Issue #201, opened 2023-07-21
Rename db.client.connections.usage to db.client.connections.count?


trask:
This seems more consistent with:

* `system.processes.count`

(though also could be `db.client.connections` based on outcome of #50)

trask:
Examples of `.count`:

* `process.runtime.jvm.threads.count`
* `system.processes.count`

Examples of `.usage`:

* `db.client.connections.usage`
* `system.filesystem.usage` (`Bytes`)
* `system.memory.usage` (`Bytes`)
* `system.paging.usage`  (`Bytes`)
* `process.runtime.jvm.memory.usage` (`Bytes`)
* `process.runtime.jvm.memory.usage_after_last_gc` (`Bytes`)
* `hw.gpu.memory.usage` (`Bytes`)
* `hw.logical_disk.usage` (`Bytes`)

Proposal: recommend `.usage` when units are `Bytes`, otherwise recommend `.count`

lmolkova:
Related: #703

gregkalapos:
I'd be happy to create a PR for this.

PR #384, opened 2023-10-11
BREAKING: Use seconds as default duration for FaaS duration histograms

pyohannes:
Fixes #382

## Changes

[General guidelines for metric instrumentation units](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#instrument-units) require that for measuring durations, seconds should be used:

> When instruments are measuring durations, seconds (i.e. s) SHOULD be used.

This PR brings the FaaS metrics `faas.invoke_duration`, `faas.init_duration`, and `faas.cpu.usage` in line with this requirement.

As I couldn't find a place where those metrics are currently implemented or used, this change is proposed without recourse to any compatibility mechanism.

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x] [CHANGELOG.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md) updated for non-trivial changes.
* [x] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

Issue #381, opened 2023-10-11
Use seconds as default duration for DB client metrics


pyohannes:
This issue was created based on a discussion about #176.

[General guidelines for metric instrumentation units](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#instrument-units) require that for measuring durations, seconds should be used:

> * When instruments are measuring durations, seconds (i.e. s) SHOULD be used.

However, the following DB client metrics still use milliseconds for measuring durations:
* [`db.client.connections.create_time`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#metric-dbclientconnectionscreate_time)
* [`db.client.connections.wait_time`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#metric-dbclientconnectionswait_time)
* [`db.client.connections.use_time`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#metric-dbclientconnectionsuse_time)

For consistency, those histograms should be changed to use seconds. Those metrics are already in use, and to avoid unnecessary disruptions the change should happen during stabilization of DB semantic conventions and should be covered by the `OTEL_SEMCONV_STABILITY_OPT_IN` mechanism.

trask:
the initial database semantic convention stability isn't targeting these metrics, but it makes sense to me to make these changes as part of the stability effort so that they can be grouped with the other database semconv breaking changes

jack-berg:
> the initial database semantic convention stability isn't targeting these metrics, but it makes sense to me to make these changes as part of the stability effort so that they can be grouped with the other database semconv breaking changes

I agree. Consumers of db semantic conventions will already be dealing with significant breaking changes. Withholding changes to part of the conventions because they're not targeted for stability is probably detrimental to the user experience.

gregkalapos:
As suggested by @trask https://github.com/open-telemetry/semantic-conventions/pull/966#pullrequestreview-2025811321 doing it in https://github.com/open-telemetry/semantic-conventions/pull/966

PR #966, opened 2024-04-24
Rename `db.client.connections.usage` to `db.client.connections.count`

gregkalapos:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/201 

## Changes

As described in #201: `db.client.connections.usage` ->  `db.client.connections.count` 

### DIFF
--- a/docs/database/database-metrics.md
+++ b/docs/database/database-metrics.md
@@ -154,13 +154,13 @@ The following metric instruments describe database client connection pool operat
 
 This metric is [required][MetricRequired].
 
-<!-- semconv metric.db.client.connections.usage(metric_table) -->
+<!-- semconv metric.db.client.connections.count(metric_table) -->
 | Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
 | -------- | --------------- | ----------- | -------------- | --------- |
-| `db.client.connections.usage` | UpDownCounter | `{connection}` | The number of connections that are currently in state described by the `state` attribute | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `db.client.connections.count` | UpDownCounter | `{connection}` | The number of connections that are currently in state described by the `state` attribute | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 <!-- endsemconv -->
 
-<!-- semconv metric.db.client.connections.usage(full) -->
+<!-- semconv metric.db.client.connections.count(full) -->
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
 | [`db.client.connections.pool.name`](../attributes-registry/db.md) | string | The name of the connection pool; unique within the instrumented application. In case the connection pool implementation doesn't provide a name, instrumentation should use a combination of `server.address` and `server.port` attributes formatted as `server.address:server.port`. | `myDataSource` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
--- a/docs/database/database-metrics.md
+++ b/docs/database/database-metrics.md
@@ -18,7 +18,7 @@ and attributes but more may be added in the future.
 - [Database operation](#database-operation)
   - [Metric: `db.client.operation.duration`](#metric-dbclientoperationduration)
 - [Connection pools](#connection-pools)
-  - [Metric: `db.client.connections.usage`](#metric-dbclientconnectionsusage)
+  - [Metric: `db.client.connections.count`](#metric-dbclientconnectionscount)
   - [Metric: `db.client.connections.idle.max`](#metric-dbclientconnectionsidlemax)
   - [Metric: `db.client.connections.idle.min`](#metric-dbclientconnectionsidlemin)
   - [Metric: `db.client.connections.max`](#metric-dbclientconnectionsmax)
@@ -150,7 +150,7 @@ If a database operation involved multiple network calls (for example retries), t
 
 The following metric instruments describe database client connection pool operations.
 
-### Metric: `db.client.connections.usage`
+### Metric: `db.client.connections.count`
 
 This metric is [required][MetricRequired].
 
--- a/model/metrics/database-metrics.yaml
+++ b/model/metrics/database-metrics.yaml
@@ -8,9 +8,9 @@ groups:
     stability: experimental
     extends: attributes.db.client
 
-  - id: metric.db.client.connections.usage
+  - id: metric.db.client.connections.count
     type: metric
-    metric_name: db.client.connections.usage
+    metric_name: db.client.connections.count
     stability: experimental
     brief: "The number of connections that are currently in state described by the `state` attribute"
     instrument: updowncounter
--- a/schema-next.yaml
+++ b/schema-next.yaml
@@ -4,6 +4,9 @@ versions:
   next:
     metrics:
       changes:
+        #
+        - rename_metrics:
+            db.client.connections.usage: db.client.connections.count
         # https://github.com/open-telemetry/semantic-conventions/pull/909
         - rename_attributes:
             attribute_map:
--- a/.chloggen/db_client_connection_metric.yaml
+++ b/.chloggen/db_client_connection_metric.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Rename `db.client.connections.usage` to `db.client.connections.count`.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [201]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:


model/metrics/database-metrics.yaml
```diff
@@ -8,9 +8,9 @@ groups:
     stability: experimental
     extends: attributes.db.client
 
-  - id: metric.db.client.connections.usage
```
gregkalapos, line 11: First time doing such a PR, so I rather ask here: shouldn't a deprecated metric be added with the old name? 

I see [it's done for attributes](https://github.com/open-telemetry/semantic-conventions/blob/main/model/registry/deprecated/db.yaml), but I don't see any for metrics - closest is [metric related attributes](https://github.com/open-telemetry/semantic-conventions/blob/main/model/registry/deprecated/db.yaml#L88), but that's still not the metric itself. Also looked at some similar PRs that rename a metric, and from what I have seen, the old metric name never shows up as deprecated.

gregkalapos: According to `make compatibility-check` and [CONTRIBUTING.md#version-compatibility-check](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#version-compatibility-check), `metric.db.client.connections.usage` still needs to show up.

I don't really find any example for deprecated metrics 🤔 - I added it to the end of `database-metrics.yaml`. Let me know if there is a better place for it.

model/metrics/database-metrics.yaml
```diff
@@ -108,3 +108,16 @@ groups:
     attributes:
       - ref: db.client.connections.pool.name
         requirement_level: required
+
+  - id: metric.db.client.connections.usage
+    type: metric
+    metric_name: db.client.connections.usage
+    stability: experimental
```
trask, line 115: is this supported by the yaml?

```suggestion
    deprecated: "Replaced by `db.client.connections.count`."
```

model/metrics/database-metrics.yaml
trask: good question, i'm not sure we've done this before either, let's see what @lmolkova thinks

trask (commented):
it may be worth fixing #967 in this PR also

model/metrics/database-metrics.yaml
lmolkova, line 115: yep, but in this form 
```suggestion
    stability: experimental
    deprecated: "Replaced by `db.client.connections.count`."
```

model/metrics/database-metrics.yaml
lmolkova: we never deprecated metrics so far (which was not great and we started enforcing it very recently). So it's the first metrics to be deprecated.

Let's create `/model/metrics/deprecated/database.yaml` file and add deprecated metric there?

model/metrics/deprecated/database.yaml
```diff
@@ -0,0 +1,110 @@
+groups:
+  - id: metric.deprecated.db.client.connections.count
```
gregkalapos, line 2: Is this id format ok? 

Alternatively `metric.db.deprecated.client.connections.count` seems also reasonable.

model/metrics/database-metrics.yaml
gregkalapos: Thanks @lmolkova - moved deprecated metrics to `/model/metrics/deprecated/database.yaml`. Should I also generate an `.md` file from that?

lmolkova: @jsuereth I remember you wanted to have MD for all deprecated attributes to be generated in the registry.
Do you have a strong opinion for metrics?

I have a soft preference to not generate experimental-deprecated metrics in the MD to reduce noise while keeping codegen intact.

docs/database/database-metrics.md
```diff
@@ -18,7 +18,7 @@ and attributes but more may be added in the future.
 - [Database operation](#database-operation)
   - [Metric: `db.client.operation.duration`](#metric-dbclientoperationduration)
 - [Connection pools](#connection-pools)
-  - [Metric: `db.client.connections.usage`](#metric-dbclientconnectionsusage)
+  - [Metric: `db.client.connections.count`](#metric-dbclientconnectionscount)
   - [Metric: `db.client.connections.idle.max`](#metric-dbclientconnectionsidlemax)
   - [Metric: `db.client.connections.idle.min`](#metric-dbclientconnectionsidlemin)
   - [Metric: `db.client.connections.max`](#metric-dbclientconnectionsmax)
```
trask, line 24: `connections` -> `connection`

model/metrics/database-metrics.yaml
```diff
@@ -8,9 +8,9 @@ groups:
     stability: experimental
     extends: attributes.db.client
 
-  - id: metric.db.client.connections.usage
+  - id: metric.db.client.connections.count
     type: metric
-    metric_name: db.client.connections.usage
+    metric_name: db.client.connections.count
```
trask, line 13: `connections` -> `connection`

trask (commented):
@gregkalapos sorry, I realized it would be good to add one more change to this PR: #381

bundling #381 here would have the advantage that we know the old name is always associated with milliseconds and the new name is always associated with seconds

(and it would affect the schema translation, where the metrics that changed from milliseconds to seconds would no longer be considered renames)

model/metrics/deprecated/database.yaml
lmolkova: We follow the <name>.deprecated pattern for attribute groups in the registry, so I guess 
`metric.db.client.connections.count.deprecated`? Either way works for me.

gregkalapos: Thanks. Let's follow the `.deprecated` pattern that attribute groups already use. I adapted the PR.

trask (approved):

schema-next.yaml
```diff
@@ -4,6 +4,14 @@ versions:
...
     metrics:
       changes:
+        # https://github.com/open-telemetry/semantic-conventions/pull/966
+        - rename_metrics:
+            db.client.connections.usage: db.client.connection.count
+            db.client.connections.idle.max: db.client.connection.idle.max
+            db.client.connections.idle.min: db.client.connection.idle.min
+            db.client.connections.max: db.client.connection.max
+            db.client.connections.pending_requests: db.client.connection.pending_requests
+            db.client.connections.timeouts: db.client.connection.timeouts
```
trisch-me, line 14: these are not all deprecated metrics, it should be 9 entries according to the file above

trask: 3 of them are not renames since the unit was changed from `ms` to `s`

model/metrics/database-metrics.yaml
lmolkova: based on the offline discussion, we decided to skip md for deprecated metrics now and add it later if it's ever necessary

lmolkova (approved):

schema-next.yaml
gregkalapos: So the tooling forces me to add all previous metric names in one of the yaml files - that's why we have all 9 in [deprecated/database.yaml](https://github.com/open-telemetry/semantic-conventions/pull/966/files#diff-1d811a06d82d206d944c19019884c092fcf92747f1cf6f014f3cdcd5e1f6c1aa). As I learned in a thread above, this is a recent change, previously this wasn't enforced.

And as @trask says, 6 are just renames, the other 3 change both the unit and the metric name. I wonder if this difference should be visible in [deprecated/database.yaml](https://github.com/open-telemetry/semantic-conventions/pull/966/files#diff-1d811a06d82d206d944c19019884c092fcf92747f1cf6f014f3cdcd5e1f6c1aa). From what I know there is no syntax for just renaming there, right? Only think we can do there is to mark it with `deprecated:`.

trisch-me: But they are renames as well? we do change from `db.client.connections.use_time` to `db.client.connection.use_time` and In addition to this we also do change the unit. 

In this case we shouldn't add it to the `rename_metrics` in schema?

model/metrics/deprecated/database.yaml
```diff
@@ -0,0 +1,110 @@
...
+    attributes:
+      - ref: db.client.connections.pool.name
+        requirement_level: required
+
+  - id: metric.db.client.connections.use_time.deprecated
+    type: metric
+    metric_name: db.client.connections.use_time
+    stability: experimental
+    deprecated: "Replaced by `db.client.connection.use_time`."
+    brief: "Deprecated, use `db.client.connection.use_time` instead."
```
trisch-me, line 105: I think we need to update description of deprecated items to announce unit change as well

schema-next.yaml
gregkalapos: I looked at some of the PRs (not all of them) that added `rename_metrics` - those seem to be purely renames. If the unit changes, I'd interpret it more than a rename.

I don't have full context on how schema translation works, but my guess is that everything that is under `rename_metrics`, could be just renamed and be 100% compatible with the new schema. If so, then the ones where we change the unit should not be under `rename_metrics` - they should be kept as is and then be marked as deprecated. In theory there could be a correct translation for rename+unit change, but from what I know, there is no syntax for that and that would still be different from just renaming.

> In this case we shouldn't add it to the rename_metrics in schema?

So, unfortunately I'm not sure, so I'd defer to @trask and @lmolkova.

gregkalapos:
CI fails, `dotnet` doesn't ❤️  me:

```
Run dotnet tool update -g docfx --version "3.0.0-*" --add-source https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-public-packages/nuget/v3/index.json
/tmp/837413a5-82a5-4ef9-abd4-39c1160faf4a/restore.csproj : error NU1102: Unable to find package docfx with version (>= 3.0.0)
/tmp/837413a5-82a5-4ef9-abd4-39c1160faf4a/restore.csproj : error NU1102:   - Found 224 version(s) in nuget.org [ Nearest version: 2.76.0 ]
/tmp/837413a5-82a5-4ef9-abd4-39c1160faf4a/restore.csproj : error NU1102:   - Found 0 version(s) in https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-public-packages/nuget/v3/index.json
Tool 'docfx' failed to update due to the following:
The tool package could not be restored.
Tool 'docfx' failed to install. This failure may have been caused by:

* You are attempting to install a preview release and did not use the --version option to specify the version.
* A package by this name was found, but it was not a .NET tool.
* The required NuGet feed cannot be accessed, perhaps because of an Internet connection problem.
* You mistyped the name of the tool.

For more reasons, including package naming enforcement, visit https://aka.ms/failure-installing-tool
Error: Process completed with exit code 1.
```

schema-next.yaml
jsuereth: We will need to  add a new "change units" schema transformation to schema files for a unit change.

cc @lmolkova @trask

lmolkova: Created https://github.com/open-telemetry/opentelemetry-specification/issues/4028, I don't think we need to block this PR on this since it's an experimental metric and we don't have schema transformation in practice.

pyohannes: There already existed a similar issue: https://github.com/open-telemetry/opentelemetry-specification/issues/3729

Also, we did unit changes for experimental metrics in the past (for HTTP, and for [FaaS](https://github.com/open-telemetry/semantic-conventions/pull/384)), I think it's fine as long as it's part of stabilization effort.

However, I wonder whether it should be covered by some backward compatibility mechanisms like `OTEL_SEMCONV_STABILITY_OPT_IN`, or whether implementations switching units are supposed to do it via "hard" breaking changes?

Issue #922, opened 2024-04-15
Add k8s.container.status.last_terminated_reason to semantic conventions.


povilasv:
### Area(s)

area:k8s

### Is your change request related to a problem? Please describe.

We recently added `k8s.container.status_last_terminated_reason` to kubernetes cluster receiver. And we would like to add resource attribute ~`k8s.container.status_last_terminated_reason`~ to existing semantic conventions.

**Edit: It was agreed that this attribute is better named  `k8s.container.status.last_terminated_reason`** 

Reference https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31505#discussion_r1521881960

Also issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/31282

### Describe the solution you'd like

Add resource attribute `k8s.container.status_last_terminated_reason` to existing semantic conventions.

joaopgrassi:
Given the example and structure is complex for `status`

https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/31282#issue-2136401464

kubectl get pod X -o yaml

```
apiVersion: v1
kind: Pod
...
status:
  containerStatuses:
  - containerID: containerd://07ff7db2706d20ddd26c6257c1c1dbf176917f855c7feef0e05db88159b1584d
    image: image
    imageID:  imageId
    lastState:
      terminated:
        containerID: containerid
        exitCode: 2
        finishedAt: "2024-02-05T23:17:38Z"
        reason: Error
        startedAt: "2024-02-05T23:17:33Z"
```

I think we should probably have it more structured. Maybe something like `k8s.container.status.last_terminated_reason`. Because then we could add more things under status, such as  the container id and etc. 

You also added a metric to the collector, right? This metric should also probably be added to the semconv.

joaopgrassi:
Great, then I will assign this to you if that's ok :)

ChrsMark:
The issue's tittle talks about adding `k8s.container.status_last_terminated_reason` but the [PR](https://github.com/open-telemetry/semantic-conventions/pull/968) adds the `k8s.container.status.last_terminated_reason` which is of course aligned with the [Collector's implementation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d291a327f78201a98d7ce3d545916f3d15f0a2b7/receiver/k8sclusterreceiver/documentation.md?plain=1#L431).

[...]

PR #968, opened 2024-04-25
feat: add k8s.container.status.last_terminated_reason resource attribute

povilasv:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/922

### DIFF
--- a/.chloggen/k8s-status.yaml
+++ b/.chloggen/k8s-status.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: "k8s"
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: add container.status.last_terminated_reason resource attribute
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [922]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/model/registry/k8s.yaml
+++ b/model/registry/k8s.yaml
@@ -98,7 +98,12 @@ groups:
           Number of times the container was restarted. This attribute can be
           used to identify a particular container (running or stopped) within a
           container spec.
-        examples: [0, 2]
+      - id: container.status.last_terminated_reason
+        type: string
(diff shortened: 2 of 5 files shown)


model/registry/k8s.yaml
```diff
@@ -98,7 +98,12 @@ groups:
           Number of times the container was restarted. This attribute can be
           used to identify a particular container (running or stopped) within a
           container spec.
-        examples: [0, 2]
+      - id: container.status.last_terminated_reason
+        type: string
+        stability: experimental
+        brief: >
+          Last terminated reason of the Container.
+        examples: ["Evicted", "Error"]
```
ChrsMark, line 106: nit: Should we specify an enum with the possible values if there is such a list?

ChrsMark: My point is that if we know the possible values this attribute can take we can specify them explicitly like what we do in `cpu.state`s for example: https://github.com/open-telemetry/semantic-conventions/blob/4d80b4ce2b17c84856f39c73cef41ae54d087d32/model/registry/system.yaml#L19-L43

TylerHelmuth (approved):

model/registry/k8s.yaml
ChrsMark: I see, thank's for clarifying!

mx-psi (approved):

ChrsMark:
FYI @open-telemetry/specs-semconv-approvers @open-telemetry/specs-semconv-maintainers

.chloggen/k8s-status.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
```
lmolkova, line 7: I believe it should be `enhancement`, `new_component` is for new root namespaces

model/resource/k8s.yaml
```diff
@@ -45,6 +45,8 @@ groups:
     attributes:
       - ref: k8s.container.name
       - ref: k8s.container.restart_count
+      - ref: k8s.container.status.last_terminated_reason
+        requirement_level: opt_in
```
lmolkova, line 49: any reason for it to be opt-in? does it have sensitive info, is it expensive to collect, is it too verbose? 
If not, I'd consider `recommended` so users can leverage it by default and instrumentations wouldn't need to add config options.

lmolkova (approved):
Left a couple of minor comments

model/resource/k8s.yaml
TylerHelmuth: I am good with `recommended` as I don't feel high cardinality (or in this case the potential of high cardinality) should prevent an attribute from being recommended. I think in practice there will be a pretty small set of values.

ChrsMark: > it might cause high cardinality

Is this true 🤔? Do we have an indicator of how many different values this attribute could come with?

I would expect what @TylerHelmuth already mentioned:
> in practice there will be a pretty small set of values

PR #970, opened 2024-04-26
Recording exception as standalone events

scheler:
## Changes

Added semantic convention for recording exception on standalone events.

NOTE on when to use what for recording exceptions:
1. When exception occurs in the context of a span, it MAY be recorded as a Span Event with name `exception` and attributes `exception.*`.
2. When Log Bridge API is used to map exceptions from logs to OpenTelemetry LogRecords, the exception MAY be recorded as a LogRecord using the `exception.*` attributes.
3. In other cases, the exception MAY be recorded as an Event with name `exception` and attributes `exception.*`.

--

Note: if the PR is touching an area that is not listed in the [existing areas](https://github.com/open-telemetry/semantic-## Merge requirement checklist


docs/exceptions/exceptions-events.md
```diff
@@ -0,0 +1,48 @@
...
+
+# Semantic Conventions for Exceptions in Logs
+
+**Status**: [Experimental][DocumentStatus]
+
+This document defines semantic conventions for recording exceptions on
+[events](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.31.0/specification/logs/event-api.md#emit-event)
+emitted through the [EventLogger](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md#eventlogger) API.
+
+
```
martinkuba, line 13: nit: two blank lines, probably failing lint

docs/exceptions/exceptions-events.md
```diff
@@ -0,0 +1,48 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Events
+--->
+
+# Semantic Conventions for Exceptions in Logs
```
martinkuba, line 5:
```suggestion
# Semantic Conventions for Exceptions in Events
```

martinkuba:
I feel like we may need to document somewhere the difference of capturing exceptions as events vs logs. IMO, it is clear that "event exceptions" are the same as span exceptions - both captured as events with name `exception`. Log exceptions, I think, are captured only from logging frameworks as ERROR logs? Should backends interpret event exceptions and log exceptions as the same thing, or is it important to note the difference?

docs/exceptions/exceptions-logs.md
```diff
@@ -52,4 +56,13 @@ The table below indicates which attributes should be added to the
 Same as [Trace Semantic Conventions for Exceptions - Stacktrace
 Representation](exceptions-spans.md#stacktrace-representation).
 
+## Guidelines for Recording Exceptions in Spans, Logs, and Events
```
reyang, line 59: Is this part intended to be Experimental or Stable for now?

scheler: Recording of exceptions in events is experimental as it's being added only now.

trask: here's an example: https://github.com/open-telemetry/semantic-conventions/pull/989/files#diff-e1471b4097d9b5e1547732314a879b440712be6e2aa5d8b61013161018cafd90R7

docs/exceptions/exceptions-logs.md
```diff
@@ -52,4 +56,13 @@ The table below indicates which attributes should be added to the
...
+
+When it comes to recording exceptions, it's crucial to distinguish between Spans, Logs, and Events:
+
+* Spans and Events: Exception recording in Spans and Events follows a similar pattern. Both utilize the event name `exception`. Exceptions MAY be recorded within Span Events if they occur during the span's lifecycle.
+* Logs: Exceptions SHOULD be recorded in logs when utilizing the log bridge API to map application logs to OpenTelemetry logs.
+
+Note that in all cases the exception data is stored in attributes.
```
lmolkova, line 66: Can we entertain the idea of recording stack trace in the payload and not in the attribute to avoid all the common problems (parsing, truncation, etc) ?

scheler: @lmolkova  Can you elaborate a bit on how recording the stacktrace in the Event Payload (LogRecord Body) address those issues? Even within the `any` type field, the data would still have to be put into the standard primitive types.

trask: Check out https://github.com/open-telemetry/opentelemetry-specification/issues/2839

scheler: @trask thanks for the pointer. I guess we were taking the simplistic approach of keeping the `exception` event consistent with `exception` spans and `exception` logs. However, I can see that moving the exception data to Event body/payload would allow us to have richer data in the Event. For eg., in addition to the structured vs plaintext stacktrace representation, we could add `context` or `additonal_data` to the Event to capture arbitrary objects which is not possible with attributes which only allow primitive types currently.

docs/exceptions/exceptions-logs.md
```diff
@@ -52,4 +58,13 @@ The table below indicates which attributes should be added to the
 Same as [Trace Semantic Conventions for Exceptions - Stacktrace
 Representation](exceptions-spans.md#stacktrace-representation).
 
+## Guidelines for Recording Exceptions in Spans, Logs, and Events
+
+When it comes to recording exceptions, it's crucial to distinguish between Spans, Logs, and Events:
+
+* Spans and Events: Exception recording in Spans and Events follows a similar pattern. Both utilize the event name `exception`. Exceptions MAY be recorded within Span Events if they occur during the span's lifecycle.
```
arminru, line 65:
```suggestion
* Spans and Events: Exception recording in Spans and Events follows a similar pattern. Both utilize the event name `exception`. Exceptions SHOULD be recorded within Span Events if they occur during the span's lifecycle.
```

Is there any reason why we'd diverge from the stable spec statement that makes this a SHOULD here in semconv?
https://opentelemetry.io/docs/specs/otel/trace/exceptions/

CHANGELOG.md
```diff
@@ -6,12 +6,15 @@
...
 
 <!-- next version -->
+
 ## Unreleased
 
 ### Breaking
 
 ### Features
 
+- Add semantic conventions for recording exceptions in events [#970](https://github.com/open-telemetry/semantic-conventions/pull/970)
```
trisch-me, line 16: We are not updating changelog directly. Please check requirements in the description of your PR

trisch-me (changes requested):
please add changelog entry correctly

PR #961, opened 2024-04-24
URL query string values should be redacted by default

trask:
Fixes #860

## Changes

Query string values are now redacted by default due to concerns around leaking sensitive data.

This is not considered a breaking change because the [OTel semantic conventions definition of stability](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#semantic-conventions-stability) specifically allows changing attribute values:

> Things not listed in the above are not expected to remain stable via semantic convention and are allowed (or expected) to change. A few examples:
>
> * The values of attributes

Issue #860, opened 2024-03-29
Http client and server span default collection behavior for `url.full` and `url.query` attributes


vishweshbankwar:
### Area(s)

area:url

### Is your change request related to a problem? Please describe.

The HTTP client and server spans have `url.full` and `url.query` attributes, respectively, tagged as **Required**. These attributes can contain sensitive information, as mentioned in the spec. Even though the spec states, `Sensitive content provided SHOULD be scrubbed when instrumentations can identify it`, it isn't possible to identify this in a deterministic way.

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md

This is an issue because users can onboard to the instrumentation library without being aware of the data collected by the library, resulting in a leak of sensitive information.

### Describe the solution you'd like

The attributes that have the possibility to contain sensitive information are not collected by default. Changing this now will be **breaking** change, an alternate is to have a consistent opt-out option defined in the specification to opt out of the default behavior. 

### Describe alternatives you've considered

As a mitigation, users can provide processors to scrub information. However this is likely to be a reactive thing, we should have a safe-by-default stance instead.

### Additional context

Related issue: https://github.com/open-telemetry/semantic-conventions/issues/128

lmolkova:
There is nothing officially documented, but I believe this https://github.com/open-telemetry/semantic-conventions/issues/128#issuecomment-1658809083 and https://github.com/open-telemetry/semantic-conventions/issues/128#issuecomment-1775594259 summarize the approach we follow now:
- it's impossible to prevent PII/sensitive data collection on the instrumentation level without compromising observability usefulness
- it cannot be an instrumentation author concern to scrub sensitive data and be aware of GDPR and other potentially relevant  concerns.
- we're missing:
  - a way to mark some attributes as ones that's more likely to contain sensitive/pii data
  - central SDK/collector component that'd do scrubbing to comply with GDPR and or other requirements relevant to a certain application

None of this is specific to `url.full`/`url.query` and if we stopped collecting those, it'd be more than just a breaking change - it'd render http client tracing useless.

reyang:
Might be worthy to have dedicated columns for each attribute indicating the potential concerns from privacy/security perspective. For example:

| attribute name | privacy notes | security notes |
| -------------- | ------------- | -------------- |
| client.address | this field could be a privacy concern especially when the client.address is coming from the end-user / consumer | there is no known security concern |
[...]

PR #971, opened 2024-04-26
Specific URL query string values should be redacted

trask:
Related to #860

Alternative proposal to #961

## Changes

Specific URL query string values should now be redacted by default due to concerns around leaking credentials.

The list of problematic query string keys is documented in the semantic conventions and is subject to change over time as we learn about additional common usages of credentials found in query string values.


austinlparker:
Should we also add a few 'obvious' ones? e.g., 'password', 'secret', 'private', etc.?

trask:
here's the default list used by Elastic APM: https://www.elastic.co/guide/en/apm/agent/java/1.x/config-core.html#config-sanitize-field-names

lmolkova (approved):
I believe we need to let users/sdks/distros to expand the blocklist via configuration, but this would be a purely incremental change.

austinlparker:
> here's the default list used by Elastic APM: https://www.elastic.co/guide/en/apm/agent/java/1.x/config-core.html#config-sanitize-field-names

I do find it ironic that there's a big "query string values are considered non-sensitive" info box in that section. 

But yeah, that's a good list of additions too.

tylerbenson (approved):
I think this should be easily configured in a standard way so users can add more to it.

I'm also surprised that this list is as short as it is.

docs/attributes-registry/url.md
```diff
@@ -26,9 +26,29 @@ linkTitle: URL
...
+
+Sensitive content provided in `url.full` SHOULD be scrubbed when instrumentations can identify it.
+
+In particular, query string values for the following keys SHOULD be redacted by default and replaced by the
+value `REDACTED`:
+
+* `AWSAccessKeyId`
+* `Signature`
+* `sig`
+* `X-Goog-Signature`
```
tylerbenson, line 45: Should we give some added context as to why these were added?  What was the decision process and what criteria should be used in the future to evaluate if something new should be added?

gregkalapos:
I'd definitely prefer this over https://github.com/open-telemetry/semantic-conventions/pull/961, since this is less intrusive for users who already capture query strings and unlike https://github.com/open-telemetry/semantic-conventions/pull/961 this really handles values that are problematic from security point of view. With https://github.com/open-telemetry/semantic-conventions/pull/961 the only chance for people to capture query strings is to turn it back on, but that again would capture problematic values as well. So I definitely prefer an approach that tries to handle problematic values over an all or nothing config.

Since configuration came up: I think this list could be generalized and applied to multiple other things (e.g. Db query parameters, Elastic also applies this to x-form encoded request bodies (not yet part of SemConv, but could be), and there are probably many other attribtues). So the config could be a list of keys that are applied for sanitization to every attribute that can potentially contain sensitive information, not only for query strings. That could be also introduced incrementally; for now I think what the PR proposes is ok.

reyang:
> I'm also surprised that this list is as short as it is.

If we look at what the user provided in the [original report](https://github.com/open-telemetry/opentelemetry-dotnet/security/advisories/GHSA-vh2m-22xx-q94f), here are some keys:

_It is important to ensure that sensitive query parameters are scrubbed by default. While this list is not exhaustive, the following parameters should be scrubbed as a starting point: "sig", "code", "key", "access_token", "client_secret", "password", "appkey", "app_key", "apikey", "apiKey", "api_key", "accesskey", "access_key", and "token"._


image

docs/attributes-registry/url.md
reyang: +1 on put some clarification around the criteria (doesn't have to be perfect in this PR though).

I have a general concern regarding the process, for example, if this list got updated everyday, are we going to play the catch-up game from all language implementations, and do we publish security advisories and deprecate previous versions?

gregkalapos (approved):

docs/attributes-registry/url.md
trask: I've added links now from each of these query string keys to the public rest APIs that use them. This is the initial criteria I'm thinking of: _known_ query string keys that are used by _known_ public rest APIs.

trask:
There are many potential candidates for additional query string keys to include.

From initial CVE report against OpenTelemetry .Net:

* `code`
* `key`
* `access_token`
* `client_secret`
* `password`
* `appkey`
* `app_key`
* `apikey`
* `apiKey`
* `api_key`
* `accesskey`
* `access_key`
* `token`

From https://www.elastic.co/guide/en/apm/agent/java/1.x/config-core.html#config-sanitize-field-names:

* `password`
* `passwd`
* `pwd`
* `secret`
* `*key`
* `*token*`
* `*session*`
* `*credit*`
* `*card*`
* `*auth*`
* `*principal*`
* `set-cookie`

From a process perspective of how to pick and choose (both now and in the future), it might be nice to say something along the lines of

> our default redaction list only includes _known_ query string keys used to pass credentials to _known_ public rest APIs

docs/attributes-registry/url.md
```diff
@@ -26,9 +26,29 @@ linkTitle: URL
...
+
+In particular, query string values for the following keys SHOULD be redacted by default and replaced by the
+value `REDACTED`:
+
+* [`AWSAccessKeyId`](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth)
+* [`Signature`](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth)
+* [`sig`](https://learn.microsoft.com/en-us/azure/storage/common/storage-sas-overview#sas-token)
+* [`X-Goog-Signature`](https://cloud.google.com/storage/docs/access-control/signed-urls)
+
+This list is subject to grow over time, but once a key is added to the list, removing it will be considered a
```
reyang, line 47: Why would this be considered as breaking changes?
For example, items in the current list have links to the official doc, one can imagine over years some of these technologies will be abandoned and docs would be retired.

svrnm (approved):

svrnm:
I prefer this solution over #961, 👍 

Question: In [my comment](https://github.com/open-telemetry/semantic-conventions/pull/961#issuecomment-2078047508) and in [@austinlparker's comment](https://github.com/open-telemetry/semantic-conventions/pull/961#issuecomment-2079628921) and in [@reyang's comment](https://github.com/open-telemetry/semantic-conventions/pull/961#issuecomment-2079761854) (in chronological order), we discussed the idea of having generalized SDK capabilities for redaction, is this still something we want? If so I am happy to work towards a proposal for that.

pyohannes:
> we discussed the idea of having generalized SDK capabilities for redaction, is this still something we want? If so I am happy to work towards a proposal for that.

I'd definitely answer that with a yes. Users might want to have redaction capabilities they can use for data that doesn't conform to OTel semantic conventions.

There have been related previous attempts:
* https://github.com/open-telemetry/oteps/pull/187
* https://github.com/open-telemetry/oteps/pull/100

TylerHelmuth:
> our default redaction list only includes known query string keys used to pass credentials to known public rest APIs

I agree with adding this verbiage as it helps clarify how the list was decided.

austinlparker (approved):

svrnm:
As discussed during the SemConv SIG meeting I am going to provide a proposal for the "general problem", I will try to have a draft by the end of this week to share

svrnm:
> As discussed during the SemConv SIG meeting I am going to provide a proposal for the "general problem", I will try to have a draft by the end of this week to share

Here's the proposal: https://github.com/open-telemetry/oteps/pull/255

MSNev (approved):

trask:
I've added this PR to next week's semantic convention SIG agenda to discuss resurrecting it given that https://github.com/open-telemetry/oteps/pull/255 has stalled

svrnm:
> I've added this PR to next week's semantic convention SIG agenda to discuss resurrecting it given that [open-telemetry/oteps#255](https://github.com/open-telemetry/oteps/pull/255) has stalled

Thanks, @trask! commenting on the OTEP and why it stalled: I tried to rework it but there are so many moving pieces that I couldn't get a good handle of it, while also running into some bandwidth issues on my end, since other things required my attention. I still want to follow up on this, but it will take much more time than anticipated. So I am in favor with moving forward with this PR 👍

svrnm (approved):

jmacd (approved):

docs/attributes-registry/url.md
trask: Resolving with https://github.com/open-telemetry/semantic-conventions/pull/971/commits/853c4851be738792f8cd5d409e049e2f78498d84

CEbbinghaus:
One thing everybody is instantly keen on addressing is more and more keys or ways to include keys, But I would like to point out that it is equally important to allow developers to explicitly allow certain query parameters. Ultimately I think that Otel should give the developer implementing it the ultimate control of what is and isn't redacted as there will always be edge cases in either direction that cannot be addressed generically

PR #345, opened 2023-09-26
feat: Add MySQL instance address attribute

arielvalentin:
Fixes # N/A

## Changes

At GitHub we run MySQL in multi-node clusters and are often interested in knowing the hostname of a node that is executing a query or command.

In order to acheive this we added an attribute to the Trilogy instrumentation named `db.mysql.instance.address` which is populated with the hostname of the node:

<https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/487>

This change adds the attribute to the database model as a MySQL specific connection attribute.

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] [CHANGELOG.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md) updated for non-trivial changes.
* [ ] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

PR #972, opened 2024-04-26
Remove `db.instance.id`

trask:
## Changes

Removes `db.instance.id` and reverts the Elasticsearch usage of it to `db.elasticsearch.node.name`.

## Reason

In many database systems, the idea of `db.instance.id` corresponds to `server.address` or `network.peer.address`.

After adding `db.namespace`, we're currently only using `db.instance.id` in one place, and so I think it may be too early to try to add this generalized attribute.

Also, "instance id" is a potentially confusing name given the different meanings it has in SQL Server and Oracle.


lmolkova:
I didn't find any usage of `db.instance.id` in existing instrumentation libraries in otel repos.

lmolkova (approved):

trisch-me:
should we update schema as well?

trask:
> should we update schema as well? 

Can you clarify what you had in mind? Thanks

jack-berg (approved):

reyang (approved):

arielvalentin:
@trask @lmolkova I proposed this attribute and use it here. 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/743796d425de3aa6c13fd89dffb3ad869106d261/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb#L76

The reason why `server.address` or `net.peer.name` at the time did not work for us was discussed here: https://github.com/open-telemetry/semantic-conventions/pull/345

I suppose we should follow the pattern that elasticsearch uses and have a separate name under db.mysql then?

lmolkova:
thanks for the context @arielvalentin ! 

I'd suggest using `server.address` for cluster address and `network.peer.address` for the individual node. You can provide either one or both of them together on one span. 

In case of mySQL I assume what you have there 
 as [`@connected_host`](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/743796d425de3aa6c13fd89dffb3ad869106d261/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb#L76) is the network address of the mysql node? 

Looking at this comment https://github.com/open-telemetry/semantic-conventions/pull/345#discussion_r1337175786 it seems it is

> The configured hostname would appear as `server.address: mysql-db.example.com` vs node instance `db.mysql.instance.address: mysql-e26b99z.example.com` 

And then `network.peer.address` is a good fit:
- `server.address: mysql-db.example.com` 
- `network.peer.address: mysql-e26b99z.example.com` 

You may populate either or both of them for single-db deployments, just in a consistent way.

I believe elastic uses `db.elasticsearch.node.name` because it's not an address. 

PS: we don't always pay attention to comments on the closed issues/PRs, so please don't hesitate to create new issues.

arielvalentin:
Understood. We're very behind on semconv and haven't started work on introducing stability opt-in flags so I've been reluctant to start to migrate individual attributes over so we can avoid being in a more unspecified state than we already are. 

I'll see what we can do next quarter to address stability migrations.

PR #743, opened 2024-02-14
Database span name clarification and improvements

lmolkova:
Fixes #704

## Changes

- Clarifies span name patterns for databases: 
  - `db.statement` (if low-cardinality) -> `db.name db.statement` (order matters because `ShopDb SELECT * FROM orders WHERE order_id = ?` is better than `SELECT * FROM orders WHERE order_id = ? ShopDb`
  - `<db.operation> <db.name>.<db.sql.table>` -> `{db.name}.{table name} {operation}` if no statement (or of a high cardinality)
- Updates span names for some individual systems to follow the same pattern
  - also adds `db.sql.table` to mssql - at least in java it's set for everything JDBC and makes sense on mssql
- Fallback to `db.system` if nothing else is available

To clarify:
- [ ] case normalization (`SELECT * FROM foo` vs `sElEcT fROm FOO`)
- [ ] elastic search uses <endpoint id> - can/should we change?
- [ ] hbase - https://github.com/open-telemetry/semantic-conventions/issues/742
- [ ] aws dynamodb - not clear what should be used as a span name
- [ ] CouchDB - probably has other attributes, currently doing what java instrumentation does
- [ ]  Redis uses db index as database name, but index looks bad in the span names

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [ ] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

PR #974, opened 2024-04-26
Database: update span name

### LINKED ISSUE #704: Better guidance on semantic conventions for database client call span names in case of missing information
When tracing an activity from Npgsql (.NET Database provider for PostgreSQL) we want to set a valid span name according to the semantic conventions for database client calls without having to parse the database statement on the client side and without having to guess (or rather infer from our knowledge on the current default behavior of PostgreSQL) information that we don't have when issuing a database client call.

**What did you expect to see?**
Guidance on creating a valid span name in cases where all three of `db.operation`, `db.name` and `db.sql.table` are not available to the database client.
Per discussion following https://github.com/npgsql/npgsql/issues/4757#issuecomment-1316724458 we think that it is probably pretty safe to infer the database name from our knowledge on PostgreSQL's default behavior but we may still want to research/discuss what good alternatives could be.

**Additional context.**
https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/database/
https://github.com/npgsql/npgsql/issues/4757#issuecomment-1316599424

lmolkova:
Fix #704

(Resurrecting #743)

## Changes

- clarifies span name format in a format similar to HTTP
- changes the name format from `{db.operation.name} {db.namespace}.{db.collection.name}` to  `{db.operation.name {db.collection.name}`. the motivation:
  - `db.namespace` is the current one associated with connection, but the collection name is the first one that appears in the query.  So the collection does not necessarily belong to that database namespace.
  - `db.collection.name` can be fully qualified and then `{db.namespace}.{db.collection.name}` will have namespace twice (potentially different namespaces, see prev point)
- provides a chain of fallbacks in case information is not available


docs/database/elasticsearch.md
```diff
@@ -95,7 +95,7 @@ Tracing instrumentations that do so, MUST also set `http.request.method_original
 
 | Key                                 | Value                                                                                                                               |
 |:------------------------------------|:------------------------------------------------------------------------------------------------------------------------------------|
-| Span name                           | `"search"`                                                                                                                          |
+| Span name                           | `"search elasticsearch.mydomain.com:9200"`                                                                                          |
```
gregkalapos, line 98: I'd suggest not to change the span name for the default case for Elasticsearch. `search` is the name of the endpoint id, so that is a fairly meaningful span name (it indicates that this was a standard sync search query). The [current spec for Elasticsearch already defines this](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/elasticsearch.md#span-name) - so I'd keep it.

Few other example span name from the current spec: `async_search.get`, `async_search.delete` `bulk`. I think it's worth deviating from the general span name scheme in case of Elasticsearch.

docs/database/database-spans.md
```diff
@@ -51,21 +51,31 @@ linkTitle: Client Calls
...
 For example, if a transient issue happened and was retried within this database call, the corresponding span should cover the duration of the logical operation
 with all retries.
 
+## Name
+
+Database spans MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.31.0/specification/trace/api.md#span).
+
+<!-- markdown-link-check-disable -->
+<!-- HTML anchors are not supported https://github.com/tcort/markdown-link-check/issues/225-->
+The **span name** span names SHOULD be `{db.operation.name} {target}` if there is a
```
trask, line 64:
```suggestion
The **span name** SHOULD be `{db.operation.name} {target}` if there is a
```

docs/database/database-spans.md
```diff
@@ -51,21 +51,31 @@ linkTitle: Client Calls
...
+If there is no (low-cardinality) `db.operation.name` available, database span names
+SHOULD be [`{target}`](#target-placeholder).
+<!-- markdown-link-check-enable -->
+
+The <span id="target-placeholder">`{target}`</span> SHOULD adhere to one of the following values, arranged in prioritized order, provided they are accessible:
+
+- `db.collection.name`
+- `db.namespace`
+- `server.address:server.port`
+- `network.peer.address:network.peer.port`
```
trask, line 76: can `network.peer.addres` be available when `server.address` isn't?

lmolkova: perhaps in this case we'll still set `server.address` to `192.1.1.1`

docs/database/elasticsearch.md
lmolkova: thanks for pointing it out. I don't mind giving individual semconv more freedom to define how the span name should look like.

I wonder what's the motivation not to include `server.address:port` in the span name and whether we should remove it from the generic fallback values then?

joaopgrassi (approved):

trask (approved):

gregkalapos (approved):

PR #976, opened 2024-04-26
Parameterized query text does not need to be sanitized by default

trask:
## Changes

Clarify that parameterized query text does not need to be sanitized by default.

Even though parameterized query text can potentially have sensitive data, by using a parameterized query
the user is giving a strong signal that any sensitive data will be passed as parameters, and the benefit
to observability of capturing it by default outweighs the risk.

### DIFF
--- a/docs/database/database-spans.md
+++ b/docs/database/database-spans.md
@@ -79,11 +79,11 @@ These attributes will usually be the same for all operations performed over the 
 | [`db.operation.name`](../attributes-registry/db.md) | string | The name of the operation or command being executed. | `findAndModify`; `HMSET`; `SELECT` | `Conditionally Required` [4] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`server.port`](../attributes-registry/server.md) | int | Server port number. [5] | `80`; `8080`; `443` | `Conditionally Required` [6] | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 | [`db.instance.id`](../attributes-registry/db.md) | string | An identifier (address, unique name, or any other identifier) of the database instance that is executing queries or mutations on the current connection. This is useful in cases where the database is running in a clustered environment and the instrumentation is able to record the node executing the query. The client may obtain this value in databases like MySQL using queries like `select @@hostname`. | `mysql-e26b99z.example.com` | `Recommended` If different from the `server.address` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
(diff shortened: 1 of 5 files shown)


model/trace/database.yaml
```diff
@@ -9,7 +9,16 @@ groups:
...
+            Non-parameterized query text SHOULD NOT be collected by default unless there is sanitization that excludes
+            sensitive data, e.g. by redacting all literal values present in the query text.
+
+            Parameterized query text SHOULD be collected by default
+            (the query parameter values themselves are opt-in,
+            see [`db.query.parameter.<key>`](../../docs/attributes-registry/db.md)).
+        note:
+          Even though parameterized query text can potentially have sensitive data, by using a parameterized query
+          the user is giving a strong signal that any sensitive data will be passed as parameters, and the benefit
+          to observability of capturing it by default outweighs the risk.
```
lmolkova, line 21: maybe

```suggestion
          Even though parameterized query text can potentially have sensitive data, by using a parameterized query
          the user is giving a strong signal that any sensitive data will be passed as parameter values, and the benefit
          to observability of capturing the static part of the query text by default outweighs the risk.
```

?

lmolkova (approved):

docs/database/database-spans.md
```diff
@@ -101,14 +101,17 @@ Semantic conventions for individual database systems SHOULD document what `db.na
...
+
+**[10]:** Semantic conventions for individual database systems SHOULD document whether `network.peer.*` attributes are applicable. Network peer address and port are useful when the application interacts with individual database nodes directly.
 If a database operation involved multiple network calls (for example retries), the address of the last contacted node SHOULD be used.
 
-**[10]:** When observed from the client side, and when communicating through an intermediary, `server.address` SHOULD represent the server address behind any intermediaries, for example proxies, if it's available.
+**[11]:** When observed from the client side, and when communicating through an intermediary, `server.address` SHOULD represent the server address behind any intermediaries, for example proxies, if it's available.
 
-**[11]:** Query parameters should only be captured when `db.query.text` is parameterized with placeholders.
+**[12]:** Query parameters should only be captured when `db.query.text` is parameterized with placeholders.
```
jack-berg, line 114: If we expect the parameters to contain sensitive information, then i believe we need to make  capturing those keys via `db.query.parameter.<key>` opt in. Not much value in having sensitive data omitted from `db.query.text` just to be included in a different parameter.

jack-berg: Oops just realized that's already the case!

jack-berg (approved):

PR #954, opened 2024-04-24
Add `event.data` attribute to provide mapping from span events to events

lmolkova:
Related to #829, #695, https://github.com/open-telemetry/opentelemetry-specification/issues/3406

## Changes

In GenAI WG, we'd like to be able to document semantic conventions for events, prototype and ship instrumentation libraries that emit them.

Event API is not available in the majority of languages yet, log are also still in development in many of them and the instrumentation libraries should not use logging bridge and produce log records anyway.

As explained in #829, we'd like to record events payload as if they were reported with Event API, but record them as span events until Event API is widely available.

This PR introduces `event.body` attribute that can be used to report event payload on a span event (available and stable everywhere).

The attribute records the payload as a JSON string.
Note that a `"plain string"` is a valid JSON  - see [RFC 7159](https://datatracker.ietf.org/doc/html/rfc7159#appendix-A)

>    o  Changed the definition of "JSON text" so that it can be any JSON
      value, removing the constraint that it be an object or array.

TODO
- [ ] Send spec PR to introduce mapping to transform span events to events. 

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* ~~[schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.~~

Issue #829, opened 2024-03-21
LLM semconv: how to capture prompts and completions


lmolkova:
### Area(s)

llm, events

### Is your change request related to a problem? Please describe.

Prompts and completions are essential part of debugging experience when developing LLM applications. 
They could be verbose and contain sensitive information, so they need some opt-in mechanism such as feature-flag, configuration property on the instr library. Specific OTel distros, debugging and diagnostic tools within IDEs may decide to enable the collection by default.

There are multiple ways to capture LLM events: as attributes on spans, as attributes on span events, as log events.

1. Span Attributes:
   - span attributes are limited in size (not on the OTel SDK, but on the vendor side) and could be as small as few kilobytes
   - they are not structured and we'd have to capture the format (e.g. mime type) to allow parsing them
   - but they always stay along with the span making visualization tooling and queries easier. Users can't forward them to a separate backend along with the logs.

2. Span event attributes:
   - span event's don't have a notion of body and are not structured either
   - event attributes are subject to the same problems as span attributes.
   - they are also exported along with the span and arguably are easy for visualization tooling to use
   - their future is uncertain - see https://github.com/open-telemetry/semantic-conventions/issues/695 and https://github.com/open-telemetry/opentelemetry-specification/issues/3406

3. Log events/logs:
   - Log events have a notion of body - tracing/logging vendors usually have bigger limits for it than for attributes
   - Log events are structured. We can specify the structure for each event and simplify parsing/querying 
   - They are exported separately from spans, but are still correlated with them. If user sends logs and traces to different backends, LLM-debugging experience would need cross-vendor queries which would be challenging.
   - They are not implemented/released/stabilized by [most of the languages](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#logs), [eventlogger](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md#eventlogger) implementations can only be found in PHP and C++

Since logs/event logs are not ready to be used by instrumentation libraries in popular languages  used by LLM applications (python, JS), at least the initial version of LLM semconv cannot use them leaving us with attributes vs span events choice.

### Describe the solution you'd like

LLM semantic conventions should allow to capture prompts and completions as span events, but also be future-proof and support new log events:
- I'd like to be able to describe LLM event structure/format in the semantic conventions
[...]

PR #980, opened 2024-04-29
Introduce per-message structured GenAI events instead of prompt/completion span events

### LINKED ISSUE #834: LLM: define common/system-specific event structure
### Area(s)

llm

See https://github.com/open-telemetry/semantic-conventions/pull/825#discussion_r1533040974

#825 is going to introduce LLM events and (currently) requires to put prompts and completions as json string in OpenAI format. Some LLM vendors might not use json and instrumentation would be forced to use specific structure.

We should decide if:
- we define a common format and all LLM vendor instrumentations are forced to use it. 
- we want one event with OpenAI JSON messages format (or another common format) and other vendors are forced to use it. 
- vendor-specific events. E.g. `gen_ai.{vendor}.prompt` and then vendors  could document their own structure or document that they emit OpenAI-formatted events
- we make format dynamic with additional attributes representing mime type.

This is closely coupled with https://github.com/open-telemetry/semantic-conventions/issues/829 

TBD

lmolkova:
Fixes #834

## Changes

- Breaks down prompt and completion events to system, user, tool, assistant events as well as the choice event 
- Specifies their structure
- Switches them to be log-based events instead of span events

Related to #954, #829 

* [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions - N/A


docs/gen-ai/llm-spans.md
```diff
@@ -36,49 +50,356 @@ Instrumentations that support it, MUST offer the ability to turn off capture of
...
 
 **[1]:** The name of the LLM a request is being made to. If the LLM is supplied by a vendor, then the value must be the exact name of the model requested. If the LLM is a fine-tuned custom model, the value should have a more specific name than the base model that's been fine-tuned.
 
 **[2]:** If not using a vendor-supplied model, provide a custom friendly name, such as a name of the company or project. If the instrumetnation reports any attributes specific to a custom model, the value provided in the `gen_ai.system` SHOULD match the custom attribute namespace segment. For example, if `gen_ai.system` is set to `the_best_llm`, custom attributes should be added in the `gen_ai.the_best_llm.*` namespace. If none of above options apply, the instrumentation should set `_OTHER`.
 
-**[3]:** If available. The name of the LLM serving a response. If the LLM is supplied by a vendor, then the value must be the exact name of the model actually used. If the LLM is a fine-tuned custom model, the value should have a more specific name than the base model that's been fine-tuned.
+**[3]:** If there is more than one finish reason in the response, the last one should be reported.
```
TaoChenOSU, line 71: Does it recommend to only report one finish reason even when there are multiple response candidates returned by the model?

docs/gen-ai/llm-spans.md
```diff
@@ -36,49 +50,356 @@ Instrumentations that support it, MUST offer the ability to turn off capture of
...
 
-<!-- semconv gen_ai.content.prompt -->
-The event name MUST be `gen_ai.content.prompt`.
+The event payload describes message content sent to or received from GenAI and depends on specific messages described in the following sections.
+
+Instrumentations for individual GenAI systems MAY add system specific fields into corresponding events payload.
+It's RECOMMENDED to document them in system-specific extensions.
+Telemetry consumers SHOULD expect to receive unknown payload fields.
+
+### System message
```
TaoChenOSU, line 100: Are we not going to support legacy completion APIs that accept single prompts?

docs/gen-ai/llm-spans.md
```diff
@@ -36,49 +50,356 @@ Instrumentations that support it, MUST offer the ability to turn off capture of
...
+`gen_ai.system` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `openai` | OpenAI | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- endsemconv -->
+
+| Body Field | Type | Description | Examples | Requirement Level | Sensitive |
+|---|---|---|---|---|---|
+| `role` | string | The role of the messages author | `assistant` | `Conditionally Required`: when the message is returned by GenAI model. |  |
```
TaoChenOSU, line 170: This says `Conditionally Required` while the `role` in `gen_ai.user.message` says `Opt-in`. Are they different?

docs/gen-ai/llm-spans.md
```diff
@@ -36,49 +50,356 @@ Instrumentations that support it, MUST offer the ability to turn off capture of
...
+     |---------------------|-------------------------------------------------------|
+     | `gen_ai.system`     | `"openai"`                                            |
+     | Event payload       | `{"role":"user","content":"How to instrument GenAI library with OTel?"}` |
+
+   - `gen_ai.assistant.message` (could be reported when capturing sensitive content is disabled)
+
+     |   Property          |                     Value                             |
+     |---------------------|-------------------------------------------------------|
+     | `gen_ai.system`     | `"openai"`                                            |
+     | Event payload (with sensitive content) | `{"role":"assistant","tool_calls":[{"id":"call_VSPygqKTWdrhaFErNvMV18Yl","function":{"name":"get_link_to_otel_semconv","arguments":"{\"semconv\":\"GenAI\"}"},"type":"function"}]}` |
```
TaoChenOSU, line 386: The event payload here contains the role attribute in the stringified value, while in the definition the `role` is an attribute of itself. Do we need the `role` attribute?

docs/gen-ai/llm-spans.md
```diff
@@ -36,49 +50,356 @@ Instrumentations that support it, MUST offer the ability to turn off capture of
...
+
+| Value  | Description | Stability |
+|---|---|---|
+| `openai` | OpenAI | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 <!-- endsemconv -->
 
 ## Events
 
-In the lifetime of an LLM span, an event for prompts sent and completions received MAY be created, depending on the configuration of the instrumentation.
+In the lifetime of an GenAI span, an event for each message application sends to GenAI and receives in response from it MAY be created, depending on the configuration of the instrumentation.
```
TaoChenOSU, line 84: Have we considered the debugging experience when an app leverages long running chats? If I have a chat that already has 100 iterations between the user and the AI, from that point of time, each invocation to the model will have 100+ events attached to the span.

docs/gen-ai/llm-spans.md
lmolkova: The individual finish reasons is reported in individual events for each choice. 

I decided to change finish reason to a single string since:
- It seems only(?) OpenAI returns more than one choice and it's not quite a popular scenario (most samples are based on the assumption that n=1)
- I could  not really simulate a case when different finish reasons would be returned

I.e. in most cases there will be one choice or one reason on the span.
What if it's not enough?

The alternatives are:
1. remove finish reason from the span altogether, but then it won't even be available on metrics
2. keep an array attribute. It won't be usable on metrics and in hard to use in custom queries
3. join all finish reasons to a single string and keep a comma-separated list sorted by the choice index. 

I think options 1 and 2 are bad since they don't allow to use finish reason on metrics.
Option 3 seems reasonable too.

docs/gen-ai/llm-spans.md
lmolkova: Legacy completion fits into the user message quite well.

docs/gen-ai/llm-spans.md
lmolkova: I think we chatted about it offline, sharing thoughts here for visibility:
100 iterations as a single huge event will be quite problematic:
- it will be hard to use when debugging
- it's mulch more likely to hit backend limit on event payload size

Having 100 events (not attached to spans, as log-based events are not) is not  a great experience either, but
- it's easier to read 100 independent messages than 1 huge message
- each event will be smaller and less likely to hit limits
- we can give events different severity allowing users to enable e.g. only responses, but not system or tool messages
- we can support configuring a limit on last X events

docs/gen-ai/llm-spans.md
lmolkova: I removed role for all events. We can always add it back later if it's needed. For now it seems to duplicate the event name.

docs/gen-ai/llm-spans.md
lmolkova: same here, no more role for now - we can always add it back later if needed

docs/gen-ai/llm-spans.md
```diff
@@ -65,26 +82,380 @@ These attributes track input data and metadata for a request to an LLM. Each att
...
+
+| Body Field | Type | Description | Examples | Requirement Level | Sensitive |
+|---|---|---|---|---|---|
+| `finish_reason` | string | The reason the model stopped generating tokens. | `stop`, `tool_calls`, `content_filter` | `Required` |  |
+| `index` | int | The index of the choice in the list of choices. | `1` | `Required` |  |
+| `message` | [Message](#message-object) | GenAI response message | `{"content":"The OpenAI semantic conventions are available at opentelemetry.io"}` | `Recommended` | ![Mixed](https://img.shields.io/badge/-mixed-orange) |
+
+#### `Message` object
+
+The message structure matches one of the messages defined in this document depending on the role:
```
langchain4j, line 240: Hi! Shouldn't it only be an assistant message? AFAIK LLMs do not output tool (result) messages, only tool calls, but those are located inside the assistant message.

docs/gen-ai/llm-spans.md
cartermp: I think that chat sessions run into the same problem we have today with using OTel to model browser sessions, data processing jobs, etc. where there's meaningful operations that could each be a trace, but the end-user experience is really a collection of many traces and could be unbounded in length.

To that end, I prefer that we model spans (and, well, traces) for some...kind...of....err..."meaningful" group of operations:

- A single chat request results in one LLM call or an entire network of agents or chains making N calls, this can be one trace
- A "user presses a button to produce some output" is a trace of N spans, where N is the number of operations involved to produce that output, be it a complex RAG pipeline, multiple LLM calls, or just a basic pass-through to an LLM to produce a response in text directly

What I think is missing in all of this though is a better correlation ID that ties traces together. This would have to be domain-specific, though, and I don't really know how (or if) we want to consider modeling that right now.

docs/gen-ai/llm-spans.md
```diff
@@ -65,26 +82,380 @@ These attributes track input data and metadata for a request to an LLM. Each att
...
+
+Events:
+
+1. `gen_ai.system.message`.
+   Not reported if when capturing sensitive content is not enabled.
+
+   |   Property          |                     Value                             |
+   |---------------------|-------------------------------------------------------|
+   | `gen_ai.system`     | `"openai"`                                            |
+   | Event payload       | `{"content":"You're a friendly bot that answers questions about OpenTelemetry."}` |
```
cartermp, line 299: A thought I had here is how we might go about giving these messages more structure over time. I recognize that a lot of this really just comes down to the system you're using, as different model APIs will return more or less structure than others, different keys for values, etc. I don't know how we might address this over time but I'd love to get into a world where the experience for most people involves nice, system-readable, structured information.

docs/gen-ai/llm-spans.md
lmolkova: In general, I agree - we'll need to add context identifier (e.g. gen-ai.thread.id) to correlate calls within the same conversation. 

For this specific case: the conversation is carried over in each request. So the instrumentation can't do a lot - it can only  record messages in the provided prompt (optionally allowing to filter or limit them).

It seems the be the API problem though. I might be wrong, but it feels like OpenAI assistant API solves this problem and take care of storing the context (all the conversation). Given how fast APIs are evolving, we might not need to solve the message duplication problem.

docs/gen-ai/llm-spans.md
lmolkova: This is an example that shows payload as a json - I did add a bare-minimum structure in the event definitions above and  agree that we need to keep extending it

docs/gen-ai/llm-spans.md
lmolkova: good catch, thanks!

docs/gen-ai/llm-spans.md
cartermp: Yeah, I think maybe for now even we could punt on the issue, since we technically do have that kind of causality already with span links.

drewby (changes requested):
I think most, if not all, conversations are resolved. There are some branch conflicts and and build issues. Once those are resolved, I think we can merge this.

docs/gen-ai/llm-spans.md
danielsc: It is quite easy to get different finish reasons, between `stop`, `length` and `content_filter`.

I would just let people pull it out of the choice event.

model/registry/gen-ai.yaml
```diff
@@ -20,55 +20,46 @@ groups:
...
         examples: ['chatcmpl-123']
-        tag: llm-generic-response
       - id: response.model
         stability: experimental
         type: string
         brief: The name of the LLM a response was generated from.
         examples: ['gpt-4-0613']
-        tag: llm-generic-response
-      - id: response.finish_reasons
+      - id: response.finish_reason
```
codefromthecrypt, line 53: @patcher9 this accidentally bumps you into compatible lol https://github.com/openlit/openlit/issues/318

codefromthecrypt (commented):
I'm in favor of progress here, though I might peel off the unrelated change into an easier to merge PR (finish_reason part). My main concern is around events and adoptability.

We are defining a span attribute `gen_ai.event.content` for a potentially large and sensitive value. I can tell this is compensating for missing language features, but it also seems a reason to push for those to complete. Particularly, there are two top languages today: python and javascript. It is hard to reason with why a small api like event is too much to implement. Having it conditional causes this tech debt and would be better to get rid of that. If it is conditional, I would consider raising a tracking issue and then citing that in the docs, so it is easy to tell when the special case can be removed.

Next is adoptability: even the prior wasn't converged and also I've had feedback that feels like some tools will not change the approach they are using today. Especially from SIG members, but not just SIG members, I would like to see who controls implementations and would actually transition to this format once merged.

If no one will actually use what's defined, and we feel they are compatible even if they don't, then maybe we should consider removing this prompt/completion recording vs porting it to structured events. This piece is already security and operationally sensitive, so we should make sure the effort to maintain the specs around it is worth it (in terms of use).

copying in directly @nirga (openllmetry) @patcher9 (openlit) @karthikscale3 (langtrace) as they are a few implementors who would have to change their code.

docs/gen-ai/llm-spans.md
```diff
@@ -79,52 +96,473 @@ These attributes track input data and metadata for a request to an LLM. Each att
 
 ## Events
 
-In the lifetime of an LLM span, an event for prompts sent and completions received MAY be created, depending on the configuration of the instrumentation.
+In the lifetime of an GenAI span, an event for each message application sends to GenAI and receives in response from it MAY be created, depending on the configuration of the instrumentation.
+The generic events applicable to multiple GenAI models follow `gen_ai.{type}.message` naming pattern.
+GenAI vendor-specific instrumentations SHOULD follow `gen_ai.{gen_ai.system}.{type}.message` pattern to record system-specific events.
+
+It's RECOMMENDED to use [Event API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md) to record GenAI events once it's implemented in corresponding language.
```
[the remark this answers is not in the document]
cartermp: @codefromthecrypt it's recommended in part because events (i.e., named logs) are sent as a separate batch from trace data. Many telemetry backends have harder restrictions on trace data, mostly because of attribute count and size limitations, but they typically do not have anywhere near those kinds of restrictions on log data. In a perfect world, this data would probably be captured as a part of a trace tracking the call/response to an LLM, but tracing systems at large are not design to handle large payloads.[0]

That said, the events API is very new, and it's not likely that a given implementation language will have events implemented. It would be expected that a given language that logs this data as a part of a trace (or something else) eventually moves to the events API once it's available for them.

[0] There are several exceptions to this on the market today, but it's not generally the case, and even these exceptions have limits you can run into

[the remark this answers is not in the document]
codefromthecrypt: @cartermp sorry for sidetracking as I'm sure folks have thought about this, but could we establish an event bridge instead of special casing here? at least then it would be easier to reason with and then a mechanical concern vs special to this LLM.

I wrote a spec for wasm which had similar feature concerns. e.g. is buffering available? basically if we know that events aren't available and it doesn't matter the reason albeit lang or backend, then we have a compensating alternative. Some may choose to not enable that alternative.

ack feel free to punt me.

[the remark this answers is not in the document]
cartermp: > but could we establish an event bridge instead of special casing here?

Can you clarify what you mean by this? In OTel, logs are already defined as a part of the "logs bridge" concept, and events are just logs with a name. But this may be a different terminology from what you're referring to.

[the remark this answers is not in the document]
codefromthecrypt: Sorry got very sick and was offline.

So for the highlighted line, we are saying to use events when they are implemented:

> It's RECOMMENDED to use [Event API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md) to record GenAI events once it's implemented in corresponding language.

I meant to ask if there was (or could be) a canonical way to adapt a spec aimed for the Event API until such time as it is available. If there was a way (call it X)..

> It's RECOMMENDED to use [Event API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md) to record GenAI events. Instrumentation can use X to adapt to the Event API in languages or integrations where it is not yet available.

What had me mostly on this line is the inability to link to a resolution of the event api being unavailable. If an interim approach X was linked, it could be updated later saying "this is deprecated as all languages now have the event API" and those looking at this doc could navigate to that conclusion.

It is about the stale doc problem for a concern that appears temporary. Hope this helps.

docs/gen-ai/llm-spans.md
lmolkova: is it common to do more than one choice? The main motivation for this change is that it's not common (some models support one choice only, some openai SDKs like openai-dotnet don't even allow to request more than one).
 
But this is not strictly related to this change. I'll return finish_reasons array and create an  issue to discuss it further.

drewby (approved):
I think most feedback is addressed. Unless there is any final critical items I think we should merge this and allow further feedback/updates via new Issues/PRs.

docs/gen-ai/llm-spans.md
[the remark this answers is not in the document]
codefromthecrypt: you can close this comment. I can follow-up later on which languages support log events API. It isn't blocking feedback.

codefromthecrypt:
there's a semantic release pending (1.27.0) and big q on my mind is if this is blocking or not?

For me, I'm not a maintainer so can't make blocking feedback, but https://github.com/open-telemetry/semantic-conventions/pull/980#discussion_r1669421525 is still curious and I'm not sure if it is a bug/fuzz or not.

'gen_ai.event.content' moves back into span attribute size/sensitivity and also doesn't indicate what this is about (prompt or completion). It seems to imply it is the input content, so not sure why if this is intentional it isn't named as such.

docs/gen-ai/llm-spans.md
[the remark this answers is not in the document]
lmolkova: There is a matrix here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#events (short answer - none, but it's in development), it sometimes possible to abuse logs api to send propoer log-based events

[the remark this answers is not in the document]
lmolkova: I'll add  link to the text

docs/gen-ai/llm-spans.md
lmolkova: Removed the change and created https://github.com/open-telemetry/semantic-conventions/issues/1277 to follow up

docs/gen-ai/llm-spans.md
lmolkova: Resolving this - we can reuse severity on individual events and allow users to filter by it. We can also introduce some options to report last N messages, etc.

Regardless of the approach we take, long chats would be expensive and verbose.

docs/gen-ai/llm-spans.md
lmolkova: I have the following wording in the event section

> Instrumentations for individual GenAI systems MAY add system specific fields into corresponding events payload. It's RECOMMENDED to document them in system-specific extensions. Telemetry consumers SHOULD expect to receive unknown payload fields.

I can explicitly call out that we expect to add new generic body fields into the payloads, but I feel it's implied and we already warn consumers. WDYT?

docs/gen-ai/llm-spans.md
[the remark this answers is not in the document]
lmolkova: done, closing for now

docs/attributes-registry/gen-ai.md
```diff
@@ -10,28 +10,23 @@
...
-
-**[2]:** It's RECOMMENDED to format prompts as JSON string matching [OpenAI messages format](https://platform.openai.com/docs/guides/text-generation)
-
-**[3]:** The actual GenAI product may differ from the one identified by the client. For example, when using OpenAI client libraries to communicate with Mistral, the `gen_ai.system` is set to `openai` based on the instrumentation's best knowledge.
+| Attribute                        | Type   | Description                                                                | Examples                                                          | Stability                                                        |
+| -------------------------------- | ------ | -------------------------------------------------------------------------- | ----------------------------------------------------------------- | ---------------------------------------------------------------- |
+| `gen_ai.event.content`           | string | The GenAI event payload serialized into JSON string.                       | `[{'content': 'What is the capital of France?'}]`; `plain string` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
lmolkova: updated, PTAL

lmolkova:
> there's a semantic release pending (1.27.0) and big q on my mind is if this is blocking the release or not?

I don't believe this PR is blocking the release. Attributes that are being deprecated have been there is the previous release and we won't save anything by deprecating them in 1.27.0 vs 1.28.0

lmolkova:
> We are defining a span attribute gen_ai.event.content for a potentially large and sensitive value. I can tell this is compensating for missing language features, but it also seems a reason to push for those to complete. Particularly, there are two top languages today: python and javascript. It is hard to reason with why a small api like event is too much to implement. Having it conditional causes this tech debt and would be better to get rid of that. If it is conditional, I would consider raising a tracking issue and then citing that in the docs, so it is easy to tell when the special case can be removed.

I'm all in to go and help OTel SDKs implement event API (if they're ready to review and accept contributions in this area) and would encourage anyone who has some capacity for it.

It's unrealistic to expect this API to become stable in the next several months. Using temporary attribute that's never going to be stabilized gives us time and freedom to experiment, design and evolve event structure.

Having Event API available and stable would be prerequisite for stability, but this is an experimental spec and we still need a way to make progress in GenAI area without EventAPI available.

I created #1278 and added a todo comment to the attribute to remove it.

Note:
it's possible to build a processor that transforms span even with specific attribute into proper log-event with structured payload for future-compatibility.

docs/gen-ai/gen-ai-spans.md
```diff
@@ -109,45 +105,173 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
...
+#### `Message` object
+
+| Body Field | Type | Description | Examples | Requirement Level | Sensitive |
+|---|---|---|---|---|---|
+| `content` | `AnyValue` | The contents of the assistant message. | `Spans, events, metrics defined by the GenAI semantic conventions.` | `Opt-In` | ![Sensitive](https://img.shields.io/badge/-sensitive-red) |
+| `tool_calls` | [ToolCall](#toolcall-object)[] | The tool calls generated by the model, such as function calls. | `[{"id":"call_mszuSIzqtI65i1wAUOE8w5H4", "function":{"name":"get_link_to_otel_semconv", "arguments":"{\"semconv\":\"gen_ai\"}"}, "type":"function"}]` | `Conditionally Required`: if available | ![Mixed](https://img.shields.io/badge/-mixed-orange) |
+
+### Recording GenAI events as span events
+
+TODO: Remove this attribute once Event API is widely available https://github.com/open-telemetry/semantic-conventions/issues/1278
```
lmolkova, line 253: Add explanation that we should use the same event names as for proper-events

docs/gen-ai/gen-ai-spans.md
```diff
@@ -109,45 +105,173 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
 
 ## Events
```
lmolkova, line 106: move to a separate doc, events can be used without tracing

docs/gen-ai/gen-ai-spans.md
```diff
@@ -109,45 +105,173 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
...
-In the lifetime of a GenAI span, an event for prompts sent and completions received MAY be created, depending on the configuration of the instrumentation.
+GenAI instrumentation MAY capture user inputs send to the model and responses received from it.
+
+It's RECOMMENDED to use [Event API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md) to record GenAI events once it's implemented in corresponding language (check out [Spec Compliance Matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#events) for the details).
 
-<!-- semconv gen_ai.content.prompt -->
+The event payload describes message content sent to or received from GenAI and depends on specific messages described in the following sections.
+
+Instrumentations for individual GenAI systems MAY add system specific fields into corresponding events payload.
+It's RECOMMENDED to document them in system-specific extensions.
```
lmolkova, line 115: mention that redaction is instrumentation concern

docs/gen-ai/gen-ai-spans.md
```diff
@@ -109,45 +105,173 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
...
+
+### Configuration
+
+Instrumentations that support capturing inputs and outputs, SHOULD NOT capture them by default and SHOULD offer the ability to turn on capturing of prompts and completions. This is for three primary reasons:
+
+1. Data privacy concerns. End users of GenAI applications may input sensitive information or personally identifiable information (PII) that they do not wish to be sent to a telemetry backend.
+2. Data size concerns. Although there is no specified limit to sizes, there are practical limitations in programming languages and telemetry systems. Some GenAI systems allow for extremely large context windows that end users may take full advantage of.
+3. Performance concerns. Sending large amounts of data to a telemetry backend may cause performance issues for the application.
+
+Instrumentations MAY offer additional configuration options allowing to control sensitive content is capture.
```
lmolkova, line 126: change to "SHOULD offer", default to events with content redacted - we can change later based on the feedback

codefromthecrypt (approved):
:shipit: when you're ready (noticed you have some clarifying TODOs I agree with). Thanks for detailing the stuff on the call.

a4doks (approved):
Looks good to me. Thanks @lmolkova . This has been long time coming!

karthikscale3 (approved):
Ship it! Thank you @lmolkova

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,403 @@
...
+
+**Status**: [Experimental][DocumentStatus]
+
+<!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->
+
+<!-- toc -->
+
+- [Naming pattern](#naming-pattern)
+- [Common attributes](#common-attributes)
+- [System message](#system-message)
```
lmolkova, line 15: event names are lost

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,403 @@
...
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## System message
+
+This event describes the instructions passed to the GenAI model.
+It's NOT RECOMMENDED to report this event when capturing content is disabled, unless there are additional system-specific fields
+instrumentation adds to the event body.
+
+| Body Field | Type | Description | Examples | Requirement Level |
```
lmolkova, line 105: table is broken

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,403 @@
...
+
+Instrumentations MAY capture inputs and outputs if and only if application has enabled the collection of this data.
+This is for three primary reasons:
+
+1. Data privacy concerns. End users of GenAI applications may input sensitive information or personally identifiable information (PII) that they do not wish to be sent to a telemetry backend.
+2. Data size concerns. Although there is no specified limit to sizes, there are practical limitations in programming languages and telemetry systems. Some GenAI systems allow for extremely large context windows that end users may take full advantage of.
+3. Performance concerns. Sending large amounts of data to a telemetry backend may cause performance issues for the application.
+
+Body fields that contain user input, model output, or other potentially sensitive and verbose data
+SHOULD NOT be captured by default. Corresponding values, when present, MAY be replaced with `*`.
```
lmolkova, line 41: consider not capturing the content object at all if it's not enabled

codefromthecrypt (commented):
*I agree with the direction*, so please take the following feedback as supportive of re-approval. Please do not interpret any of this as arguing or argue back unless you disagree with the move to log events. In that case, give me all the angst!

This PR is now a change from span events to log events, and does not explicitly handle the migration story. This still exists.

* this issue changed in scope and should have its description and name rewritten
* I have personally asked people to help with a schema or transform processor which could mechanically change span events to log events. We need to make or develop issues for this transition
* This no longer references any external issue about the very obvious status quo that events are not supported in any way besides experimental, and not landed in any primary language of genai. I believe an effective tracking issue (a la how @jack-berg rolls) should be created with a clear updated task list of what's left
* folks who click approve honestly should be ready to implement this concretely. This means you *will* change your own code or at least contrib to make this change possible to manifest. We need to change the status quo where approve is decoupled from action. You may hate me for saying this aloud, but it needs to be said.

These are non-binding comments as so far no one has invited me to be an approver on this sig

lmolkova:
> this issue changed in scope and should have its description and name rewritten

could you please point to exactly place that should be rewritten? I don't see anything that needs to be changed in the title or description of this PR or the original issue.

> This no longer references any external issue about the very obvious status quo that events are not supported in any way besides experimental, and not landed in any primary language of genai.

I added a link to the spec-compliance matrix in the doc. 

If you believe there are other things on this PR I should improve, I'm happy to do so and would appreciate specific suggestions on the related changes.

codefromthecrypt:
suggestion: "Introduce per-message structured GenAI events" -> "Change prompt-completion recording from span events to log events"

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,379 @@
...
+This event describes the assistant message.
+
+The event name MUST be `gen_ai.assistant.message`.
+
+| Body Field   | Type                           | Description                            | Examples                                        | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) |
+|--------------|--------------------------------|----------------------------------------|-------------------------------------------------|-------------------|
+| `content`    | `AnyValue`                     | The contents of the assistant message. | `Spans, events, metrics defined by the GenAI semantic conventions.` | `Opt-In` |
+| `tool_calls` | [ToolCall](#toolcall-object)[] | The tool calls generated by the model, such as function calls. | `[{"id":"call_mszuSIzqtI65i1wAUOE8w5H4", "function":{"name":"get_link_to_otel_semconv", "arguments":"{\"semconv\":\"gen_ai\"}"}, "type":"function"}]` | `Conditionally Required`: if available |
+
+### `ToolCall` object
```
codefromthecrypt, line 134: copying some folks from last LLM call @nirga @karthikscale3 @AmanAgarwal041 @ThomasVitale

unless I am mistaken, this is a big feature change of the spec. Before, there was to my knowledge no attributes or other developments on tool calls. I'm not sure how compatible this is across the board, but I noticed for example openllmetry has an open issue about [anthropic tool calls](https://github.com/traceloop/openllmetry/issues/879).

Aside, I think the description could be fleshed out @lmolkova to specify what features this PR supports (modeling tool calls is a very big deal for interop, and somewhat independent from being represented in log events). This is a suggestion you can feel free to ignore if you think it is implied.

In summary, while I understand not all backends will be able to support log events soon, and may shrug at this PR thinking they will just convert back to span events. Even if that's the case, you should still look closely at the model here. It isn't just changing to log events, it is also fleshing out features not currently tracked in a specific issue (notably tool calls).

lmolkova: tool calls are hidden  in underspecified prompt and completion events.

https://platform.openai.com/docs/guides/function-calling

they are part of gen-ai responses and input sent by users. This PR describes the structure and clarifies the content of the events. You can't address it without specifying all the basic pieces. 

I.e. this PR does not add any new big feature.

codefromthecrypt: The reason I raised this comment is besides [one comment](https://github.com/open-telemetry/semantic-conventions/pull/980#discussion_r1594321576) from @langchain4j in an early draft of this PR, I don't see any discussion about tool calling. I took time to make this thread because it isn't obvious in the title or description that tool call modeling is in the same scope as the mechanics of span event -> log events. Almost all discussion was on the latter point.

While it may indeed be a small change, so far I don't see any one discussing it, much less looking up cohere, anthropic or other non-openai change. There are a number of fields around tool calls which may be brittle if not looked at independent of the other parts of this change.

Two suggestions FWIW:
* fork off tool call specification into a separate issue PR and remove it from this one
* redo the description and on the next call get a diverse opinion on how the changes here apply

The latter is tricky because of the same reason the main PR is tricky. As the api isn't available, people need to theorize about the change vs practice it.

My 2p, and last opinion until someone else pipes up. Thanks so far!

lmolkova: the title of this PR says "introduce per-message structured GenAI events". 

The tool calls is just on of them, I can't separate it from this PR because then GenAI conventions will not be complete.

I.e. previously we assumed tool calls to be in the prompt and completion span event. Now they what? should not be captured?

I will keep tool calls inside the scope of this PR. This is an experimental spec and I'm sure we will need to specify and update the structure further.

lmolkova: > While it may indeed be a small change, so far I don't see any one discussing it, much less looking up cohere, anthropic or other non-openai change. There are a number of fields around tool calls which may be brittle if not looked at independent of the other parts of this change.

if you have some specific proposal on what needs to be added to better support non-openai and what currently in this PR does not work for them, please bring specific suggestions up.

jsuereth: To reinforce this -

This PR should be evaluated based on the proposed EVENT being defined.  This is replacing attributes with an Event, and that event should serve the same purpose.   This *is* a feature change.  If you disagree with the shape of the Event being proposed - comment as such.

I think we (OpenTelemetry) fragment problems a bit too often to the point where we're no longer solving an issue, but solving a fragment of an issue and causing larger ones.  If you have a proposal for a better shaped event, please make it.  If you think ToolCalls shouldn't exist, please say so, and provide justification.

TL;DR; please make this clear as a reviewer of this PR.  What is the current issue with ToolCalls?  That should be an obvious statement on this line of code for maintainers to understand the concern and address it.  I'm unable to read your statement and figure out what you want to see in this design.

codefromthecrypt: First, the reason I pinged folks on this topic was an attempt to get current implementors to consider this issue, as they'll have experience. There are many already modeling span attributes or otherwise to represent tool events. These may or may not be easy to index. I'm not sure which are instrumenting the tool calls, and if that's done via spans, if they are adding their version of tool call ID to the LLM spans, or if they are making child spans for each tool call etc. I was hoping that people writing instrumentation can mention what if any concerns they have. I think those working on this every day will see what I'm saying and decide if this is an issue or not.

I'm being asked to present a problem on my own, so no issues doing so. Here goes. I mentioned Anthropic because I noticed openllmetry having an open issue for it. It is a good example to elaborate why it should be a separate topic vs being a sidecar of something to switch from span events to log events.

In [anthropic docs](https://docs.anthropic.com/en/docs/build-with-claude/tool-use#handling-tool-use-and-tool-result-content-blocks), they say this:

> Unlike APIs that separate tool use or use special roles like tool or function, Anthropic’s API integrates tools directly into the user and assistant message structure.

1. In our spec, we hint there is a `tool` role (e.g. `gen_ai.tool.message`), but that's not the case in anthropic. Is that ok? What does cohere do?

While there is an example, fields or events themselves have no description of what a tool call is or any implications of it. In anthropic "All tools are user-provided", so there are no server-side tools. This means it is a very good question, if these should be spans or not.. more on this later.

2. Should there be a description of tool calls in general, or if they are user provided or not? While we do hint that they are always user provided in the examples, I believe some description should exist outside the examples.

At least in anthropic, tool calls are logic defined in the same process as what calls the LLM.  At that point you are told to "Execute the actual tool code client-side." So what we call a `gen_ai.response.finish_reasons`? In the example, we hint at `tool_calls` (`tool_use` in anthropic) but it isn't listed as a choice

3. consider adding `tool_calls` to finish_reasons or explain why not

Next, so at that point you invoke a call. Now, if instrumenting, perhaps some instrument these tool calls also as a span. Right now, in the examples we say "Application executes the tool call. Application may create span which is not covered by this semantic convention." I don't personally think this should be only defined in an example. Also, knowing how any of the existing implementations work is helpful to guide this part of the spec. Do they set the span's parent as the LLM span? Do they add an attribute of `tool_call_id` to explain latency or errors?

4. Should we make any relationship note about instrumentation of user-provided tool calls? For example, the span should be a child of the LLM one and it should have an attribute of the `tool_call_id`. I think saying that spans should be parented to the LLM if done, is reasonable to say (in the spec, not just hinted in an example)

Next, right now there is no mention about tool errors, a type of message or anything. In Anthropic there is an `is_error` Do current implementors not care about this part, because for example everything is done sequentially and the error would bubble up into the calling span, or capturing it would be too expensive? If we don't care about errors, saying why raises less questions.

5. Should there be a way to know if a tool call was an error? How would we represent that?

I stopped at five. While I'm happy to hear @lmolkova's take on this, I really do want to hear what folks who are currently implementing this feel about it also. If I fail at getting feedback from implementors so be it, it was worth a shot!

codefromthecrypt: @jsuereth to your question also:

> This PR should be evaluated based on the proposed EVENT being defined. This is replacing attributes with an Event, and that event should serve the same purpose. This is a feature change. If you disagree with the shape of the Event being proposed - comment as such.

"This is replacing attributes with an Event, and that event should serve the same purpose" and basically that's my whole point. This PR is not limiting itself to replacement of existing model. I goes into modeling of something not yet modeled (tool calls). Before this change, we had basic prompt/completion span events. The change as done now extends that to tool calls, etc.

I feel this point is being treated as a nit, or draconian. The request to at least say that this is also getting into new models, if not branch off a new issue. I honestly wonder how my feedback is resisted so much? @jack-berg can you help me understand from a different POV, why I am incorrect to raise issues like this? I really want to understand and use my time wisely. I tend to spend a lot of hours on non-technical back and forth and really want to get to a point where I can get more done. Raise an issue, help solve it, and done. Less about me doing something wrong for raising it.

lmolkova: I'll address your technical comments one-by-one. To offer some POV in general: your feedback was not specific or actionable so far. You've dismissed my explanation that tool calls were already inside existing attributes and this PR makes things obvious instead of implicit. And that tool calls therefor cannot be separated for it. 

I'll be happy to hear domain-specific feedback based on facts with specific suggestions. I appreciate it's being left as a separate comments that we can effectively discuss and resolve independently.

I.e. the problem is the form of your feedback and lack of clarity in it.

docs/attributes-registry/gen-ai.md
```diff
@@ -13,34 +13,28 @@
...
+| `gen_ai.request.temperature`       | double   | The temperature setting for the GenAI request.                                                   | `0.0`                     | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `gen_ai.request.top_k`             | double   | The top_k sampling setting for the GenAI request.                                                | `1.0`                     | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `gen_ai.request.top_p`             | double   | The top_p sampling setting for the GenAI request.                                                | `1.0`                     | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `gen_ai.response.finish_reasons`   | string[] | Array of reasons the model stopped generating tokens, corresponding to each generation received. | `["stop"]`                | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 27: > consider adding tool_calls to finish_reasons or explain why not

finish_reasons is an array of a free form strings - we didn't get to defining a unified list yet (since this spec is in early development). If you check examples, you'll find finish_reason=tool_call used there.

So there is no place to add it to (yet). This PR does not change anything around it. I'll happily review a PR which normalizes a list of such reasons, but won't add it to the scope of this PR.

model/trace/gen-ai.yaml
```diff
@@ -56,32 +56,3 @@ groups:
           The `error.type` SHOULD match the error code returned by the Generative AI provider or the client library,
```
lmolkova, line 56: > Should there be a way to know if a tool call was an error? How would we represent that?

Yes, the error is captured on the span, not inside the event. `error.type` attribute is a generic way to capture span errors.

If there is a need to add such attribute on the event, it can be done, but in case of LLM:
- when model responds, if there is a tool call there, then there is no error - response was successfully received. 
- when app executes a tool and returns the response back to genai, then if the request fails, it will fail later

I.e. I don't think there is a  need for `error.type` on specific events, but such attribute can be added if the need arrives

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,379 @@
...
+
+   - `gen_ai.choice`
+
+     |   Property          |                     Value                             |
+     |---------------------|-------------------------------------------------------|
+     | `gen_ai.system`     | `"openai"`                                            |
+     | Event body (with content)    | `{"index":0,"finish_reason":"tool_calls","message":{"tool_calls":[{"id":"call_VSPygqKTWdrhaFErNvMV18Yl","function":{"name":"get_link_to_otel_semconv","arguments":"{\"semconv\":\"GenAI\"}"},"type":"function"}]}` |
+     | Event body (without content) | `{"index":0,"finish_reason":"tool_calls","message":{"tool_calls":[{"id":"call_VSPygqKTWdrhaFErNvMV18Yl","function":{"name":"get_link_to_otel_semconv"},"type":"function"}]}` |
+
+2. Application executes the tool call. Application may create span which is not covered by this semantic convention.
```
lmolkova, line 282: > Should we make any relationship note about instrumentation of user-provided tool calls? For example, the span should be a child of the LLM one and it should have an attribute of the tool_call_id. I think saying that spans should be parented to the LLM if done, is reasonable to say (in the spec, not just hinted in an example)

maybe, but this will be out of scope of this PR - it currently only specifies what's in the prompt and completion events. Please create an issue if you'd be interested.; PRs are also appreciated.

model/trace/gen-ai.yaml
lmolkova, line 56: > Next, right now there is no mention about tool errors, a type of message or anything. In Anthropic there is an is_error Do current implementors not care about this part, because for example everything is done sequentially and the error would bubble up into the calling span, or capturing it would be too expensive? If we don't care about errors, saying why raises less questions.

We should have a separate document that clarifies how generic conventions should apply to openai, anthropic, etc.

Check out how it's done for [databases](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/README.md) or messaging. 

There is an explicit clause allowing such extensions to add vendor-specific fields

> Semantic conventions for individual systems which extend content events SHOULD document all additional body fields and specify whether they
should be captured by default or need application to opt into capturing them.

Again, if you want to volunteer and send PR to add a document for anthropic, it'd be much appreciated.

This PR, again, does not change status quo - the prompts and completions were de-facto undefined for anthropic, now they are defined a bit more clearly. there is more work to do.

model/logs/gen-ai-events.yaml
```diff
@@ -0,0 +1,42 @@
...
+    type: event
+    brief: >
+      This event describes the assistant message passed to GenAI system or received from it.
+    extends: gen_ai.common.event.attributes
+
+  - id: gen_ai.tool.message
+    name: gen_ai.tool.message
+    type: event
+    brief: >
+      This event describes the tool or function response message.
```
lmolkova, line 34: > In [anthropic docs](https://docs.anthropic.com/en/docs/build-with-claude/tool-use#handling-tool-use-and-tool-result-content-blocks), they say this:

> Unlike APIs that separate tool use or use special roles like tool or function, Anthropic’s API integrates tools directly into the user and assistant message structure.

> In our spec, we hint there is a tool role (e.g. gen_ai.tool.message), but that's not the case in anthropic. Is that ok? What does cohere do?

As I mentioned in the other comment, we'll need to add a document that shows how to apply generic semconv to individual systems.

Given that OpenAI API is de-facto embraced by the community and differnt LLM vendors as universal one, we, based on the SIG discussions, decided to keep things more open-ai friendly.

For Anthropic it probably means that the response

```json
{
  "role": "assistant",
  "content": [
    {
      "type": "text",
      "text": "<thinking>To answer this question, I will: 1. Use the get_weather tool to get the current weather in San Francisco. 2. Use the get_time tool to get the current time in the America/Los_Angeles timezone, which covers San Francisco, CA.</thinking>"
    },
    {
      "type": "tool_use",
      "id": "toolu_01A09q90qw90lq917835lq9",
      "name": "get_weather",
      "input": {"location": "San Francisco, CA"}
    }
  ]
}
```

would be represented in choice event in the same way as openai one does:

```
`{"index":0,"finish_reason":"tool_calls","message":{"tool_calls":[{"id":"toolu_01A09q90qw90lq917835lq9","function":{"name":"get_weather","arguments":"{\"location\":\"San Francisco, CA\"}"},"type":"tool_use"}]}`
```

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,379 @@
...
+### `Function` object
+
+| Body Field  | Type       | Description                            | Examples                   | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) |
+|-------------|------------|----------------------------------------|----------------------------|-------------------|
+| `name`      | string     | The name of the function to call       | `get_link_to_otel_semconv` | `Required`        |
+| `arguments` | `AnyValue` | The arguments to pass the the function | `{"semconv": "gen_ai"}`    | `Opt-In`          |
+
+## Tool message
+
+This event describes the output of the tool or function submitted back to the model.
```
lmolkova, line 150: > While there is an example, fields or events themselves have no description of what a tool call is or any implications of it. In anthropic "All tools are user-provided", so there are no server-side tools. This means it is a very good question, if these should be spans or not.. more on this later.

This PR describes how tool calls appear in prompts and completions. If server-side functions need extra fields, we can add them incrementally.

This PR does not describe (and does not intend to) describe how applications should record tool execution.

If you believe it's not clear or need additional explanations, the fastest way to achieve it would be to leave a specific suggestions at specific lines of this PR.

docs/gen-ai/gen-ai-events.md
lmolkova: I believe I addressed your questions in separate comments on this PR. As you may notice many of them seem to be out of scope of this PR and represent important work we need to do in llm semconv. 

I prefer us to focus on this work rather than nitpicking on PR descriptions. If you want to volunteer and help us do this work, you're more than welcome.

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,379 @@
...
+| `tool_calls` | [ToolCall](#toolcall-object)[] | The tool calls generated by the model, such as function calls. | `[{"id":"call_mszuSIzqtI65i1wAUOE8w5H4", "function":{"name":"get_link_to_otel_semconv", "arguments":"{\"semconv\":\"gen_ai\"}"}, "type":"function"}]` | `Conditionally Required`: if available |
+
+### `ToolCall` object
+
+| Body Field | Type                           | Description                            | Examples                                        | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) |
+|------------|--------------------------------|----------------------------------------|-------------------------------------------------|-------------------|
+| `id`       | string                         | The id of the tool call                | `call_mszuSIzqtI65i1wAUOE8w5H4`                 | `Required`        |
+| `type`     | string                         | The type of the tool                   | `function`                                      | `Required`        |
+
```
trentm, line 140:
```suggestion
| `function` | string                         | The type of the tool                   | `function`                                      | `Required`        |
| `function` | [Function](#function-object) | | `{"name":"get_link_to_otel_semconv", "arguments":"{\"semconv\":\"gen_ai\"}"}` | `Required` |
```

I think

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,379 @@
...
+## Choice message
+
+This event describes the model-generated chat response message (choice).
+If GenAI model returns multiple choices, each choice SHOULD be recorded as an individual event.
+
+When response is streamed, instrumentations that report response events MUST reconstruct and report the full message and MUST NOT report individual chunks as events.
+If the request to GenAI model fails with an error before content is received, instrumentation SHOULD report an event with truncated content (if enabled). If `finish_reason` was not received, it MUST be set to `error`.
+
+The event name MUST be `gen_ai.choice`.
```
trentm, line 167:
```suggestion
The event name MUST be `gen_ai.choice.message`.
```

docs/gen-ai/gen-ai-events.md
codefromthecrypt: I don't appreciate this comment being dismissed as nitpicking, as I made it exactly on the line where the content was discussed. https://github.com/open-telemetry/semantic-conventions/pull/980#discussion_r1714651646

I tagged people who are implementing this to please take a close look, only the part addressed to you personally was maybe find a way to make more visible some changes that didn't have much or any discussion yet. Then, I was told to do that for them, and I did. I still think we should be more interested in the implementations vs telling me that calling attention to things is a nit or non-work.

codefromthecrypt: @nirga @karthikscale3 @AmanAgarwal041 @ThomasVitale I did my best to raise issues on the content of things around tool calling. I will not be making comments on the follow-up threads here, as I have to respect my boundaries. I have unsubscribed from this issue.

I would encourage you to actually try this PR in javascript if you can't yet in python due to missing event api. I know everyone is very busy, so probably not all of you will be able to. If you decide to approve without trying it, that's your prerogative also. That said, you have the most experience, many of you dealing with multiple non-openai implementations. I think things are best if you can participate. Good luck.

lmolkova: to summarize my understanding (whether you see it or not): you wanted other people to take a look and share the feedback. You don't want things to be sent as PRs without wide discussion. You are not interested in discussing technical details and expect others to discuss them. 
I assumed you wanted me to take some actions (remove `ToolCall` type description?), but it seems the main intention was to raise awareness,. Good to know.

The best discussion indeed happens when people implement things. Since there were no concerns/discussions on this in this PR (or in LLM calls), I assume there are no major concerns. 

This spec stays experimental for the time being and, as in other specs, we'll adjust things based on the implementation feedback. Implementation in a diverse set of languages and specific products is a prereq to stability.
If anyone discovers anything that feels wrong, confusing, suboptimal, etc, please share now or later once you're implementing it.

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,379 @@
...
+
+## User message
+
+This event describes the prompt message specified by the user.
+
+The event name MUST be `gen_ai.user.message`.
+
+| Body Field | Type | Description | Examples | Requirement Level |
+|---|---|---|---|---|
+| `content` | `AnyValue` | The contents of the user message. | `What telemetry is reported by OpenAI instrumentations?` | `Opt-In` |
```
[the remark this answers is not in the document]
lmolkova: I agree, it would be more future-proof to have `AnyValue` for system message as well. 

The user and system messages are the same for chat completion, but are different for assistants, where [user or assistant messages](https://platform.openai.com/docs/api-reference/messages/createMessage#messages-createmessage-content) are structured:


```json
"content": [
    {
      "type": "text",
      "text": {
        "value": "How does AI work? Explain it in simple terms.",
        "annotations": []
      }
    }
  ],
```

and [system messages (instructions)](https://platform.openai.com/docs/api-reference/assistants/createAssistant#assistants-createassistant-instructions) are not 

```json
"instructions": "You are a personal math tutor. When asked a question, write and run Python code to answer the question.",
```

But that's probably just a question of time until we need structured instructions too.

docs/gen-ai/gen-ai-events.md
lmolkova: I wanted to have a different name pattern for choice, because choice is not a message - it wraps assistant message and includes some metadata about it.

E.g.

```json
{
    "index": 0,
    "finish_reason": "stop"
    "message": {
      "role": "assistant",
      "content": "\n\nHello there, how may I assist you today?",
    },
  }
```

I feel calling it a `message` would be misleading. 

Happy to hear any suggestions or thoughts

docs/gen-ai/gen-ai-events.md
lmolkova: good catch, thanks!

docs/gen-ai/gen-ai-events.md
trentm: I don't have a strong opinion either way. It felt like an accidental omission given the "Naming pattern" section above:

> The generic events applicable to multiple GenAI models follow `gen_ai.{type}.message` naming pattern.
> GenAI vendor-specific instrumentations SHOULD follow `gen_ai.{gen_ai.system}.{type}.message` pattern to record system-specific events.

and this section is called "Choice message".

I agree with you that having a "message" Body property of an event named `(something).message` would be odd.

lmolkova: good catch on the naming pattern - I'll try to rephrase it

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,380 @@
...
+
+This event describes the assistant message.
+
+The event name MUST be `gen_ai.assistant.message`.
+
+| Body Field   | Type                           | Description                            | Examples                                        | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) |
+|--------------|--------------------------------|----------------------------------------|-------------------------------------------------|-------------------|
+| `content`    | `AnyValue`                     | The contents of the assistant message. | `Spans, events, metrics defined by the GenAI semantic conventions.` | `Opt-In` |
+| `tool_calls` | [ToolCall](#toolcall-object)[] | The tool calls generated by the model, such as function calls. | `[{"id":"call_mszuSIzqtI65i1wAUOE8w5H4", "function":{"name":"get_link_to_otel_semconv", "arguments":{"semconv":"gen_ai"}}, "type":"function"}]` | `Conditionally Required`: if available |
+
```
trentm, line 133: (This is perhaps a question that could get sorted out later.)

I notice that Anthropic tool calls (they call them `tool_use`) are included in the array of `content`, e.g.:

https://docs.anthropic.com/en/docs/build-with-claude/tool-use#example-of-empty-tool-result
```json
{
  "id": "msg_01Aq9w938a90dw8q",
  "model": "claude-3-5-sonnet-20240620",
  "stop_reason": "tool_use",
  "role": "assistant",
  "content": [
    {
      "type": "text",
      "text": "<thinking>I need to use the get_weather, and the user wants SF, which is likely San Francisco, CA.</thinking>"
    },
    {
      "type": "tool_use",
      "id": "toolu_01A09q90qw90lq917835lq9",
      "name": "get_weather",
      "input": {"location": "San Francisco, CA", "unit": "celsius"}
    }
  ]
}
```

Should this section state a preference on how to handle that? Should the `content` be kept in full *and* the `tool_calls` duplicate some of that info?

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,380 @@
...
+### `ToolCall` object
+
+| Body Field | Type                        | Description                        | Examples                                        | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) |
+|------------|-----------------------------|------------------------------------|-------------------------------------------------|-------------------|
+| `tool_call_id` | string                  | The id of the tool call            | `call_mszuSIzqtI65i1wAUOE8w5H4`                 | `Required`        |
```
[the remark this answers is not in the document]
lmolkova: > I suppose it was to match the tool_call_id field name used in the gen_ai.tool.message event?

Yes, I don't have a strong preference between `tool_call_id` and `id`, but we should stay consistent within this doc.
I'm happy to change to `id` if it's better for perf and to keep things simple.

In the long term, I don't think it's safe to assume that specific client lib object  (like `msg.tool_calls`) will stay compatible with our event definition, won't add extra properties or won't change something around serialization. Also, we might not want to capture it all as is.

In any case, I'll update the PR to use `id`.

docs/gen-ai/gen-ai-events.md
lmolkova: We should add a separate file for anthropic that would describe how all genai semconv would apply to anthropic. 

e.g. we use this pattern for dbs or messaging systems: 

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/elasticsearch.md contains all you need to know about elasticsearch database spans/instrumentation.

Or https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/kafka.md documents what applies to kafka and adds some examples.

We could describe how exactly events are recorded for anthropic there, but I believe they fit nicely into the type system defined in this PR:

- choice event will contain both - the `content` and `tool_calls`: 

```json
{
  ...
  "message":  {
    "content": "<thinking>I need to call the get_link_to_otel_semconv function, and the user wants genai, which is likely GenAI.</thinking>",  
    "tool_calls":  [ { 
        "id":"toolu_01A09q90qw90lq917835lq9",
        "function":{"name":"get_link_to_otel_semconv","arguments":"{\"semconv\":\"GenAI\"}"},"type":"tool_use"
       }
     ]
  }
}
```

Based on docs, it seems it won't return more than one content block 
> When Claude decides to use one of the tools you’ve provided, it will return a response with a stop_reason of tool_use and one or more tool_use content blocks in the API response...

- when tool result it sent back to the model, we'll record `gen_ai.assistant.message` with the message object above and also the `gen_ai.tool.message` with the result of the tool:

```json
{"content":"opentelemetry.io/semconv/gen-ai","id":"toolu_01A09q90qw90lq917835lq9"}
```

If there are multiple tools, each response should be submitted as individual event.

The full example would look like this:

1. Chat completion resulting in a tool call.

   |   Attribute name    |                     Value                             |
   |---------------------|-------------------------------------------------------|
   | Span name           | `"chat claude-3-5"`                             |
   | `gen_ai.system`     | `"anthropic"`                                            |
   | `gen_ai.request.model`| `"claude-3-5`                                           |
   | `gen_ai.request.max_tokens`| `200`                                          |
   | `gen_ai.request.top_p`| `1.0`                                               |
   | `gen_ai.response.id`| `"msg_01Aq9w938a90dw8q"`            |
   | `gen_ai.response.model`| `"claude-3-5"`                                     |
   | `gen_ai.response.finish_reasons`| `["tool_calls"]`                          |

   Events parented to this span:

   - `gen_ai.user.message` 

     |   Property          |                     Value                             |
     |---------------------|-------------------------------------------------------|
     | Event body          | `{"content":"How to instrument antropic client with OTel?"}` |

   - `gen_ai.choice`

     |   Property          |                     Value                             |
     |---------------------|-------------------------------------------------------|
     | Event body    | `{"finish_reason":"tool_calls","message":{"content": "<thinking>I need to call the get_link_to_otel_semconv function, and the user wants antropic , which is likely Anthropic .</thinking>",  "tool_calls":[{"id":"toolu_01A09q90qw90lq917835lq9","function":{"name":"get_link_to_otel_semconv","arguments":"{\"semconv\":\"Anthropic\"}"},"type":"tool_use"}]}}` |

2. Application executes the tool call. Application may create span which is not covered by this semantic convention.
3. Final chat completion call

   |   Attribute name                |                     Value                             |
   |---------------------------------|-------------------------------------------------------|
   | Span name                       | `"chat claude-3-5"`                                        |
   | `gen_ai.system`                 | `"anthropic"`                                            |
   | `gen_ai.request.model`          | `"claude-3-5"`                                             |
   | `gen_ai.request.max_tokens`     | `200`                                                 |
   | `gen_ai.request.top_p`          | `1.0`                                                 |
   | `gen_ai.response.id`            | `"msg_01Aq9w938a90dw8q"`            |
   | `gen_ai.response.model`         | `"claude-3-5"`                                        |
   | `gen_ai.response.finish_reasons`| `["stop"]`                                            |

   Events parented to this span:
   (in this example, the event content matches the original messages, but applications may also drop messages or change their content)

   - `gen_ai.user.message` ...
   - `gen_ai.assistant.message`

     |   Property                       |                     Value                                                                                                                  |
     |----------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------|
     | Event body  | `{"content": "<thinking>I need ...", "tool_calls":[{"id":"toolu_01A09q90qw90lq917835lq9","function":{"name":"get_link_to_otel_semconv","arguments":"{\"semconv\":\"GenAI\"}"},"type":"tool_use"}]}` |

   - `gen_ai.tool.message`

     |   Property                       |                     Value                                                                      |
     |----------------------------------|------------------------------------------------------------------------------------------------|
     | Event body    | `{"content":"opentelemetry.io/semconv/gen-ai","id":"toolu_01A09q90qw90lq917835lq9"}` |

lmolkova: Maybe I should rename `gen_ai.tool.message` to `gen_ai.tool_respose.message` or  `gen_ai.tool_result.message` so it's more obvious that it represents the outcome and not the call?

docs/gen-ai/gen-ai-events.md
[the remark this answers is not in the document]
lmolkova: done

docs/gen-ai/gen-ai-events.md
lmolkova: I renamed "naming pattern" to "custom events" and removed general part - there is no need to describe naming pattern for existing events. I kept the suggestion for vendor-specific ones. LMK if it helps

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,380 @@
...
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## System event
+
+This event describes the instructions passed to the GenAI model.
+It's NOT RECOMMENDED to report this event when capturing content is disabled, unless there are additional system-specific fields
```
xrmx, line 99: Is this specific to the system event or other events? e.g. the user events

xrmx: I think it's fine, thanks for the clarification

karthikscale3:
Looks like it's going to be Logs API for emitting Events going forward and they are recommending interoperability with other log sinks. Should we note Logs API instead of Event API?
https://github.com/open-telemetry/oteps/pull/265/files

lmolkova:
> Looks like it's going to be Logs API for emitting Events going forward and they are recommending interoperability with other log sinks. Should we note Logs API instead of Event API? https://github.com/open-telemetry/oteps/pull/265/files

OTel events will remain events, but will likely be emitted by  using `Logger` instead of `EventLogger` class. It does not change anything for semantic conventions or this PR.

docs/gen-ai/gen-ai-events.md
lmolkova: I removed this sentence to simplify things.

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,378 @@
...
+# Semantic Conventions for GenAI events
+
+**Status**: [Experimental][DocumentStatus]
+
+<!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->
+
+<!-- toc -->
+
+- [Common attributes](#common-attributes)
+- [System event](#system-event)
```
lmolkova, line 14: add role to the attribute or to the payload with actual system-specific role

mattmccleary (approved):

nirga (approved):
🎉

TaoChenOSU (approved):

karthikscale3 (approved):
👏🏽

docs/gen-ai/gen-ai-events.md
```diff
@@ -0,0 +1,383 @@
...
+## Tool event
+
+This event describes the output of the tool or function submitted back to the model.
+
+The event name MUST be `gen_ai.tool.message`.
+
+| Body Field     | Type   | Description                                   | Examples                        | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) |
+|----------------|--------|-----------------------------------------------|---------------------------------|-------------------|
+| `role`         | string | The actual role of the message author as passed in the message. | `"tool"`, `"function"` | `Conditionally Required`: if available and if not equal to `tool` |
+| `content`      | string | The contents of the tool message.             | `opentelemetry.io`              | `Opt-In`          |
```
yeesian, line 154: For my understanding, why is `content` (in Tool event) not `AnyValue` (for consistency with "User event" and "Assistant event")?

codefromthecrypt (approved):
FWIW: tested internally with js and soon python

Internal testing at ES had `enableContentCapture` to capture events or not (regardless of the semantic) and choice of mechanism via `eventsSemconv` (log vs span or default to whatever semver implies)

docs/gen-ai/gen-ai-events.md
lmolkova: good point, let's convert it to AnyValue too so it can be extended in the future

jsuereth (approved):

yeesian (approved):

Issue #535, opened 2023-11-16
Semantic conventions for go runtime metrics


dashpole:
For runtime semantic conventions, we currently only have JVM metrics defined: https://github.com/open-telemetry/semantic-conventions/tree/main/docs/runtime#semantic-conventions-for-runtime-environment.  The Go SIG (discussed on 11/16/23) would like to agree on conventions in this repository before expanding the set of metrics offered by the go runtime instrumentation package: https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/runtime#pkg-overview to minimize the churn to users.

Additional context:

* There is a separately maintained (thanks @jmacd!) package for go runtime metrics with an expanded set of supported metrics that may also help inform these conventions: https://github.com/lightstep/otel-launcher-go/tree/main/lightstep/instrumentation/runtime

Starting Questions/potential issues:

* Which metrics from the go runtime are stable?  What are the stability guarantees?
* Which metrics should be on-by-default?  Go is possibly considering recommending a default set: https://github.com/golang/go/issues/57175#issuecomment-1804230372.
* Go runtime metrics are currently named `runtime.go.*`, which conflicts with the runtime metrics guidance: `Metrics specific to a certain runtime environment should be prefixed with the runtime's top-level namespace {environment}.*, e.g., jvm.* `

@MrAlias @pellared

trask:
> Which metrics from the go runtime are stable? What are the stability guarantees?

in case the prior art helps, we are putting the breaking JVM metrics changes that we are implementing as part of JVM metric stability behind `OTEL_SEMCONV_STABILITY_OPT_IN=jvm` (and will make them the default in upcoming Java Instrumentation 2.0)

Issue #316, opened 2023-09-12
Provide some guidance for coming up with default histogram buckets for various metrics


trask:
Coming up with default histogram buckets for various metrics can be challenging, as we've seen in #274.

I'd like to propose some guidance for coming up with default histogram buckets which we can lean on in each time the issue of defining default histogram buckets comes up.

Proposal:
* If values align with service timings, use our default buckets, translated from millis to seconds, `[ 0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10 ]` [(dropping the zero bucket?)](https://github.com/open-telemetry/semantic-conventions/issues/298)
* Otherwise determine roughly the smallest and largest buckets that you care about, e.g. <0.01 seconds and >10 seconds, and use an exponential range in between (consider using base 2 if you need higher granularity and base 10 if you need lower granularity).

These would only be recommendations, and so if there is compelling reason to come up with a completely custom set of buckets for a particular metric that would always be ok.

trask:
another option could be to follow the pattern `[ .1, .25, .5, .75, 1]` in between the low and high buckets, e.g.

`[ 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10 ]`

and another similar option with less buckets:

`[ 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10 ]`

or even fewer:

`[ 0.01, 0.05, 0.1, 0.5, 1, 5, 10 ]`

jack-berg:
Generalizing your last comment into an algorithm we could encode into the spec might look like:

* Start with base 10 "outer" bucket boundaries.
* How many and which base 10 orders of magnitude do measurements typically span? E.g. two orders of magnitude (three buckets) starting at 1 would be `[1, 10], or `[10, 100]` if starting at ten. Three order of magnitude (four buckets) starting at 1 would be `[1, 10, 100]` or `[10, 100, 1_000]` if starting at ten.
* Decide how many inner buckets you need to get a decent distribution useful to most users most of the time. For each outer bucket boundary (skipping the first), the even bucket boundaries are distributed evenly between `[0, out_bucket_bound]`. Maybe we constrain it to multiples of two so the buckets boundaries don't fall on numbers with infinitely repeating decimals. For example:
[...]

Issue #269, opened 2023-08-17
Move existing Kafka metrics to a JMX specific namespace


pyohannes:
### Problem

The [Kafka metrics in the messaging semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/kafka.md#general-metrics) are based on metrics as they are reported by the JMX reporter. Although the JMX reporter is the [default reporter](https://docs.confluent.io/platform/current/kafka/monitoring.html), different reporters can be configured.

In their current form, the definition of those metrics causes confusion, as they suggest being an authoritative definition of metrics as they are reported by Kafka. However, they look different from the [metrics in the official Kafka documentation](https://kafka.apache.org/documentation/#monitoring), and they are different from the [OTLP metrics defined in KIP-714](https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability#KIP714:Clientmetricsandobservability-Standardproducermetrics).

Based on feedback from @AndrewJSchofield, people involved in the Kafka project currently don't have an interest in defining and maintaining a definition of Kafka metrics as part of the OTel semantic conventions.

### Desired solution

The semantic conventions should make it clear that the existing metrics are JMX specific, ideally in the metric name. This allows support for conventions for Kafka metrics that are sent via a different reporter.

Even more importantly, this avoids confusion, as _it makes clear that the OTel semantic conventions are not the place that provides an authoritative definition of Kafka metrics_.

pyohannes:
> The lack of mapping to semantic conventions doesn't mean this data is not useful.

I fully agree, [.NET runtime metrics](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Runtime#metrics) are a great example of this.

The conception that telemetry data needs to be covered by OTel semantic conventions in order to be useful leads to unwanted side effects:
1. Semantic conventions are written to comply with existing instrumentation, whereas it should be the other way round.
2. As a consequence of the point above, we end up with a bunch of experimental semantic conventions with no clearly defined ownership.

My take is that telemetry conventions referring to a specific component (e. g. Kafka broker, or .NET runtime) _can_ be part of OTel semantic conventions, but can as well live apart of them. In any case, they should be maintained by people working on that specific component, and not by the wider community.

joaopgrassi:
[...]

PR #981, opened 2024-04-29
Add experimental go runtime metrics semantic conventions

dashpole:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/535

## Changes

Add experimental runtime metrics based on the proposed set of recommended metrics from the Go team: https://github.com/golang/go/issues/67120

Please direct feedback on the _set_ of metrics on the issue above.  If the recommended set is changed, I will update these conventions.  Feedback on the OTel representation of those metrics is welcome.

## Alternatives considered

**Separate metrics for memory usage**

Instead of a single go.memory.usage metric, we could have one metric for the total, and other metrics for the stack and released amounts.  Separate metrics would potentially be more "future proof", if a new sub-set of memory that overlapped with released or stack memory was ever introduced.  I chose the existing design because using labels is easier to visualize, and understand for users.

**go.gc.\* instead of go.memory.gc.***

This PR uses `go.memory.gc.*`.  Instead, we could use `go.gc.*`, similar to java's `jvm.gc.duration`.  I went this direction because it is garbage collecting memory, and I felt it was most appropriate to compare the GC target to the `go.memory.usage` and `go.memory.limit` metrics.

**Different naming for "total" memory metrics**

`/gc/heap/allocs:bytes` and `/gc/heap/allocs:objects` from the go runtime are both cumulative totals for number and size of allocations made.  I've opted to name them `go.memory.allocated` and `go.memory.allocations` to be concise, but users may think `go.memory.allocated` sounds like the same thing as `go.memory.usage`.  Some alternative names I considered:

* `go.memory.size_allocated` and `go.memory.objects_allocated`

Note that we cannot use the `count` or `total` suffixes per https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#naming-rules-for-counters-and-updowncounters

**go.config.gomemlimit instead of go.memory.limit**

To be consistent with the other configuration metric, `go.config.gogc`, we could name the memory limit `go.config.gomemlimit`.  I've opted for the `go.memory.limit` naming because it aligns with `jvm.memory.limit`, and because it helps users understand that they should compare memory usage with the limit.

@open-telemetry/go-approvers @felixge @mknyszek


model/metrics/go-metrics.yaml
```diff
@@ -0,0 +1,87 @@
...
+    brief: "Heap size target for the end of the GC cycle."
+    instrument: updowncounter
+    unit: "By"
+    prefix: go.gc
+    stability: experimental
+
+  - id: metric.go.memory.gc.user_target
+    type: metric
+    metric_name: go.memory.gc.user_target
+    brief: "Heap size target percentage for the end of the GC cycle, as configured by the user."
```
reyang, line 67: Why is this "percentage"? (the unit says "By")

dashpole: Sorry, forgot to push my changes.  The unit is now "1", and the description says "ratio"

model/metrics/go-metrics.yaml
```diff
@@ -0,0 +1,88 @@
+groups:
+  - id: metric.go.memory.used
+    type: metric
+    metric_name: go.memory.used
+    brief: "Virtual memory mapped by the Go runtime."
```
reyang, line 5: Does "mapped" mean committed or reserved?

reyang (approved):
LGTM with some non-blocking clarification questions.

docs/runtime/go-metrics.md
```diff
@@ -0,0 +1,173 @@
...
+
+This metric is [recommended][MetricRecommended].
+This metric is obtained from Go's [`runtime/metrics`][RuntimeMetrics] package using `/gc/gogc:percent`.
+
+<!-- semconv metric.go.memory.gc.user_target(metric_table) -->
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `go.memory.gc.user_target` | UpDownCounter | `1` | Heap size target ratio for the end of the GC cycle, as configured by the user. [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1]:** The value range is [0.0,1.0].
```
MrAlias, line 129: Why not use the native `[0.0, 100.0]` exported by the runtime package?

trask:
> Add experimental runtime metrics based on the proposed set of recommended metrics from the Go team: https://docs.google.com/document/d/1NBeEYxld5lA_UUXoFniBO0a6_ryHYgFC0hgUAjpfdDY/edit?resourcekey=0-NBd-0IZAYTtiPej124rR2w&tab=t.0

Can this doc be made publicly accessible?

docs/runtime/go-metrics.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Metric: `go.memory.limit`
+
+This metric is [recommended][MetricRecommended].
+This metric is obtained from Go's [`runtime/metrics`][RuntimeMetrics] package using `/gc/gomemlimit:bytes`.
+
+<!-- semconv metric.go.memory.limit(metric_table) -->
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `go.memory.limit` | UpDownCounter | `By` | Go runtime memory limit configured by the user, otherwise math.MaxInt64. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
dmathieu, line 67: How is this memory limit retrieved? `GOMEMLIMIT`?

`math.MaxInt64` seems a bit unhelpful if someone hasn't explicitly set a limit.
Could this be the available memory? (it may cause performance issues to retrieve it though)

felixge (commented):
Thanks for raising this @dashpole . But as mentioned by others, the document you linked hasn't been made public yet. It might also make sense to allow some time for discussing the contents of the document in the Go community. (But since your PR here uses "status: experimental" I guess I don't feel strongly about it)

docs/runtime/go-metrics.md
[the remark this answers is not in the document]
felixge: >  Given the computationally expensive nature of retrieving these

Retrieving `runtime/metrics` in Go is pretty [light weight](https://github.com/golang/go/blob/16ce8b3925deaeb72541ee96b6ee23a08fc21dea/src/runtime/metrics.go#L830-L842). Or are you talking about the computational expense of sending metrics in general?

docs/runtime/go-metrics.md
```diff
@@ -0,0 +1,173 @@
...
+|---|---|---|---|---|---|
+| `go.memory.type` | string | The type of memory. | `released`; `stack` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+`go.memory.type` MUST be one of the following:
+
+| Value  | Description | Stability |
+|---|---|---|
+| `released` | Memory that is completely free and has been returned to the underlying system. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `stack` | Memory allocated from the heap that is reserved for stack space, whether or not it is currently in-use. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `other` | All memory mapped by the Go runtime into the current process as read-write, excluding other categories of memory usage. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
felixge: I agree, the calculation of `other` seems under-specified. I'm assuming it is:

```
/memory/classes/total:bytes - /memory/classes/heap/released:bytes - /memory/classes/heap/stacks:bytes
```

If that's the case, I'm not sure `released` should be included as a memory.type here.

docs/runtime/go-metrics.md
felixge: > How is this memory limit retrieved? GOMEMLIMIT?

I think via a `runtime/metrics.Read()` call for `/gc/gomemlimit:bytes`.

docs/runtime/go-metrics.md
```diff
@@ -0,0 +1,173 @@
...
+<!-- endsemconv -->
+
+<!-- semconv metric.go.memory.allocations(full) -->
+<!-- endsemconv -->
+
+## Go Garbage Collection
+
+**Description:** Go metrics captured under the namespace `go.memory.gc.*`
+
+### Metric: `go.memory.gc.target`
```
felixge, line 105: Is there a reason for using a different naming schema in this proposal than what is used by `runtime/metrics` itself? E.g. why is this not the `go.memory.gc.goal` metric or something similar?

docs/runtime/go-metrics.md
dashpole: There are no other % metrics I can find in the semantic conventions.  All others are ratios.  For example, process.cpu.utilization is a ratio, rather than a percentage. It also aligns with the prometheus conventions of using base units.

dashpole:
> But as mentioned by others, the document you linked hasn't been made public yet. It might also make sense to allow some time for discussing the contents of the document in the Go community. (But since your PR here uses "status: experimental" I guess I don't feel strongly about it)

I'm hoping this will help provide feedback on the proposal.  These conventions will be kept experimental until the proposal is finalized.

docs/runtime/go-metrics.md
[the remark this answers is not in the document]
MrAlias: I guess I'm more concerned about polling. Having these defined as synchronous instruments means they will need to be detected at a polling cycle. Should this be done once per collection cycle or are we expecting the instrumentation to run its own cycle?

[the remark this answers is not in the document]
dashpole: @MrAlias see https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#instrument-types.  "Counter" here can mean either sync or async, and it is up to us how we implement it.

docs/runtime/go-metrics.md
dashpole: changed target to goal.  I was trying to balance consistency with other otel metrics (mostly the current JVM metrics) with aligning with the go runtime naming.  Let me know if anything else stands out to you as off.

My hope was also to combine related go runtime metrics together, where appropriate, particularly for memory usage metrics, as metrics with labels are generally easier to use individual counters.

docs/runtime/go-metrics.md
dmathieu: Right. My concern was more around the use of `math.MaxInt64` as the default limit, which doesn't provide much value. Could we use `/proc/meminfo` to retrieve the total amount of memory instead?

docs/runtime/go-metrics.md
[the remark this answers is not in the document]
MrAlias: Oh, gotcha. :+1:

docs/runtime/go-metrics.md
MrAlias: I don't seem to be able to resolve this. Feel free to if you can.

model/metrics/go-metrics.yaml
dashpole: This metric is now no longer virtual memory usage (VSS), and is instead the amount of memory used by the program.  I believe that makes it committed memory.

docs/runtime/go-metrics.md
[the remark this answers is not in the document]
dashpole: I removed released memory from this metric, and split it out separately.  That means `go.memory.usage` is now defined as `/memory/classes/total:bytes - /memory/classes/heap/released:bytes`, which is all memory usage subject to `go.memory.limit` (based on https://go.dev/doc/gc-guide).  I think that makes using the two metrics intuitive.

As far as what is now included in `go.memory.usage`, [MemStats](https://pkg.go.dev/runtime#MemStats) seems to have the clearest description, since it is Sys - HeapReleased.  It includes the heap (excluding released), the stack (including OS thread stacks; excluding any stacks allocated by C code), mpsan structures, mcache structures, profiling bucket hash tables, gc metadata, and miscellaneous off-heap runtime allocations.

docs/runtime/go-metrics.md
dashpole: I would prefer not recording a metric at all, if there is no limit imposed by the go runtime than trying to discover an external memory limit.  I've updated the PR to exclude max int-valued metrics.

mknyszek:
I have posted the proposal on the Go issue tracker; no need to look at the doc anymore. Please comment at https://github.com/golang/go/issues/67120.

docs/runtime/go-metrics.md
```diff
@@ -0,0 +1,218 @@
...
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `go.memory.gc.goal` | UpDownCounter | `By` | Heap size target for the end of the GC cycle. [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1]:** Computed from `/gc/heap/goal:bytes`.
+<!-- endsemconv -->
+
+<!-- semconv metric.go.memory.gc.goal(full) -->
+<!-- endsemconv -->
+
+### Metric: `go.memory.gc.user_goal`
```
mknyszek, line 144: Could this just be called `gogc` or `gcpercent`? The concept is fairly specific, so calling it "goal" feels misleading to me.

dashpole: Yeah, I agree we need a better name.  I was almost thinking it might be better to put metrics which track user-provided configuration in the OTel resource if they are constant for the lifetime of the program, or group them together in a single metric (e.g. `go.config` with labels `gogc = 100, gomemlimit=12345`).  Can gogc or gomemlimit be modified during the execution of the program?

mknyszek: Yeah, both can be modified during execution, and possibly with some frequency (the API is fairly responsive by design). That's another reason that exporting it is useful (though changing it frequently is less common). The relevant APIs are `runtime/debug.SetGCPercent` and `runtime/debug.SetMemoryLimit` for `GOGC` and `GOMEMLIMIT` respectively.

dashpole: updated to `go.config.gogc`.  I don't love that it isn't consistent with the memory limit (which is still go.memory.limit), but I think it is at least clearer what the metric is.

MrAlias (approved):

dmathieu (approved):

docs/runtime/go-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+<!-- endsemconv -->
+
+<!-- semconv metric.go.goroutine.count(full) -->
+<!-- endsemconv -->
+
+## Go Threads
+
+**Description:** Go metrics captured under the namespace `go.thread.*`
+
+### Metric: `go.thread.limit`
```
felixge, line 166: I think this should be `go.procs.limit`:

1. This is not a limit on threads, but a limit on procs, which is an important difference. Calling this out in the metric description is too subtle IMO.
2. Diverging from the names chosen by `runtime/metrics` is a level of indirection that should be avoided.
3. What if Go decides to impose an actual thread limit in the future and export a metric for it?

docs/runtime/go-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `go.memory.released` | UpDownCounter | `By` | Memory that is completely free and has been returned to the underlying system. [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1]:** Computed from `/memory/classes/heap/released:bytes`.
+<!-- endsemconv -->
+
+<!-- semconv metric.go.memory.released(full) -->
+<!-- endsemconv -->
+
+### Metric: `go.memory.limit`
```
felixge, line 79: As per [this thread](https://github.com/open-telemetry/semantic-conventions/pull/981#discussion_r1586600729), I think it would be better to have this under `go.config.memlimit`. 

The reason for this as follows:

1. It's a soft rather than hard limit. This clashes with common CNCF nomenclature in k8s where a memory limit is the point where a container gets killed or evicted.
2. What if Go introduces a hard limit in the future?
3. Having all config options under config makes the most sense to me.

I don't feel strongly about this, so feel free to ignore.

felixge (commented):
Thanks for the updates. A few more thoughts from my end!

docs/runtime/go-metrics.md
dashpole: WDYT of `go.memory.soft_limit` to mitigate points 1 and 2?  I also don't feel strongly, so we can leave this open to see if others have thoughts.

docs/runtime/go-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+**Description:** Go metrics captured under the namespace `go.schedule.*`
+
+### Metric: `go.schedule.duration`
+
+This metric is [recommended][MetricRecommended].
+
+<!-- semconv metric.go.schedule.duration(metric_table) -->
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `go.schedule.duration` | Histogram | `s` | The time goroutines have spent in the scheduler in a runnable state before actually running. [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
trask, line 192: for any histograms, you'll want to specify the default buckets, check out https://github.com/open-telemetry/semantic-conventions/issues/316#issuecomment-1783528869

docs/runtime/go-metrics.md
dashpole: Ah, I had used this naming based on the description: "number of operating system threads" from https://pkg.go.dev/runtime/metrics.  Is `go.process.limit` OK?  It aligns better with otel conventions.

docs/runtime/go-metrics.md
```diff
@@ -0,0 +1,221 @@
...
+
+`go.memory.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `stack` | Memory allocated from the heap that is reserved for stack space, whether or not it is currently in-use. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `other` | Memory used by the Go runtime, excluding other categories of memory usage. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- endsemconv -->
+
+### Metric: `go.memory.released`
```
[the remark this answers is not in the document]
dashpole: removed

docs/runtime/go-metrics.md
dashpole: Bucket boundaries are provided by the runtime, so we can't actually change them via views.  They appear to be:

* 6.399999999999999e-08
* 6.399999999999999e-07
* 7.167999999999999e-06
* 8.191999999999999e-05
* 0.0009175039999999999
* 0.010485759999999998
* 0.11744051199999998

dashpole: I'm asking if those are something that we can rely on: https://github.com/golang/go/issues/67120#issuecomment-2100709396

dashpole: Buckets are not stable, and will change across go versions.  Should we still document them here?

trask: Bucket changes are allowed in stable semconv (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md).

Btw, how are you planning to emit pre-aggregated buckets?

I think the semconv story is a bit murky when it's not treated as a specification, but rather a bridging of an underlying metric system, e.g. check out https://github.com/open-telemetry/semantic-conventions/issues/269#issuecomment-1686756064

dashpole: We were planning to expose histogram metrics using the MetricProducer interface.  This is something in-between a bridge to an underlying metric system and instrumentation.  Go runtime metrics are fairly verbose, and lack attributes, so there is at least some curation and transformation that is needed to make the metrics useful to OTel users.  OTOH, we are definitely re-exposing metric data here.

trask: Makes sense, and very similar to the JVM metrics story, I guess we just haven't dealt with preaggregated buckets before. Do you have the option to get individual GC durations that could then be pushed into our own shaped histogram? (Also to allow metric view buckets to work)

dashpole: We don't have that option today.

trask: Maybe then just doc the situation in semconv, explains that this histogram has preaggregated buckets from the underlying data source. 

Also not sure but may be worth mentioning that metric views won't work, since that may be a common confusion.

docs/runtime/go-metrics.md
felixge: `proc` is short for `processor` not `process`. So `go.processors.limit` would be okay I guess. But I think `procs` is the most common way to refer to this scheduler entity in the community.

dashpole: changed to processor for now. I notice the metric is named `/sched/gomaxprocs:threads`.  Does that mean the unit should be `{thread}`?

docs/runtime/go-metrics.md
dashpole: Documented here: https://github.com/open-telemetry/semantic-conventions/pull/981/commits/7b0717230641508d33896c2e18c553fbedff091d.  We can document that views won't work in the documentation for the package.

docs/runtime/go-metrics.md
dashpole: Raised this in the SIG meeting.  Additional thoughts/feedback from others mostly supported `go.memory.limit`:

* Don't love `soft_limit`, since people might be confused that it is named differently from GOMEMLIMIT. "I didn't know Go had a soft limit?  Is that the same as GOMEMLIMIT?"
* We could make the new metric `hard_limit`, if a hard limit was introduced in the future.
* How likely is it actually that a hard limit is added?
* Given the configuration is GOMEMLIMIT, `go.memory.limit` seems reasonably safe.  Go would need to choose something different e.g. GOHARDMEMLIMIT, etc to introduce a new limit anyways.

felixge: Your last comment mitigates my worries about 2. 1 and 3 still apply tho. Was the `go.config.memlimit` option I suggested discussed in the SIG meeting?

Anyway, I don't feel very strongly about this, so feel free to go ahead with `go.memory.limit` if you think it has some consensus behind it.

docs/runtime/go-metrics.md
felixge: > changed to processor for now. I notice the metric is named `/sched/gomaxprocs:threads`. Does that mean the unit should be `{thread}`?

I guess so. But I'd still stick with the term `processor` for the metric name.

dashpole: done

docs/runtime/go-metrics.md
dashpole: I raised it as one of the options.  I don't believe there were any specific objections, but that it didn't seem as useful to users.  We should have plenty of time to revisit it if the fact that it is actually a soft limit is confusing for users.

felixge: Ok, works for me.

docs/runtime/go-metrics.md
```diff
@@ -0,0 +1,205 @@
...
+
+`go.memory.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `stack` | Memory allocated from the heap that is reserved for stack space, whether or not it is currently in-use. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `other` | Memory used by the Go runtime, excluding other categories of memory usage. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- endsemconv -->
+
+### Metric: `go.memory.soft_limit`
```
felixge, line 63: I thought the conclusion from [this thread](https://github.com/open-telemetry/semantic-conventions/pull/981#discussion_r1593693033) was to go with `go.memory.limit`?

dashpole: yes, sorry.  Fixed.

felixge (approved):
LGTM now. Thanks for all the iterations 🙇

model/registry/go.yaml
```diff
@@ -0,0 +1,25 @@
...
+          members:
+            - id: stack
+              value: 'stack'
+              brief: 'Memory allocated from the heap that is reserved for stack space, whether or not it is currently in-use.'
+              note: >
+                Computed from `/memory/classes/heap/stacks:bytes`.
+              stability: experimental
+            - id: other
+              value: 'other'
+              brief: 'Memory used by the Go runtime, excluding other categories of memory usage.'
```
trisch-me, line 21: I don't understand the meaning of this phrase - what is `other categories of memory` ? I see here only stack above and that's all?

dashpole: Thats correct. It currently includes all memory except `stack`.  I phrased it that way to make it possible to extend the metric with future values.  If other categories are introduced (e.g. `heap`), the `other` category would then be all memory except `heap` and `stack`. It exists to ensure the total of all categories is the total memory usage.

trisch-me:
```suggestion
              brief: 'Memory used by the Go runtime, excluding other categories of memory usage described in this enumeration.'
```

what about this option?

dashpole: sure, thats fine.

trisch-me (approved):

dashpole:
@reyang anything else needed for this to merge?

reyang:
> @reyang anything else needed for this to merge?

Nope, you're all set 👍

PR #982, opened 2024-04-29
Define Nginx Ingress Controller events

gregkalapos:
## Changes

Following the events convention, adding definition for Nginx Ingress Controller events. All this information is from the [Elastic Nginx Ingress Controller Logs integration](https://docs.elastic.co/en/integrations/nginx_ingress_controller#access-logs).

### DIFF
--- a/docs/nginx/events.md
+++ b/docs/nginx/events.md
@@ -1,4 +1,4 @@
- Semantic conventions for Nginx Ingress Controller events
+Semantic conventions for Nginx Ingress Controller events
 
 **Status**: [Experimental](../../../document-status.md)
 
@@ -7,6 +7,9 @@
 <!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->
 
 <!-- toc -->
+
+- [Access](#access)
+- [Error](#error)
 
 <!-- tocstop -->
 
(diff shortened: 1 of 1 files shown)


docs/nginx/events.md
```diff
@@ -0,0 +1,54 @@
...
+## Access
+
+The event name MUST be `nginx_ingress_controller.access`.
+
+This event describes the details of an the Nginx access.
+
+| Body Field | Type | Description | Examples | Requirement Level |
+|---|---|---|---|---|
+| `http.request.id` | string | The randomly generated ID of the request | `529a007902362a5f51385a5fa7049884` | Required |
+| `http.request.length` | int | The request length (including request line, header, and request body) | `89` | Recommended |
```
felixbarny, line 29: We should be using the Semantic Convention field [`http.request.size`](https://github.com/open-telemetry/semantic-conventions/blob/02ecf0c71e9fa74d09d81c48e04a132db2b7060b/model/registry/http.yaml#L109) for this.

docs/nginx/events.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Access
+
+The event name MUST be `nginx_ingress_controller.access`.
+
+This event describes the details of an the Nginx access.
+
+| Body Field | Type | Description | Examples | Requirement Level |
+|---|---|---|---|---|
+| `http.request.id` | string | The randomly generated ID of the request | `529a007902362a5f51385a5fa7049884` | Required |
```
felixbarny, line 28: There's a [`http.request.id`](https://www.elastic.co/guide/en/ecs/master/ecs-http.html#field-http-request-id) field in ECS. I think we should bring that over to ECS and use that attribute instead.

felixbarny (commented):
Instead of using the integration specific fields form the Elastic integration, we should try to use equivalent Semantic Convention fields, add new Semantic Convention fields, or port over existing ECS fields to Semantic Conventions.

Overall, I'm not really sold on the idea to place the parsed content of nginx logs under body rather than attributes. It's not obvious to me when something should become an attribute vs added to the body. All of the content is coming from parsing the string `body` of stemming from a plain-text logs file, rather than a pre-existing structured event. Therefore, I don't think it makes sense to parse some things into attributes and others into a structured body. Whether or not an attribute is defined in Semantic Conventions or is applicable to other, similar technologies, such as Apache or HAProxy shouldn't matter whether something is an attribute or a log record body. It should be fine if a parsed log record of a specific technology has attributes specific to that technology.

docs/nginx/events.md
```diff
@@ -0,0 +1,54 @@
...
+| `response.time_list` | array | An array of upstream response durations. It is a list because it is common that several upstream servers were contacted during request processing. | | Recommended |
+| `upstream_address_list` |  array | An array of the upstream addresses. It is a list because it is common that several upstream servers were contacted during request processing. |  | Recommended |
+
+# Error
+
+The event name MUST be `nginx_ingress_controller.error`.
+
+This event describes the details of an the error in Nginx Ingress Controller.
+
+| Body Field | Type | Description | Examples | Requirement Level |
```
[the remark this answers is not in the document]
rogercoll: Or `source.file.log.level`

docs/nginx/events.md
```diff
@@ -0,0 +1,54 @@
...
+
+<!-- tocstop -->
+
+</details>
+
+This document describes the semantic conventions for events in Nginx Ingress Controller. All Nginx Ingress Controller events MUST use a namespace of `nginx_ingress_controller` in the `event.name` property.
+
+## Access
+
+The event name MUST be `nginx_ingress_controller.access`.
```
trask, line 22: can we link to the definition of the nginx fields below, e.g. https://nginx.org/en/docs/http/ngx_http_log_module.html#log_format?

docs/nginx/events.md
```diff
@@ -0,0 +1,54 @@
...
+This event describes the details of an the Nginx access.
+
+| Body Field | Type | Description | Examples | Requirement Level |
+|---|---|---|---|---|
+| `http.request.id` | string | The randomly generated ID of the request | `529a007902362a5f51385a5fa7049884` | Required |
+| `http.request.length` | int | The request length (including request line, header, and request body) | `89` | Recommended |
+| `http.request.time` | double | Time elapsed since the first bytes were read from the client | `0.001`| Recommended |
+| `remote_ip_list` | array | An array of remote IP addresses. It is a list because it is common to include, besides the client IP address, IP addresses from headers like X-Forwarded-For. Real source IP is restored to source.ip. | `["192.168.64.1"]` | Recommended |
+| `upstream.alternative_name` | string | The name of the alternative upstream. |  | Recommended |
+| `upstream.ip` | string | The IP address of the upstream server. If several servers were contacted during request processing, their addresses are separated by commas. | `172.17.0.5` |Recommended |
```
trask, line 33: In my brain things are opposite: `upstream.*` is the HTTP client and `downstream.*` is the service that gets called to process the request and returns the response back to nginx.

I did google and find https://en.wikipedia.org/wiki/Upstream_server though which matches the definition here.

Given what appears to be [widespread ambiguity](https://stackoverflow.com/a/59458012), could we avoid the term "upstream" somehow?

Maybe something like:

`client.*` = HTTP client
`server.*` = nginx server itself
`backend.*` = (what's referred to as "upstream" in this PR)

I guess it depends though on whether the goal is to define a (mostly) generic access log event, or to define a very specific nginx access log, in which case it could make more sense to stick with whatever terms / fields names nginx uses.

docs/nginx/events.md
```diff
@@ -0,0 +1,54 @@
...
+- [Access](#access)
+- [Error](#error)
+
+<!-- tocstop -->
+
+</details>
+
+This document describes the semantic conventions for events in Nginx Ingress Controller. All Nginx Ingress Controller events MUST use a namespace of `nginx_ingress_controller` in the `event.name` property.
+
+## Access
```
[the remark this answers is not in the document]
vedantshetty: Why not include these fields as event attributes?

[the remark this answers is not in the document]
trask: hi @vedantshetty! indeed, this is latest direction we are going with OpenTelemetry event semantic conventions, check out the recent https://github.com/open-telemetry/semantic-conventions/issues/1651#issuecomment-2532915486:

> we have a consensus on using attributes for everything structured we'd define in semantic conventions

trask:
linking: https://github.com/open-telemetry/semantic-conventions/pull/1434#issuecomment-2624195997

Issue #916, opened 2024-04-10
Move yaml schema definition to semantic conventions


lmolkova:
## Description

[Schema definition](https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/semconv.schema.json) for semconv yaml currently lives in the build tools repo along with the documentation that describes [yaml format](https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/syntax.md).

They should be part of semantic conventions and can be shared between different tooling (weaver/build-tools) and easily discoverable by semconv authors.

Additions/removals/changes should be tracked in the semconv repo as they affect all conventions and should be discussed/reviewed by wider community.

### Additional context

We'll need to find a good process to update semconv yaml schema: new schema features would still need to be implemented in tooling repos before they can be brought in to semconv.

lmolkova:
> We'll need to find a good process to update semconv yaml schema

The ways to achieve it could be:

1. Separate vCurrent and vNext schemas
- keep two separate files - `semconv.schema.json` (vCurrent) and `semconv.schema.next.json` in the semconv repo 
- update `semconv.schema.json` with new tooling release that adds some new features

Cons: hard to keep track of difference between vNext schema and what tooling actually supports. 
We don't want vNext schema to list features that are not coming in the next tooling version

2. Keep vCurrent in semconv and vNext in tooling (only weaver)
- when new tooling is released, `semconv.schema.json` in semconv is updated to one in the tooling repo

Cons: schema changes/proposals are less discoverable in the weaver.

3. Require schema changes to be done in parallel in the tooling and in the semconv
- Draft PR1 to semconv with schema change
- Draft PR2 to tooling to support the new feature
- Once there is a consensus on PR1 to proceed, PR2 can be reviewed and merged
- After PR2 is merged, PR1 can be merged as well

Cons: difficult process

jsuereth:
Is there an option where:

- We associate our *usage* of schema to a previous tag of semconv repository
- We allow updates to semconv.schema.json *and* specification of it, but we don't allow that change to be used in YAML until a release of semconv repository.

lmolkova:
@jsuereth I believe all of the options do this (with a different file, not tag, but the same).
The key question is how to keep imaginary vnext in sync with actual implementation.

I.e. if we got 10 contributions to schema in semconv on top of tag A, but implemented 2 of them in weaver, it'd be a problem when we're going to release tag A + 1

jsuereth:
[...]

PR #985, opened 2024-04-30
Update build-tools version to 0.25.0 and add `display_name` to all registry attribute groups

### LINKED ISSUE #978: Add `name` property on registry attribute groups
In order to automatically render readable titles for generated attributes registry docs, it would be useful to have display names for attribute groups in the YAML files (see https://github.com/open-telemetry/semantic-conventions/pull/917#discussion_r1580512059).

Proposal is to add a property `name` (or `display_name` or `title`) to each attributes group in the registry so that the weaver can use that property for rendering headings. 

Example:

```

groups:
  - id: registry.aws.dynamodb
    prefix: aws.dynamodb
    type: attribute_group
    name: AWS DynamoDB
    brief: >
            This document defines attributes for AWS DynamoDB.
    attributes:
         ...
```

AlexanderWert:
Fixes #978

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,7 +83,7 @@ environment configured:
 ### 1. Modify the YAML model
 
 Refer to the
-[Semantic Convention YAML Language](https://github.com/open-telemetry/build-tools/blob/v0.24.0/semantic-conventions/syntax.md)
+[Semantic Convention YAML Language](https://github.com/open-telemetry/build-tools/blob/v0.25.0/semantic-conventions/syntax.md)
 to learn how to make changes to the YAML files.
 
 #### Schema files
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ CHLOGGEN_CONFIG  := .chloggen/config.yaml
 
 # see https://github.com/open-telemetry/build-tools/releases for semconvgen updates
 # Keep links in model/README.md and .vscode/settings.json in sync!
-SEMCONVGEN_VERSION=0.24.0
+SEMCONVGEN_VERSION=0.25.0
 WEAVER_VERSION=0.5.0
 
 # From where to resolve the containers (e.g. "otel/weaver").
--- a/docs/attributes-registry/android.md
+++ b/docs/attributes-registry/android.md
@@ -6,8 +6,8 @@
 
 # Android
 
-- [Android](#android-attributes)
-- [Android Deprecated](#android-deprecated-attributes)
+- [Android Attributes](#android-attributes)
+- [Deprecated Android Attributes](#deprecated-android-attributes)
 
 ## Android Attributes
 
@@ -17,7 +17,7 @@ The Android platform on which the Android application is running.
 | ---------------------- | ------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------- | ---------------------------------------------------------------- |
 | `android.os.api_level` | string | Uniquely identifies the framework API revision offered by a version (`os.version`) of the android operating system. More information can be found [here](https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels). | `33`; `32` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
-## Android Deprecated Attributes
+## Deprecated Android Attributes
 
 This document defines attributes that represents an occurrence of a lifecycle transition on the Android platform.
 
--- a/docs/attributes-registry/aspnetcore.md
+++ b/docs/attributes-registry/aspnetcore.md
@@ -6,7 +6,7 @@
 
 # Aspnetcore
 
-## Aspnetcore Attributes
+## ASP.NET Core Attributes
 
 ASP.NET Core attributes
 
--- a/docs/attributes-registry/aws.md
+++ b/docs/attributes-registry/aws.md
@@ -7,7 +7,7 @@
 # AWS
 
 - [General AWS Attributes](#general-aws-attributes)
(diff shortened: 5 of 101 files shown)


AlexanderWert:
Depends on: https://github.com/open-telemetry/build-tools/pull/316

trisch-me (approved):
thanks for quick implementing it. @jsuereth could add this now into the final PR to populate correct name

ChrsMark (approved):
LGTM

trisch-me:
One question though - should we enforce then a presence of this property for registry?

AlexanderWert:
> One question though - should we enforce then a presence of this property for registry?

yeah, I guess we could extend https://github.com/open-telemetry/build-tools/pull/316 to include it in the validity checks

lmolkova (approved):
Could you please create an issue to update schema definition (or just update the schema definition) in the https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/semconv.schema.json to include it?

We want to move it to this repo https://github.com/open-telemetry/semantic-conventions/issues/916, but it's not done yet

noahfalk (approved):
LGTM

noahfalk:
@JamesNK (Just fyi - This looks like a minor cosmetic change)

jsuereth:
> One question though - should we enforce then a presence of this property for registry?

Once we get the rule-engine/checker working in weaver, this is a simple configuration for semantic-conventions.  Let's see if we can get that prioritized as a the next feature we need from weaver for semconv.

jsuereth (approved):
Thanks for adding all of these.

We need to update both build-tools and weaver.  I merged your build-tools change (will need a release) but we'll also need to update weaver.

Since the weaver PR merged, you should be able to actually update the weaver `attribute_group.md.j2` template to *use* this value in this PR.

AlexanderWert:
> Could you please create an issue to update schema definition (or just update the schema definition) in the https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/semconv.schema.json to include it?

Done, see: https://github.com/open-telemetry/build-tools/pull/317

joaopgrassi (approved):

AlexanderWert:
> Since the weaver PR merged, you should be able to actually update the weaver `attribute_group.md.j2` template to _use_ this value in this PR.

@jsuereth Done with the last commit.

lmolkova:
@AlexanderWert I think you need to run `make table-generation` to re-generate registry MD files with display name.

jsuereth:
@AlexanderWert Yes - I think that change (and propagation of its value) should be what you need to do in weaver.

There's also a TemplateGroup: https://github.com/open-telemetry/weaver/blob/main/crates/weaver_forge/src/registry.rs#L29 

Once you have Group / TemplateGroup updated and wired - we can immediately start rendering this attribute in the registry.

joaopgrassi:
I added the new field `display_name` to Weaver https://github.com/open-telemetry/weaver/pull/202. So when we have a new weaver release we can go further with this. 

The only thing is: Should we mark the new field as required? The Jinja template now assumes it is but if one attribute group misses it the md rendering will fail (I think).

AlexanderWert:
> The only thing is: Should we mark the new field as required? The Jinja template now assumes it is but if one attribute group misses it the md rendering will fail (I think).

Can we make it required only for registry attribute groups?

trisch-me:
we can enforce through the policies that this attribute is required for registry

joaopgrassi:
>Can we make it required only for registry attribute groups?

Hum I think not, specially in the JSON schema but also on Weaver. What we can do is in the jinja template we check if display name exists and if not, we use the old behavior. I gave that a try and it worked for me locally.

joaopgrassi:
> we can enforce through the policies that this attribute is required for registry

Ah really? Then forget my comment above 👍

trisch-me:
> Then forget my comment above

but until we have it implemented we can just use the workaround you proposed

joaopgrassi:
@AlexanderWert are you able to continue this work, or should one of us take over?

AlexanderWert:
> @AlexanderWert are you able to continue this work, or should one of us take over?

I'll update the PR with your proposal

AlexanderWert:
@joaopgrassi updated this PR

I think we need both releasing and updating the weaver version AND build tools, right?

joaopgrassi:
> I think we need both releasing and updating the weaver version AND build tools, right?

Weaver for sure we need a new release. For build-tools I'm not sure.. but I assume we might (maybe bcs of the JSON schema?)

@lquerel can we have a new Weaver release?

lmolkova:
Oh, I think we still need a build-tools release for this.

lquerel:
> > I think we need both releasing and updating the weaver version AND build tools, right?
> 
> Weaver for sure we need a new release. For build-tools I'm not sure.. but I assume we might (maybe bcs of the JSON schema?)
> 
> @lquerel can we have a new Weaver release?

@joaopgrassi The release v0.5.0 of Weaver contains this feature.

joaopgrassi:
Yes, we have the release and already updated Weaver version in semconv. @AlexanderWert will update the PR because it needs some changes before merging.

AlexanderWert:
@joaopgrassi @lmolkova @lquerel 
Updated my PR.

PTAL, I also changed the order of the generated groups to make sure deprecated attributes are always listed at the end (instead of pure alphabetical order)

AlexanderWert:
@lmolkova The compatibility check still fails.

That has been actually fixed in these PRs, but we did not have a build-tools release since then:
https://github.com/open-telemetry/build-tools/pulls?q=is%3Apr+display_name+is%3Aclosed

So, we need to do a new release of the build tools and update it here.

lmolkova:
@AlexanderWert build-tools v0.25.0 were just published, could you try updating? Thanks!

PR #681, opened 2024-02-01
Move process metric attributes to the registry

braydonk:
Fixes #650

## Changes

The guidance is for all attributes to be moved to global registry instead of being in metric specification files. This PR moves all attributes from `process.yaml` spec to the Process registry.

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x] [CHANGELOG.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md) updated for non-trivial changes.
* [N/A] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

Issue #840, opened 2024-03-27
Consider merging `*.cpu.state` metric attributes


ChrsMark:
### Area(s)

area:container
area:system
area:process

### Is your change request related to a problem? Please describe.

At the moment we have 3 different metric attributes representing the `cpu.state`:

- [`system.cpu.state`](https://github.com/open-telemetry/semantic-conventions/blob/4368358cdcdd78f37291317a31cbc413080738ec/model/metrics/system-metrics.yaml#L20)
- [`process.cpu.state`](https://github.com/open-telemetry/semantic-conventions/blob/4368358cdcdd78f37291317a31cbc413080738ec/model/metrics/process-metrics.yaml#L7)
- [`container.cpu.state`](https://github.com/open-telemetry/semantic-conventions/pull/282#discussion_r1528808420)

Based on the global flat registry idea it would be nice if we had one single definition for this attribute and then be able to re-use it in the various namespaces.

### Describe the solution you'd like

As it is explained at https://github.com/open-telemetry/semantic-conventions/pull/681#discussion_r1511310673 we could consider merging into one single one called `cpu.state` similar to the [`disk.io.direction`](https://github.com/open-telemetry/semantic-conventions/blob/4368358cdcdd78f37291317a31cbc413080738ec/model/registry/disk.yaml#L8) and [`network.io.direction`](https://github.com/open-telemetry/semantic-conventions/blob/4368358cdcdd78f37291317a31cbc413080738ec/model/registry/network.yaml#L191).

### Describe alternatives you've considered

Maybe the `embed` concept would be sth to think of here: https://github.com/open-telemetry/build-tools/issues/240

cc @open-telemetry/semconv-system-approvers @open-telemetry/semconv-container-approvers @open-telemetry/semconv-k8s-approvers

ChrsMark:
So to summarize the situation here, we have the following:

## System Metrics

Values for [`system.cpu.state`](https://github.com/open-telemetry/semantic-conventions/blob/4368358cdcdd78f37291317a31cbc413080738ec/model/metrics/system-metrics.yaml#L20):
- `user`
- `system`
- `nice`
- `idle`
- `iowait`
- `interrupt`
- `steal`

## Process Metrics

Values for [`process.cpu.state`](https://github.com/open-telemetry/semantic-conventions/blob/4368358cdcdd78f37291317a31cbc413080738ec/model/metrics/process-metrics.yaml#L7):
- `user`
- `system`
- `wait`

[...]

PR #988, opened 2024-04-30
[chore] move process metrics to the registry

trisch-me:
Replacement for #681 which is stale and outdated


lmolkova (approved):

model/registry/process.yaml
```diff
@@ -170,3 +170,53 @@ groups:
...
+          members:
+            - id: voluntary
+              value: 'voluntary'
+              stability: experimental
+            - id: involuntary
+              value: 'involuntary'
+              stability: experimental
+        stability: experimental
+      - id: paging.fault_type
+        brief: >
```
lmolkova, line 186:
```suggestion
        brief: >
```

docs/attributes-registry/process.md
```diff
@@ -34,4 +36,32 @@
...
+| Value  | Description | Stability |
+|---|---|---|
+| `major` | major | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `minor` | minor | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 <!-- endsemconv -->
+
+<!-- semconv registry.process.cpu(omit_requirement_level) -->
+| Attribute  | Type | Description  | Examples  | Stability |
+|---|---|---|---|---|
+| `process.cpu.state` | string | The CPU state for this data point. A process SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels. | `system` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
ChrsMark, line 58: Shall we honor the signal-generic rule here?
This was discussed at https://github.com/open-telemetry/semantic-conventions/pull/681/files#r1476037740 and we have followed that so far in other namespaces as well like in `system` and `container`.

trisch-me: we have discussed it here https://github.com/open-telemetry/semantic-conventions/pull/681/files#r1484331195
with resolution to leave as is. Do you think we need to rewrite it?

ChrsMark: I wouldn't consider this as a clear and future-proof resolution 🤔 .
I would prefer being consistent for now and go with the overwrite approach like we did in `system` and `container`. 
That will also bring us less friction at https://github.com/open-telemetry/semantic-conventions/issues/840.

joaopgrassi: Yeah I'd prefer if we keep things in the registry signal-agnostic, since the whole point of it is sharing the attribute across signals.

trisch-me: I have removed description from global registry and added to specific metrics usage

ChrsMark (approved):

jsuereth (approved):

PR #991, opened 2024-05-01
Add node.js runtime metrics semantic conventions

### LINKED ISSUE #990: Add Node.js runtime metrics specific semantic conventions
### Area(s)

area:nodejs

Currently, only JVM have detailed semantic conventions for its metrics. Other languages would also required, such as Node.js

Addition of Node.js runtime metric conventions

maryliag:
Part Of #990

## Changes

Adding semantic conventions for Node.js runtime metrics


maryliag:
I wasn't sure if I need to update something on the next, since there was no changes to existing semantic. This is a new one.

trask:

docs/runtime/nodejs-metrics.md
```diff
@@ -0,0 +1,149 @@
...
+| `nodejs.active_handles.count` | UpDownCounter | `{handles}` | Number of active handles. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- endsemconv -->
+
+<!-- semconv metric.nodejs.active_handles.count(full) -->
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`nodejs.version`](/docs/attributes-registry/nodejs.md) | string | Node.js version. | `v22.0.0`; `v21.7.3` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- endsemconv -->
+
+### Metric: `nodejs.active_requests.count`
```
Flarna, line 46: What exactly is an `active_request` here?

model/registry/nodejs.yaml
```diff
@@ -0,0 +1,94 @@
...
+  - id: registry.nodejs
+    type: attribute_group
+    brief: Describes Node.js related attributes.
+    prefix: nodejs
+    attributes:
+      - id: version
+        stability: experimental
+        type: string
+        brief: Node.js version.
+        examples: ["v22.0.0", "v21.7.3"]
```
Flarna, line 11: Why the leading `v`?

model/registry/nodejs.yaml
```diff
@@ -0,0 +1,94 @@
...
+              stability: experimental
+            - id: p90
+              value: 'p90'
+              brief: 'Event loop 90 percentile latency.'
+              stability: experimental
+            - id: p99
+              value: 'p99'
+              brief: 'Event loop 99 percentile latency.'
+              stability: experimental
+        brief: The type of memory.
```
Flarna, line 41: seems a copy paste issue

docs/runtime/nodejs-metrics.md
maryliag: Number of active libuv requests

model/registry/nodejs.yaml
maryliag: ops, fixed!

model/metrics/nodejs-metrics.yaml
```diff
@@ -0,0 +1,66 @@
...
+    stability: experimental
+    attributes:
+      - ref: nodejs.version
+        requirement_level: recommended
+
+  - id: metric.nodejs.eventloop.lag
+    type: metric
+    metric_name: nodejs.eventloop.lag
+    brief: "Event loop lag."
+    instrument: counter
```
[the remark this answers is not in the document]
maryliag: Good point, indeed gauge would be a better fit. Updated

Netail:
The only thing I am wondering is, whether or not it should be called NodeJS. What if someone uses Deno or Bun?

maryliag:
> I am wondering whether or not it should be called NodeJS. What if someone uses Deno or Bun?

what about just `js` ?

Netail:
> > I am wondering whether or not it should be called NodeJS. What if someone uses Deno or Bun?
> 
> what about just `js` ?

Well that would hint towards a language, while they all are javascript runtime environments

maryliag:
> I mean I guess, but also feel like it can be confused with the language instead of the runtime environment

if we want to go be definition, it would be `runtimejs`

Netail:
> > I mean I guess, but also feel like it can be confused with the language instead of the runtime environment
> 
> if we want to go be definition, it would be `runtimejs`

jsruntime? idk, might need some extra opinions

.chloggen/jsruntime-metrics.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: jsruntime
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Introducing semantic conventions for Runtime Javascript (e.g. NodeJS, DenoJS, BunJS) runtime metrics.
```
Netail, line 13: I don't believe Deno and Bun use the JS suffix

maryliag: removed

pikalovArtemN:
For my opinion different runtimes need to have diferent preffix.
For example:
node - nodejs
bun - bun
deno - deno
etc...
Or some attribute that specify what runtime is currently run.
Or metrics idk.

Also i think for version we need to set metric like it does in the prom-client library. Attribute in all metrics with version, for my opinion, it's not a good solution.

docs/runtime/jsruntime-metrics.md
```diff
@@ -0,0 +1,192 @@
...
+<!-- endsemconv -->
+
+### Metric: `jsruntime.gc.duration`
+
+This metric is [recommended][MetricRecommended].
+
+<!-- semconv metric.jsruntime.gc.duration(metric_table) -->
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `jsruntime.gc.duration` | Gauge | `s` | Garbage collection duration. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
pikalovArtemN, line 100: i think it's not a gauge it's a histogram

docs/runtime/jsruntime-metrics.md
```diff
@@ -0,0 +1,192 @@
...
+<!-- semconv metric.jsruntime.heap.space(metric_table) -->
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `jsruntime.heap.space` | UpDownCounter | `By` | Heap space. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- endsemconv -->
+
+<!-- semconv metric.jsruntime.heap.space(full) -->
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`jsruntime.heap.space.state`](/docs/attributes-registry/jsruntime.md) | string | The space of heap memory. | `total`; `used` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
pikalovArtemN, line 180: i think need to add 'jsruntime.heap.space' 
https://nodejs.org/api/v8.html#v8getheapstatistics 
In documentation it's space_name field

docs/attributes-registry/jsruntime.md
```diff
@@ -0,0 +1,62 @@
...
+| ------- | ----------------------- | ---------------------------------------------------------------- |
+| `total` | Total heap memory size. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `used`  | Used heap memory size.  | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+`jsruntime.heap.space.state` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value   | Description              | Stability                                                        |
+| ------- | ------------------------ | ---------------------------------------------------------------- |
+| `total` | Total heap memory space. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `used`  | Used heap memory space.  | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
pikalovArtemN, line 55: i think need to add available

pikalovArtemN (changes requested):

docs/attributes-registry/jsruntime.md
maryliag: can't you just get that from `total - used`?

Netail: I believe they use used + free for memory, no total

pikalovArtemN: > can't you just get that from `total - used`?

I don't think so, in [doc](https://nodejs.org/api/v8.html#v8getheapstatistics) if count total - used (space_size - space_used_size) example it's different than space_available_size

dyladan:
> > > I mean I guess, but also feel like it can be confused with the language instead of the runtime environment
> > 
> > 
> > if we want to go be definition, it would be `runtimejs`
> 
> jsruntime sounds maybe better? idk, might need some extra opinions

I think we should make it clear that it is only for `NodeJS` not other runtimes like `deno`, `bun`, or something else. JS is just a language but the underlying runtimes differ a lot. I would argue for something along the lines of `nodejs`, `noderuntime`, or `nodejsruntime`.

docs/runtime/jsruntime-metrics.md
```diff
@@ -0,0 +1,192 @@
...
+<!-- endsemconv -->
+
+### Metric: `jsruntime.eventloop.utilization`
+
+This metric is [recommended][MetricRecommended].
+
+<!-- semconv metric.jsruntime.eventloop.utilization(metric_table) -->
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `jsruntime.eventloop.utilization` | Gauge | `s` | Event loop utilization. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
dyladan, line 84: Utilization is usually reported as a float between 0 and 1 as it represents a percentage of the time the event loop is utilized. `s` doesn't make sense in this context.

https://nodejs.org/api/perf_hooks.html#performanceeventlooputilizationutilization1-utilization2 returns 3 numbers:

- `idle` - number of ms event loop is idle
- `active` - number of ms event loop is active
- `utilization` - float between 0 and 1 calculated from idle and active

I would suggest either reporting just the utilization percent, or just the idle and active components as a gauge split by a `active=0/1` dimension. Both could be reported, but it would be duplicate data.

dyladan (commented):
This is a great start thanks for putting this together

dyladan:
> > > > I mean I guess, but also feel like it can be confused with the language instead of the runtime environment
> > > 
> > > 
> > > if we want to go be definition, it would be `runtimejs`
> > 
> > 
> > jsruntime sounds maybe better? idk, might need some extra opinions
> 
> I think we should make it clear that it is only for `NodeJS` not other runtimes like `deno`, `bun`, or something else. JS is just a language but the underlying runtimes differ a lot. I would argue for something along the lines of `nodejs`, `noderuntime`, or `nodejsruntime`.

It may be worth splitting the `v8` semconv from the NodeJS semconv. The `v8` engine is used by a few different JS runtimes and it would be worth making them consistent between runtimes. Most of the metrics returned by the [`v8`](https://nodejs.org/api/v8.html) module are returned directly from the `v8` engine such as [`getHeapStatistics`](https://v8docs.nodesource.com/node-13.2/d5/dda/classv8_1_1_isolate.html#a5593ac74687b713095c38987e5950b34). The node metrics returned by the [`perf_hooks`](https://nodejs.org/api/perf_hooks.html) module however, are specific to NodeJS.

docs/runtime/jsruntime-metrics.md
dyladan: I agree. Each collection is an entry in a histogram so you can view a summary of collection times over time

maryliag:
thank you for all the comments, I'm looking and will be replying to them 👀

docs/runtime/jsruntime-metrics.md
maryliag: I had this as a histogram, and because on the SIG meeting on May 1st, I understood that I cannot use histograms here, so I changed to gauge, maybe I misunderstood.

trask: check out [jvm.gc.duration](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/jvm-metrics.md#metric-jvmgcduration)

also, for any histograms, you'll want to specify the default buckets, check out https://github.com/open-telemetry/semantic-conventions/issues/316#issuecomment-1783528869

maryliag: Changed back to histogram and added info about bucket

docs/runtime/jsruntime-metrics.md
```diff
@@ -0,0 +1,192 @@
...
+
+| Value  | Description | Stability |
+|---|---|---|
+| `total` | Total memory. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `used` | Used memory. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- endsemconv -->
+
+### Metric: `jsruntime.heap.size`
+
+This metric is [recommended][MetricRecommended].
```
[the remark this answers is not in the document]
maryliag: You think this (and the following one) should be marked as opt-in or you're saying to completely remove it form here?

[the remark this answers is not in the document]
maryliag: I didn't see it as required, so this is why I used the recommended. Since the opt-in option is to be used for expensive metric or that have a security risk, I didn't think it would apply here.

docs/runtime/jsruntime-metrics.md
maryliag: Fixed with a note for the range [0.0-1.0]

docs/runtime/jsruntime-metrics.md
```diff
@@ -0,0 +1,192 @@
...
+| `jsruntime.active_handles.count` | UpDownCounter | `{handles}` | Number of active handles. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- endsemconv -->
+
+<!-- semconv metric.jsruntime.active_handles.count(full) -->
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`jsruntime.version`](/docs/attributes-registry/jsruntime.md) | string | JS Runtime version. | `v22.0.0`; `v21.7.3` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- endsemconv -->
+
+### Metric: `jsruntime.eventloop.lag`
```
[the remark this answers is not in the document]
maryliag: yes, it was coming from there. Renamed to use `delay` instead

docs/runtime/jsruntime-metrics.md
maryliag: New metrics created

docs/attributes-registry/jsruntime.md
maryliag: Changed to separated metrics of the different types

docs/attributes-registry/jsruntime.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Jsruntime
+
+## Jsruntime Attributes
+
+Describes JS Runtime related attributes.
+
+| Attribute                      | Type   | Description                         | Examples                        | Stability                                                        |
+| ------------------------------ | ------ | ----------------------------------- | ------------------------------- | ---------------------------------------------------------------- |
+| `jsruntime.eventloop.lag.type` | string | The type of the event loop latency. | `min`; `max`; `mean`            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
maryliag: Replaces this with a metric for each oh those types

docs/attributes-registry/jsruntime.md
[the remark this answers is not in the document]
maryliag: Replaced with separated metrics

docs/runtime/jsruntime-metrics.md
```diff
@@ -0,0 +1,192 @@
...
+
+This metric is [recommended][MetricRecommended].
+
+<!-- semconv metric.jsruntime.eventloop.lag(metric_table) -->
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `jsruntime.eventloop.lag` | Gauge | `s` | Event loop lag. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
maryliag: Separated on different metrics

maryliag:
New updates:
- Metrics separated in 2: node.js (containing only the node.js specific metrics), v8 js engine metrics (that can be used by other js runtime, such as deno)
- replaced metrics where I was using type attributes to define the type, now they're separate metrics (for example, heap size and eventloop delay)

docs/runtime/v8jsengineruntime-metrics.md
```diff
@@ -0,0 +1,212 @@
...
+<!-- endsemconv -->
+
+### Metric: `v8jsengineruntime.memory.size`
+
+This metric is [recommended][MetricRecommended].
+
+<!-- semconv metric.veightjsengineruntime.memory.size(metric_table) -->
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `v8jsengineruntime.memory.size` | UpDownCounter | `By` | External memory size. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
pikalovArtemN, line 68: I think it's Gauge. like in prom-client library

pikalovArtemN (changes requested):

maryliag:
From the discussion on the SIG meeting, keeping this PR just for node and created a new one for the v8 on https://github.com/open-telemetry/semantic-conventions/pull/1066

docs/attributes-registry/nodejs.md
```diff
@@ -0,0 +1,15 @@
+<!--- Hugo front matter used to generate the website version of this page:
+--->
+
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/attribute_namespace.md.j2 -->
+
+# Nodejs
```
trisch-me, line 7: I think it makes sense to add this to the abbreviations list such as NodeJS to the weaver.yaml

maryliag: Ah cool! I wasn't aware of it, and it looks a lot better now!

docs/attributes-registry/nodejs.md
```diff
@@ -0,0 +1,15 @@
...
+--->
+
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/attribute_namespace.md.j2 -->
+
+# NodeJS
+
+## NodeJS Attributes
+
+Describes Node.js Runtime related attributes.
```
trisch-me, line 11: are those always only runtime attributes? I wonder if registry as agnostic entity should have `runtime` word
```suggestion
Describes Node.js related attributes.
```

trisch-me (approved):

docs/attributes-registry/nodejs.md
maryliag: good point, removed!

model/metrics/nodejs-metrics.yaml
```diff
@@ -0,0 +1,98 @@
...
+    brief: "Event loop standard deviation delay."
+    instrument: gauge
+    unit: "s"
+    stability: experimental
+    prefix: nodejs.eventloop.delay.stddev
+    attributes:
+      - ref: nodejs.version
+        requirement_level: recommended
+
+  - id: metric.nodejs.eventloop.delay.pfifty
```
david-luna, line 50: nit: any reason for the leafing `p` in `pfifty`? if so, why not then use it in the following percentiles (`pninety`, `pninety_nine`)?

maryliag: I added these as p50, but there is a limitation on the readme generation that couldn't read numbers, so I just switched to `fifty`, and then kept the `p`, since that is the name of the metric. I must have deleted on the others by accident, so I added back to keep consistent.
 
This value doesn't really matter, is just used on the readme generation, but doesn't even show on the readme itself when reading it.

model/metrics/nodejs-metrics.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - id: metric.nodejs.eventloop.delay.min
+    type: metric
+    metric_name: nodejs.eventloop.delay.min
+    brief: "Event loop minimum delay."
+    instrument: gauge
+    unit: "s"
+    stability: experimental
+    prefix: nodejs.eventloop.delay.min
+    attributes:
+      - ref: nodejs.version
```
legendecas, line 11: I think `nodejs.version` would be better be a resource attribute. Like https://github.com/open-telemetry/semantic-conventions/blob/main/model/resource/webengine.yaml#L9

maryliag: I had this as a resource attribute, and then a prior comment on this PR suggested as its own metric: 
> Also i think for version we need to set metric like it does in the prom-client library. Attribute in all metrics with version, for my opinion, it's not a good solution.

I'm open to suggestions on which option is best, I have nos strong preferences either way

legendecas: Are you referring to https://github.com/open-telemetry/semantic-conventions/pull/991#issuecomment-2093541602? I think @pikalovArtemN is suggesting that making a constant metric for resource attributes rathar than including these attributes in every metrics, maybe they can elaborate on their idea.

maryliag: I just realized that I read your initial comment wrong, my mistake. I agree it makes more sense. I made the changes to be a resource attribute now, let me know what you think 😄

legendecas: I'd think `process.runtime.version` is sufficient to represent nodejs version.

maryliag: good point! Removed

legendecas (approved):

david-luna:
One last question from my side. I've checked that other metrics of time make use of `histogram` as instrument. Some examples:

- https://github.com/open-telemetry/semantic-conventions/blob/cde003cd371ef8cc802f275bcfed15cff8fd6a48/model/metrics/jvm-metrics.yaml#L49
- https://github.com/open-telemetry/semantic-conventions/blob/cde003cd371ef8cc802f275bcfed15cff8fd6a48/model/metrics/go-metrics.yaml#L77

Here all the metrics related to delay and using `s` as unit use the `gauge` instrument. Probably a naive question but why is that?

maryliag:
> Here all the metrics related to delay and using s as unit use the gauge instrument. Probably a naive question but why is that?

That is a limitation of this type of metric for node, meaning, I'm using this more because is from the event loop and not because is a duration (I even use histogram on duration on other metrics). 
The gist is that the values for min, max, p50, p90, p99 for these metrics come from the builtin histogram from the node runtime and is not possible to convert it to otel histogram, because we can only query the distribution not get the entire histogram object. This is why you will notice I have a separate metric for each value (min, max, etc), instead of just one called delay that would have everything.

Hope that helps clarify!

docs/runtime/nodejs-metrics.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Experimental
+
+**Status**: [Experimental][DocumentStatus]
+
+**Description:** Experimental Node.js Runtime metrics captured under `nodejs`.
+
+### Metric: `nodejs.eventloop.delay.min`
+
+This metric is [recommended][MetricRecommended].
```
pyohannes, line 35: Could you give a brief description of how this metric is recorded, like it's done for Java:

> This metric is obtained from [`MemoryPoolMXBean#getUsage()`](https://docs.oracle.com/javase/8/docs/api/java/lang/management/MemoryPoolMXBean.html#getUsage--).

Or Go:

> [1]: Computed from `/sched/goroutines:goroutines`.

david-luna:
> > Here all the metrics related to delay and using s as unit use the gauge instrument. Probably a naive question but why is that?
> 
> That is a limitation of this type of metric for node, meaning, I'm using this more because is from the event loop and not because is a duration (I even use histogram on duration on other metrics). The gist is that the values for min, max, p50, p90, p99 for these metrics come from the builtin histogram from the node runtime and is not possible to convert it to otel histogram, because we can only query the distribution not get the entire histogram object. This is why you will notice I have a separate metric for each value (min, max, etc), instead of just one called delay that would have everything.
> 
> Hope that helps clarify!

It helps a lot. Thank you! :)

david-luna (approved):

docs/runtime/nodejs-metrics.md
```diff
@@ -0,0 +1,310 @@
...
+<!-- toc -->
+
+- [Experimental](#experimental)
+  - [Metric: `nodejs.eventloop.delay.min`](#metric-nodejseventloopdelaymin)
+  - [Metric: `nodejs.eventloop.delay.max`](#metric-nodejseventloopdelaymax)
+  - [Metric: `nodejs.eventloop.delay.mean`](#metric-nodejseventloopdelaymean)
+  - [Metric: `nodejs.eventloop.delay.stddev`](#metric-nodejseventloopdelaystddev)
+  - [Metric: `nodejs.eventloop.delay.p50`](#metric-nodejseventloopdelayp50)
+  - [Metric: `nodejs.eventloop.delay.p90`](#metric-nodejseventloopdelayp90)
+  - [Metric: `nodejs.eventloop.delay.p99`](#metric-nodejseventloopdelayp99)
```
pyohannes, line 22: This has been answered in previous comments, but I'd like to reiterate: it would be really nice to unify all this into one single histogram.

This is not blocking for merging those metrics as experimental, but I think should be thoroughly evaluated before declaring those metrics as stable. I wonder if there's any possibility for the runtime to expose the raw event loop delay and to feed this into a native OTel histogram.

maryliag: Agree that this is the ideal scenario, so until then I will keep those separate and if we find a out to convert, I will update the semantic here.

pyohannes (approved):

lmolkova (approved):

lmolkova:
> That is a limitation of this type of metric for node, meaning, I'm using this more because is from the event loop and not because is a duration (I even use histogram on duration on other metrics). The gist is that the values for min, max, p50, p90, p99 for these metrics come from the builtin histogram from the node runtime and is not possible to convert it to otel histogram, because we can only query the distribution not get the entire histogram object. This is why you will notice I have a separate metric for each value (min, max, etc), instead of just one called delay that would have everything.

Could you please add a similar note to the md file itself?

@open-telemetry/javascript-approvers any last minute feedback?

maryliag:
> Could you please add a similar note to the md file itself?

Done

PR #992, opened 2024-05-01
Finalize migration requirement for stable database semconv

### LINKED ISSUE #719: Add warning on database semconv about not adopting any new breaking changes until database semconv stability is reached
(similar to the HTTP warning that is in place)

trask:
Fixes #719

## Changes

Finalizes the migration requirement for instrumentations to follow when updating to stable database semconv.


pyohannes:
> **Warning**
> Existing Database instrumentations that are using
> [v1.20.0 of this document](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/database.md)
> (or prior):
>
> * SHOULD NOT change the version of the networking conventions that they emit
>   until the HTTP semantic conventions are marked stable (HTTP stabilization will
>   include stabilization of a core set of networking conventions which are also used
>   in Database instrumentations).

Does it makes sense that this part still coexists with the warning proposed in this PR?

If I understand correctly, this would require users to set both `OTEL_SEMCONV_STABILITY_OPT_IN=database,http` to reliably get data according to new (stable) conventions?

docs/database/database-metrics.md
```diff
@@ -32,10 +32,28 @@ and attributes but more may be added in the future.
 
 > **Warning**
 > Existing database instrumentations that are using
-> [v1.24.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/database/database-metrics.md)
-> (or prior) SHOULD NOT change the version of the database conventions that they emit
-> until a transition plan to the (future) stable semantic conventions has been published.
-> Conventions include, but are not limited to, attributes, metric and span names, and unit of measure.
+> [v1.24.0 of this document](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md)
```
gregkalapos, line 35: The link seems to be a copy-paste leftover. Same in `database-spans.md`. 

```suggestion
> [v1.24.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/database/database-metrics.md)
```

gregkalapos (commented):
nit - somewhat related to @pyohannes's comment. In the [`http`](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/http) docs in the part about `OTEL_SEMCONV_STABILITY_OPT_IN` it currently says:
> The only values defined so far are:

If these new values are added due to DB, I think that sentence should be removed from the http docs.

trask:
> Does it makes sense that this part still coexists with the warning proposed in this PR?

good point, it does make it much more confusing

I've removed the http part. That part still exists in prior tags, so still can apply if someone is upgrading to earlier version of semconv.

lmolkova (approved):

gregkalapos (approved):

pyohannes (approved):

Issue #648, opened 2022-09-20
Add system uptime metric


andrzej-stencel:
**What are you trying to achieve?**

I want to add a metric to the semantic conventions that will describe the system uptime. How about `system.uptime`?

**Additional context.**

This is reported by Telegraf as `uptime` field of the `system` metric (in seconds).

Here's a related proposal on the `hostmetrics` receiver to add this metric: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14130.

minuk-dev:
Q. Is there any plan to do it?

mx-psi:
@kernelpanic77 offered to do it here https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/31627#issuecomment-2133882411 :)

dmitryax:
Looks like we have an agreement here. Just need someone to submit a PR

tigrannajaryan:
Is there any way we can generalize this to be an "uptime" of any entity, not just of "system"? What if we make this an `uptime` metric with the Resource describing what it is about (e.g. Resource can have "host.name=foo" to indicate that it is an uptime of a host).

andrzej-stencel:
> Is there any way we can generalize this to be an "uptime" of any entity, not just of "system"? What if we make this an `uptime` metric with the Resource describing what it is about (e.g. Resource can have "host.name=foo" to indicate that it is an uptime of a host).

I suppose we could do it, I wonder what others think.

The `uptime` attribute name would not be namespaced, unlike `system.uptime` that is namespaced to `system`. Looking at the [Attributes Registry](https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/attributes-registry/README.md), it doesn't look like we currently have any non-namespaced attributes in the semantic conventions. Is that true?

rogercoll:
> Is there any way we can generalize this to be an "uptime" of any entity, not just of "system"?

I think this is part of a broader discussion taking place in https://github.com/open-telemetry/semantic-conventions/issues/1161 (`system.uptime` vs `process.uptime` vs `container.uptime`). The main benefit of using the metric without namespace seems to be dashboards correlation and avoiding deduplication. But it comes at the cost of implying resource attributes to corresponding metrics (not sure if this is possible in semconv), for example, the `uptime` metric should always be linked to either `host.name`, `process.pid` or `container.id`. 

[...]

Issue #626, opened 2024-01-03
Adjust `process.command_args` and `process.command_line` to be opt-in


inssein:
By default, auto instrumentation ships `process.command_args`, which is very dangerous as a lot of java services pass in secrets via command line arguments. Example:

```
java \
  -Dkeycloak.clientSecret="${KEYCLOAK_SECRET:-test}" \
  -jar app.jar
```

I have opened an issue against the java auto instrumentation repo (https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/10151), but I was pointed to https://opentelemetry.io/docs/specs/semconv/resource/process/ which indicates that the information I am asking to be made opt-in is marked "Conditionally Required" in the specification.

Does it make sense to give an out for languages where passing in secrets via command line arguments is common? Curious also if other languages have this problem and how they deal with it.

joaopgrassi:
@open-telemetry/semconv-system-approvers FYI

mx-psi:
I am in favor of making these opt-in

joaopgrassi:
@inssein would you like to send a PR to address it?

inssein:
@joaopgrassi happy to, but I just have a quick clarifying question.

For the `container` resource, `comand_line` and `command_args` are both `opt_in`. Are we good with doing that to the `process` resource (here), or just `command_args`?

trask:
cc @breedx-splk since you had thoughts in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/10151#issuecomment-1967818321

breedx-splk:
> cc @breedx-splk since you had thoughts in [open-telemetry/opentelemetry-java-instrumentation#10151 (comment)](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/10151#issuecomment-1967818321)

Thanks @trask. Yeah, I don't love the motivation here...tho I'm probably outnumbered. 

We already jump through so many hoops to mask and hide information that users shouldn't be exposing in the first place, and I think that can come at the expense of troubleshooting utility. Like especially in the java world, where configuration items are often passed on the commandline via system properties. Not being able to see an agent's configuration (via commandline) after the fact is frustrating....and I use the commandline to help troubleshoot quite often.

In the other thread, there is also mention of size/length/waste, which I think is maybe a stronger reason for considering making these opt-in (off by default). I also pointed out that I think this is probably fine, since users and vendors can always choose to override it.

mx-psi:
[...]

PR #993, opened 2024-05-01
Add additional process fields from ECS

mjwolf:
## Changes

Add additional process fields to the attribute registry, from ECS.

The newly added fields are:
* [process.args_count](https://www.elastic.co/guide/en/ecs/current/ecs-process.html#field-process-args-count)
* [process.title](https://www.elastic.co/guide/en/ecs/current/ecs-process.html#field-process-title)
* [process.uptime](https://www.elastic.co/guide/en/ecs/current/ecs-process.html#field-process-uptime)
* [process.working_directory](https://www.elastic.co/guide/en/ecs/current/ecs-process.html#field-process-working-directory)

* ~~[ ] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.~~


model/registry/process.yaml
```diff
@@ -147,6 +157,15 @@ groups:
           An additional description about the runtime of the process, for example
           a specific vendor customization of the runtime environment.
         examples: 'Eclipse OpenJ9 Eclipse OpenJ9 VM openj9-0.21.0'
+      - id: title
+        type: string
+        stability: experimental
+        brief: >
+          Process title
+        note: >
+          The proctitle, some times the same as process name. Can also be different, for
```
trisch-me, line 166: should it be `process title`?

mjwolf: proctitle is the common term for it, for example see https://manpages.ubuntu.com/manpages/trusty/man3/setproctitle.3.html

I've reworked the note to have both terms though

model/registry/process.yaml
```diff
@@ -165,8 +184,20 @@ groups:
         brief: >
           The exit code of the process.
         examples: [127]
+      - id: uptime
```
joaopgrassi, line 187: Isn't this better modeled as a metric?

model/registry/process.yaml
```diff
@@ -165,8 +184,20 @@ groups:
...
+        stability: experimental
+        brief: >
+          Seconds the process has been up.
+        examples: [2880]
       - id: interactive
         type: boolean
         stability: experimental
         brief: >
           Whether the process is connected to an interactive shell.
+      - id: working_directory
```
joaopgrassi, line 198: hum couldn't this be considered sensitive information? 🤔

mjwolf: What sort of sensitive information are you thinking of?

It's possible that it contains something that someone considers sensitive, but it's also possible that a few of the other existing fields also have similar sensitive information, like `executable.path`, `user.id`, `command_args`, etc.

Do you think this needs any special call-out on potentially having sensitive information?

mjwolf: I see from [here](https://github.com/open-telemetry/semantic-conventions/pull/731#issuecomment-2072875648) there's already sensitive tags, which I didn't know about. Maybe it should be a larger conversation on how to apply them to the process fields, since there are a lot of other fields that could potentially have sensitive data

trisch-me: there is already a larger [discussion](https://github.com/open-telemetry/oteps/pull/255/files) on sensitive data
After resolution for this otep we would need to mark probably all sensitive fields in our registry. But for now I don't think we need to change this PR and remove this field. Maybe add a note to it. But I agree with you, this field is as sensitive as many other process fields, for example command args could have some credentials passed over etc

model/registry/process.yaml
mjwolf: I've changed this to a metric

trisch-me (approved):

docs/attributes-registry/process.md
```diff
@@ -15,6 +15,7 @@ An operating system process.
...
 | ----------------------------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------- | ---------------------------------------------------------------- |
+| `process.args_count`          | int      | Length of the process.command_args array [1]                                                                                                                                                                                                                                                                                                                | `4`                                                 | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
joaopgrassi, line 18: This is also more like a metric, no? 🤔

mjwolf: This could be a metric, but to me, it doesn't feel like it should. I don't think it represents any measurement that anyone would care to track over time with a metric. The arg_count generally won't change (outside of an exec), and it's not a measurement that indicates system performance or utilization.

It does have value as an attribute in a log/event though. This is intended to be a convenience/performance field. You could also get arg count by counting the `process.args` length, but having a separate `arg_count` field makes it easier to write queries on this field directly.

joaopgrassi: >The arg_count generally won't change (outside of an exec), and it's not a measurement that indicates system performance or utilization.

Hum, but this then contradicts the brief of the field, no?

> This field can be useful for querying or performing bucket analysis on how many
> arguments were provided to start a process. More arguments may be an indication
> of suspicious activity.

If it's important to recognize suspicious activity like that, I feel having it in a log or span, rather than being a metric is more cumbersome? Because one would have to extract the attribute from logs/spans and then plot it somehow as a metric?

mjwolf: I've talked with the team at Elastic that uses this field to work on detections, and they've told me it has value as both a metric and an attribute attached to an event.

I've now added it as both a metric and a field in the attribute registry in this PR. Although I didn't see any prior art for a field being used as both, so I'm not sure if I've done it in the correct way

joaopgrassi: Hum I'm also not sure, having both the attribute and the metric seems a bit.. unnecessary but I don't have strong opinions against it. I left another comment because you used a Histogram for it and I'm not sure why.

model/metrics/process-metrics.yaml
```diff
@@ -100,3 +101,19 @@ groups:
...
+    brief: "The number of seconds the process has been running."
+    instrument: counter
+    unit: "s"
+
+  - id: metric.process.args_count
+    type: metric
+    metric_name: process.args_count
+    stability: experimental
+    brief: "Number of process arguments."
+    instrument: histogram
```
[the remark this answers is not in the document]
mjwolf: I chose histogram because this is a non-additive property, and according to [this](https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/#additive-property), the only metric instruments to use for non-additive properties are Histogram and Asynchronous gauge. Of these two, I think the histogram is better, since there's value in working with the statistics of args_count. For example, an outlier in the number of arguments for a process name could be an indication of attack.

If there's another instrument type you think is better, I will change it. I don't know the metric instrument types well, and if there's something more appropriate I can change it.

model/metrics/process-metrics.yaml
```diff
@@ -100,3 +101,19 @@ groups:
...
+    type: metric
+    metric_name: process.uptime
+    stability: experimental
+    brief: "The number of seconds the process has been running."
+    instrument: counter
+    unit: "s"
+
+  - id: metric.process.args_count
+    type: metric
+    metric_name: process.args_count
```
lmolkova, line 115: what would be the use case for this metric? Does it make any sense, especially without any attributes?

I.e. I can see P95 of my processes have 2 arguments and P99 is 10 arguments - is this helpful at all?

model/metrics/process-metrics.yaml
```diff
@@ -100,3 +101,19 @@ groups:
...
+    stability: experimental
+    brief: "Number of process arguments."
+    instrument: histogram
+    unit: "{count}"
```
lmolkova, line 119:
```suggestion
    unit: "{argument}"
```

model/metrics/process-metrics.yaml
mjwolf: I've taken args_count out for now.  This is useful for threat hunting, but it's really not useful for system performance or other use cases.

In the future we can consider adding it back in a more use-case specific metrics namespace.

joaopgrassi:
Looping in @open-telemetry/semconv-system-approvers as this also adds a process metric.

model/registry/process.yaml
```diff
@@ -147,6 +157,15 @@ groups:
           An additional description about the runtime of the process, for example
           a specific vendor customization of the runtime environment.
         examples: 'Eclipse OpenJ9 Eclipse OpenJ9 VM openj9-0.21.0'
+      - id: title
+        type: string
+        stability: experimental
+        brief: >
+          Process title
+        note: >
+          The process title (proctitle), sometimes the same as process name. It can also be different, for
```
ChrsMark, line 166: Could we add more examples here or enhance the description with more specific details? For someone who is not really familiar with this attribute might be hard to interpret.

lmolkova (approved):
Left a few last cosmetic comments, looks great otherwise

model/metrics/process-metrics.yaml
```diff
@@ -100,3 +101,11 @@ groups:
     unit: "{fault}"
     attributes:
       - ref: process.paging.fault_type
+
+  - id: metric.process.uptime
```
lmolkova, line 105: we should have the metric definition in markdown

docs/attributes-registry/process.md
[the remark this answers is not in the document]
mjwolf: `arg_count` can be an important part of security queries. Without this field, backends would need to count each `command_args` field, which can be computationally expensive.

So it is a bit of duplication, but having both for query optimizations is worthwhile.

model/metrics/process-metrics.yaml
```diff
@@ -100,3 +101,11 @@ groups:
...
+    type: metric
+    metric_name: process.uptime
+    stability: experimental
+    brief: "The number of seconds the process has been running."
```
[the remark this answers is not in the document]
mjwolf: I agree "time" is better here.

Precision isn't currently part of the metrics spec, so we can't really specify it now. It also might not be good to specify. Different systems might capture this with different precision, so I don't think we'd want to force a value here.

model/registry/process.yaml
mjwolf: I've rewritten the description and added a few more examples, which hopefully makes it more clear

model/registry/process.yaml
```diff
@@ -84,6 +84,16 @@ groups:
...
+      - id: args_count
+        type: int
+        stability: experimental
+        brief: >
+          Length of the process.command_args array
+        note: >
+          This field can be useful for querying or performing bucket analysis on how many
+          arguments were provided to start a process. More arguments may be an indication
+          of suspicious activity.
+        examples: [4]
```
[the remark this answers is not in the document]
braydonk: Perhaps you mean only if `process.command_args` *is* populated? Making it only recommended when the args metric is not populated doesn't make much sense to me

model/metrics/process-metrics.yaml
```diff
@@ -102,3 +103,11 @@ groups:
     unit: "{fault}"
     attributes:
       - ref: process.paging.fault_type
+
+  - id: metric.process.uptime
```
braydonk, line 107: We have a separate issue open for this metric: #648 

Would be fine by me if this ended up being covered by this PR. I don't have a good grasp on how semconv generally wants these types of metrics to be designed; I see arguments for these kinds of `uptime` metrics being not useful since the metrics should be reported with a start time anyway. (I might not be fully understanding this here, so hopefully I'm not misrepresenting). Given that, I'd like to make sure I'm understanding the intended usage of the metric. I see it's a `counter`; does that mean each reported point is essentially reporting `currentTime` - `processStartTime`?

docs/attributes-registry/process.md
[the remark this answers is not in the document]
joaopgrassi: I think what @lmolkova is that if `command_args` is not populated, there's little sense in populating `arg_count` as both goes hand in hand.

model/registry/process.yaml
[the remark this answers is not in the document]
joaopgrassi: Yes I'm also confused. To me both goes hand in hand. I'd consider _not_ setting `arg_count` when `command_args` is not set.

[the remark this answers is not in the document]
mjwolf: Yes, I think both should be set together. I've added a recommended level, but have set the test to:
`if `process.command_args` is populated.`

model/metrics/process-metrics.yaml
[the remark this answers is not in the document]
lmolkova: I understand that it could be hard to measure things with actual millisecond precision, but getting current time from the platform in milliseconds is possible on all system. So why not recommend millisecond precision here? What's the downside?

model/metrics/process-metrics.yaml
mjwolf: I can understand the argument that the uptime metric might not be too useful. OSs don't track process uptime specifically, so it would be calculated by `currentTime - processStartTime`. However this has been requested a few times in various issues, and there are existing users of process.uptime in ECS, so I think that shows there's value in having this.

I'm not sure if this should be a counter or gauge. I picked counter because it's monotonic, but it is also non-additive, so it could also be a gauge.  It was previously discussed [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2824#discussion_r978765127), but I don't see a resolution to that discussion. An individual process's uptime is monotonic, but in aggregation, a group of processes' uptime could decrease (if processes are stopped and replaced by younger processes). I don't know if the metric would be more useful to track individual process uptime, or an aggregation. @reyang, since you were involved in that discussion, do you remember the outcome?

model/registry/process.yaml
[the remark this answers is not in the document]
lmolkova: if `process.command_args` are populated, it's trivial to count them and calculate `process.args_count`. Also `process.args_count` does not have any security, verbosity, sensitivity or other concerns. 
Also, there are discussions about making args opt-in (https://github.com/open-telemetry/semantic-

So I can see why someone would want the count, but not the actual arguments.

When both attributes (args and count) are present, there is a duplication - https://github.com/open-telemetry/semantic-conventions/pull/993#discussion_r1683245851. I'm ok with it on the experimental spec.

so if we are ok with duplication, it makes sense to have 

```yaml
- ref: process.args_count

- ref: process.command_args
  requirement_level: whatever it is now, but will probably become opt_in later
```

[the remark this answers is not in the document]
ChrsMark: Usually having metrics that can be derived from other metrics as `opt-in` makes sense.

model/metrics/process-metrics.yaml
ChrsMark: Without knowing what can be the objections here, a metric that would report a specific's process uptime makes sense to me as a user.

ChrsMark (approved):

model/metrics/process-metrics.yaml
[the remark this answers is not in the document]
mjwolf: I'll add in the suggested note, I agree that will be good

model/registry/process.yaml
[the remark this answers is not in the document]
mjwolf: Existing tools send both `args` and `args_count` from the edge, so telemetry would need to support both. I agree it's trivial to calculate at the storage backend too, but sematic conventions should support the existing use-cases. Sysdig, Falco and Elastic endpoint products all send both args and arg_count.

I'll leave things as they are right now in the attribute registry, I think addressing https://github.com/open-telemetry/semantic-conventions/issues/626 will be done by modifying the resource definition

mjwolf:
I've made a small change to add a note to use at least millisecond precision time to the process.uptime metric. I think the `args` vs. `args_count` discussion should be addressed more in the actual resource definition.

I'll wait a few days, if anyone want to discuss it more, then I plan to merge this PR

Issue #996, opened 2024-05-02
add k8s.container.status.waiting_reason resource attribute


povilasv:
### Is your change request related to a problem? Please describe.

Original request to add such attribute in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32457

Basically user wants to monitor/alert CrashLoopBackOff reason.

Example Pod in such state:
```
  containerStatuses:
  - containerID: containerd://6bb73f56d9f34abc2015629d594625e8e637582ecf94b7e5067171c647a773fd
    image: docker.io/library/nginx:latest
    imageID: docker.io/library/nginx@sha256:ed6d2c43c8fbcd3eaa44c9dab6d94cb346234476230dc1681227aa72d07181ee
    lastState:
      terminated:
        containerID: containerd://6bb73f56d9f34abc2015629d594625e8e637582ecf94b7e5067171c647a773fd
        exitCode: 0
        finishedAt: "2024-05-02T12:30:53Z"
        reason: Completed
        startedAt: "2024-05-02T12:30:41Z"
    name: test
    ready: false
    restartCount: 2
    started: false
    state:
      waiting:
        message: back-off 20s restarting failed container=test pod=dns-example_default(dbd4171d-fcef-45d5-8caf-be07213e1bf6)
        reason: CrashLoopBackOff
```

### Describe the solution you'd like

Add  k8s.container.status.waiting_reason resource attribute.

PR #997, opened 2024-05-02
add k8s.container.status.current_waiting_reason resource attribute

povilasv:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/996

## Changes

### DIFF
--- a/docs/resource/k8s.md
+++ b/docs/resource/k8s.md
@@ -167,6 +167,7 @@ to a running container.
 |---|---|---|---|---|---|
 | [`k8s.container.name`](/docs/attributes-registry/k8s.md) | string | The name of the Container from Pod specification, must be unique within a Pod. Container runtime usually uses different globally unique name (`container.name`). | `redis` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`k8s.container.restart_count`](/docs/attributes-registry/k8s.md) | int | Number of times the container was restarted. This attribute can be used to identify a particular container (running or stopped) within a container spec. |  | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`k8s.container.status.current_waiting_reason`](/docs/attributes-registry/k8s.md) | string | Current waiting reason of the Container. | `CrashLoopBackOff` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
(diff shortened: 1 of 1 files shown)


model/registry/k8s.yaml
```diff
@@ -104,6 +104,13 @@ groups:
         brief: >
           Last terminated reason of the Container.
         examples: ["Evicted", "Error"]
+      - id: container.status.waiting_reason
```
TylerHelmuth, line 107: Similar to `container.status.last_terminated_reason` should the name of this attribute convey a sense of time, like `container.status.latest_waiting_reason` or `container.status.current_waiting_reason`?

ChrsMark (approved):
LGTM

ChrsMark:
@open-telemetry/semconv-k8s-approvers @open-telemetry/specs-semconv-approvers please take a look.

TylerHelmuth (approved):

joaopgrassi (approved):

trask:
sorry for the late comment, was just curious if there was any discussion already about this resource attribute being non-immutable?

ChrsMark:
https://github.com/open-telemetry/semantic-conventions/pull/968 introduced the `container.status.last_terminated_reason`. This one adds a different attribute.
From my perspective these attributes represent "states" of the container. If those cannot be expressed as resource attributes because of the immutability constraint then what would be the right way to report them?

joaopgrassi:
We don't allow removing things anymore and just deprecating, but since this wasn't released yet I think we can just remove it. What do you think @lmolkova ? The other one that was merged `container.status.last_terminated_reason` is part of the last release though, so that one cannot be removed and just deprecated.

ChrsMark:
> I don't think container.status.last_terminated_reason should be removed, as it's AFAIK immutable.

While that's most probably true, I wonder if we should find a consistent way to model both of them. It could be weird to find `k8s.container.status.last_terminated_reason` being a resource attribute while `k8s.container.status.current_waiting_reason` being a metric or event?

lmolkova:
> We don't allow removing things anymore and just deprecating, but since this wasn't released yet I think we can just remove it.

it's ok to remove things that hasn't been released - we know that nothing depends on it.

jsuereth:
I'm way late to the party, but I think we need to back out the resource-attribute portion of this change.

See: https://github.com/open-telemetry/semantic-conventions/issues/1181#issuecomment-2229168814 for rationale.   If I had caught this earlier I would have blocked it earlier, so apologies for being late.  At a minimum, this is still experimental, so we CAN deprecate the resource attribute.

PR #1002, opened 2024-05-03
Adapt Elasticsearch spec to stability changes

gregkalapos:
Follow up from https://github.com/open-telemetry/semantic-conventions/pull/974#discussion_r1584002801. 

## Changes

- Align Elasticsearch span names with the format defined in the general db spec. MongoDB is comparable and it already follows this format. The idea is to not only have the operation (endpoint id) in the span name, but also the target.
- Adding `db.collection.name` to the Elasticsearch spec, which is used in the span name as well.
- Changed the last fallback in the span name from `http.request.method` to `db.system`. The thinking behind this is that `db.system` is usually the last fallback for other DBs. This was also the outcome in https://github.com/open-telemetry/semantic-conventions/issues/704. In practice this fallback is used extremely rarely since endpoint id is usually available.

## Questions
- [x]  `db.elasticsearch.cluster.name` could be replaced with `db.namespace`. Do we want to do that? 
- [ ] I think we could consider making this stable at this point. Any thoughts on this? I'd prefer to try to make this stable.


model/trace/database.yaml
```diff
@@ -198,6 +198,14 @@ groups:
         brief: The request body for a [search-type query](https://www.elastic.co/guide/en/elasticsearch/reference/current/search.html), as a json string.
         examples: [ '"{\"query\":{\"term\":{\"user.id\":\"kimchy\"}}}"' ]
         tag: tech-specific
+      - ref: db.collection.name
+        requirement_level: recommended
+        brief: The index or data stream against which the query is executed.
+        note: >
+          If the query targets multiple indices or data streams, then the name of those should be added as a comma separated list.
+          If the query doesn't target a specific index, this field should remain empty.
```
lmolkova, line 206:
```suggestion
          If the query doesn't target a specific index, this field MUST NOT be set.
```

docs/database/elasticsearch.md
```diff
@@ -14,12 +14,13 @@ described on this page.
 
 ## Span Name
 
-The **span name** SHOULD be of the format `<endpoint id>`.
+The **span name** SHOULD be of the format `{db.operation.name} {db.collection.name}`.
```
lmolkova, line 17: I wonder if this section is necessary, it seems to be perfectly aligned with the general db span name guidance now.

docs/database/elasticsearch.md
```diff
@@ -31,12 +32,13 @@ If the endpoint id is not available, the span name SHOULD be the `http.request.m
...
-| [`db.query.text`](/docs/attributes-registry/db.md) | string | The request body for a [search-type query](https://www.elastic.co/guide/en/elasticsearch/reference/current/search.html), as a json string. | `"{\"query\":{\"term\":{\"user.id\":\"kimchy\"}}}"` | `Recommended` [9] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| [`network.peer.address`](/docs/attributes-registry/network.md) | string | Peer address of the database node where the operation was performed. [10] | `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`db.collection.name`](/docs/attributes-registry/db.md) | string | The index or data stream against which the query is executed. [7] | `my_index`; `index1, index2` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.elasticsearch.cluster.name`](/docs/attributes-registry/db.md) | string | Represents the identifier of an Elasticsearch cluster. | `e9106fc68e3044f0b1475b04bf4ffd5f` | `Recommended` [8] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
trask, line 36: can this be replaced by `db.namespace`?

docs/database/elasticsearch.md
gregkalapos: Yeah, that's a good point, but a bit tricky as well. In general what I propose here is very similar to the general db span name guideline, but it differs a bit, specifically: `db.collection.name` isn't available for all queries - there could be queries that target all indices or no index - in that case we won't have this value and it'd be very unfortunate to fall back to the next item in the `{target}` list from the general spec. I think the `target` list is very useful when all spans have the same set of attributes, but if the list of attributes changes within the same instrumentation (and potentially within the same trace), then we need a solution for that, otherwise the span name format will be mixed. 

I opened https://github.com/open-telemetry/semantic-conventions/issues/1045 to discuss this.

joaopgrassi: I think https://github.com/open-telemetry/semantic-conventions/pull/1069 is going to get merged. Then this is addressed, no?

docs/database/elasticsearch.md
gregkalapos: Yes. I changed it.

The only downside I can think of is that prior to this we had `db.elasticsearch.cluster.name` and `db.elasticsearch.node.name` (<- which we almost deprecated to use `db.instance.id`) - those are related fields. Now it changes, so we only have `db.elasticsearch.node.name`, but the cluster is in `db.namespace`.

Question is what's our preference? 
- To follow general attributes and accept that the cluster is stored in the general `db.namespace` attribute, but the instance is in the Elasticsearch specific `db.elasticsearch.node.name`. I'd say this would prefer consistency across the overall db attributes.
- To prefer consistency within Elasticsearch and have `db.elasticsearch.node.name` and `db.elasticsearch.cluster.name`, but loose some consistency with the overall db attributes on the cluster name.

I'd go with the 1. one (that's the current status of the PR).

https://github.com/open-telemetry/semantic-conventions/pull/972 is related. Without that we'd have a good story on these attributes and both cluster and instance id would be stored in a general, non-elastic searchspecific attribute.

docs/database/elasticsearch.md
gregkalapos: Yes, this is addressed - I removed this part. 

Nevertheless I still have 1 sentence about the span name calling out specifically that we want to use the endpoint id and the index name, instead of the url path. I think that's useful info for instrumentation. But other than that, I just reference the general span name now.

.chloggen/align_es_spec.yaml
```diff
@@ -0,0 +1,24 @@
...
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: >
+  Align Elasticsearch span name to the general database span name guidelines. 
+  Deprecates `b.elasticsearch.cluster.name` in favor of db.namespace.
```
jack-berg, line 15:
```suggestion
  Deprecates `db.elasticsearch.cluster.name` in favor of `db.namespace`.
```

jack-berg (approved):
Just a couple of minor comments.

docs/database/elasticsearch.md
```diff
@@ -14,12 +14,7 @@ described on this page.
 
 ## Span Name
 
-The **span name** SHOULD be of the format `<endpoint id>`.
-
-The elasticsearch endpoint identifier is used instead of the url path in order to reduce the cardinality of the span
-name, as the path could contain dynamic values. The endpoint id is the `name` field in the
-[elasticsearch schema](https://raw.githubusercontent.com/elastic/elasticsearch-specification/main/output/schema/schema.json).
-If the endpoint id is not available, the span name SHOULD be the `http.request.method`.
+The **span name** follows the [general database span name guidelines](database-spans.md#name) with the endpoint identifier stored in `db.operation.name`, instead of the url path in order to reduce the cardinality of the span and the index stored in `db.collection.name`.
```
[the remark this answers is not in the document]
gregkalapos: Good point - I added back the original link.

model/registry/deprecated/db.yaml
```diff
@@ -84,6 +84,13 @@ groups:
         brief: 'Deprecated, no general replacement at this time. For Elasticsearch, use `db.elasticsearch.node.name` instead.'
         deprecated: 'Deprecated, no general replacement at this time. For Elasticsearch, use `db.elasticsearch.node.name` instead.'
         examples: 'mysql-e26b99z.example.com'
+      - id: elasticsearch.cluster.name
+        type: string
+        stability: experimental
+        deprecated: Use `db.namesapce` instead.
```
trask, line 90:
```suggestion
        deprecated: Use `db.namespace` instead.
```

model/trace/database.yaml
```diff
@@ -240,11 +240,21 @@ groups:
             sensitive information.
         brief: The request body for a [search-type query](https://www.elastic.co/guide/en/elasticsearch/reference/current/search.html), as a json string.
         examples: [ '"{\"query\":{\"term\":{\"user.id\":\"kimchy\"}}}"' ]
+        tag: tech-specific
```
lmolkova, line 243: nit: tags are not needed here anymore
```suggestion
```

model/trace/database.yaml
```diff
@@ -240,11 +240,21 @@ groups:
...
+      - ref: db.collection.name
+        requirement_level: recommended
+        brief: The index or data stream against which the query is executed.
+        note: >
+          If the query targets multiple indices or data streams, then the name of those should be added as a comma separated list.
+          If the query doesn't target a specific index, this field MUST NOT be set.
+        examples: [ 'my_index', 'index1, index2' ]
+        tag: tech-specific
```
lmolkova, line 251:
```suggestion
```

docs/database/elasticsearch.md
```diff
@@ -72,13 +69,19 @@ Tracing instrumentations that do so, MUST also set `http.request.method_original
 
 **[7]:** If using a port other than the default port for this DBMS and if `server.address` is set.
 
-**[8]:** When communicating with an Elastic Cloud deployment, this should be collected from the "X-Found-Handling-Cluster" HTTP response header.
+**[8]:** If the query targets multiple indices or data streams, then the name of those should be added as a comma separated list. If the query doesn't target a specific index, this field MUST NOT be set.
 
 **[9]:** When communicating with an Elastic Cloud deployment, this should be collected from the "X-Found-Handling-Instance" HTTP response header.
 
-**[10]:** Should be collected by default for search-type queries and only if there is sanitization that excludes sensitive information.
+**[10]:** If a database system has multiple namespace components, they SHOULD be concatenated (potentially using database system specific conventions) from most general to most specific namespace component, and more specific namespaces SHOULD NOT be captured without the more general namespaces, to ensure that "startswith" queries for the more general namespaces will be valid.
```
lmolkova, line 76: this note came from the generic `db.namespace` note in the registry, but it does not really make sense for elastic. 

You can override it when referencing `db.namespace` if you have anything to add there, or remove it (maybe possible with weaver).
It was not possible with older tooling and we have hacks like this one to remove the note

https://github.com/open-telemetry/semantic-conventions/blob/9381a31e07306453b38ea827f81885160ad8cada/model/trace/database.yaml#L158-L159

docs/database/elasticsearch.md
```diff
@@ -14,12 +14,8 @@ described on this page.
 
 ## Span Name
 
-The **span name** SHOULD be of the format `<endpoint id>`.
-
-The elasticsearch endpoint identifier is used instead of the url path in order to reduce the cardinality of the span
-name, as the path could contain dynamic values. The endpoint id is the `name` field in the
+The **span name** follows the [general database span name guidelines](database-spans.md#name) with the endpoint identifier stored in `db.operation.name`, instead of the url path in order to reduce the cardinality of the span and the index stored in `db.collection.name`. The endpoint id is the `name` field in the
```
lmolkova, line 17: > instead of the url path in order to reduce the cardinality of the span

Is this part still relevant? why would users expect url path here? Maybe there was some prior art, but we don't really have to mention it here.

> The endpoint id is the `name` field in the
[elasticsearch schema](https://raw.githubusercontent.com/elastic/elasticsearch-specification/main/output/schema/schema.json).

since we're reusing `db.operation.name` attribute as is in the span name, consider moving this to the `db.operation.name` note and removing it from the span name section.

.chloggen/influxdb.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add semantic convention of InfluxDB
```
lmolkova, line 13: oops

lmolkova:
@gregkalapos 
I was able to regenerate the registry and tables and pushed a commit with changes. Hope you don't mind.

I think you have some problem with an npm module `npm run fix:format` in https://github.com/open-telemetry/semantic-conventions/blob/0dfe9fb53d9888e9f7776aab06ffcfa80159cfde/Makefile#L1119-L126

PTAL at the error logs - maybe you need to install npm or there is some version conflict. If you believe there is something we can do better on the makefile, please create an issue or send a PR to fix :)

docs/database/elasticsearch.md
lmolkova: Just tried: 
```yaml
        note: ""
```

works like a charm now

gregkalapos:
> @gregkalapos I was able to regenerate the registry and tables and pushed a commit with changes (hope you don't mind). So it works on my machine™️.
> 
> I think you have some problem with `npm run fix:format` in
> 
> https://github.com/open-telemetry/semantic-conventions/blob/0dfe9fb53d9888e9f7776aab06ffcfa80159cfde/Makefile#L119-L126
> 
> PTAL at the error logs - maybe you need to install npm or there is some version conflict. If you believe there is something we can do better on the makefile/tooling, please create an issue or send a PR to fix :)

Great, thanks for looking into this. Installing Node.js solves the issue (just reinstalled my machine... )

.chloggen/influxdb.yaml
gregkalapos: Sorry, leftover from rebasing - `main` moves very fast today :) Should be ok now.

docs/database/elasticsearch.md
gregkalapos: Updated. 

> Is this part still relevant? why would users expect url path here?

That's a good point. Naturally people (and by that I mostly mean people who work on Elasticsearch and created the first instrumentations) pick the url path. But probably less relevant now. Removed it.


Also moved the link with the endpoint id into a note.

docs/database/elasticsearch.md
```diff
@@ -14,12 +14,7 @@ described on this page.
 
 ## Span Name
 
-The **span name** SHOULD be of the format `<endpoint id>`.
-
-The elasticsearch endpoint identifier is used instead of the url path in order to reduce the cardinality of the span
-name, as the path could contain dynamic values. The endpoint id is the `name` field in the
-[elasticsearch schema](https://raw.githubusercontent.com/elastic/elasticsearch-specification/main/output/schema/schema.json).
-If the endpoint id is not available, the span name SHOULD be the `http.request.method`.
+The **span name** follows the [general database span name guidelines](database-spans.md#name) with the endpoint identifier stored in `db.operation.name`, and the index stored in `db.collection.name`.
```
trask, line 17: is this needed? it seems like it is already following the general database span name guidelines? (and documenting what to put into `db.operation.name` and `db.collection.name` for elasticsearch can be done in overrides for those attributes)

gregkalapos: This info is definitely covered in the attributes, so in that sense we'd not loose any information if we drop this line. My thinking was that calling out specifically that we put the endpoint id into the span name could be useful for readers.

If we drop this part, then there is nothing said in `elasticsearch.md` about the span name. And since this is a bit different from other databases, I thought it's worth calling out. Otherwise the reader needs to put together the fact, that operation name is something special (the endpoint id) for Elasticsearch and that is the one in the span name.

I just though it's worth calling out that we expect span names in `endpoint id + index` format at the top of the `.md` file.

lmolkova: > And since this is a bit different from other databases, I thought it's worth calling out. Otherwise the reader needs to put together the fact, that operation name is something special (the endpoint id) for Elasticsearch and that is the one in the span name.

In the context of the span name there is no difference though? The difference is captured in the corresponding attribute briefs/notes.

gregkalapos: Yes, that's true.

I thought this is worth calling out extra. For people familiar with SemConv this is obvious, but if someone just implements this spec, I think that short sentence is very useful.

But if others think this is not needed I'm ok dropping this.

model/trace/database.yaml
```diff
@@ -240,11 +243,20 @@ groups:
             sensitive information.
         brief: The request body for a [search-type query](https://www.elastic.co/guide/en/elasticsearch/reference/current/search.html), as a json string.
         examples: [ '"{\"query\":{\"term\":{\"user.id\":\"kimchy\"}}}"' ]
+      - ref: db.collection.name
+        requirement_level: recommended
+        brief: The index or data stream against which the query is executed.
+        note: >
+          If the query targets multiple indices or data streams, then the name of those should be added as a comma separated list.
```
gregkalapos, line 250: Any thoughts on this?

Other parts of the db spec typically use the 1. item and drop the rest. Same for [`db.collection.name` in the general db span spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#common-attributes).

I don't see a huge benefit of comma separating index names here - that's already captured in `db.elasticsearch.path_parts.<key>`. I think it's better to be consistent with the general spec here, so I'd change this to only capture the 1. index name.

docs/database/elasticsearch.md
trask: I'm ok either way, we can always change our minds later as we build out more database system specific conventions since this kind of change can be done post stability

trask (approved):

model/trace/database.yaml
trask: > I think it's better to be consistent with the general spec here

👍

docs/database/elasticsearch.md
```diff
@@ -117,16 +116,17 @@ Tracing instrumentations that do so, MUST also set `http.request.method_original
...
+| `db.collection.name`                | `"my-index"`                                                                                                                        |
 | `url.full`                          | `"https://elasticsearch.mydomain.com:9200/my-index-000001/_search?from=40&size=20"`                                                 |
 | `db.elasticsearch.path_parts.index` | `"my-index-000001"`                                                                                                                 |
-| `db.elasticsearch.cluster.name`     | `"e9106fc68e3044f0b1475b04bf4ffd5f"`                                                                                                |
-| `db.instance.id`                    | `"instance-0000000001"`                                                                                                             |
+| `db.namespace`                      | `"e9106fc68e3044f0b1475b04bf4ffd5f"`                                                                                                |
```
lmolkova, line 129: nit: do people tend to name their clusters as guids? If not, maybe we can provide a friendly example)?

model/registry/deprecated/db.yaml
```diff
@@ -84,6 +84,13 @@ groups:
         brief: 'Deprecated, no general replacement at this time. For Elasticsearch, use `db.elasticsearch.node.name` instead.'
         deprecated: 'Deprecated, no general replacement at this time. For Elasticsearch, use `db.elasticsearch.node.name` instead.'
         examples: 'mysql-e26b99z.example.com'
+      - id: elasticsearch.cluster.name
+        type: string
+        stability: experimental
+        deprecated: Use `db.namespace` instead.
+        brief: >
+          Represents the identifier of an Elasticsearch cluster.
```
lmolkova, line 92: nit (to be consistent with other deprecated attributes): 

```suggestion
         Deprecated,  use `db.namespace` instead
```

model/trace/database.yaml
```diff
@@ -228,7 +228,10 @@ groups:
         requirement_level: required
       - ref: db.operation.name
         requirement_level: required
-        note: This SHOULD be the endpoint identifier for the request.
+        note: >
+          This SHOULD be the endpoint identifier for the request.
+
+          The endpoint id is the `name` field in the [elasticsearch schema](https://raw.githubusercontent.com/elastic/elasticsearch-specification/main/output/schema/schema.json).
```
lmolkova, line 234: nit:

maybe?
```suggestion
        note: >
          The `db.operation.name` SHOULD match the endpoint name provided in the request (see the [Elasticsearch schema](https://raw.githubusercontent.com/elastic/elasticsearch-specification/main/output/schema/schema.json)).
```

model/trace/database.yaml
```diff
@@ -240,11 +243,20 @@ groups:
...
+          If the query doesn't target a specific index, this field MUST NOT be set.
+        examples: [ 'my_index', 'index1, index2' ]
       - ref: server.address
       - ref: server.port
-      - ref: db.elasticsearch.cluster.name
+      - ref: db.namespace
+        note: ""
         requirement_level:
           recommended: >
+            The name of the Elasticsearch cluster which the client connects to.
```
lmolkova, line 259: this sentence should be in the brief, it's ok to override the inherited one.

model/trace/database.yaml
```diff
@@ -247,16 +246,16 @@ groups:
...
         examples: [ 'my_index', 'index1, index2' ]
       - ref: server.address
       - ref: server.port
       - ref: db.namespace
         note: ""
+        brief: The name of the Elasticsearch cluster which the client connects to.
         requirement_level:
           recommended: >
-            The name of the Elasticsearch cluster which the client connects to.
             When communicating with an Elastic Cloud deployment, this should be collected from the "X-Found-Handling-Cluster" HTTP response header.
```
lmolkova, line 259: nit: this sentence should probably be on the `note`, does not seem to be relevant to the `requirement_level`

lmolkova (approved):

swallez:
About the new `db.collection.name` attribute: the Elasticsearch API generally accepts a list of names and wildcards for index names, e.g. not only `foo`, but also `foo,bar,baz-*`. Will that be an issue? Expanding wildcards on the client side is not really an option because of the additional request it would require to have the list of matching indices (which can also be large).

trask:
> About the new `db.collection.name` attribute: the Elasticsearch API generally accepts a list of names and wildcards for index names, e.g. not only `foo`, but also `foo,bar,baz-*`. Will that be an issue? Expanding wildcards on the client side is not really an option because of the additional request it would require to have the list of matching indices (which can also be large).

my interpretation of the spec is that `db.collection.name` should capture the first index name

if the first index name has a wildcard, I think that's ok, and `db.collection.name` would include the wildcard, e.g. `foo*`

gregkalapos:
Thanks for raising this @swallez . 

> > About the new `db.collection.name` attribute: the Elasticsearch API generally accepts a list of names and wildcards for index names, e.g. not only `foo`, but also `foo,bar,baz-*`. Will that be an issue? Expanding wildcards on the client side is not really an option because of the additional request it would require to have the list of matching indices (which can also be large).
> 
> my interpretation of the spec is that `db.collection.name` should capture the first index name

Correct, so the example of `foo,bar,baz-*` is fine - it's just `foo` in that case. 

> if the first index name has a wildcard, I think that's ok, and `db.collection.name` would include the wildcard, e.g. `foo*`

That'd be my suggestion as well. Should we call this out explicitly? Happy to push a change.

gregkalapos:
@swallez there is also `db.elasticsearch.path_parts.<key>` with definition:

> A dynamic value in the url path. 

and examples:
> `db.elasticsearch.path_parts.index=test-index`; `db.elasticsearch.path_parts.doc_id=123`

So in your example, I'd expect `db.elasticsearch.path_parts.index=foo,bar` - so users would know from `db.elasticsearch.path_parts.index` why the span deleted from `bar`. 

So this is only for `db.collection.name` and for the span name.

Having said that, we already had an iteration of the comma separated vs. only keeping the first one question here: https://github.com/open-telemetry/semantic-conventions/pull/1002#discussion_r1622025093

I agree with you, this is a bit different since this is coming from the request url. @swallez what do you think: does `db.elasticsearch.path_parts.index` address this, or you still think we should revert back and add the comma separated list into `db.collection.name`, even though other database semantic conventions only take the 1. one?

swallez:
First, asking users to look at `db.elasticsearch.path_parts.index` for multi-index requests because `db.collection.name` is incomplete in that case would be a pretty bad experience IMHO.

Also, there are some deviations from using `index` as the path parameter: operations on data streams use `name` and operations on aliases use `alias`. Data streams and aliases can be considered as collections, as many operation equally accept indices, data streams and aliases.

I really think we should consider multi-index targets as a kind of runtime alias (or an SQL partitioned table) as it is semantically equivalent, and all target indices are expected to have the same schema. This is very different from comma separated table names in a SQL query like `select * from FOO, BAR where...` where there is a clear distinction between the left and right parts of a join.

So IMHO keeping the index parameter as is is _semantically consistent with the spec_, since it defines what the target of the operation is, which - in the case of wildcards or comma-separated names - is equivalent to a runtime-defined alias.

docs/database/elasticsearch.md
```diff
@@ -38,12 +33,13 @@ If the endpoint id is not available, the span name SHOULD be the `http.request.m
 | [`db.elasticsearch.path_parts.<key>`](/docs/attributes-registry/db.md) | string | A dynamic value in the url path. [4] | `db.elasticsearch.path_parts.index=test-index`; `db.elasticsearch.path_parts.doc_id=123` | `Conditionally Required` when the url has dynamic values | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [5] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` If and only if the operation failed. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 | [`server.port`](/docs/attributes-registry/server.md) | int | Server port number. [6] | `80`; `8080`; `443` | `Conditionally Required` [7] | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
-| [`db.elasticsearch.cluster.name`](/docs/attributes-registry/db.md) | string | Represents the identifier of an Elasticsearch cluster. | `e9106fc68e3044f0b1475b04bf4ffd5f` | `Recommended` [8] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
trask, line 41: is it accurate to add a schema transform from `db.elasticsearch.cluster.name` -> `db.namespace`?

gregkalapos:
> Also, there are some deviations from using index as the path parameter: operations on data streams use name and operations on aliases use alias. Data streams and aliases can be considered as collections, as many operation equally accept indices, data streams and aliases.

Yeah, that indeed makes this too complicated to figure out for users.

I pushed an update to the PR now stating that it should be a comma separated list.

At the same time I also opened https://github.com/open-telemetry/semantic-conventions/issues/1132 to discuss this in general. We had a conversation about this in the working group and we thought it'd be useful to reiterate the decision that only the 1. collection name should be captured. Regardless of the outcome of that issue, my current proposal would be to move on with this PR and explicitly stating here that collection name should be a comma separated list of all targets.

docs/database/elasticsearch.md
gregkalapos: Yes 👍  Pushed a change with it.

model/trace/database.yaml
```diff
@@ -238,16 +240,24 @@ groups:
...
+        examples: [ 'my_index', 'index1, index2' ]
       - ref: server.address
       - ref: server.port
-      - ref: db.elasticsearch.cluster.name
-        requirement_level:
-          recommended: >
-            When communicating with an Elastic Cloud deployment, this should be collected from the "X-Found-Handling-Cluster" HTTP response header.
+      - ref: db.namespace
+        note: >
+          When communicating with an Elastic Cloud deployment, this should be collected from the "X-Found-Handling-Cluster" HTTP response header.
```
trask, line 254: this could prevent us from making `db.namespace` a sampling relevant attribute: https://github.com/open-telemetry/semantic-conventions/issues/1019#issuecomment-2142324021

trask: I forgot, we changed MUST to SHOULD here:

> The following attributes can be important for making sampling decisions and SHOULD be provided at span creation time (if provided at all)

so this is not a big problem

trask (approved):

docs/database/elasticsearch.md
lmolkova: I believe this discussion is resolved.

PR #1005, opened 2024-05-04
Add registry md check and fix violations

lmolkova:
Apparently we never check that registry is up to date


jsuereth (approved):
Wierd - I thought I had updated a `check` command in the makefile for this, but it looks like I must have blown that away in one of the merge-conflicts.


Thanks for fixing!

joaopgrassi (approved):

Issue #937, opened 2024-04-18
Review naming conventions used for messaging metrics


pyohannes:
Currently, [messaging metric names](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-metrics.md) adhere to the convention of `messaging.<operation type>.[duration|messages]`.

It should be revisited whether these logic is extensible and expressive enough. Some things to consider:

* `messaging.operation` was renamed to `messaging.operation.type`, `messaging.operation.name` was introduced to allow system-specific names for operations. It needs to be discussed whether this system-specific information in `messaging.operation.name` is also relevant for metrics.
* For database, the metric name [`db.client.operation.duration`](https://github.com/open-telemetry/semantic-conventions/pull/735) is proposed. It needs to be discussed in how far DB and messaging should be consistent in that regard.
* There is #895 which proposed a metric for measuring message latency (and lag). It needs to be investigated how such metrics fit in with currently defined metrics.

pyohannes:
I think it makes sense to follow the current path insofar as we keep separate metrics for different operation types (`publish`, `receive`, and `process`) and not use a generic metric like it is done for DB metrics.

The main reason for this is the fact that I don't see a use case for aggregating metrics for both `publish` and `process` operations without regarding the operation type. This means, one with likely not ever use a metric `messaging.operation.duration` without filtering for a specific operation type.

pyohannes:
### Proposal 1

We stick with the current list of metrics and the pattern of having `messaging.<operation name>.[duration|messages]`:

 * `messaging.publish.duration`
 * `messaging.publish.messages`
 *  `messaging.receive.duration`
 * `messaging.receive.messages`
 * `messaging.process.duration`
 * `messaging.process.messages`

We could extend this by adding different "synthetic" operation types for metrics that don't map to any defined operation (for example metrics for message latency or lag):

* `messaging.enqueued.duration` (latency)
* `messaging.enqueued.messages` (lag)

### Proposal 2

We follow a pattern of `messaging.[producer|consumer].operation.[duration|messages]`, which would be consistent with the pattern proposed for DB metrics. The operation name or type would be an attribute on the metric.

Here we would end up with the following metrics:

* `messaging.producer.operation.duration`
* `messaging.producer.operation.messages`
* `messaging.consumer.operation.duration`
* `messaging.consumer.operation.messages`

Metrics for latency and lag could be grouped under the `consumer` namespace, for example:

[...]

PR #1006, opened 2024-05-04
Revamp messaging metrics to support generic operations

### LINKED ISSUE #947: Messaging.operation.type attribute should not be required
Operation type identifies common messaging operations (which may be called differently in specific systems): e.g. `publish` can be called `send`, `schedule`, or `enqueue`.

In such  cases we use `messaging.operation.name` to represent system-friendly name.

However there are other types of operations that don't match any common type. E.g, `cancel scheduled message`, `delete`, `create topic`, etc...

While those are not targeted for stability, we should make sure `messaging.operation.type` requirement level is adequate and future-proof.

_Originally posted by @lmolkova in https://github.com/open-telemetry/semantic-conventions/pull/942#discussion_r1575412911_

lmolkova:
Fixes #947, #937

## Changes

- make `messaging.operation.name` required and `messaging.operation.type` conditionally required
- update span name accordingly (and follow other semconv pattern of `{operation} {destination}`)
- update metrics:
    - Make `messaging.operation.name` required and `messaging.operation.type` conditionally required when type is applicable.
    - Unify `messaging.publish.duration` and `messaging.receive.duration` metrics into `messaging.client.operation.duration`
    - Unify `messaging.receive.messages` and `messaging.process.messages` metrics into `messaging.client.consumed.messages`


docs/messaging/messaging-metrics.md
```diff
@@ -93,92 +92,76 @@ the broker doesn't have such notion, the destination name SHOULD uniquely identi
...
+### Metric: `messaging.client.consumed.messages`
 
-This metric is [required][MetricRequired] for operations that are initiated by the application code (pull-based).
+This metric measures number of messages that were consumed by the application. If messaging client supports pull mode, this metric SHOULD represent the number of successfully pulled
+messages. If messaging client pushes messages to the application, it SHOULD represent the number of messages that were dispatched to the application.
 
-This metric SHOULD be specified with
-[`ExplicitBucketBoundaries`](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.31.0/specification/metrics/api.md#instrument-advice)
-of `[ 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10 ]`.
+The message SHOULD be counted each time it's received and consumed.
```
lmolkova, line 135:
```suggestion
The message SHOULD be counted each time it's consumed.
```

docs/messaging/messaging-metrics.md
```diff
@@ -93,92 +92,76 @@ the broker doesn't have such notion, the destination name SHOULD uniquely identi
...
 
-When this metric is reported alongside a messaging receive span, the metric value SHOULD be the same as the corresponding span duration.
+If instrumentation tracks pull-receiving and processing for one message, it SHOULD track this message once within `messaging.client.consumed.messages` metric.
```
lmolkova, line 137: this is probably not right. Knowing if messages are dropped between receiving and processing would be really useful, so maybe we need to count both operations.

docs/messaging/messaging-metrics.md
```diff
@@ -7,14 +7,12 @@
...
 - [Producer metrics](#producer-metrics)
-  - [Metric: `messaging.publish.duration`](#metric-messagingpublishduration)
-  - [Metric: `messaging.publish.messages`](#metric-messagingpublishmessages)
+  - [Metric: `messaging.client.published.messages`](#metric-messagingclientpublishedmessages)
 - [Consumer metrics](#consumer-metrics)
-  - [Metric: `messaging.receive.duration`](#metric-messagingreceiveduration)
-  - [Metric: `messaging.receive.messages`](#metric-messagingreceivemessages)
-  - [Metric: `messaging.process.duration`](#metric-messagingprocessduration)
-  - [Metric: `messaging.process.messages`](#metric-messagingprocessmessages)
+  - [Metric: `messaging.client.consumed.messages`](#metric-messagingclientconsumedmessages)
```
pyohannes, line 14: If have some thoughts about the counters `messaging.client.[published|consumed].messages`.

1. I wonder if we should change those from opt-in and required for batching scenarios to always required? As now we don't anymore have a system-independent way of determining those values (we'd need to slice the `operation.duration` metrics by the exact and system specific operation name).
2. The pairing of the terms "publish" and "consume" seems a little bit odd to me, usually it's "produce" and "consume". Maybe "published" and "received" would be better?

pyohannes (commented):
Thanks @lmolkova, I think this is going in a good direction.

docs/attributes-registry/messaging.md
```diff
@@ -89,7 +76,8 @@ Describes deprecated messaging attributes.
...
 | `messaging.kafka.destination.partition` | int    | Deprecated, use `messaging.destination.partition.id` instead. | `2`                               | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `messaging.destination.partition.id`. |
-| `messaging.operation`                   | string | Deprecated, use `messaging.operation.type` instead.           | `publish`; `create`; `process`    | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `messaging.operation.type`.           |
+| `messaging.operation`                   | string | Deprecated, use `messaging.operation.name` instead.           | `publish`; `create`; `process`    | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `messaging.operation.name`.           |
+| `messaging.operation.type`              | string | Deprecated, use `messaging.operation.name` instead.           | `publish`; `create`; `process`    | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `messaging.operation.name`.           |
```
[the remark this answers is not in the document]
lmolkova: unfortunately we did :)

docs/messaging/messaging-metrics.md
```diff
@@ -32,6 +30,7 @@ All messaging metrics share the same set of attributes:
 <!-- semconv metric.messaging.attributes(full) -->
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
+| [`messaging.operation.name`](/docs/attributes-registry/messaging.md) | string | The system-specific name of the messaging operation. | `ack`; `nack`; `send` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
lmolkova: I can still publish messages using different operations like `schedule` or `send` within a single system. It also captures system-specific terminology, so does not seem to hurt.

docs/messaging/messaging-metrics.md
lmolkova: > I wonder if we should change those from opt-in and required for batching scenarios to always required? As now we don't anymore have a system-independent way of determining those values (we'd need to slice the operation.duration metrics by the exact and system specific operation name).

Agreed, done

lmolkova: >The pairing of the terms "publish" and "consume" seems a little bit odd to me, usually it's "produce" and "consume". Maybe "published" and "received" would be better?

I really like `consumed` because it describes both receiving and processing. 
If I search for "publish and consume" I still see them used together, but I'm fine changing to `messaging.client.produced.messages`

.chloggen/1006.yaml
```diff
@@ -0,0 +1,14 @@
+change_type: breaking
+
+component: messaging
+
+note: |
+  Support generic operations in messaging rename_metrics:
```
trask, line 6: (?)
```suggestion
  Support generic operations in messaging and rename metrics:
```

.chloggen/1006.yaml
```diff
@@ -0,0 +1,14 @@
...
+component: messaging
+
+note: |
+  Support generic operations in messaging and rename metrics:
+
+    - Deprecate `messaging.operation.type` in favor of now required `messaging.operation.name`.
+    - Rename `messaging.publish.messages` to `messaging.client.produced.messages`
+    - Rename `messaging.process.duration` to `messaging.client.process.duration`
+    - Unify `messaging.publish.duration` and `messaging.receive.duration` into `messaging.client.operation.duration`
+    - Unify `messaging.receive.messages` and `messaging.process.messages` into `messaging.client.consumed.messages`
```
trask, line 12: does this render ok in change log? (I didn't see prior examples)

docs/messaging/messaging-spans.md
```diff
@@ -153,38 +153,32 @@ in such a way that it cannot be changed by intermediaries.
...
-```
+The <span id="destination-placeholder">`{destination}`</span> SHOULD adhere to one of the following values, arranged in prioritized order, provided they are accessible:
 
-The destination name SHOULD only be used for the span name if it is known to be of low cardinality (cf. [general span name guidelines](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.31.0/specification/trace/api.md#span)).
-This can be assumed if it is statically derived from application code or configuration.
-Wherever possible, the real destination names after resolving logical or aliased names SHOULD be used.
-If the destination name is dynamic, such as a [conversation ID](#conversations) or a value obtained from a `Reply-To` header, it SHOULD NOT be used for the span name.
-In these cases, an artificial destination name that best expresses the destination, or a generic, static fallback like `"(anonymous)"` for [anonymous destinations](#temporary-and-anonymous-destinations) SHOULD be used instead.
+- `messaging.destination.template`
+- `messaging.destination.name` if it is known to be of a low cardinality.
```
trask, line 168: can we say anything more specific than "low cardinality"? my specific concern is that generic messaging instrumentation may not be able to guarantee someone doesn't have 10,000 destinations and so the generic instrumentation might choose not to put this in the span name which I don't think is what we want

docs/messaging/messaging-metrics.md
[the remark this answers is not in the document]
pyohannes: Should we go a step further and make the counters as generic as the duration metrics? So we'd have two symmetric pairs of metrics:
* `messaging.client.operation.duration`
* `messaging.client.operation.messages`
* `messaging.client.process.duration`
* `messaging.client.process.messages`

This has the advantage of providing proper counters for _all_ operations. On the other hand, it is a bit more cumbersome to use (one needs to know the system-specific operation name to be able to count produced messages).

> so does not seem to hurt.

I agree it doesn't hurt and this isn't a blocker, I just think that for most cases this (required) attribute doesn't provide much use on counter metrics.

docs/messaging/messaging-spans.md
pyohannes:
```suggestion
- `messaging.destination.name` if it is statically derived from application code or configuration.
```

Maybe that's better? This will result in a low cardinality in most cases, and avoids that instrumentation authors need to reason about what high or low cardinality means.

docs/messaging/messaging-spans.md
```diff
@@ -153,38 +153,32 @@ in such a way that it cannot be changed by intermediaries.
...
+* `subscribe shop.orders`
+* `ack shop.orders`
+* `nack print_jobs `
+* `process topic with spaces `
+* `settle AuthenticationRequest-Conversations`
 
-### Operation types
+### Common messaging operations
 
 The following operation types related to messages are defined for these semantic conventions:
```
pyohannes, line 184: As we remove the operation type attribute, we shouldn't write about "operation types" anymore but only about "operation names".

Also we'd need to clarify the meaning of the table above. Would we recommend using those well-known names (if applicable)? Should we require systems to document other names that are used and not in this list? Do we always require the operation name "process" for processing operations?

docs/messaging/messaging-metrics.md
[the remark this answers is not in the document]
pyohannes: As discussed in the workgroup meeting, we decided that we will further explore the generic metrics proposed above (exact names are still open for discussion):
* `messaging.client.operation.duration`
* `messaging.client.operation.messages`
* `messaging.client.process.duration`
* `messaging.client.process.messages`

This will also require the presence of the `messaging.operation.type` attribute, so that generic operation counts can be aggregated towards standardized operation types.

### Example

In this example two messages are published (one via a "send" operation, the other one via a "schedule" operation), for one message also a "create" operation was instrumented:
```
messaging.client.operation.messages{messaging.operation.type=create} 1
messaging.client.operation.messages{messaging.operation.type=publish, messaging.operation.name=send} 1
messaging.client.operation.messages{messaging.operation.type=publish, messaging.operation.name=schedule} 1
```
Here a count of all operations with the type "publish" would give the number of all published messages (initially the dedicated `messaging.client.published.message` was intended for this).

Then two messages are received and both are settled via one "commit" operation:

```
messaging.client.operation.messages{messaging.operation.type=receive, messaging.operation.name=consume} 2
messaging.client.operation.messages{messaging.operation.type=settle, messaging.operation.name=commit} 1
```

Here a count of all operations with the type "receive" would give the number of all consumed messages (initially the dedicated `messaging.client.consumed.message` was intended for this).

### Problem

It's not clear to me how to reliable count consumed messages with this approach. There could possibly be just a single "receive" operation for a message (pull-based consumption with out "process" instrumentation), there could be just a single "process" operation for a message (push-based consumption without "receive" instrumentation), or there could be both "receive" and "process" operations for a single message (pull-based consumption with "process" instrumentation).

Given this situation, it's not clear to me how to deduce the number of consumed messages from `messaging.client.operation.messages` and `messaging.client.process.messages`.

docs/messaging/messaging-spans.md
lmolkova: whether we remove it or not, we still need to establish terminology for the rest of the document. If we remove the attribute, we can add a note that it's the only goal and that type does not appear on the telemetry.

docs/messaging/messaging-spans.md
lmolkova: I rephrased to 

> The `messaging.destination.template` SHOULD be used whenever it's available. Otherwise, `messaging.destination.name` SHOULD be used only when
  the destination is known to be neither [temporary nor anonymous](#temporary-and-anonymous-destinations).

WDYT?

(also aligned with our DB span name changes in #1069)

docs/messaging/messaging-metrics.md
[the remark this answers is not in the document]
pyohannes: Given that the number of consumed messages can't be deduced from the number of processed and the number of received messages, I propose to keep generic operation metrics for flexibility, but to additionally add counters for produced and consumed messages:

* `messaging.client.operation.duration`
* `messaging.client.operation.messages`
* `messaging.client.process.duration`
* `messaging.client.process.messages`
* `messaging.client.produced.messages`
* `messaging.client.consumed.messages`

Reasons:
* We agreed that "process" is a special operation that's important for default dashboards, so we keep this as a separate metric.
* All other operations are tracked via the generic "operation" metrics (similar to database metrics).
* We keep track of counts of produced and consumed messages independent of operations. This might create some redundancy, as the count of produced messages could be deduced from operation counts, given we keep `messaging.operation.type`. However, keeping dedicated counters here makes the whole model more clear and user-friendly, and also enables us to go ahead without `messaging.operation.type`.

@lmolkova @joaopgrassi @trask  What do you think?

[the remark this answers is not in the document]
joaopgrassi: Also: In pull based scenarios, we will always have a "receive" operation, so for that case we can consistently count the received messages. The problem seems to be on push-based scenarios, as there we can't guarantee that a process span/operation will happen as it's in user land code.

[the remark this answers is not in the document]
lmolkova: it feels like we're trying to make metrics work with existing operation types. What if we change operation types instead?
what if we introduced a `consume` operation type instead of `process`. `process` is not a messaging operation type - it's special. We need the terminology, but not necessarily a type

Let's do a mental experiment:

- `consume` operation captures that a message is dispatched to the application
   - it's pretty much the same as `delivery` operation, just a different word since `delivered.messages` is ambiguous
   - message is consumed when it's dispatched to the application, it's counted once either at the end of `receive` or right before `process` starts
  - we need to document this moment anyway to tell when to update the message counter and to provide time-in-queue/delay metrics
- spans: 
    - push-based: `consume` type should be set on the process span. 
    - pull-based: no `consume` span
- metrics: 
  - `messaging.client.operation.duration` is never reported for `consume` operation type - it's in the `process.duration`
  - `messaging.client.operation.messages` is reported for it 

So we'd have:
- `messaging.client.operation.duration`: for `publish`, `receive` and untyped operations
- `messaging.client.operation.messages`: for all defined operation types and untyped operations except processing
- `messaging.client.process.duration`: for processing duration
- `messaging.client.process.messages`: for processing throughput

WDYT?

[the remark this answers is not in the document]
lmolkova: it feels like we're trying to make metrics work with existing operation types. What if we change operation types instead?
what if we used `receive` for all kinds of deliveries (push and pull). Maybe we can come up with a better name for it?

Let's do a mental experiment:

- `receive` operation captures a process or receiving a message in one way or another. 
- spans: 
    - `receive` is always created. It's always a consumer span.
    - `process` can be created in addition to `receive`
- metrics: 
  - `messaging.client.operation.duration` is not reported for  push-based `receive` operation type - it's in the `process.duration`. 
  - `messaging.client.operation.messages` is reported for all operations that deal with messages

So we'd have:
- `messaging.client.operation.duration`: for all operations except push-based `receive` and `process`
- `messaging.client.operation.messages`: for all operations except processing
- `messaging.client.process.duration`: for processing duration
- `messaging.client.process.messages`: for processing throughput

TBH I no longer understand why we want to differentiate process operation. 

WDYT?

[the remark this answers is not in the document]
lmolkova: Update the PR with this revolutionary suggestion to demonstrate it. Let's discuss on the sig meeting.

docs/attributes-registry/messaging.md
```diff
@@ -63,10 +63,8 @@ size should be used.
 | Value     | Description                                                                                                                                                                                                            | Stability                                                        |
 | --------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------- |
 | `create`  | A message is created. "Create" spans always refer to a single message and are used to provide a unique creation context for messages in batch publishing scenarios.                                                    | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `process` | One or more messages are delivered to or processed by a consumer.                                                                                                                                                      | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 66: not sure if we strictly need removed operations types but happy to bring them back.

docs/messaging/azure-messaging.md
```diff
@@ -24,23 +24,22 @@ The following additional attributes are defined:
...
 | [`messaging.servicebus.message.delivery_count`](/docs/attributes-registry/messaging.md) | int | Number of deliveries that have been attempted for this message. | `2` | `Conditionally Required` [7] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`server.address`](/docs/attributes-registry/server.md) | string | Server domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name. [8] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Conditionally Required` If available. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`messaging.client.id`](/docs/attributes-registry/messaging.md) | string | A unique identifier for the client that consumes or produces a message. | `client-5`; `myhost@8742@s8083jm` | `Recommended` If a client id is available | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 36: todo: not applicable, remove

docs/messaging/azure-messaging.md
```diff
@@ -129,21 +126,20 @@ The following additional attributes are defined:
...
+| [`messaging.operation.type`](/docs/attributes-registry/messaging.md) | string | A string identifying the type of the messaging operation. [6] | `publish`; `create`; `receive` | `Conditionally Required` If applicable. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`server.address`](/docs/attributes-registry/server.md) | string | Server domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name. [7] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Conditionally Required` If available. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`messaging.client.id`](/docs/attributes-registry/messaging.md) | string | A unique identifier for the client that consumes or produces a message. | `client-5`; `myhost@8742@s8083jm` | `Recommended` If a client id is available | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 137: todo: not applicable, remove

docs/messaging/messaging-metrics.md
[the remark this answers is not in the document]
pyohannes: Thanks for your feedback and ideas.

> what if we used receive for all kinds of deliveries (push and pull). Maybe we can come up with a better name for it?

Wouldn't that make it meaningless to compare the duration of "receive" operations, when it could measure either a push or a pull operation? Also, would this mean that one needs two spans to capture "process" operations (as "receive" will always be required)?

> it feels like we're trying to make metrics work with existing operation types.

From my point of view that's not a strict requirement. At a minimum we need to be able to align duration metrics with span durations.

We should strive to design metrics so that the following questions can be easily answered, ideally without knowing many details about system-specifics:
* How many message were produced?
* How many messages were consumed?
* What's the processing duration?
* What's the error rate for any of the operations above (producing, consuming, and processing)?

For other questions I wouldn't mind if it would be a bit more complicated (and maybe involve system-specific knowledge) to extract information from metrics. Examples include receive duration, publish duration, or settlement related metrics.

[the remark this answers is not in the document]
lmolkova: Based on the discussion I changed the PR to 
- keep operation.type
- no changes around operation types
- introduce following metrics:
  - messaging.client.request.duration (for all operations except processing)
  - messaging.process.duration 
  - messaging.client.published.messages 
  - messaging.client.consumed.messages (all consumed messages, either received or processed, counted once per message delivery)

As a result:

Spans: 
- consumer span represents either receive or process operation depending on the scenario. `msg.operation.type=process` tells it's processing. It can be used to convert spans to metrics.

Metrics:
- Message throughput: `messaging.client.published.messages` for produced, `messaging.client.consumed.messages` for consumed. No metric for count of processed messages. We can always add one later.
- Request throughput is derived from `messaging.client.request.duration` grouped by operation name/type and `messaging.process.duration`
- Error rate: derived from `messaging.client.request.duration`  and `messaging.process.duration`


Naming and other suggestions are very welcome.

docs/messaging/messaging-metrics.md
```diff
@@ -1,20 +1,18 @@
...
 
 **Status**: [Experimental][DocumentStatus]
 
 <!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->
 
 <!-- toc -->
 
-- [Common attributes](#common-attributes)
+- [Common metrics](#common-metrics)
+  - [Metric: `messaging.client.request.duration`](#metric-messagingclientrequestduration)
```
pyohannes, line 10: Why using `messaging.client.request.duration` instead of `messaging.client.operation.duration` (and be consistent with DB metrics)?

docs/messaging/messaging-metrics.md
```diff
@@ -88,6 +119,18 @@ the broker doesn't have such notion, the destination name SHOULD uniquely identi
 | `_OTHER` | A fallback error value to be used when the instrumentation doesn't define a custom value. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 
 
+`messaging.operation.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `create` | A message is created. "Create" spans always refer to a single message and are used to provide a unique creation context for messages in batch publishing scenarios. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `deliver` | Deprecated. Use `process` instead. | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `process`. |
```
pyohannes, line 127: Do we need this? It might make sense to have it in the attribute registry, but I think it's confusing to have it (indefinitely?) in the conventions.

model/metrics/messaging-metrics.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+  - id: metric.messaging.client.consumed.messages
+    type: metric
+    metric_name: messaging.client.consumed.messages
+    brief: "Number of messages that were delivered to the application."
+    note: >
+      Records the number of messages pulled from the broker or number of messages dispatched to the application in push-based scenarios.
+
+      The metric MUST be reported once per message delivery. For example, if receiving and processing operations are both instrumented for a single message delivery, the number of received messages
+      is reported when then message is received and not reported for the processing operation.
```
pyohannes, line 73:
```suggestion
    brief: "Number of messages that were forwarded to the application."
    note: >
      Records the number of messages pulled from the broker or number of messages dispatched to the application in push-based scenarios.

      The metric MUST be reported once for each forwarded message. For example, if receiving and processing operations are both instrumented for a single message delivery, the number of received messages
      is reported when then message is received and not reported for the processing operation.
```

Maybe avoiding the term "delivery", as it connotates push-based scenarios?

docs/messaging/messaging-spans.md
```diff
@@ -193,32 +192,28 @@ The following operation types related to messages are defined for these semantic
...
-| `process`      | `CONSUMER` for push-based scenarios where no `receive` span exists. |
-
-For cases not covered by the table above, the span kind should be set according
-to the [generic specification about span kinds](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.33.0/specification/trace/api.md#spankind),
-e. g. it should be set to CLIENT for the "Publish" span if its context is not
-used as creation context and if the "Publish" span models a synchronous call to
-the intermediary.
+| `publish`      | `PRODUCER` if the context of the "Publish" span is used as creation context, `CLIENT` otherwise. |
+| `receive`    | `CONSUMER` |
+| `process`    | `CONSUMER` for push-based scenarios where no "Receive" span exists, `SERVER` otherwise. |
```
pyohannes, line 207:
```suggestion
| `publish`      | `PRODUCER` if the context of the "Publish" span is used as creation context. |
| `receive`    | `CONSUMER` |
| `process`    | `CONSUMER` for push-based scenarios where no "Receive" span exists. |
```

Could we leave this out of this PR, and discuss this in a separate issue/PR?

I know of some tools (including Grafana Application Observability) that assume ephemeral (uninstrumented) services for client spans without corresponding server spans (and vice versa). This assumption can be valid for the "publish" case (where a possible server span would be part of a broker trace), however, it's not valid for the "process" case (there isn't really any upstream client span).

I'm not in principal against this change, but I'd like to think and reason more about it independently of metrics.

docs/messaging/messaging-metrics.md
lmolkova: there is no way to not have it 🤐

model/metrics/messaging-metrics.yaml
lmolkova: I use it it to emphasize that it's not per message, but per message delivery. A message can be received/delivered multiple times.

'Delivery' is used in general messaging scenarios without pull/push assumption - e.g. 
- JMS - https://javaee.github.io/javaee-spec/javadocs/javax/jms/Message.html#getJMSRedelivered--
- ActiveMQ - https://activemq.apache.org/components/artemis/documentation/latest/undelivered-messages
- RabbitMQ - https://www.rabbitmq.com/docs/consumers#message-properties

docs/messaging/messaging-metrics.md
lmolkova: I want to make the difference between operations that application initiates and processing obvious in the metric name.

Processing is also an operation, so someone might try to look for `process` operation in `messaging.client.operation.duration`. 

Names like `messaging.client.request|call.duration` would make it obvious.

While databases do support change-feeds and may have process-like operation, it's not the key scenario and database semconv does not really have this problem.

pyohannes: Ok, this makes sense (and is consistent with HTTP conventions).

> Processing is also an operation, so someone might try to look for `process` operation in `messaging.client.operation.duration`.

On the other hand, someone might not expect `create` operations in `messaging.client.request.*`, as it doesn't involve a broker request?

docs/messaging/messaging-metrics.md
pyohannes: I see you submitted https://github.com/open-telemetry/semantic-conventions/pull/1110, which will take care of this. Thanks!

model/metrics/messaging-metrics.yaml
pyohannes: Makes sense. I was wondering about different terms we're using ("consume", "receive", "deliver", "forward") and being clear about the usage. It makes sense to use "deliver" for the whole process of getting a message from the broker to the application code, regardless of pull/push scenarios.

docs/messaging/messaging-spans.md
pyohannes: I created https://github.com/open-telemetry/semantic-conventions/issues/1112 to discuss this further.

docs/messaging/messaging-spans.md
lmolkova: resolving since we're keeping the operation type

docs/messaging/messaging-spans.md
```diff
@@ -193,20 +192,19 @@ The following operation types related to messages are defined for these semantic
...
 | `receive`      | One or more messages are requested by a consumer. This operation refers to pull-based scenarios, where consumers explicitly call methods of messaging SDKs to receive messages. |
-| `process`      | One or more messages are delivered to or processed by a consumer. |
+| `process`      | One or more messages are processed by a consumer. |
 | `settle`       | One or more messages are settled. |
 
 ### Span kind
 
-[Span kinds](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.33.0/specification/trace/api.md#spankind)
-SHOULD be set according to the following table, based on the operation type a span describes.
+Span kind SHOULD be set according to the following table, based on the operation type a span describes.
```
lmolkova, line 200: removing a link here for brevity because it's also mentioned under the table.

docs/messaging/messaging-spans.md
lmolkova: make sense, removed

docs/messaging/messaging-metrics.md
lmolkova: > On the other hand, someone might not expect create operations in messaging.client.request.*, as it doesn't involve a broker request?

I'd say there is no point in reporting `create` duration since it's 0. So it checks out :) I can mention it explicitly.

docs/messaging/messaging-spans.md
```diff
@@ -153,36 +153,35 @@ in such a way that it cannot be changed by intermediaries.
...
-If the destination name is dynamic, such as a [conversation ID](#conversations) or a value obtained from a `Reply-To` header, it SHOULD NOT be used for the span name.
-In these cases, an artificial destination name that best expresses the destination, or a generic, static fallback like `"(anonymous)"` for [anonymous destinations](#temporary-and-anonymous-destinations) SHOULD be used instead.
+The <span id="destination-placeholder">`{destination}`</span> SHOULD describe the entity that the operation is performed against
+and SHOULD adhere to one of the following values, provided they are accessible:
 
-Examples:
+- `messaging.destination.template` or `messaging.destination.name` SHOULD be used for operations on messages or messaging destinations.
+  The `messaging.destination.template` SHOULD be used whenever it's available. Otherwise, `messaging.destination.name` SHOULD be used only when
+  the destination is known to be neither [temporary nor anonymous](#temporary-and-anonymous-destinations).
+- `server.address:server.port` SHOULD be used for other operations not targeting any specific destination(s).
```
pyohannes, line 173:
```suggestion
1. `messaging.destination.template` SHOULD be used when it is available.
2. `messaging.destination.name` SHOULD be used when the destination is known to be neither [temporary nor anonymous](#temporary-and-anonymous-destinations).
3. `server.address:server.port` SHOULD be used for other operations not targeting any specific destination(s).
4. Otherwise, no destination SHOULD be specified.
```

docs/messaging/messaging-spans.md
```diff
@@ -153,36 +153,35 @@ in such a way that it cannot be changed by intermediaries.
...
 
-The destination name SHOULD only be used for the span name if it is known to be of low cardinality (cf. [general span name guidelines](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.33.0/specification/trace/api.md#span)).
-This can be assumed if it is statically derived from application code or configuration.
-Wherever possible, the real destination names after resolving logical or aliased names SHOULD be used.
-If the destination name is dynamic, such as a [conversation ID](#conversations) or a value obtained from a `Reply-To` header, it SHOULD NOT be used for the span name.
-In these cases, an artificial destination name that best expresses the destination, or a generic, static fallback like `"(anonymous)"` for [anonymous destinations](#temporary-and-anonymous-destinations) SHOULD be used instead.
+The <span id="destination-placeholder">`{destination}`</span> SHOULD describe the entity that the operation is performed against
+and SHOULD adhere to one of the following values, provided they are accessible:
```
pyohannes, line 168:
```suggestion
and SHOULD adhere to the first applicable option from the following list:
```

Maybe that makes it a bit clearer?

docs/messaging/messaging-spans.md
```diff
@@ -153,36 +153,35 @@ in such a way that it cannot be changed by intermediaries.
...
 
-The span name SHOULD be set to the message destination name and the name of the operation being performed (as captured in [`messaging.operation.name`](../attributes-registry/messaging.md)) in the following format:
+Messaging spans SHOULD follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.31.0/specification/trace/api.md#span).
 
-```
-<destination name> <operation name>
-```
+<!-- markdown-link-check-disable -->
+<!-- HTML anchors are not supported https://github.com/tcort/markdown-link-check/issues/225-->
+The **span name** SHOULD be `{messaging.operation.name} {destination}` (see below for the exact definition of the [`{destination}`](#destination-placeholder) placeholder).
```
pyohannes, line 160: What's the main reason for turning this around (the operation name before the destination)? Consistency with DB?

pyohannes (approved):
I have some smaller remarks, but nothing that should block merging this. I'm looking forward to first prototypes based on these changes.

docs/messaging/messaging-spans.md
lmolkova: Yes, with DB and HTTP - they all share the same pattern {verb} {target} which seems to be more English-friendly.

docs/messaging/messaging-spans.md
lmolkova: We discovered fallback to be problematic in DB semconv.

Let's say someone publishes a message to a temp destination and there is no template.
This operation usually (on other destinations) is called `publish {detaintion.name}`, but since it's a temp operation, the fallback would result in `publish {address:port}`.
So fallback would result in inconsistent operation names.

What we want is the entity operation is done against (broker or a specific queue|topic).
So instead of falling back, we'd rather not populate destination part of the span name.

docs/messaging/messaging-spans.md
lmolkova: rephrased a bit, thanks!

.chloggen/1006.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+component: messaging
+
+note: |
+  Support generic operations in messaging and rename metrics:
+
+    - Make `messaging.operation.name` required and `messaging.operation.type` conditionally required when type is applicable.
+    - Rename `messaging.publish.messages` metric to `messaging.client.published.messages`
+    - Unify `messaging.publish.duration` and `messaging.receive.duration` metrics into `messaging.client.request.duration`
+    - Unify `messaging.receive.messages` and `messaging.process.messages` metrics into `messaging.client.consumed.messages`
```
joaopgrassi, line 11: Shouldn't these also be added to the schema change (renamed)?

joaopgrassi (approved):

.chloggen/1006.yaml
lmolkova: good question. It's a rename with two new attributes, so we don't have a way to transform old -> new (`messaging.operation.name` was not available in any shape on the old metric).

trask:
> * deprecate `messaging.operation.type`

looks like drift on this part of PR description

model/metrics/messaging-metrics.yaml
```diff
@@ -0,0 +1,80 @@
...
+        requirement_level:
+          conditionally_required: if and only if `messaging.destination.name` is known to have low cardinality. Otherwise, `messaging.destination.template` MAY be populated.
+      - ref: messaging.destination.template
+        requirement_level:
+          conditionally_required: if available.
+
+  # durations
+  - id: metric.messaging.client.request.duration
+    type: metric
+    metric_name: messaging.client.request.duration
```
trask, line 21: naming it "request" seems like a good choice over "operation" in order to help mitigate the confusion around whether this also includes "process" times (which have "messaging.operation.name" and so seem like they should be included in anything called "operation")

model/metrics/messaging-metrics.yaml
trask, line 73: maybe MUST -> SHOULD?

this could be tricky to implement / coordinate, e.g. when we instrument Spring messaging to capture "process" operations, and Spring uses Kafka polling under the hood which is also instrumented to capture "receive" operation

model/metrics/messaging-metrics.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - ref: messaging.operation.name
+        examples: ["process", "consume", "handle"]
+
+  # counters
+  - id: metric.messaging.client.published.messages
+    type: metric
+    metric_name: messaging.client.published.messages
+    brief: "Number of messages producer attempted to publish to the broker."
+    note: >
+      This metric MUST NOT count messages that were created, but aren't being published yet.
```
trask, line 56:
```suggestion
      This metric MUST NOT count messages that were created but haven't been published yet.
```

model/metrics/messaging-metrics.yaml
```diff
@@ -0,0 +1,80 @@
...
+        requirement_level:
+          conditionally_required: if available.
+
+  # durations
+  - id: metric.messaging.client.request.duration
+    type: metric
+    metric_name: messaging.client.request.duration
+    brief: "Duration of messaging operation initiated by a producer or consumer."
+    note: >
+      This metric SHOULD NOT be used to report processing duration - processing duration is reported in `messaging.process.duration` metric.
```
trask, line 24:
```suggestion
      This metric SHOULD NOT be used to report processing duration - processing duration is reported in `messaging.client.process.duration` metric.
```

lmolkova: we kept the same name for `messaging.process.duration` (due to concern over `client.process`). We can revisit the naming for it.

model/metrics/messaging-metrics.yaml
lmolkova: > maybe MUST -> SHOULD?

done 👍 

In the spring-kafka case, Spring integration should not report the number of consumed messages since kafka instrumentation does. But I agree it's could be impossible to know in some cases.

model/metrics/messaging-metrics.yaml
trask: oh, I don't know what I was thinking, I agree with `messaging.process.duration` 👍

trask: since `messaging.process.duration` is not a "client" metric, wdyt of renaming `messaging.client.request.duration` to `messaging.client.operation.duration` (it is all "client" operations, which doesn't include process operations)

lmolkova: I like it. @pyohannes any objections?

trask (approved):

PR #1010, opened 2024-05-07
Update markdownlint version and fix violation

lmolkova:


trisch-me (approved):

trask (approved):

.markdownlint.yaml
```diff
@@ -13,3 +13,6 @@ ol-prefix:
 no-inline-html: false
 fenced-code-language: false
 no-space-in-code: false
+first-line-h1: false
```
chalin, line 16: @lmolkova - why disable the check for page title / h1 heading? This is the source of the problem of #1971.

lmolkova: created https://github.com/open-telemetry/semantic-conventions/issues/1976 - `first-line-h1` literally checks if heading is the first line ignoring comments. There maybe some way to customize, but I didn't succeed on the first few attempts.

PR #1014, opened 2024-05-08
Prototype:  Use Weaver to enforce semantic convention policies

jsuereth:
This is a prototype of using Weaver to enforce semantic convention policies via the experimental Open-Policy-Agent support.

This defines three policies:

- `nonregistry_with_id_attr`: We only allow new attribute definitions inside the attribute registry, all other attributes must be `ref`.
- `registry_must_be_attribute_group`: All groups in attribute registry MUST be type `attribute_group`.
- `registry_with_ref_attr`: Attribute registry is not allowed to contain `ref` - An attribute can only belong to one group. 

Example output on today's repository:

```
$ make check-policies 
docker run --rm -v /home/joshuasuereth/src/open-telemetry/semantic-conventions/model:/source -v /home/joshuasuereth/src/open-telemetry/semantic-conventions/docs:/spec -v /home/joshuasuereth/src/open-telemetry/semantic-conventions/policies:/policies \
	otel/weaver:latest registry check \
	--registry=/source \
	--before-resolution-policies=/policies/before_resolution/registry.rego
✔ SemConv registry loaded (129 files)
✖ Policy violation: id=nonregistry_with_id_attr, category=attribute_registry, group=attributes.jvm.buffer, attr=pool.name, provenance: /source/metrics/jvm-metrics-experimental.yaml
✖ Policy violation: id=nonregistry_with_id_attr, category=attribute_registry, group=metric.process.context_switches, attr=process.context_switch_type, provenance: /source/metrics/process-metrics.yaml
✖ Policy violation: id=nonregistry_with_id_attr, category=attribute_registry, group=metric.process.paging.faults, attr=process.paging.fault_type, provenance: /source/metrics/process-metrics.yaml
✖ Policy violation: id=nonregistry_with_id_attr, category=attribute_registry, group=attributes.process.cpu, attr=state, provenance: /source/metrics/process-metrics.yaml
make: *** [Makefile:38: check-policies] Error 1
```
Each of these current violations are something we should fix.

## Changes

- Add new `check-policies` command to the makefile which can check weaver policies
- Add `policies/before_resolution/` folder for policies we can enforce on raw YAML model.

### DIFF
--- a/policies/before_resolution/registry.rego
+++ b/policies/before_resolution/registry.rego
@@ -30,7 +30,7 @@ attr_registry_violation(violation_id, group_id, attr_id) = violation {
 	violation := {
 		"id": violation_id,
 		"type": "semconv_attribute",
-		"category": "attrigute_registry",
+		"category": "attribute_registry",
 		"group": group_id,
 		"attr": attr_id,
 	}
--- a/policies/before_resolution/registry.rego
+++ b/policies/before_resolution/registry.rego
@@ -1,0 +1,37 @@
+package otel
+
+# A registry `attribute_group` containing at least one `ref` attribute is
+# considered invalid if it's not in the registry group.
+deny[attr_registry_violation("registry_with_ref_attr", group.id, attr.ref)] {
+	group := input.groups[_]
+	startswith(group.id, "registry.")
+	attr := group.attributes[_]
+	attr.ref != null
+}
+
+# We only allow attribute groups in the attribute registry.
+deny[attr_registry_violation("registry_must_be_attribute_group", group.id, "")] {
+	group := input.groups[_]
+	startswith(group.id, "registry.")
+	group.type != "attribute_group"
+}
+
+# Any group that is NOT in the attribute registry that has an attribute id is
+# in violation of not using the attribute registry.
+deny[attr_registry_violation("nonregistry_with_id_attr", group.id, attr.id)] {
+	group := input.groups[_]
+	not startswith(group.id, "registry.")
+	attr := group.attributes[_]
+	attr.id != null
+}
+
+# Build an attribute registry violation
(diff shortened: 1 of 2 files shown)


policies/before_resolution/registry.rego
```diff
@@ -0,0 +1,37 @@
+package otel
+
+# A registry `attribute_group` containing at least one `ref` attribute is
```
lmolkova, line 3: not blocking: would it be possible to link some docs on the policy syntax and the properties available on group/attribute/etc?

policies/before_resolution/registry.rego
```diff
@@ -0,0 +1,37 @@
...
+# considered invalid if it's not in the registry group.
+deny[attr_registry_violation("registry_with_ref_attr", group.id, attr.ref)] {
+	group := input.groups[_]
+	startswith(group.id, "registry.")
```
lmolkova, line 7: non-blocking, just some thoughts for the future:
 
I hope we will invent a more formal way to define a registry.
E.g. 
- now someone can add `registry.foo.bar` in any place in the repo. 
- or we'll add metrics registry and will need a different prefix.
- the attribute definition in the registry won't even have `ref` from the schema perspective
- etc

policies/before_resolution/registry.rego
lquerel: @lmolkova, you will find some documentation I wrote on this topic here https://github.com/open-telemetry/weaver/tree/main/crates/weaver_checker 

The definition of the attributes for the violation entity is defined here https://github.com/open-telemetry/weaver/blob/main/crates/weaver_checker/src/violation.rs

In general, the policy file is based on the Rego language and its documentation is available here https://www.openpolicyagent.org/docs/latest/policy-language/

PR #698, opened 2024-02-07
Messaging: clarify network attributes usage on common semconv 

lmolkova:
## Changes

Addresses messaging portion of #690

Removes `network.transport|type` attributes from general messaging semantic conventions.

Based on the findings from Java instrumentation libraries, the only messaging instrumentation that populates network attributes is RabbitMQ (which also sets them in native .NET instrumentation).  

This PR clarifies that `network.server.address|port` should be only set when they represent a specific broker instance (when individual instances are visible from the application side) and recommends to use them on the RabbitMQ (as one of the systems that's usually self-hosted)

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* ~~schema-next.yaml~~

PR #1016, opened 2024-05-08
Clarify that `*.system` captures the client perception

### LINKED ISSUE #813: `db|messaging|*.system` describes the client library but not necessarily the underlying system
Usually client library is specific to a certain DB/broker/etc, but it's also common to support non-native libs/backends. Examples:
- Azure Event Hubs broker [supports Kafka client libraries](https://learn.microsoft.com/en-us/azure/event-hubs/azure-event-hubs-kafka-overview)
- Azure CosmosDB [can be used with MongoDB libs](https://www.mongodb.com/docs/drivers/cosmosdb-support/)
- Several LLM solutions provide OpenAI-compatible REST APIs ([1](https://aws.amazon.com/marketplace/pp/prodview-wutjvjmubfkk6#:~:text=OpenAI%20API%20Compatibility%3A%20Our%20Mistral,AI%20experience%20intuitive%20and%20consistent.), [2](https://ollama.com/blog/openai-compatibility), [3](https://github.com/tensorchord/modelz-llm)) and are used via OpenAI SDKs

So `*.system` might not be the perfect word since the attribute value describes the client lib only.
Keeping `system` might also be good enough, but we need to clarify the attribute definition.

lmolkova:
Fixes #813 

## Changes

Clarifies that `db|messaing|gen_ai.system` captures the product name from the client's perspective which may not match the actual product used on the server side.


.chloggen/1016.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: db, messaging, gen_ai
```
trask, line 2: does this work? (asking for future me)

trask (approved):

.chloggen/1016.yaml
lmolkova: should be fine, based on @joaopgrassi experiment in https://github.com/open-telemetry/semantic-conventions/pull/698#discussion_r1496194623 it should render as

- `messaging, db, gen_ai`: <title> (#PR)

gregkalapos (approved):
Definitely a useful addition to the spec. Given these fields are `experimental` I think it's fine to merge as is.

Only thing I was thinking about is if there are maybe some higher level frameworks already instrumented that have a smarter logic to detect the `*.system`. If so, then for them this may be a breaking change.

I looked into [the EF Core instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs#L86-L141) that has some logic around populating `db.system`, but that ultimately ends up using [`ProviderName`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.infrastructure.databasefacade.providername?view=efcore-8.0), which based on the doc would be aligned with this PR:
> The name is typically the name of the provider assembly.

So that's ok.

Maybe someone knows about other similar cases - those would be nice to check.

lmolkova:
@gregkalapos 

thanks for checking EF!
what do you think could be breaking? There is no semantical change for the attribute, just the clarification that it's what client thinks the DB system is. Let me know if there is something I can do to clarify it further.

gregkalapos:
> @gregkalapos
> 
> thanks for checking EF! what do you think could be breaking? There is no semantical change for the attribute, just the clarification that it's what client thinks the DB system is. Let me know if there is something I can do to clarify it further.

I was thinking that [SemConv stability](https://opentelemetry.io/docs/specs/otel/versioning-and-stability/#semantic-conventions-stability) lists the following as a breaking change:
> The attribute values provided for the event that are defined in a list of well-known values.

Although [`db.system` has  `allow_custom_values: true`](https://github.com/open-telemetry/semantic-conventions/blob/main/model/registry/db.yaml#L63C10-L63C36), so I'm not sure this applies - but I was thinking more from a real user's experience. 

So if there is some higher level instrumentation that e.g. connects to MongoDb, but has some smart logic to set `db.system` to MongoDB, ConsomsDB, etc. based on what it connects to and let's assume it's a ConsosDB library, then for that instrumentation could be breaking - because prior to this PR it was ok to set `db.system` based on what it really connects to and now it must be `CosmosDb`. But this is completely hypothetical and the field is experimental. Just raising this so maybe someone has a real example. But very unlikely.

nirga (approved):

model/registry/messaging.yaml
```diff
@@ -137,8 +137,11 @@ groups:
           The system-specific name of the messaging operation.
         examples: [ "ack", "nack", "send" ]
       - id: system
-        brief: >
-          An identifier for the messaging system being used. See below for a list of well-known identifiers.
+        brief: Identifies the messaging product being used as recognized from the client side.
+        note: >
+          The actual messaging system may differ from the one known by the client.
+          For example, when using Kafka client library to communicate with Azure Event Hubs, the `messaging.system`
+          is set to `kafka` based on the instrumentation best knowledge.
```
pyohannes, line 144:
```suggestion
          The actual messaging system may differ from the one identified by the client.
          For example, when using Kafka client libraries to communicate with Azure Event Hubs, the `messaging.system`
          is set to `kafka` based on the instrumentation's best knowledge.
```

model/registry/messaging.yaml
```diff
@@ -137,8 +137,11 @@ groups:
           The system-specific name of the messaging operation.
         examples: [ "ack", "nack", "send" ]
       - id: system
-        brief: >
-          An identifier for the messaging system being used. See below for a list of well-known identifiers.
+        brief: Identifies the messaging product being used as recognized from the client side.
```
pyohannes, line 140:
```suggestion
        brief: The messaging system as identified by the client instrumentation.
```

pyohannes (approved):
This looks good to me, just some minor wording nits.

PR #1020, opened 2024-05-09
Extend GenAI system well-known values

TaoChenOSU:
Add more well-known GenAI model providers to the list of possible values that the `gen_ai.system` attribute can take. 

## Changes
The following are added:
- `microsoft` 
- `huggingface`
- `google`
- `meta`
- `anthropic`
- `mistral`
- `cohere` 

### DIFF
--- a/docs/attributes-registry/gen-ai.md
+++ b/docs/attributes-registry/gen-ai.md
@@ -21,7 +21,7 @@ This document defines the attributes used to describe telemetry in the context o
 | `gen_ai.response.finish_reasons` | string[] | Array of reasons the model stopped generating tokens, corresponding to each generation received. | `stop`                                                                  | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `gen_ai.response.id`             | string   | The unique identifier for the completion.                                                        | `chatcmpl-123`                                                          | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `gen_ai.response.model`          | string   | The name of the LLM a response was generated from.                                               | `gpt-4-0613`                                                            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `gen_ai.system`                  | string   | The name of the LLM foundation model vendor.                                                     | `openai`                                                                | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `gen_ai.system`                  | string   | The name of the LLM foundation model vendor.                                                     | `openai`; `microsoft`; `huggingface`                                    | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `gen_ai.usage.completion_tokens` | int      | The number of tokens used in the LLM response (completion).                                      | `180`                                                                   | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `gen_ai.usage.prompt_tokens`     | int      | The number of tokens used in the LLM prompt.                                                     | `100`                                                                   | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
@@ -30,6 +30,13 @@ This document defines the attributes used to describe telemetry in the context o
 
 `gen_ai.system` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
 
-| Value    | Description | Stability                                                        |
-| -------- | ----------- | ---------------------------------------------------------------- |
-| `openai` | OpenAI      | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| Value         | Description  | Stability                                                        |
+| ------------- | ------------ | ---------------------------------------------------------------- |
+| `openai`      | OpenAI       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `microsoft`   | Microsoft    | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `huggingface` | Hugging Face | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `google`      | Google       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `meta`        | Meta         | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `anthropic`   | Anthropic    | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `mistral`     | Mistral      | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `cohere`      | Cohere       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
--- a/model/registry/gen-ai.yaml
+++ b/model/registry/gen-ai.yaml
@@ -26,6 +26,14 @@ groups:
               stability: experimental
               value: "google"
               brief: 'Google'
+            - id: meta
+              stability: experimental
+              value: "meta"
+              brief: 'Meta'
+            - id: anthropic
+              stability: experimental
+              value: "anthropic"
+              brief: "Anthropic"
             - id: mistral
               stability: experimental
               value: "mistral"
--- a/model/registry/gen-ai.yaml
+++ b/model/registry/gen-ai.yaml
@@ -14,6 +14,26 @@ groups:
               stability: experimental
               value: "openai"
               brief: 'OpenAI'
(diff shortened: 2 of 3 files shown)


model/registry/gen-ai.yaml
```diff
@@ -14,6 +14,34 @@ groups:
...
+              brief: 'Microsoft'
+            - id: huggingface
+              stability: experimental
+              value: "huggingface"
+              brief: 'Hugging Face'
+            - id: google
+              stability: experimental
+              value: "google"
+              brief: 'Google'
+            - id: meta
```
nirga, line 29: The problem I see here is that Meta is not a provider. They don't have an SDK. You can host their models on replicate, on Azure, your own infra. How should the instrumentation infer that it should set this value to Meta? From the model name?

model/registry/gen-ai.yaml
```diff
@@ -14,6 +14,34 @@ groups:
               stability: experimental
               value: "openai"
               brief: 'OpenAI'
+            - id: microsoft
```
nirga, line 17: Are we using the company name or the service name? Because maybe this should be `azure`

model/registry/gen-ai.yaml
```diff
@@ -14,6 +14,34 @@ groups:
...
+              stability: experimental
+              value: "microsoft"
+              brief: 'Microsoft'
+            - id: huggingface
+              stability: experimental
+              value: "huggingface"
+              brief: 'Hugging Face'
+            - id: google
```
nirga, line 25: Same here - VertexAI?

docs/attributes-registry/gen-ai.md
```diff
@@ -30,6 +30,13 @@ This document defines the attributes used to describe telemetry in the context o
 
 `gen_ai.system` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
 
-| Value    | Description | Stability                                                        |
-| -------- | ----------- | ---------------------------------------------------------------- |
-| `openai` | OpenAI      | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| Value         | Description  | Stability                                                        |
+| ------------- | ------------ | ---------------------------------------------------------------- |
+| `openai`      | OpenAI       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `microsoft`   | Microsoft    | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 36: we should have a specific service here, but are there any MS-owned GenAI solutions? Where would someone set it?

docs/attributes-registry/gen-ai.md
```diff
@@ -30,6 +30,13 @@ This document defines the attributes used to describe telemetry in the context o
...
+| `huggingface` | Hugging Face | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `google`      | Google       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `meta`        | Meta         | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 39: We should have a specific produce name rather than company here. 
Is there a common name for Meta AI? Llama?

model/registry/gen-ai.yaml
TaoChenOSU: Are we defining model providers as the infrastructure providers or SDK providers?

I think users should use Meta as the model provider when they are using their Llama series models.

model/registry/gen-ai.yaml
TaoChenOSU: VertexAI is a machine learning platform for building and using Gen AI.

docs/attributes-registry/gen-ai.md
TaoChenOSU: Microsoft has its own series of models. I think we need a clear definition of `model provider`. 

Soon, there will be model as a service on Azure where users can run any models. Do we want all of those to be under Azure? In fact, you can already run Llama2 on Azure, do we want to categorize Llama2 under Azure?

TaoChenOSU:
I think HuggingFace will be the most controversial one. It's neither a model provider nor a service provider. You can use its inference endpoint but for production people may use their dedicated endpoints, which can be hosted on Azure, AWS or GCP.

docs/attributes-registry/gen-ai.md
```diff
@@ -30,6 +30,13 @@ This document defines the attributes used to describe telemetry in the context o
...
 
-| Value    | Description | Stability                                                        |
-| -------- | ----------- | ---------------------------------------------------------------- |
-| `openai` | OpenAI      | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| Value         | Description  | Stability                                                        |
+| ------------- | ------------ | ---------------------------------------------------------------- |
+| `openai`      | OpenAI       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `microsoft`   | Microsoft    | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `huggingface` | Hugging Face | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `google`      | Google       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
TaoChenOSU: Google also has other GenAI models as well, such as Gemma.

model/registry/gen-ai.yaml
joaopgrassi: maybe we already have a problem with this attribute, as if I understood it correctly there can be two distinct actors on this - the provider of the model, and the provider where things are running? What are we trying to define here of the two?

TaoChenOSU:
I removed some of the most controversial values. The remaining ones are known to have client SDKs.

docs/attributes-registry/gen-ai.md
lmolkova: we'd probably have a constant for `maas` since it's the API provides that serves requests and proxies things to the actual model. 
It's the GenAI product client application interacts with. 

The name here should be the name of the GenAI product, not a generic company name. 
I'd suggest to remove Microsoft for now unless MaaS is already being instrumented.

docs/attributes-registry/gen-ai.md
```diff
@@ -33,6 +33,10 @@ This document defines the attributes used to describe telemetry in the context o
 
 `gen_ai.system` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
 
-| Value    | Description | Stability                                                        |
-| -------- | ----------- | ---------------------------------------------------------------- |
-| `openai` | OpenAI      | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| Value          | Description  | Stability                                                        |
+| -------------- | ------------ | ---------------------------------------------------------------- |
+| `openai`       | OpenAI       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `azure_openai` | Azure OpenAI | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 39: Do we have to differentiate? Do we have a reliable way to differentiate? 
AFAIK not:

- we don't capture anything specific to azure
- client SDK does not care and the intention is to keep it this way
- domain name is not a reliable way to differentiate

nirga (approved):
@TaoChenOSU I've run the CI for you, there're some things you need to fix 😅

model/registry/gen-ai.yaml
```diff
@@ -14,6 +14,18 @@ groups:
...
+              value: "vertex_ai"
+              brief: 'Vertex AI'
+            - id: anthropic
+              stability: experimental
+              value: "anthropic"
+              brief: 'Anthropic'
+            - id: cohere
+              stability: experimental
+              value: "cohere"
+              brief: 'Cohere'
```
lmolkova, line 28:
```suggestion
              brief: 'Cohere'
```

lmolkova (approved):
thank you!

Could you please also add a changelog entry?
(just add a [yaml file in the expected format ](https://github.com/open-telemetry/semantic-conventions/blob/main/.chloggen/TEMPLATE.yaml) to `.chloggen` folder)

See https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#adding-a-changelog-entry

lmolkova (approved):

PR #1028, opened 2024-05-10
Add Azure SDK attributes & Logs events

### LINKED ISSUE #1027: New Azure Logs area
### Area(s)

area:new

As discussed in the following PR that contributes Azure logs transformations to the OpenTelemetry Collector Contrib, this issue proposes to add new semantic conventions for Azure Logs.
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32486

New Azure Logs section for semantic conventions.

_No response_

### Additional context

_No response_

MikeGoldsmith:
Fixes #1027, #1299

## Changes

Adds the new Azure Logs semantic conventions.

* [ ] ~[schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.~


model/registry/azure.yaml
```diff
@@ -0,0 +1,67 @@
+groups:
+  - id: registry.azure
+    prefix: azure
```
TylerHelmuth, line 3: The PR also defines `az.service_request_id`.

martinjt: It's defined here: 

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32486/files#diff-fbffe677c472f0c2f632302b809514af9861cd93c2490025a1be3d80b76a7e72R23

martinjt: Also, the prefix on that is `az` not `azure`

model/registry/azure.yaml
```diff
@@ -0,0 +1,100 @@
...
+      This document defines generic attributes for Azure SDK.
+    attributes:
+      - id: service_request_id
+        type: string
+        stability: experimental
+        brief: "The unique reference string that identifies a request"
+        examples:
+          - "00000000-0000-0000-0000-000000000000"
+  - id: registry.azure.logs
+    prefix: azure
```
JamieDanielson, line 15:
```suggestion
    prefix: az
```

MikeGoldsmith: There are two proposed azure prefixes, `az` is used by the Azure SDK and `azure` by the Azure logs.

model/registry/azure.yaml
```diff
@@ -0,0 +1,100 @@
...
+        brief: "The ID of the resource"
+        examples:
+          - "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg/providers/Microsoft.ApiManagement/service/apim"
+      - id: result.type
+        type: string
+        stability: experimental
+        brief: "The type of the result"
+        examples:
+          - "success"
+          - "failure"
```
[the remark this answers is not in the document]
MikeGoldsmith: It's not super important as the generated markdown files are the same. However, I'd prefer to be consistent in the same file and we use the same the list style for other fields that have multiple examples.

MikeGoldsmith:
Hi @lmolkova - please can we get your input on adding these Azure SDK / logs semantic conventions? Thanks 😄

lmolkova:
Thank you for for working on this! 

I'll share more comments, but the main question is whether we need to define those as attributes.
My assumption that azure logs would be mapped to OTel logs/log-based-events.

[Logs/events](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/events.md) have attributes and payload. We can take azure log as is, document the payload format and send it as it over OTLP with no mapping. 

Event payload fields can be documented, here's an [example](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/mobile/events.md#event-details), they don't need a namespace.

If there are no use-case for these attributes to appear on spans, metrics, resources, then event payload fields would work better.

So, was this option was considered? What would be the reasons to send those as attributes and not event payload fields?

/cc @MSNev

model/registry/azure.yaml
lmolkova: Unfortunately Azure SDK `az` ship has sailed, we shipped some stable libraries that emit attributes with this prefix and there are integrations that depend on it.

If Azure services would start to emit otel-compatible telemetry, I'd recommend them to use `az` as well, for the sake of consistency. 

So, unless there are strong reasons to use `azure` and not `az`,  we should be using the same prefix consistently for different signals.

model/registry/azure.yaml
```diff
@@ -0,0 +1,100 @@
...
+        brief: "The unique reference string that identifies a request"
+        examples:
+          - "00000000-0000-0000-0000-000000000000"
+  - id: registry.azure.logs
+    prefix: azure
+    type: attribute_group
+    brief: >
+      This document defines generic attributes for Azure Logs.
+    attributes:
+      - id: category
```
lmolkova, line 20: If we are modeling azure log as an event, the category is effectively an event name. 

If there are cases it's needed as an attribute, it'd suggest `az.log.category` to make it more precise.

TylerHelmuth:
@lmolkova the PR went with attributes because that has been the standard for collector translators/components so far. I believe the `k8sobjectsreceiver` is an early implementor of the event concept, but is currently suffering for being an early adopter ([it sets event.domain still which was removed](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/30fca096d6695fb5279c0a4a17614e73f56a4bc6/receiver/k8sobjectsreceiver/unstructured_to_logdata.go#L38)).

I'm not very up-to-date with the event spec. I see it is marked as experimental still, is it changing a lot still? I see this PR as an attempt to provide some clarity/stability for azure -> OTel semantics and I'm wondering if attributes gives more stability than events while azure -> OTel semantics become stable.

model/registry/azure.yaml
MikeGoldsmith: Okay, I thought we were already seeing telemetry with both `az` and `azure` prefixes. I don't see a strong need to have both, so will move everything under the `az` prefix for now. We can look to re-introduce the `azure` prefix later if needed.

model/registry/azure.yaml
MikeGoldsmith: This is fixed now we've addd the `az.log` prefix.

model/registry/azure.yaml
```diff
@@ -0,0 +1,100 @@
...
+        brief: "The correlation ID of the log entry"
+        examples:
+          - 607964b6-41a5-4e24-a5db-db7aab3b9b34
+      - id: duration
+        type: int
+        stability: experimental
+        brief: "The duration of the log entry in milliseconds"
+        examples:
+          - 1000
+      - id: identity
```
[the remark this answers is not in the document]
MikeGoldsmith: Updated the field description and added a note to the transform processor PR to make it opt-in:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32486#issuecomment-2137448044

trask:
> whether we need attributes for azure log events or they should be populated as event payload fields (elaborated in the comment above)

I think(?) this was recently answered in https://github.com/open-telemetry/opentelemetry-specification/pull/3969:

```
We do not define semantic conventions for LogRecords that are not Events.
```

> I'm not very up-to-date with the event spec. I see it is marked as experimental still, is it changing a lot still?

It changed a lot in the past 2 months, but I believe the big churn is over now

> I see this PR as an attempt to provide some clarity/stability for azure -> OTel semantics and I'm wondering if attributes gives more stability than events while azure -> OTel semantics become stable.

I believe the Event SIG and semantic convention SIG are interested in pushing forward with Events, and this seems like a great use case to ensure the Event Spec is meeting the mark

lmolkova:
> My understanding is that for any event.name value, the payload (log body in this case) should always be same. Do we have that guarantee with the data being targeted in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32486?

the event name identifies the event structure - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/events.md#general-event-semantics and not the content.

>the PR went with attributes because that has been the standard for collector translators/components so far. I believe the k8sobjectsreceiver is an early implementor of the event concept, but is currently suffering for being an early adopter ([it sets event.domain still which was removed](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/30fca096d6695fb5279c0a4a17614e73f56a4bc6/receiver/k8sobjectsreceiver/unstructured_to_logdata.go#L38)).

It made perfect sense before events were defined. 

Echoing Trask's comment https://github.com/open-telemetry/semantic-conventions/pull/1028#issuecomment-2161585093 we're now settling on events where they make sense. 
It could mean a simpler integration for existing logs in the collector
- little-to-no mapping for payload is required
- all we need to do is to define an event name and document the existing structure (maybe even as a link to external documentation if it's versioned)

lmolkova:
@MikeGoldsmith 

I'm happy to chat about the attributes, but I suggest to settle on the event discussion. The events are the same as logs and the way to define semconv for them is to:

- define event name
- define and reference attributes
- define payload fields

E.g. I can have an event "foo.bar" with payload `{"key":"value", "anotherKey":42}` (it's not a string, but an object), `key` and `anotherKey` are payload fields. They don't need to be defined as attributes. They don't need to be namespaced, there is no expectation for them to be reused across different events. 
Attributes are expected to be reused. E.g. Azure resource id is expected to be captured under `cloud.resource_id` on spans, metrics, or events if it's captured as an attribute.

So, the option for Azure logs is to not do any mapping and take existing Azure log and put all the stuff into payload properties and just document it. 
If we believe that something needs to be an attribute and not a payload field, we can probably promote them to attributes. I'd assume `cloud.resource_id` to be a reasonable candidate for it since people would probably query different telemetry using it.

MikeGoldsmith:
I plan on working on moving log attributes to an event structure this week.

model/registry/azure.yaml
```diff
@@ -0,0 +1,100 @@
...
+      - id: properties
+        type: string
+        stability: experimental
+        brief: "The properties of the log entry"
+        examples:
+          - "{'string': 'value'}"
+          - "{'int': 1234 }"
+          - "{'float': '1.234'}"
+          - "{'bool': true}"
+      - id: resource.id
```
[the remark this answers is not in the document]
MikeGoldsmith: Updated to use ref to recommended field.

MikeGoldsmith:
@lmolkova I think this is ready for another review.

I've made the following changes:
- update type to event
- update attributes to use known references instead of creating new ones
- moved the other attributes to a commented out body.fields section like I saw [here](https://github.com/open-telemetry/semantic-conventions/blob/main/model/logs/mobile-events.yaml).

Thanks 😄

MikeGoldsmith:
Also, as event body fields are not stored in the spec - where do you think we should store them?

model/registry/azure.yaml
```diff
@@ -0,0 +1,115 @@
...
+    #     - id: operation.version
+    #       type: string
+    #       stability: experimental
+    #       brief: "The version of the operation"
+    #       examples:
+    #         - "1.0"
+    #     - id: properties
+    #       type: template[string[]]
+    #       stability: experimental
+    #       brief: "The properties of the log entry"
```
lmolkova, line 74: The properties provided in the Azure Resource Log, `<key>` being the normalized property name (lowercase), the value being the property value.

model/registry/azure.yaml
```diff
@@ -0,0 +1,115 @@
...
+    #       examples:
+    #         - "{'string': 'value'}"
```
lmolkova, line 76: please use example format as in https://github.com/open-telemetry/semantic-conventions/blob/d7115c8aecbfaa6ed2f5d57def9761bca3364af2/model/registry/http.yaml#L138

model/registry/azure.yaml
```diff
@@ -0,0 +1,115 @@
...
+    #         - "{'int': 1234 }"
+    #         - "{'float': '1.234'}"
+    #         - "{'bool': true}"
+    #     - id: resource.id
```
lmolkova, line 80: do we want do have it as a payload field and as an attribute?

model/registry/azure.yaml
```diff
@@ -0,0 +1,115 @@
...
+    attributes:
+      - id: service_request_id
+        type: string
+        stability: experimental
+        brief: "The unique reference string that identifies a request"
+        examples:
+          - "00000000-0000-0000-0000-000000000000"
+  - id: az.log
+    stability: experimental
+    type: event
```
lmolkova, line 16: we should not define events in the registry (it's only for attributes for the time being), please move `az.log` definition to another file in `/model/logs`

lmolkova (commented):
great progress, thank you! I left some minor comments, the major one is how to document it in markdown

I believe we need to:
- add a new md file for event definition, see https://github.com/open-telemetry/semantic-conventions/blob/main/docs/mobile/events.md as an example
- we'd need to document event payload in markdown - no need to preserve perfect look-and-feel.  E.g. you could add a TODO note with link to yaml file for reference to avoid duplication.
- I think the log processor should be prepared to receive other properties not listed in the yaml. E.g. resource-specific properties such as [partitionId](https://learn.microsoft.com/en-us/azure/event-hubs/monitor-event-hubs-reference#resource-logs) for Event Hubs. We should document how they will handled (dropped, populated but name normalized, if normalized then how)

model/registry/azure.yaml
MikeGoldsmith: I think this was meant to be a suggestion for the description. Let me know if you meant something else.

model/registry/azure.yaml
```diff
@@ -0,0 +1,115 @@
...
+    #         - SecretGet
+    #         - Microsoft.ApiManagement/GatewayLogs
+    #         - ApplicationGatewayAccess
+    #     - id: operation.version
+    #       type: string
+    #       stability: experimental
+    #       brief: "The version of the operation"
+    #       examples:
+    #         - "1.0"
+    #     - id: properties
```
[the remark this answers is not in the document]
MikeGoldsmith: I did orinally take the, then when reviewing the [example log](https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/activity-log-schema#sample-event) schema again, the field is called `properties`. Would still prefer it to be `property` instead?

MikeGoldsmith:
@lmolkova I think this is ready for (hopefully) final review 😢

model/logs/azure.yaml
```diff
@@ -0,0 +1,98 @@
...
+    name: az.log
+    brief: >
+      Describes Azure Resource Log event, see
+      [Azure Resource Log Top-level Schema](https://learn.microsoft.com/azure/azure-monitor/essentials/resource-logs-schema#top-level-common-schema)
+      for more details.
+    attributes:
+      - ref: event.name
+      - ref: az.service_request_id
+      - ref: cloud.resource_id
+    # Future Note: When the build tools support this definition please uncomment and validate the details
```
lmolkova, line 14: nit: could you please indent the comment?

docs/attributes-registry/azure.md
```diff
@@ -0,0 +1,15 @@
...
+
+# Azure
+
+## Azure Sdk Attributes
+
+This document defines generic attributes for Azure SDK.
+
+| Attribute               | Type   | Description                                           | Examples                               | Stability                                                        |
+| ----------------------- | ------ | ----------------------------------------------------- | -------------------------------------- | ---------------------------------------------------------------- |
+| `az.service_request_id` | string | The unique reference string that identifies a request | `00000000-0000-0000-0000-000000000000` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
lmolkova: the `service_request_id` is the service-side identifier that comes from response headers (there is also a client-side identifier generated on the client-side). It's an important distinction and I wonder if we can explain it here with something like 

```suggestion
| `az.service_request_id` | string | The unique reference string that identifies a request. It's generated by the Azure service and returned with the response. | `00000000-0000-0000-0000-000000000000` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

model/logs/azure.yaml
```diff
@@ -0,0 +1,98 @@
+groups:
+  - id: az.resource.log
+    stability: experimental
+    type: event
+    name: az.log
```
lmolkova, line 5:
```suggestion
    name: az.resource.log
```

model/logs/azure.yaml
lmolkova, line 1: could you please add an md file for this event? 

e.g. with the content similar to:

# Semantic Conventions for Azure Resource Logs

Status: Experimental

This document describes Azure Resource Log event, see [Azure Resource Log Top-level Schema](https://learn.microsoft.com/azure/azure-monitor/essentials/resource-logs-schema#top-level-common-schema).

## Event details

The event name MUST be `az.resource.log`

### Attributes 


this will be auto-generated 


### Body Fields

TODO: auto-generate event body description. 
See [Azure Resource Log definition](/model/logs/azure.yaml) for the details.

model/logs/azure.yaml
```diff
@@ -0,0 +1,98 @@
...
+    brief: >
+      Describes Azure Resource Log event, see
+      [Azure Resource Log Top-level Schema](https://learn.microsoft.com/azure/azure-monitor/essentials/resource-logs-schema#top-level-common-schema)
+      for more details.
+    attributes:
+      - ref: event.name
+      - ref: az.service_request_id
+      - ref: cloud.resource_id
+    # Future Note: When the build tools support this definition please uncomment and validate the details
+    # included here and what has been added to the manual markdown table
```
trask, line 15: I didn't find this manual markdown table?

MikeGoldsmith: Added markdown files in https://github.com/open-telemetry/semantic-conventions/pull/1028/commits/7a9d4fe591d01051dfa34873cca6182eecf68044 which addes the markdown tables 👍🏻

model/logs/azure.yaml
MikeGoldsmith: Added in https://github.com/open-telemetry/semantic-conventions/pull/1028/commits/7a9d4fe591d01051dfa34873cca6182eecf68044

model/logs/azure.yaml
MikeGoldsmith: The comment is for the body tag below so I added the comment to the same level. I think if I indented again it'll look like a comment to the above `ref:` entry. Is that what you mean?

docs/azure/events.md
```diff
@@ -0,0 +1,44 @@
...
+| [`cloud.resource_id`](/docs/attributes-registry/cloud.md) | string | Cloud provider-specific native identifier of the monitored cloud resource (e.g. an [ARN](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html) on AWS, a [fully qualified resource ID](https://learn.microsoft.com/rest/api/resources/resources/get-by-id) on Azure, a [full resource name](https://cloud.google.com/apis/design/resource_names#full_resource_name) on GCP) [4] | `arn:aws:lambda:REGION:ACCOUNT_ID:function:my-function`; `//run.googleapis.com/projects/PROJECT_ID/locations/LOCATION_ID/services/SERVICE_ID`; `/subscriptions/<SUBSCIPTION_GUID>/resourceGroups/<RG>/providers/Microsoft.Web/sites/<FUNCAPP>/functions/<FUNC>` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- end of manually added table -->
+
+### Body Fields
+
+<!-- manually added table until body fields can be autogenerated -->
```
lmolkova, line 26: 👍

docs/azure/events.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Azure Resource Log
+
+### Event details
+
+The event name MUST be `az.resource.log` in the `event.name` property.
+
+### Attributes
+
+<!-- manually added table until body fields can be autogenerated -->
```
lmolkova, line 16: these are attributes, you can generate them automatically (for `az.resource.log` group)

lmolkova (approved):
Left a couple of cosmetic comments, looks great otherwise!

model/logs/azure.yaml
lmolkova: I mean GH warning above, I think GH yaml linter wants the comment to be indented on the same level as `ref`.  

I.e.
```yaml
- ref: cloud.resource_id
  # Future Note: When the build tools support this definition please uncomment and validate the details
```

model/registry/azure.yaml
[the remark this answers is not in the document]
lmolkova: sorry, missed this.

If you want to preserve `properties`, the resulting event would look like (if it was serialized to json) 

```json
{
... 
"properties" : {
  "foo": "bar",
  "baz": 42
  }
}
```

To achieve it, you probably want to say that `properties` type is [`KeyValueList`](https://github.com/open-telemetry/opentelemetry-proto/blob/484241a016d7b81f20b6e19d00ffbc4a3b864a22/opentelemetry/proto/common/v1/common.proto#L37C5-L37C18) - it'll probably need to change once tooling support events, but the format of the event should not change.

the current type used for `properties` is `template[string[]]`, which implies different structure. It's be 

```json
{
... 
"property.foo" : "bar",
"property.baz" : 42
}
```

So depending on the outcome you want, you'd pick the name (`properties` vs `property`) and type (dictionary vs flattened individual properties).

docs/azure/events.md
MikeGoldsmith: I thought the attributes table would have been autogenerated using the make targets `table-generation` or `attribute-registry-generation` but it didn't work so created manually.

Is there another make target that I should use instead?

model/registry/azure.yaml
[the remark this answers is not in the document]
MikeGoldsmith: Updated to use keyvaluelist in https://github.com/open-telemetry/semantic-conventions/pull/1028/commits/3b1dd09c94d9a4b364d50d5c2318bcfe14b773d1

MikeGoldsmith:
Updated properties to be a keyvaluelist field type.

docs/azure/events.md
lmolkova: outside if attribute registry, markdown files are mostly written manually, only tables between 
`<!-- semconv  az.resource.log -->` and `<!-- endsemconv -->` are auto-generated (with `table-generation`) target.

Do you just need to add the comments above and run the target.

MikeGoldsmith: Nice, thanks you for the help. I've updated to autogenerate the attributes in [0c98f83](https://github.com/open-telemetry/semantic-conventions/pull/1028/commits/0c98f834a9146bfc7ba60795d8a143b1f9b1c3ac)

MikeGoldsmith:
I've addressed feedback - @lmolkova @trask. Let me know if there's anything else you'd like to see.

model/logs/azure.yaml
```diff
@@ -0,0 +1,100 @@
...
+    type: event
+    name: az.resource.log
+    brief: >
+      Describes Azure Resource Log event, see
+      [Azure Resource Log Top-level Schema](https://learn.microsoft.com/azure/azure-monitor/essentials/resource-logs-schema#top-level-common-schema)
+      for more details.
+    attributes:
+      - ref: event.name
+      - ref: az.service_request_id
+      - ref: cloud.resource_id
```
lmolkova, line 13: maybe:

```suggestion
      - ref: cloud.resource_id
         brief:  The [Fully Qualified Azure Resource ID](https://docs.microsoft.com/rest/api/resources/resources/get-by-id) the log is emitted for.
         note: ""
```

lmolkova (approved):
Thank you!

trask (approved):

Issue #940, opened 2024-04-17
Issues with Hardware Metrics semantic conventions


eero-t:
**What are you trying to achieve?**

Follow recommendations on: https://opentelemetry.io/docs/specs/semconv/system/hardware-metrics/

**What did you expect to see?**

Consistent and implementable specification.

**Additional context.**

Ran across following issues when trying to map GPU metrics to the semantics...

Non-implementable / unspecified items:
* Single `firmware_version` won't work:
  * Devices have multiple types of firmware (display, media, scheduling/power management etc)
* `hw.errors` have just `hw.error.type` attribute, although:
  * Errors can be correctable, uncorrectable (data lost), or fatal (functionality lost, at least reset needed to recover)
  * Errors can originate from different parts of the SW stack (FW, kernel, userspace driver)
  * Errors can originate from different parts of the device HW (display, media, compute, 3d etc)
  * => I suggest adding `.category` attribute[1], similarly to Level-Zero spec:
    * https://spec.oneapi.io/level-zero/latest/sysman/api.html#zes-ras-error-cat-t

Inconsistencies:
* `hw.gpu.power` vs. `hw.power{hw.type="gpu"}` confusion
  *  Unless both are valid, why there's no no `hw.gpu.energy` to match `hw.energy{hw.type="gpu"}`?
* Common HW `name` & `id` attributes vs. GPU `model` & `serial` attributes
  * Are all of these supposed to be used despite overlap?
* Why `vendor` attribute is used for GPU devices, but `manufacturer` for (host) device: https://opentelemetry.io/docs/specs/semconv/resource/device/
* Inconsistent attribute examples for missing GPU metrics:
  * `system.cpu.frequency` vs. `hw.cpu.speed`
  * `.utilization` vs. `.*_ratio` [2] for things like (frequency) throttling
  * whether in addition to base metric one should provide `.utilization` / `.*_ratio`, or `.limit` value?

----

[1] Summing errors is not very meaningful (rate is more interesting), so maybe `all` category could be used also for indicating whether there are any errors (within query period), not just when more fine-categories are not provided?
[2] Used e.g. in: https://opentelemetry.io/docs/specs/semconv/system/hardware-metrics/#hwfan---fan-metrics

joaopgrassi:
FYI @bertysentry

bertysentry:
@eero-t Thank you for the feedback, give me a few hours to answer your questions. We will address all issues!

bertysentry:
[...]

PR #1030, opened 2024-05-10
move common hardware attributes and metrics to registry

trisch-me:
This is the first PR out of many for hardware metrics.

The first PR is about generic attributes used by all hardware metrics. I decided to split it into multiple PRs to make review easier because it's easy to miss something with a lot of changed files (I have started doing it with 5 metrics and lost myself track on things :) ). 

I delete portions of `docs/system/hardware-metrics.md` file so that reviewers could see what has been changed now. This file will be deleted later as soon as it will be ported completely.

Hardware wasn't moved to the yaml schema, so I'm doing here both introducing it to the registry and creating appropriate yaml files for each hardware metric

I need this PR merged before I will create other PRs (half of them is ready but based on this PR)


model/metrics/hardware/common.yaml
```diff
@@ -0,0 +1,59 @@
...
+    extends: hardware.attributes.common
+    attributes:
+      - ref: hw.type
+        requirement_level: required
+
+  - id: metric.hardware.energy
+    type: metric
+    metric_name: hw.energy
+    stability: experimental
+    brief: "Energy consumed by the component, in joules"
```
lmolkova, line 15:
```suggestion
    brief: "Energy consumed by the component"
```

model/metrics/hardware/common.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - id: metric.hardware.errors
+    type: metric
+    metric_name: hw.errors
+    stability: experimental
+    brief: "Number of errors encountered by the component"
+    instrument: counter
+    unit: "{errors}"
+    extends: metric.hardware.attributes
+    attributes:
+      - ref: hw.error.type
```
lmolkova, line 29:
```suggestion
      - ref: error.type
```

model/metrics/hardware/common.yaml
```diff
@@ -0,0 +1,59 @@
...
+        requirement_level: recommended
+
+  - id: metric.hardware.power
+    type: metric
+    metric_name: hw.power
+    stability: experimental
+    brief: "Instantaneous power consumed by the component, in Watts"
```
lmolkova, line 36:
```suggestion
    brief: "Instantaneous power consumed by the component"
```

model/metrics/hardware/common.yaml
```diff
@@ -0,0 +1,59 @@
...
+    type: attribute_group
+    brief: "Attributes for hardware metrics"
+    extends: hardware.attributes.common
+    attributes:
+      - ref: hw.type
+        requirement_level: required
+
+  - id: metric.hardware.energy
+    type: metric
+    metric_name: hw.energy
```
lmolkova, line 13:
```suggestion
    metric_name: hw.energy.consumption
```
?

model/metrics/hardware/common.yaml
lmolkova, line 36: is it derivable from `hw.energy` ? why is this one necessary?

model/registry/hardware.yaml
```diff
@@ -0,0 +1,113 @@
...
+        brief: >
+          An identifier for the hardware component, unique within the monitored host
+        examples: ['win32battery_battery_testsysa33_1']
+      - id: name
+        type: string
+        stability: experimental
+        brief: >
+          An easily-recognizable name for the hardware component
+        examples: ['eth0']
+      - id: parent
```
lmolkova, line 20:
```suggestion
      - id: parent.id
```

model/registry/hardware.yaml
```diff
@@ -0,0 +1,113 @@
...
+        type: string
+        stability: experimental
+        brief: >
+          Unique identifier of the parent component (typically the `id` attribute of the enclosure, or disk controller)
+        examples: ['dellStorage_perc_0']
+      - id: type
+        type:
+          allow_custom_values: true
+          members:
+            - id: battery.yaml
```
lmolkova, line 30:
```suggestion
            - id: battery
```

model/registry/hardware.yaml
```diff
@@ -0,0 +1,113 @@
...
+              brief: 'Temperature'
+              stability: experimental
+            - id: voltage
+              value: 'voltage'
+              brief: 'Voltage'
+              stability: experimental
+        stability: experimental
+        brief: >
+          Type of the component
+      - id: error.type
```
lmolkova, line 89: there is no need to declare `hw.error.type`, we should reuse the existing one

model/registry/hardware.yaml
```diff
@@ -0,0 +1,113 @@
...
+              stability: experimental
+            - id: power_supply
+              value: 'power_supply'
+              brief: 'Power supply'
+              stability: experimental
+            - id: tape_drive
+              value: 'tape_drive'
+              brief: 'Tape drive'
+              stability: experimental
+            - id: temperature
```
lmolkova, line 78: hw.type = temperature? What it be used for? cooling system? temperature sensor?

model/registry/hardware.yaml
```diff
@@ -0,0 +1,113 @@
+groups:
+  - id: registry.hardware
+    prefix: hw
+    type: attribute_group
+    brief: >
+      Attributes for hardware.
```
lmolkova, line 6: It sounds the goal is to describe hardware component rather than a generic hardware, so I'd prefer to have `hw.component` or `hardware.component` as a namespace

model/registry/hardware.yaml
```diff
@@ -0,0 +1,113 @@
...
+              stability: experimental
+            - id: tape_drive
+              value: 'tape_drive'
+              brief: 'Tape drive'
+              stability: experimental
+            - id: temperature
+              value: 'temperature'
+              brief: 'Temperature'
+              stability: experimental
+            - id: voltage
```
lmolkova, line 82: same here  - `voltage` seems confusing as a type

docs/system/hardware-metrics.md
```diff
@@ -12,9 +12,7 @@ when creating instruments not explicitly defined in the specification.
 
 <!-- toc -->
 
-- [Common hardware attributes](#common-hardware-attributes)
```
lmolkova, line 15: it should still be relevant for other metrics defined in this document

docs/system/hardware-metrics.md
```diff
@@ -44,43 +42,8 @@ when creating instruments not explicitly defined in the specification.
...
-
-| Attribute Key | Description                                                                                                   | Example                             | Requirement Level |
-| ------------- | ------------------------------------------------------------------------------------------------------------- | ----------------------------------- | ----------------- |
-| `id`          | An identifier for the hardware component, unique within the monitored host                                    | `win32battery_battery_testsysa33_1` | **Required**      |
-| `name`        | An easily-recognizable name for the hardware component                                                        | `eth0`                              | Recommended       |
-| `parent`      | Unique identifier of the parent component (typically the `id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0`                | Recommended       |
-
 ## Metric Instruments
 
-### `hw.` - Common hardware metrics
```
lmolkova, line 62: Can we keep HW metrics in the same file? 
If we're splitting them in 2, we should at least have cross-references and some explanation why they are different

/cc @open-telemetry/semconv-system-approvers

docs/hardware/common.md
```diff
@@ -0,0 +1,223 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Hardware Common Metrics
+--->
+
+# Semantic Conventions for Hardware Common Metrics
+
+**Status**: [Experimental][DocumentStatus]
+
+All metrics in `hw.` instruments should be attached to a [Host Resource](/docs/resource/host.md)
```
lmolkova, line 9: Is is a responsibility of instrumentation author or the end user?

docs/hardware/common.md
```diff
@@ -0,0 +1,223 @@
...
+| [`hw.parent`](/docs/attributes-registry/hw.md) | string | Unique identifier of the parent component (typically the `id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+<!-- endsemconv -->
+
+## Common hardware metrics
+
+The below metrics apply to any type of hardware component.
+
+These common `hw.` metrics include the below attributes to describe the
+monitored component:
+<!-- semconv metric.hardware.attributes(full) -->
```
lmolkova, line 28: we have another attribute table defined above, which of them should be applied to metrics?

lmolkova (commented):
Please also add a changelog entry.

trask:
cc @bertysentry

trisch-me:
@lmolkova I'm confused by your comments. I haven't created anything new. I have extracted hardware from one md file into yaml. As I stated in description I will do this in batches because there will be open questions and to make review easier, so I have extracted only first portion

docs/hardware/common.md
trisch-me: This is refactoring (copy-paste) from original file.

docs/hardware/common.md
trisch-me: In original md file here was defined only `type` 
common attributes extend the table above therefore when you render it it has both previous attributes and `type`. Do you know how to skip it or should we skip it?

docs/system/hardware-metrics.md
trisch-me: This file will be removed. I have deleted this portion only for visibility on what is moved out

docs/system/hardware-metrics.md
trisch-me: Sorry, I don't understand your question? In which file?

model/metrics/hardware/common.yaml
trisch-me: I don't change in this PR anything, I do only refactoring (in cases where it's possible), if we would like to change names and introduce a breaking change - this should be done later for simplicity

model/registry/hardware.yaml
trisch-me: this error type has specific enum value (will be updated in the future PRs). I can't redefine enum values from existing one in current build system.

model/registry/hardware.yaml
trisch-me: it is defined as temperature sensor in the hardware-metrics. The same goes for voltage

model/registry/hardware.yaml
trisch-me: This will introduce a breaking change but probably we can't go without one as `parent` is very common

model/registry/hardware.yaml
trisch-me: Can we change the whole namespace? It will go from current `hw.*` to `hw.XX.*` which means breaking changes for **all** fields

model/metrics/hardware/common.yaml
trisch-me: I have answered in another similar comment

model/metrics/hardware/common.yaml
bertysentry: @lmolkova This can be discussed in a separate issue. The `.consumption` suffix is not necessary. Also, this counter may also cover energy production (batteries, for example), in which case we will use an attribute to indicate the *direction* (`direction="out|in"`).

model/metrics/hardware/common.yaml
bertysentry: You're right! It has been decided to specify both `hw.power` and `hw.energy` because:

* `hw.energy` (counter in Joules) is preferred because it provides more accurate aggregation (just like network traffic or CPU time utilization is better calculated with cumulative counters instead of averaging rates)
* Unfortunately, most systems are only capable of reporting their instantaneous electrical power consumption in Watts. To make sure existing systems all use the same metric name for electrical power, we specified `hw.power` in the semantic conventions.

Again: It can be discussed in a separate issue.

model/registry/hardware.yaml
bertysentry: Personal opinion: Let's not put namespaces everywhere. Will there ever be `parent.name`, or `parent.parent`?

`parent` is indeed very generic. If it belongs to the hardware semantic conventions only, it could be later changed to `hw.parent`. 

However, I think this attribute could be beneficial to other semconv, like process, system disks and filesystems, databases, etc. Maybe it could be specified as a general attribute, applicable everywhere.

This will be discussed in a separate issue.

model/registry/hardware.yaml
bertysentry: We need extensible enums, because the semantic conventions cannot cover all possible values of error types.

model/registry/hardware.yaml
bertysentry: I strongly disagree with this. In hardware, we're specifying metrics that apply to different levels: `hw.energy` may apply to a processor, a server, a UPS or even a data center. Same for temperature. Hardware does include components, but it's not just that, and it doesn't need to be namespaced as there are no collisions here.

docs/hardware/common.md
bertysentry: Responsibility is on the instrumentation developer.

bertysentry:
@trisch-me Thank you so much for taking the time to move the hardware semconv to YAML! (I was supposed to do it but couldn't find the time... and I was afraid the tooling wouldn't work well on some of the specificities of the hardware metrics, with common attributes, etc.).

Let me know if you need any help with that!

If you want to see these metrics in real life (to get a better understanding of their purpose and use), you can check these Grafana dashboards:

* [https://hws-demo.sentrysoftware.com/d/SHFBpSH7z/hardware-sentry---site?var-hw_type=All&var-site=Sentry-Ottawa&var-host_name=idrac-r640-01.internal.sentrysoftware.net&var-agent=demo-hws-02.internal.sentrysoftware.net&orgId=1&refresh=30s](Aggregated view of several systems)
* [https://hws-demo.sentrysoftware.com/d/c2K8skm4k/hardware-sentry---host?orgId=1&var-site=Sentry-Paris&var-host_name=idrac-r640-01.internal.sentrysoftware.net&var-hw_type=All&var-agent=demo-hws-02.internal.sentrysoftware.net](Detailed view of 1 system)

Thanks!

model/registry/hardware.yaml
trisch-me: @bertysentry there is a existing `process.parent` fields from ECS. But I suspect parent could be used also somewhere else. So I am also for introducing a separate namespace for it. But I don't want to do this in this PR. I want this PR to be as plain as possible - only move things out to the yaml files and the registry without functional changes

model/registry/hardware.yaml
trisch-me: @bertysentry we have discussed extensible enums already, maybe they will be there in the near future 
cc @jsuereth @lquerel 
But as for now I will just redefine enum with all possible options as soon as I will move them over

bertysentry:
@trisch-me I confirm we met in Paris! It's funny when you realize usernames in GitHub are actual people! 😅 

I'll be happy to review any PR on this topic!

model/registry/hardware.yaml
bertysentry: Alright, thank you!

model/registry/hardware.yaml
bertysentry: Understood, thank you @trisch-me!

bertysentry (approved):

trisch-me:
blocked by https://github.com/open-telemetry/semantic-conventions/pull/1049

docs/system/hardware-metrics.md
lmolkova: In the meantime, while it is not removed, could you please add a link to the section in the new file so that someone looking at this file will know that there are some common attributes?

model/metrics/hardware/common.yaml
lmolkova: Since there was no defined attribute for `hw.error.type`, my strong preference would be not to introduce one and instead use generic `error.type`.

Otherwise, we'd end up with another attribute the should be immediately deprecated.

model/metrics/hardware/common.yaml
```diff
@@ -0,0 +1,59 @@
...
+    unit: "J"
+    extends: metric.hardware.attributes
+
+  - id: metric.hardware.errors
+    type: metric
+    metric_name: hw.errors
+    stability: experimental
+    brief: "Number of errors encountered by the component"
+    instrument: counter
+    unit: "{errors}"
```
lmolkova, line 26:
```suggestion
    unit: "{error}"
```

model/registry/hardware.yaml
lmolkova: so the suggestion is to use `voltage_sensor`

model/registry/hardware.yaml
lmolkova: error.type is intended to be common attribute describing an error, individual semconv can define additional domain specifc attributes, or specify what `error.type` means in their context. 

Let's reuse the generic one as the first step.

docs/hardware/common.md
lmolkova: Since each metric in this file defines extra attributes in addition to common ones, I suggest removing common attributes and just generating full attribute tables for each metric.

Otherwise it becomes hard to understand from the reader perspective which attribute apply to a specific metric without scrolling back and forth

docs/system/hardware-metrics.md
lmolkova: Currently hardware metrics are spread across multiple files. I understand the intention to move all of them, but we should keep things consistent in the mean time.

Since you're removing some metrics from this file, please at least add a link to the new doc from this one.

model/registry/hardware.yaml
lmolkova: well, the `hw.id` seems to be better called `hw.component.id`.  While some things can stay in the `hw` root namespace, some might need more love to become descriptive

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -44,6 +44,7 @@ body:
         - area:gcp
         - area:gen-ai
         - area:graphql
+        - area:hardware
```
lmolkova, line 47: could we be consistent with the root namespace?
```suggestion
        - area:hw
```

(and in other places)

model/metrics/hardware/common.yaml
lmolkova: created https://github.com/open-telemetry/semantic-conventions/issues/1071 to follow up

lmolkova:
> @lmolkova I'm confused by your comments. I haven't created anything new. I have extracted hardware from one md file into yaml. As I stated in description I will do this in batches because there will be open questions and to make review easier, so I have extracted only first portion

I created #1071 to keep track of HW naming issues. I still suggest to do a bare minimum changes to existing HW metrics so that we don't need to deprecated a bunch of things later (`hw.error.type`, and `hw.type` values).

model/registry/hardware.yaml
bertysentry: @lmolkova It's a matter of opinion: `temperature` is not confusing to me. 

Appending `_sensor` doesn't provide additional meaning, as most metrics in hardware are reported by "sensors" anyway: power status, memory presence, temperature, fan speed, etc. Why add `_sensor` for temperatures and voltages, but not for fan speed or power supplies? Let's keep this simple and consistent.

model/registry/hardware.yaml
bertysentry: @lmolkova See my previous comment for temperature.

model/registry/hardware.yaml
bertysentry: @lmolkova The intent of this PR is just to migrate from Markdown to YAML. Semantic changes must be brought in a separate PR.

docs/hardware/common.md
bertysentry: @lmolkova I'm sorry, but there must be a misunderstanding here. Let's say I'm HPE developing an OpenTelemetry instrumentation for my ProLiant servers, I will use an Otel SDK to push metrics that I captured through specific hardware busses, and while developing this software layer, I will need to create a Resource to attach my metrics to, so that both Resources and Metrics are sent over OTLP.

It's simply a matter of making sure all metrics have the proper attributes to attach them to a host.

Without these, it's practically impossible to use these metrics in real life.

docs/hardware/common.md
bertysentry: I agree with @lmolkova's suggestion here.

model/metrics/hardware/common.yaml
bertysentry: We used to use plural for non-standard units (in curly brackets). Has this changed?

model/registry/hardware.yaml
lmolkova: `hw.type = fan`  makes sense, fan is a hardware component, same with `power_supply`.
`temperature` is not a component.

bertysentry: Okay @lmolkova now I understand why it sounds confusing. Actually, `hw.type` is not to be read as "hardware type", but as the "type" attribute used by various metrics (notably `hw.status`), which happens to be in the `hw.` namespace.

`hw.status` can be used for disks, network cards, memory modules, etc. and temperatures, voltages, etc. `hw.status{hw.type="temperature"}` will report whether temperature is okay or not, and not whether the temperature sensor is operating properly. `hw.status{hw.type="temperature_sensor"}` would actually be very confusing to me, as I'd interpret it as whether the sensor is working fine or not.

I'm not sure however if we can clarify this in the description of the attribute. I still hope my explanation above makes it clearer! 😅

docs/hardware/common.md
bertysentry: @lmolkova I understand. 

For my own knowledge, could you please clarify what you call "instrumentation"?

.github/ISSUE_TEMPLATE/bug_report.yaml
trisch-me: hw as abbreviation seems strange to me in registry so I prefer hardware, but I haven't changed the namespace.
@bertysentry do you think we should/can change common namespace for hardware to be indeed `hardware` and not `hw`?

docs/hardware/common.md
trisch-me: Should I rephrase this sentence then such as 
```suggestion
Application developers should attach all the metrics in the `hw` namespace to a `Host Resource](/docs/resource/host.md)."
```

model/registry/hardware.yaml
trisch-me: @lmolkova I would be ok to use existing `error.type` but I don't see how I can redefine enum unless I redefine the global one.

I would need to make `error.type` an enum of fields such as for example `zero_buffer_credit`, `crc`, `corrected` etc depending on the type of hardware. There is no reusable enums currently. So the only thing I could do here would be update comments and put examples there. 

@bertysentry is it ok from your perspective if type of error is not defined as enum but could have any value? We do not restrict enums currently anyway but sometimes it is nice to have predefined set of values. Here I can put set of values only to examples session

model/registry/hardware.yaml
lmolkova: I see, thanks for the explanation!

I agree that `temperature_sensor` would not make sense either then.

I guess that's another big thing to revisit in the https://github.com/open-telemetry/semantic-conventions/issues/1071 and should not block this PR.

So for the future
- `hw.type` name could be more precise. But also wouldn't it be clearer to report temperature as a separate metric (attribute on spans) and let someone decided if it's good/bad at a query time?  `hw.type=temperature, hw.status=failed` does not quite make sense either
- we might need to break this enum down into a component type (cpu|fan|etc) & component state|indicator|etc?. E.g. if I want to report CPU temperature, how do I do it today? `name="AMD Rizen 42",type=temperature,status=ok` ?

For this PR:

can we clarify what's `hw.type` is in the `hw.type` note?  E.g.

> Describes the category of hardware component for which `hw.state` is reported. For example, `hw.type=temperature` along with `hw.state=degraded` describes that the temperature of the hardware component has degraded.

docs/hardware/common.md
```diff
@@ -158,11 +166,42 @@ This metric is [recommended][MetricRecommended].
...
+Telemetry consumers that aggregate data from multiple instrumentation libraries and applications
+should be prepared for `error.type` to have high cardinality at query time when no
+additional filters are applied.
+
+If the operation has completed successfully, instrumentations SHOULD NOT set `error.type`.
+
+If a specific domain defines its own set of error identifiers (such as HTTP or gRPC status codes),
+it's RECOMMENDED to:
+
+* Use a domain-specific attribute
```
bertysentry, line 193: Given the description of the `error.type` attribute, I suggest that we enforce `error.type="hardware"` (or `error.type="hw"`?) and let the instrumentation use `hw.error.type` to specify the type of hardware error.

In issue #940, @eero-t suggested to add another attribute `hw.error.category`, to define even more precisely the type of error.

docs/hardware/common.md
```diff
@@ -158,11 +166,42 @@ This metric is [recommended][MetricRecommended].
...
 |---|---|---|---|---|---|
 | [`hw.id`](/docs/attributes-registry/hardware.md) | string | An identifier for the hardware component, unique within the monitored host | `win32battery_battery_testsysa33_1` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| [`hw.type`](/docs/attributes-registry/hardware.md) | string | Type of the component | `battery`; `cpu`; `disk_controller` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| [`hw.error.type`](/docs/attributes-registry/hardware.md) | string | The type of error encountered by the component | `` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`hw.type`](/docs/attributes-registry/hardware.md) | string | Type of the component [1] | `battery`; `cpu`; `disk_controller` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [2] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
bertysentry, line 170: The description is confusing in the context of hardware metrics.

docs/attributes-registry/hardware.md
```diff
@@ -10,14 +10,15 @@
...
+| ----------- | ------ | ------------------------------------------------------------------------------------------------------------- | ----------------------------------- | ---------------------------------------------------------------- |
+| `hw.id`     | string | An identifier for the hardware component, unique within the monitored host                                    | `win32battery_battery_testsysa33_1` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `hw.name`   | string | An easily-recognizable name for the hardware component                                                        | `eth0`                              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `hw.parent` | string | Unique identifier of the parent component (typically the `id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0`                | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
bertysentry, line 17:
```suggestion
| `hw.parent` | string | Unique identifier of the parent component (typically the `hw.id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0`                | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

docs/hardware/common.md
```diff
@@ -97,10 +101,14 @@ This metric is [recommended][MetricRecommended].
...
 | [`hw.id`](/docs/attributes-registry/hardware.md) | string | An identifier for the hardware component, unique within the monitored host | `win32battery_battery_testsysa33_1` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| [`hw.type`](/docs/attributes-registry/hardware.md) | string | Type of the component | `battery`; `cpu`; `disk_controller` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`hw.type`](/docs/attributes-registry/hardware.md) | string | Type of the component [1] | `battery`; `cpu`; `disk_controller` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`hw.name`](/docs/attributes-registry/hardware.md) | string | An easily-recognizable name for the hardware component | `eth0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`hw.parent`](/docs/attributes-registry/hardware.md) | string | Unique identifier of the parent component (typically the `id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
bertysentry, line 106:
```suggestion
| [`hw.parent`](/docs/attributes-registry/hardware.md) | string | Unique identifier of the parent component (typically the `hw.id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

docs/hardware/common.md
```diff
@@ -158,11 +166,42 @@ This metric is [recommended][MetricRecommended].
...
+| [`hw.type`](/docs/attributes-registry/hardware.md) | string | Type of the component [1] | `battery`; `cpu`; `disk_controller` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [2] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 | [`hw.name`](/docs/attributes-registry/hardware.md) | string | An easily-recognizable name for the hardware component | `eth0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`hw.parent`](/docs/attributes-registry/hardware.md) | string | Unique identifier of the parent component (typically the `id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
bertysentry, line 172:
```suggestion
| [`hw.parent`](/docs/attributes-registry/hardware.md) | string | Unique identifier of the parent component (typically the `hw.id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

docs/hardware/common.md
```diff
@@ -224,10 +263,14 @@ This metric is [recommended][MetricRecommended].
...
 | [`hw.id`](/docs/attributes-registry/hardware.md) | string | An identifier for the hardware component, unique within the monitored host | `win32battery_battery_testsysa33_1` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| [`hw.type`](/docs/attributes-registry/hardware.md) | string | Type of the component | `battery`; `cpu`; `disk_controller` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`hw.type`](/docs/attributes-registry/hardware.md) | string | Type of the component [1] | `battery`; `cpu`; `disk_controller` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`hw.name`](/docs/attributes-registry/hardware.md) | string | An easily-recognizable name for the hardware component | `eth0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`hw.parent`](/docs/attributes-registry/hardware.md) | string | Unique identifier of the parent component (typically the `id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
bertysentry, line 268:
```suggestion
| [`hw.parent`](/docs/attributes-registry/hardware.md) | string | Unique identifier of the parent component (typically the `hw.id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

docs/hardware/common.md
```diff
@@ -290,10 +333,14 @@ This metric is [recommended][MetricRecommended].
...
 | [`hw.state`](/docs/attributes-registry/hardware.md) | string | The current state of the component | `ok`; `degraded`; `failed` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| [`hw.type`](/docs/attributes-registry/hardware.md) | string | Type of the component | `battery`; `cpu`; `disk_controller` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`hw.type`](/docs/attributes-registry/hardware.md) | string | Type of the component [1] | `battery`; `cpu`; `disk_controller` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`hw.name`](/docs/attributes-registry/hardware.md) | string | An easily-recognizable name for the hardware component | `eth0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`hw.parent`](/docs/attributes-registry/hardware.md) | string | Unique identifier of the parent component (typically the `id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
bertysentry, line 338:
```suggestion
| [`hw.parent`](/docs/attributes-registry/hardware.md) | string | Unique identifier of the parent component (typically the `hw.id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

docs/hardware/common.md
trisch-me: I can overwrite it here - what do you propose?

docs/hardware/common.md
trisch-me: later error.type would have precise specific values i.e. https://opentelemetry.io/docs/specs/semconv/system/hardware-metrics/#hwcpu---physical-processor-metrics or https://opentelemetry.io/docs/specs/semconv/system/hardware-metrics/#hwgpu---gpu-metrics so `error.type` will be defined for specific hardware component

eero-t: > In issue #940, @eero-t suggested to add another attribute `hw.error.category`, to define even more precisely the type of error.

I changed my suggestion later to: https://github.com/open-telemetry/semantic-conventions/issues/940#issuecomment-2085109515
* `.severity`: correctable, uncorrectable (data loss), fatal (functionality loss)
* `.type`: type / category of the error; timeout, memory-parity-error, hw-programming-error etc
* `.location` / `.origin` / `.cause` / `.whatever`: e.g. http-endpoint, display (part of GPU with some memory), compute (GPU pipeline setting)

trisch-me: This type of changes should be an additional PR after this one gets merged. So we don't mix refactoring (with minor changes) and major changes/new functionality

bertysentry:
We're not done yet!

.github/ISSUE_TEMPLATE/bug_report.yaml
trisch-me: @bertysentry could you please check this comment?

trisch-me:
@lmolkova @bertysentry I have updated all the comments except the naming one. could you look once more

ChrsMark (approved):

.github/ISSUE_TEMPLATE/bug_report.yaml
bertysentry: I'm not quite sure. I don't think the `area` field is supposed to match the namespace (see `gen-ai` which is not namespace for generative AI metrics).

So I think it's okay to keep `area:hardware` value, as it's just one of the possible options in the issue template.

bertysentry:
Approved! Let's merge this so we can work the remaining issues!

bertysentry (approved):

model/registry/hardware.yaml
```diff
@@ -0,0 +1,111 @@
...
+        examples: ['eth0']
+      - id: parent
+        type: string
+        stability: experimental
+        brief: >
+          Unique identifier of the parent component (typically the `hw.id` attribute of the enclosure, or disk controller)
+        examples: ['dellStorage_perc_0']
+      - id: type
+        type:
+          allow_custom_values: true
```
lmolkova, line 28:
```suggestion
```

model/registry/hardware.yaml
```diff
@@ -0,0 +1,111 @@
...
+        stability: experimental
+        brief: >
+          Type of the component
+        note: >
+          Describes the category of the hardware component for which `hw.state` is being reported. For example,
+          `hw.type=temperature` along with `hw.state=degraded` would indicate that the temperature of the hardware
+          component has been reported as `degraded`.
+      - id: state
+        type:
+          allow_custom_values: true
```
lmolkova, line 95:
```suggestion
```

docs/system/hardware-metrics.md
```diff
@@ -44,43 +47,8 @@ when creating instruments not explicitly defined in the specification.
 > * SHOULD introduce a control mechanism to allow users to opt-in to the new
 >   conventions once the migration plan is finalized.
 
-## Common hardware attributes
```
lmolkova, line 47: could you please add link from this document to the common.md#common-attributes and mention that attributes are defined there. Attributes declared here  apply to all metrics in this file and with them removed, this file is not complete.

docs/system/hardware-metrics.md
lmolkova: Please add a link to the common metrics section of the new document, so people would see it among other metrics (and not just in the note above)

model/hardware-common.yaml
```diff
@@ -0,0 +1,12 @@
+groups:
+  - id: hardware.attributes.common
+    type: attribute_group
+    stability: experimental
+    brief: 'Hardware common attributes'
```
lmolkova, line 5:
```suggestion
    brief: 'Common hardware attributes'
```

docs/hardware/common.md
```diff
@@ -0,0 +1,380 @@
...
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `hw.energy` | Counter | `J` | Energy consumed by the component | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+<!-- semconv metric.hardware.energy(full) -->
```
lmolkova, line 94:
```suggestion
<!-- semconv metric.hardware.energy -->
```

(and in other places)

model/metrics/hardware/common.yaml
```diff
@@ -0,0 +1,61 @@
...
+    stability: experimental
+    brief: "Number of errors encountered by the component"
+    instrument: counter
+    unit: "{error}"
+    extends: metric.hardware.attributes
+    attributes:
+      - ref: error.type
+        brief: "The type of error encountered by the component"
+        examples: ['uncorrected', 'zero_buffer_credit', 'crc', 'bad_sector']
+        requirement_level: recommended
```
lmolkova, line 32:
```suggestion
        requirement_level:
          conditionally_required: if and only if an error has occured
        note: >
          The `error.type` SHOULD match the error code reported by the component,
          the canonical name of exception that occurred, or another low-cardinality error identifier.
          Instrumentations SHOULD document the list of errors they report.  
```

model/metrics/hardware/common.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+  - id: metric.hardware.power
+    type: metric
+    metric_name: hw.power
+    stability: experimental
+    brief: "Instantaneous power consumed by the component"
+    note: >
+      `hw.energy` is preferred
```
lmolkova, line 40:
```suggestion
      It's RECOMMENDED to report `hw.energy` metric instead of `hw.power` when possible. 
```

docs/hardware/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Status**: [Experimental][DocumentStatus]
+
+This document describes instruments and attributes for common hardware level
+metrics in OpenTelemetry. Consider the [general metric semantic conventions](/docs/general/metrics.md#general-metric-semantic-conventions)
+when creating instruments not explicitly defined in the specification.
+
+Semantic conventions for hardware are defined as following:
+
+* [Common Hardware Metrics](common.md): Semantic Conventions for hardware *common* metrics.
```
lmolkova, line 15:
```suggestion
* [Common Hardware Metrics](common.md): Semantic Conventions for  *common* hardware metrics.
```

docs/hardware/common.md
```diff
@@ -0,0 +1,380 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Hardware Common Metrics
+--->
+
+# Semantic Conventions for Hardware Common Metrics
```
lmolkova, line 5:
```suggestion
# Semantic Conventions for Common Hardware Metrics
```

docs/hardware/common.md
```diff
@@ -0,0 +1,380 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Hardware Common Metrics
```
lmolkova, line 2:
```suggestion
linkTitle: Common Hardware Metrics
```

docs/system/hardware-metrics.md
```diff
@@ -10,11 +10,14 @@ This document describes instruments and attributes for common hardware level
 metrics in OpenTelemetry. Consider the [general metric semantic conventions](/docs/general/metrics.md#general-metric-semantic-conventions)
 when creating instruments not explicitly defined in the specification.
 
+This document is being converted to specific hardware metrics, parts of this document that have already been
+converted are now located in the [Hardware] (/docs/hardware/README.md) folder and are no longer present in this file.
```
lmolkova, line 14:
```suggestion
converted are now located in the [Common Hardware Metrics](/docs/hardware/common.md) folder and are no longer present in this file.
```

docs/system/hardware-metrics.md
```diff
@@ -10,11 +10,14 @@ This document describes instruments and attributes for common hardware level
...
+
+Please note that this is an ongoing process and may take some time to complete.
```
lmolkova, line 16: do we have an issue that outlines what needs to happen? Can we add a todo comment here and add link to that issue?

trisch-me: I have created a tracking issue for this https://github.com/open-telemetry/semantic-conventions/issues/1309

model/metrics/hardware/common.yaml
trisch-me: I will update it as it makes more sense but I need to note that originally it was a `recommended`

docs/system/hardware-metrics.md
trisch-me: Updated

docs/system/hardware-metrics.md
trisch-me: Updated

bertysentry (approved):

docs/system/hardware-metrics.md
trisch-me: This is not correct, it is about all hardware metrics and not only common

lmolkova (approved):

PR #1042, opened 2024-05-15
Add session.change event

breedx-splk:
## Changes

This adds semantic conventions for an event called `session.change`.

* [x] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions. (**I think this is N/A due to automation lacking for events**)


.chloggen/session_change_event.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: 'session'
```
breedx-splk, line 10: I have no idea if this `component` is real or correct.

jsuereth: I think @joaopgrassi's tooling infers components and generates them.

Could create a YAML in the model for this event with just the name?  Then you can have a (probably empty) snippet of the form:

```
<!-- semconv session.change_event -->
<!-- endsemconv -->
```

breedx-splk: Do we think that's worth doing now, or wait until the event tooling is baked?

docs/general/session.md
```diff
@@ -35,4 +35,14 @@ backends can link the two sessions.
...
 
+## Session Change Event
+
+![Experimental](https://img.shields.io/badge/-experimental-blue)
+
+When a session changes (due to time-based expiry or another mechanism), an event MAY be emitted.
+
+The emitted event that represents a session change MUST have the `event.name=session.change`.
+The event body MUST be empty and the attributes MUST include both the `session.id` and `session.previous_id`
+attributes, as described above. The values MUST be different.
```
scheler, line 46: The wording `session.change` indicates there is another event to represent when a new session starts. However, there is no such event currently. I suggest we call this `session.start` or just `session`, and this event MUST include `session.id` and MAY include `session.previous_id` when applicable.

Structurally, this event should be described using the Events format. See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/mobile/events.md for an example.

scheler (changes requested):

docs/general/session.md
breedx-splk: > The wording `session.change` indicates there is another event to represent when a new session starts.

That's a bit of a stretch to me, but is probably highlighting the fact that a larger spec around session handling might be missing. It's perfectly reasonable to also have other telemetry include the session without a session start event, IMO.

> However, there is no such event currently. I suggest we call this `session.start` or just `session`, and this event MUST include `session.id` and MAY include `session.previous_id` when applicable.

Yeah, I think this demonstrates that what you're suggesting is semantically something different than what I was trying to define here.

 
> Structurally, this event should be described using the Events format. See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/mobile/events.md for an example.

Agreed, however there's no precedent for an empty body. Maybe you're suggesting that this event should have those attributes as body fields instead?

bidetofevil: Given that there's no notion of a session right now beyond that all signals that contain the same `session.id` attribute with a given value conceptually belong to the same session, having this to signal that the previously set `session.id` is no longer valid seems reasonable to have even if there's no equivalent `session.start` event. 

The latter would be useful if there was additional metadata attached to said session, but in the absence of that, I don't think we necessarily need that. Or if we do, it should be part of a different proposal to define what a session is and how it relates to signals.

docs/general/session.md
```diff
@@ -35,4 +35,14 @@ backends can link the two sessions.
...
 <!-- endsemconv -->
 
+## Session Change Event
+
+![Experimental](https://img.shields.io/badge/-experimental-blue)
+
+When a session changes (due to time-based expiry or another mechanism), an event MAY be emitted.
+
+The emitted event that represents a session change MUST have the `event.name=session.change`.
+The event body MUST be empty and the attributes MUST include both the `session.id` and `session.previous_id`
```
bidetofevil, line 45: Should we allow `session.id` to be null to denote the change resulting in a session expiring but no new session beginning?

breedx-splk: Maybe. What's the use case? Session end? Seems like we could have a separate event then, keeping parity with a hypothetical session start event.

bidetofevil: The use case is when the "change" doesn't lead to a new session - just the ending of the previous one, like when there is a requirement for a new session that is not fulfilled.

For instance, if a session times out, but the app is in the background, and no telemetry will be recorded while that is the case, do we want to generate a new `session.id`?

An alternative is to use some non-null stub to indicate that there's no session, e.g. empty-string. If we go down that route, we should define what that non-null stub should be.

docs/general/session.md
breedx-splk: Ok so I'm not opposed to using body fields, but if we have previously standardized on "`session.id` is an attribute that gets slapped onto all client telemetry`, we either need to make an exception for this specific event (eg. "The session attribute isn't here because it's in the body") or we need to exclude it from the body, which means there's just the one `session.previous.id` in the body). 

I thought both of these options were more confusing than just having attributes.

.chloggen/session_change_event.yaml
joaopgrassi: We already have a `session.yaml` in the registry` https://github.com/open-telemetry/semantic-conventions/blob/main/model/registry/session.yaml so this is correct already :)

docs/general/session.md
joaopgrassi: I agree with having separated events for single cases like session change/end/start.

docs/general/session.md
joaopgrassi: Sorry if what I will say is completely dumb, but reading the changes and the discussions here I'm thinking: What if we he only have `session.start` and `session.end`? And to signal a "change in session", we emit a `session.end` and a `session.start` having a `session.previous_id`? Then when a new start even doesn't have the `session.previous_id` we know it's a brand new session, and when we have the attribute it indicates a change. Was this option discussed?

breedx-splk: It wasn't discussed. That approach sounds viable as well, although it has two minor concerns for me, neither of which is a deal-breaker:
1. there is a bit of redundancy in the "session change" case -- that is, an end and and start are both always emitted, while the presence of `session.previous_id` in the start case already suggests that the prior session is ending (or has just ended)
2. does the event receiver then have to pay slightly more attention to event ordering? If a session change results in a stop/start sequence that is out of order (start/stop), is that MUCH harder to handle (or require keeping state?). I don't know if it's a problem, but I know enough to be cautious about that choice.

In any case, it seems that folks are thinking more in terms of start/stop than "change", so I'll mark this one draft and write up another PR that tries this. Thanks for the suggestions folks.

bidetofevil: > Ok so I'm not opposed to using body fields, but if we have previously standardized on "session.id is an attribute that gets slapped onto all client telemetry

Perhaps we are overloading the meaning of `session.id` then. The one in attributes indicates what session this event belongs to, while for the proposed body field, it's the NEW sessionId being introduced - so it might make more sense as `session.new_id` or something?

FWIW, I also agree that making this event "atomic" to express both the end and beginning at the same time is a good idea, in case the telemetry comes in out of order, or just delayed, which will happen pretty frequently).

bidetofevil (approved):
I think I still prefer making `session.next_id` and `session.previous_id` body fields and reserving `session.id` to denote what the session the event actually belongs to, so the meaning of `session.id` is not overloaded.

That said, I understand wanting to minimize duplication, so I'll leave it up to you to make the call after saying my piece.

breedx-splk:
> I think I still prefer making `session.next_id` and `session.previous_id` body fields and reserving `session.id` to denote what the session the event actually belongs to, so the meaning of `session.id` is not overloaded.
> 
> That said, I understand wanting to minimize duplication, so I'll leave it up to you to make the call after saying my piece.

Yeah, it gets weird in this liminal space for sure. Have a look at the start/end events over in #1091 and see if that maybe makes a little more sense.

breedx-splk:
Closing in favor of #1091. Thanks y'all!

PR #1046, opened 2024-05-17
[chore] Move last 'loose' attribute to registry.

jsuereth:
As of this PR - every single attribute is now located in the registry and all attribute_group, event or metric types reference their attributes from the registry.


reyang (approved):

arminru (approved):

trisch-me:
@jsuereth you were faster in creating PR :)

PR #1047, opened 2024-05-17
Propose semantic conventions for GCP client library identification

michaelsafyan:
Propose semantic conventions for GCP client library identification and attribution.

## Changes

Defines a set of `gcp.client.*` scope attributes for identifying Google Cloud client libraries.

* ✅ [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* ✅ Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
* N/A: [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.


model/trace/instrumentation/gcp-client.yml
```diff
@@ -0,0 +1,43 @@
...
+    type: scope
+    brief: >
+      The `gcp.client` conventions are intended to be used by Google Cloud
+      client libraries (e.g. https://github.com/googleapis/* and other
+      Google-owned/Google-provided client libraries for Google Cloud) to
+      identify such clients and to track their operations and performance.
+    prefix: gcp.client
+    stability: experimental
+    attributes:
+      - id: gcp.client.service
```
jsuereth, line 12: I'm not sure this is well documented, but we have a requirement now that attributes are part of the attribute registry.

Specifically, you should create a file `model/registry/gcp.yaml` or `model/reigstry/gcp-client.yaml`.

In that registry file you can define a group with type `attribute_group`, that defines these attributes but without a `requirement_level`.

That will do the minimum to reserve the attribute names.  Next, you'll want a second attribute_group (not scope I think) which using `ref` to pull in the attributes and give them a requirement level.  This group would be inhereted by GCP client library instrumentation once it has spans defined on its behalf.

I think for now, I'd just reserve the attributes.

jsuereth: Also, after you create the YAML file for the registry, make sure to run `make attribute-registry-generation`

trisch-me: We have already gcp in registry, so you could add your data there, to the yaml file

trisch-me: Also if you have prefix defined, you don't need to repeat it in `id`

model/trace/instrumentation/gcp-client.yml
```diff
@@ -0,0 +1,43 @@
...
+        type: string
+        stability: experimental
+        brief: >
+          Identifies the repository where the client library source comes from.
+          The prefix `github.com/` should be not be included in the name. If 
+          (`gcp.client.service`, `telemetry.sdk.language`) do not uniquely
+          identify the client such as when multiple client variants exist
+          for the same service in a given programming language, this field
+          should be supplied in order to disambiguate between clients.
+      - id: gcp.client.artifact
```
lmolkova, line 34: what's the relationship between instrumentation scope name and the artifact name?

We used to think about instrumentation scope name as an artifact name, but it's not a requirement  (and not a right level of granularity), I believe we miss a general-purpose attribute to identify the artifact name and the repository. there is nothing specific to GCP here.

I'd suggest adding attributes like `instrumentation.artifact.name` and `instrumentation.artifact.repo` (in case of instrumentation library, they should match the instrumentation library name, in case of native instrumentation, they should match the name of the original library).

model/trace/instrumentation/gcp-client.yml
```diff
@@ -0,0 +1,43 @@
...
+        stability: experimental
+        requirement_level: required
+        brief: >
+          Short name of the GCP service for which the library is a client.
+          All variants of the client across programming languages should use an
+          identical value for this property. The short name should be intuitive
+          and aligned with either the name of the client (e.g. X for client
+          library packages named `google-cloud-X`) or the name of the API
+          subdomain (e.g. X for APIs hosted at `X.googleapis.com`). 
+        examples: ['cloudsql', 'datastore', 'spanner', 'firestore']
```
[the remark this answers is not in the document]
michaelsafyan: Thanks for the review and for your input!

Just to provide a bit more context regarding our goals/motivations, we are trying to enable shared, centralized analysis, dashboards, etc. across GCP client libraries written by various different teams.

We want to be able to easily identify when instrumentation comes from a GCP client library as well as to be able to easily identify that client in a uniform, domain-agnostic manner that does not require tweaking/adapting analysis, dashboards, etc. for each different kind of shape of client that may arise.

It is possible that some attributes may at times contain overlapping information. I am not sure that these meet these same goals or necessarily overlap all of the time, though. For example, let's consider the case of a higher level abstraction layer library that seeks to provide a vendor-neutral interface to interoperation with a database. Such a library might set `db.system` to `firestore` when `firestore` is configured as the backend system, but we would not expect it to set `gcp.client.service` if it is not a client library for that service provided by Google.

model/trace/instrumentation/gcp-client.yml
michaelsafyan: Thanks! I like the suggestion of making these into general scope properties.

As for "name", there seemed to be some variability into how it was being interpreted/set.

lmolkova: > As for "name", there seemed to be some variability into how it was being interpreted/set.

agreed, we can consider `full_name` or something similar, suggest to use whatever is idiomatic in specific language/ecosystem and add a few examples (e.g. `com.google.cloud:google-cloud-storage`, `@azure/storage-blob`, `opentelemetry-instrumentation-boto`)

model/trace/instrumentation/gcp-client.yml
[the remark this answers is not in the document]
lmolkova: thank you for the explanation! yes, it makes sense. JFYI, we have a similar thing [on Azure](https://github.com/Azure/azure-sdk/blob/main/docs/tracing/distributed-tracing-conventions.md#public-api-calls) and call it `az.namespace`

model/trace/instrumentation/gcp-client.yml
```diff
@@ -0,0 +1,43 @@
...
+    stability: experimental
+    attributes:
+      - id: gcp.client.service
+        type: string
+        stability: experimental
+        requirement_level: required
+        brief: >
+          Short name of the GCP service for which the library is a client.
+          All variants of the client across programming languages should use an
+          identical value for this property. The short name should be intuitive
```
dbolduc, line 19: You say that these all need to be identical. I agree. I think using the endpoint name will be deterministic. (e.g. `foo` from `foo.googleapis.com`).

Context from the client library perspective:
I think the short names of clients vary across the client library languages. At least in C++ we did a lot of ad-hoc naming. `google-cloud-X` will not be consistent, so let's not mention it.

model/trace/instrumentation/gcp-client.yml
michaelsafyan: Done:

 - Removed the repetition of the `prefix` in `id`.
-  Executed `make attribute-registry-generation` and verified that it passed.

model/trace/instrumentation/gcp-client.yml
```diff
@@ -0,0 +1,43 @@
...
+          and aligned with either the name of the client (e.g. X for client
+          library packages named `google-cloud-X`) or the name of the API
+          subdomain (e.g. X for APIs hosted at `X.googleapis.com`).
```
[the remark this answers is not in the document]
michaelsafyan: Thanks for that suggestion. I've shortened each `brief` and put the larger commentary into `note`.

model/trace/instrumentation/gcp-client.yml
michaelsafyan: I've updated this code based on your suggestions to generalize. I ended up at `instrumentation.source.` instead of `instrumentation.artifact.`, because the intention was to describe the original source code more so than where it is being packaged up and published. However, I think this falls in line with where you were suggesting/heading.

To summarize, the new attributes are:

  - `gcp.client.service`
  -  `instrumentation.source.repo`
  -  `instrumentation.source.target`

I made one slight semantic change with `repo`; previously, I had it omit the domain (because GCP client libraries are always published on GitHub). With this generalization, I felt that it was appropriate to include the `github.com/` prefix so as to make this more generalizable to code which may be hosted elsewhere.

model/trace/instrumentation/gcp-client.yml
michaelsafyan: Thanks! I changed the guidance so that the derivation is done solely from the API domain name.

model/registry/gcp.yaml
```diff
@@ -1,4 +1,28 @@
...
+  - id: registry.gcp.client
+    prefix: gcp.client
+    type: attribute_group
+    brief: >
+      Attributes for Google Cloud client libraries.
+    attributes:
+      - id: service
+        type: string
+        stability: experimental
+        brief: Identifies official Google Cloud client libraries.
```
lmolkova, line 15: maybe
```suggestion
        brief: Identifies the Google Cloud service for which the official client library is intended.
```

model/registry/instrumentation.yaml
```diff
@@ -0,0 +1,24 @@
+groups:
+- id: registry.instrumentation
+  prefix: instrumentation
+  type: attribute_group
+  brief: >
+    Describes properties about the library that is being instrumented or that
+    is responsible for performing the instrumentation.
```
lmolkova, line 7:
```suggestion
    Describes the instrumentation library - the original library if it's instrumented natively or the instrumentation library.
```

model/registry/instrumentation.yaml
```diff
@@ -0,0 +1,24 @@
...
+  attributes:
+    - id: source.repo
+      type: string
+      stability: experimental
+      brief: The repository associated with the InstrumentationScope.
```
lmolkova, line 12:
```suggestion
      brief: Link to the instrumentation library source code.
```

or 

```suggestion
      brief: Instrumentation library project URL.
```

model/registry/instrumentation.yaml
```diff
@@ -0,0 +1,24 @@
...
+      examples: [github.com/organization/repo, gitlab.com/group/repo]
+    - id: source.target
+      type: string
+      stability: experimental
+      brief: The build target associated with the InstrumentationScope.
```
lmolkova, line 17:
```suggestion
      brief: The instrumentation library artifact name
```

model/registry/instrumentation.yaml
```diff
@@ -0,0 +1,24 @@
...
+      examples: ['//in/bazel/path/to/my/folder:target', 'inGradleYourLib']
```
lmolkova, line 18: This information seems to be an internal implementation detail - why is this important on the end-user telemetry? 
I was hoping to have the artifact name.

model/trace/instrumentation/instrumentation.yml
```diff
@@ -0,0 +1,13 @@
...
+- id: instrumentation.scope
+  type: scope
+  brief: >
+    The `instrumentation` conventions are intended to provide common
+    mechanisms for triangulating and identifying instrumentation libraries.
+  prefix: instrumentation
+  stability: experimental
+  attributes:
+  - ref: instrumentation.source.repo
+    requirement_level: opt_in
```
lmolkova, line 11: does it need to be opt-in? I.e. users would need to enable it explicitly. Why not `recommended: if not captured otherwise, e.g. in the instrumentation scope name attribute`? (i.e. instrumentation libraries may decide to populate it if it makes sense)

model/registry/instrumentation.yaml
```diff
@@ -0,0 +1,24 @@
+groups:
+- id: registry.instrumentation
+  prefix: instrumentation
+  type: attribute_group
+  brief: >
+    Describes properties about the library that is being instrumented or that
+    is responsible for performing the instrumentation.
+  attributes:
+    - id: source.repo
```
lmolkova, line 9: npm, python, maven, nuget allow a general-purpose link to anything and call it:
- `homepage` - npm
- `project_urls` - pypi
- `url` - maven
- `projectUrl` - nuget

This 1) describes the format (URL) 2) allows any link to be added, not just the library repo which can be closed source, etc.

So I'd prefer to call this attribute something like `instrumentation.scope|library.project_url`

model/registry/instrumentation.yaml
```diff
@@ -0,0 +1,24 @@
...
+      type: string
+      stability: experimental
+      brief: The repository associated with the InstrumentationScope.
+      examples: [github.com/organization/repo, gitlab.com/group/repo]
```
lmolkova, line 13: Why not full url?

model/registry/instrumentation.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - id: source.target
```
lmolkova, line 14: See my comment below - I'm not sure why build target would be interesting to end users and would prefer to see an artifact name here.

I'd suggest `instrumentation.scope|library.artifact_name`

model/registry/instrumentation.yaml
michaelsafyan: Between these two options, I prefer the former; however, this is somewhat dependent on the other conversation regarding whether it should be a full URL or not. Leaving this open pending resolution of https://github.com/open-telemetry/semantic-conventions/pull/1047#discussion_r1617758202

model/registry/instrumentation.yaml
michaelsafyan: To me, "artifact" implies that it names the output/result; while "target" implies the structure of the build in the source. The problem with the "artifact" nomenclature is that it requires that the project has reached a sufficient level of maturity as to have packaging in place, precluding complete instrumentation before that...

I think it would be ideal to be able to track and differentiate these components prior to packaging/deployment being completed. It might be nice to have additional attributes available related to packaging/deployment, however, it would be good to be able to begin tracking / differentiating at an earlier lifecycle stage.

model/registry/instrumentation.yaml
michaelsafyan: I think there are two separate things that are of interest:

  1. The _source_ / _identity_ / _provenance_ of the code.
  2. The _distribution channel_ through which it was installed/downloaded.

It is possible for the same original source to be bundled, packaged, distributed, etc. through different distribution channels (e.g. installation through `homebrew` vs `fink` or through `pacman` vs `apt`) ...

For the particular use case that I have in mind, I'm interested more in case 1. I feel like `artifact`, `asset`, etc. are more properties associated with case 2. Both can be useful, but for different purposes.

For the telemetry end user, the build target can be relevant if the same repo contains multiple different variants of the same thing. For example, you could imagine that there is  `fast_debug_build`, `asan_enabled_build`, `optimized_deployment_build` of the component. If there are unusual behaviors or problems that are happening, knowing which of these variants was used could provide some insight regarding what is going on. It also provides useful information to include in bug reports raised to the maintainers of the instrumentation library to identify specifically which component is the one that is being complained about.

model/trace/instrumentation/gcp-client.yml
michaelsafyan: I've moved the general, vendor-agnostic conventions to a separate pull request:

  - https://github.com/open-telemetry/semantic-conventions/pull/1094

Let's continue the conversation there.

In the meantime, I hope we can make progress on and get in `gcp.client.service` (now the sole focus of this PR).

Thanks!

model/registry/instrumentation.yaml
michaelsafyan: I followed this suggestion of making it a full URL in https://github.com/open-telemetry/semantic-conventions/pull/1094

model/registry/instrumentation.yaml
michaelsafyan: Given this feedback, I added some distribution-related attributes in:

 - https://github.com/open-telemetry/semantic-conventions/pull/1094

Let's continue that conversation there.

model/registry/instrumentation.yaml
michaelsafyan: No longer applicable in the scope of this pull request; moved to https://github.com/open-telemetry/semantic-conventions/pull/1094

model/registry/instrumentation.yaml
michaelsafyan: My biggest concern with this suggestion is that it seems to be language-specific, whereas the aim of that work was towards facilitating cross-language analysis of cross-platform / multi-platform libraries that have different per-language variants. Regardless, this is no longer in the scope of this PR. Let's continue that discussion in https://github.com/open-telemetry/semantic-conventions/pull/1094

model/registry/instrumentation.yaml
michaelsafyan: Ultimately went with that suggestion in https://github.com/open-telemetry/semantic-conventions/pull/1094

model/trace/instrumentation/instrumentation.yml
michaelsafyan: I did not realize earlier that `recommended` could be conditional. I followed that pattern in https://github.com/open-telemetry/semantic-conventions/pull/1094

model/registry/instrumentation.yaml
michaelsafyan: I believe I addressed that question here:

 - https://github.com/open-telemetry/semantic-conventions/pull/1047#discussion_r1619399324

"the build target can be relevant if the same repo contains multiple different variants of the same thing. For example, you could imagine that there is fast_debug_build, asan_enabled_build, optimized_deployment_build of the component. If there are unusual behaviors or problems that are happening, knowing which of these variants was used could provide some insight regarding what is going on"

The "artifact" or "asset" implies some kind of bundling/packaging/distribution, which may not always be applicable. But it may still be useful for other contexts.

I'm going to resolve this here on this PR, because these changes have been extracted to:

 - https://github.com/open-telemetry/semantic-conventions/pull/1094

I've narrowed the scope of _this_ pull request to just `gcp.client.service` under the philosophy that smaller, incremental, more narrowly focused changes are easier to discuss, review, and update compared to large ones.

docs/attributes-registry/gcp.md
```diff
@@ -6,9 +6,28 @@
...
+| -------------------- | ------ | ------------------------------------------------------------------------------------------ | ----------------------------------------------------- | ---------------------------------------------------------------- |
+| `gcp.client.service` | string | Identifies the Google Cloud service for which the official client library is intended. [1] | `appengine`; `run`; `firestore`; `alloydb`; `spanner` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1]:** Intended to be a stable identifier for Google Cloud client libraries that is uniform across implementation languages. The value should be derived from the canonical service domain for the service; for example, 'foo.googleapis.com' should result in a value of 'foo'. This attribute should be accompanied by 'instrumentation.source.repo' and/or 'instrumentation.source.target' when the combination of 'gcp.client.service' and 'telemetry.sdk.language' is insufficient to uniquely identify the particular client library.
```
lmolkova, line 29: Since `instrumentation.source.*` attributes are not defined in the PR, maybe this sentence can be removed?

```suggestion
**[1]:** Intended to be a stable identifier for Google Cloud client libraries that is uniform across implementation languages. The value should be derived from the canonical service domain for the service; for example, 'foo.googleapis.com' should result in a value of 'foo'. 
```

model/trace/instrumentation/gcp-client.yml
```diff
@@ -0,0 +1,12 @@
+groups:
+- id: gcp.client.scope
+  type: scope
```
lmolkova, line 3: please don't use `scope` type, it's going to be removed. Also, many languages don't support scope attributes and the only way to report it would be through regular span/metric/event attribute.

```suggestion
  type: attribute_group
```

docs/attributes-registry/gcp.md
michaelsafyan: Done.

model/trace/instrumentation/gcp-client.yml
michaelsafyan: Thanks! Good to know that this is being removed. Updated to use `attribute_group` instead of `scope`. Also changed the `id` to not reference `scope` given this change in type.

lmolkova (approved):

PR #1048, opened 2024-05-17
move jvm attribute to the registry

trisch-me:
While implementing bugfix for a weaver I found out that https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/jvm-metrics.md#metric-jvmsystemcpuload_1m has link to the 
```
 [`jvm.buffer.pool.name`](/docs/attributes-registry/jvm.md
```

But we don't define this attribute there. I have fixed it, but probably this is one of cases where we need to add constraint policy to the weaver? WDYT @jsuereth 

### DIFF
--- a/model/registry/jvm.yaml
+++ b/model/registry/jvm.yaml
@@ -79,3 +79,11 @@ groups:
               value: 'terminated'
               brief: 'A thread that has exited is in this state.'
               stability: stable
+      - id: buffer.pool.name
+        type: string
+        stability: experimental
+        brief: Name of the buffer pool.
+        examples: [ "mapped", "direct" ]
+        note: >
+          Pool names are generally obtained via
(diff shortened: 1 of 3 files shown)


model/registry/jvm.yaml
```diff
@@ -79,3 +79,11 @@ groups:
               value: 'terminated'
               brief: 'A thread that has exited is in this state.'
               stability: stable
+      - id: buffer.pool.name
```
reyang, line 82: Looks like duplicated work? https://github.com/open-telemetry/semantic-conventions/pull/1046

reyang (changes requested):
Please take a look at https://github.com/open-telemetry/semantic-conventions/pull/1046

trisch-me:
close as duplicated :)

PR #1066, opened 2024-05-23
Add v8 js engine runtime metrics semantic conventions

maryliag:
Part Of #990

## Changes

Adding semantic conventions for V8 Js Engine metrics


maryliag:
friendly ping for reviews

joaopgrassi (commented):
Did a first pass, I will try to get more familiar with Nodejs v8 and then look at it again.

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,84 @@
...
+    stability: experimental
+    prefix: v8js.gc
+    attributes:
+      - ref: v8js.gc.type
+        requirement_level: recommended
+
+  - id: metric.veightjs.memory.size
+    type: metric
+    metric_name: v8js.memory.size
+    brief: "External memory size."
```
legendecas, line 17: This metric is a little bit confusing. The name didn't reflect it is a figure for external memory size.

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,84 @@
...
+    brief: "Garbage collection duration."
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    prefix: v8js.gc
+    attributes:
+      - ref: v8js.gc.type
+        requirement_level: recommended
+
+  - id: metric.veightjs.memory.size
```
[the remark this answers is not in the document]
maryliag: Switched to be separate metrics

model/metrics/v8js-metrics.yaml
maryliag: Updated description

model/registry/v8js.yaml
```diff
@@ -0,0 +1,84 @@
...
+              stability: experimental
+            - id: incremental
+              value: 'incremental'
+              brief: 'Incremental (Incremental Marking).'
+              stability: experimental
+            - id: weakcb
+              value: 'weakcb'
+              brief: 'Weak Callbacks (Process Weak Callbacks).'
+              stability: experimental
+      - id: memory.state
```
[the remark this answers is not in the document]
maryliag: Switched to be separate metrics

model/registry/v8js.yaml
```diff
@@ -0,0 +1,84 @@
...
+              value: 'total'
+              brief: 'Total memory.'
+              stability: experimental
+            - id: used
+              value: 'used'
+              brief: 'Used memory.'
+              stability: experimental
+        brief: The state of memory.
+        examples: ["total", "used"]
+      - id: heap.size.state
```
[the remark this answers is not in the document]
maryliag: Switched to be separate metrics

legendecas (approved):

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,92 @@
...
+    unit: "By"
+    stability: experimental
+
+  - id: metric.veightjs.heap.total
+    type: metric
+    metric_name: v8js.heap.total
+    brief: "Total heap memory size."
+    instrument: updowncounter
+    unit: "By"
+    stability: experimental
```
joaopgrassi, line 36: Wouldn't `v8js.heap.space.name` make sense in the `.heap.*` metrics? 🤔

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,92 @@
...
+    prefix: v8js.gc
+    attributes:
+      - ref: v8js.gc.type
+        requirement_level: recommended
+
+  - id: metric.veightjs.memory.total
+    type: metric
+    metric_name: v8js.memory.total
+    brief: "Total memory size."
```
joaopgrassi, line 17: To what "total size" here refers to? Is it the total size it has used? it is the total size available? I think this needs a better explanation.

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,92 @@
...
+    metric_name: v8js.heap.used
+    brief: "Used heap memory size."
+    instrument: updowncounter
+    unit: "By"
+    stability: experimental
+
+  - id: metric.veightjs.heap.space_size
+    type: metric
+    metric_name: v8js.heap.space_size
+    brief: "Heap space size."
```
joaopgrassi, line 49: Same with this one. It would be better to describe a bit more what "heap space size" here means.

joaopgrassi (commented):
Press enter too soon 😓

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,92 @@
...
+    instrument: updowncounter
+    unit: "By"
+    stability: experimental
+    prefix: v8js.heap.space_size
+    attributes:
+      - ref: v8js.heap.space.name
+        requirement_level: recommended
+
+  - id: metric.veightjs.heap.space_used_size
```
joaopgrassi, line 58: Hum I wonder if this metric shouldn't be like the JVM one `jvm.buffer.memory.usage`. This is to track the usage of heap space, right? Maybe these can be like `v8js.heap.usage` and `v8js.heap.size`?

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,92 @@
...
+    unit: "By"
+    stability: experimental
+    prefix: v8js.heap.space_used_size
+    attributes:
+      - ref: v8js.heap.space.name
+        requirement_level: recommended
+
+  - id: metric.veightjs.heap.space_available_size
+    type: metric
+    metric_name: v8js.heap.space_available_size
```
joaopgrassi, line 72: What is the difference between this one and `v8js.heap.space_size`? 🤔

joaopgrassi (commented):
I see we have these metrics

```
v8js.memory.total
v8js.memory.used

v8js.heap.total
v8js.heap.used
```

That are basically duplicated. Did you consider grouping them in a single `v8js.memory.total|used`, and then having a `v8js.memory.type = heap|non_heap` like the JVM one has?  https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/jvm-metrics.md#metric-jvmmemoryused 

Then you are down to 2 metrics instead of 4.

joaopgrassi:
I also see we use `veightjs` and `v8js` in `id` and `metric name`. It would be good to be consistent.. and probably best to avoid the number in the metric name. I'd suggest to use `veightjs` all around.

maryliag (commented):
> I also see we use veightjs and v8js in id and metric name. It would be good to be consistent.. and probably best to avoid the number in the metric name. I'd suggest to use veightjs everywhere.

I only used the `veight` version on the ids used to generate the readme, they don't show up in the docs for the final user. I only used because there was a limitation on the docs generation that was not working properly when there were numbers on the ID name. Once a new version of the plugin with the fix is added, this can be updated to use v8. I prefer keeping v8 on the metric name, and not force a limitation of our tooling on the final result to the user. What do you think?

In case you're curios, the PR with the fix: https://github.com/open-telemetry/weaver/pull/152

legendecas:
I'd prefer keeping `v8` or `V8` since it is the name of the engine.

model/metrics/v8js-metrics.yaml
joaopgrassi: Would be great to have these links in here somewhere!

joaopgrassi:
@maryliag can you please file an issue so we don't forget to change the ids to use `v8js`?

model/metrics/v8js-metrics.yaml
```diff
@@ -14,10 +14,15 @@ groups:
...
-    brief: "Total memory size."
+    brief: "Total memory size pre-allocated."
     instrument: updowncounter
     unit: "By"
     stability: experimental
+    attributes:
+      - ref: v8js.memory.type
+        requirement_level: recommended
+      - ref: v8js.heap.space.name
+        requirement_level: recommended
```
joaopgrassi, line 25: Given the comment:

> Available only when memory type is heap

Should the requirement level `v8js.heap.space.name` here be conditionally required? So only there if `memory.type = heap`?

joaopgrassi (commented):
Thanks for the links @maryliag those were helpful for non JS experts like me :D. I think this looks pretty good. 

One thing I'm thinking is: there's several metrics in the "space" namespace and I wonder if we shouldn't convert that to an actual namespace. So like `v8js.heap.space_used_size` becomes `v8js.heap.space.used_size` and etc.

What do you think?

maryliag:
> can you please file an issue so we don't forget to change the ids to use v8js?

@joaopgrassi created the issue https://github.com/open-telemetry/semantic-conventions/issues/1173

model/metrics/v8js-metrics.yaml
maryliag: Updated

maryliag:
Included notes with the links for the v8 functions, so it's more clear where they come from and their names.
Updated the `space` names, to have `space.*` instead of `space_*`

> Do you think having some sort of direction on how these metrics are collected makes sense?

Not sure what else could be added for guidance. Someone else has actually started to implement these metrics on https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2136, so for someone who knows this area, should be more straight forward.

joaopgrassi (approved):

model/registry/v8js.yaml
```diff
@@ -0,0 +1,71 @@
+groups:
+  - id: registry.v8js
+    type: attribute_group
+    brief: Describes V8 JS Engine Runtime related attributes.
+    prefix: v8js
+    attributes:
+      - id: memory.type
+        stability: experimental
+        type:
+          allow_custom_values: false
```
lmolkova, line 10: nit:

this value is ignored

```suggestion
```

model/registry/v8js.yaml
```diff
@@ -0,0 +1,71 @@
...
+              value: 'non_heap'
+              brief: 'Non-heap memory'
+              stability: experimental
+        brief: The type of memory.
+        examples: ["heap", "non_heap"]
+      - id: gc.type
+        stability: experimental
+        brief: The type of garbage collection.
+        type:
+          allow_custom_values: false
```
lmolkova, line 26:
```suggestion
```

model/registry/v8js.yaml
```diff
@@ -0,0 +1,71 @@
...
+              value: 'weakcb'
+              brief: 'Weak Callbacks (Process Weak Callbacks).'
+              stability: experimental
+      - id: heap.space.name
+        stability: experimental
+        brief: The name of the space type of heap memory.
+        note: >
+          Value can be retrieved from value `space_name` of [`v8.getHeapSpaceStatistics()`](https://nodejs.org/api/v8.html#v8getheapspacestatistics)
+        type:
+          allow_custom_values: false
```
lmolkova, line 50:
```suggestion
```

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,96 @@
...
+    unit: "s"
+    stability: experimental
+    prefix: v8js.gc
+    attributes:
+      - ref: v8js.gc.type
+        requirement_level: recommended
+
+  - id: metric.veightjs.memory.total
+    type: metric
+    metric_name: v8js.memory.total
```
lmolkova, line 16:
```suggestion
    metric_name: v8js.memory.limit
```

similar to [jvm.memory.limit](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/jvm-metrics.md#metric-jvmmemorylimit) and also 

> limit - an instrument that measures the constant, known total amount of something should be called entity.limit. For example, system.memory.limit for the total amount of memory on a system.

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#instrument-naming

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,96 @@
...
+    attributes:
+      - ref: v8js.memory.type
+        requirement_level: recommended
+      - ref: v8js.heap.space.name
+        requirement_level:
+          conditionally_required: If memory type is `heap`.
+
+  - id: metric.veightjs.heap.space.size
+    type: metric
+    metric_name: v8js.heap.space.size
```
lmolkova, line 44: It seems to be a total size according to the brief, can we reflect it in the name?
E.g. 
```suggestion
    metric_name: v8js.heap.space.limit
```

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,96 @@
...
+    brief: "Total heap space size pre-allocated."
+    instrument: updowncounter
+    unit: "By"
+    stability: experimental
+    prefix: v8js.heap.space.size
```
lmolkova, line 49:
```suggestion
```

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,96 @@
...
+    attributes:
+      - ref: v8js.heap.space.name
+        requirement_level: recommended
+    note: >
+      Value can be retrieved from value `space_size` of [`v8.getHeapSpaceStatistics()`](https://nodejs.org/api/v8.html#v8getheapspacestatistics)
+
+  - id: metric.veightjs.heap.space.used_size
```
lmolkova, line 56: is my understanding correct that this metric would be the same as `v8js.memory.used{v8js.memory.type="heap"}` ?

If it is, do we need both? Otherwise, would it be possible to explain the difference in the brief/notes/markdown?

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,96 @@
...
+    type: metric
+    metric_name: v8js.heap.space.used_size
+    brief: "Heap space used size."
+    instrument: updowncounter
+    unit: "By"
+    stability: experimental
+    prefix: v8js.heap.space.used_size
```
lmolkova, line 63:
```suggestion
```

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,96 @@
...
+      Value can be retrieved from value `space_used_size` of [`v8.getHeapSpaceStatistics()`](https://nodejs.org/api/v8.html#v8getheapspacestatistics)
+
+  - id: metric.veightjs.heap.space.available_size
+    type: metric
+    metric_name: v8js.heap.space.available_size
+    brief: "Heap space available size."
+    instrument: updowncounter
+    unit: "By"
+    stability: experimental
+    prefix: v8js.heap.space.available_size
```
lmolkova, line 77:
```suggestion
```

(and in other places - prefix is not applicable to metrics, I'm surprised that tooling didn't complain)

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,96 @@
...
+    prefix: v8js.heap.space.used_size
+    attributes:
+      - ref: v8js.heap.space.name
+        requirement_level: recommended
+    note: >
+      Value can be retrieved from value `space_used_size` of [`v8.getHeapSpaceStatistics()`](https://nodejs.org/api/v8.html#v8getheapspacestatistics)
+
+  - id: metric.veightjs.heap.space.available_size
+    type: metric
+    metric_name: v8js.heap.space.available_size
```
lmolkova, line 72: it this metric derivable from other ones?  `v8js.heap.space.size` - `v8js.heap.space.used_size` ?

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,96 @@
...
+    prefix: v8js.heap.space.available_size
+    attributes:
+      - ref: v8js.heap.space.name
+        requirement_level: recommended
+    note: >
+      Value can be retrieved from value `space_available_size` of [`v8.getHeapSpaceStatistics()`](https://nodejs.org/api/v8.html#v8getheapspacestatistics)
+
+  - id: metric.veightjs.heap.space.physical_size
+    type: metric
+    metric_name: v8js.heap.space.physical_size
```
lmolkova, line 86: It is possible to add some explanation on how it's different from `v8js.heap.space.size|limit` ?

model/metrics/v8js-metrics.yaml
maryliag: Unfortunately it's not. 
The size is the value pre-allocated that can grow, and the used is the already allocated. The value of available is the total capacity - used, but the heap stats function from v8 doesn't return the capacity, just the allocated size. The value of capacity can probably be found in another place, but that would mean that I need to add a new metric to be able to calculate the available. But since the getHeapSpaceStatistics already returns the available, I'm using that directly

docs/runtime/v8js-metrics.md
```diff
@@ -0,0 +1,329 @@
...
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `v8js.memory.limit` | UpDownCounter | `By` | Total memory size pre-allocated. [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+
+**[1]:** If the memory type is `heap`, the value can be retrieved from value `space_size` of [`v8.getHeapSpaceStatistics()`](https://nodejs.org/api/v8.html#v8getheapspacestatistics)
```
pyohannes, line 98: Will the `v8js.memory.*` metrics used to report anything else but _heap_ memory usage and limits? It seems they are well-defined if the memory type is `heap`, but I wonder if those even make sense with a memory type of `non_heap`.

docs/runtime/v8js-metrics.md
```diff
@@ -0,0 +1,329 @@
...
+<!-- semconv metric.veightjs.gc.duration(metric_table) -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `v8js.gc.duration` | Histogram | `s` | Garbage collection duration. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
pyohannes, line 47: Could you add any recommendation/guidance of how to retrieve this value?

docs/runtime/v8js-metrics.md
maryliag: V8 has its memory space categorized into 2: heap and stack (non-heap). The heap is then divided into multiple spaces, this is why we can get more details into what type of heap is exactly, but the non-heap is grouped into one.

An example of their difference: The static values like number and string are pushed directly into the Stack memory space in order, while the object value is stored into Heap memory, and its Heap memory address is pushed into the Stack.

docs/runtime/v8js-metrics.md
maryliag: done

docs/runtime/v8js-metrics.md
pyohannes: Makes sense. However, I wonder more about whether this metric is also used to report `non_heap` memory limits.

If so, could you extend the footnote to also say something like.

> If the memory type is `non_heap`, the value can be retrieved from ...

Otherwise, if reporting a `non_heap` (stack) memory limit doesn't make sense, I'd rather not use an attribute for signifying that it's a heap limit, but put it into the name (e. g. `v8js.memory.heap.limit`).

maryliag: I decided to remove the non-heap metrics, since it's what makes more sense

joaopgrassi: Now that we only have the heap metrics, would it make sense that the attribute `v8js.heap.space.name` is changed to be required?

maryliag: done

pyohannes (approved):

MSNev (approved):

model/metrics/v8js-metrics.yaml
```diff
@@ -0,0 +1,69 @@
...
+    type: metric
+    metric_name: v8js.gc.duration
+    brief: "Garbage collection duration."
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    prefix: v8js.gc
+    attributes:
+      - ref: v8js.gc.type
+        requirement_level: recommended
```
lmolkova, line 12: nit:

are there cases when it's not available? If it's always available, would it make sense to make it required?

lmolkova (approved):

lmolkova:
@maryliag curious about CPU - is the plan to report `process.cpu.time` or follow up with JS-specific CPU metric?

model/metrics/v8js-metrics.yaml
maryliag: changed to required

maryliag:
> curious about CPU - is the plan to report `process.cpu.time`

yes, the plan is to use `process.cpu.time`

PR #1069, opened 2024-05-24
Update database span name:  clarify `target` and remove fallback

### LINKED ISSUE #1045: DB Span name format with alternating fields
Currently the [spec for db span names](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#name) defines the span name in the format of `{db.operation.name} {target}` with a specific list in prioritized order.

That works very well when within an instrumentation all spans have always the same attributes. However it's possible that some spans have the highest priority `{target}`, and some don't. That'd lead to traces where e.g. some have span names like `search Users` and others  with names like `search myDb:9200`. This is specifically an issue in the Elasticsearch instrumentation, but may be relevant to others as well.

We discussed this with @jack-berg and @trask in the last DB stabilisation working group.

A few potential resolutions to this problem that came up so far:

1) We let Elasticsearch define its own naming scheme, which will be similar to the general DB span names, but differ a bit. Specifically: it'd allow just setting `{db.operation.name}` as the span name, when `db.collection.name` is not present. This is relevant, because some queries may have a specific index name, but some target all (or none) indices, so there will be some spans with just the `{db.operation.name}`.
2) We force Elasticsearch to stick to the current naming schema potentially resulting in mixed span names when `db.operation.name` is not present.
3) We introduce a common `db.operation.name` value for cases when we know that the instrumentation in general would be able to capture it, but for the give query it can't set it - this would result in the same span name format. Question is, do we add this to the general DB spec, or to the DB specific one?

The list is open ended - there may be better solutions.

Above I focus on Elasticsearch, but similar issue can happen with other databases, so I feel we can generalise this. Somewhat related is the [HTTP span name](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#name) - although in its current form this is less relevant, but as the `{target}` list may grow in the future, similar issue can arise there.

lmolkova:
Fixes #1045

Alternative to #1061

## Changes

Clarifies that target depends on the operation and if target is not available, instrumentation should not fallback to other options.

E.g. for query like `SELECT * FROM (VALUES (1, 'a'), (2, 'b'), (3, 'c')) AS t (id, name)`  that uses anonymous table, the span name should be `SELECT` and not `SELECT {db.namespace}`.

### DIFF
--- a/docs/database/database-spans.md
+++ b/docs/database/database-spans.md
@@ -73,7 +73,7 @@ and SHOULD adhere to one of the following values, provided they are accessible:
 - `db.namespace` SHOULD be used only for operations on a specific database namespace.
 - `server.address:server.port` SHOULD be used for other operations not targeting any specific database(s) or collection(s)
 
-If a corresponding `{target}` value is not available for a specific operation, the instrumentation SHOULD NOT populate `{target}`.
+If a corresponding `{target}` value is not available for a specific operation, the instrumentation SHOULD omit the `{target}`.
 For example, for an operation describing SQL query on an anonymous table `SELECT * FROM (VALUES 'a', 'b', 'c') AS t (name)`, span name should be `SELECT`.
 
 ## Common attributes
--- a/docs/database/database-spans.md
+++ b/docs/database/database-spans.md
@@ -73,6 +73,9 @@ and SHOULD adhere to one of the following values, provided they are accessible:
 - `db.namespace` SHOULD be used only for operations on a specific database namespace.
 - `server.address:server.port` SHOULD be used for other operations not targeting any specific database(s) or collection(s)
 
+If a corresponding `{target}` value is not available for a specific operation, the instrumentation SHOULD NOT populate `{target}`.
+For example, for an operation describing SQL query on an anonymous table `SELECT * FROM (VALUES 'a', 'b', 'c') AS t (name)`, span name should be `SELECT`.
+
 ## Common attributes
 
 These attributes will usually be the same for all operations performed over the same database connection.
--- a/docs/database/database-spans.md
+++ b/docs/database/database-spans.md
@@ -56,13 +56,13 @@ Database spans MUST follow the overall [guidelines for span names](https://githu
 <!-- markdown-link-check-disable -->
 <!-- HTML anchors are not supported https://github.com/tcort/markdown-link-check/issues/225-->
 The **span name** SHOULD be `{db.operation.name} {target}` if there is a
-(low-cardinality) `db.operation.name` available (see below for the exact definition of the [`{target}`](#target-placeholder) placeholder).
+(low-cardinality) `{db.operation.name}` available (see below for the exact definition of the [`{target}`](#target-placeholder) placeholder).
 
 If there is no (low-cardinality) `db.operation.name` available, database span names
 SHOULD be [`{target}`](#target-placeholder).
 <!-- markdown-link-check-enable -->
 
-If neither `db.operation.name` nor `{target}` are available, span name SHOULD be `db.system`.
+If neither `{db.operation.name}` nor `{target}` are available, span name SHOULD be `{db.system}`.
 
 Semantic conventions for individual database systems MAY specify different span name format.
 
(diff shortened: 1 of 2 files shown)


docs/database/database-spans.md
```diff
@@ -62,14 +62,16 @@ If there is no (low-cardinality) `db.operation.name` available, database span na
 SHOULD be [`{target}`](#target-placeholder).
 <!-- markdown-link-check-enable -->
 
+If neither `db.operation.name` nor `{target}` are available, span name SHOULD be `db.system`.
+
 Semantic conventions for individual database systems MAY specify different span name format.
 
-The <span id="target-placeholder">`{target}`</span> SHOULD adhere to one of the following values, arranged in prioritized order, provided they are accessible:
+The <span id="target-placeholder">`{target}`</span> SHOULD describe the entity operation is performed against
```
trask, line 69:
```suggestion
The <span id="target-placeholder">`{target}`</span> SHOULD describe the entity that the operation is performed against
```
```suggestion
The <span id="target-placeholder">`{target}`</span> SHOULD describe the entity operation is performed against
```

docs/database/database-spans.md
```diff
@@ -62,14 +62,16 @@ If there is no (low-cardinality) `db.operation.name` available, database span na
...
+and SHOULD adhere to one of the following values, provided they are accessible:
 
-- `db.collection.name`
-- `db.namespace`
-- `server.address:server.port`
-- `db.system`
+- `db.collection.name` SHOULD be used for data manipulation operations or operations on database collections.
+- `db.namespace` SHOULD be used only for operations on a specific database namespace.
+- `server.address:server.port` SHOULD be used for other operations not targeting a specific database or collection
```
trask, line 74: (trying to avoid instrumentation authors thinking that they should use this if operations target multiple databases or multiple collections as opposed to a single specific one)

```suggestion
- `server.address:server.port` SHOULD be used for other operations not targeting any specific database(s) or collection(s)
```

gregkalapos (commented):
Nice 👍 


I'd maybe add 2 things here:
- The example in the PR description with the anonymous table is very useful, I think that could be part of the spec.
- The goal here is to avoid falling back to a less useful `target`, (that's also part of the PR description here). So my understanding is, we are ok to omit `target` in edge cases. But it's not really written into the spec and this PR doesn't really says that explicitly. I think it'd be worth adding it as well. 

I had this in https://github.com/open-telemetry/semantic-conventions/pull/1061: 

```
A given instrumentation SHOULD strive for consistency in span names.


If certain queries lack this specific attribute, the instrumentation SHOULD use `{db.operation.name}` as the span name and omit `{target}`, 
rather than falling back to the next attribute in the [`{target}`](#target-placeholder) list.
```

(dropped the part which is different in this PR). Do we want to maybe add that part?

lmolkova:
@gregkalapos I updated the doc with your suggestion, lmk what you think

gregkalapos (approved):
> @gregkalapos I updated the doc with your suggestion, lmk what you think

Thanks, lgtm.

docs/database/database-spans.md
```diff
@@ -62,14 +62,19 @@ If there is no (low-cardinality) `db.operation.name` available, database span na
...
+
+- `db.collection.name` SHOULD be used for data manipulation operations or operations on database collections.
+- `db.namespace` SHOULD be used only for operations on a specific database namespace.
+- `server.address:server.port` SHOULD be used for other operations not targeting any specific database(s) or collection(s)
 
-- `db.collection.name`
-- `db.namespace`
-- `server.address:server.port`
-- `db.system`
+If a corresponding `{target}` value is not available for a specific operation, the instrumentation SHOULD NOT populate `{target}`.
```
trask, line 76:
```suggestion
If a corresponding `{target}` value is not available for a specific operation, the instrumentation SHOULD omit the `{target}`.
```

docs/database/database-spans.md
```diff
@@ -62,14 +62,19 @@ If there is no (low-cardinality) `db.operation.name` available, database span na
 SHOULD be [`{target}`](#target-placeholder).
 <!-- markdown-link-check-enable -->
 
+If neither `db.operation.name` nor `{target}` are available, span name SHOULD be `db.system`.
```
trask, line 65: (maybe?)
```suggestion
If neither `db.operation.name` nor `{target}` are available, span name SHOULD be `{db.system}`.
```

trask (approved):

joaopgrassi (approved):

PR #1072, opened 2024-05-24
Add support for database batch operations

### LINKED ISSUE #712: DB Convention does not cover batch/multi/envelope operations
Our API has a small alphabet of relatively simple operations key-value operations (`get`, `put`, `delete`, &c.) For these, the operation names seem clear. We also have a set of operations that support bulk/batching of operations. these can be homogeneous or heterogeneous. For example, `batch` can accept a set of any combination of `get`, `put`, `delete`, &c. We also support a generic system for server-side compare-and-mutate, where some predicate based on a query over existing data is provided, and when the predicate returns true, some operation is applied — that operation can be a simple or a batch operation. for these collections of heterogenous operations, how should be annotate the span?

trask:
Fixes #712

## Changes

Adds support for database batch operations.


model/registry/db.yaml
```diff
@@ -10,10 +10,18 @@ groups:
...
-            It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.
+          If the collection name is parsed from the query text, it SHOULD be the first collection name found in the query
+          and it SHOULD match the value provided in the query text including any schema and database name prefix.
+
+          For [homogeneous batch operations](/docs/database/database-spans.md#homogeneous-batches), since individual operations will all have the same collection name,
+          that collection name SHOULD be used.
+          For [heterogeneous batch operations](/docs/database/database-spans.md#heterogeneous-batches) where individual operations are known to all have the same collection name,
+          that collection name SHOULD be used.
+          For [heterogeneous batch operations](/docs/database/database-spans.md#heterogeneous-batches) where individual operations are not known to all have the same collection name,
+          `db.collection.name` SHOULD NOT be captured.
```
trask, line 24: can simplify: if known to have the same collection name use that, otherwise don't capture

model/registry/db.yaml
```diff
@@ -38,12 +46,29 @@ groups:
...
+
+          For [homogeneous batch operations](/docs/database/database-spans.md#homogeneous-batches),
+          individual operations will all have the same operation name and so that operation name SHOULD be used,
+          prepended by `BATCH `.
+          For [heterogeneous batch operations](/docs/database/database-spans.md#heterogeneous-batches)
+          where individual operations are known to all have the same operation name, that operation name SHOULD be used,
+          prepended by `BATCH `.
+          For [heterogeneous batch operations](/docs/database/database-spans.md#heterogeneous-batches)
+          where individual operations are not known to all have the same operation name,
+          `db.operation.name` SHOULD be `BATCH` or some other database system specific term if more applicable.
```
trask, line 58: simplify: if known to have the same operation name, prepend by `BATCH `, otherwise capture `BATCH`

model/registry/db.yaml
```diff
@@ -38,12 +46,29 @@ groups:
...
         stability: experimental
         brief: >
           The database query being executed.
+        note: >
+          For [homogeneous batch operations](/docs/database/database-spans.md#homogeneous-batches),
+          individual operations will all have the same query text and so that query text SHOULD be used.
+          For [heterogeneous batch operations](/docs/database/database-spans.md#heterogeneous-batches),
+          individual operations will not all have the same query text,
+          and all of the individual query texts SHOULD be concatenated with separator `; `
+          or some other database system specific separator if more applicable.
```
trask, line 71: simplify: if known to have the same query text use that, otherwise concatenate

maryliag (approved):

model/registry/db.yaml
```diff
@@ -10,10 +10,14 @@ groups:
         stability: experimental
         brief: The name of a collection (table, container) within the database.
         note: >
-            If the collection name is parsed from the query, it SHOULD match the value provided
-            in the query and may be qualified with the schema and database name.
+          If the collection name is parsed from the query, it SHOULD match the value provided
```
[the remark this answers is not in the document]
trask: I think this got cleaned up in #1070, ptal

jack-berg (approved):

PR #1074, opened 2024-05-25
Add event.summary as a recommendation for events

alexvanboxel:
## Changes

Events are identified by an `event.name` and attributes and fields in the body that carry specific meaning. However, since these events will be combined with other logs, `event.summary` allows a backend to display a human-readable representation of the event.


alexvanboxel:
Proposed PR for issue https://github.com/open-telemetry/semantic-conventions/issues/1076

As discussed in the Eventing SIG on 2024-05-24 cc: @MSNev @trask @JamieDanielson

trisch-me (approved):

model/registry/event.yaml
```diff
@@ -16,3 +16,13 @@ groups:
           separation of semantics for events in separate domains like browser, mobile, and
           kubernetes.
         examples: ['browser.mouse.click', 'device.app.lifecycle']
+      - id: message
```
trask, line 19: maybe `event.description`?

alexvanboxel: I took `message` as it's a convention that already exists... but granted it could be confusing with messaging.: `description` is not bad idea.

model/registry/event.yaml
```diff
@@ -16,3 +16,13 @@ groups:
...
+        type: string
+        stability: experimental
+        brief: >
+          Human-readable message representing an event.
+        note: >
+          Events are identified by an `event.name` and a set of attributes and fields in the body that carry specific
+          meaning. However, since these events will be combined with other logs, `event.message` allows a backend to
+          display a human-readable representation of the event.
```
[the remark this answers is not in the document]
alexvanboxel: I'll try to add more context and align the examples with the examples from the name

breedx-splk (dismissed):
Respectfully, I don't think we should do this.

> Events are identified by an event.name and attributes and fields in the body that carry specific meaning. 

Yes. Events have defined semantics and attributes are spec'd to convey those semantics.

> However, since these events will be combined with other logs, event.message allows a backend to display a human-readable representation of the event.

The fact that Events happen to be an extension of Logs should be irrelevant here. Users and consumers of Events should not be concerned with the underlying transport mechanism, and there are APIs to help treat events as a legitimate, separate signal type.

Why are we trying to allow a backend to display something human readable? Backends don't typically display much anyway. It sounds like this use case is already covered by structured logging, and I don't think we should muddy Events with a `message` field whose human-readable content could be misleading or otherwise conflict the semantics of the event itself.

alexvanboxel:
> Why are we trying to allow a backend to display something human readable? Backends don't typically display much anyway. It sounds like this use case is already covered by structured logging, and I don't think we should muddy Events with a message field whose human-readable content could be misleading or otherwise conflict the semantics of the event itself.

Maybe the definition of backend is not clear and could be better specified. I meant as backend logging backends (like DataDog, Google Logging, Elastic, ... those are the systems I have experience with), because they display the logs to the user: For us (as an end-user) it's important to mix our normal logs without events in the same stream. We use the convention of `message` in our structured logs, so when our teams do exploratory work they don't see blurps of JSON inter-mingeled with normal logs lines.

We defined our internal business event based ontop of structured logs with success (as we mandate the `message` field in the body). Imagine the backlash we'll get that when we upgrade to OTel events that the experience for the user degrades. That will not help adoption.

alexvanboxel:
> Respectfully, I don't think we should do this.
> 
> > Events are identified by an event.name and attributes and fields in the body that carry specific meaning.
> 
> Yes. Events have defined semantics and attributes are spec'd to convey those semantics.
> 
> > However, since these events will be combined with other logs, event.message allows a backend to display a human-readable representation of the event.
> 
> The fact that Events happen to be an extension of Logs should be irrelevant here. Users and consumers of Events should not be concerned with the underlying transport mechanism, and there are APIs to help treat events as a legitimate, separate signal type.
> 
> Why are we trying to allow a backend to display something human readable? Backends don't typically display much anyway. It sounds like this use case is already covered by structured logging, and I don't think we should muddy Events with a `message` field whose human-readable content could be misleading or otherwise conflict the semantics of the event itself.

Maybe the definition of backend is not clear and could be better specified. I meant as backend logging backends (like DataDog, Google Logging, Elastic, ... those are the systems I have experience with), because they display the logs to the user: For us (as an end-user) it's important to mix our normal logs without events in the same stream. We use the convention of `message` in our structured logs, so when our teams do exploratory work they don't see blurps of JSON inter-mingeled with normal logs lines.

We defined our internal business event based ontop of structured logs with success (as we mandate the `message` field in the body). Imagine the backlash we'll get that when we upgrade to OTel events that the experience for the user degrades. That will not help adoption.

alexvanboxel:
> We defined our internal business event based ontop of structured logs with success (as we mandate the `message` field in the body). Imagine the backlash we'll get that when we upgrade to OTel events that the experience for the user degrades. That will not help adoption.

Maybe an extra note: I'm afraid that if we cannot leverage the fact that logs are also meant for human consumption, for events, my feeling on the fact that events are built on top of logs will flip to unsupportive. Then, it would probably have been better to make logs their own signal. At least then, if that signal had a message/description in the proto, at least a log-backend could also consume those signals and do something with it.

alexvanboxel:
@MSNev I've removed the workflow started example, and added examples parallel to the examples in `event.name`
@trask I've changed the PR to change `event.message` -> `event.description` (It very much aligns with how a metric is defined, it also has a name and description). Some systems may display the description, others not.

breedx-splk:
I think I could be talked into this if there was wording to say something like "If the `event.name` for the event is well known and the structure and semantics of the event are described by OpenTelemetry, then the description field SHOULD NOT be populated."

docs/general/events.md
```diff
@@ -69,6 +72,8 @@ structure and semantics will also be defined.
   and the semantic conventions will define the expected structure of the _payload_
   (data) for the event.
 * The _payload_ (data) SHOULD be used to represent the structure of the event.
+* The event MAY have an `event.message` attribute that is a human-readable
```
JamieDanielson, line 75:
```suggestion
* The event MAY have an `event.description` attribute that is a human-readable
```
missed in the rename I think?

model/registry/event.yaml
```diff
@@ -16,3 +16,13 @@ groups:
           separation of semantics for events in separate domains like browser, mobile, and
           kubernetes.
         examples: ['browser.mouse.click', 'device.app.lifecycle']
+      - id: description
+        type: string
+        stability: experimental
+        brief: >
+          Human-readable message representing an event.
```
JamieDanielson, line 23:
```suggestion
          Human-readable description representing an event.
```
another missed rename I think?

trask:
one potential concern with the name `event.description` is that it's not quite analogous to metric "description" since metric description is a constant value as opposed to this new proposal which varies per event instance

alexvanboxel:
As asked in the Eventing SIG, I'll add a few redacted events with its message/summary/description:

**Event**: import:dgc_asset_import
**Summary**: Exported 1057 nodes from 166 roots in 0.051s

```json
{
  "message": "Exported 1057 nodes from 166 roots in 0.051s",
  "event_class": "generic",
  "event_name": "export:dgc_asset_export",
  "root_nodes_count": 166,
  "authenticated_id": "REDACTED",
  "duration": 51,
  "output_format": "JSON",
  "root_resource_type": "terms",
  "c4_container": "backend",
  "max_page_nodes_count": 1057,
  "send_notification": false,
  "query_limit": 200,
  "resource_types": [
    "vocabulary",
    "terms",
    "attributes",
    "source",
    "type",
    "status",
    "target"
  ],
  "timestamp": "2024-06-07T19:11:44.563Z",
  "data_size": 404475,
  "trace_id": "8853d755645a95f3809341f2c41ed92c",
  "output_type": "TEXT",
  "all_nodes_count": 1057,
  "span_id": "bcad170d7e71c0d0",
  "count_limit": 10001,
  "query_type": "TABLE_VIEW_CONFIG",
  "validation_enabled": false,
  "request_source": "JAVA_API",
  "valid_query": true,
  "c4_system": "knowledgegraph",
  "status": "INFO"
}
```

**Event**: import:dgc_asset_import
**Summary**: Asset import completed with result: SUCCESS

```json
{
  "message": "Asset import completed with result: SUCCESS",
  "operation_type": "IMPORT",
  "community_by_mapping_count": 0,
  "simulation": false,
  "resources_updated": {
    "MP": 5
  },
  "job_state": "COMPLETED",
  "authenticated_id": "REDACTED",
  "duration": 25,
  "community_by_name_count": 0,
  "domain_by_id_count": 0,
  "asset_by_name_count": 0,
  "domain_by_mapping_count": 0,
  "file_type": "JSON",
  "job_result": "SUCCESS",
  "send_notification": false,
  "continue_on_error": false,
  "timestamp": "2024-06-07T19:11:44.568Z",
  "commands_executed": 0,
  "trace_id": "bcdfe64836f084dac3edd79b33546fd4",
  "span_id": "6d3936b98a8181e4",
  "community_by_id_count": 0,
  "domain_by_name_count": 0,
  "asset_by_mapping_count": 5,
  "request_source": "REST_API",
  "commands_skipped": 0,
  "save_result": false,
  "asset_by_id_count": 0,
  "service": "dgc",
  "job_id": "REDACTED",
  "event_class": "generic",
  "event_name": "import:dgc_asset_import",
  "commands_count_in_request": 7,
  "errors_count": 0,
  "status": "INFO"
}
```

**Event**: catalogjdbc:edge_jdbc_connect
**Summary**: connection to the datasource has been established

```json
{
  "message": "connection to the datasource has been established",
  "database_name": "REDACTED",
  "timing": {
    "total_calls": 2,
    "time_unit": "MILLISECONDS",
    "max_time": 16994,
    "mean_time": 8606,
    "total_time": 17212
  },
  "schema_connection_id": "REDACTED",
  "schema_name": "REDACTED",
  "driver_class_name": "REDACTED",
  "duration": 17212,
  "tmp_msg": "",
  "database_id": "REDACTED",
  "engine": "JAVA_API",
  "service": "jdbc-ingestion",
  "schema_id": "REDACTED",
  "event_name": "catalogjdbc:edge_jdbc_connect",
  "event_class": "generic",
  "correlation_id": "REACTED",
  "status": "INFO",
  "timestamp": "2024-06-07T19:27:10.826Z"
}
```

breedx-splk:
Thanks for the examples, @alexvanboxel . I think these do a good job of demonstrating what these are for, and I think they also demonstrate:

1. The data in the message is redundant with the data contained in the event body. In the first example, node count, root count, and time are all fields within the event. In the second example, the message effectively restates the `job_state` (COMPLETED) and `job_result` (SUCCESS). In the third example, the message seems to be explaining the semantic behind the event name itself. I definitely understand that this can be useful to a human operator who might otherwise be unfamiliar with the event structure....however, I think in practice this is better solved by having schema-aware (event-specific) `toString()` type implementations. Because...
2. The free-form nature of the message field opens itself up to inaccurate or misleading information! And because the field is just  any text, it is impossible to constrain/limit the contents via conventions or specification. For example, there's nothing to prevent the message in the second example from reading `"Asset import failed with result: MEGAFAILURE"`, even though this directly conflicts with the true event content.

I will remove my blocking request, because I think there isn't too much harm in this, but unfortunately I cannot offer a supportive approval. I ask that other reviewers please consider some phrasing that discourages the use of the message/description/summary for well-specified otel events. 

I also still think that an attribute is a reasonable place to put a summary of the event, and we don't need it to be first-class Event field.

alexvanboxel:
> Thanks for the examples, @alexvanboxel . I think these do a good job of demonstrating what these are for, and I think they also demonstrate:
> 
> 1. The data in the message is redundant with the data contained in the event body. In the first example, node count, root count, and time are all fields within the event. In the second example, the message effectively restates the `job_state` (COMPLETED) and `job_result` (SUCCESS). In the third example, the message seems to be explaining the semantic behind the event name itself. I definitely understand that this can be useful to a human operator who might otherwise be unfamiliar with the event structure....however, I think in practice this is better solved by having schema-aware (event-specific) `toString()` type implementations. Because...
> 2. The free-form nature of the message field opens itself up to inaccurate or misleading information! And because the field is just  any text, it is impossible to constrain/limit the contents via conventions or specification. For example, there's nothing to prevent the message in the second example from reading `"Asset import failed with result: MEGAFAILURE"`, even though this directly conflicts with the true event content.
> 
> I will remove my blocking request, because I think there isn't too much harm in this, but unfortunately I cannot offer a supportive approval. I ask that other reviewers please consider some phrasing that discourages the use of the message/description/summary for well-specified otel events.
> 
> I also still think that an attribute is a reasonable place to put a summary of the event, and we don't need it to be first-class Event field.

Thanks for the feedback. I've added a section that I hope formulates your concerns. I've also renamed everything to `summary` as this is till now the best name that was suggested.

trask:
> `event.summary` allows a backend to display a human-readable representation of the event

does the same issue exist for how to display a human-readable representation of a structured log?

tedsuo:
Had a discussion about the `summary` attribute today in the Event SIG. It's a very useful field, Alex's example above (https://github.com/open-telemetry/semantic-conventions/pull/1074#issuecomment-2155453445) really shows that.

One question is whether the content of the summary attribute should be strictly defined as part of the semantic convention for the event. We (Me, Nev, Trask) feel that the value should not be strictly defined. Like `severity`, the event summary is context specific: whatever message the person writing the event wants to communicate to the person reading the event is fine.

For example, a commonly defined database event in a particular library may want to include a library-specific summary. An end user tracking down a production problem may flag a particular failure or exception as suspicious and mention some context specific information in the summary. 

Because the content of this attribute is not intended to be indexed and used as part of machine analysis, strictly defining its content is not necessary. This is different from the Span `name` field, which is used both as a primary index and a UI label and thus ends up creating a lot of fuss about its structure and content. I appreciate having the concept of a summary field that has no purpose other than to deliver a short message from one human to another.

Personally, I would prefer `summary` to be a proto field, not an attribute, in order to keep it completely separate from the attributes which are intended to be used as indexes. But I'm not going to die on that hill, I don't think that there is real harm in keeping it as an attribute. :)

model/registry/event.yaml
```diff
@@ -16,3 +16,18 @@ groups:
...
+      - id: summary
+        type: string
+        stability: experimental
+        brief: >
+          Human-readable summary representing an event.
+        note: >
+          Events are identified by an `event.name` and a set of attributes and fields in the body that carry specific
+          meaning. However, since these events will be combined with other logs, `event.summary` allows a
+          centralized logging system to display a human-readable representation of the event.
+
```
lmolkova, line 28: This paragraph does not seem to provide new information - it's already captured in the event name, the markdown, and the brief. Could it be removed?

```suggestion
```

model/registry/event.yaml
```diff
@@ -16,3 +16,18 @@ groups:
...
+          When summaries are generated, they are not expected to include every attribute and field that is part of the
+          event but could contain those that are meaningful for a human operator when visually navigating a centralized
+          log system. Instrumentation libraries that produce events defined within the standard OpenTelemetry are not
+          expected to add an `event.summary` attribute, as these are well-known events.
```
lmolkova, line 32: Could we make it normative? E.g.
```suggestion
          It's NOT RECOMMENDED to record full string representation of the event on the `event.summary` attribute. The summary SHOULD be concise and human-readable as it's expected to be used by a human operator when visually navigating a centralized log system.
```

I removed 

> Instrumentation libraries that produce events defined within the standard 
OpenTelemetry are not expected to add an `event.summary` attribute, as these are well-known events.

I'm not sure how it could work - if we have 100+ events defined, no backend would be able to special-case each event and define human-readable representation of it. I.e. I don't understand why we would allow to define a summary for at least some standards events.

thompson-tomo:
I would also agree to not making it strongly typed and instead would propose an attribute for the outcome as suggested in #

MSNev:
@alexvanboxel Do you want this reopened?

alexvanboxel:
> @alexvanboxel Do you want this reopened?

yes, please. I've been on holiday and couldn't (we'll it holiday) pick this up. I can pick it up again.

JamieDanielson:
In looking at this again, my understanding of the general consensus so far is that having `event.summary` as an optional property on events is fine, and in some cases especially useful for human operators to get a succinct summary of valuable details within the event. The open questions I can see are:

1. Should this be a top-level field as it is here, or should it be an attribute?
2. What is the expectation for "well-known" events in OTel? Should summary be optional, required, or "disallowed"?

Is there anything else I am missing that needs to be resolved for this?

.chloggen/event-message.yaml
```diff
@@ -0,0 +1,25 @@
...
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [ 1076 ]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Events are identified by an `event.name` and a set of attributes and fields in the body that carry specific
+  meaning. However, since these events will be combined with other logs, `event.description` allows a
```
thompson-tomo, line 24: Should be event.summary

thompson-tomo (changes requested):

Issue #522, opened 2023-11-14
Remove `shared` from `system.memory.state` values?


trask:
Should `shared` be removed from [`system.memory.state`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemmemoryusage) values?

I'm reviewing the [`usage` naming convention](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#instrument-naming):

> Where appropriate, the sum of **usage** over all attribute values SHOULD be equal to the **limit**.

and at least on my system, the `shared` dimensions appears to be double counted:

```
root@972900a4c455:/# free
               total        used        free      shared  buff/cache   available
Mem:        32847132     2214804    30020716       80216      611612    30172244
```

doing the math:

```
32847132 = 2214804 + 30020716 + 611612
```

cc @open-telemetry/semconv-system-approvers

mx-psi:
Discussed on January 18th System Semantic Conventions WG meeting, similar to #531 this blocks stability

ChrsMark:
Thank's @braydonk! 

@open-telemetry/semconv-system-approvers any thoughts about this?

PR #409, opened 2023-10-16
Clarify how to report the total amount of memory

mx-psi:
Updates #127

## Changes

- Removes `total` from list of well-known values for `system.memory.state`
- Adds note to `system.memory.usage` clarifying that adding over `system.memory.state` SHOULD equal the total amount of memory available in the system.
- Adds opt-in `system.memory.limit` metric that reports the total amount of memory available in the system.

This tries to follow the guidance on https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#instrument-naming:

> an instrument that measures the constant, known total amount of something should be called entity.limit. For example, `system.memory.limit` for the total amount of memory on a system.

and 

> Where appropriate, the sum of usage over all attribute values SHOULD be equal to the limit.

The `total` value was added on #89 but was not listed before the refactoring. This rewording avoids an inconsistency with the general guidelines and adds a metric that may be useful in some implementations given that total is reported separately by certain OSs (e.g. Linux).
 
## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x] [CHANGELOG.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md) updated for non-trivial changes.
* [x] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

Issue #50, opened 2023-05-24
Should metric namespaces be restricted from also being metric names?


trask:
Summarizing reasons for [not introducing this restriction](https://github.com/open-telemetry/opentelemetry-specification/pull/3477):

* allows more flexibility in naming metrics, e.g.
  * allows dropping `.count` suffixes for more concise metric names, e.g. `system.processes` instead of `system.processes.count`

Summarizing reasons for [introducing this restriction](https://github.com/open-telemetry/opentelemetry-specification/pull/3476):

* [makes metrics documentation simpler for both author and reader](https://github.com/open-telemetry/opentelemetry-specification/pull/3476#pullrequestreview-1418919384)
* consistency with same restriction on attribute namespaces
  * (the reason for attribute name restrictions was to leave open the possibility for attribute names to be represented as nested maps, [simlar to ECS](https://www.elastic.co/guide/en/elasticsearch/reference/8.7/object.html))
* [allows converting bidirectionally between our dot-notation and common structured data formats such as json or yaml](https://github.com/open-telemetry/opentelemetry-specification/pull/3476#issuecomment-1559704071)

I will update these summaries based on any comments made below.

trask:
If "Option C" is chosen in https://github.com/open-telemetry/semantic-conventions/issues/51#issuecomment-1561642471, I think that this reason not introducing the restriction will no longer be relevant:

* allows dropping `.count` suffixes for more concise metric names, e.g. `system.processes` instead of `system.processes.count`

because `system.processes.count` has attribute `state`, which would implicitly become `system.processes.state`, while if the metric was `system.processes`, the attribute `state` would implicitly become `system.state`.

reyang:
The TC voted and here is the result https://github.com/open-telemetry/semantic-conventions/issues/51#issuecomment-1607892958.

trask:
@reyang this issue is about namespaces for instrument names as opposed to namespaces for attribute names, did the TC vote also on this? thx

Oberon00:
To me this looks like a mix-up. Re-opening.

trask:
On second look, maybe the decision has been made(?), based on https://github.com/open-telemetry/semantic-conventions/issues/50#issuecomment-1564648535

> If "Option C" is chosen in https://github.com/open-telemetry/semantic-conventions/issues/51#issuecomment-1561642471, I think that this reason not introducing the restriction will no longer be relevant:
>
[...]

PR #1078, opened 2024-05-27
add system.linux.memory.slab to system metrics

rogercoll:
Fixes  https://github.com/open-telemetry/semantic-conventions/issues/531

Related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7417

## Changes

Adds the `slab` memory metric that is only available on Linux systems as `system.linux.memory.slab`. 


rogercoll:
For the attributes, it references the `system.memory.state` which will **only** match the `total` well-known value.
`reclaimable` and `unreclaimable` are additional custom state attributes normally associated with slab memory. For example, the current (0.101.0) version of the opentelemetry-collector-contrib uses `system.memory.slab_reclaimable` and `system.memory.slab_unreclaimable` to indicate those states. 

Should this metric reference the `system.memory.state` attributes or create a new registry (e.g `system.linux.memory.slab.state`) that explicitly indicates the `total`, `reclaimable` and `unreclaimable` states?

ChrsMark:
Thank's @rogercoll for this PR!

> Should this metric reference the system.memory.state attributes or create a new registry (e.g system.linux.memory.slab.state) that explicitly indicates the total, reclaimable and unreclaimable states?

I think it makes sense to have a specific `*.state` attribute which would include the `reclaimable` and `unreclaimable`.
As it was pointed out at https://github.com/open-telemetry/semantic-conventions/issues/522 (from the [usage naming convention](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#instrument-naming)):
> Where appropriate, the sum of usage over all attribute values SHOULD be equal to the limit.

mx-psi:
Agreed with @ChrsMark!

docs/system/system-metrics.md
```diff
@@ -1466,6 +1467,25 @@ See also `MemAvailable` in [/proc/meminfo](https://man7.org/linux/man-pages/man5
...
 
 
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `system.linux.memory.slab`
+
+<!-- semconv metric.system.linux.memory.slab(metric_table) -->
```
trisch-me, line 1477: as your metric has attributes they should be noted here i.e. add after metric table
```
<!-- semconv metric.system.linux.memory.slab(full) -->
```

braydonk:
I agree with adding a new attribute for this, and even though the best name for that attribute would be `system.linux.memory.slab.state`, that name may be restricted by #50 (a metric name can't also be a namespace). I'm not sure what other name to give this though, especially because it has to go in the registry when I have a hard time imagining it will ever end up on another metric.

Not sure what to recommend. I'll check in on #50 again to get clarification that this restriction is definitely in place; it's hard to tell from the discussion history exactly where it ended up.

rogercoll:
> I think it makes sense to have a specific *.state attribute which would include the reclaimable and unreclaimable.

Sounds good to me. To be aligned with the `system.memory` model, I think the following could work:

```
Metric name: system.linux.memory.slab.usage       
Attributes: system.linux.memory.slab.state (reclaimable and unreclaimable)
```

```
Metric name: system.linux.memory.slab.limit 
```

model/metrics/system-metrics.yaml
```diff
@@ -340,3 +340,26 @@ groups:
...
+    attributes:
+      - ref: system.linux.memory.slab.state
+
+  - id: metric.system.linux.memory.slab.limit
+    type: metric
+    metric_name: system.linux.memory.slab.limit
+    stability: experimental
+    brief: "Total slab memory used in the system."
+    note: |
+      Its value SHOULD equal the sum of `system.linux.memory.slab.state` over all states.
```
ChrsMark, line 362: I see we have the same wording for the `system.memory.limit` but something bugs me here, since we don't sum the `state` but the `usage` 🤔 .

Shouldn't that be
```suggestion
      Its value SHOULD equal the sum of `system.linux.memory.slab.usage` over all states.
```
?

rogercoll: Good catch! Docs regenerated in https://github.com/open-telemetry/semantic-conventions/pull/1078/commits/b03736c7b84ea3243f8ebe22f758e426c57c3181

ChrsMark (commented):
From SemConv perspective looks good to me.

One question: do we actually plan to implement the `system.linux.memory.slab.limit` as well in the Collector?

rogercoll:
> One question: do we actually plan to implement the `system.linux.memory.slab.limit` as well in the Collector?

I would say so, the current sum of all `system.memory.usage` [states](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/memoryscraper/metadata.yaml#L12) do not equal to the `limit` which is needed for v1. If this PR moves forward, I can open an issue in the collector to track the implementation.

cc @dmitryax

dmitryax (approved):
LGTM

docs/system/system-metrics.md
ChrsMark: I guess this can be resolved now.

ChrsMark (approved):

mx-psi (approved):

docs/system/system-metrics.md
```diff
@@ -1466,6 +1468,77 @@ See also `MemAvailable` in [/proc/meminfo](https://man7.org/linux/man-pages/man5
...
 
 
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `system.linux.memory.slab.usage`
+
+<!-- semconv metric.system.linux.memory.slab.usage(metric_table) -->
```
lmolkova, line 1478: Could you please document [metric requirement level](https://github.com/open-telemetry/semantic-conventions/blob/c80cd1fb36f21eae0321ea1e765d7ae9d036b6d6/docs/general/metric-requirement-level.md#metric-requirement-levels)?

We usually add some form of "This metric is [required](/docs/general/metric-requirement-level.md#required)." beofre metric definition. 

It'll be very helpful when we yamlify metric requirement levels.

model/registry/system.yaml
```diff
@@ -254,3 +254,23 @@ groups:
...
+  # system.linux.* attribute group
+  # system.linux.memory.slab* attribute group
+  - id: registry.system.linux.memory.slab
+    prefix: system.linux.memory.slab
+    type: attribute_group
+    brief: "Describes Linux Slab Memory attributes"
+    attributes:
+      - id: state
+        type:
+          allow_custom_values: false
```
lmolkova, line 266: nit:

```suggestion
```

this value is ignored and will be removed https://github.com/open-telemetry/build-tools/issues/294

model/metrics/system-metrics.yaml
```diff
@@ -340,3 +340,26 @@ groups:
...
+    attributes:
+      - ref: system.linux.memory.slab.state
+
+  - id: metric.system.linux.memory.slab.limit
+    type: metric
+    metric_name: system.linux.memory.slab.limit
+    stability: experimental
+    brief: "Total slab memory used in the system."
+    note: |
+      Its value SHOULD equal the sum of `system.linux.memory.slab.usage` over all states.
```
lmolkova, line 362:
```suggestion
      Its value SHOULD be equal to the sum of `system.linux.memory.slab.usage` over all states.
```

model/metrics/system-metrics.yaml
```diff
@@ -340,3 +340,26 @@ groups:
...
     unit: "By"
     attributes: []
+
+  - id: metric.system.linux.memory.slab.usage
+    type: metric
+    metric_name: system.linux.memory.slab.usage
+    stability: experimental
+    brief: "Reports the memory used by the Linux kernel for managing caches of frequently used objects by state."
+    note: |
+      The sum over all `system.linux.memory.slab.usage` state values SHOULD equal the total slab memory available on the system, that is `system.linux.memory.slab.limit`.
```
lmolkova, line 350:
```suggestion
      The sum over all `system.linux.memory.slab.usage` state values SHOULD be equal to the total slab memory available on the system, that is `system.linux.memory.slab.limit`.
```

model/metrics/system-metrics.yaml
```diff
@@ -340,3 +340,26 @@ groups:
     instrument: updowncounter
     unit: "By"
     attributes: []
+
+  - id: metric.system.linux.memory.slab.usage
+    type: metric
+    metric_name: system.linux.memory.slab.usage
+    stability: experimental
+    brief: "Reports the memory used by the Linux kernel for managing caches of frequently used objects by state."
```
lmolkova, line 348: maybe?
```suggestion
    brief: "Reports the memory used by the Linux kernel for managing caches of frequently used objects."
```

model/metrics/system-metrics.yaml
```diff
@@ -340,3 +340,26 @@ groups:
...
+    instrument: updowncounter
+    unit: "By"
+    attributes:
+      - ref: system.linux.memory.slab.state
+
+  - id: metric.system.linux.memory.slab.limit
+    type: metric
+    metric_name: system.linux.memory.slab.limit
+    stability: experimental
+    brief: "Total slab memory used in the system."
```
lmolkova, line 360: Is it a limit or a total used? 

Limit should be a constant

https://github.com/open-telemetry/semantic-conventions/blob/c80cd1fb36f21eae0321ea1e765d7ae9d036b6d6/docs/general/metrics.md?plain=1#L174-L176

model/metrics/system-metrics.yaml
rogercoll: Yes, more readable and the state attributes are already included in the generated docs. Thanks!

docs/system/system-metrics.md
rogercoll: Makes sense to me, added in https://github.com/open-telemetry/semantic-conventions/pull/1078/commits/bd3fd25411f68d5929a590cd294661d911110143

model/metrics/system-metrics.yaml
rogercoll: Good point. Actually, it is not a constant value, it can grow/shrink over time depending on the used resources. From the shared docs, it seems that we should not be using "limit" for non-constant values, and consequently neither "usage".  As defined in the docs, "total" [should not be](https://github.com/open-telemetry/semantic-conventions/blob/c80cd1fb36f21eae0321ea1e765d7ae9d036b6d6/docs/general/metrics.md?plain=1#L154) used neither.

That makes me wonder if we should rollback the "usage" approach and remove the "limit" metric. I think having a "slab" metric with specific *.state attributes might be the most versatile approach:
```
Metric name: system.linux.memory.slab       
Attributes: system.linux.memory.slab.state (recommended attributes: reclaimable and unreclaimable)
```
Wdyt? cc @open-telemetry/semconv-system-approvers

ChrsMark: In that case I agree that the `.limit` one can be skipped/removed.

From https://github.com/open-telemetry/semantic-conventions/issues/50, metric names might not be allowed to be namespaces that's why it's most probably better to have `.slab.usage` and `.slab.state`?

So we can report:
```
Metric name: system.linux.memory.slab.usage       
Attributes: system.linux.memory.slab.state (recommended attributes: reclaimable and unreclaimable)
```

lmolkova: BTW, do you need limit metric at all since it's derivable from the `system.linux.memory.slab.usage` ?

mx-psi: @lmolkova I guess we can make it opt-in instead of recommended, but I think it makes sense to have `.limit`. See #409 for prior discussion on this area for `system.memory.limit`, which is also derivable from `system.memory.usage`

lmolkova: +1 for opt-in (if we need it). I believe it's a choice to add a deriveable total-like metric or not add it at all. If there is no use-case for it, then I'd rather not add it, but if you believe it's useful, let's just find a different name and make it opt-in.

rogercoll: After discussing this topic in the system SIG, we would opt for removing the limit metric as missing a clear use-case. The total used slab memory is not constant (not a limit), but it can be derivative from the usage state sum (reclaimable + unreclaimable). 

Removed in https://github.com/open-telemetry/semantic-conventions/pull/1078/commits/806a9b6d3fe2b55b560b45e26ccbff9ad7942df5

model/metrics/system-metrics.yaml
```diff
@@ -340,3 +340,16 @@ groups:
...
     attributes: []
+
+  - id: metric.system.linux.memory.slab.usage
+    type: metric
+    metric_name: system.linux.memory.slab.usage
+    stability: experimental
+    brief: "Reports the memory used by the Linux kernel for managing caches of frequently used objects."
+    note: |
+      The sum over all `system.linux.memory.slab.usage` state values SHOULD be equal to the total slab memory available on the system.
+      Note that the total slab memory is not constant and may vary over time.
```
ChrsMark, line 351: Would be super helpful if there was any links to any kind of docs that mention more about that.

ChrsMark (approved):

model/registry/system.yaml
```diff
@@ -254,3 +254,22 @@ groups:
         brief: >
           The process state, e.g., [Linux Process State Codes](https://man7.org/linux/man-pages/man1/ps.1.html#PROCESS_STATE_CODES)
         examples: ["running"]
+  # system.linux.* attribute group
+  # system.linux.memory.slab* attribute group
+  - id: registry.system.linux.memory.slab
```
trisch-me, line 259: nit: I feel like it's too narrow group. Maybe it makes sense to have `system.linux.memory` grouping? So that if someone will add more attributes he could reuse this group.

joaopgrassi: I agree, and this is even disallowed according to the rules here: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#metric-attributes

The attributes should be under `system.linux.memory`.

joaopgrassi (changes requested):

model/registry/system.yaml
rogercoll: Sounds good to me. Indeed, there is another Linux memory metric (Kernel) that could benefit from the `reclaimable` state attribute too.

model/metrics/system-metrics.yaml
rogercoll: Definitely, added a couple of links in https://github.com/open-telemetry/semantic-conventions/pull/1078/commits/6b01a6fbcd79fed011cb7eae3be9f474dbe7e2c9

model/registry/system.yaml
rogercoll: Attributes moved to `system.linux.memory` in https://github.com/open-telemetry/semantic-conventions/pull/1078/commits/6b01a6fbcd79fed011cb7eae3be9f474dbe7e2c9

trisch-me (approved):

model/registry/system.yaml
```diff
@@ -254,3 +254,21 @@ groups:
         brief: >
           The process state, e.g., [Linux Process State Codes](https://man7.org/linux/man-pages/man1/ps.1.html#PROCESS_STATE_CODES)
         examples: ["running"]
+  # system.linux.memory.* attribute group
+  - id: registry.system.linux.memory
+    prefix: system.linux.memory
+    type: attribute_group
+    brief: "Describes Linux Memory attributes"
+    attributes:
+      - id: state
```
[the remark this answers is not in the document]
lmolkova: (or `linux.memory.slab` since slab only applies to linux memory)

model/metrics/system-metrics.yaml
```diff
@@ -340,3 +340,17 @@ groups:
     instrument: updowncounter
     unit: "By"
     attributes: []
+
+  - id: metric.system.linux.memory.slab.usage
```
lmolkova, line 344: should it be `linux` or `unix`?

rogercoll: As far as I am aware, it should be `linux`. Although, some other Unix-like kernels like Solaris uses the same memory management strategy, others use a different implementation or name. For example, FreeBSD uses UMA (Universal Memory Allocator) which is a distinct implementation optimized for the FreeBSD kernel. (OpenBSD uses the Pool Allocator)

model/registry/system.yaml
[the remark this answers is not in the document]
rogercoll: Great! On Linux there is a memory metric named `Kreclaimable` which stands for the memory `allocations that the kernel will attempt to reclaim under memory pressure`. That metric contains the currently added `system.linux.memory.slab.usage` metric with the `reclaimable` state. In that sense, I think we could reuse the reclaimable `system.linux.memory.state` with a future `system.linux.memory.kernel.usage` metric.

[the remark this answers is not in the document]
lmolkova: Sorry I misunderstood the comment. My suggestion is to do `linux.memory.state` - it does not seem like `system.*` prefix is useful here. It'd be important to remove `system` if there are reasons to believe the same attribute will be applicable on other metrics (e.g. process, container, etc)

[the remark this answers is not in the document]
rogercoll: I see, although I am not aware that processes typically report these states, they can be retrieved from cgroups (group of processes/containers). Actually, Elastic is retrieving those states for system processes: https://www.elastic.co/guide/en/beats/metricbeat/8.13/metricbeat-metricset-system-process.html

It looks reasonable to me to decouple it from the `system` namespace. :+1: 

cc @open-telemetry/semconv-system-approvers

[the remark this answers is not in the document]
rogercoll: `linux.memory.state` added in https://github.com/open-telemetry/semantic-conventions/pull/1078/commits/6e116d22b36a3eb5ace28a9e3e830d0c8f333424

docs/system/system-metrics.md
```diff
@@ -1464,6 +1465,58 @@ See also `MemAvailable` in [/proc/meminfo](https://man7.org/linux/man-pages/man5
 
 
 
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `system.linux.memory.slab.usage`
```
trask, line 1473: currently we have both of these namespaces in this PR:

* `linux.*`
* and `system.linux.*`

should we rename `system.linux.*` to `linux.*`?

that would seem to align with our decision to move `process.runtime.jvm.*` to `jvm.*`

rogercoll: I am not sure about this one, the idea of these metric was to fix the current structure of "system.memory.*" metrics for Linux only ones. First step, was to add the `system.linux.memory.available` metric, and secondly, move the slab metric into the "linux" namespace too. I think it is still important to differentiate from where this metric is extracted, whether it is the system or a process. How would this play with the current system metrics? How could I indicate that the metric is related to the system if we remove the "system" namespace?

lmolkova: Is slab memory usage reported per-process? Would it be reported as a container metric?
If it's highly unlikely that we'd add it to other namespaces, I'd suggest to define it in the root `linux` namespace.

Also curious what @open-telemetry/semconv-container-approvers think.

I think it's a part of a bigger discussion on why we have so many similar metrics - 
https://github.com/open-telemetry/semantic-conventions/issues/1161

ChrsMark: I don't think the Docker Stats API provides such a metric: https://docs.docker.com/engine/api/v1.45/#tag/Container/operation/ContainerStats
`/stats/summary` that `kubeletstats` receiver scrapes from does not provide that as well from what I can see: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kubeletstatsreceiver/testdata/stats-summary.json

lmolkova: I believe this PR does not change the status-quo on how `system.{os}` metrics are reported  - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#systemos---os-specific-system-metrics 

so I'm going to resolve this, thanks for clarifications

docs/attributes-registry/linux.md
```diff
@@ -0,0 +1,22 @@
...
+
+# Linux
+
+## Linux Memory Attributes
+
+Describes Linux Memory attributes
+
+| Attribute            | Type   | Description            | Examples                       | Stability                                                        |
+| -------------------- | ------ | ---------------------- | ------------------------------ | ---------------------------------------------------------------- |
+| `linux.memory.state` | string | The Linux memory state | `reclaimable`; `unreclaimable` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 15: I think this attribute is specific to slab memory, correct? Not all memory would have these states, so I'm suggesting to rename to `linux.slab.memory.state`.

lmolkova (approved):
Left a suggestion on attribute naming, look good otherwise!

docs/system/system-metrics.md
lmolkova: Also created https://github.com/open-telemetry/semantic-conventions/issues/1255 to discuss `system.linux` vs `linux`

docs/attributes-registry/linux.md
rogercoll: Sounds good to move it into `linux.slab.memory.state`, but I think there could be another `linux` memory metric that could reuse the `reclaimable` state. See `KReclaimable` in https://man7.org/linux/man-pages/man5/proc_meminfo.5.html (Kernel allocations that the kernel will attempt to reclaim under memory pressure.). Given this future use case, should we keep it `linux` generic or it will make sense to create a `linux.kernel.memory.state` and duplicate the attribute?

lmolkova: Ah, thanks for the explanation!

It seems the attribute will have different values depending on the metric, which seems to be confusing:
- slab memory would have values mapped to `sreclaimable` and `sunreclaim`
- kernel would only have one value - `kreclaimable`. Perhaps kernel memory metric would not even have such attribute and would be called something like `system.linux.kernel.reclaimable.memory` 

So on the first sight there won't be any duplication and `linux.slab.memory.state` makes the most sense for now.

lmolkova: agree, `linux.memory.slab.state` looks good.

as for groups: the group in yaml is implementation detail + the thing being rendered as an individual table. I think rendering all `linux.memory` attributes in one table makes sense. Attribute definitions and grouping is irrelevant to how these attributes are used on metrics.

We still render metric-specific attributes here https://github.com/open-telemetry/semantic-conventions/blob/48b89870580518b175b924e36284678c263e7c22/docs/system/system-metrics.md#metric-systemlinuxmemoryslabusage, so it all looks good to me.

rogercoll: Makes sense, `slab` namespace included in the attribute: https://github.com/open-telemetry/semantic-conventions/pull/1078/commits/58fad8529029cfbf2e8a47e442179895e128cd8b

Thanks pointing this out.

trisch-me:
@rogercoll there are still some failing checks that prevent from merging

trisch-me:
@joaopgrassi could you check this one as you have requested changes but I think it's resolved now

rogercoll:
> @rogercoll there are still some failing checks that prevent from merging

I think the failing checks are unrelated to the PR but to endpoints blocking the CI IP. Error message:
```
ERROR: 1 dead links found in ./docs/resource/k8s.md !
  [✖] https://www.itu.int/ITU-T/studygroups/com17/oid.html → Status: 0
```
I can reach the failed endpoint on my end.

.chloggen/add_linux_memory_slab.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: system
```
joaopgrassi, line 10:
```suggestion
component: linux
```

joaopgrassi (approved):

PR #1091, opened 2024-05-29
Session start and end events

breedx-splk:
## Changes

This serves as an alternative/successor to #1042. Instead of describing a session change event, we have a session start and session end event instead.


docs/general/session.md
```diff
@@ -35,4 +35,36 @@ backends can link the two sessions.
...
+
+### Session End Event
+
+![Experimental](https://img.shields.io/badge/-experimental-blue)
+
+`event.name`=`session.end`
+
+For instrumentation that tracks users behavior during user sessions, a `session.end` event SHOULD be emitted
+every time a session ends. When a session ends and continues as a new session, this event SHOULD also be
+emitted prior to the `session.start` event.
```
bidetofevil, line 64: If a start event arrives before an end event for the previous session, how should the server handle it? Is receiving a start event with a previous session an indication that that previous session has ended?

If the answer is yes, then what I raised in the previous PR would be handled with these two events - end can be used if there is no new session begins, and a start can be used if one does.

If that is the correct interpretation, maybe we can make that explicit in this definition?

docs/general/session.md
```diff
@@ -35,4 +35,36 @@ backends can link the two sessions.
...
+`event.name`=`session.start`
+
+For instrumentation that tracks users behavior during user sessions, a `session.start` event MUST be emitted
+every time a session is created. When a new session is created as a continuation of a prior session,
+the `session.previous_id` SHOULD be included in the event. The values of `session.id` and `session.previous_id`
+MUST be different.
+
+| Body field                                                    | Type   | Description                                          | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) |
+|---------------------------------------------------------------|--------|------------------------------------------------------|-------------------------------------------------------------------------------------------------------|
+| [`session.id`](/docs/attributes-registry/session.md)          | string | The ID of the new session being started.             | `Required`                                                                                            |
```
bidetofevil, line 53: Having this in the body field, will this preclude it from also being an attribute?

bidetofevil (approved):
LGTM. One non-blocking question, and a suggestion to mention how the previous session ID can be used to signify that the previous session has ended - but again, non-blocking

docs/general/session.md
breedx-splk: It's a good question. As written now, it's intended to be in the body because I think that's what people were suggesting in the other PR, and we have the body. It wouldn't make sense to me to have it in both places .

bidetofevil: The way this is being used, it definitely belongs in the body given it has a specific meaning in the context of this event, which is why we might want to use a different name to disambiguate it from a `session.id` being in the attributes, which signals the session the event was fired in.

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
...
+
+`event.name`=`session.end`
+
+For instrumentation that tracks users behavior during user sessions, a `session.end` event SHOULD be emitted
+every time a session ends. When a session ends and continues as a new session, this event SHOULD also be
+emitted prior to the `session.start` event.
+
+| Body field                                                    | Type   | Description                                          | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) |
+|---------------------------------------------------------------|--------|------------------------------------------------------|-------------------------------------------------------------------------------------------------------|
+| [`session.id`](/docs/attributes-registry/session.md)          | string | The ID of the new session being started.             | `Required`                                                                                            |
```
bidetofevil, line 73: This is the ID of the session that has ended, right? The description should said "started". 

Also, does it make more sense to use `session.previous_id` to mirror the `start` event?

breedx-splk: Oh yeah good catch. Copypasta artifact. Will fix.

I don't like using `session.previous_id` for this. This event is intended to signify the ending of a session. Which session? The one identified by the `session.id` in the body. Let's keep this simple.

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Session Events
```
jsuereth, line 38: From an overall semconv maintainer standpoint - I *would* like to avoid the state where we can add conventions without YAML / machine-readable backing.

Is the event and/or client WG tracking all of these contributions with plans to immediately move to YAML when able?

If, not could you at least define an empty `event` group with the id of this event?

I'd also be happy if you updated `Event` template to declare that `event.name` gets filled out with the name of the group in the YAML: https://github.com/open-telemetry/semantic-conventions/blob/main/templates/registry/markdown/event_macros.j2#L2

breedx-splk: I've created https://github.com/open-telemetry/semantic-conventions/issues/1133 to track these events that are being described in markdown before the yaml tooling is ready.

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
...
+
+### Session Start Event
+
+![Experimental](https://img.shields.io/badge/-experimental-blue)
+
+`event.name`=`session.start`
+
+For instrumentation that tracks users behavior during user sessions, a `session.start` event MUST be emitted
```
joaopgrassi, line 46:
```suggestion
For instrumentation that tracks user behavior during user sessions, a `session.start` event MUST be emitted
```

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
...
+| [`session.id`](/docs/attributes-registry/session.md)          | string | The ID of the new session being started.             | `Required`                                                                                            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`session.previous_id`](/docs/attributes-registry/session.md) | string | The previous `session.id` for this user, when known. | `Conditionally Required`                                                                              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+### Session End Event
+
+![Experimental](https://img.shields.io/badge/-experimental-blue)
+
+`event.name`=`session.end`
+
+For instrumentation that tracks users behavior during user sessions, a `session.end` event SHOULD be emitted
```
joaopgrassi, line 67:
```suggestion
For instrumentation that tracks user behavior during user sessions, a `session.end` event SHOULD be emitted
```

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
...
+every time a session ends. When a session ends and continues as a new session, this event SHOULD also be
```
joaopgrassi, line 68:
```suggestion
every time a session ends. When a session ends and continues as a new session, this event SHOULD be
```

joaopgrassi (approved):

joaopgrassi:
@breedx-splk I think it would be good to have folks from the event and client working groups to take a look and approve this? Are you in touch with them? We don't seem to have a GitHub team for them so not sure how to get attention.

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
...
+that the previous session has ended. If the session ID in `session.previous_id` has not yet ended via explicit
+`session.end` event, then the receiver SHOULD treat this continuation event as semantically equivalent to
+`session.end(session.previous_id)` and `session.start(session.id)`.
+
+| Body field                                                    | Type   | Description                                          | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability                                                        |
+|---------------------------------------------------------------|--------|------------------------------------------------------|-------------------------------------------------------------------------------------------------------|------------------------------------------------------------------|
+| [`session.id`](/docs/attributes-registry/session.md)          | string | The ID of the new session being started.             | `Required`                                                                                            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
MSNev, line 58: nit: as this is the "session.x" event do you really want / need the specific body fields to also be prefixed with "session."

MSNev: As part of the semantic conventions for events I am adding support for "alias" so that body fields (both in this case) could reference the global `session.id` and then provide an alias (name) for inclusion in the body along with overriding the inherited brief, note, examples etc. to further clarify what this specific usage is about. The general approach should be the narrowing of scope (from referenced attributes) rather than opening / broadening / loosening the global semantics.

LikeTheSalad (approved):

docs/general/session.md
scheler: My preference is to keep the `session.` prefix for `session.id` here, as it helps easily connect with its usage in other places.  This event is sent only once in a while, so payload size should be less of a concern.

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
...
+
+### Session End Event
+
+![Experimental](https://img.shields.io/badge/-experimental-blue)
+
+`event.name`=`session.end`
+
+For instrumentation that tracks user behavior during user sessions, a `session.end` event SHOULD be emitted
+every time a session ends. When a session ends and continues as a new session, this event SHOULD be
+emitted prior to the `session.start` event.
```
scheler, line 69: since this event is optional, s/SHOULD/MAY/g

scheler (approved):
Mostly looks good, left a couple comments.  I envision having named sessions in the long term, but it can be added in the future using a `session.name` field.

docs/general/session.md
breedx-splk: I'm with @scheler on this one -- prefer to keep `session.id` here, at least until aliases are available. When that happens, I am happy to revisit.

docs/general/session.md
breedx-splk: I thought SHOULD was a bit more suggestive of "Oh hey this is a good idea you probably want to do this..." versus MAY which is like "Doesn't matter, do this if you want".

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
...
+| [`session.id`](/docs/attributes-registry/session.md)          | string | The ID of the new session being started.             | `Required`                                                                                            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`session.previous_id`](/docs/attributes-registry/session.md) | string | The previous `session.id` for this user, when known. | `Conditionally Required`                                                                              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+### Session End Event
+
+![Experimental](https://img.shields.io/badge/-experimental-blue)
+
+`event.name`=`session.end`
+
+For instrumentation that tracks user behavior during user sessions, a `session.end` event SHOULD be emitted
```
scheler, line 67: nit: is it really an instrumentation that will emit these events? If I am not wrong, it would be some SDK level software. If correct, please consider using another term for instrumentation.

docs/general/session.md
scheler: For the browser case, there will be a scenario of session timing out when `session.end` event can be emitted. However, it's better to avoid this since it's not possible to emit the event in all the scenarios when session ends. To avoid confusion, it's better to not emit this event at all.  It is for this reason I suggest using MAY instead of SHOULD.

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Session Events
+
+### Session Start Event
+
+![Experimental](https://img.shields.io/badge/-experimental-blue)
+
+`event.name`=`session.start`
```
lmolkova, line 44: let's use normative language as in other places. I believe in this case it's "MUST be"

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
...
+
+For instrumentation that tracks user behavior during user sessions, a `session.start` event MUST be emitted
+every time a session is created. When a new session is created as a continuation of a prior session,
+the `session.previous_id` SHOULD be included in the event. The values of `session.id` and `session.previous_id`
+MUST be different.
+
+When the `session.start` event contains both `session.id` and `session.previous_id` fields, the event then implies
+that the previous session has ended. If the session ID in `session.previous_id` has not yet ended via explicit
+`session.end` event, then the receiver SHOULD treat this continuation event as semantically equivalent to
```
lmolkova, line 53: nit:

I *think* we use "consumer" term instead of "receiver" in other places

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
...
+|---------------------------------------------------------------|--------|------------------------------------------------------|-------------------------------------------------------------------------------------------------------|------------------------------------------------------------------|
+| [`session.id`](/docs/attributes-registry/session.md)          | string | The ID of the new session being started.             | `Required`                                                                                            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`session.previous_id`](/docs/attributes-registry/session.md) | string | The previous `session.id` for this user, when known. | `Conditionally Required`                                                                              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+### Session End Event
+
+![Experimental](https://img.shields.io/badge/-experimental-blue)
+
+`event.name`=`session.end`
```
lmolkova, line 65:
```suggestion
`event.name` MUST be `session.end`
```

docs/general/session.md
breedx-splk, line 44:
```suggestion
`event.name` MUST be`session.start`
```

docs/general/session.md
```diff
@@ -35,4 +35,41 @@ backends can link the two sessions.
...
+`event.name` MUST be`session.start`
+
+For instrumentation that tracks user behavior during user sessions, a `session.start` event MUST be emitted
+every time a session is created. When a new session is created as a continuation of a prior session,
+the `session.previous_id` SHOULD be included in the event. The values of `session.id` and `session.previous_id`
+MUST be different.
+
+When the `session.start` event contains both `session.id` and `session.previous_id` fields, the event then implies
+that the previous session has ended. If the session ID in `session.previous_id` has not yet ended via explicit
+`session.end` event, then the receiver SHOULD treat this continuation event as semantically equivalent to
```
breedx-splk, line 53:
```suggestion
`session.end` event, then the consumer SHOULD treat this continuation event as semantically equivalent to
```

breedx-splk:
3 approvals, are we waiting for anything else on this? 🙏🏻

lmolkova:
@breedx-splk please resolve open discussions (if they are resolved) and it should be ready to go

lmolkova (approved):

docs/general/session.md
breedx-splk: The word `instrumentation` is used there in regard to software that is tracking user behavior, not in regard to the thing that generates events.

docs/general/session.md
breedx-splk: > To avoid confusion, it's better to not emit this event at all.

While I definitely agree that there will be cases where the client instrumentation cannot send the `session.end`, I disagree that this should be discouraged. Backends that receive a session.end can begin processing a complete session immediately, instead of waiting for a much longer timeout. Prefer to keep it SHOULD.

Issue #1053, opened 2024-05-21
DB Grouping variables from statement


maryliag:
### Is your change request related to a problem? Please describe.

As part of sanitization, one improvement is to also do a grouping of the replacements. Splitting this issue from https://github.com/open-telemetry/semantic-conventions/issues/717 to focus on the grouping itself.

### Describe the solution you'd like

For example: 
When there was IN clause, it would be replaced by one of the values: 
- `__more1_10__`
- `__more10_100__`
- `__more100_200__`
- `__more200_300__`
- `__more300_400__`
- `__more400_500__`
- `__more500_600__`
- `__more600_700__`
- `__more700_800__`
- `__more800_900__`
- `__more900_1000__`
-  `__more1000_plus__`. 
 
That created a nice balance of separating groups that would use different plan executions, but at the same time keeping cardinality lower of different possible final strings, since the list can be quite big (I saw cases with 20k+ values in a list)

### Describe alternatives you've considered

Another solution would be to always replace with the exact value is being grouped, such as `__more23__`, but that would increase cardinality and this level of details is not that helpful. A solution creating buckets would make more sense.

maryliag:
@trask I created the issue as we discussed on the last SIG, but I don't have permission to add this to the DB Client Semantic Convention project

joaopgrassi:
I added to the project now and removed the triage label :)

maryliag:
thank you @joaopgrassi ! And yes, you can assign it to me

trask:
let's add something after #1100 to mention in lists MAY be collapsed in _some_ way

trask:
Discussed previously in DB semconv meeting:

I sent #1243 to address https://github.com/open-telemetry/semantic-conventions/issues/1053#issuecomment-2183051443.

After that is merged we can postpone the remaining portions of this issue until after stability.

trask:
should we change MAY to SHOULD collapse in-clauses during sanitization?

I think the reason we said MAY is that collapsing in-clauses is not related to sanitization...

at the same time though, if you are going through the trouble of sanitizing, this is a nice extra...

PR #1100, opened 2024-05-31
Define sanitization for db.query.text

### LINKED ISSUE #717: DB sanitization uniform format

According to the [db spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes), the `db.statement` value can be sanitized, but it is not defined how to do so.

Currently, the sanitization is being dealt with differently in few places.
I suggest to add a uniform format that will describe how to do the sanitization.
(It will be best if this format will apply to all different DB's and syntaxes)

Different implementations examples:
- [JS mongo db](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts#L606) - implements the sanitization by replacing the information with question marks
- [Python pymongo](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py#L202) - implements the sanitization by deleting the information completely, and leaving the query method name only 
- [Python elasticsearch (WIP) ](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1598) - suggests to replace the information with a string that will explain that the data is sanitized

I suggest a few options to replace the value with:

- **Keep the method name, and add a sanitized text.** 
**for example:** `db.statement = "SELECT {query information is sanitized}"`
**advantages:** quite easy to implement, easy to keep consistent across libraries.
**dis-advantages:** will require to research whether all different libraries can handle this format effectively.

- **Simple text that will describe that the value is sanitized.**
**for example:** `db.statement = "query information is sanitized"`
**advantages:** easy to implement, easy to keep consistent across different libraries.
**dis-advantages:** doesn't supply basic information about the query that could be useful.

- **Replace the values with question marks.**
 **for example:** `db.statement = "SELECT ? FROM ?"`
 **advantages:** keeps more amount of information, while still not exposing sensitive or private data.
 **dis-advantages:** harder to implement, harder to keep consistent across libraries. 
 
 I would like to hear opinions about the suggested solutions, or hear different ideas.

**Additional context.**

open-telemetry/opentelemetry-specification#3104 - Issue regarding changing the recommendation to sanitize the information by default.
open-telemetry/semantic-conventions#708 - Issue about missing examples for sanitization in specs.

gregkalapos:
Fixes #717

## Changes

Describes how to sanitize `db.query.text`. 


model/trace/database.yaml
```diff
@@ -5,18 +5,7 @@ groups:
...
-            Non-parameterized query text SHOULD NOT be collected by default unless there is sanitization that excludes
-            sensitive data, e.g. by redacting all literal values present in the query text.
-
-            Parameterized query text SHOULD be collected by default
-            (the query parameter values themselves are opt-in,
-            see [`db.query.parameter.<key>`](../../docs/attributes-registry/db.md)).
-        note:
-          Even though parameterized query text can potentially have sensitive data, by using a parameterized query
-          the user is giving a strong signal that any sensitive data will be passed as parameter values, and the benefit
-          to observability of capturing the static part of the query text by default outweighs the risk.
```
gregkalapos, line 19: I'm not sure this was intentionally defined here in `attributes.db.client.minimal`. This was overwritten below in `trace.db.common.query_and_collection` (which extends `attributes.db.client.minimal`), which resulted in this text never being rendered in any of the `.md` files. Some specific db specs extend `attributes.db.client.minimal`, but none of them seem to use `db.query.text`.

So I moved the whole text below to `trace.db.common.query_and_collection`, which seem to make more sense (e.g. that way `database-spans.md` is correct, but may not be ok for some other specs).

I think this should be cleaned up - let's discuss in the WG meeting.

docs/database/cosmosdb.md
```diff
@@ -61,9 +61,10 @@ For batch operations, if the individual operations are known to have the same op
 
 **[7]:** If using a port other than the default port for this DBMS and if `server.address` is set.
 
-**[8]:** For batch operations, if the individual operations are known to have the same query text then that query text SHOULD be used, otherwise all of the individual query texts SHOULD be concatenated with separator `; ` or some other database system specific separator if more applicable.
+**[8]:** Even though parameterized query text can potentially have sensitive data, by using a parameterized query the user is giving a strong signal that any sensitive data will be passed as parameter values, and the benefit to observability of capturing the static part of the query text by default outweighs the risk.
```
lmolkova, line 64: this change seems off. We used to document what it should be be batch operations and now we talk about parametrization.

Perhaps we want both parts of this note to be present (here and in the registry)?

I have no objections addressing it here or moving to a separate PR.

docs/database/cosmosdb.md
```diff
@@ -61,9 +61,10 @@ For batch operations, if the individual operations are known to have the same op
...
 
-**[9]:** SHOULD be collected by default only if there is sanitization that excludes sensitive information.
+**[9]:** Non-parameterized query text SHOULD NOT be collected by default unless there is sanitization that excludes sensitive data. Sanitization SHOULD be done by replacing all literals with a placeholder value. Such literals include, but are not limited to, String literals, Numeric literals, Date and Time literals, Boolean literals, Interval literals, Binary literals, and Hexadecimal literals. The placeholder value SHOULD be `?`, unless it already has a defined meaning in the given database system, in which case the instrumentation MAY choose a different placeholder.
```
lmolkova, line 66: I think this text belongs on the attribute note (and not on the requirement level) since it talks about the format, not the condition under which the attribute should be collected. 
Also examples in the second sentence are definitely related to the existing parametrization note.

maryliag: I wonder if the rules/format of how to sanitize should be it's own section (or maybe file), since can be complex. This way we can add examples, considerations, etc, and then all places can just point to that link.

Then I can add the rules of grouping variables (issue #1053) on the same place.

gregkalapos: Yeah, totally agree, we should move this.

> should be it's own section (or maybe file)

So where should we move it? The big advantage of having this in yaml is that it'll be always auto generated (like this md file) - the downside is that the best we have there is to put this into a `note`, right? 

If we move it into a file, then we'll need to manually maintain it.

What should we do?

matt-hensley: From June 5 SIG call: Is there any way to determine whether a query has been sanitized? `?` is a common placeholder for query parameters.

An explicit flag that a query was sanitized could work, with the benefit of not require any string parsing to determine if the query was originally parameterized or not.

maryliag: > If we move it into a file, then we'll need to manually maintain it.

I see this as the equivalent of manually maintaining the yaml file. If there is a new md file, your yaml could have the note pointing to it, so that part continues to be auto generated.

The new file could be on `docs/database/sanitization.md`

model/trace/database.yaml
gregkalapos: Should be ok now.

docs/database/cosmosdb.md
gregkalapos: Discussed in the WG meeting - we add it to `database-spans.md` for now. Happy to move it out later if needed. We link to that section from yaml.

maryliag: Awesome, I like how it looks now! Thank you for making the update!

maryliag (approved):

docs/database/database-spans.md
```diff
@@ -76,6 +77,17 @@ and SHOULD adhere to one of the following values, provided they are accessible:
 If a corresponding `{target}` value is not available for a specific operation, the instrumentation SHOULD omit the `{target}`.
 For example, for an operation describing SQL query on an anonymous table like `SELECT * FROM (SELECT * FROM table) t`, span name should be `SELECT`.
 
+## Sanitization of `db.query.text`
```
lmolkova, line 80: Can it be moved down to be under the attribute table? Then we'll first introduce the attribute and then talk about its sanitization.

docs/database/database-spans.md
```diff
@@ -76,6 +77,17 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+
+Sanitization MUST be done by replacing all literals with a placeholder value. Such literals
```
lmolkova, line 82: It could be interpreted as that sanitization is a MUST and I hope we can clarify it with something like

```suggestion
The `db.query.text` SHOULD be collected by default only if there is sanitization that excludes sensitive information.

Sanitization SHOULD replace all literals with a placeholder value. Such literals
```

I intentionally changed MUST to SHOULD since the sanitization so far meant removing sensitive information, not all literals. Happy to discuss more.

But also it feels that we use MUST to tell **how** to do sanitization and it's too strong.

docs/database/database-spans.md
```diff
@@ -125,7 +137,8 @@ For batch operations, if the individual operations are known to have the same op
 
 **[9]:** If using a port other than the default port for this DBMS and if `server.address` is set.
 
-**[10]:** For batch operations, if the individual operations are known to have the same query text then that query text SHOULD be used, otherwise all of the individual query texts SHOULD be concatenated with separator `; ` or some other database system specific separator if more applicable.
+**[10]:** For sanitization see [Sanitization of `db.query.text`](../../docs/database/database-spans.md#sanitization-of-dbquerytext)
+For batch operations, if the individual operations are known to have the same query text then that query text SHOULD be used, otherwise all of the individual query texts SHOULD be concatenated with separator `; ` or some other database system specific separator if more applicable.
 
 **[11]:** SHOULD be collected by default only if there is sanitization that excludes sensitive information.
```
lmolkova, line 143: let's reference the new section here instead of the current text in the note?

docs/database/database-spans.md
gregkalapos: `SHOULD` sounds ok to me here as well. I was going back and forth between `SHOULD` and `MUST` - so let's just move on with `SHOULD`.

docs/database/database-spans.md
gregkalapos: Makes sense, moved this part below.

docs/database/cosmosdb.md
gregkalapos: > From June 5 SIG call: Is there any way to determine whether a query has been sanitized? `?` is a common placeholder for query parameters.
> 
> An explicit flag that a query was sanitized could work, with the benefit of not require any string parsing to determine if the query was originally parameterized or not.

That's a good point, however I'm not sure if this is important enough to address it at this point. We can always add an additional attribute to store if the query was sanitized or not as an incremental change.

There is the `db.query.parameter.<key>` to store parameters - so if the query is parametrized, then that attributes is used - however that attribute is opt-in, so admittedly that'll not help in a default setup.

I personally would wait for user feedback on this and if needed, then I'd add the additional flag on-demand if we really see the need for it.

docs/database/database-spans.md
```diff
@@ -225,6 +227,18 @@ Back ends could, for example, use the provided identifier to determine the appro
...
+## Sanitization of `db.query.text`
+
+The `db.query.text` SHOULD be collected by default only if there is sanitization that excludes sensitive information.
+Sanitization SHOULD replace all literals with a placeholder value.
+Such literals include, but are not limited to, String literals, Numeric literals, Date and Time literals,
+Boolean literals, Interval literals, Binary literals, and Hexadecimal literals.
+The placeholder value SHOULD be `?`, unless it already has a defined meaning in the given database system,
+in which case the instrumentation MAY choose a different placeholder.
+
+Placeholders in a parametrized query SHOULD not be sanitized. E.g. `where id = $1` can be captured as is,
```
trask, line 239:
```suggestion
Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id = $1` can be captured as is,
```

docs/database/database-spans.md
```diff
@@ -225,6 +227,18 @@ Back ends could, for example, use the provided identifier to determine the appro
...
+and there is no need to sanitize the value `$1`.
```
trask, line 240:
```suggestion
and there is no need to sanitize the text `$1`.
```

model/trace/database.yaml
```diff
@@ -26,9 +28,15 @@ groups:
...
+        note: >
+          For sanitization see [Sanitization of `db.query.text`](../../docs/database/database-spans.md#sanitization-of-dbquerytext)
+
+          For batch operations, if the individual operations are known to have the same query text
+          then that query text SHOULD be used, otherwise all of the individual query texts SHOULD be concatenated
+          with separator `; ` or some other database system specific separator if more applicable.
         requirement_level:
           recommended: >
-            SHOULD be collected by default only if there is sanitization that excludes sensitive information.
+            See [Sanitization of `db.query.text`](../../docs/database/database-spans.md#sanitization-of-dbquerytext).
```
trask, line 39: what happens if we inherit note and requirement_level from `attributes.db.client.minimal`?

gregkalapos: In short: for note, we'd loose the part that talks about batch operations, but for `requirement_level`, we could probably do that. A few things inherit from `trace.db.common.query_and_collection` (e.g. cassandra, sql, cosmos), so the effect of doing it for `requirement_level` will be that all those will have a longer text for `requirement_level` talking about parameterized and Non-parameterized queries. But I think that's fine.

Regarding `note`: there is also a note in the registry, which we seem to overwrite everywhere. I don't really understand yet why. Wouldn't it be easier to just handle the note for db.query.text in the registry and just inherit that everywhere? So far I don't see the logic of differentiating that note for the different ids.

docs/database/cassandra.md
```diff
@@ -60,15 +60,15 @@ For batch operations, if the individual operations are known to have the same op
...
 
-**[9]:** For batch operations, if the individual operations are known to have the same query text then that query text SHOULD be used, otherwise all of the individual query texts SHOULD be concatenated with separator `; ` or some other database system specific separator if more applicable.
+**[9]:** For sanitization see [Sanitization of `db.query.text`](../../docs/database/database-spans.md#sanitization-of-dbquerytext).
+For batch operations, if the individual operations are known to have the same query text then that query text SHOULD be used, otherwise all of the individual query texts SHOULD be concatenated with separator `; ` or some other database system specific separator if more applicable.
+Even though parameterized query text can potentially have sensitive data, by using a parameterized query the user is giving a strong signal that any sensitive data will be passed as parameter values, and the benefit to observability of capturing the static part of the query text by default outweighs the risk.
 
-**[10]:** SHOULD be collected by default only if there is sanitization that excludes sensitive information.
```
gregkalapos, line 65: This was a note on the `requirement_level`, which got removed here. We have a general note now for the attribute which starts with `For sanitization see Sanitization of db.query.text...`, which starts with the exact same sentence. Having this twice is not needed -> so this is dropped in other md files below as well.

model/trace/database.yaml
gregkalapos: I moved the general note to the registry and consolidated it there - I really wonder why it was overwritten [here](https://github.com/open-telemetry/semantic-conventions/blob/main/model/trace/database.yaml#L16-L19). 

With this, we just inherit the note from the registry which now covers the link to sanitization, batching, and parameterized query.

Let me know if this is not an ok move.

model/trace/database.yaml
```diff
@@ -26,9 +22,6 @@ groups:
     brief: This group defines the attributes used to perform database client calls.
     attributes:
       - ref: db.query.text
-        requirement_level:
-          recommended: >
-            SHOULD be collected by default only if there is sanitization that excludes sensitive information.
```
lmolkova, line 31: I feel it's important to preserve.

I understand it's not affecting markdown much, but at least in the yaml we specify this under requirement-level. 

Also we can always do better job changing how markdown is rendered and this may become significant there too.

docs/database/database-spans.md
```diff
@@ -225,6 +226,18 @@ Back ends could, for example, use the provided identifier to determine the appro
 
 When additional attributes are added that only apply to a specific DBMS, its identifier SHOULD be used as a namespace in the attribute key as for the attributes in the sections below.
 
+## Sanitization of `db.query.text`
+
+The `db.query.text` SHOULD be collected by default only if there is sanitization that excludes sensitive information.
+Sanitization SHOULD replace all literals with a placeholder value.
+Such literals include, but are not limited to, String literals, Numeric literals, Date and Time literals,
+Boolean literals, Interval literals, Binary literals, and Hexadecimal literals.
```
lmolkova, line 236: super-nit: 

maybe
```suggestion
Such literals include, but are not limited to, String, Numeric, Date and Time,
Boolean, Interval, Binary, and Hexadecimal literals.
```

model/trace/database.yaml
gregkalapos: Ok, added it back.

docs/database/cassandra.md
gregkalapos: Discussed in https://github.com/open-telemetry/semantic-conventions/pull/1100#discussion_r1649060948 - we add it back.

docs/database/database-spans.md
```diff
@@ -225,6 +226,18 @@ Back ends could, for example, use the provided identifier to determine the appro
...
+The placeholder value SHOULD be `?`, unless it already has a defined meaning in the given database system,
+in which case the instrumentation MAY choose a different placeholder.
+
+Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id = $1` can be captured as is,
+and there is no need to sanitize the text `$1`.
```
lmolkova, line 241: nit:

```suggestion
Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id = $1` can be captured as is.
```

model/trace/database.yaml
maryliag: if is only required with condition, shouldn't this be using something like
```
 requirement_level:
          conditionally_required: If there is sanitization that excludes sensitive information.
```

docs/database/database-spans.md
```diff
@@ -225,6 +228,18 @@ Back ends could, for example, use the provided identifier to determine the appro
 
 When additional attributes are added that only apply to a specific DBMS, its identifier SHOULD be used as a namespace in the attribute key as for the attributes in the sections below.
 
+## Sanitization of `db.query.text`
+
+The `db.query.text` SHOULD be collected by default only if there is sanitization that excludes sensitive information.
+Sanitization SHOULD replace all literals with a placeholder value.
+Such literals include, but are not limited to, String, Numeric, Date and Time,
+Boolean, Interval, Binary, and Hexadecimal literals.
```
maryliag, line 236: nit: you can probably remove the "literals" on the "Hexadecimal" too, since it already starts the sentence saying those are type of literals (I saw you removed from the other on your latest commit)

model/trace/database.yaml
gregkalapos: That'd be a stronger requirement level - it'd mean, if it's sanitized, then it must be captured. Right?

Now we just say, it's recommended, so it may not be captured and I can imagine there will be configs to opt in and out to capture this. So I feel recommended is appropriate here. That was already the case prior to the PR - so there would be no change on that here.

lmolkova: The PR would not change the status quo, so I don't want to block it on the req_level discussion, but I agree that

> conditionally_required: If there is sanitization that excludes sensitive information.

would be more precise - let's chat on the SIG call and maybe change it in a follow up

docs/database/database-spans.md
gregkalapos: I'm not a native speak, so I just asked AI and it says it's ok to drop it, but keeping it there is more precise. It says  `"literals" at the end makes it explicitly clear that all the items in the list are types of literals`. That's also how I'd understand it - the "literals" is not on the "Hexadecimal", but on the whole list.

docs/database/cosmosdb.md
trask: > I personally would wait for user feedback on this and if needed, then I'd add the additional flag on-demand if we really see the need for it.

let's create an issue for this 👍

trask (approved):

lmolkova (approved):

docs/database/cosmosdb.md
lmolkova: Created https://github.com/open-telemetry/semantic-conventions/issues/1179

PR #1103, opened 2024-05-31
Add LLM model server metrics

### LINKED ISSUE #1102: Semantic conventions for LLM model server metrics
### Area(s)

area:gen-ai

GenAI metrics today has client metrics. It would be good to add model server metrics to provide a standard set of performance and operational metrics for model servers. We have a proposal to standardize them as a part of Kubernetes Serving WG - https://docs.google.com/document/d/1SpSp1E6moa4HSrJnS4x3NpLuj88sMXr2tbofKlzTZpk/edit?resourcekey=0-ob5dR-AJxLQ5SvPlA4rdsg&tab=t.0#heading=h.qmzyorj64um1 (please request access if you are not able to access it). Having them standardized via LLM semantic conventions in open telemetry would be ideal.

Previous discussion on this in the LLM Semantic Conventions WG for context - https://docs.google.com/document/d/1EKIeDgBGXQPGehUigIRLwAUpRGa7-1kXB736EaYuJ2M/edit#bookmark=id.f4yu21sfndxj.

Solution is to include these common model server metrics in the GenAI LLM semantic conventions

_No response_

### Additional context

cc @lmolkova @SergeyKanzhelev

achandrasekar:
This change adds common model server metrics that we want to standardize on. It starts of with two common latency metrics - time per output token and time to first token.

Fixes #1102 

## Changes

This change adds common model server metrics that we want to standardize on. It starts of with two common latency metrics - time per output token and time to first token.


achandrasekar:
cc @lmolkova @SergeyKanzhelev

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -171,6 +171,75 @@ Instrumentations SHOULD document the list of errors they report.
...
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `gen_ai.server.latency.time_per_output_token` | Histogram | `s` | Mean time per output token generated | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `gen_ai.server.latency.time_per_output_token`
```
SergeyKanzhelev, line 216: same metric name as before?

SergeyKanzhelev: copy/paste error?

model/metrics/gen-ai.yaml
```diff
@@ -43,3 +43,19 @@ groups:
...
           Instrumentations SHOULD document the list of errors they report.
+  - id: metric.gen_ai.server.latency.time_per_output_token
+    type: metric
+    metric_name: gen_ai.server.latency.time_per_output_token
+    brief: 'Mean time per output token generated'
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    extends: metric_attributes.gen_ai
+  - id: metric.gen_ai.server.latency.time_to_first_token
```
SergeyKanzhelev, line 54: should we rename the `metric.gen_ai.client.operation.duration` to `metric.gen_ai.client.latency.time_to_last_token`?

SergeyKanzhelev: otherwise it is very inconsistent between client and server

model/metrics/gen-ai.yaml
```diff
@@ -43,3 +43,19 @@ groups:
           The `error.type` SHOULD match the error code returned by the Generative AI provider or the client library,
           the canonical name of exception that occurred, or another low-cardinality error identifier.
           Instrumentations SHOULD document the list of errors they report.
+  - id: metric.gen_ai.server.latency.time_per_output_token
+    type: metric
+    metric_name: gen_ai.server.latency.time_per_output_token
+    brief: 'Mean time per output token generated'
```
SergeyKanzhelev, line 49: why `mean time` and not just time? I am not quite sure how the `mean` will be calculated here

model/metrics/gen-ai.yaml
```diff
@@ -43,3 +43,19 @@ groups:
           The `error.type` SHOULD match the error code returned by the Generative AI provider or the client library,
           the canonical name of exception that occurred, or another low-cardinality error identifier.
           Instrumentations SHOULD document the list of errors they report.
+  - id: metric.gen_ai.server.latency.time_per_output_token
```
SergeyKanzhelev, line 46: do we need an `error_type` attribute here? Or this metrics only for successful replies?

docs/gen-ai/gen-ai-metrics.md
achandrasekar: Yes, good catch. Fixed it.

model/metrics/gen-ai.yaml
achandrasekar: Agree, mean is a little confusing. Fixed it. Mainly it is the (total tokens generated per request) / (request latency - time to first token). This is then separated into buckets.

model/metrics/gen-ai.yaml
achandrasekar: These metrics are not per request and are cumulative for all successful requests only. So, I don't think `error_type` is needed here. Let me know if that makes sense.

model/metrics/gen-ai.yaml
achandrasekar: `time_to_last_token` isn't the widely used terminology here. It is usually referred to as `request_duration` or `request_latency`. Ideally I would prefer to use `gen_ai.client.request.latency`. `gen_ai.client.request.duration` might be ok too. If `operation.duration` is the Otel standard derived from other metrics, then it might be ok to keep it the same too if that's the recommendation.

For context, today vLLM calls it `request_latency` and TGI and Triton call this `request_duration` - https://docs.google.com/document/d/1SpSp1E6moa4HSrJnS4x3NpLuj88sMXr2tbofKlzTZpk/edit?resourcekey=0-ob5dR-AJxLQ5SvPlA4rdsg&tab=t.0#bookmark=id.bdl7ba5eaec.

For the server metric, I haven't included it in this PR. But I'm planning to call it `gen_ai.server.latency.time_per_request`.

SergeyKanzhelev: Should the names be symmetrical? Like: `gen_ai.client.latency.request_duration` and `gen_ai.server.latency.request_duration`. In future if we need a client's time to first token it will be `gen_ai.client.latency.time_to_first_token`

model/metrics/gen-ai.yaml
SergeyKanzhelev: ok, perhaps add this assumption into the metric description? Something like 'Time per output token generated for successful responses'

achandrasekar: Makes sense. Added.

model/metrics/gen-ai.yaml
achandrasekar: Having the names be symmetrical makes sense. I've created the server metric called `gen_ai.server.latency.request_duration` in this PR. I'll make the client change in a separate PR since it could be considered breaking. 

@lmolkova let me know if you have any objection to changing the client metric name to `gen_ai.client.latency.request_duration` in a separate PR.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -2,28 +2,32 @@
...
-
 <!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->
 
 <!-- toc -->
 
 - [Generative AI Client Metrics](#generative-ai-client-metrics)
   - [Metric: `gen_ai.client.token.usage`](#metric-gen_aiclienttokenusage)
   - [Metric: `gen_ai.client.operation.duration`](#metric-gen_aiclientoperationduration)
+- [Generative AI Model Server Metrics](#generative-ai-model-server-metrics)
+  - [Metric: `gen_ai.server.latency.request_duration`](#metric-gen_aiserverlatencyrequest_duration)
```
lmolkova, line 17: Having `latency` and `duration` in the same metric name sounds confusing. 

I'd prefer `gen_ai.server.request.duration` - we already have time-to-first-token in `gen_ai.server.latency.time_to_first_token`, so we need time-to-last token. 

We can introduce a symmetrical `gen_ai.client.time_to_first_token`

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -2,28 +2,32 @@
...
+  - [Metric: `gen_ai.server.latency.time_per_output_token`](#metric-gen_aiserverlatencytime_per_output_token)
+  - [Metric: `gen_ai.server.latency.time_to_first_token`](#metric-gen_aiserverlatencytime_to_first_token)
```
lmolkova, line 19: why not `gen_ai.server.time_to_first_token` ? `latency` seems redundant

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -2,28 +2,32 @@
...
 <!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->
 
 <!-- toc -->
 
 - [Generative AI Client Metrics](#generative-ai-client-metrics)
   - [Metric: `gen_ai.client.token.usage`](#metric-gen_aiclienttokenusage)
   - [Metric: `gen_ai.client.operation.duration`](#metric-gen_aiclientoperationduration)
+- [Generative AI Model Server Metrics](#generative-ai-model-server-metrics)
+  - [Metric: `gen_ai.server.latency.request_duration`](#metric-gen_aiserverlatencyrequest_duration)
+  - [Metric: `gen_ai.server.latency.time_per_output_token`](#metric-gen_aiserverlatencytime_per_output_token)
```
lmolkova, line 18: similarly to `gen_ai.server.time_to_first_token`, why not `gen_ai.server.time_per_token`? 
Also if do `time_per_output_token`, we should probably do `time_to_first_output_token`

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -171,6 +175,106 @@ Instrumentations SHOULD document the list of errors they report.
...
+latency in terms of time spent per request.
+
+For example, if a GenAI model server reports latency information, it SHOULD be
+used.
+
+If instrumentation cannot obtain this information at a request level and break
+it down into the buckets mentioned below, then it MUST NOT report this metric.
+
+This metric SHOULD be specified with [ExplicitBucketBoundaries] of
+[1.0, 2.5, 5.0, 10.0, 15.0, 20.0, 30.0, 40.0, 50.0, 60.0].
```
lmolkova, line 200: this one does not seem to be exponential.

Can we reuse the same boundaries as for the client one? If you'd rather change them, let's do it consistently too - I assume there is no reason to expect drastically different time-to-last-token between client and server in case of LLM request.

(I agree lower boundaries are not so helpful for LLM, but there could still be simple operations such as list models or error responses that are in ms range)

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -171,6 +175,106 @@ Instrumentations SHOULD document the list of errors they report.
...
+<!-- semconv metric.gen_ai.server.latency.time_per_output_token(metric_table) -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `gen_ai.server.latency.time_per_output_token` | Histogram | `s` | Time per output token generated for successful responses | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 243: Can we provide a more strict definition (either in the brief or in the note above)?

Does it capture the time difference between two generate token for all tokens except the first one?
Should it be reported for streaming requests only or for all requests?

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -171,6 +175,106 @@ Instrumentations SHOULD document the list of errors they report.
...
+
+This metric is [recommended][MetricRecommended] to report the model server
+latency in terms of time spent to generate the first token of the response for
+any model servers which support serving LLMs.
+
+For example, if a model server which serves LLMs reports latency information,
+it SHOULD be used.
+
+If instrumentation cannot obtain this information at a request level and break
+it down into the buckets mentioned below, then it MUST NOT report this metric.
```
lmolkova, line 261: Do we need this formula on each metric?

I believe it's implied for Recommended metric level - instrumentations SHOULD report it if they can and if they report it they MUST/SHOULD follow the semantics. 

So maybe just omit it for brevity?

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -171,6 +175,106 @@ Instrumentations SHOULD document the list of errors they report.
...
+<!-- semconv metric.gen_ai.server.latency.request_duration(metric_table) -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `gen_ai.server.latency.request_duration` | Histogram | `s` | Time (end to end latency) to complete a request | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 211: maybe 
```suggestion
| `gen_ai.server.latency.request_duration` | Histogram | `s` | Generative AI server request duration such as time-to-last byte or last output token | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

model/metrics/gen-ai.yaml
```diff
@@ -43,3 +43,27 @@ groups:
...
           the canonical name of exception that occurred, or another low-cardinality error identifier.
           Instrumentations SHOULD document the list of errors they report.
+  - id: metric.gen_ai.server.latency.request_duration
+    type: metric
+    metric_name: gen_ai.server.latency.request_duration
+    brief: 'Time (end to end latency) to complete a request'
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    extends: metric_attributes.gen_ai
```
lmolkova, line 53: this one does not include `error.type` which would also apply to server metrics.

You might want to create a new attribute group and extend it, e.g.:

```yaml
  - id: metric_attributes.gen_ai.server
    type: attribute_group
    brief: 'This group describes GenAI server metrics attributes' # not rendered 
    extends: metric_attributes.gen_ai
    attributes:
      - ref: error.type
        requirement_level:
          conditionally_required: "if the operation ended in an error"
        note: | 
          The `error.type` SHOULD match the error code returned by the Generative AI service,
```

It can also be reused on client duration metric which now references attribute directly, but then we'd need to make note applicable to both.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -171,6 +175,106 @@ Instrumentations SHOULD document the list of errors they report.
...
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `gen_ai.server.latency.time_to_first_token`
+
+This metric is [recommended][MetricRecommended] to report the model server
+latency in terms of time spent to generate the first token of the response for
```
lmolkova, line 254: should it be reported if token is never generated? I.e. should the metric have `error.type` attribute or it only captures successful generations?
similar question for `time_per_output_token`

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -171,6 +175,106 @@ Instrumentations SHOULD document the list of errors they report.
...
 
 
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## Generative AI Model Server Metrics
+
+The following metric instruments describe Generative AI model servers'
```
lmolkova, line 185: can we please render attribute table for each metric? (Or if all attributes are the same, could we render them here just once?)

model/metrics/gen-ai.yaml
lmolkova: I'd prefer to keep `gen_ai.client.operation.duration`. Reasons:
- duration implies time-to-last semantics
- there are operations that don't produce tokens - e.g. list models, upload file, other control plane things.
- some operation fail before producing any tokens,
- etc.

It does not prevent us from adding more metrics for time-to-first/time-to-last-tokens.

Note: in case of clients, these metrics would only makes sense for streaming requests.

achandrasekar:
cc @jsuereth to take a look as well

docs/gen-ai/gen-ai-metrics.md
achandrasekar: Makes sense. Updated.

docs/gen-ai/gen-ai-metrics.md
achandrasekar: The reason to add latency is to group model server metrics into different categories like `latency`, `capacity`, `admission`, etc. But I'm okay with not having the grouping. I've updated the names.

docs/gen-ai/gen-ai-metrics.md
achandrasekar: Time to first token (TTFT) and time per output token (TPOT) are widely used terminologies as you can see from the best practices guides and usage across companies and organizations - [Databricks](https://www.databricks.com/blog/llm-inference-performance-engineering-best-practices), [Microsoft](https://learn.microsoft.com/en-us/azure/databricks/machine-learning/foundation-models/prov-throughput-run-benchmark#--llm-inference-introduction), [ MLCommons ](https://mlcommons.org/2024/03/mlperf-llama2-70b/) and [vLLM](https://github.com/vllm-project/vllm/blob/828da0d44e9124d949909477d6018fc08469a31e/vllm/engine/metrics.py#L77). So, I'd like to keep them as is.

model/metrics/gen-ai.yaml
achandrasekar: The metrics on the server side unlike the client metrics are not at a single request level, but are cumulative for all the requests that have been processed by the server till that point in time which are then bucketed accordingly and reported. So, I'm not sure if `error.type` is needed for it since individual requests might fail, but the cumulative metric will report data on all the requests processed so far.

docs/gen-ai/gen-ai-metrics.md
achandrasekar: Both of these are cumulative and report only on successful token generations.

docs/gen-ai/gen-ai-metrics.md
achandrasekar: Added

model/metrics/gen-ai.yaml
achandrasekar: Makes sense. We can keep the client metric the same.

model/metrics/gen-ai.yaml
lmolkova: the client metrics are also aggregates, but they are aggregated within one combination of attributes.

If we don't report error.type, it's not possible to distinguish successful responses from non-successful, get error rate, the duration would be skewed, etc. 
So it's be hard to use such metric. 

Are there some blockers that does not allow servers to use error type as a dimension?

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -2,28 +2,32 @@
...
-and attributes but more may be added in the future.
-
 <!-- Re-generate TOC with `markdown-toc --no-first-h1 -i` -->
 
 <!-- toc -->
 
 - [Generative AI Client Metrics](#generative-ai-client-metrics)
   - [Metric: `gen_ai.client.token.usage`](#metric-gen_aiclienttokenusage)
   - [Metric: `gen_ai.client.operation.duration`](#metric-gen_aiclientoperationduration)
+- [Generative AI Model Server Metrics](#generative-ai-model-server-metrics)
```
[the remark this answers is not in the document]
drewby: > non-blocking: I think eventually we should have client and server in different files (we keep HTTP metrics in one file, so this PR uses existing practice).
> 
> Adding a comment to gather more feedback. Either way, it's a cosmetic change that can be done later.

I agree we can keep it one for now and separate later.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -171,6 +175,217 @@ Instrumentations SHOULD document the list of errors they report.
...
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`gen_ai.operation.name`](/docs/attributes-registry/gen-ai.md) | string | The name of the operation being performed. | `chat`; `completion` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.request.model`](/docs/attributes-registry/gen-ai.md) | string | The name of the LLM a request is being made to. | `gpt-4` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.system`](/docs/attributes-registry/gen-ai.md) | string | The Generative AI product as identified by the client instrumentation. [1] | `openai` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
drewby, line 230: Should we remove the client reference from this Description? Or change to "client or server instrumentation"?

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -171,6 +175,217 @@ Instrumentations SHOULD document the list of errors they report.
...
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `gen_ai.server.time_per_output_token`
+
+This metric is [recommended][MetricRecommended] to report the model server
+latency in terms of time per token generated for any model servers which
```
drewby, line 259: Maybe a dumb question, but is this intended be an average of the time between all tokens? So essentially the (request_duration) / (output tokens) for a each request?

docs/gen-ai/gen-ai-metrics.md
achandrasekar: Makes sense. Made it the same as client buckets.

docs/gen-ai/gen-ai-metrics.md
achandrasekar: Updated the brief to clarify it and added a better description.

docs/gen-ai/gen-ai-metrics.md
achandrasekar: Makes sense. Included it following the other metric which had it. I've dropped it now from all the metrics.

model/metrics/gen-ai.yaml
achandrasekar: Created a new attribute group as discussed and used it for request duration.

docs/gen-ai/gen-ai-metrics.md
achandrasekar: Changed to include server as well.

docs/gen-ai/gen-ai-metrics.md
achandrasekar: Added more information in the description. TPOT per request = (request duration - TTFT) / (output tokens - 1). It is then aggregated into buckets.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -2,28 +2,32 @@
...
 
 <!-- toc -->
 
 - [Generative AI Client Metrics](#generative-ai-client-metrics)
   - [Metric: `gen_ai.client.token.usage`](#metric-gen_aiclienttokenusage)
   - [Metric: `gen_ai.client.operation.duration`](#metric-gen_aiclientoperationduration)
+- [Generative AI Model Server Metrics](#generative-ai-model-server-metrics)
+  - [Metric: `gen_ai.server.request_duration`](#metric-gen_aiserverrequest_duration)
+  - [Metric: `gen_ai.server.time_per_output_token`](#metric-gen_aiservertime_per_output_token)
+  - [Metric: `gen_ai.server.time_to_first_token`](#metric-gen_aiservertime_to_first_token)
```
gyliu513, line 19: Can you please take a look at https://github.com/vllm-project/vllm/blob/main/vllm/tracing.py#L88-L94 , this is for server observability with vLLM, it has request id, best_of, time_in_queue etc, do you want to add those metrics as well?

gyliu513: This was introduced by https://github.com/vllm-project/vllm/pull/4687/

achandrasekar: Yes, we can add those as well. But I'd do it in a separate change. Do you want to raise a PR for the tracing attributes?

gyliu513: Yes, I can do that, thanks @achandrasekar

model/metrics/gen-ai.yaml
```diff
@@ -16,6 +16,18 @@ groups:
         requirement_level: required
       - ref: gen_ai.operation.name
         requirement_level: required
+  - id: metric_attributes.gen_ai.server
+    type: attribute_group
+    brief: 'This group describes GenAI server metrics attributes'
+    extends: metric_attributes.gen_ai
+    attributes:
+    - ref: error.type
```
drewby, line 24:
```suggestion
      - ref: error.type
```

model/metrics/gen-ai.yaml
```diff
@@ -16,6 +16,18 @@ groups:
...
+      requirement_level:
```
drewby, line 25:
```suggestion
        requirement_level:
```

drewby (changes requested):

model/metrics/gen-ai.yaml
```diff
@@ -43,3 +55,27 @@ groups:
           The `error.type` SHOULD match the error code returned by the Generative AI provider or the client library,
           the canonical name of exception that occurred, or another low-cardinality error identifier.
           Instrumentations SHOULD document the list of errors they report.
+  - id: metric.gen_ai.server.request_duration
```
lmolkova, line 58:
```suggestion
  - id: metric.gen_ai.server.request.duration
```

model/metrics/gen-ai.yaml
```diff
@@ -43,3 +55,27 @@ groups:
...
+    type: metric
+    metric_name: gen_ai.server.request_duration
```
lmolkova, line 60:
```suggestion
    metric_name: gen_ai.server.request.duration
```

drewby (approved):
LGTM

lmolkova (approved):

model/metrics/gen-ai.yaml
```diff
@@ -43,3 +55,27 @@ groups:
           The `error.type` SHOULD match the error code returned by the Generative AI provider or the client library,
           the canonical name of exception that occurred, or another low-cardinality error identifier.
           Instrumentations SHOULD document the list of errors they report.
+  - id: metric.gen_ai.server.request.duration
```
joaopgrassi, line 58: Was it considered that we will have two duration histograms, when these http calls are [also instrumented ](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestduration)? I assume since here there's extra attributes this is fine, but just wanted to clarify.

joaopgrassi (approved):

model/metrics/gen-ai.yaml
jsuereth: It's a good point.  However - I think we have this general problem around  "layers" of instrumentation.   Ideally - users would have a good default, but they can pick which layers to keep and which to drop.  My ideal default would be redunandant instrumentation would always drop the "furthest semantic meaning" -i.e. if I have HTTP request metrics *AND* AI/ML - the AI/ML metrics are closer in semantic to what I'm doing so by default they'd be preserved.

I believe @lmolkova has a write-up on this issue somewhere in the specification at some point.   For now, I think this is ok.   The need for users to be able to control semantic-layers of information and the fact that we'll have overlap conflicts like this is not new, nor is this PR the first to cause the issue.

lmolkova: On the client side we separate all 'logical' from all 'physical' because of the retries, redirect, auth, response streaming. 
It's slightly different on the server.

@achandrasekar wonder what your thoughts are - what are the instrumentation point for LLM metrics? are they coming from the model or from the HTTP server?  
I assume it could be possible to use LLM with a different protocol, would the same LLM metrics apply/make sense?

Also, echoing @jsuereth comment I'd assume LLM servers would not expose HTTP metrics if they are quite similar to HTTP. If they are different, then reporting both could make sense.

achandrasekar: This server metric is coming from the model server. And there can be different protocols supported by the model server - gRPC is a popular one that model servers support in addition to HTTP. In some cases, gRPC is the only one supported by default and a separate HTTP FastAPI server is provided as an additional component that users can deploy if needed. So, I think having a model server level "request duration" different from the HTTP one makes sense to report this along with the other model server level latency metrics like time to first token.

joaopgrassi: Thanks for the insights, makes sense to me!

PR #1110, opened 2024-06-02
Don't render deprecated enum members in semconv tables

lmolkova:
Came up in #https://github.com/open-telemetry/semantic-conventions/pull/1006#discussion_r1622076165

Deprecated enum members are not useful in semconv docs and should only appear in the registry.


pyohannes (approved):

ChrsMark (approved):
I wonder if we'd rather preserve the deprecated ones in the semconv tables but only for limited period of time, i.e. for 2 releases before removing them.
However I don't know if that would be technically doable. Maybe if we were encoding the version that a member/attribute was deprecated, `deprecated-1.26.0`.

lmolkova:
> I wonder if we'd rather preserve the deprecated ones in the semconv tables but only for limited period of time, i.e. for 2 releases before removing them. However I don't know if that would be technically doable. Maybe if we were encoding the version that a member/attribute was deprecated, `deprecated-1.26.0`.

You're right, we don't have a way to express in which version a property was deprecated. It came up in the past and I created https://github.com/open-telemetry/semantic-conventions/issues/1117 to track it. If we had an option to remove them after X releases, I'd be happy to keep them around for a bit.

Until we have this option, I'd prefer not to render them in the specific semconvs - that's what we do for the attributes.

templates/registry/markdown/enum_macros.j2
```diff
@@ -1,10 +1,11 @@
 {% import 'stability.j2' as stability %}
+{% macro filter(member) %}{% if snippet_type is not defined or (member.deprecated is none or member.deprecated == "") %}{{ "True" }}{% else %}{{ "False" }}{% endif %}{% endmacro %}
```
trisch-me, line 2: nit: I might be late to the review, but filter usually filters things out, so I was stumbled upon usage, where we `filter` member but still continue to process it later
So semantically it should be `False` for good member and `True` for deprecated one.

trisch-me:
```suggestion
{% macro filter_deprecated(member) %}{% if snippet_type is not defined or (member.deprecated is none or member.deprecated == "") %}{{ "False" }}{% else %}{{ "True" }}{% endif %}{% endmacro %}
```

something like this

PR #1115, opened 2024-06-03
remove k8s.container.status.current_waiting_reason resource attribute

povilasv:
Reverts https://github.com/open-telemetry/semantic-conventions/pull/997 due to immutability constraint.

Ref https://github.com/open-telemetry/semantic-conventions/pull/997


ChrsMark (approved):

joaopgrassi (commented):
Please re-generate the markdown tables.

lmolkova (approved):
Note: attribute definitions are not related to a signal. I.e. this attribute can still be used, just not as a resource attribute.

If the only use case for it was to be used as a resource attribute, then yes, it's best to remove it before it was released and re-introduce if it's ever needed.

joaopgrassi (approved):

PR #351, opened 2023-09-27
Add geo fields from ECS into client and server attributes

kaiyan-sheng:
Fixes https://github.com/open-telemetry/opentelemetry-specification/issues/2834

## Changes

This PR is to add `geo` fields from Elastic Common Schema(ECS) into otel client and server attributes. These fields can be used to carry data about a specific location such as city name, country name, longitude, latitude, postal code, and ect.

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x] [CHANGELOG.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md) updated for non-trivial changes.
* [ ] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

PR #1116, opened 2024-06-03
Add geo fields into attribute registry

### LINKED ISSUE #1033: Add Geo fields from Elastic Common Schema

Some observability use cases require localization of resources. Examples are:

- Observability of mobile devices: Geographically localizing the mobile devices provides valuable insights into how mobile apps perform depending on their location. 
- For applications that are geographically distributed across multiple regions and data centers, the geo location can provide useful insights as well.

Having an additional `geo` namespace on the resource attributes serves use cases as outlined above and allows to filter and group data by geo locations. 

As part of the [OTEP to support the Elastic Common Schema (ECS)](https://github.com/open-telemetry/oteps/pull/199), we propose to start with adopting the [rich set of the ECS geo fields](https://www.elastic.co/guide/en/ecs/current/ecs-geo.html) as a new namespace in the resource attributes. 

**Additional context.**

This is related to this OTEP: https://github.com/open-telemetry/oteps/pull/199

See attached Draft PR for detailed proposal / context: https://github.com/open-telemetry/opentelemetry-specification/pull/2835

gregkalapos:
Fixes #1033

## Changes

Adds geo fields into the attribute registry into its own file.

This is another take on https://github.com/open-telemetry/semantic-conventions/pull/351, which basically copied the fields into both `client` and `server`. Here I aim for adding the fields only into the registry and not using it yet. At the same time, we have https://github.com/open-telemetry/build-tools/issues/240 with [a few suggestions](https://github.com/open-telemetry/build-tools/issues/240#issuecomment-1988592513) from @trisch-me.

So this PR will add the fields into the registry - then once https://github.com/open-telemetry/build-tools/issues/240 is implemented, we could use these geo fields by updating `client.yaml` and `server.yaml` and embed these fields into those.


.chloggen/geo_fields.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
trisch-me, line 7: it's a new component

model/registry/geo.yaml
```diff
@@ -0,0 +1,114 @@
...
+        type: string
+        brief: >
+          City name.
+        examples: [ 'Montreal', 'Berlin' ]
+      - id: continent_code
+        stability: experimental
+        brief: >
+          Two-letter code representing continent’s name.
+        type:
+          allow_custom_values: false
```
trisch-me, line 20: we don't have restrictions on custom values anymore and this check will be removed. 
Although in this case it makes non difference :)

model/registry/geo.yaml
```diff
@@ -0,0 +1,114 @@
...
+              stability: experimental
+              value: 'OC'
+              brief: 'Oceania'
+            - id: 'sa'
+              stability: experimental
+              value: 'SA'
+              brief: 'South America'
+      - id: continent_name
+        stability: experimental
+        type: string
```
trisch-me, line 52: should it be also an enum then?

gregkalapos: I took this from https://github.com/open-telemetry/semantic-conventions/pull/351 - so back then it seemed to be ok to have `continent_code` as enum and `continent_name` as string. At least I don't see any discussion or rejection for this. [In ECS both are `keyword`](https://www.elastic.co/guide/en/ecs/current/ecs-geo.html) - so we can't get inspired from ECS in this case.

What I have found is that `continent_code` was added afterwords: https://github.com/elastic/ecs/pull/1229 - although the motivation mentioned there doesn't really help to answer this question either.

I don't feel super strongly about this, but if we make this into an enum as well, then what's the point of having both `continent_code` and `continent_name` both as enum with fixed values? We'd have basically the same information, one would be the short code (like `eu`), the other one the full name (like `Europe`), but it'd be the exact same information. So if this becomes an enum as well, then my question is: what's the point having both and why don't we drop one?

If however this remains a string, then this will enable implementation to not parse the values into the fixed enum values and only use `continent_name` as string (just store it as-is without any parsing) and not use `continent_code`. Whether this is a good thing and we want to have this is still a question.

trisch-me: For me enums are useful if we want to have a limited set of answers and also low cardinality - if we save it just as string without any parsing we could have different spelling for the same thing. For the fixed set, and continents and their name is a small fixed set I feel enum has the best fit. 

I could see that both enums are basically the same thing, but I could see the benefit depending on the specific use cases, where one is using code and another - name.

gregkalapos: I think this comes down to the question: what do we want to allow users when it comes to continent? 

- Force them to only have a small set of fixed values? (due to the benefits you mentioned) -> enum
- Allow them to save parsing (and as you say may have different spelling, etc) -> string

> I could see that both enums are basically the same thing, but I could see the benefit depending on the specific use cases, where one is using code and another - name.

Do you think is this benefit worth the duplicated storage and the potential confusion of having 2 attributes with the same info?

The more I think about this the more I'd lean towards leaving `continent_name` as string. Also - that's more aligned with other fields: e.g. `country_name` is a string, but could be enum as well - of course that has more values, but still somewhat analogue to `continent_name`.

trisch-me: I would say countries are more to have changes in its number/naming than continents but that's not really relevant for this PR.
I don't see the benefit of having it as string other than saving on parsing, but for small subset it's not a big deal.

Saying all this I don't have a strong opinion that it should be enum, if we decide to do otherwise.

gregkalapos: Thanks for the feedback - I also don't feel super strongly either, so enum is totally acceptable as well.

Let's see what others say. @rogercoll since you worked on this in the collector: any thoughts on this?

rogercoll: I would also be tempted to use an enum for greater restriction on user input, but from a user perspective, the only unique field in any natural language seems to be the `continent_code`. Actually, while using the GeoIP database from MaxMind, you can directly retrieve the `continent_code` for any given an IP. However, if you want to retrieve the `continent_name`, you must specify a locale/name. See the [example](https://github.com/maxmind/MaxMind-DB/blob/main/source-data/GeoIP2-City-Test.json#L5).

In the [geoipprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/geoipprocessor) we are going to add the `continent_name` using the "english" name, but it makes me wonder the possibility of allowing the user to configure which name to use. In that case, being a string would be more suitable.

trisch-me (approved):

model/registry/geo.yaml
AlexanderWert: > In the [geoipprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/geoipprocessor) we are going to add the continent_name using the "english" name, but it makes me wonder the possibility of allowing the user to configure which name/locale to use. In that case, being a string would be more suitable.

Yes, that's a good reason for it to be a string

AlexanderWert (approved):

trisch-me:
Are there any comments from other maintainers regarding this new namespace?

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+      Geo fields can carry data about a specific location related to an event.
+      This geolocation information can be derived from techniques such as Geo IP, or be user-supplied.
+    attributes:
+      - id: city_name
+        stability: experimental
+        type: string
+        brief: >
+          City name.
+        examples: [ 'Montreal', 'Berlin' ]
+      - id: continent_code
```
joaopgrassi, line 15: Given the recent problems with attribute names with `_`, I think we should think about changing `continent` to be a namespace, specially since there's already multiple attributes under it. `continent.code|name`.

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+              stability: experimental
+              value: 'SA'
+              brief: 'South America'
+      - id: continent_name
+        stability: experimental
+        type: string
+        brief: >
+          Name of the continent.
+        examples: [ 'North America', 'Europe' ]
+      - id: country_iso_code
```
joaopgrassi, line 55:
```suggestion
      - id: country.iso_code
```

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+        stability: experimental
+        type: string
+        brief: >
+          Two-letter ISO Country Code ([ISO 3166-1 alpha2](https://en.wikipedia.org/wiki/ISO_3166-1#Codes)).
+        examples: [ 'CA' ]
+      - id: country_name
```
joaopgrassi, line 61:
```suggestion
      - id: country.name
```

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+          Not typically used in automated geolocation.
+        examples: [ 'boston-dc' ]
+      - id: postal_code
+        stability: experimental
+        type: string
+        brief: >
+          Postal code associated with the location.
+          Values appropriate for this field may also be known as a postcode or ZIP code and will vary widely from country to country.
+        examples: [ '94040' ]
+      - id: region_iso_code
```
joaopgrassi, line 96:
```suggestion
      - id: region.iso_code
```

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+        stability: experimental
+        type: string
+        brief: >
+          Region ISO code ([ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2)).
+        examples: [ 'CA-QC' ]
+      - id: region_name
```
joaopgrassi, line 102:
```suggestion
      - id: region.name
```

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+        stability: experimental
+        type: string
+        brief: >
+          Region name.
+        examples: [ 'Quebec' ]
+      - id: timezone
+        stability: experimental
+        type: string
+        brief: >
+          The time zone of the location, such as [IANA time zone name](https://nodatime.org/TimeZones).
```
joaopgrassi, line 112: I think we should be more strict on which format we expect the timezone identifier. Also is just `timezone` enough? This I supposed is the identifier, but there's also for ex the tz abbreviation (GMT, CET). Don't we also want to have that? If so then this should also become a namespace.

joaopgrassi (changes requested):
I think we need to look into these further and think about changing the attributes to namespaces.

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+        type: double
+        brief: >
+          Longitude of the geo location in [WGS84](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84).
+        examples: [ -73.614830 ]
+      - id: location.lat
+        stability: experimental
+        type: double
+        brief: >
+          Latitude of the geo location in [WGS84](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84).
+        examples: [ 45.505918 ]
```
KarstenSchnitter, line 78: I am wondering, whether the reference to WGS84 should be replaced by an [EPSG identifier as an SRID](https://en.wikipedia.org/wiki/EPSG_Geodetic_Parameter_Dataset). But this might be a step to far into [georeferencing](https://en.wikipedia.org/wiki/Georeferencing), which might not be required at this point in time. The convention could be refined and extended to support other reference systems, that require more fields than just a longitude/lattitude pair.

lmolkova:
Would it be possible to add a use case for those?

Do we expect them to be on span/metrics attributes? If the goal is to record events, they may go into the payload and should not be defined in the registry then.

rogercoll:
> Would it be possible to add a use case for those?

In the OpenTelemetry collector we are adding a processor that would add those attributes to any trace, metric or log. The processor looks for an IP within the resource attributes and appends the related geolocation attributes. The component is still a nop (we are just adding support for multiple geolocation providers): https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/geoipprocessor

Additional context: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32663

lmolkova:
thanks for the clarification @rogercoll 

Looking into the collector:
`source.address` is not a resource attribute, it's dynamic and describes the network packet source.

https://github.com/open-telemetry/semantic-conventions/blob/9adff435c76ea3b8cba89babefc832d3dd3c1ab9/docs/general/attributes.md?plain=1#L143-L150

Is you intention to enrich individual telemetry signals or resources?

AlexanderWert:
> Is you intention to enrich individual telemetry signals or resources?

@lmolkova The intention of the GeoIP processor is to have a flexible processor that can derive geo information from an IP.

What the source of the IP is (i.e. which attribue it's coming from) and what the destination (i.e. namespace) depends on the use case.

For example, on web server access logs the geoIP processor can be used to derive the geo information for each individual requests source.ip, allowing endusers to visualize where (geographically) requests come from.

But there could also be other use cases, that would require to read the IP from other attributes (either signal level or resource), etc.

lmolkova:
@AlexanderWert  thanks for the context!

So if I understand correctly the intention is to
- have ability to add geo attributes to any signal (traces, metrics, logs, resources)
- based on the pre-existing arbitrary IP attribute on that signal

I.e. we do need to define them as attributes and not event payload fields.

Correct ?

Looking only from the semconv perspective, it could be somewhat misleading. I can have more than one IP on the telemetry item (e.g. `host|device.ip`, `client.address`, `network.peer.address`, ...)  - if I add `geo.*`, it's not clear which one they apply to. 
What if some instrumentation (or PII redactor) reuses `geo` attributes to report end-user geography? 

I assume embedding could be useful here and we could embed the geo namespace under related IPs?

E.g. if I configure geo-processor for `client.address`, it adds `client.geo.*`, or if I do it for `source.address`, processor adds `source.geo.*`

Effectively, `geo` namespace becomes something that's usually embedded and not used as stand-alone.
This way we can enrich signals with geo information unambiguously and on any level (application, instr library, processor, backend).

Thoughts?

trisch-me:
@lmolkova you are correct, geo namespace is usually embedded into `parent` namespace and enriches it with geo information 

In ECS we don't use geo namespace as standalone in the root of events

AlexanderWert:
> Effectively, `geo` namespace becomes something that's usually embedded and not used as stand-alone.
> This way we can enrich signals with geo information unambiguously and on any level (application, instr library, processor, backend).

Yes, that's absolutely right!

I think we should consider it as a two steps process to add geo attributes to semconv:

1. This PR defines the core / plain geo attributes in the registry.
2. Once we have the `embedding` functionality, we will use that to extend those to be used under `source.*`, `client.*`, etc.

````
groups:
  - id: client
    prefix: client
    type: attribute_group
    brief: ...
    embed_namespaces:
      - id: geo
````

BTW, @trisch-me what's the status with `embed`? Do we have any blockers with that?

**Comment by trisch-me** [comment]
> what's the status with embed?

An option to embed not the full namespace but just particular fields (for example create `client.geo.lat`) is merged and we are preparing weaver release for it. 
Embedding the whole namespace at once is the next step

File: model/registry/geo.yaml (hunk 10 above)
Comment (AlexanderWert) [reply model/registry/geo.yaml:78]: I'd recommend to add that later when we have a need / use case for that specifically.

File: model/registry/geo.yaml (hunk 9 above)
Comment (gregkalapos) [reply model/registry/geo.yaml:112]: I think in this context forcing very accurate timezone reporting is super hard. So to me the question is: How much accuracy do we want to push here on implementors of this vs how much free room do we want to leave?

To me the point of this attribute would be to report the timezone of the event which was enriched with geo fields, so users know about the location of the event - which potentially includes the timezone of the location as well.

The potential problem I see with some of common timezone abbreviations is the daylight saving issue. If we have CET, then we'd need to also have CEST and if we limit this field, then implementations should deal with that. And this may force implementations to look at timestamp and figure out the current time zone name based on that.

But that's still not 100% clear - if an event was recorded in January and has `geo.timezone: CET`, but you look at the event in June, is it then correct? Or shouldn't it be `CEST`? Probably `CET` is correct, but are we 100% sure?

What I like about the [nodatime link](https://nodatime.org/TimeZones) is that it contains a list of official names and to me it seems you can (almost?) always populate `timezone` with a city based name (e.g. `Europe/Vienna`). This enables reporting the time-zone of the event in a format where the implementation does not need to care about daylight saving. In the context of geo, I feel this is the best way to go.

At the same time, it's still a string field, so if some implementation wants to go further and always report the common tz abbreviation, it can do so. But I agree, this introduces some ambiguity.

Given the context here, I'd vote for leaving it as is.

Alternatively we could also drop this field for now, unless someone has a specific use-case for it. I'd argue this field is actually redundant: I'd bet most (all?) implementations would just calculate the time-zone based on some other fields (e.g. `geo.location.lat/lon` or `geo.city.name`). Those are the more important fields here and this calculation can then happen later in the backend or even in the UI if someone wants to have a feature that shows the time-zone of the event.

Happy to hear other opinions.

**Comment by lmolkova** [comment]
Thanks again for the context @AlexanderWert  and @trisch-me !

So maybe we can move forward by adding some disclaimer into the group brief like 

> Note:
> Geo attributes are typically used under another namespace, such as `client.*` and describe the location of the corresponding entity (device, end-user, etc). Semantic conventions that reference `geo` attributes (as a root namespace) or embed them (under their own namespace) SHOULD document what `geo` attributes describe in the scope of that convention.

?

Also, it would be great if we could start embedding some common attributes under `client`, `source` or other namespaces.

File: model/registry/geo.yaml [hunk 11]
```diff
@@ -0,0 +1,113 @@
+groups:
+  - id: registry.geo
+    type: attribute_group
+    prefix: geo
+    brief: >
+      Geo fields can carry data about a specific location related to an event.
+      This geolocation information can be derived from techniques such as Geo IP, or be user-supplied.
+    attributes:
+      - id: city.name
```
lmolkova, line 9: would we have a dedicated field for village/town/county name? 

Can we look into popular Maps API and see how this and other geo attributes would align with their abstractions?

E.g. [Google Maps](https://developers.google.com/maps/documentation/geocoding/requests-geocoding) or [Azure Maps](https://learn.microsoft.com/en-us/rest/api/maps/search/get-search-address?view=rest-maps-1.0&tabs=HTTP#geographicentitytype)

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+        brief: >
+          Region ISO code ([ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2)).
+        examples: [ 'CA-QC' ]
+      - id: region.name
+        stability: experimental
+        type: string
+        brief: >
+          Region name.
+        examples: [ 'Quebec' ]
+      - id: timezone
```
lmolkova, line 108: I'd expect timezone to be a thing on it's own. Someone can report timezone without geo context. E.g. when scheduling a meeting, the event that describes the meeting can have a timezone, but no physical location.

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+          Latitude of the geo location in [WGS84](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84).
+        examples: [ 45.505918 ]
+      - id: name
+        stability: experimental
+        type: string
+        brief: >
+          User-defined description of a location.
+        note: >
+          User-defined description of a location, at the level of granularity they care about.
+          Could be the name of their data centers, the floor number, if this describes a local physical entity, city names.
```
lmolkova, line 86: we have dedicated attributes for data center names. I assume there could be other attributes added in the future that could intersect with this one.

Depending on the signal it can be more suitable on  a span name, log payload field. There is also no instrumentation that's expected to add it.

Given all these potential issues, could we postpone adding this attribute until we have an actual use-case?

model/registry/geo.yaml
gregkalapos: There was a discussion on this in https://github.com/open-telemetry/semantic-conventions/pull/1116#discussion_r1662310967.

Let's handle it separately - I remove this attribute from the PR.

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+              brief: 'North America'
+            - id: 'oc'
+              stability: experimental
+              value: 'OC'
+              brief: 'Oceania'
+            - id: 'sa'
+              stability: experimental
+              value: 'SA'
+              brief: 'South America'
+      - id: continent.name
```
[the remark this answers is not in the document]
gregkalapos: This PR is [based on ECS](https://github.com/open-telemetry/oteps/blob/main/text/0199-support-elastic-common-schema-in-opentelemetry.md). I tried to figure out the history of this, what I found is that `code` was added later here: https://github.com/elastic/ecs/pull/1229. But from the PR description I still don't really see why we need both.

Information stored in these 2 fields are clearly duplicated. The only upside I can see is that `continent.code` is an enum with fixed values, `continent.name` is a string. So implementation can decide if it wants to parse to the fixed enum values or leave it as string.

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+        brief: >
+          Longitude of the geo location in [WGS84](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84).
+        examples: [ -73.614830 ]
+      - id: location.lat
+        stability: experimental
+        type: double
+        brief: >
+          Latitude of the geo location in [WGS84](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84).
+        examples: [ 45.505918 ]
+      - id: name
```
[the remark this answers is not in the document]
gregkalapos: Since this is based on ECS, I'd prefer to not break compatibility on this field because of this. So I'd keep it as `name` for backwards compatibility with ECS.

model/registry/geo.yaml
```diff
@@ -0,0 +1,113 @@
...
+          Postal code associated with the location.
+          Values appropriate for this field may also be known as a postcode or ZIP code and will vary widely from country to country.
+        examples: [ '94040' ]
+      - id: region.iso_code
+        stability: experimental
+        type: string
+        brief: >
+          Region ISO code ([ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2)).
+        examples: [ 'CA-QC' ]
+      - id: region.name
```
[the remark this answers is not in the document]
gregkalapos: From what I found, ECS added those because the [geoip processor collects](https://www.elastic.co/guide/en/elasticsearch/plugins/current/using-ingest-geoip.html#using-ingest-geoip) both: https://github.com/elastic/ecs/issues/177.

I think this is very similar to https://github.com/open-telemetry/semantic-conventions/pull/1116#discussion_r1668319586: information stored in both attributes is definitely redundant with the benefit of fixed values ([ISO 3166-2 code) vs. not needing to parse; and with having both, implementation can just choose.

model/registry/geo.yaml
gregkalapos: Removed this field for now and opened https://github.com/open-telemetry/semantic-conventions/issues/1213 to discuss.

lmolkova:
related: https://github.com/open-telemetry/semantic-conventions/issues/1228

model/registry/geo.yaml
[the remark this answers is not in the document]
lmolkova: do we need to add both attributes? can we add only those that we'd use in the collector processor?

model/registry/geo.yaml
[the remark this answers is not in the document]
lmolkova: would this attribute be used by the collector processor? If not and we don't have a use-case for it in otel yet, could we postpone adding it?

model/registry/geo.yaml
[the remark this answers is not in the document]
lmolkova: same here, let's not add attributes that we don't have a use-case for yet. Especially if they are controvercial.

model/registry/geo.yaml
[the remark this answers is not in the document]
gregkalapos: Yes, both are used by the collector processor from what I can see: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33451/files#diff-0e91371ea4833d69a2fe3ac9e0ccab8db487e56af914d1717b0763a973eaffd4R84-R85

model/registry/geo.yaml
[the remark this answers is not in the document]
gregkalapos: This one I don't see being used - @rogercoll do you know if this field is used currently? If not, I'm ok to remove this.

model/registry/geo.yaml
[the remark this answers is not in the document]
gregkalapos: These are used by the collector processor: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33451/files#diff-0e91371ea4833d69a2fe3ac9e0ccab8db487e56af914d1717b0763a973eaffd4R92-R93

model/registry/geo.yaml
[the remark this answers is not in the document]
rogercoll: This field is not used in the collector processor atm, neither in the Elastic's geoip processor: https://www.elastic.co/guide/en/elasticsearch/reference/current/geoip-processor.html

I agree to postpone the addition until we find a use case :+1:

model/registry/geo.yaml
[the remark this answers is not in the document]
lmolkova: does it need to record both? would users appreciate having both and paying more for the same data?

That's exactly the reason I'm pushing back on adding duplicative attributes - when people see a list of attributes, they populate them all. 

Let's pick one and add it. Collector process should remove the other one unless some good reason to populate both.

Suggesting iso_code since it's a standard. Backends can maintain a mapping to human-friendly names for all continents.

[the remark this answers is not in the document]
rogercoll: > does it need to record both? would users appreciate having both and paying more for the same data?

That's a good point. The collector tends to disable metrics that can be derived by others in its default configurations, the same could be applied to attributes. The user would need to explicitly change the processor configuration to populate this attribute `region.name::enabled: true`.

Nonetheless, I support only including the `iso_code` attribute for this initial PR and leave the human-friendly name for a future PR or when we have an explicit use case. If needed, we could provide a `transformer` processor configuration to do a sample mapping.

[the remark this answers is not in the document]
gregkalapos: Thanks for the input. Removed `region.name`.

model/registry/geo.yaml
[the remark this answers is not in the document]
gregkalapos: Thanks - removed `name`.

model/registry/geo.yaml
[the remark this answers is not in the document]
gregkalapos: Similarly to `region` - here i removed `continent.name` and kept `continent.iso_code`.

model/registry/geo.yaml
gregkalapos: https://github.com/open-telemetry/semantic-conventions/pull/1116#discussion_r1697244285

gregkalapos:
Failing CI seems to be unrelated:

```
4m 56s
Run make markdown-link-check
semantic-conventions@ /home/runner/work/semantic-conventions/semantic-conventions
└── markdown-link-check@3.11.2


make: *** [Makefile:66: markdown-link-check] Error 1
Error: Process completed with exit code 2.
```

rogercoll:
I reckon this PR is still on review (unstale)

model/registry/geo.yaml
```diff
@@ -0,0 +1,84 @@
+groups:
+  - id: registry.geo
+    type: attribute_group
+    prefix: geo
```
lmolkova, line 4: we no longer use prefix in semconv - should now be part of each attribure name

```suggestion
```

model/registry/geo.yaml
```diff
@@ -0,0 +1,84 @@
...
+    brief: >
+      Geo fields can carry data about a specific location related to an event.
+      This geolocation information can be derived from techniques such as Geo IP, or be user-supplied.
+
+      Note: Geo attributes are typically used under another namespace, such as client.* and describe the location of the corresponding
+      entity (device, end-user, etc). Semantic conventions that reference geo attributes (as a root namespace) or embed them
+      (under their own namespace) SHOULD document what geo attributes describe in the scope of that convention.
+
+    attributes:
+      - id: city.name
```
lmolkova, line 14:
```suggestion
      - id: geo.city.name
```

(and in other places)

model/registry/geo.yaml
lmolkova: let's either remove the attribute for now or consider more general term. Azure maps use `municipality`, google uses `locality`.

ChatGPT also prefers `locality` 

> Yes, the official term for the name of a city, town, village, or similar populated place is often referred to as a "locality" or "place name." In some contexts, it's also called a "geographic name" or "toponym."
>
> If you’re working with geographic data, “locality” is commonly used to encompass all such divisions, but “place name” is frequently used in casual contexts.

lmolkova (approved):
LGTM with some suggestions. The key part is `geo.city.name` (which I'm suggesting to remove or rename to `geo.locality.name`

gregkalapos:
> LGTM with some suggestions. The key part is `geo.city.name` (which I'm suggesting to remove or rename to `geo.locality.name`

Agreed. ECS calls it `city_name`, so it's a bit unfortunate to see that this will diverge from ECS; on the other hand, I also wasn't satisfied  with `city` - calling this `locality` is clearly an improvement. I'd be ok to proceed with that, so I kept the field and changed it to `locality`.

gregkalapos:
:( 
```
```

gregkalapos:
I think this is ready to be merged - I'm not authorized, so I can't do so.

Could someone please hit merge?

Issue #1032, opened 2024-05-13
Define semantic conventions for k8s metrics


ChrsMark:
### Area(s)

area:k8s

### Is your change request related to a problem? Please describe.

At the moment there are not Semantic Conventions for k8s metrics. 

### Describe the solution you'd like

Even if we cannot consider the k8s metrics as stable we can start considering adding metrics that are not controversial to get some progress here.  This issue aims to collect the existing k8s metrics that exist in the Collector and keep track of any related work.
Bellow I'm providing an initial list with metrics coming from the `kubeletstats` and `k8scluster` receivers. Note that these are matter to change with time being so we should get back to the Collector to verify the current state.

cc: @open-telemetry/semconv-k8s-approvers 

### kubeletstats metrics

 `k8s.node.cpu.usage`
 `k8s.node.cpu.utilization`
 `k8s.node.cpu.time`
 `k8s.node.memory.available`
 `k8s.node.memory.usage`
 `k8s.node.memory.rss`
 `k8s.node.memory.working_set`
 `k8s.node.memory.page_faults`
 `k8s.node.memory.major_page_faults`
 `k8s.node.filesystem.available`
 `k8s.node.filesystem.capacity`
 `k8s.node.filesystem.usage`
 `k8s.node.network.io`
 `k8s.node.network.errors`
 `k8s.node.uptime`
 `k8s.pod.cpu.usage`
 `k8s.pod.cpu.utilization`
 `k8s.pod.cpu.time`
 `k8s.pod.memory.available`
 `k8s.pod.memory.usage`
 `k8s.pod.cpu_limit_utilization`
 `k8s.pod.cpu_request_utilization`
 `k8s.pod.memory_limit_utilization`
 `k8s.pod.memory_request_utilization`
 `k8s.pod.memory.rss`
 `k8s.pod.memory.working_set`
 `k8s.pod.memory.page_faults`
 `k8s.pod.memory.major_page_faults`
 `k8s.pod.filesystem.available`
 `k8s.pod.filesystem.capacity`
 `k8s.pod.filesystem.usage`
 `k8s.pod.network.io`
 `k8s.pod.network.errors`
 `k8s.pod.uptime`
 `container.cpu.usage`
 `container.cpu.utilization`
 `container.cpu.time`
 `container.memory.available`
 `container.memory.usage`
 `k8s.container.cpu_limit_utilization`
 `k8s.container.cpu_request_utilization`
 `k8s.container.memory_limit_utilization`
[...]

PR #282, opened 2023-08-24
Add container metric fields (from ECS)

ChrsMark:
This PR adds container related metrics fields as part of https://github.com/open-telemetry/semantic-conventions/issues/72.

Continues https://github.com/open-telemetry/semantic-conventions/pull/87 on top of the project restructuring. 

Based on the comparison between ECS and Otel collector the result is the following (see https://github.com/open-telemetry/semantic-conventions/pull/282#issuecomment-1757735366). 

### Summary of added fields

1) `container.cpu.time` (ok for the Collector implementation: https://github.com/open-telemetry/semantic-conventions/pull/282#issuecomment-1909118310)
2) `container.memory.usage` (more accurate and aligned with system metrics) 
3) `container.disk.io` which has as dimension the already existent `disk.io.direction` (Otel collector only provides the Linux specific metric and hence a generic one capable to cover Windows as well would be better)
4) `container.network.io` which has as dimension the already existent `network.io.direction` (better alignment with disko.io and use of dimension).

All of those also refer to `container.id` as attribute as well.

Issue #226, opened 2023-08-02
Document the difference between a host and a system metric


frzifus:
This issue is intended to discuss the difference between the host and system namespace. The outcome should be a clear description of when which ns is used.

This question was raised during the [Aug 2, 2023, 08:30 PT SIG MTG](https://docs.google.com/document/d/1p5TH57t43XpxA48onLzX4PIr3g6ydYKCtR_AUlsCnQk/edit#heading=h.ldhv1fb8w5d) and needs to be answered to continue on this [PR](https://github.com/open-telemetry/semantic-conventions/pull/209#discussion_r1279445608).

mx-psi:
cc @ChrsMark

trask:
just in case it helps, there was a somewhat related prior discussion here: https://github.com/open-telemetry/opentelemetry-specification/pull/2388

ChrsMark:
We had a discussion about this during the August 30th System Semantic Convention WG meeting. The main points were the following:

1) Keep using `host.*` namespace for resource attributes like `host.id`, `host.name` etc. This means that the [PR](https://github.com/open-telemetry/semantic-conventions/pull/209) which only adds "static" host's resource attributes is good to go from this perspective.
2) For metrics we can keep using the `system.*` namespace whenever we collect metrics from the system directly.
3) For centralized metrics that are collected from outside of the resource's scope (ie container metrics collected centrally for a k8s cluster) we can use resource specific namespace like for example `container.*`, `pod.*`, `vm.*`, `instance.*` etc. Containers, VMs etc are actually processes for a bare metal host similarly to what a `jvm` process is. This is also explained at https://github.com/open-telemetry/semantic-conventions/pull/87#discussion_r1282885340 and aligns with https://github.com/open-telemetry/opentelemetry-specification/pull/2388#issuecomment-1057120767.

@mx-psi @dmitryax @kaiyan-sheng feel free to correct me if I miss anything here. If we agree on this we can maybe transfer this conclusion somewhere for future reference (and then close this issue).

cc: @trask

mx-psi:
Thanks for the summary @ChrsMark! I think there is agreement within the System semantic conventions WG. Given there has not been any disagreement voiced on this issue, I think we can open a PR to put this in writing somewhere: for (1) and (2) maybe on the README.md of the relevant folders on this repository, while I am not sure as to where to put (3)

ChrsMark:
[...]

PR #1128, opened 2024-06-06
Add container.cpu.usage metric

ChrsMark:

## Changes

This PR adds the `container.cpu.usage` metric.
This metric is implemented by the [`kubeletstats`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.102.0/receiver/kubeletstatsreceiver/documentation.md#containercpuusage) receiver.
This value comes from the `stats/summary` endpoint of Kubelet and its calculated at https://github.com/kubernetes/kubernetes/blob/v1.30.0/pkg/kubelet/stats/cri_stats_provider.go#L791-L800.

It's not implemented by the [`dockerstats`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/dockerstatsreceiver/documentation.md) receiver but that would be doable with something similar to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.102.0/receiver/dockerstatsreceiver/metric_helper.go#L30, using the formula from the Kubelet's stats mentioned above.

ref: https://docs.docker.com/engine/api/v1.45/#tag/Container/operation/ContainerStats

/cc @open-telemetry/semconv-container-approvers @open-telemetry/semconv-k8s-approvers


trisch-me (approved):

model/metrics/container.yaml
```diff
@@ -14,6 +14,20 @@ groups:
         brief: "The CPU state for this data point. A container SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels."
         requirement_level: opt_in
 
+  - id: metric.container.cpu.usage
```
lmolkova, line 17: what would be the reason to record it as a `container.cpu.usage` and not one of the existing metrics in `system.cpu.*`? 

I understand there are other metrics in `container` namespace already, and it's be nice to clarify whether we need different ones and if we do, then document why.

model/metrics/container.yaml
```diff
@@ -14,6 +14,20 @@ groups:
...
+    type: metric
+    metric_name: container.cpu.usage
+    stability: experimental
+    brief: "CPU usage"
+    note: >
+      CPU usage of the specific container on all available CPU cores, averaged over the sample window
+    instrument: gauge
+    unit: "{cpu}"
```
lmolkova, line 25: is this right? should it be time? ticks? percentage?

model/metrics/container.yaml
```diff
@@ -14,6 +14,20 @@ groups:
         brief: "The CPU state for this data point. A container SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels."
         requirement_level: opt_in
 
+  - id: metric.container.cpu.usage
+    type: metric
+    metric_name: container.cpu.usage
+    stability: experimental
+    brief: "CPU usage"
+    note: >
+      CPU usage of the specific container on all available CPU cores, averaged over the sample window
```
lmolkova, line 23: is it normalized per CPU or sum across all of them?

model/metrics/container.yaml
```diff
@@ -14,6 +14,20 @@ groups:
         brief: "The CPU state for this data point. A container SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels."
         requirement_level: opt_in
 
+  - id: metric.container.cpu.usage
+    type: metric
+    metric_name: container.cpu.usage
+    stability: experimental
+    brief: "CPU usage"
```
lmolkova, line 21: could we clarify a range of this value in the note?

model/metrics/container.yaml
```diff
@@ -14,6 +14,20 @@ groups:
...
+    note: >
+      CPU usage of the specific container on all available CPU cores, averaged over the sample window
+    instrument: gauge
+    unit: "{cpu}"
+    attributes:
+      - ref: container.cpu.state
+        brief: "The CPU state for this data point. A container SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels."
```
lmolkova, line 28: > A container SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels."

Could it be clarified?
cpu state does not characterizes a container, but a state of the CPU. Also, does it mean that cpu.state should be added on all data points or not applied at all for a given container?

docs/system/container-metrics.md
```diff
@@ -54,6 +54,57 @@ This metric is [opt-in][MetricOptIn].
...
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`container.cpu.state`](/docs/attributes-registry/container.md) | string | The CPU state for this data point. A container SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels. | `user`; `kernel` | `Opt-In` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+`container.cpu.state` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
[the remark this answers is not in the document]
lmolkova: I think `cpu.mode` would be more accurate, but we can follow up in the https://github.com/open-telemetry/semantic-conventions/issues/1139

model/metrics/container.yaml
ChrsMark: [kubeletstats](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.102.0/receiver/kubeletstatsreceiver/documentation.md#containercpuusage) reports the total on all available CPUs, does not break down per cpu.
[dockerstats](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/dockerstatsreceiver/documentation.md) does the same but maybe in the future it can become more granular since [dockerAPI](https://docs.docker.com/engine/api/v1.45/#tag/Container/operation/ContainerStats) can provide per cpu metrics. In any case, if we'd like to change it the future we can add the `cpu.logical_number` attribute as `opt-in` similarly to [`cpu.state`](https://github.com/open-telemetry/semantic-conventions/pull/1128#discussion_r1634449851)? (that was also discussed at https://github.com/open-telemetry/semantic-conventions/pull/282#issuecomment-1909118310).

model/metrics/container.yaml
ChrsMark: Based on the implementation from [kubeletstats](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.102.0/receiver/kubeletstatsreceiver/documentation.md#containercpuusage) it is `{cpu}`, see https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/25901#issuecomment-1885979641 for more context.

model/metrics/container.yaml
ChrsMark: It is documented at https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/docs/system/system-metrics.md?plain=1#L15-L17.

lmolkova: thanks! would you mind adding a similar statement to container metrics (that they are not the same as system ones)?

do you know why - would system metrics be different than container metrics? or it's a way to distinguish them if both sets of metrics are reported?

docs/system/container-metrics.md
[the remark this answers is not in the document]
TylerHelmuth: Is there a conclusion from https://github.com/open-telemetry/semantic-conventions/issues/840 that we'll stick with `container.cpu.state`?

model/metrics/container.yaml
ChrsMark: Mostly a way to distinguish them since a container (or Pod) is not a "system"(=="host") neither a standalone process (it's a group of processes, aka cgroup). I wouldn't be surprised if some API specific metrics might differ and extra transformations would be needed to make them match the `system.*` metrics, so that would be an extra pain/hassle for the implementations I guess. Also it depends from which point of view we observe a container. If we run a collector inside a container treating it as a host then from this point of view the container is a "host" and hence `system.*` metrics should be collected using the `hostmetrics` receiver. When we observe containers/pods from outside, using a centralized API, we treat them mostly as "processes" (but strictly speaking a container can include more than 1 process). Since those entities are technology specific, we decided back then that for centralized metrics that are collected from outside of the resource's scope (ie container metrics collected centrally for a k8s cluster) we use resource specific namespace like `container.*` or `k8s.pod`.
I think this is somehow similar to the `jvm.*` namespace.

Some relevant discussions, issues:
- https://github.com/open-telemetry/semantic-conventions/issues/226#issuecomment-1700529192
- https://github.com/open-telemetry/opentelemetry-specification/pull/2388#issuecomment-1057120767

lmolkova: thank you for the context!

docs/system/container-metrics.md
```diff
@@ -8,6 +8,10 @@ linkTitle: Container
 
 ## Container Metrics
 
+This document describes instruments and attributes for common container level
+metrics in OpenTelemetry. Collected from technology-specific,
+well-defined APIs (e.g. Kubelet's API or container runtimes).
```
heyams, line 13:
```suggestion
metrics in OpenTelemetry. These metrics are collected from technology-specific,
well-defined APIs (e.g. Kubelet's API or container runtimes).
```

model/metrics/container.yaml
lmolkova: In all other utilization metrics we use `1` as a unit. Having `{cpu}` there creates an impression that we measure number of CPUs.

model/metrics/container.yaml
lmolkova: So if I have 10 CPUs that are busy 100% of time, would this metric show 100% (1) or 1000% (10) ? Could you please clarify the brief so it's clear to the future readers. See my other comment on the value range as well.

model/metrics/container.yaml
```diff
@@ -11,7 +11,21 @@ groups:
...
+    metric_name: container.cpu.usage
+    stability: experimental
+    brief: "Container's CPU usage, measured in cpus. Range from 0 to the number of allocatable cpus"
+    note: >
+      CPU usage of the specific container on all available CPU cores, averaged over the sample window
+    instrument: gauge
+    unit: "{cpu}"
+    attributes:
+      - ref: container.cpu.state
+        brief: "The CPU state for this data point. A container's cpu usage SHOULD be characterized _either_ by data points with no `state` labels, _or only_ data points with `state` labels."
```
lmolkova, line 28: maybe (to make it a bit more clear, also feel free to move it back to brief if you like it better)
```suggestion
        brief: "The CPU state for this data point."
        note: >
          `container.cpu.state` SHOULD be reported consistently on all measurements if it's reported at all.
```

model/metrics/container.yaml
```diff
@@ -11,7 +11,21 @@ groups:
...
         requirement_level: opt_in
```
lmolkova, line 29: Why not `recommended`? Having one dimension with 3 possible values on the metric should be fine.

model/metrics/container.yaml
lmolkova: In other places we call a similar metric `*.cpu.utilization`  and measure the normalized one (divided by cpu count) and in a [0, 1] range.  (see JVM, process, system metrics)

It'd be inconsistent if we had a different name and measured a sum across all CPUs here. 

We also report `*.cpu.count` in other places (as a metric, it has high cardinality and can't be used as an attribute).
The `cpu.time` and `cpu.count` metrics can be used together to calculate utilization. Utilization is still reported as a relatively cheap convenience.

By introducing non-normalized `cpu.usage` in addition to `cpu.time` here we provide no way for users to calculate normalized thing and create any alerts.

model/metrics/container.yaml
ChrsMark: As explained also at https://github.com/open-telemetry/semantic-conventions/pull/1128#discussion_r1634449851, `kubeletstats` cannot distinguish the state/mode while the `dockerstats` can. That's why we marked it as opt-in when we [introduced that](https://github.com/open-telemetry/semantic-conventions/pull/282).

model/metrics/container.yaml
ChrsMark: I'm fine with changing this, however we define it like that in [system](https://github.com/open-telemetry/semantic-conventions/blob/main/model/metrics/system-metrics.yaml#L12) and [process](https://github.com/open-telemetry/semantic-conventions/blob/main/model/metrics/process-metrics.yaml#L12) metrics. Do you think diverging them would be ok?

model/metrics/container.yaml
ChrsMark: But this is not a utilization metric, this is usage metric.
Correct me if I'm wrong here but the intention is to measure the number of CPUs used by the container over a reporting interval. That means essentially that if a container uses 2 complete CPUs constantly during the internal, the metric would be `2`. 
Then if we divide this with the number of Node's cores for example (ie `4` cores) it would give us `2/4=0.5` utilization.
In `kubeletstats` we have other metrics specifically for reporting the utilization against the container's limits and/or the Node's capacity.

cc @TylerHelmuth @dmitryax @povilasv who were involved in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/25901.

model/metrics/container.yaml
ChrsMark: > So, the suggestion:
>
> add container.cpu.logical_count or similar metric
> add container.cpu.utilization metric (in [0,1] range) instead of container.cpu.usage

It has been preferred to use the `container.cpu.usage` in `kubeletstats` receiver after a change at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/25901.

As I also mention at https://github.com/open-telemetry/semantic-conventions/pull/1128#discussion_r1642699619 the `*utilization` metrics can be calculated in various different ways in K8s against various different limits. For sure we are planning to add those `*utilization` metrics as SemConvs as part of the https://github.com/open-telemetry/semantic-conventions/issues/1032. 
So I guess that `container.cpu.usage` can be added independently of the `*utilization` ones.

docs/system/container-metrics.md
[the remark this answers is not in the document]
ChrsMark: Switched to `cpu.mode` since https://github.com/open-telemetry/semantic-conventions/issues/840 was merged. We can resolve this thread I guess.

model/metrics/container.yaml
dmitryax: Correct. We use `utilization` when there is a limit available and count it as a percentage [0,1]. Aggregation of those metrics would be `avg` which result in the same units `1`.

Here, we have an average usage of CPUs. If we had `cpu.logical_number` attribute from docker, the aggregation method would be `sum`, which gives us the same `{cpu}` unit.

dmitryax (approved):

ChrsMark:
@open-telemetry/semconv-k8s-approvers @open-telemetry/semconv-container-approvers @open-telemetry/specs-semconv-approvers could you take a look?

model/metrics/container.yaml
jsuereth: You could also use conditionally required (where condition is dockerstats)

model/metrics/container.yaml
```diff
@@ -15,6 +15,21 @@ groups:
         note: "Following states SHOULD be used: `user`, `system`, `kernel`"
         requirement_level: opt_in
 
+  - id: metric.container.cpu.usage
+    type: metric
+    metric_name: container.cpu.usage
+    stability: experimental
+    brief: "Container's CPU usage, measured in cpus. Range from 0 to the number of allocatable cpus"
```
mx-psi, line 22:
```suggestion
    brief: "Container's CPU usage, measured in cpus. Range from 0 to the number of allocatable CPUs"
```

mx-psi (approved):

model/metrics/container.yaml
ChrsMark: Thank's for the suggestion. Added at [9bc986b](https://github.com/open-telemetry/semantic-conventions/pull/1128/commits/9bc986b361db6144c768d53590fed1ac86a757f3). For some reason it has no impact on the MD table though 🤔 (even after running `make fix` or `make table-generation `).

ChrsMark:
@open-telemetry/specs-semconv-approvers since there are quite a few approvals by the k8s/container approvers can we move this one forward?

joaopgrassi (approved):

ChrsMark:
Thank's @joaopgrassi! My only concern is about https://github.com/open-telemetry/semantic-conventions/pull/1128#discussion_r1686534350. Does this look normal?

model/metrics/container.yaml
joaopgrassi: We spoke on slack, @lquerel will investigate what looks like a bug in the tooling.

ChrsMark: I think we can move this forward by changing the requirement level in the `container.cpu.time` as well, which is something needed anyways. And the issue in the tooling can be investigated independently.

docs/system/container-metrics.md
```diff
@@ -42,10 +46,74 @@ This metric is [opt-in][MetricOptIn].
...
+**[1]:** CPU usage of the specific container on all available CPU cores, averaged over the sample window
+
+
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+<!-- semconv metric.container.cpu.time(full) -->
```
joaopgrassi, line 102: This is still wrong, as @lquerel spotted. @ChrsMark curious now, did you add this manually or was it produced by the markdown generation? We aren't supposed to edit these markdown files manually anymore so maybe that was it? 🤔

ChrsMark: Good catch, I didn't notice that line. It was automatically generated by the tooling. I just changed the `requirement_level` of the `container.cpu.time` metric.
I see, we need to block this until the tooling is fixed.

joaopgrassi: Credits to @lquerel good eyes 😅

lmolkova: wait, why are we not supposed to edit these files? It's not the registry, so we added `container.cpu.time` and the fix would be

```suggestion
<!-- semconv metric.container.cpu.usage(full) -->
```

what am I missing?

model/metrics/container.yaml
lmolkova: As I mentioned in the other comment, there seem to be no issue on tooling side, we just need to use proper group id (related to cpu.usage and not time)

lmolkova (approved):
Please fix anchor name and this PR should be good to go.

docs/system/container-metrics.md
lmolkova: no worries! It seems everything is fixed now.

PR #1146, opened 2024-06-11
Add authentication sub-namespace to user

### LINKED ISSUE #1104: User.id for authenticated user id
### Area(s)

area:user

enduser.id has been deprecated and replaced with user.id. #731 

enduser.id had this old description: 
image

user.id has this new description: 
image

The new description is confusing now. Is it for authenticated user id or anonymous user id? What are your thoughts on creating a new attribute called `user.anonymous_id`?

Our telemetry solution tracks both authenticated user id and anonymous user id.

1. update the user.id description to make it clear that it's intended for authenticated user id.
2. create a new attribute call `user.anonymous_id` for the anonymous user id.

n/a

### Additional context

n/a

heyams:
Fix #1104

I chose `user.anonymous_id` because of https://github.com/open-telemetry/semantic-conventions/pull/1121


model/registry/user.yaml
```diff
@@ -28,8 +28,14 @@ groups:
...
         brief: >
-          Unique identifier of the user.
+          Unique identifier of an authenticated user.
         examples: ['S-1-5-21-202424912787-2692429404-2351956786-1000']
+      - anonymous_id
+        type: string
+        stability: experimental
+        brief: >
+          Unique identifier of an anonymous user who interacts with a system or service without providing personal information or authentication credentials. It allows the system to track user activity while maintaining anonymity.
+        examples: ['QdH5CAWJgqVT4rOr0qtumf']
```
MSNev, line 38: It may also be a guid

model/registry/user.yaml
```diff
@@ -28,8 +28,14 @@ groups:
...
         stability: experimental
         brief: >
-          Unique identifier of the user.
+          Unique identifier of an authenticated user.
         examples: ['S-1-5-21-202424912787-2692429404-2351956786-1000']
+      - id: anonymous_id
+        type: string
+        stability: experimental
+        brief: >
+          Unique identifier of an anonymous user who interacts with a system or service without providing personal information or authentication credentials. It allows the system to track user activity while maintaining anonymity.
```
trisch-me, line 37: please split the line to fix lint

model/registry/user.yaml
trisch-me: we could add this as a note

model/registry/user.yaml
AlexanderWert: How is `user.anonymous_id` different to the already existing `user.hash` attribute?

From the description both attributes try to achieve the same.

AlexanderWert (changes requested):

model/registry/user.yaml
heyams: `user.hash` is a hash of user.id or user.name. anonymous user id is a completely separate kind of user id, not just anonymizing a well-known user id. For users who have neither a user.id nor a user.name, anonymous user id is still can be tracked via browsers. hope that makes sense.

heyams: Added this to note.

scheler (approved):

model/registry/user.yaml
```diff
@@ -28,8 +28,17 @@ groups:
...
         stability: experimental
         brief: >
-          Unique identifier of the user.
+          Unique identifier of an authenticated user.
         examples: ['S-1-5-21-202424912787-2692429404-2351956786-1000']
+      - id: anonymous_id
+        type: string
+        stability: experimental
+        brief: >
+          Unique identifier of an anonymous user who interacts with a system or service without providing personal information or authentication credentials.
```
reyang, line 37: Would anonymous_id and user_id be mutually exclusive or not for a single event?
For example, if a user started installing an app, once the app is installed, the user provided credential and logged in, do we expect the telemetry to have BOTH anonymous_id and user_id after the successful login?

OsvaldoRosado: I think they cannot be mutually exclusive, otherwise you cannot calculate correct user counts like MAU for scenarios that need to count anonymous users. 

If mutually exclusive, `anonymous_id` will be missing users who are signed in with long term tokens (never became anonymous in the query period). On the other hand, trying to calculate MAU by `dcount(user_id)` + `dcount(anonymous_id)` will duplicate users who _did_ sign in during the query period.

reyang: I agree with you @OsvaldoRosado, the current wording sounds like user_id is for non-anonymous, and anonymous_id is for anonymous.

trisch-me: @OsvaldoRosado do I understand correctly that in your example we would have a link between anonymous and authenticated user? Because telemetry will provide both ids if available

trisch-me: @heyams how can we distinguish between not authenticated and non-anonymous user? If the user hasn't authenticate itself, do we have means to know if he is anonymous (doesn't have an account) or not anonymous (have account but hasn't logged in)?

heyams: I asked Copilot about the difference between authenticated user id, non-anonymous user id, and unauthenticated user id: 

```
Copilot
Certainly! Let’s break down the differences between authenticated user IDs, non-anonymous user IDs, and unauthenticated user IDs:

Authenticated User IDs:
Definition: Authenticated user IDs represent entities that have successfully proven their identity to the system. These entities are typically associated with user accounts or processes running in a specific user’s security context.
Unique Identifier: Each authenticated user has a unique Security Identifier (SID) issued by an authority (e.g., a Windows domain controller). The SID is stored in a security database.
Access Tokens: When a user logs in, the system creates an access token containing the user’s SID, user rights, and group SIDs. This token defines the security context for the user’s actions.
Examples: Regular user accounts, computer accounts, and other authenticated entities fall into this category.
Non-Anonymous User IDs:
Definition: Non-anonymous user IDs refer to known entities that haven’t undergone full authentication. They may have limited rights on the system.
Examples:
Guest Account: The Guest account, even if it has a password, remains non-anonymous because it lacks full authentication.
Built-in Accounts: Certain built-in accounts (e.g., LOCAL_SERVICE) are also non-anonymous.
Note: Non-anonymous users don’t need credentials to log in, but their identities are still known.
Unauthenticated User IDs:
Definition: Unauthenticated user IDs represent entities with no verified identity. These are often transient or anonymous.
Examples:
Anonymous Users: Visitors accessing public resources without logging in.
Temporary Tokens: Tokens issued for short-term access (e.g., OAuth tokens before authentication).
Note: Unauthenticated IDs lack a persistent association with a specific user.
In summary, authenticated user IDs are fully verified, non-anonymous user IDs are known but unauthenticated, and unauthenticated user IDs lack any verified identity. 
```

OsvaldoRosado: > @OsvaldoRosado do I understand correctly that in your example we would have a link between anonymous and authenticated user? Because telemetry will provide both ids if available

Yes you would. "Anonymous" is arguably a confusing term here. I see it as basically a uniquely identifying user id that is consistently available for all users, regardless of authentication state.

heyams: @trisch-me I've updated the description of `anonymous_id` for better clarity. I hope this will alleviate some of the confusion.

thompson-tomo:
I would suggest we look to finalise the discussion in #1104 especially around scenarios & use cases prior to merging this request.

heyams:
> This PR was marked stale due to lack of activity. It will be closed in 7 days.

please keep open.

heyams:
@AlexanderWert can you please join next Monday's semcon SIG to discuss this? I will be there.

model/registry/user.yaml
```diff
@@ -28,8 +28,16 @@ groups:
...
         stability: experimental
         brief: >
-          Unique identifier of the user.
+          Unique identifier of an authenticated user.
         examples: ['S-1-5-21-202424912787-2692429404-2351956786-1000']
+      - id: anonymous_id
+        type: string
+        stability: experimental
+        brief: >
+          Identifier for a user who interacts with a system. This identifier may be unique only through best-effort means and does not imply that the user is authenticated to the system.
```
heyams, line 37: @trisch-me can you comment on this description? 
I can update it to include a condition like `when auth context is present` it should be treated as authenticated user id. something like that.

let's improve this description.

trisch-me: I have updated original id to represent auth context

trisch-me:
@heyams let's wait for outcome of the discussion in the related issue 
https://github.com/open-telemetry/semantic-conventions/issues/1104 before continuing on this PR

model/registry/user.yaml
```diff
@@ -33,16 +33,11 @@ groups:
...
-          - id: id
-            type: string
-            brief: "Unique identifier of an authenticated user in the system."
-            examples: ['S-1-5-21-202424912787-2692429404-2351956786-1000']
-            stability: experimental
+      - id: authentication.id
+        type: string
+        brief: "Unique identifier of an authenticated user in the system."
+        examples: ['S-1-5-21-202424912787-2692429404-2351956786-1000']
+        stability: experimental
```
heyams, line 40: @trisch-me i tried to use nested `attribute_group` to represent new sub-namespace `authentication`. `make` command didn't like it and failed to update markdown accordingly. any suggestion?

.chloggen/add_authentication_user_subnamespace.yaml
```diff
@@ -0,0 +1,23 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
```
heyams, line 7: should this be "breaking" for users who replaced `user.id` for the deprecated `enduser.id`?

heyams:
i can't validate chlolg yaml locally.. stuck with this 

image

model/registry/user.yaml
trisch-me: hey, I'm not sure what have you tried to do? The way I see it in final yaml is good one. You don't need to extract 1 subgroup in it's own attribute_group

.chloggen/add_authentication_user_subnamespace.yaml
trisch-me: I think deprecated attributes should also be updated to use `user.authentication.id` instead of `user.id`
Also schema-next should reflect the same.

heyams: I have updated it. please check out general.yaml.

MSNev (approved):

model/registry/user.yaml
```diff
@@ -28,8 +28,15 @@ groups:
         type: string
         stability: experimental
         brief: >
-          Unique identifier of the user.
+          Identifier of a user who interacts with a system. This identifier may be unique only through best-effort means and does not imply that the user is authenticated to the system.
```
lmolkova, line 31: suggesting some clarifications:
```suggestion
          Identifies a user interactive with a system regardless of user authentication status. This identifier may be unique only through best-effort means.
        note: > 
          The `user.id`, when populated, is expected to be generated before user is authenticated and SHOULD NOT change after the user logs in. In browser scenarios `user.id` is usually stored in cookies.
          It's NOT RECOMMENDED to populate this attribute when unauthenticated users are not tracked or identified by the application.
```

model/registry/user.yaml
```diff
@@ -28,8 +28,15 @@ groups:
...
+        note: >
+          It can be a random guid or a hash of the user's IP address. This is different from `user.hash` which is a hash of a known `user.id` or `user.name`.
+        examples: ['QdH5CAWJgqVT4rOr0qtumf']
+      - id: user.authentication.id
+        type: string
+        brief: "Unique identifier of an authenticated user in the system."
```
lmolkova, line 37: I really like `authentication` since we can now use it to describe auth attempt/process and not just the result

```suggestion
        brief: "Unique identifier of an authenticated user in the system."
        note: >
          The `user.authentication.id` MAY be used to identify a user attempting to authenticate if it's known at this stage.
```

does it align with your undestanding?

.chloggen/add_authentication_user_subnamespace.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: user
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: introduce a new subnamespace `authentication` under `user` with a new attribute `user.authentication.id`
```
lmolkova, line 13: super-nit:

```suggestion
note: introduce subnamespace `user.authentication` with a new attribute `user.authentication.id`
```

(seems slightly easier to read)

docs/attributes-registry/user.md
```diff
@@ -10,13 +10,16 @@
...
+| `user.id`                | string   | Identifier of a user who interacts with a system. This identifier may be unique only through best-effort means and does not imply that the user is authenticated to the system. [2] | `QdH5CAWJgqVT4rOr0qtumf`                           | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `user.name`              | string   | Short name or login/username of the user.                                                                                                                                           | `a.einstein`                                       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `user.roles`             | string[] | Array of user roles at the time of the event.                                                                                                                                       | `["admin", "reporting_user"]`                      | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
 **[1]:** Useful if `user.id` or `user.name` contain confidential information and cannot be used.
```
lmolkova, line 23: we probably want to update this note to include `user.authenticated.id`.

But also given  that `user.id` is now random and not-authenticated one, we probably can remove the assumption that it may contain sensitive information.

docs/attributes-registry/user.md
```diff
@@ -10,13 +10,16 @@
...
+
+**[2]:** It can be a random guid or a hash of the user's IP address. This is different from `user.hash` which is a hash of a known `user.id` or `user.name`.
```
lmolkova, line 25: should `user.hash` be a hash of `user.authentication.id` now since it's the primary user id and is sensitive?

lmolkova:
@AlexanderWert could you please take another look? I believe your feedback was addressed.

docs/attributes-registry/user.md
heyams: i've updated it to 'user.authentication.id'

heyams:
image

i checked that link is still live. is this false negative?

model/general.yaml
```diff
@@ -34,15 +34,16 @@ groups:
...
     brief: >
-      These attributes may be used for any operation with an authenticated and/or authorized enduser.
+      These attributes may be used for any operation with an authenticated and/or authorized user.
     attributes:
-      - ref: enduser.id
-        requirement_level: recommended
-      - ref: enduser.role
+      - ref: user.id
+        requirement_level:
+          conditionally_required: If instrumentation supports tracking unauthenticated users and if `user.authentication.id` is set, recommended otherwise.
```
lmolkova, line 43: sorry, I think I got it wrong in my previous suggestion.

We don't want `user.id` to be used for authenticated users, so I guess it should be

```suggestion
          conditionally_required: If and only if instrumentation supports tracking unauthenticated users and if `user.authentication.id` is set.
```

wdyt?

lmolkova: the goal is to prevent someone from duplicating things like 

❌ Bad:

```
user.authentication.id = lmolkova
user.id = lmolkova
```

❌ Bad:

```
user.id = lmolkova
```

we want people to do 

✅  Good:

```
user.id = QdH5CAWJgqVT4rOr0qtumf
```

✅  Good:

```
user.id = QdH5CAWJgqVT4rOr0qtumf
user.authentication.id = lmolkova
```

✅  Good:

```
user.authentication.id = lmolkova
```

model/registry/user.yaml
```diff
@@ -22,14 +22,28 @@ groups:
...
         brief: >
-          Unique identifier of the user.
+          Identifies a user interacting with a system regardless of user authentication status. This identifier may be unique only through best-effort means.
+        note: >
+          The `user.id`, when populated, is expected to be generated before user is authenticated and SHOULD NOT change after the user logs in.
+          In browser scenarios `user.id` is usually stored in cookies.
+
+          It's NOT RECOMMENDED to populate this attribute when unauthenticated users are not tracked or identified by the system.
+
+          It can be a random guid or a hash of the user's IP address. This is different from `user.hash` which is a hash of a known `user.id` or `user.name`.
```
lmolkova, line 38: sorry, one more minor suggestion to rearrange things slightly (feel free to rephrase or not take it at all)

```suggestion
          The `user.id`, when populated, is expected to be generated before user is authenticated and SHOULD NOT change after the user logs in.

          It's NOT RECOMMENDED to populate this attribute when unauthenticated users are not tracked or identified by the system.

          In browser scenarios `user.id` is usually stored in cookies. It can also be a random guid or a hash of the user's IP address. This is different from `user.hash` which is a hash of a known `user.authentication.id` or `user.name`.
```

docs/attributes-registry/user.md
```diff
@@ -10,13 +10,20 @@
...
+| `user.roles`             | string[] | Array of user roles at the time of the event.                                                                                                           | `["admin", "reporting_user"]`                      | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1]:** The `user.authentication.id` MAY be used to identify a user attempting to authenticate if it's known at this stage.
+
+**[2]:** Useful if `user.authentication.id` or `user.name` contain confidential information and cannot be used.
+
+**[3]:** The `user.id`, when populated, is expected to be generated before user is authenticated and SHOULD NOT change after the user logs in. In browser scenarios `user.id` is usually stored in cookies.
+It's NOT RECOMMENDED to populate this attribute when unauthenticated users are not tracked or identified by the system.
+It can be a random guid or a hash of the user's IP address. This is different from `user.hash` which is a hash of a known `user.id` or `user.name`.
```
lmolkova, line 29: left another comment on yaml with suggestion that includes changing `uer.id` to `user.authentication.id` here

docs/attributes-registry/user.md
```diff
@@ -10,13 +10,20 @@
...
+| `user.roles`             | string[] | Array of user roles at the time of the event.                                                                                                           | `["admin", "reporting_user"]`                      | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1]:** The `user.authentication.id` MAY be used to identify a user attempting to authenticate if it's known at this stage.
+
+**[2]:** Useful if `user.authentication.id` or `user.name` contain confidential information and cannot be used.
+
+**[3]:** The `user.id`, when populated, is expected to be generated before user is authenticated and SHOULD NOT change after the user logs in. In browser scenarios `user.id` is usually stored in cookies.
+It's NOT RECOMMENDED to populate this attribute when unauthenticated users are not tracked or identified by the system.
```
MSNev, line 28: I would remove this line, OpenTelemetry should not be making these statement.

As it's up to the application to determine whether or not they want to track the user.id when the user in authenticated. For client applications this is almost ALWAYS required as they will use this to identify the "user" in some random way

docs/general/attributes.md
```diff
@@ -409,9 +409,19 @@ These attributes may be used for any operation with an authenticated and/or auth
...
+| [`user.id`](/docs/attributes-registry/user.md) | string | Identifies a user interacting with a system regardless of user authentication status. This identifier may be unique only through best-effort means. [1] | `QdH5CAWJgqVT4rOr0qtumf` | `Conditionally Required` [2] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`user.authentication.id`](/docs/attributes-registry/user.md) | string | Unique identifier of an authenticated user in the system. [3] | `S-1-5-21-202424912787-2692429404-2351956786-1000` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`user.roles`](/docs/attributes-registry/user.md) | string[] | Array of user roles at the time of the event. | `["admin", "reporting_user"]` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1]:** The `user.id`, when populated, is expected to be generated before user is authenticated and SHOULD NOT change after the user logs in. In browser scenarios `user.id` is usually stored in cookies.
+It's NOT RECOMMENDED to populate this attribute when unauthenticated users are not tracked or identified by the system.
```
MSNev, line 417: Again remove

model/general.yaml
MSNev: Maybe another way to same this is that this value SHOULD NOT be identifying in anyway to the real user id. So this value MUST not contain PII which can directly identify the authenticated user.

model/registry/user.yaml
```diff
@@ -22,14 +22,28 @@ groups:
...
         type: string
         stability: experimental
         brief: >
-          Unique identifier of the user.
+          Identifies a user interacting with a system regardless of user authentication status. This identifier may be unique only through best-effort means.
+        note: >
+          The `user.id`, when populated, is expected to be generated before user is authenticated and SHOULD NOT change after the user logs in.
+          In browser scenarios `user.id` is usually stored in cookies.
+
+          It's NOT RECOMMENDED to populate this attribute when unauthenticated users are not tracked or identified by the system.
```
MSNev, line 36: Remove

MSNev (changes requested):
Please see comments about being not recommended.

docs/attributes-registry/user.md
lmolkova: why?

we want authenticated user id to always be in the same attribute `user.authentication.id`. If you don't have means to track unauthenticated users, they should never populate `user.id`.

Otherwise how you see people using those attributes?

model/general.yaml
lmolkova: PII is once concern, but my main point is is avoid duplication (you should not generate new guid for the sake of populating user.id) and consistency - everyone should be using `user.*.id` in the same way

heyams: for azure monitor javascript SDK, browser always has a cookie named 'ai_user=QdH5CAWJgqVT4rOr0qtumf`. this will be tracked via `user.id` regardless of whether user is authenticated or not. user can choose to use `user.authentication.id` for the real user's id/name.

can we make it not required and remove the condition completely?

lmolkova: I'm missing something. 

The condition here is if user tracking is supported, so azmon distro should set `user.id` based on it.

this condition says nothing about authentication.

heyams: all i'm saying that `user.id` is not a required field.

lmolkova: would this work?

```suggestion
          recommended: If and only if instrumentation supports tracking unauthenticated users.
```

docs/attributes-registry/user.md
heyams: only authenticated user id can be tracked through `user.authentication.id` and everything else is tracked through 'user.id', it can be anonymous, random, pseudo, unauthenticated, unauthorized and etc. can we not make any recommendation how OpenTelemetry customers use this attribute? as long as we have a crystal-clear description, it's up to the users to decide. does it help?

lmolkova: We're recommending to use `user.id` only if app has some means to identify them (without authentication),
Assuming app has no means to identify users before authentication, `user.id` should not be used.

App can identify them by generating guid on every call, that's also a way of identification/tracking. 

So from what I can tell this `NOT RECOMMENDED` just reinforces the description and does not contradict anything you're saying. 

I want to keep this sentence since I expect people to put things they don't need and want to do as much as possible to avoid it. What's the problem in keeping it?

heyams: I don't have a strong opinion on this; your argument seems valid. let's try to resolve it in Monday's SIG with @MSNev and everyone else.

trask: I’m finding these names a bit confusing, and agree with this:

> I expect people to put things they don't need into `user.id` and want to do as much as possible to avoid it

It seems that unauthenticated/tracking/pseudo user ids are fairly specific to client (browser/mobile) instrumentation, so maybe it’s better to have a more specific name for that concept (and not use `user.id` for it)?

model/general.yaml
trisch-me: > i'm saying that user.id is not a required field.

It might be not required for this specific use case. But for anything related to the users in OS `user.id` is well known and important field

heyams:
> @heyams hey, do you have any updates for this issue? Can you ping me in slack if you have worked on it while I was away?

will do

lmolkova:
Do we want to close this one in favor of https://github.com/open-telemetry/semantic-conventions/pull/1456?

PR #1154, opened 2024-06-13
[editorial] Fix link by dropping anchor to title

chalin:
This is the fix necessary to be able to drop the patch that was added via:

- https://github.com/open-telemetry/opentelemetry.io/pull/1851

/cc @open-telemetry/docs-maintainers


lmolkova (approved):

docs/general/logs.md
```diff
@@ -61,7 +61,7 @@ The id MAY be an [Universally Unique Lexicographically Sortable Identifier (ULID
 
 This section describes attributes for log media in OpenTelemetry. Log media are mechanisms by which logs are transmitted. Types of media include files, streams, network protocols, and os-specific logging services such as journald and Windows Event Log.
 
-**Note:** The OpenTelemetry specification defines a [Resource](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.33.0/specification/resource/sdk.md#resource-sdk) as `an immutable representation of the entity producing telemetry`.
```
joaopgrassi, line 64: So, this means we can never refer to anchors in markdown? 🤔

chalin: @joaopgrassi - no, anchors are ok. The problem here is that `#resource-sdk` refers to the `<h1>` heading of the markdown page (i.e., the page "title"), which exists when viewed from the GitHub web interface, but doc page titles don't have anchors when viewed from opentelemetry.io -- so this results in a broken link when the spec gets published (which then requires that I manually apply file conversion patches, etc). Does that make sense?

I'm slowly working towards a proposal to catch these issues earlier, but I don't know when that'll be ready. This PR is a manual fix in the meantime.

joaopgrassi (approved):

PR #1166, opened 2024-06-19
Define sampling relevant attributes for database client spans

### LINKED ISSUE #1019: Database: review sampling_relevant attributes
### Area(s)

area:db

### What happened?

We should identify and document attributes that are relevant for sampling

1.25.0

### Additional context

_No response_

trask:
Fixes #1019

## Changes

Defines sampling relevant attributes for database client spans


model/trace/database.yaml
```diff
@@ -1,10 +1,23 @@
...
     type: attribute_group
     brief: This group defines the attributes used to perform database client calls.
+    attributes:
+      - ref: db.operation.name
+        sampling_relevant: true
+      - ref: server.address
+        sampling_relevant: true
+      - ref: server.port
+        sampling_relevant: true
+
```
lmolkova, line 13: mind adding a TODO comment like
```suggestion
# TODO: add db.system once https://github.com/open-telemetry/build-tools/issues/192 is possible
# - ref: db.system
#    sampling_relevant: true  

```
?


Unfortunately we don't mention `db.system` as sampling-relevant in system-specific conventions, but if we add it, it'll appear as 50+ systems in each extension.

Also maybe mention it's being sampling-relevant in the beginning of each doc? E.g.

> db.system MUST be set to "cassandra" and SHOULD be provided at span creation time.

lmolkova (approved):

model/trace/database.yaml
```diff
@@ -196,33 +224,37 @@ groups:
...
     type: span
-    extends: attributes.db.client.minimal
+    extends: trace.db.common.minimal
     brief: >
       Attributes for Elasticsearch
     attributes:
       - ref: http.request.method
+        sampling_relevant: true
         requirement_level: required
       - ref: db.operation.name
```
gregkalapos, line 259: This isn't marked with `sampling_relevant: true`, but it's listed as sampling relevant in the md file. We definitely have this during span creation.

I think `sampling_relevant: true` should be added here.

model/trace/database.yaml
```diff
@@ -196,33 +224,37 @@ groups:
...
           The MongoDB collection being accessed within the database stated in `db.namespace`.
         requirement_level: required
       - ref: db.namespace
+        sampling_relevant: true
         brief: The MongoDB database name.
         requirement_level:
           conditionally_required: If available.
         note: ""  # overriding the base note
 
   - id: db.elasticsearch
```
gregkalapos, line 250: For Elasticsearch, we also have `db.elasticsearch.path_parts.<key>` available during span creating, so potentially we could also consider this for sampling relevant. I'm not sure we want to have these db specific fields, or just common ones.

model/trace/database.yaml
```diff
@@ -21,17 +37,19 @@ groups:
...
+    extends: trace.db.common.minimal
     type: attribute_group
     brief: This group defines the attributes used to perform database client calls.
     attributes:
       - ref: db.query.text
+        sampling_relevant: true
         requirement_level:
           recommended: >
             SHOULD be collected by default only if there is sanitization that excludes sensitive information.
       - ref: db.query.parameter
```
gregkalapos, line 49: Was there a specific reason to not include this? If we have `db.query.text` marked as sampling relevant and users opt in to collect query parameters, then this will be naturally also captured during spar start.

Definitely not pushing to add this as sampling relevant - just raising for consideration.

model/trace/database.yaml
trask: > it's listed as sampling relevant in the md file

this means it's being inherited from an attribute group

model/trace/database.yaml
trask: from discussion:
* sampling relevant attributes incur some performance because can't defer setting them until after sampling decision
* sampling relevant attributes are somewhat related to metrics (which need to be captured on start)
* we think we can change an attribute from "not sampling relevant" to "sampling relevant" in the future (but not vice versa)

model/trace/database.yaml
```diff
@@ -1,10 +1,26 @@
 groups:
-  - id: trace.db.common.query
+  - id: trace.db.common.minimal
     extends: attributes.db.client.minimal
     type: attribute_group
     brief: This group defines the attributes used to perform database client calls.
+    attributes:
+      # TODO: add db.system once https://github.com/open-telemetry/build-tools/issues/192 is possible
```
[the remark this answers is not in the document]
trask: done, #1232

model/trace/database.yaml
```diff
@@ -1,10 +1,26 @@
...
+      # - ref: db.system
+      #   sampling_relevant: true
+      - ref: db.operation.name
```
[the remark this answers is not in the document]
trask: their requirement levels are inherited from the parent attribute group, these refs are only needed for overriding `sampling_relevant`. the attribute group inheritance is confusing, I usually rely on looking at the markdown to see the final result

docs/database/redis.md
```diff
@@ -10,7 +10,7 @@ The Semantic Conventions for [Redis](https://redis.com/) extend and override the
 that describe common database operations attributes in addition to the Semantic Conventions
 described on this page.
```
jack-berg, line 11: Is this to say that attributes `database-spans.md` which do not appear on `redis.md` are not part of redis DB semantic conventions? If so (and I believe this is the case or else I'd ask why `db.collection.name` is omitted), then this is not immediately obvious based on the text.

jack-berg (approved):

docs/database/redis.md
lmolkova: Good catch, let me follow up on this

gregkalapos (approved):

maryliag (approved):

docs/database/redis.md
lmolkova: resolving and will follow up in a separate PR

PR #1175, opened 2024-06-21
Policy checks with jq

### LINKED ISSUE #1068: Add attribute naming checks
According to attribute naming:

https://github.com/open-telemetry/semantic-conventions/blob/e29907969685fb784c653e3e083da21f3eba644b/docs/general/attribute-naming.md?plain=1#L52-L58

we SHOULD NOT allow adding attributes that would collide with namespaces (e.g. in presence of `log.iostream` attribute, we should not allow adding `log.iostream.something_else`).

But we don't have an automated check for it

## Steps to Reproduce (if any)

https://github.com/open-telemetry/semantic-conventions/pull/1067

## Expected Result

There should be a check preventing adding an attribute like `log.iostream.something_else` in presence of existing attribute `log.iostream` or vice versa

## Actual Result

Semconv checks are passed and attribute can be added

lmolkova:
Fixes #1068, detect constant name collisions

## Changes
```
Errors:
	Multiple attributes with the same constant name are defined: messaging.client_id, messaging.client.id
	Attribute and namespace name collision: messaging.operation
	Attribute and namespace name collision: db.operation
```

### DIFF
--- a/internal/tools/codegen_collisions_check.jq
+++ b/internal/tools/codegen_collisions_check.jq
@@ -1,0 +1,23 @@
+def to_const_name: .| split(".")|join("_");
+def exclude(list):
+  map(select( .name as $name | list | all(. != $name) ) );
+def to_const_name: .| split(".")|join("_");
+def get_namespaces: . | split(".") as $parts  | reduce range(1; ($parts |length)) as $i ([]; . + [$parts[0:$i] | join(".")]  );
+def fail_on_const_name_collision: . as $original
+    | exclude(["messaging.client_id"])
+    | group_by(.const_name)
+    | map(select(length>1))
+    | map({error: ("Multiple attributes with the same constant name are defined: " + (. | map(.name) | join(", ")))});
+def fail_on_attr_name_collision: . as $original
+    | exclude(["messaging.operation", "db.operation"])
+    | {names: map(.|.name), namespaces: map(.|.name|get_namespaces[] ) | unique}
+    | .namespaces as $namespaces
+    | .names | map(select( . as $name | $namespaces | index($name)))
+    | map({error: ("Attribute and namespace name collision: " + .)});
+.groups
+    | map(select(.type == "attribute_group"))
+    | map(select(.id | startswith("registry")))
+    | map({attributes: .attributes | map(.const_name= (.name | to_const_name)) })
+    | [.[].attributes[]] as $original
+    | [($original | fail_on_const_name_collision)[][], ($original | fail_on_attr_name_collision)[][]]
+    | if length > 0 then error("\nErrors:\n\t" + (. | join("\n\t"))) else "success" end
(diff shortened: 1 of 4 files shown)


internal/tools/codegen_collisions_check.jq
```diff
@@ -0,0 +1,23 @@
+def to_const_name: .| split(".")|join("_");
+def exclude(list):
+  map(select( .name as $name | list | all(. != $name) ) );
+def to_const_name: .| split(".")|join("_");
```
trisch-me, line 4: this one is repeating the function above

jsuereth (commented):
I think this may be easier to express/undertand in Rego / OPA.

From what I can tell, this works for collision checking, but given your own review of JQ, I am worried about long-term maintenance of policy checking w/ JQ vs. a more dedicated tool.

This might be worth a live pair-programming session to try out Rego policy rules and see which one we think will be easier to maintain, let me know.

lmolkova:
> I think this may be easier to express/undertand in Rego / OPA.
> 
> From what I can tell, this works for collision checking, but given your own review of JQ, I am worried about long-term maintenance of policy checking w/ JQ vs. a more dedicated tool.
> 
> This might be worth a live pair-programming session to try out Rego policy rules and see which one we think will be easier to maintain, let me know.

Well, once we'll have all the groups accessible in rego, I'd write something like 

```rego
package otel

to_const_name(prefix, id) = const_name {
    const_name := replace(trim(concat(".", [prefix, id]), "."), ".", "_")
}

deny[schema_evolution_violation("const name collision", group.id, attr.id)] {
    group := input.groups[_]
    attr := group.attributes[_]

    group.type == "attribute_group"
    not attr.ref

    const_names := [n | g := input.groups[_]; a := g.attributes[_]; n := to_const_name(g.prefix, a.id)]

    const_name := to_const_name(group.prefix, attr.id)
    count({i | const_names[i] == const_name}) > 1
}

attr_registry_violation(violation_id, group_id, attr_id) = violation {
	violation := {
		"id": violation_id,
		"type": "semconv_attribute",
		"category": "attribute_registry",
		"group": group_id,
		"attr": attr_id,
	}
}
```

(not sure if it works, but might?)

Do I like it more than JQ? Is it easier? Is it more readable?  Not sure in any of those. 
But it's definitely a 3rd cryptic language someone  needs to learn in addition to jinja and jq to efficiently support semconv. The reason I tried JQ is to show that we can do it with 2 cryptic languages instead of 3.

Of course we can hide the complexity in weaver, but that's defeats the purpose of having policies right in the semconv.

I'd prefer to write python over resolved registry and keep this python right here.

lmolkova:
I'm closing this - I have no intention to continue down this route. My concerns about (unnecessary) complexity of rego, lack of testability or debuggability remains.

PR #1180, opened 2024-06-24
Add metrics for the BEAM-VM

RoadRunnr:
Follow up from #827 because I can't reopen the PR.

## Changes

Add metrics for the BEAM-VM


RoadRunnr:
@tsloughter @ferd replacement PR for #827

ferd (approved):
I can't exactly see the changes from the previous PR since the commits have been merged into a single one, but from the spot checks I've done based on my previous comments, this seems to hold up.

RoadRunnr:
@tsloughter could you have look at this PR, please?

RoadRunnr:
> I think every use of `updowncounter` should be `gauge`. Are there any you think should be kept as updowncounter over gauge?

Most should, following the guide in https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/#instrument-selection, I agree that most should indeed be gauges, but a few should IMHO stay`updowncounter`.

I'll go through them and change the ones that I think should be changed and we can then have a discussion about the remaining ones.

RoadRunnr:
@ferd @tsloughter most updowncounters replaced with gauges

tsloughter:
@RoadRunnr ugh, sorry, I was wrong. I forgot the way to know which to use is if makes sense to sum the values across attributes. So like ` metric.beam.memory.allocators.carrier.count` probably makes sense to sum across different allocators if you wanted a total carrier count?

I'm less sure about like `metric.beam.process.limit`... Maybe it makes sense to sum the limits across multiple beams in a cluster. I mean, the value would have meaning I suppose, so that makes it an `updowncounter`?

`metric.beam.system.wordsize` is probably one that should stay a gauge. I wouldn't say that the sume of all `wordsize` across beam instances makes any sense.

RoadRunnr:
@tsloughter

Just to verify that we have the same understanding: `updowncounter` for metrics where is make sense to sum them across multiple instance in a cluster, `gauge` for metrics where aggregation makes no sense.

`metric.beam.memory.allocators.carrier.count` should IMHO be  `gauge`. The size of an carrier depends on the settings of the beam VM. So the sizes could vary between instances on a cluster and even on the same VM the carrier size depends on the carrier type. The `count` is therefor only applicable on a per beam per carrier type basis.

`metric.beam.process.limit` is somewhat borderline. But what information would the total process limit aggregated across multiple beams carry ? The only sensible use is IMHO on a per beam VM basis, therefore also `gauge`.

`metric.beam.system.wordsize` ack, adding word sizes is pointless.

The only metrics that IMHO make sense to aggregate  are total memory used values. That is:
* `metric.beam.memory.allocators.carrier.size`
* `metric.beam.memory.atoms`
* `metric.beam.memory.allocated`
* `metric.beam.memory.processes`
* `metric.beam.memory.system`

PR #1182, opened 2024-06-26
CloudEvents should re-use existing HTTP or messaging conventions

### LINKED ISSUE #654: Clarify possible incompatibilities with CloudEvents Distributed Tracing extension
[CloudEvents](https://github.com/cloudevents/spec) is a specification for describing event data in common formats to provide interoperability across services, platforms and systems. It is a CNCF project which is very popular with messaging systems.

CloudEvents provides an [extension for Distributed Tracing](https://github.com/cloudevents/spec/blob/v1.0.1/extensions/distributed-tracing.md). This extension specifies an additional channel for context propagation (in addition to context propagation via e. g. HTTP or AMQP). It is not clear whether and how this extension should be used when instrumenting with OpenTelemetry and when utilizing OpenTelemetry semantic conventions.

When stabilizing OpenTelemetry semantic conventions for messaging and HTTP, it needs to be worked out how this CloudEvents extension can be utilized in OpenTelemetry scenarios, and whether it's needed at all.

joaopgrassi:
Fixes #654

## Changes

We are approaching the stability of the messaging conventions, and as discussed we want to clarify the relation between CloudEvents convention with messaging. The conclusion we have is that applications/instrumentations working with CloudEvents should generally follow either the HTTP conventions or messaging. Those already offer mechanisms for context propagation and also are prepared to handle asynchronous scenarios such as when CloudEvents are sent via messaging systems.


docs/cloudevents/cloudevents-spans.md
```diff
@@ -35,168 +31,38 @@ consult the
...
-on the event. This applies when:
-
-- A CloudEvent is created by the instrumented library.
-  It may be impossible or impractical to create the Span during event
-  creation (e.g. inside the constructor or in a factory method),
-  so instrumentation MAY create the Span later, when passing the event to the transport layer.
-- A CloudEvent is created outside of the instrumented library
-  (e.g. directly constructed by the application owner, without calling a constructor or factory method),
-  and passed without the Distributed Tracing Extension populated.
+CloudEvents can be sent via various transports, such as HTTP or via messaging
```
lmolkova, line 36: transport context != message context, so the protocol used to carry data is not relevant for messaging or CloudEvents

docs/cloudevents/cloudevents-spans.md
```diff
@@ -35,168 +31,38 @@ consult the
...
+systems. Each transport defines its own mechanism of
```
lmolkova, line 37: I left a comment on https://github.com/cloudevents/spec/pull/912#issuecomment-2195429167 - not each protocol. 

There are at least 2 Azure messaging services that don't have any place to stick trace context to. 

I support removing context propagation section above, but here I'd say  something like

When CloudEvents are used in messaging scenarios and corresponding messaging system supports propagating trace context along with the message independently of CloudEvents, the corresponding messaging instrumentation SHOULD propagate the context using this messaging-system capabilities.
Such instrumentations may not be aware of the payload format or could support CloudEvents along with other formats.

If CloudEvents are instrumented independently of messaging, the CloudEvent-specific instrumentation SHOULD use CloudEvents distributed context propagation extension to inject the context and is encouraged to follow span structure described in messaging semantic conventions.

joaopgrassi: Right, I think this is very close to what I tried to say below:

> When the transport being used does not have a well established
propagation mechanism, instrumentations MAY rely on the
[CloudEvents Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/extensions/distributed-tracing.md).
as means to propagate the trace context, while still following either the HTTP
or Messaging span semantic conventions.

I think we are aligned on the objective. I will try to rephrase it a bit and incorporate some of your suggestions :)

docs/cloudevents/cloudevents-spans.md
joaopgrassi: Not sure I got it? What I tried to say here and that one can sent CloudEvents via different transport/protocols, so http, messaging, etc and that each of those defines its own way of context propagation. Maybe I should say MAY define its own mechanism instead, as you pointed out some don't define anything. The way it is written now implies that all do which is false. Is that what you meant or?

docs/cloudevents/cloudevents-spans.md
joaopgrassi: It seems also you want to leave out completely the mention about CloudEvents used in HTTP scenarios?

joaopgrassi: I made some changes, let me know what you think!

docs/cloudevents/cloudevents-spans.md
```diff
@@ -35,168 +31,26 @@ consult the
...
-Once the trace context is set on the event
-via the Distributed Tracing Extension, it MUST not be modified.
-
-The remainder of this section describes the semantic conventions for Spans
-required to achieve the proposed trace.
-
-### Spans
+CloudEvents can be sent via various communication protocols. When the protocol
+used to send CloudEvents has a well established support for trace context propagation,
+instrumentations SHOULD propagate the context using this built-in mechanism.
```
pyohannes, line 38: We shouldn't say that, it's misleading.

Given CloudEvents are sent via HTTP, then the established context propagation mechanism for HTTP doesn't satisfy [requirements for messaging context propagation](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md#context-propagation):

> If possible, the message creation context SHOULD be attached in such a way that it cannot be changed by intermediaries.

A context attached to a HTTP request by the producer will not make it through to the consumer. I'd remove this remark.

docs/cloudevents/cloudevents-spans.md
```diff
@@ -35,168 +31,26 @@ consult the
...
 
-#### Creation
-
-Instrumentation SHOULD create a new span and populate the
+If CloudEvents are instrumented independently of the protocol used,
+CloudEvent-specific instrumentations SHOULD use the
 [CloudEvents Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/extensions/distributed-tracing.md)
```
pyohannes, line 42:
```suggestion
The CloudEvents specification defines a context propagation mechanism via CloudEvents in the [CloudEvents Distributed Tracing Extension](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/extensions/distributed-tracing.md). This mechanism satisfies [context propagation requirements of OpenTelemetry messaging semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md#context-propagation).
```

I'd avoid SHOULD wording, and just hint at the existence of the context propagation extension.

docs/cloudevents/cloudevents-spans.md
```diff
@@ -35,168 +31,26 @@ consult the
...
-
-- Span kind: CONSUMER
+as means to propagate the trace context.
 
-- Span attributes: Instrumentation MUST add the required attributes defined
-  in the [table below](#attributes).
+When CloudEvents are used in messaging scenarios, where an event can go through
+many [hops](https://en.wikipedia.org/wiki/Hop_%28networking%29),
+CloudEvent-specific instrumentations SHOULD follow the span structure described in
+[Semantic Conventions for Messaging Spans](../messaging/messaging-spans.md).
```
pyohannes, line 48:
```suggestion
No dedicated CloudEvents spans should be created, but CloudEvents-specific attributes should be used to enrich scenario-specific spans (for example [messaging](../messaging/messaging-spans.md) or [HTTP](../http/http-spans.md) spans).
```

docs/cloudevents/cloudevents-spans.md
joaopgrassi: I agree, but I'm not sure about removing it completely. It is perfectly fine to attach the context to the HTTP request, if the communication is done via HTTP and not via messaging systems. Previously I had paragraphs saying explicitly that if sending HTTP in per-to-per/HTTP the HTTP conventions/context propagation should be used, otherwise the messaging conventions are to be used. But I tried to incorporate Liudmilla's suggestions https://github.com/open-telemetry/semantic-conventions/pull/1182#discussion_r1657656417. 

The thing is, I'm not convinced that we should just talk about messaging here, as I'm sure people also use CloudEvents in synchronous HTTP use cases.

joaopgrassi: Maybe we should have clear sections on this document to give directions on specific protocols/methods of sending CloudEvents? If in synchronous/peer-to-peer/HTTP -> use HTTP conventions and propagate via header. If using asynchronous/messaging -> Use Messaging conventions, propagate via existing messaging system mechanisms otherwise use trace extension from CloudEvents.

pyohannes: I'd prefer to turn this around a bit. Instead of taking the usage of CloudEvents a starting point, I'd rather like to treat this as an implementation detail. Instrumentation authors should follow the following train of thought:

1. I'm doing peer-to-peer HTTP, or I'm doing messaging, or I'm doing something else.
2. Therefore, I'm instrumenting according to HTTP semantic conventions, to messaging semantic conventions, or according to another semantic conventions area (RPC, or whatever).
3. I'm using CloudEvents as data format.
4. Therefore, I'm adding CloudEvents attributes to the spans I obtained from following (2).

> Maybe we should have clear sections in this document to give directions on specific protocols/methods of sending CloudEvents?

I'd like to avoid that, as we can't cover all cases. What about CloudEvents via gRPC? AMQP? Stored in databases?

pyohannes (approved):
I don't want to make this too complicated, so as long as we avoid duplication with messaging semantic conventions (which the current approach already does), I'm happy to approve this.

Just one additional remark below.

docs/cloudevents/cloudevents-spans.md
joaopgrassi: But if they follow the train of thought you laid out above, wouldn't make sense that we have sections for each mechanism? 🤔 . For HTTP and Messaging we have it all covered, and for the rest, it seems fit to rely on the distributed tracing extension from cloud events as the propagation mechanism.

joaopgrassi:
@lmolkova @pyohannes I took your suggestions and also from the discussions in the CloudEvents PR and tried to simplify things a bit. Would be great if you can take another look. Thanks!

lmolkova (approved):

Issue #1124, opened 2024-06-05
Attribute names: unicode on OTLP, only `[a-z0-9._]` in OTel semcov


lmolkova:
Attribute names can be any unicode sequence

https://github.com/open-telemetry/semantic-conventions/blob/dd277f62f66f3342be33aec2c432f6bd959b379b/docs/general/attribute-naming.md?plain=1#L24

It makes sense for user apps using OTel API and OTLP, but is not accepted by our ([build-tools](https://github.com/open-telemetry/build-tools/blob/14261ddd0b14fecb94e9fca3e0690008b83de836/semantic-conventions/src/opentelemetry/semconv/model/utils.py#L20-L23)) tooling 

```python
ID_RE = re.compile("([a-z](\\.?[a-z0-9_-]+)+)")
"""Identifiers must start with a lowercase ASCII letter and
contain only lowercase, digits 0-9, underscore, dash (not recommended) and dots.
Each dot must be followed by at least one allowed non-dot character."""
```

We should document and enforce the rules that we have for semantic convention definitions in this repo:
- only a-z, 0-9, `.` and `_` are accepted
- starts with a letter
- ends with a letter or number
- (no dashes - there are no existing attributes with it)

These rules are necessary for code-generation. They should also apply to metric names, units, event names, event payload fields, or other properties that are likely to be represented as a code.

We can expand the list of allowed characters if we can find a way to support code generation for them.

tigrannajaryan:
The use of `[a-z0-9._]` is currently merely a recommendation, not a strict restriction. I think it is fine if we want to rely on that recommendation and make it the default behavior for our tools but we may need the tools to be able to deal with exceptions.

I think use cases [like this](https://github.com/open-telemetry/semantic-conventions/pull/1188#discussion_r1659224378) show that strictly prohibiting it may create problems with interoperability with other standards.

lmolkova:
Agreed. Our existing tooling imposes such limitations - CI checks would flag and fail. 
One of the reasons to have this limitation to be able to translate attribute/metrics/etc names need to constant names in the code. 

To support other characters, we'd need some mechanism to define a code-friendly name for such identifiers. We might a similar mechanism for https://github.com/open-telemetry/semantic-conventions/issues/1118#issuecomment-2173803006 (phase 2).

Issue #1118, opened 2024-06-03
Code generation: how to avoid naming collisions


lmolkova:
As we discovered in #1031, certain semantic conventions changes (together with default code generator behavior) result in ambiguous constant names generated in the code.

For example, when `foo.bar_baz` is renamed to `foo.bar.baz`, the code generator produces the same constant name for both -`FOO_BAR_BAR` , `FooBarBaz`, etc depending on the preferred casing.

We don't remove attributes/enum members/metrics from the semantic conventions anymore (old property is deprecated), so both constants would exist in the same version.

So, the semantic conventions together with the codegen should prevent such collisions from happening:
- The default tooling behavior should result in collision-free code.
- There could be additional policies added to semconv to prevent certain renames.

lmolkova:
Option 1:

- any renames are allowed (for experimental conventions)
- `_` is preserved in constant names:
  - `foo.bar_baz` is generated as `FOO_BAR__BAZ`, `FooBar_Baz`, etc

Cons:
- `__` is ugly in snake_case and `_` is ugly in PascalCase

lmolkova:
Option 1.5:

`foo.bar_baz` is generated as `FOO__BAR_BAZ`, `Foo_BarBaz`, ...

lmolkova:
Option 2:

`foo.bar_baz` is generated as `FOO_BARBAZ`, `FooBarbaz`, ...

- renames that only remove or add `_` are not allowed even for experimental attributes. 
- `_` is removed during code generation

Cons:
- limited rename options. Partially mitigated by the fact that we didn't recall making such renames in the past.

lmolkova:
Option 3 (do nothing):

- any renames are allowed (for experimental conventions)
- `_` and `.` are not distinguishable in constant names

Cons:

- attribute name and schema version mismatch (instrumentation thinks it sets `foo.bar_baz`, but it sets `foo.bar.baz`.

lmolkova:
Option 4:

`foo.bar_baz` is generated as `FOO_BAR_BAZ`, `FooBarBaz`.
When `foo.bar.baz` is added, the collision is detected, so the new attribute is called `FOO_BAR_BAZ_NEW`, `FooBarBazNew`.

- any renames are allowed (for experimental conventions)
- resolve collisions by giving new names 

Cons:

- new attribute would have a 'bad' name and deprecated attribute would have a `good` name forever.  

The rename can be done for the old attribute, resulting in Option 3.

[...]

PR #1188, opened 2024-06-27
[profiles] introduce semantic convention

florianl:
## Changes

With https://github.com/open-telemetry/opentelemetry-proto/pull/534 a new OTel signal was introduced.

This change introduced semantic conventions specific to the OTel Profiling signal.

FYI: @open-telemetry/profiling-maintainers  @aalexand 


.chloggen/profiles-convention.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
trisch-me, line 7:
```suggestion
change_type: new_component
```

.chloggen/profiles-convention.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: profiles
```
trisch-me, line 10:
```suggestion
component: profile
```

petethepig (approved):

model/registry/profile.yaml
```diff
@@ -0,0 +1,102 @@
...
+          The Go build ID.
+        examples: ['foh3mEXu7BLZjsN9pOwG/kATcXlYVCDEFouRMQed_/WwRFB1hPo9LBkekthSPG/x8hMC8emW2cCjXD0_1aY']
+
+  - id: registry.profile.link
+    type: attribute_group
+    prefix: profile.link
+    brief: >
+      Connect Profile information to Spans and Traces.
+    attributes:
+      - id: trace_id
```
trask, line 28: @open-telemetry/specs-semconv-approvers thoughts?

* `trace.id` is in line with the direction we seem to be heading: https://github.com/open-telemetry/semantic-conventions/issues/1118#issuecomment-2159136053
* `trace_id` matches the logging spec's [Trace Context in Legacy Formats](
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/README.md#trace-context-in-legacy-formats) (see https://github.com/open-telemetry/opentelemetry-specification/issues/3303)

model/registry/profile.yaml
```diff
@@ -0,0 +1,102 @@
...
+              stability: experimental
+            - id: ruby
+              brief: >
+                [Ruby](https://wikipedia.org/wiki/Ruby_(programming_language))
+              value: "ruby"
+              stability: experimental
+            - id: v8
+              brief: >
+                [V8](https://wikipedia.org/wiki/V8_(JavaScript_engine))
+              value: "v8"
```
trask, line 101: might be nice if this aligned with the namespace in #1066 (either changing that to `v8` or this to `v8js`)

cc @maryliag

florianl: `v8js` sounds also good :+1: `v8` was chosen as this is the value that has been used by Elastic in the past. But going forward and aligning with other components, `v8js` sounds right.

model/registry/profile.yaml
AlexanderWert: Today, we have `trace_id` in legacy formats for logs and the corresponding OTLP-native field.

So, `trace.id` would be a new (yet another representation) of the same thing. That's why I'm slightly more in favor of  `trace_id`. Also, `trace.id` would make sense if we expect more attributes under the `trace` namespace, which I think is not the case here.

docs/general/profiles.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## Spans and Traces
+
+**Description:** Connect Profile information to Spans and Traces.
```
lmolkova, line 46: It seems profiles will have dedicated properties for span-id and trace-id https://github.com/open-telemetry/opentelemetry-proto/blob/40b3c1b746767cbc13c2e39da3eaf1a23e54ffdd/opentelemetry/proto/profiles/v1experimental/pprofextended.proto#L94

What would be the use-case for `profile.link.*`? what they will be stamped on? spans? logs? profiles themselves?

docs/general/profiles.md
```diff
@@ -0,0 +1,110 @@
...
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## Reserved attributes
+
+**Description:** For cross tooling compatibility the following attributes SHOULD be avoided.
+
+| Attribute  | Type | Description  | Examples  |
+|---|---|---|---|
+| `pprof::*` | string | Special indicator for [pprof](https://github.com/google/pprof). | `pprof::base` |
```
lmolkova, line 108: semconv allows [a-z0-9_\.] chars https://github.com/open-telemetry/semantic-conventions/issues/1124, we also have no tooling support to reserve the namespace at the moment.

But also, who/how/when will use these attributes? what would they be stamped on?

lmolkova: created https://github.com/open-telemetry/semantic-conventions/issues/1197 to track reserving namespace

docs/attributes-registry/profile.md
```diff
@@ -0,0 +1,51 @@
...
+- [Profile Frame](#profile-frame-attributes)
+- [Profile Link](#profile-link-attributes)
+
+## Profile Build Id Attributes
+
+Unique identifier for an executable.
+
+| Attribute              | Type   | Description                                                               | Examples                                                                              | Stability                                                        |
+| ---------------------- | ------ | ------------------------------------------------------------------------- | ------------------------------------------------------------------------------------- | ---------------------------------------------------------------- |
+| `profile.build_id.gnu` | string | The GNU build ID as can be found in the `.note.gnu.build-id` ELF section. | `c89b11207f6479603b0d49bf291c092c2b719293`                                            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 19: is there some documentation on this? could it be referenced? So someone not familiar with profiling could go and learn what it means?

docs/general/profiles.md
florianl: One task of the OTel Profiling SIG is to have some compatibility between OTel profiling and pprof (tool + protocol) - see also  https://github.com/open-telemetry/opentelemetry-proto/issues/567.

The semantic convention for OTel profiles allows to address this topic. And so the idea is to define the connection between profiles and spanes/traces within the semantic convention and remove `message Link` from the experimental OTel profiling protocol. This allows interoperability between OTel profiling and pprof.

docs/general/profiles.md
florianl: Thanks for pointing out the allowed array of characters`[a-z0-9_.]` in Attributes. I wasn't aware of that, when writing this. Is there another way to express colons?

Compatibility between OTel profiling and pprof (tool + protocol) is important to the OTel profiling SIG. pprof is a well established tool and protocol in the industry. And there are special features, like `pprof::*` that pprof (tool) interprets in a special way. To avoid conflicts and guarantee compatibility, `pprof::*` should be avoided.

I'm aware that this is a special case and to my knowledge, semantic convention does not cover such special cases at the moment. This is also the reason, this section is written manually here and not automatically generated like other attributes.

docs/attributes-registry/profile.md
florianl: Would it help to link here the GNU build system article from wikipedia?  https://wikipedia.org/wiki/GNU_build_system

tigrannajaryan: We typically don't use signal-specific prefixes for attribute names, unless the attribute is only meaningful just for that one signal. Is `build_id` strictly a profiling concept? If it is an identifier of a build I think it may possibly be used in other signals (e.g. log an event and include information about the build of the executable that logged it). In that case we should not use `profile.` namespace and should put this attribute in an appropriate top-level namespace (or create one if none exists).
The same comment applies to other attributes introduced.

model/registry/profile.yaml
tigrannajaryan: Whatever we choose let's make sure it is consistent with what we do for logs. For now that means using `trace_id`/`span_id`.

docs/general/profiles.md
jsuereth: These compatibility layer things MUST use `otel` namespace and are generally defined elsewhere.

See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/semantic-conventions.md#reserved-namespace

docs/attributes-registry/profile.md
```diff
@@ -0,0 +1,51 @@
...
+| `profile.build_id.go`  | string | The Go build ID.                                                          | `foh3mEXu7BLZjsN9pOwG/kATcXlYVCDEFouRMQed_/WwRFB1hPo9LBkekthSPG/x8hMC8emW2cCjXD0_1aY` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
aalexand, line 20: For the record, it looks like Go build ID is not handled correctly in Go's runtime/pprof and pprof CLI itself. I filed https://github.com/golang/go/issues/68652.

@felixge FYI

model/registry/profile.yaml
florianl: Removed the attributes around span/trace ID as there is a dedicated field in the OTel profiling signal.

docs/general/profiles.md
florianl: This compatability section around google/pprof got removed.

docs/attributes-registry/profile.md
florianl: @tigrannajaryan I do need some advice on resolving a conflict.

With `process.executable` there exists an attribute group, where `process.executable.build_id` could be another field. But for the Profiling use case, there is a conflict, that a single process can have multiple executables - e.g. a stack trace of a Python script can have libc, libpython, kernel and interpreted python frames (and maybe even more). The former three, libc, libpython and kernel, will have their dedicated `build_id`. So a 1:1 mapping of process to executable is not possible.
How could the 1:1 mapping become a 1:n mapping where one process can have multiple executables in its stack?  

FYI: @christos68k @AlexanderWert

florianl: Removed the build_id from this proposed change with https://github.com/open-telemetry/semantic-conventions/pull/1188/commits/2ae1ac9c7156438f3698e9b0fb1f5ca182157926 and will open a new issue/PR around it as a follow up.

model/registry/profile.yaml
```diff
@@ -0,0 +1,102 @@
+groups:
+
+  - id: registry.profile.build_id
+    type: attribute_group
+    prefix: profile.build_id
```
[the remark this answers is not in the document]
florianl: build_id got removed from this PR with https://github.com/open-telemetry/semantic-conventions/pull/1188/commits/2ae1ac9c7156438f3698e9b0fb1f5ca182157926.

florianl:
@open-telemetry/profiling-maintainers the PR got updated to only contain the frame types.

christos68k (approved):

model/registry/profile.yaml
```diff
@@ -0,0 +1,63 @@
...
+            - id: ruby
+              brief: >
+                [Ruby](https://wikipedia.org/wiki/Ruby_(programming_language))
+              value: "ruby"
+              stability: experimental
+            - id: v8js
+              brief: >
+                [V8JS](https://wikipedia.org/wiki/V8_(JavaScript_engine))
+              value: "v8js"
+              stability: experimental
```
athre0z, line 63: This one is a bit of an outlier: all the other frame types here indicate a *language*, not an *implementation*. For example, we have `python` instead of `CPython`, `dotnet` instead of `.NET core`, and `jvm` instead of `hotspot`. V8, on the other hand, is a specific runtime. Should we generalize this to "js"?

florianl: The change from `v8` to `v8js` originates from a request in https://github.com/open-telemetry/semantic-conventions/pull/1188#discussion_r1658010097 and as https://github.com/open-telemetry/semantic-conventions/pull/1066 did land first, this proposed change did adapt.

athre0z: `v8js` makes sense in the context of what #1066 added because the attributes are actually runtime specific, but this isn't really the case here. 

I can't speak for @trask, but I suspect that his comment was based on the fact that the field already had the runtime-specific name "v8" in your draft. So if we went with that, it'd certainly make sense to go with the already established term `v8js`.

I think both approaches, reporting the general language OR the specific runtime would be raesonable for this field, and I don't have a strong preference. However, we should be *consistent* with that, and not mix it up at random.

brancz (approved):

model/registry/profile.yaml
```diff
@@ -0,0 +1,63 @@
...
+        stability: experimental
+        brief: >
+          Describes the interpreter or compiler of a single frame.
+        examples: ['python']
+        type:
+          allow_custom_values: true
+          members:
+            - id: dotnet
+              brief: >
+                [.Net](https://wikipedia.org/wiki/.NET)
```
joaopgrassi, line 18:
```suggestion
                [.NET](https://wikipedia.org/wiki/.NET)
```

model/registry/profile.yaml
joaopgrassi: .NET is not the language but the runtime/framework. You can program in C#, F# and Visual Basic with .NET. Also there's no .NET Core anymore, it's all only .NET.

athre0z: Fair point. The same is also true for JVM: it can also run e.g. Scala, so it's a runtime as well. For most languages in the list, the name of the reference implementation runtime is equal to the language itself. So I think if we'd just rename `Python` to `CPython`, it may already be consistent then.

florianl: https://github.com/open-telemetry/semantic-conventions/pull/1188/commits/608fb5c84773a3c7d060ebafd351677336e0839b

AlexanderWert (approved):

athre0z (approved):

model/registry/profile.yaml
```diff
@@ -0,0 +1,63 @@
...
+              brief: >
+                [Kernel](https://wikipedia.org/wiki/Kernel_(operating_system))
+              value: "kernel"
+              stability: experimental
+            - id: native
+              brief: >
+                [C](https://wikipedia.org/wiki/C_(programming_language)),
+                [C++](https://wikipedia.org/wiki/C%2B%2B),
+                [Go](https://wikipedia.org/wiki/Go_(programming_language)),
+                [Rust](https://wikipedia.org/wiki/Rust_(programming_language))
```
felixge, line 36: Are these 4 languages grouped together on purpose, or because it's difficult for the ebpf profiler to distinguish them? Did we write down any use cases for these frame types somewhere? IIRC from the SIG discussions, the use case is to visually highlight different languages when visualizing the data (e.g. with different colors in a flame graph). If that's the case, distinguishing between C, C++, Go and Rust would probably be useful?

Apologies if this has been discussed already.

felixge (approved):

model/registry/profile.yaml
florianl: These four named languages are just an example for `native` frames and the list is not exhaustive.
From a pure stack unwinding point of view, I think it might be hardly possible to distinguish them. But by examining the backing executable/shared library, it should be possible. At the moment, this additional inspection is not done and needs more research.

Overall, `native` is just a general grouping and I think there should be no issue being more specific in the future and adding more specific languages, like `rust`.

felixge: Sounds good. Thanks!

AlexanderWert:
@open-telemetry/specs-semconv-approvers Can we get more general approvals on this one please?
We already have a bunch of the profiling SIG's approvals, just missing an additional general semconv SIG approval.

model/profiles/frame.yaml
```diff
@@ -0,0 +1,8 @@
+groups:
+  - id: profile.frame
+    type: attribute_group
+    brief: >
+      Describes the origin of a single frame in a Profile.
+    attributes:
+      - ref: profile.frame.type
+        requirement_level: opt_in
```
joaopgrassi, line 8: Is it really opt in here? Wouldn't we always want to know the frame type? 🤔

docs/general/profiles.md
```diff
@@ -0,0 +1,55 @@
...
+
+| Value  | Description | Stability |
+|---|---|---|
+| `dotnet` | [.NET](https://wikipedia.org/wiki/.NET) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `jvm` | [JVM](https://wikipedia.org/wiki/Java_virtual_machine) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `kernel` | [Kernel](https://wikipedia.org/wiki/Kernel_(operating_system)) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `native` | [C](https://wikipedia.org/wiki/C_(programming_language)), [C++](https://wikipedia.org/wiki/C%2B%2B), [Go](https://wikipedia.org/wiki/Go_(programming_language)), [Rust](https://wikipedia.org/wiki/Rust_(programming_language)) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `perl` | [Perl](https://wikipedia.org/wiki/Perl) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `php` | [PHP](https://wikipedia.org/wiki/PHP) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `python` | [Python](https://wikipedia.org/wiki/Python_(programming_language)) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
joaopgrassi, line 44: You need to re-generate the table with `make fix` after changing to `cpython`

joaopgrassi: @lmolkova shouldn't some check fail due to outdated markdown? The id of one of the enum names changed, but it is not reflected in the markdown here 🤔

florianl: Did run `make fix` which fixed the name issue - see https://github.com/open-telemetry/semantic-conventions/pull/1188/commits/b6aed34588cc8313eb789884f87d7acc54a3f7c4. Not sure, why some commands did generate the correct output, like in `docs/attributes-registry/profile.md`, but this file did not get updated.

model/profiles/frame.yaml
florianl: The understanding and agreement is, that `profile.frame.type` is optional, as it is "just" required for visualization of the data. Data without `profile.frame.type` can still be processed and provides its value.

joaopgrassi: I see. Maybe then `recommended` is better here? `opt-in` requires the explicit opt-in from the user via configuration/sdk setup whereas recommended is "on" by default, but can be not added if there's good reason not to, like if it's expensive to collect and etc https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attribute-requirement-level.md#recommended

docs/general/profiles.md
joaopgrassi: Most likely you ran only the command to update the registry. They are separated and `make fix` executes both :)

joaopgrassi (approved):
Looks good to me, apart from the `opt-in` requirement, but I don't have a strong argument against it, so I will defer the decision to the profiling sig.

model/profiles/frame.yaml
florianl: Updated the level to `recommended` with https://github.com/open-telemetry/semantic-conventions/pull/1188/commits/94bb7ed8f9f3288bffa02744058071c3d36ca4af.

Issue #454, opened 2023-10-26
Common socket connection conventions


lmolkova:
Connection-level instrumentation is useful for all kinds of protocols to detect/investigate any connectivity issues, optimize connection pools, detect leaks, etc. The following telemetry signals are common across protocols:
- when the connection is established/terminated/closed
  - how long it takes to attempt opening a connection
  - if a connection is terminated with an error, what was the error
- connection duration and counters

The conventions could include:
- spans
  - span per attempt to connect
  - span per connection 
  - higher-level spans (such as HTTP) might be able to link to them
- metrics
  - connection attempt duration
  - connection duration
  - number of active connections/connection pool size
  - number of opened connections
  - received/transmitted bytes?
  - etc

We should be able to reuse `network.*` and other existing common attributes to define generic connection conventions.

Individual protocols on top of sockets may extend such conventions and add protocol-specific attributes (e.g. AMQP may define some other attributes, spans/metrics).

Socket connection conventions should be reused across databases, http, messaging, and other conventions
See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md where we already define a bunch of metrics for DB connections specifically.

trask:
this was accidentally auto-closed (via merging an old commit that closed spec issue number 454)

PR #1192, opened 2024-06-27
Add .NET network + HTTP connection spans

lmolkova:
This PR introduces additional network-level spans to-be emitted by .NET 9 🤞 

Related to #454, https://github.com/open-telemetry/semantic-conventions/issues/1226

In particular:
1. DNS, TLS, and socket connect spans that track individual stages of connection establishment
2. wrapper HTTP connection setup span for above spans
3. wait-for-connection span that signals when a connection is not immediately available for the request
   - link that connects wait-for-connection to the connection setup
   - is expected to be off by default

On .NET every `ActivitySource` (aka tracer) is opt-in, so users have full control over what they are enabling.

The volume of new telemetry is expected to be low (with caveats):
- The connection-level spans (groups 1 and 2 above)  are rare - connections are created rarely.
- Wait-for-connection spans are only reported when connection is not readily available and would be present when connection pool is overloaded. 
- Links to connection setup are most concerning since we don't know how backends are handling incoming links. See https://github.com/open-telemetry/opentelemetry-specification/discussions/4130 for the context. This is off by default docs)?

image

Overall, the feature is intended to help debug issues like:
- misconfigured connection pool
- DNS/TLS
- abandoned response streams (or slow response stream reading) that result in connection staying busy
- ...


docs/dotnet/dotnet-connection-traces.md
```diff
@@ -0,0 +1,399 @@
...
+
+## Overview
+
+.NET reports spans related to HTTP connection establishment and its stages starting from .NET 9.
+Application developers are encouraged to enable corresponding instrumentation in development and test environments.
+
+Enabling connection-level details in production may result in increased telemetry consumption and reduce performance.
+
+While such spans represent low-level details, connection lifetime is usually measured in minutes, therefore in common case
+when application is under the load and connections are established for every request, the rate of connection-related spans
```
lmolkova, line 30:
```suggestion
when application is under the load and connections are not established for every request, the rate of connection-related spans
```

docs/dotnet/dotnet-connection-traces.md
```diff
@@ -0,0 +1,399 @@
...
+## DNS
+
+The span describes DNS lookup duration and outcome.
+
+DNS lookup duration is also reported by [`dns.lookup.duration` metric](/docs/dotnet/dotnet-dns-metrics.md#metric-dnslookupduration).
+
+Span name SHOULD be `DNS {dns.question.name}`.
+Span kind SHOULD be `CLIENT`
+
+Corresponding `Activity.OperationName` is `System.Net.NameResolution.DnsLookup`, `ActivitySource` name - `System.Net.NameResolution`.
```
antonfirsov, line 277: Should a future [custom DNS resolver](https://github.com/dotnet/runtime/issues/19443), emit the same span? Should it come from the same activity source?

docs/dotnet/dotnet-connection-traces.md
```diff
@@ -0,0 +1,399 @@
...
+<!-- semconv dotnet.dns.lookup -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`dns.answer`](/docs/attributes-registry/dns.md) | string[] | The list of resolved IPv4 or IPv6 addresses. | `["10.0.0.1", "2001:0db8:85a3:0000:0000:8a2e:0370:7334"]` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
antonfirsov, line 289: In context of the previous question, should this be more generic given that a proper DNS client might return all kinds of records?

docs/dotnet/dotnet-connection-traces.md
```diff
@@ -0,0 +1,399 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: HTTP client and server
+--->
+
+# Semantic Conventions for HTTP client connection spans emitted by .NET
```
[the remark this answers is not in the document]
lmolkova: qq  - do we need all the attributes (that are on http.client.request.time_in_queue) on that span? It'd be right under the request span which has everything available.

I'll add it with attributes and add a todo note for  now

docs/dotnet/dotnet-connection-traces.md
antonfirsov: @lmolkova actually, this is even more tricky. The current `System.Net.Dns` telemetry (including the `dns.lookup.duration` metric) covers protocol-independent lookup methods (eg. reverse lookup).

It would make sense to include those in our spans as well. We have to sort out our naming strategy here!

docs/attributes-registry/dns.md
```diff
@@ -10,8 +10,9 @@
 
 This document defines the shared attributes used to report a DNS query.
 
-| Attribute           | Type   | Description                 | Examples                              | Stability                                                        |
-| ------------------- | ------ | --------------------------- | ------------------------------------- | ---------------------------------------------------------------- |
-| `dns.question.name` | string | The name being queried. [1] | `www.example.com`; `opentelemetry.io` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| Attribute           | Type     | Description                                  | Examples                                                  | Stability                                                        |
```
antonfirsov, line 13: We can also add `error.type` here.

docs/dotnet/dotnet-connection-traces.md
```diff
@@ -0,0 +1,504 @@
...
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [1] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.peer.address`](/docs/attributes-registry/network.md) | string | Peer address of the network connection - IP address or Unix domain socket name. | `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.protocol.version`](/docs/attributes-registry/network.md) | string | The actual version of the protocol used for network communication. [2] | `1.1`; `2` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
antonfirsov, line 243: I assume this is a copy-paste error?

docs/dotnet/dotnet-connection-traces.md
```diff
@@ -0,0 +1,504 @@
...
+| [`network.transport`](/docs/attributes-registry/network.md) | string | [OSI transport layer](https://osi-model.com/transport-layer/) or [inter-process communication method](https://wikipedia.org/wiki/Inter-process_communication). [3] | `tcp`; `udp` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
antonfirsov, line 244: Does this bring enough value to be included without being asked explicitly by users? It will be tcp in 99.9% of the cases.

docs/dotnet/dotnet-connection-traces.md
```diff
@@ -0,0 +1,504 @@
...
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## TLS
+
+The span describes TLS handshake.
+
+Span name SHOULD be `TLS {dns.question.name}`. // TODO - decide on TLS!
+Span kind SHOULD be `CLIENT`
```
antonfirsov, line 380: It should depend on `SslStream.IsServer`.

docs/dotnet/dotnet-connection-traces.md
```diff
@@ -0,0 +1,504 @@
...
+
+The span describes the establishment of the socket connection.
+It's different from [*HTTP connection setup*](#http-connection-setup) span, which covers the DNS and TLS handshake duration in addition to socket connection setup.
+
+When *Socket connection setup* span is reported along with *HTTP connection setup* span, the socket span becomes a child of HTTP connection setup.
+
+Span name SHOULD be `socket connection_setup {network.peer.address}:{network.peer.port}`.
+Span kind SHOULD be `CLIENT`.
+
+Corresponding `Activity.OperationName` is `System.Net.Sockets.ConnectionSetup`, `ActivitySource` name - `System.Net.Sockets`.
```
antonfirsov, line 229: @noahfalk recommended to rename this to `socket connect`. I don't have a preference here, but we should make a decision.

docs/dotnet/dotnet-connection-traces.md
```diff
@@ -0,0 +1,504 @@
...
+Added in .NET Core 9.
+
+<!-- semconv dotnet.socket.connection_setup -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
```
antonfirsov, line 239: `network.peer.port` is missing.

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,399 @@
...
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [1] | `TODO` | `Conditionally Required` if and only if an error has occurred. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`server.address`](/docs/attributes-registry/server.md) | string | The [server name indication (SNI)](https://en.wikipedia.org/wiki/Server_Name_Indication) used in the 'Client Hello' message during TLS handshake. [2] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Recommended` when authenticating the client. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
antonfirsov, line 318: Shouldn't this be `tls.client.server_name`?

https://github.com/open-telemetry/semantic-conventions/blob/eb80b71d7f3dee6cf6df696333b2ce87e65816f2/docs/attributes-registry/tls.md

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,399 @@
...
+<!-- semconv dotnet.tls.handshake -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [1] | `TODO` | `Conditionally Required` if and only if an error has occurred. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
antonfirsov, line 317: This will be exception name, typically `System.Net.Security.Authentication.AuthenticationException` or `System.OperationCanceledException`.

Instead of `AuthenticationException` we can consider reporting platform-specific numeric error codes when available, but there would be little or no clue how to interpret them.

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,399 @@
...
+| [`server.address`](/docs/attributes-registry/server.md) | string | The [server name indication (SNI)](https://en.wikipedia.org/wiki/Server_Name_Indication) used in the 'Client Hello' message during TLS handshake. [2] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Recommended` when authenticating the client. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`tls.protocol.name`](/docs/attributes-registry/tls.md) | string | Normalized lowercase protocol name parsed from original string of the negotiated [SSL/TLS protocol version](https://www.openssl.org/docs/man1.1.1/man3/SSL_get_version.html#RETURN-VALUES) | `ssl`; `tls` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`tls.protocol.version`](/docs/attributes-registry/tls.md) | string | Numeric part of the version parsed from the original string of the negotiated [SSL/TLS protocol version](https://www.openssl.org/docs/man1.1.1/man3/SSL_get_version.html#RETURN-VALUES) | `1.2`; `3` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
antonfirsov, line 320: These will not be guaranteed to be available on failing authentication.

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,399 @@
...
+a new connection is established, so the duration of *Wait For Connection* span is different from duration of the [*HTTP connection setup*](#http-connection-setup) span.
+
+If a new connection was created to serve the request and the corresponding [*HTTP connection setup*](#http-connection-setup) was reported, the instrumentation adds the
+link to the *HTTP connection setup* from the *Wait for connection* span.
+
+Span name SHOULD be `wait_for_connection {server.address}:{server.port}`.
+
+Span kind SHOULD be `INTERNAL`
+
+Corresponding `Activity.OperationName` is `System.Net.Http.TODO.`, `ActivitySource` name - `System.Net.Http.TODO`.
```
antonfirsov, line 64: `System.Net.Http.Connections.RequestLink.WaitForConnection` ?

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,399 @@
...
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`error.type`](/docs/attributes-registry/error.md) | string | The full name of exception type. | `System.OperationCanceledException`; `System.Net.Sockets.SocketException` | `Conditionally Required` if and only if an error has occurred. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.peer.address`](/docs/attributes-registry/network.md) | string | Peer address of the network connection - IP address or Unix domain socket name. | `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.protocol.version`](/docs/attributes-registry/network.md) | string | The actual version of the protocol used for network communication. [1] | `1.1`; `2` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
antonfirsov, line 119: This is very complicated to fetch during connection setup due to possible downgrades, I'd prefer to drop it.

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,399 @@
...
+
+The span describes the establishment of the socket connection.
+It's different from [*HTTP connection setup*](#http-connection-setup) span, which also covers the DNS lookup and TLS handshake.
+
+When *Socket connect* span is reported along with *HTTP connection setup* span, the socket span becomes a child of HTTP connection setup.
+
+Span name SHOULD be `socket connect {network.peer.address}:{network.peer.port}`.
+Span kind SHOULD be `CLIENT`.
+
+Corresponding `Activity.OperationName` is `System.Net.Sockets.ConnectionSetup`, `ActivitySource` name - `System.Net.Sockets`.
```
antonfirsov, line 155: Given that the DispayName is `socket connect`, it's probably better to also change the OperationName.

```suggestion
Corresponding `Activity.OperationName` is `System.Net.Sockets.Connect`, `ActivitySource` name - `System.Net.Sockets`.
```

docs/attributes-registry/dns.md
lmolkova: it's registry section for attributes in `dns` namespace, we have `error.type` on the corresponding spans.

docs/dotnet/dotnet-connection-traces.md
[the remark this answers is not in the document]
lmolkova: based on the offline discussion let's not add any attributes to wait-for-connection for the time being since it's a child of HTTP request which has all the details

docs/dotnet/dotnet-connection-traces.md
lmolkova: I updated the description, could you please check if it's addressed now?

(also would appreciate span names suggestions) for forward/reverse lookup.

docs/dotnet/dotnet-connection-traces.md
lmolkova: I believe I addressed it

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,375 @@
...
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`error.type`](/docs/attributes-registry/error.md) | string | Socket error code. [1] | `connection_refused`; `address_not_available` | `Conditionally Required` if and only if an error has occurred. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.peer.address`](/docs/attributes-registry/network.md) | string | Peer address of the network connection - IP address or Unix domain socket name. | `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.peer.port`](/docs/attributes-registry/network.md) | int | Peer port number of the network connection. | `65123` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.type`](/docs/attributes-registry/network.md) | string | [OSI network layer](https://osi-model.com/network-layer/) or non-OSI equivalent. [2] | `ipv4`; `ipv6` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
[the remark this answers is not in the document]
lmolkova: we should report `network.transport` for unix domain sockets. Could we return it (report it if it's not TCP)?

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,412 @@
...
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## TLS handshake
+
+**Status**: [Experimental][DocumentStatus]
+
+The span describes TLS client or server handshake performed with [System.Net.Security.SslStream](https://learn.microsoft.com/dotnet/api/system.net.security.sslstream).
+
+Span name SHOULD be `TLS client {server.address}` when authenticating on the client side and `TLS server` when authenticating the server.
```
antonfirsov, line 317:
```suggestion
Span name SHOULD be `TLS client handshake {server.address}` when authenticating on the client side and `TLS server handshake` when authenticating the server.
```

docs/dotnet/dotnet-network-traces.md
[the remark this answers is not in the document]
antonfirsov: Replied in https://github.com/dotnet/runtime/pull/103922#discussion_r1673974879 so networking folks can see the discussion.

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,412 @@
...
+## TLS handshake
+
+**Status**: [Experimental][DocumentStatus]
+
+The span describes TLS client or server handshake performed with [System.Net.Security.SslStream](https://learn.microsoft.com/dotnet/api/system.net.security.sslstream).
+
+Span name SHOULD be `TLS client handshake {server.address}` when authenticating on the client side and `TLS server handshake` when authenticating the server.
+Span kind SHOULD be `INTERNAL` in both cases.
+
+When *TLS* span is reported for client-side authentication along with *HTTP connection setup* and *Socket connect* span, the *TLS* span becomes a child of *HTTP connection setup*.
```
bidetofevil, line 320: If I'm not mistaken, isn't the TLS handshake part of the socket creation process? If that's the case, should it be the child of that span if it exists?

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,412 @@
...
+Using connection-level instrumentation in production environments should be done after appropriate validation.
+
+## HTTP client request
+
+**Status**: [Stable][DocumentStatus]
+
+.NET `HttpClient` reports client request spans according to [HTTP Client Semantic Conventions](/docs/http/http-spans.md#http-client) with the following
+specific details:
+
+- `network.protocol.name`, `network.peer.port`, and `http.request.resend_count` are not reported
```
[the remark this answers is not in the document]
antonfirsov: It is very difficult to implement because of the way HttpClientHandler / SocketsHttpHandler internal handler chains are  structured.

docs/dotnet/dotnet-network-traces.md
antonfirsov: No, `System.Net.Sockets.Socket` is a thin wrapper over OS sockets, which do not do encryption.

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,412 @@
...
+
+
+`error.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `_OTHER` | A fallback error value to be used when the instrumentation doesn't define a custom value. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+
+
+`tls.protocol.name` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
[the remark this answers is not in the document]
lmolkova: sure, it's defined on the line 337

docs/dotnet/dotnet-network-traces.md
bidetofevil: But you can't write to a secure socket until the TLS handshake is done, right? I suppose if you're modelling the socket span as being representative of the creation of the socket but before protocol configuration is done, thus allowing it to be used to transmit data, this makes sense. 

I suppose my world view is influenced by how OkHttp's lifecycle events model this, bundling the configuration with the creation given a secure socket isn't truly ready to go until the TLS handshake is done.

If this makes more sense for the .NET, fair enough 😅

docs/dotnet/dotnet-network-traces.md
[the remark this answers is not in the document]
bidetofevil: Oops! I don't know how I missed this! It's... right below this...... 🤦

docs/dotnet/dotnet-network-traces.md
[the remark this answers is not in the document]
bidetofevil: Got it. So it's a practical limitation due to the implementation rather than a lack of utility? Fair enough

docs/dotnet/dotnet-network-traces.md
antonfirsov: >  bundling the configuration with the creation given a secure socket isn't truly ready to go until the TLS handshake is done.

We model the creation period with the new spans, or in our terminology, "connection setup".
There is no such thing as "secure socket" in lower level layers. The counterpart that best matches the "creation of a secure socket" is our "HTTP connection_setup" span, where "TLS client" has been already renamed to `TLS client handshake`:

image

codefromthecrypt:
@trustin @swankjesse @minwoox  I know this is a tall order, but I'll ask anyway since years ago you all had helped with this and led to zipkin annotations or otherwise.

Do you have some cycles to review otel semantics  (aka data recording conventions like tag names which are called attributes in otelngish)? I know you have all very deep work in interception and lifecycle of HTTP and there are some subtleties here which are more than meets the eye 🤖

lmolkova:
@antonfirsov could you please review and approve if it aligns with the implementation? Thanks!

docs/dotnet/dotnet-network-traces.md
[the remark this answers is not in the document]
lmolkova: added a note (`if `network.peer.address` is an IP address.`)

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,412 @@
...
+| [`network.peer.address`](/docs/attributes-registry/network.md) | string | Peer address of the network connection - IP address or Unix domain socket name. | `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.peer.port`](/docs/attributes-registry/network.md) | int | Peer port number of the network connection. | `65123` | `Recommended` [2] | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.transport`](/docs/attributes-registry/network.md) | string | [OSI transport layer](https://osi-model.com/transport-layer/) or [inter-process communication method](https://wikipedia.org/wiki/Inter-process_communication). [3] | `tcp`; `udp`; `unix` | `Recommended` [4] | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.type`](/docs/attributes-registry/network.md) | string | [OSI network layer](https://osi-model.com/network-layer/) or non-OSI equivalent. [5] | `ipv4`; `ipv6` | `Recommended` if `network.peer.address` is an IP address. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+
+**[1]:** The following errors codes are reported:
```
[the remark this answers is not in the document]
lmolkova: I'm fine not to list them, but curious why not (since we explicitly list them here https://github.com/dotnet/runtime/blob/c9148c8783ee953da8da033ae3222b9f24963e0c/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketsTelemetry.cs#L234 - is it expected to change?

lmolkova:
@antonfirsov I think I addressed you comments (all except listing error codes), could you please take another look?

model/dns/registry.yaml
```diff
@@ -16,3 +16,9 @@ groups:
           as escaped base 10 integers (\DDD). Back slashes and quotes should be escaped.
           Tabs, carriage returns, and line feeds should be converted to \t, \r, and
           \n respectively.
+      - id: dns.answers
```
jsuereth, line 19: This feels like a request/response logging mechanism.

Is there a reason .NET prefers these as span attributes vs. Events?  

Note: I don't want to open the giant can of worms of Events vs. Span Attributes.  I'm just trying to understand rationale here *and* from the non-.NET portion of semantic conventions, I don't think this impacts choice of span-attribute vs. event-attribute for `dns.answers`.

lmolkova: I just noticed your comment @jsuereth 

what would be the reason to report `dns.answers` in event?

Here's an example of this PR in action

image

DNS lookup takes 30ms and can fail, it's a valid span. And then there are a few addresses lookup returns, they make sense on the span that captures DNS lookup. 

Am I missing something?

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,427 @@
...
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`error.type`](/docs/attributes-registry/error.md) | string | The error code or exception name returned by [System.Net.Dns](https://learn.microsoft.com/dotnet/api/system.net.dns). [1] | `host_not_found`; `try_again` | `Conditionally Required` if and only if an error has occurred. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`dns.answers`](/docs/attributes-registry/dns.md) | string[] | List of resolved IP addresses (for DNS lookup) or a single element containing domain name (for reverse lookup). | `["10.0.0.1", "2001:0db8:85a3:0000:0000:8a2e:0370:7334"]` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
antonfirsov, line 183: Only reported on success.

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,427 @@
...
+
+- `host_not_found`
+- `try_again`
+- `no_recovery`
+- `address_family_not_supported`
+- the full exception type name
+
+See [SocketError](https://learn.microsoft.com/dotnet/api/system.net.sockets.socketerror) for more details.
+
+**[2] `dns.question.name`:** If the name field contains non-printable characters (below 32 or above 126), those characters should be represented as escaped base 10 integers (\DDD). Back slashes and quotes should be escaped. Tabs, carriage returns, and line feeds should be converted to \t, \r, and \n respectively.
```
antonfirsov, line 196: I'm not aware of any logic in the implementation that would do the `\DDD` conversion.

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,427 @@
...
+
+---
+
+`error.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
antonfirsov, line 200: Duplicate?

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,427 @@
...
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`error.type`](/docs/attributes-registry/error.md) | string | One of the [HTTP Request errors](https://learn.microsoft.com/dotnet/api/system.net.http.httprequesterror) in snake_case, or a full exception type. | `name_resolution_error`; `System.OperationCanceledException` | `Conditionally Required` if and only if an error has occurred. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.peer.address`](/docs/attributes-registry/network.md) | string | Peer IP address of the socket connection. [1] | `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`server.address`](/docs/attributes-registry/server.md) | string | Server domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name. [2] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
antonfirsov, line 128: Only on client

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,427 @@
...
+
+### HTTP request has to wait for connection setup and other requests on that connection to complete
+
+If connection was not immediately available for the request, `HttpClient` creates span for HTTP request and
+*Wait for connection* span. In this example, request was performed on a new connection, but this connection
+served other requests in the queue before it became available for this request.
+
+```
+<- HTTP connection_setup - (trace=t1, span=s1) ->
+                                <--------------------- GET / (trace=t2, span=s2) ------------------------------>
```
antonfirsov, line 403: these should not overlap

antonfirsov (approved):

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,427 @@
...
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`error.type`](/docs/attributes-registry/error.md) | string | One of the [HTTP Request errors](https://learn.microsoft.com/dotnet/api/system.net.http.httprequesterror) in snake_case, or a full exception type. | `version_negotiation_error`; `System.OperationCanceledException` | `Conditionally Required` if and only if an error has occurred. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+
+---
+
+`error.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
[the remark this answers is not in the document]
lmolkova: created https://github.com/open-telemetry/semantic-conventions/issues/2141 to record the feedback on the error.type table being confusing

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,427 @@
...
+Applications are encouraged to enable *HTTP client request* spans by default in production environments.
+
+Connection-level spans are experimental - their semantics may be changed in the future in a breaking manner.
+Using connection-level instrumentation in production environments should be done after appropriate validation.
+
+## HTTP client request
+
+**Status**: [Stable][DocumentStatus]
+
+.NET `HttpClient` reports client request spans according to [HTTP Client Semantic Conventions](/docs/http/http-spans.md#http-client) with the following
```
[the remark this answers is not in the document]
lmolkova: rephrased a bit in https://github.com/open-telemetry/semantic-conventions/pull/1192/commits/11e71b3555c3686cc91d0900eee89ea184344468

docs/dotnet/dotnet-network-traces.md
lmolkova: tooling limitation, captured in https://github.com/open-telemetry/semantic-conventions/issues/2141

docs/dotnet/dotnet-network-traces.md
lmolkova: I think it's a bug in the common attribute definition, removed the note from .net span and created https://github.com/open-telemetry/semantic-conventions/issues/2143

docs/dotnet/dotnet-network-traces.md
```diff
@@ -0,0 +1,462 @@
...
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The span describes the time it takes for the HTTP request to obtain a connection from the connection pool.
+
+The span is reported only if there was no connection readily available when request has started.
+It's reported as a child of *HTTP client request* span.
+
+The span ends when the connection is obtained - it could happen when an existing connection becomes available or once
+a new connection is established, so the duration of *Wait For Connection* span is different from duration of the
```
lmolkova, line 79: let's try to generalize it and try to address similar problems - concurrency, mutexes, etc, look at absl

model/dns/registry.yaml
jsuereth: I agree the time taken to call DNS is a span, it's more whether the answers should be on the span.

I think given the direction we're now taking, we're not moving request/response to Event signal generically, so consider this comment resolved.

jsuereth (approved):

docs/dotnet/dotnet-network-traces.md
lmolkova: let's follow up in  https://github.com/open-telemetry/semantic-conventions/issues/2395

PR #1196, opened 2024-06-28
Experiment with `db.operation.target`

### LINKED ISSUE #1132: db.collection.name - reconsider only capturing the first collection name
From the DB stabilisation working group.

Currently the DB SemConv for [`db.collection.name`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#common-attributes) states that: 

> The collection name MAY be parsed from the query text, in which case it SHOULD be the first collection name found in the query.

It came up in the Elasticsearch PR that there are very good reasons to store all collection names and storing the first one would be confusing: https://github.com/open-telemetry/semantic-conventions/pull/1002#issuecomment-2150436821

Although other database systems may be different, we discussed this in the WG and a few things came up:
- This decision opens questions, we anticipate that users often will ask why only the first one is captured? Currently this can't be answered by looking at the spec
- Even in the WG we were not able to recall why this was the decision 

Notes from meeting:

- Ordering - is a,b,c the same as c,b,a
  - Alphabetical
  - As present in the query
- Cardinality
- Format (csv?)
  - db.collection.names
- It’s part of the span name (back to format)
- For metrics, make query text as opt-in

Goal of the issue: re-iterate this decision
- [ ] Try to come up with a solution that does not raise above questions and potentially stored all collection names. 
- [ ] If we stick to this decision at least try to document why and make sure the spec has good reasoning for this, plus it should also enable to store all collection names in specific db spec (see Elasticsearch:  https://github.com/open-telemetry/semantic-conventions/pull/1002)

lmolkova:
Fixes #1132


docs/database/database-spans.md
```diff
@@ -56,25 +56,25 @@ Database spans MUST follow the overall [guidelines for span names](https://githu
...
 
 Semantic conventions for individual database systems MAY specify different span name format.
 
-The <span id="target-placeholder">`{target}`</span> SHOULD describe the entity that the operation is performed against
+The <span id="entity-placeholder">`{entity}`</span> SHOULD describe the entity that the operation is performed against
 and SHOULD adhere to one of the following values, provided they are accessible:
 
-- `db.collection.name` SHOULD be used for data manipulation operations or operations on database collections.
+- `db.operation.target` SHOULD be used for data manipulation operations or operations on database collections.
 - `db.namespace` SHOULD be used for operations on a specific database namespace.
```
lmolkova, line 74: I think it's confusing:

E.g. I have `CREATE DATABASE` query or `cosmosDatabase.listContainers()` operation, the `db.operation.target` semantics would imply that database namespace is a target.

I believe we still want distinct attributes for db namespace and table/container/collection/index(es).

gregkalapos (commented):
I agree with https://github.com/open-telemetry/semantic-conventions/pull/1196#discussion_r1659181421.

Another thing which comes to my mind when I see this is that `db.collection.name` is an attribute name that very well describes what ends up in this field. I'd say for most of the cases (let's take you have any SQL based DB library), it's easy to understand what ends up there. I'd argue `db.operation.target` is not that easy to understand for most cases. Yes, it may potentially solves the issue from https://github.com/open-telemetry/semantic-conventions/issues/1132, but my impression is that we'd pay the price in complexity, since this field name is less obvious.

So, overall I'd rather keep this as is.

lmolkova:
Closing this based on the DB WG discussion on 7/10. The consensus is that `db.operation.target` is confusing and does not allow to separate database name and table name.

Issue #1174, opened 2024-06-21
GenAI: how to record embeddings


lmolkova:
- Which request parameters are important (in addition to model)?
  - encoding_format?
  - dimensions?
  - should they be in `gen_ai.request` namespace and mixed with chat params?

- Should we record input in an event? 
  - embeddings calls are probably much cheaper and frequent, recording input for each one could be too expensice
  - if we record inputs, they should probably be enabled independently from chats

Issue #963, opened 2024-04-24
LLM: Add support for other modalities


TaoChenOSU:
### Area(s)

area:gen-ai

### Is your change request related to a problem? Please describe.

With the initial version of LLM conventions now established, it is an opportunity to begin contemplating the addition of support for other modalities in the gen-ai domain, including text-to-image and image-to-text.

### Describe the solution you'd like

Incorporate additional gen-ai modalities within the LLM conventions framework.

lmolkova:
closing in favor of #1556

PR #1202, opened 2024-07-02
Convert `gen_ai.operation.name` to enum and use it on spans 

lmolkova:
Replaces open-telemetry/semantic-conventions#995

## Changes

- Converts `gen_ai.operation.name` to enum. Clarifies name for the `chat` operation. 
- Adds it to span attributes
- Uses it in span name
- adds minor clarifications for `gen_ai.system`


docs/attributes-registry/gen-ai.md
```diff
@@ -13,8 +13,8 @@ This document defines the attributes used to describe telemetry in the context o
...
 | `gen_ai.completion`                | string   | The full response received from the GenAI model. [1]                                             | `[{'role': 'assistant', 'content': 'The capital of France is Paris.'}]` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `gen_ai.operation.name`            | string   | The name of the operation being performed.                                                       | `chat`; `completion`                                                    | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `gen_ai.prompt`                    | string   | The full prompt sent to the GenAI model. [2]                                                     | `[{'role': 'user', 'content': 'What is the capital of France?'}]`       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `gen_ai.operation.name`            | string   | The name of the operation being performed. [2]                                                   | `chat`; `completions`                                                   | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
nirga, line 16: why `completions` and not `completion`?
https://docs.anthropic.com/en/api/complete

docs/attributes-registry/gen-ai.md
```diff
@@ -26,18 +26,27 @@ This document defines the attributes used to describe telemetry in the context o
...
 
-**[2]:** It's RECOMMENDED to format prompts as JSON string matching [OpenAI messages format](https://platform.openai.com/docs/guides/text-generation)
+**[2]:** If one of the predefined values applies, but specific system uses a different name it's RECOMMENDED to document it in the semantic conventions for specific GenAI system and use system-specific name in the instrumentation. If a different name is not documented, instrumentation libraries SHOULD use applicable predefined value.
 
-**[3]:** The actual GenAI product may differ from the one identified by the client. For example, when using OpenAI client libraries to communicate with Mistral, the `gen_ai.system` is set to `openai` based on the instrumentation's best knowledge.
+**[3]:** It's RECOMMENDED to format prompts as JSON string matching [OpenAI messages format](https://platform.openai.com/docs/guides/text-generation)
+
+**[4]:** The actual GenAI product may differ from the one identified by the client. For example, when using OpenAI client libraries to communicate with Mistral, the `gen_ai.system` is set to `openai` based on the instrumentation's best knowledge.
```
cartermp, line 40: Recommend making reference to `gen_ai.request.model` here to clarify that instrumentation won't be oblivious to the actual model used (even though, yes, it's detailed further in the spec)

docs/attributes-registry/gen-ai.md
```diff
@@ -26,18 +26,27 @@ This document defines the attributes used to describe telemetry in the context o
...
 For custom model, a custom friendly name SHOULD be used. If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
 
+`gen_ai.operation.name` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value         | Description                                                                                                               | Stability                                                        |
+| ------------- | ------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------- |
+| `chat`        | Chat completion operation such as [OpenAI Chat API](https://platform.openai.com/docs/api-reference/chat)                  | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `completions` | Legacy completions operation such as [OpenAI Completions API](https://platform.openai.com/docs/api-reference/completions) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
cartermp, line 48:
```suggestion
| `completions` | Completions operation such as [Legacy OpenAI Completions API](https://platform.openai.com/docs/api-reference/completions) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

Recommend not calling `completions` as a name legacy, since an API provider may choose to use this noun.

docs/attributes-registry/gen-ai.md
lmolkova: Just to stay  consistent with OpenAI name [`completions` (plural)](https://platform.openai.com/docs/api-reference/completions)
Cohere calls it `generate` - https://docs.cohere.com/reference/generate-1

If it's `completion` then it does not align with anyone, which is also fine (since it's legacy everywhere), but then I'd then prefer not to define it at all

docs/attributes-registry/gen-ai.md
drewby: May be out of scope for this PR. This is the generated text which was updated from [3] to [4] which is why its showing up as a change.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -71,22 +71,32 @@ This metric SHOULD be specified with [ExplicitBucketBoundaries] of [1, 4, 16, 64
...
-**[3]:** When observed from the client side, and when communicating through an intermediary, `server.address` SHOULD represent the server address behind any intermediaries, for example proxies, if it's available.
 
 
+`gen_ai.operation.name` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `chat` | Chat completion operation such as [OpenAI Chat API](https://platform.openai.com/docs/api-reference/chat) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `completions` | Legacy completions operation such as [OpenAI Completions API](https://platform.openai.com/docs/api-reference/completions) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
```
drewby, line 99: Should we add `images` and `embeddings` since they are well known today?

docs/gen-ai/gen-ai-spans.md
```diff
@@ -20,8 +21,11 @@ A request to an Generative AI is modeled as a span in a trace.
 
 **Span kind:** MUST always be `CLIENT`.
 
-The **span name** SHOULD be set to a low cardinality value describing an operation made to an Generative AI system.
-For example, the API name such as [Create chat completion](https://platform.openai.com/docs/api-reference/chat/create) could be represented as `completion gpt-4` to include the API and the model name.
+## Name
+
+GenAI spans MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.33.0/specification/trace/api.md#span).
+The **span name** SHOULD be `{gen_ai.operation.name} {gen_ai.request.model}`.
+Semantic conventions for individual GenAI systems MAY specify different span name format.
```
drewby, line 28: Customers I've worked with want to add a meaningful span name here based on their workflow (some workflow engines like LangChain and Promptflow may do this too). In the context of the application, its low cardinality. For example, "Summarize" instead of "Completions" to differentiate from "Generate response".  This may be in a trace flow such as `Embeddings` -> `Retrieval` -> `Summarize` -> `Generate response`. The last two both being Completions (or Chat) operations. 

Should we be clearer that this is OK?

lmolkova: would users instrument such calls themselves? 
I.e. is the ask to allow it more explicitly or to change the operation names for everyone?

docs/attributes-registry/gen-ai.md
nirga: We agreed on `text_completion` in our weekly meeting

docs/gen-ai/gen-ai-spans.md
cartermp: Hmmm. I think it would be good for more domain-specific instrumentation (like using specific langchain or llamaindex abstractions) to have a generic-ish name they can use if that's the abstraction they're providing. We could offer guidance on what that name should be, and ask that it's a verb name of low cardinality, and list some examples like above.

drewby: I think a library (langchain, etc) may want to provide a default that is more descriptive than Operation Name and Model. And a user may want to override it with a domain specific name, but the guidance should remain that it is low cardinality.

lmolkova: I'm updating this sentence to be 
> Semantic conventions for individual GenAI systems and frameworks MAY specify different span name format.

Do you feel it gives enough freedom to use a different pattern? Or is there something different you'd like to have there?

The cardinality requirements are part of the general guidelines for span name linked on line 26.

docs/gen-ai/gen-ai-metrics.md
lmolkova: I'd add embeddings in scope of https://github.com/open-telemetry/semantic-conventions/issues/1174 and images in scope of https://github.com/open-telemetry/semantic-conventions/issues/963

Adding just the operation name would not  be enough to properly instrument these operations.

docs/gen-ai/gen-ai-spans.md
drewby: Yes, I think that's enough to support the requirement.

docs/gen-ai/gen-ai-metrics.md
drewby: Works for me.

docs/attributes-registry/gen-ai.md
lmolkova: I added 

> The `gen_ai.system` describes a family of GenAI models with specific model identified
by `gen_ai.request.model` and `gen_ai.response.model` attributes.

at the beginning of the note, lmk if it helps

joaopgrassi (approved):

nirga (approved):

docs/attributes-registry/gen-ai.md
cartermp: It does!

drewby (approved):
Looks good.

codefromthecrypt:
drive by feedback, the description of this PR, while significantly changing a fundamental field (span name) does cover "what" like mechanics, but not "why"

Can we here and in the future put the motivation for renovations such as these, especially when doing things like enums we expect to break later? It will help be able to jump into a PR and not just assess, yeah.. this changes the span name, but also understand why we are considering it will drift all instrumenters.

karthikscale3:
Just wanted to add a note from Langtrace's point of view, regarding the convention for span names, it’s a bit tricky to make the change in our case for us since our SDK provides an option to disable tracing certain methods if the user wants more control and as a result we set span name as the method name. We will have to make some changes to the "disable tracing logic" if we update this - so treading a bit cautiously here. Will make this change pretty soon.

PR #1203, opened 2024-07-02
Add time and memory related db operation metrics

### LINKED ISSUE #723: DB metrics: add query-level metrics
Currently, there are some conventions for connection-level metrics defined on the [Database Metrics page](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md). Proposing that query-level metrics are added (in a different section on the same page) to define conventions for the stats and info that can be captured for individual queries.

Some examples could include:

`db.query.count` (number of times a query was executed)
`db.query.plan` (execution plan for a query)
`db.query.duration` (amount of time taken by the query)
`db.query.rows_returned` (number of rows returned by a query)

dnaik-sfx:
Fixes #723 (split off from #1186)

## Changes

Adding additional db operation metrics related to memory and time spent.


.chloggen/add-db-operation-time-and-memory-metrics.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add databse operation metrics related to memory and time spent
```
cijothomas, line 13:
```suggestion
note: Add database operation metrics related to memory and time spent
```

joaopgrassi:
CC @open-telemetry/semconv-db-approvers

model/metrics/database-metrics.yaml
```diff
@@ -33,6 +33,154 @@ groups:
...
+        requirement_level:
+          recommended: If and only if `network.peer.address` is set.
+      - ref: db.namespace
+        requirement_level:
+          conditionally_required: If available.
+      - ref: thread.id
+        requirement_level:
+          conditionally_required: If available
+
+  - id: metric.db.client.operation.pages_scanned
```
maryliag, line 112: since not all databases use pages, I don't think it makes sense to have this on the "general" view of the database. If there are things that only make sense for a specific group, is worth having those split from here.

model/metrics/database-metrics.yaml
```diff
@@ -33,6 +33,154 @@ groups:
       - ref: db.namespace
         requirement_level:
           conditionally_required: If available.
+      - ref: thread.id
```
maryliag, line 36: I'm curious about the addition of the thread.id on all of these. Do you have a use case in mind with this?

model/metrics/database-metrics.yaml
dnaik-sfx: I definitely get the reasoning there and don't necessarily disagree, but if the concept of "pages" (or another example) is shared by multiple databases, either now or in the future, would it make more sense to have a common metric name for them or multiple db-specific names? And could it be more confusing to have some metrics have common names while others would fall under db-specific namespaces? Happy to move these metrics elsewhere if these topics have already been discussed/answered though.

model/metrics/database-metrics.yaml
dnaik-sfx: My colleague mentioned the rationale in his comment [here](https://github.com/open-telemetry/semantic-conventions/pull/1186#discussion_r1662996947)

In short, for the case of single-threaded operations, the thread id could be useful for troubleshooting slow/blocked queries, and it might also be helpful to aggregate stats on a certain thread for diagnosing performance issues. But we're open to discussion on the attribute and dropping it from the spec if the consensus is that these use cases are too niche and/or not really practical.

jsuereth: Does thread id add to the cardinality in an uncontrolled way here?

model/metrics/database-metrics.yaml
```diff
@@ -33,6 +33,154 @@ groups:
...
+          Network peer address and port are useful when the application interacts with individual database nodes directly.
+
+          If a database operation involved multiple network calls (for example retries), the address of the last contacted node SHOULD be used.
+      - ref: network.peer.port
+        requirement_level:
+          recommended: If and only if `network.peer.address` is set.
+      - ref: db.namespace
+        requirement_level:
+          conditionally_required: If available.
+      - ref: thread.id
```
jsuereth, line 72: Nit: maybe say "if available and stable for the lifetime of the operation".

IF using, e.g. async IO and the handler of the response is different than the calling thread - you don't want to create new timeseries because thread.id changes.  Can we make that more explicit?

dnaik-sfx: Good call, done

PR #1208, opened 2024-07-03
Add policy enforcement for attribute registry. 

jsuereth:
- Add REGO rules to enforce attribute registry things.
- Add diagnostic rendering to improve error messages on attribute
  registry failure.

Example failure:

```
$ make check-policies
✔ SemConv registry loaded (138 files)
✔ SemConv registry resolved
Semantic Convention Model Validation Failures:

/source/metrics/jvm-metrics.yaml:
  Attribute jvm.memory.bad_name cannot be defined in Group attributes.jvm.memory
  Attribute Group not in Attribute Registry.
```

Replaces #1014


policies/registry.rego
```diff
@@ -0,0 +1,42 @@
...
+		"id": violation_id,
+		"type": "semconv_attribute",
+		"category": "attribute_registry",
+		"group": group_id,
+		"attr": attr_id,
+	}
+}
+
+# We only allow attribute groups in the attribute registry.
+deny[attr_registry_violation("registry_must_be_attribute_group", group.id, "")] {
```
lmolkova, line 19: I'm playing with collisions and would like to put something like `Attribute 'messaging.client_id' has the same constant name 'messaging_client_id' as ["messaging.client.id"]`  somewhere.

There seem to be no other place to put it except `violation.id`, so I wonder if we could use violation id as an error message.

policies/registry.rego
```diff
@@ -0,0 +1,42 @@
...
+# This file enforces policies requiring all attributes to be defined within
+# a semantic convention "registry".  This is a naming/structure convention
+# used by semantic conventions.
+
+# Helper to create attribute registry violations.
+attr_registry_violation(violation_id, group_id, attr_id) = violation {
+	violation := {
+		"id": violation_id,
+		"type": "semconv_attribute",
+		"category": "attribute_registry",
```
lmolkova, line 12: nit: not sure how to use it/if it's important, what should go there

policies/registry.rego
```diff
@@ -0,0 +1,42 @@
...
+	attr := group.attributes[_]
+	attr.id != null
+}
+
+# A registry `attribute_group` containing at least one `ref` attribute is
+# considered invalid if it's not in the registry group.
+deny[attr_registry_violation("registry_with_ref_attr", group.id, attr.ref)] {
+	group := input.groups[_]
+	startswith(group.id, "registry.")
+    attr := group.attributes[_]
```
lmolkova, line 39: nit: indentation

lmolkova (approved):

policies/registry.rego
lmolkova: Do we really need jinja for it? 😢

templates/diagnostic/ansi/errors.txt.j2
```diff
@@ -0,0 +1,31 @@
...
+  Attribute Group not in Attribute Registry.
+{%- elif item.error.violation.id == "registry_with_ref_attr" %}
+  Attribute {{ item.error.violation.attr | ansi_cyan }} cannot be defined in Group {{ item.error.violation.group | ansi_green }}
+  Registry cannot contain attribute references.
+{%- else %}
+UNKNOWN ATTRIBUTE REGISTRY ERROR! Please open a ticket agianst semconv with this info: 
+{{ debug(item) }}
+{%- endif %}
+{%- else %}
+{{ item.error.provenance | ansi_red }}:
```
lmolkova, line 21: the violation structure is rigit and doesn't work well for collisions. I'd rather not build any infra around it (at least for now). See my other comment on free-form error desription

policies/registry.rego
lmolkova: See https://github.com/open-telemetry/semantic-conventions/pull/1209 for the example

policies/registry.rego
```diff
@@ -0,0 +1,42 @@
...
+
+# This file enforces policies requiring all attributes to be defined within
+# a semantic convention "registry".  This is a naming/structure convention
+# used by semantic conventions.
+
+# Helper to create attribute registry violations.
+attr_registry_violation(violation_id, group_id, attr_id) = violation {
+	violation := {
+		"id": violation_id,
+		"type": "semconv_attribute",
```
lmolkova, line 11: also not sure about type - I assume we could do cross-signal checks (metric collision with attribute name), id characters, etc. 
What would be the benefit of having the classification? 

(I know it's required property, but I think it shouldn't be)

policies/registry.rego
jsuereth: Jinja just lets us control the error message.  If you're happy with the default, you can use it.

I wanted to do better than the default, which I don't think is super readable, so I made a format.  What I like is we don't need weaver changes to update error messages or policy violations.

If you'd rather use sprintf in rego, that's fine.  I still think the error I show above is more readable.

ALSO - once MINIJINJA supports `groupby` from JINJA we can actually sort error messages by file instance (for before_resolution), etc. etc.

policies/registry.rego
jsuereth: These are kind of all up to use to determine. I just made some random things here, but happy to alter them.

E.g. we can *just* use "type" 'attribute_registry" *or* we could use "type" "attribute_group, etc." to denote what the error is against.

templates/diagnostic/ansi/errors.txt.j2
jsuereth: We should probably push for violation to NOT be rigid, but remain as pure JSON/structured data if we go with JINJA.  I was somehow under the assumption it wasn't rigid anymore, but I didn't really dive deeply yet.

If it *is* truly rigid, then you're right, we should take your error message approach.

policies/registry.rego
lmolkova: The concern is that policies will have inconsistent output - some will return `registry_must_be_attribute_group` as violation id and other will return free-form error description. The more complicated policies are, the more likely that `Violation` would not have properties to describe all  details.

E.g. in case of collisions I have to return the problematic attribute name and a list of all collisions.

Since we'll have the same makefile check for all policies and the same jinja template, the template would need to treat the output differently depending on something. It feels like a big complication to anyone writing policies - they also need to adjust output formatting.

So I think we need consistency and given that all we need is for humans to read error messages, having something simple and obvious would make things easier for 
- policy authors
- reviewers
- someone who tries to decode butchered error message

templates/diagnostic/ansi/errors.txt.j2
lmolkova: I could not add arbitrary properties to it and the only way I could report all the details I needed was through abusing `violation.id`. I'm arguing in https://github.com/open-telemetry/semantic-conventions/pull/1208#discussion_r1665928442 that all we need (at least now) is to give a meaningful error message to human beings and I don't see the need for this error message to be spread between policy and jinja template.

policies/registry.rego
jsuereth: We discussed this in the tooling SiG.  I think you have the right requirements here, specifically:

1. We want the ability to define both POLICY and HUMAN-READABLE ERROR in the same location
2. I'd like to see some weaver improvements in this area to give us the flexibility we need.

So going forward, we're going to investigate better "string creation" in Rego *and* a better Violation structure in weaver that can take advanatage of that.

I've moved to having the readability of error messages rely on just the violation ID for now.  Once weaver supports better errors, we can improve this.

policies/registry.rego
jsuereth: I made a new name for category here, but agree it's unclear how we should use it.

policies/registry.rego
```diff
@@ -0,0 +1,42 @@
...
+	attr.id != null
+}
+
+# A registry `attribute_group` containing at least one `ref` attribute is
+# considered invalid if it's not in the registry group.
+deny[attr_registry_violation("registry_with_ref_attr", group.id, attr.ref)] {
+	group := input.groups[_]
+	startswith(group.id, "registry.")
+    attr := group.attributes[_]
+	attr.ref != null
```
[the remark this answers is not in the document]
jsuereth: Given our discussion yesterday, I'm going to leave this as-is and we can update it when we land embed/prefix into semconv.

policies/registry.rego
jsuereth: See above discussion.  We're going to spend some time fixing violation requirements in weaver, either going more flexible or *simpler* on what's included so you can do more in Rego.

policies/registry.rego
jsuereth: Oof, need to find a code formatter for rego...

joaopgrassi (approved):

PR #1209, opened 2024-07-03
Add attribute name and const names checks

### LINKED ISSUE #1068: Add attribute naming checks

lmolkova:
Fixes #1068
Implements phase 1 of https://github.com/open-telemetry/semantic-conventions/issues/1118#issuecomment-2173803006

image


Makefile
```diff
@@ -228,3 +228,11 @@ chlog-update: $(CHLOGGEN)
...
 generate-gh-issue-templates:
 	$(TOOLS_DIR)/scripts/update-issue-template-areas.sh
+
+.PHONY: check-policies
+check-policies:
+	docker run --rm -v $(PWD)/model:/source -v $(PWD)/docs:/spec -v $(PWD)/policies:/policies \
+		otel/weaver:${WEAVER_VERSION} registry check \
+		--registry=/source \
+		--policy=/policies/registry.rego \
+		--policy=/policies/attribute_name_collisions.rego
```
lquerel, line 238: Idea (@lmolkova , @jsuereth): Another option is to move all the Rego files inside the registry itself so they are automatically discovered and applied by `weaver registry check`. Weaver looks at `*.rego` files inside the registry directory by default. To make this more flexible, we could look at `**/*.rego` to retrieve all policy files, even those inside subdirectories of the registry (e.g. `$REGISTRY/policies`). Having the registry YAML and policies under the same folder makes sense to me.

lmolkova: I'd prefer to keep the definition and checks separate (I can run these checks on any data whether it's in the semconv repo or not and I can use data without checking it) .

I'd be nice to have any way to pass `*.rego` though - I was going to ask if it's possible.

lquerel (approved):
LGTM. I just left two comments that I don’t think are blocking.

Makefile
lquerel: Make sense. I created the following GH issue to keep track of this request. https://github.com/open-telemetry/weaver/issues/259

policies/attribute_name_collisions.rego
```diff
@@ -0,0 +1,46 @@
...
+    description := sprintf("Attribute '%s' is used as a namespace in attributes %s", [name, collisions])
+}
+
+attr_registry_collision(description, attr_name) = violation {
+    violation := {
+        "id": description,
+        "type": "semconv_attribute",
+        "category": "naming_collision",
+        "attr": attr_name,
+        "group": "",
```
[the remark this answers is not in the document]
lmolkova: Attribute definitions are unique, so it should not be a problem. The reason I didn't want to include group name is that there are two colliding attributes and it's not clear which group to mention.

I'm open to adding both groups into the human-readable error if we see that locating attributes becomes difficult.

policies/registry.rego
```diff
@@ -5,38 +5,59 @@ package before_resolution
 # used by semantic conventions.
 
 # Helper to create attribute registry violations.
-attr_registry_violation(violation_id, group_id, attr_id) = violation {
-	violation := {
```
joaopgrassi, line 9: Maybe we should consider adding an editorconfig here to avoid the tabs/space mixes.

joaopgrassi (approved):

policies/registry.rego
lmolkova: great point, following up in https://github.com/open-telemetry/semantic-conventions/pull/1286

policies/attribute_name_collisions.rego
```diff
@@ -0,0 +1,49 @@
...
+}
+
+attr_names_except(excluded) = names {
+    names := { n | n := input.groups[_].attributes[_].name } - excluded
+}
+
+# TODO - we'll need to specify how collision resolution happens in the schema -
+# see phase 2 in https://github.com/open-telemetry/semantic-conventions/issues/1118#issuecomment-2173803006
+# For now just allow current collisions.
+excluded_const_collisions := {"messaging.client_id"}
```
jsuereth, line 48: Can you make a big warning about how we won't allow new items in these lists?

policies/attribute_name_collisions.rego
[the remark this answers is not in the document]
jsuereth: I agree with the concern, but we can follow up on this.

I think we need more support in weaver on `Violation` and provenance of attribute/group in the model we through at Rego to improve this.

For now, let's not let perfect be the enemy of good.  We need violation checks that prevent breaking codegen.

jsuereth (approved):
Just one nit.

policies/attribute_name_collisions.rego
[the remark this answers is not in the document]
lmolkova: Created  https://github.com/open-telemetry/weaver/issues/279 to track

policies/attribute_name_collisions.rego
lmolkova: added

Issue #1181, opened 2024-06-25
Design a `process.status` metric


braydonk:
### Area(s)

area:system

### Is your change request related to a problem? Please describe.

In https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33731 we received a proposal to add `process.state` as a resource attribute. Due to the changing nature of the a process status within a process run, I believe it does not make sense as part of a process resource attribute. I'd like to engage the System Semantic Conventions working group to design it instead as a metric.

### Describe the solution you'd like

I am not sure. :slightly_smiling_face: 

Process Status is generally a string result, and on Linux at least is a restricted set: running, uninterruptible sleep, interruptable sleep, stopped, zombie. Metric design isn't my strong suit, so I would like to hear from someone with a good idea for this, or if anyone is able to point to a similar enum-style metric.

braydonk:
cc @open-telemetry/semconv-system-approvers

ishleenk17:
I do see `system.process.status` in the semantic conventions. https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/docs/attributes-registry/system.md#system-process-attributes

ChrsMark:
Thank's @braydonk!

Regarding the immutability concern: we are facing this in several other places like at https://github.com/open-telemetry/semantic-conventions/pull/997#issuecomment-2129538490 and https://github.com/open-telemetry/semantic-conventions/issues/1160#issuecomment-2176190366. This seems to be another one.

Regarding the `process.status`, do you propose another attribute different than https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/docs/attributes-registry/system.md#system-process-attributes which @ishleenk17 also mentioned?

AlexanderWert:
@tigrannajaryan This is another case for the need of mutable resource attributes.

Is there already some central place where the Entity WG is tracking these cases? If not would it make sense to have single issue to reference them all?

tigrannajaryan:
> Is there already some central place where the Entity WG is tracking these cases? If not would it make sense to have single issue to reference them all?

This repository is the best place to track them. Entity SIG is not currently working on specific cases. We are working on generic data model. The work on specific cases for concrete entity types will happen in this repo.

braydonk:
[...]

PR #1212, opened 2024-07-08
process: add process.status attribute

braydonk:
Related Issue: #1181 

## Changes
This PR deprecates `system.process.status` and replaces it with `process.status`. The previous usage of `system.process.status` has been replaced with `process.status`, and `process.status` has also been add as a resource attribute to the `process` resource.


schema-next.yaml
```diff
@@ -46,6 +46,12 @@ versions:
         # https://github.com/open-telemetry/semantic-conventions/pull/1006
         - rename_metrics:
             messaging.publish.messages: messaging.client.published.messages
+        # https://github.com/open-telemetry/semantic-conventions/pull/1212
+        - rename_attributes:
+            attribute_map:
+              system.process.status: process.status
+            apply_to_metrics:
+              - system.processes.count
```
dmitryax, line 54: Shouldn't we include all the other process metrics here?

dmitryax (approved):

schema-next.yaml
braydonk: This attribute was only previously used on `system.processes.count`, so the `rename` in particular is only to that usage of the attribute. Its usage as a resource attribute is new, which means it doesn't get represented in this file (only renames not additive changes are in `schema-next`).

ChrsMark (approved):

model/registry/deprecated/system.yaml
```diff
@@ -50,5 +50,25 @@ groups:
               stability: experimental
         brief: "Deprecated, use `cpu.mode` instead."
         stability: experimental
-        deprecated: 'Replaced by `cpu.mode`'
+        deprecated: "Replaced by `cpu.mode`"
         examples: [ "idle", "interrupt" ]
+      - id: system.process.status
+        type:
+          allow_custom_values: true
```
lmolkova, line 57: nit
```suggestion
```

value is ignored and the flag will be removed at some point

model/registry/process.yaml
```diff
@@ -196,3 +196,23 @@ groups:
               value: 'minor'
               stability: experimental
         stability: experimental
+      - id: status
+        type:
+          allow_custom_values: true
```
lmolkova, line 201:
```suggestion
```

same here, value is ignored

model/resource/process.yaml
```diff
@@ -22,7 +23,9 @@ groups:
       - ref: process.command_args
         requirement_level:
           conditionally_required: See [Selecting process attributes](#selecting-process-attributes) for details.
-      - ref: process.owner
+      - ref: process.status
+        requirement_level:
+          conditionally_required: See [Selecting process attributes](#selecting-process-attributes) for details.
```
lmolkova, line 28: I think `Selecting process attributes` applies to things that identify the process, I guess status is not one of them and would be `recommended: when applicable` or just `recommended`?

model/resource/process.yaml
```diff
@@ -8,6 +8,7 @@ groups:
       - ref: process.pid
       - ref: process.parent_pid
       - ref: process.executable.name
+      - ref: process.owner
```
lmolkova, line 11: oops, this change resulted in changing  `requirement_level` on `process.executable.name`.

ChrsMark (changes requested):
Let's wait for an outcome from https://github.com/open-telemetry/semantic-conventions/issues/1181#issuecomment-2231223216.

lmolkova:
Based on the discussion in SemConv SIG on 8/12, the only blocker is listing `process.status` as a resource attribute. We can unblock this PR (if it makes sense) by defining  `process.status` attribute in the registry, but not referencing it in `process` resource definition. It can also be used on metrics as a regular attribute.

braydonk:
I could do that if it would help. I don't suspect anything in the `process` or `system` namespace will use this attribute, but if there may be a use case for it I can adjust this PR to just be a definition in the registry.

ChrsMark:
Hey @braydonk, I wonder if this could be modeled as a metric instead. It was also suggested at https://github.com/open-telemetry/semantic-conventions/issues/1032#issuecomment-2420399295 and it seems it's already done for the `hw.status` : https://opentelemetry.io/docs/specs/semconv/system/hardware-metrics/#hwbattery---battery-metrics

braydonk:
This is going to be redone to model it the same as `hw.status` as @ChrsMark mentioned above. Since this is so old, I'm just going to close this and redo it.

PR #1216, opened 2024-07-08
Deprecate `tls.client.server_name` attribute in favor of common `server.address`

### LINKED ISSUE #1211: Could we remove `tls.client.server_name` in favor of `server.address`
`tls.client.server_name` is defined as 
https://github.com/open-telemetry/semantic-conventions/blob/5971366ae2fcaf2d01b1a52da92e872b7df5491d/docs/attributes-registry/tls.md?plain=1#L25

This should be the same as `server.address`. See https://community.f5.com/kb/technicalarticles/ssl-profiles-part-7-server-name-indication/279829

So the proposal is to deprecate the `tls.client.server_name` attribute.

lmolkova:
Fixes #1211


schema-next.yaml
```diff
@@ -4,6 +4,10 @@ versions:
   next:
     all:
       changes:
+        # https://github.com/open-telemetry/semantic-conventions/pull/1216
+        - rename_attributes:
+          attribute_map:
+            tls.client.server_name: server.address
```
joaopgrassi, line 10: The rename attribute here is because _when_ used, tls instrumentations should report `server.address` instead, right? Because since this attribute only existed in the registry we don't add now the `server.address` in `tls.yaml` so it's more of a delete than a rename?.

joaopgrassi: Yes, that's what I meant, but one could also see as delete, since it was never part of any convention that used it (it was only on the registry), but I guess this makes sense.

joaopgrassi (approved):

trisch-me (approved):

PR #1217, opened 2024-07-09
Add 'log.record.original' attribute

### LINKED ISSUE #1137: Propose new "log.original.body" attribute
### Area(s)

area:log

When working with traditional string logs, the log's Body field is sometimes overloaded. This happens because the field is defined such that it is intended for multiple purposes which were previously believed to be mutually exclusive. Paraphrasing from the data model, the body is intended as the appropriate place for any of the following:
1. A complete original representation of a structured log
2. A complete original representation of an unstructured (string) log
3. The "message" portion of a log

However, as demonstrated by examples in https://github.com/open-telemetry/opentelemetry-specification/issues/3932, it is sometimes necessary and often wise to retain more than one of the above within the same payload. Since we will not add a field to disambiguate the "message" from the "original" values, I'm proposing that we define a semantic convention to act as an optional backup location for the original log.

A new attribute: `log.original.body`
Data type: string or []byte
Description: A copy of the Log Record's original Body value. If the field contains a value, it MUST be exactly the value which the Body first contained. When this field contains a value, it SHOULD be assumed that the Body field was changed.

This follows from a lengthy [discussion about adding this as a top-level field in the log data model](https://github.com/open-telemetry/opentelemetry-specification/issues/3932). Consensus in that conversation appeared to be strongly in favor of a semantic convention instead.

### Additional context

_No response_

djaglowski:
Fixes #1137

## Changes

Add 'log.record.original' attribute.


joaopgrassi (commented):
I skimmed through the linked spec issue and I see most folks agree with a semantic conventions for this, but I wonder what will be the outcome of having this convention here? Will then receivers have to build knowledge of this semantic field? What is the overall plan? I ask because your last comment there says there's no support for it so I wonder what the convention will solve.

docs/general/logs.md
```diff
@@ -44,9 +44,12 @@ These attributes may be used for identifying a Log Record.
...
+| [`log.record.original`](/docs/attributes-registry/log.md) | string | The complete orignal Log Record. [1] | `77 <86>1 2015-08-06T21:58:59.694Z 192.168.2.133 inactive - - - Something happened`; `[INFO] 8/3/24 12:34:56 Something happened` | `Opt-In` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`log.record.uid`](/docs/attributes-registry/log.md) | string | A unique identifier for the Log Record. [2] | `01ARZ3NDEKTSV4RRFFQ69G5FAV` | `Opt-In` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
-**[1]:** If an id is provided, other log records with the same id will be considered duplicates and can be removed safely. This means, that two distinguishable log records MUST have different values.
+**[1]:** This value SHOULD be provided if the Log Record was originally transmitted as a string or equivalent data type AND the Body field of the Log Record does not contain the same value. (e.g. A syslog or log record read from a file.)
```
tigrannajaryan, line 50:
```suggestion
**[1]:** This value SHOULD be provided if the Log Record was originally transmitted as a string or equivalent data type AND the Body field of the Log Record does not contain the same value. (e.g. a syslog or log record read from a file.)
```

tigrannajaryan (approved):

model/registry/log.yaml
```diff
@@ -70,3 +70,14 @@ groups:
...
           but other identifiers (e.g. UUID) may be used as needed.
         examples: ["01ARZ3NDEKTSV4RRFFQ69G5FAV"]
+      - id: original
+        type: string
+        stability: experimental
+        brief: >
+          The complete orignal Log Record.
+        note: >
+          This value SHOULD be provided if the Log Record was originally transmitted as a string or equivalent data type AND
+          the Body field of the Log Record does not contain the same value. (e.g. a syslog or log record read from a file.)
```
lmolkova, line 80: backends impose limits on attribute value length (smaller than on log body)  and it might be truncated. E.g. Az Monitor [limits](https://learn.microsoft.com/en-us/azure/azure-monitor/app/data-model-complete#custom-properties-4) attribute value to 8KB, event payload to 32KB, New Relic has other [limits](https://docs.newrelic.com/docs/data-apis/manage-data/view-system-limits/).

Given that log bodies could be huge, I could make sense to truncate them on the collection side or at least mention that truncation might happen on the backend.

docs/general/logs.md
```diff
@@ -44,9 +44,12 @@ These attributes may be used for identifying a Log Record.
...
+| [`log.record.original`](/docs/attributes-registry/log.md) | string | The complete orignal Log Record. [1] | `77 <86>1 2015-08-06T21:58:59.694Z 192.168.2.133 inactive - - - Something happened`; `[INFO] 8/3/24 12:34:56 Something happened` | `Opt-In` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`log.record.uid`](/docs/attributes-registry/log.md) | string | A unique identifier for the Log Record. [2] | `01ARZ3NDEKTSV4RRFFQ69G5FAV` | `Opt-In` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
-**[1]:** If an id is provided, other log records with the same id will be considered duplicates and can be removed safely. This means, that two distinguishable log records MUST have different values.
+**[1]:** This value SHOULD be provided if the Log Record was originally transmitted as a string or equivalent data type AND the Body field of the Log Record does not contain the same value. (e.g. a syslog or log record read from a file.)
```
lmolkova, line 50: since it's opt-in, SHOULD seems a bit confusing.

Maybe 
```suggestion
**[1]:** This value SHOULD be provided (if provided at all) if and only if the Log Record was originally transmitted as a string or equivalent data type AND the Body field of the Log Record does not contain the same value. (e.g. a syslog or log record read from a file.)
```

lmolkova, line 50: it seems that intended component that'd implement this convention is a processor (SDK or collector) - could we mention it? 

I.e. if I'm the end user writing logs, I should not be adding this attribute - correct? If so, it'd be great to explain it.

djaglowski: I changed this to MAY along with some other rewording for your suggestion about it being a downstream concern.

djaglowski: I reworded this a bit to hopefully emphasize that the value is added by the modifier.

model/registry/log.yaml
djaglowski: I'm not sure I agree with adding any language about truncation here. Any limit we would try to establish would be arbitrary in my opinion. I suppose we could mention backends but it seems like a given that backends will consume data according to their own limitations.

lmolkova (approved):

joaopgrassi (approved):
@djaglowski thanks for the explanation!

PR #1231, opened 2024-07-10
Proposal for vector db semantic convention

ezimuel:
This is a proposal for vector db semantic convention (see open-telemetry/semantic-conventions-genai#5). I tried to expand the `db` semantic convention adding some `db.vector` attributes. I tried to focus on the basic needs of a general purpose vector database.

I proposed the following experimental attributes (updated with the feedbacks in this PR):
```
db.search.similarity_metric: specify the metric used in similarity search (e.g. cosine)
db.record.id: the ID of the record (e.g. the ID of the vector)
db.vector.field_name: the name field of the vector embedding
db.vector.dimension_count: the dimension of the vector (e.g. 1536)
db.query.limit: the number of vectors returned by a query (e.g. top-k)
```
The operations performed in a vector db, such as `insert`, `update`, `search` and `delete` can be performed using the existing [db.operation.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/db/) attribute.

Regarding the similarity search we can use the [db.query](https://opentelemetry.io/docs/specs/semconv/attributes-registry/db/) attributes, such as `db.query.parameter.<key>`.


model/registry/db.yaml
```diff
@@ -310,6 +310,10 @@ groups:
               value: 'trino'
               brief: 'Trino'
               stability: experimental
+            - id: vector
```
gregkalapos, line 313: Isn't just `vector` a bit off in this list?

I'd expect the `db.system` to either one of the well known dbs - like `db.system: elasticsearch` could be used when elasticsearch is used as a vector db. So mostly this will be the specific db system, right?

And then there are other cases, when the db system may not be (yet) on this list. In that case, similar to `other_sql` wouldn't `other_vector` be a better fit?

ezimuel: I agree, `other_vector` is a better fit here.

docs/database/dynamodb.md
```diff
@@ -83,6 +83,7 @@ These attributes are filled in for all DynamoDB request types.
 | `sybase` | Sybase | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `teradata` | Teradata | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `trino` | Trino | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `vector` | vector | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 86: Many of the databases listed here support vector search. E.g. mongoDB, cosmosDB, cassandra and probably others. A lot of databases are adding vector search due to it's high demand in AI world. Only one `db.system` can be specified on a telemetry item.

So, let's avoid adding a DB feature as a system. Instead, let's define attributes specific to vector search that could be used on any database that supports them.

model/registry/db.yaml
lmolkova: see my comment on the `db.system` above

model/registry/db.yaml
```diff
@@ -529,3 +533,67 @@ groups:
...
+          The ID of vector.
+        examples: '5c56c793-69f3-4fbf-87e6-c4bf54c28c26'
+      - id: name
+        type: string
+        stability: experimental
+        brief: >
+          The name field as of the vector (e.g. a field name).
+        examples: 'vector'
+      - id: embeddings
+        type: double[]
```
lmolkova, line 583: this would be huge and not usable on telemetry - is it necessary?

model/registry/db.yaml
```diff
@@ -529,3 +533,67 @@ groups:
...
+        brief: >
+          The values of the vector, the array of numbers.
+        examples: [0.9, 0.1, 0.1]
+      - id: dimension
+        type: int
+        stability: experimental
+        brief: >
+          The dimension of the vector.
+        examples: [3]
+      - id: model
```
lmolkova, line 594: this should be captured with `gen_ai.request.model` (and other gen-ai attributes) - it's ok to mix different attributes on the same telemetry item.

model/registry/db.yaml
```diff
@@ -529,3 +533,67 @@ groups:
...
+        type: double[]
+        stability: experimental
+        brief: >
+          The values of the vector, the array of numbers.
+        examples: [0.9, 0.1, 0.1]
+      - id: dimension
+        type: int
+        stability: experimental
+        brief: >
+          The dimension of the vector.
```
lmolkova, line 592: is it the number of dimensions? let's call it something like `db.vector.dimension_count` and also update the brief

model/registry/db.yaml
```diff
@@ -529,3 +533,67 @@ groups:
...
+        type: string
+        stability: experimental
+        brief: >
+          The ID of vector.
+        examples: '5c56c793-69f3-4fbf-87e6-c4bf54c28c26'
+      - id: name
+        type: string
+        stability: experimental
+        brief: >
+          The name field as of the vector (e.g. a field name).
```
lmolkova, line 580: do we need both - the name and the id?

model/registry/db.yaml
```diff
@@ -529,3 +533,67 @@ groups:
           reference the [elasticsearch schema](https://raw.githubusercontent.com/elastic/elasticsearch-specification/main/output/schema/schema.json)
           in order to map the path part values to their names.
         examples: ['db.elasticsearch.path_parts.index=test-index', 'db.elasticsearch.path_parts.doc_id=123']
+  - id: registry.db.vector
+    prefix: db.vector
+    type: attribute_group
+    brief: >
+      This group defines attributes for vector databases.
+    attributes:
+      - id: similarity
```
lmolkova, line 542: suggesting more explicit name like `db.vector.search.similarity_metric`

model/registry/db.yaml
```diff
@@ -529,3 +533,67 @@ groups:
...
+        type:
+          allow_custom_values: true
```
lmolkova, line 544:
```suggestion
```

we no longer use this flag, all enums are open

docs/database/dynamodb.md
ezimuel: This means you are in favor to have a generic `vector` attribute to capture a generic vector search not specific to any vendor? This was the goal of this PR, that's why I proposed only few attributes after checking some vendors. But I think the `other_vector` proposed by @gregkalapos makes more sense if the vendor is not listed. The goal is the same.

model/registry/db.yaml
ezimuel: If you refer to the fact that the embedding values can be huge (eg. 1536) I agree, but I think we need to have a way to represent a vector in the semantic convention. We can mark this as optional.

model/registry/db.yaml
ezimuel: This is the `model` for the embedding and can be different from the `gen_ai.request.model`. Vector databases are not strictly related with GenAI. There are many use cases where vector db can be used without GenAI (e.g. semantic search).

model/registry/db.yaml
ezimuel: The `id` is the identifier of the vector, the `name` is the field name that contains the vector. Many database uses both but some uses only `name` of the field. Maybe we should use a better naming here.

model/registry/db.yaml
ezimuel: There are still `allow_custom_values: true` in the existing `db.yaml`. Do we need to remove?

model/registry/db.yaml
lmolkova: why do we need to record it on the telemetry? It can be done with logs (and common ways to control verbosity), not as a span attribute

docs/database/dynamodb.md
lmolkova: do we have some real-life case of an instrumentation that does not know the product name of the database?
If so, `other_vector` would make sense.
`other_sql` exists because of JDBC/ODBC/etc drivers and ORMs which might not know what database is used underneath.

If specific databases we need are not listed here, let's just list them and use their names.

model/registry/db.yaml
karthikscale3: +1, it could be recorded with events or logs. Curious to hear what use case drives the requirement to record this?

karthikscale3:
Thanks for creating this PR. A few additional attributes which we instrument today with our SDK that we have found useful are the following:
`
db.index
db.namespace
db.collection.name
db.top_k
db.query
`

Thoughts on the ones listed above? cc @lmolkova

docs/database/dynamodb.md
ezimuel: There are some proposals for ORM with vector database and some other framework like [Spring AI](https://docs.spring.io/spring-ai/reference/api/vectordbs.html) or [Semantic Kernel](https://learn.microsoft.com/en-us/semantic-kernel/integrations/#available-connectors-to-vector-databases) that are trying to achieve this. But right now I tend to agree, there are no many real-life case.

model/registry/db.yaml
ezimuel: There's not a specific use case, as semantic convention I would expect to have the embedding values represented in a proposal about vector database (at lest as optional) . Anyway, if you think that it's better to have it in logs for me it's fine.

ezimuel:
@karthikscale3 regarding the attributes that you proposed, some already exists:
- `db.index` can be represented using `db.operation.name = index`;
- `db.namespace` already exists;
- `db.collection.name` already exists;
- `db.query` can be represented using `db.query.text` and also `db.query.parameter`;

The `top_k` proposal I think it's a good idea:
- `db.vector.query.top_k` to represent the k most similar vectors returned by a similarity serach;

Moreover, I found very interesting the proposal of [OpenLLMetry](https://github.com/traceloop/openllmetry) project especially the part regarding the attributes for vector db, [here](https://github.com/traceloop/openllmetry):
```
 # Vector DB
 VECTOR_DB_VENDOR = "db.system"
 VECTOR_DB_OPERATION = "db.operation"
 VECTOR_DB_QUERY_TOP_K = "db.vector.query.top_k"
```

ezimuel:
@lmolkova I applied all the feedbacks, thanks for the review. @karthikscale3 I added the `top_k` attribute, thanks.

Summary of the changes:
- removed `vector` in `db.system`;
- removed the `db.vector.embeddings`;
- renamed `db.vector.dimension` in `db.vector.dimension_count`;
- added the `db.vector.query.top_k` as suggested by @karthikscale3;

karthikscale3:
> @karthikscale3 regarding the attributes that you proposed, some already exists:
> 
> * `db.index` can be represented using `db.operation.name = index`;
> * `db.namespace` already exists;
> * `db.collection.name` already exists;
> * `db.query` can be represented using `db.query.text` and also `db.query.parameter`;
> 
> The `top_k` proposal I think it's a good idea:
> 
> * `db.vector.query.top_k` to represent the k most similar vectors returned by a similarity serach;
> 
> Moreover, I found very interesting the proposal of [OpenLLMetry](https://github.com/traceloop/openllmetry) project especially the part regarding the attributes for vector db, [here](https://github.com/traceloop/openllmetry):
> 
> ```
>  # Vector DB
>  VECTOR_DB_VENDOR = "db.system"
>  VECTOR_DB_OPERATION = "db.operation"
>  VECTOR_DB_QUERY_TOP_K = "db.vector.query.top_k"
> ```

Yea that sounds good! And yes, my intention was to reuse the existing ones. Wasn't sure if we needed them redefined for the sake of vector dbs or not. But sounds like its unnecessary.

karthikscale3:
> @lmolkova I applied all the feedbacks, thanks for the review. @karthikscale3 I added the `top_k` attribute, thanks.
> 
> Summary of the changes:
> 
> * removed `vector` in `db.system`;
> * removed the `db.vector.embeddings`;
> * renamed `db.vector.dimension` in `db.vector.dimension_count`;
> * added the `db.vector.query.top_k` as suggested by @karthikscale3;
> * removed the `allow_custom_values: true` in `db.yaml`, see last commit;

Thank you! From my side, everything looks good. We discussed this PR in today's working group call and @nirga wanted to take a deeper look at it once again.

karthikscale3 (approved):

docs/attributes-registry/db.md
```diff
@@ -6,12 +6,22 @@
 
 # Db
 
+<<<<<<< HEAD
```
trask, line 9: looks like a bad merge

docs/database/dynamodb.md
```diff
@@ -9,7 +9,90 @@ linkTitle: AWS DynamoDB
 The Semantic Conventions for [AWS DynamoDB](https://aws.amazon.com/dynamodb/) extend and override the general
 [AWS SDK Semantic Conventions](/docs/cloud-providers/aws-sdk.md) and [Database Semantic Conventions](database-spans.md).
 
+<<<<<<< HEAD
```
trask, line 12: and here

ezimuel:
I fixed the merge issues. Thanks @trask

model/registry/db.yaml
ezimuel: @lmolkova what do you think about this? Thanks.

nirga (commented):
Thanks, that's a great start! I wonder if we want to add specific spans that use these attributes in this PR as well?

ezimuel:
@nirga can you give me an example of specific span? FYI, I'm going offline and I'll come back August 4 for further discussion.

nirga:
Sorry, nvm I think this is already covered as part of the DB semconv

model/registry/db.yaml
maryliag: @ezimuel you marked this as resolved, but looks like you missed this change, it still shows `dimension` instead of `dimention_count`

docs/attributes-registry/db.md
```diff
@@ -199,6 +200,28 @@ This group defines attributes for Elasticsearch.
 
 **[8]:** Many Elasticsearch url paths allow dynamic values. These SHOULD be recorded in span attributes in the format `db.elasticsearch.path_parts.<key>`, where `<key>` is the url path part name. The implementation SHOULD reference the [elasticsearch schema](https://raw.githubusercontent.com/elastic/elasticsearch-specification/main/output/schema/schema.json) in order to map the path part values to their names.
 
+## Db Vector Attributes
```
maryliag, line 203: nit: Vector Database Attributes

model/registry/db.yaml
maryliag: @ezimuel I think you also missed this change, is marked as resolved without any updates :)

maryliag:
@ezimuel looks like you missed some of the changes you marked as resolved:
- rename `db.vector.dimension` to `db.vector.dimension_count`: it is still showing `db.vector.dimension`
- rename `db.vector.similarity` to `db.vector.search.similarity_metric`: it is still showing `db.vector.similarity`

model/registry/db.yaml
lmolkova: could you provide some examples of databases that can *compute* embeddings? 

How would database instrumentation know which model was used to create embeddings if it only stores and queries them?

model/registry/db.yaml
lmolkova: Could you provide some examples of id and name? 

Let's pick a few (2-3) popular databases and explain what id and name would mean in their context.

E.g. I look into [MongoDB](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/#about-the-vector-type) and I don't understand what id or name would be.

Or I look into [Azure Search](https://learn.microsoft.com/en-us/azure/search/vector-store) and don't understand what should go into id.

I look into [pinecone](https://docs.pinecone.io/guides/get-started/quickstart#5-create-a-serverless-index) and it does not talk about ids.

Also, perhaps by vector you mean index?

model/registry/db.yaml
```diff
@@ -533,3 +530,66 @@ groups:
...
+        brief: >
+          The dimension of the vector.
+        examples: [3]
+      - id: model
+        type: string
+        stability: experimental
+        brief: >
+          The model used for the embedding.
+        examples: 'text-embedding-3-small'
+      - id: query.top_k
```
lmolkova, line 590: I think we should come up with a more common attribute not specific to vector dbs. 
Many databases allow to limit number of returned rows:
- JDBC has [Statement.setMaxRows](https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html#setMaxRows(int)), 
- Mongo allows to set a [limit](https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/crud/read-operations/limit/#specify-a-limit)

Suggesting `db.query.max_returned_items` or similar

lmolkova:
We need to reference new attributes in the database spans conventions (see https://github.com/open-telemetry/semantic-conventions/blob/main/model/trace/database.yaml), specifically on the conventions for the databases we have there which support vector search. 

We should describe how new attributes apply to them.

ezimuel:
> @ezimuel looks like you missed some of the changes you marked as resolved:
> 
> * rename `db.vector.dimension` to `db.vector.dimension_count`: it is still showing `db.vector.dimension`
> * rename `db.vector.similarity` to `db.vector.search.similarity_metric`: it is still showing `db.vector.similarity`

I just fixed it, thanks @maryliag

model/registry/db.yaml
ezimuel: Fixed in https://github.com/open-telemetry/semantic-conventions/pull/1231/commits/a3330ffea04ad5d2ce8d9a4bb07a1ce259369339

model/registry/db.yaml
ezimuel: Fixed in https://github.com/open-telemetry/semantic-conventions/pull/1231/commits/a3330ffea04ad5d2ce8d9a4bb07a1ce259369339

model/registry/db.yaml
ezimuel: @lmolkova Here some examples:
- MongoDB uses name for the field (i.e. `path` in search [here](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/#ann-examples) and `field-name` in the definition [here]( https://www.mongodb.com/docs/atlas/atlas-search/field-types/knn-vector/#define-the-index-for-the-fts-field-type-type))
- Azure Search uses name ([here](https://learn.microsoft.com/en-us/azure/search/vector-search-how-to-create-index?tabs=config-2024-07-01%2Crest-2024-07-01%2Cpush%2Cportal-check-index#add-a-vector-field-to-the-fields-collection))
- Pinecone uses id ([here](https://docs.pinecone.io/guides/get-started/quickstart#6-upsert-vectors))
- Qdrant uses id (Point ID, [here](https://qdrant.tech/documentation/concepts/points/#point-ids))
- Elasticsearch uses name ([here](https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html))
- Milvus uses id ([here](https://milvus.io/docs/insert-update-delete.md#Insert-entities))
- Chroma uses id ([here](https://docs.trychroma.com/getting-started#4.-add-some-text-documents-to-the-collection))
- pgvector for PostgreSQL uses name (i.e. the field name with type vector(x), [here](https://github.com/pgvector/pgvector?tab=readme-ov-file#getting-started))
- Redis uses name (i.e. the field name that will contain the vector values in  `fieldname_embedding`, [here](https://redis.io/docs/latest/develop/get-started/vector-database/#4-generate-text-embeddings))

By **db.vector.name** I mean the **field name** in the document that will be used for embedding (most common for DB like MongoDB, Azure Search, Redis, Elasticsearch, PostgreSQL) and by **db.vector.id** I mean the identifier of the vector (most used in native vector db like Pinecone, Qdrant, Chroma, Milvus). 

Basically these attributes, **name** and **ID**, are used to identify the vector, maybe a better naming for `db.vector.name` can be `db.vector.field_name`?

model/registry/db.yaml
ezimuel: @lmolkova I see the similarity here but I think the `db.vector.query.top-k` is more specific from a semantic point of view and more related to vectors, since it specifies the top k results in order, starting from the most similar. In semantic search we have this **similarity** value that is always present in any result that we don't have in standard database. The `limit` parameter of SQL returns the first k results but not in order, it depends on how you build the query (e.g. using `ORDER BY`).
 I personally think we should keep `db.vector.query.top-k` and potentially add a `db.query.limit` (or `db.query.max_returned_items` as you suggested) in a separate PR.

model/registry/db.yaml
lmolkova: Thank you for the context!

> name

I agree that `db.vector.field_name` would be more descriptive.
It could be difficult to collect it though - for cosmos, postgres and many other dbs it would require parsing queries (and specific parsing for vector search). I.e. it'd probably mean that very few generic DB instrumentations will do it. 

Even the native instrumentation that we have in CosmosDB does not know if query supplied by user is doing vector search. 

So I wonder how critical this attribute is for the observability purposes. Also I'd be mostly static - is it important enough to justify additional costs of populating it on each span?

A  typical semconv decision making hint is: if we're not sure we need it, let's not add it. Adding attributes to spans is easy, removing them is breaking and hard.

lmolkova: > id

I still don't understand what's behind `db.vector.id` - it seems to be a generic record id and there is nothing vector-specific here. 

I support adding generic attribute like `db.record.id: string` or `db.record.ids: string[]` (needs polishing)

model/registry/db.yaml
lmolkova: Out of the few dbs I checked, they use `limit` in vector search
- pgvector uses [traditional `limit`](https://github.com/pgvector/pgvector?tab=readme-ov-file#getting-started) - same with cosmos and other sql databases
- mongo uses [imit](https://www.mongodb.com/docs/atlas/atlas-vector-search/tutorials/vector-search-quick-start/#std-label-vector-search-quick-start)
- qdrant uses [limit](https://qdrant.tech/documentation/concepts/search/#search-api)


So we're saying that DB instrumentations will need to detect if query is related to vector search or not and depending on this populate top-k or limit. That's difficult or impossible, but most importantly inconsistent and depends on instrumentation capabilities.

I.e. instrumentations that don't have vector-db specifics and those that do will use different attributes for the same thing.

So, I'd still prefer `db.query.limit` or something similar (and it should be under the same condition as `db.query.text` - we cannot require instrumentations to do query parsing)

model/registry/db.yaml
lmolkova: BTW, who/how/when is going to populate it? 

It seems it's only available at index creation time and not available at query or insert time. I.e. very rarely.

lmolkova:
> Thanks for creating this PR. A few additional attributes which we instrument today with our SDK that we have found useful are the following:

> db.index
> db.namespace
> db.collection.name
> db.top_k
> db.query

The index name I believe should be captured in the `db.collection.name` - we store and query against this collection/index

As @ezimuel already pointed out some are already defined (except top-k) and we have discussions around it (https://github.com/open-telemetry/semantic-conventions/pull/1231#discussion_r1710484074).

model/registry/db.yaml
ezimuel: Some vendors that provide in-database embedding:

- [Weaviate](https://weaviate.io/), supporting the following [provider integrations](https://weaviate.io/developers/weaviate/model-providers);
- [Elasticsearch](https://www.elastic.co/), using the [Inference API](https://www.elastic.co/guide/en/elasticsearch/reference/current/inference-apis.html);
- [PostgresML](https://postgresml.org/) provides [in-database embedding generation](https://postgresml.org/docs/open-source/pgml/guides/embeddings/in-database-generation);
- [Vespa](https://vespa.ai/) provides the embedding using the [embed()](https://docs.vespa.ai/en/embedding.html) function;
- [Supabase](https://supabase.com/) provides an embegging generator using [edge function](https://supabase.com/docs/guides/ai/quickstarts/generate-text-embeddings);

I think many other vendors will add this feature. Having the possibility to generate embeddings in the database simplify the customer use cases.

Regarding the question on how database instrumentation know which model was used to create an embedding this is an information that is specified when you create a collection or when you provide some search (e.g. in [PostgreML](https://postgresml.org/docs/open-source/pgml/guides/embeddings/in-database-generation#creating-embeddings)). Moreover, I think the instrumentation libraries can also leverage this `model` attribute since they know the embedding model used to generate the vectors.

model/registry/db.yaml
ezimuel: The `db.vector.id` is the record id but specific for the vector. The `db.record.id` will work as well, since it's just an identifier for the record (i.e. vector).

I think we should add `db.vector.field_name` since it is used in many database to specify which field has been used for the embedding.

model/registry/db.yaml
ezimuel: Correct, this is generally provided during the index creation but some databases uses it also in the query (e.g. [Qdrant](https://qdrant.tech/documentation/concepts/search/#metrics)).

I think we can also leverage the [Conditionally Required](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) level for some attributes that are not always available, like the `db.vector.search.similarity_metric`, WDYT?

model/registry/db.yaml
ezimuel: @lmolkova do you agree that `top-k` and `limit` are two different concepts, based on my previous comment? If they are I think we cannot use a single attribute (e.g. `db.query.limit`) to manage both.

ezimuel:
@lmolkova I provided the following changes:
- renamed `db.vector.name` in `db.vector.field_name`;
- removed `db.vector.id` in favor of `db.record.id`.

I think the only missing point is about `db.vector.query.top-k`, see my [last comment](https://github.com/open-telemetry/semantic-conventions/pull/1231#discussion_r1722969696).
Thanks!

model/registry/db.yaml
ezimuel: @lmolkova just a reminder for this, thanks.

model/registry/db.yaml
ezimuel: @lmolkova just a reminder for this, thanks.

lmolkova: @ezimuel I see that databases use both terms to describe the same thing (see my comment above).

Let's say you have a postgres query like `SELECT * FROM items ORDER BY embedding <-> '[3,1,2]' LIMIT 5;` - the general-purpose DB instrumentation can report `limit`. If you make it understand vector search syntax, it may be able to use `top_k` instead, but that's would be inconsistent and unfamiliar for those who use vector search in postgres.

model/registry/db.yaml
lmolkova: Thanks, it makes sense! 

BTW I think the attribute should be `db.search.similarity_metric` or similar - similarity is not always based on vectors and we don't need to limit this attribute.

lmolkova:
thanks for working on this @ezimuel !

Please make sure to update actual database semantic conventions and reference attributes you're adding under those that report them - https://github.com/open-telemetry/semantic-conventions/pull/1231#issuecomment-2251896744

model/registry/db.yaml
lmolkova: It seems all of them use some GenAI model (in-process or external) and provide an integration layer with it. 

I'm not sure what benefit defining a new attribute for databases brings. If embeddings become cross-domain concerns, let's find a generic attribute name for the vectorization model that will be reused between DB and GenAI. For now I strongly recommend to avoid adding new attribute.

model/registry/db.yaml
ezimuel: @lmolkova I see your point but I think `top-k` has a different meaning from `limit`. If you are using a relation database as vector db, `limit` is fine since you are building an SQL statement and you specify the order. But, if you are using a native vector database (e.g. Qdrant), the `top-k` is more relevant since `top` implies the order, using a similarity metric.

I think we should add both:
- `db.query.limit`
- `db.vector.query.top-k`

ezimuel:
@AlexanderWert I updated the PR and applied the suggestions from @lmolkova. I did the following changes:
- removed the `db.vector.model` as reported [here](https://github.com/open-telemetry/semantic-conventions/pull/1231#discussion_r1771747809);
- moved `db.vector.search.similarity_metric` to `db.search.similarity_metric`, as suggested [here](https://github.com/open-telemetry/semantic-conventions/pull/1231#discussion_r1771733525);

The only open question is about `limit` vs `top-k`, see my [last comment](https://github.com/open-telemetry/semantic-conventions/pull/1231#discussion_r1775087483).

@lmolkova I didn't understand what I supposed to do in [this comment](https://github.com/open-telemetry/semantic-conventions/pull/1231#issuecomment-2251896744), since the link https://github.com/open-telemetry/semantic-conventions/blob/main/model/trace/database.yaml does not work. Can you clarify? Thanks.

lmolkova:
@ezimuel  I'm sorry I did not reply earlier. I'm swamped with some work at the moment and, unfortunately, it might take me some time to reply.

If you look into https://github.com/open-telemetry/semantic-conventions/tree/main/docs/database you'd see that we have some docs for individual database systems - we reference attributes from the registry and we explain how these attributes are applied to this system (or not applied). 

This is powered by the yaml in https://github.com/open-telemetry/semantic-conventions/tree/main/model/database.

Please look at the existing database conventions and update them to include new attributes you're adding.

ezimuel:
@lmolkova thanks for the information and sorry also on my side for the late reply, very busy period.
I'm looking into the examples that you shared and I'll send a commit soon.

model/registry/db.yaml
ezimuel: @lmolkova I found that the `SELECT LIMIT` is not part of the SQL standard. A statement compliant with SQL standard is `FETCH FIRST`. Moreover, I discovered that we already have `db.operation.parameter,<key>` in the semantic convention. This means, we can have `db.operation.parameter.fetch-first`. I don't think we need to add a `limit` or `fetch-first` in `db.query`. What do you think?

At the same time, I believe we should include `db.vector.query.top-k`. As noted earlier, it differs in meaning from `fetch-first` and, additionally, this parameter is much easier to retrieve in a vector database because it's not part of any statement.

ezimuel:
@lmolkova and @AlexanderWert I finally found some time to work on this PR. I’ve added the spans definition and updated the documentation. I hope everything looks good. Looking forward to your review. Thanks!

model/database/spans.yaml
```diff
@@ -764,3 +764,81 @@ groups:
       - ref: db.cosmosdb.regions_contacted
         requirement_level:
           conditionally_required: If available.
+
+  - id: span.db.vector.client
```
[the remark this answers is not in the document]
AlexanderWert: agree, but how about doing it in a follow up issue / PR, as it is additive information?

model/database/spans.yaml
```diff
@@ -764,3 +764,81 @@ groups:
...
+          conditionally_required: If available.
+        examples: ['image_vector', 'embedding_field']
+      - ref: db.vector.dimension_count
+        brief: the dimension of the vector (e.g. 1536)
+        note: >
+          The dimension of a vector is typically defined when building an index/collection.
+        requirement_level: recommended
+        examples: [3, 1536]
+      - ref: db.vector.query.top_k
+        brief: the top-k most similar vectors returned by a query
```
AlexanderWert, line 833:
```suggestion
        brief: The top-k most similar vectors returned by a query
```

model/database/spans.yaml
```diff
@@ -764,3 +764,81 @@ groups:
...
+        examples: ['1', '5c56c793-69f3-4fbf-87e6-c4bf54c28c26']
+      - ref: db.vector.field_name
+        brief: the name field of the vector embedding
+        note: >
+          Some vector databases use a field name to store the vector.
+        requirement_level:
+          conditionally_required: If available.
+        examples: ['image_vector', 'embedding_field']
+      - ref: db.vector.dimension_count
+        brief: the dimension of the vector (e.g. 1536)
```
AlexanderWert, line 827:
```suggestion
        brief: The dimension of the vector (e.g. 1536)
```

model/database/spans.yaml
```diff
@@ -764,3 +764,81 @@ groups:
...
+        examples: ['cosine', 'dot', 'euclidean', 'manhattan']
+      - ref: db.record.id
+        brief: the ID of the record (e.g. the ID of the vector)
+        note: >
+          Some vector databases identify a vector using an ID.
+        requirement_level:
+          conditionally_required: If available.
+        examples: ['1', '5c56c793-69f3-4fbf-87e6-c4bf54c28c26']
+      - ref: db.vector.field_name
+        brief: the name field of the vector embedding
```
AlexanderWert, line 820:
```suggestion
        brief: The name field of the vector embedding
```

model/database/spans.yaml
```diff
@@ -764,3 +764,81 @@ groups:
...
+        examples: ['my_index', 'index1, index2']
+      - ref: db.search.similarity_metric
+        brief: specify the metric used in similarity search (e.g. cosine)
+        note: >
+          Some vector databases allow specifying the similarity search during a search, while others only allow it when creating an index or collection.
+        requirement_level:
+          conditionally_required: If available.
+        examples: ['cosine', 'dot', 'euclidean', 'manhattan']
+      - ref: db.record.id
+        brief: the ID of the record (e.g. the ID of the vector)
```
AlexanderWert, line 813:
```suggestion
        brief: The ID of the record (e.g. the ID of the vector)
```

model/database/spans.yaml
```diff
@@ -764,3 +764,81 @@ groups:
...
+      - ref: db.collection.name
+        sampling_relevant: true
+        requirement_level: recommended
+        brief: The index or data stream against which the query is executed.
+        note: >
+          The query may target multiple indices or data streams, in which case it SHOULD be a comma separated list of those.
+          If the query doesn't target a specific index, this field MUST NOT be set.
+        examples: ['my_index', 'index1, index2']
+      - ref: db.search.similarity_metric
+        brief: specify the metric used in similarity search (e.g. cosine)
```
AlexanderWert, line 806:
```suggestion
        brief: Specifies the metric used in similarity search (e.g. cosine)
```

model/database/spans.yaml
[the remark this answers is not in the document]
lmolkova: I don't think we should introduce `span.db.vector.client` span - it's not clear what/when/how will report it. 
Also, referencing attributes in the specific conventions is a good test if there attributes are applicable and properly named/described to fit to specific db implementations.

E.g. `db.vector.query.top_k` debate will become obvious on postgress - you can't have an attribute like this there and would have to call it `limit`. 

It's ok to limit the scope to several DBs and not cover all possible details.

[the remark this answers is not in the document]
AlexanderWert: > I don't think we should introduce `span.db.vector.client` span

Sorry, I don't get that. We also have a generic [`span.db.client`](https://github.com/open-telemetry/semantic-conventions/blob/41053b507998b6dc4effa22f657cea96ab5f8fe8/model/database/spans.yaml#L111) (which then in addition has technology specific overrides). Wouldn't it be the same thing here, so having a generic description for vector db client spans + in addition having technology-specific overrides for the DBs listed above.

> it's not clear what/when/how will report it

agree, in the readme above we need more explicit / clearer guidance on when to use `span.db.vector.client` vs. the more general `span.db.client`, but we should do it in a technology-independent way first.

> Also, referencing attributes in the specific conventions is a good test if these attributes are applicable and properly named/described to fit to specific db implementations.
> ...
> It's ok to limit the scope to several DBs and not cover all possible details.

Makes sense.

ezimuel:
Thanks @AlexanderWert for the feedback.

model/database/spans.yaml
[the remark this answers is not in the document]
ezimuel: I agree with @AlexanderWert . I think we should have a generic `span.db.vector.client` for vector databases that do not have technology-specific conventions, like the generic sql clients. 
@lmolkova  I see the issue with `db.vector.query.top_k` and `db.query.limit`, I think I can remove the `top-k` even if they are semantically different.

docs/attributes-registry/db.md
```diff
@@ -228,9 +233,38 @@ This group defines attributes for Elasticsearch.
...
+| <a id="db-search-similarity-metric" href="#db-search-similarity-metric">`db.search.similarity_metric`</a> | string | The metric used in similarity search. | `cosine` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+---
 
-**[11] `db.elasticsearch.path_parts`:** Many Elasticsearch url paths allow dynamic values. These SHOULD be recorded in span attributes in the format `db.elasticsearch.path_parts.<key>`, where `<key>` is the url path part name. The implementation SHOULD reference the [elasticsearch schema](https://raw.githubusercontent.com/elastic/elasticsearch-specification/main/output/schema/schema.json) in order to map the path part values to their names.
+`db.search.similarity_metric` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `cosine` | The cosine metric. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
TaoChenOSU, line 254: nit: 
1. There are cosine similarity and cosine distance. Better to distinguish the two in the convention.
2. There are a few common ones that are not listed here: Squared Euclidean and hamming.

docs/attributes-registry/db.md
```diff
@@ -228,9 +233,38 @@ This group defines attributes for Elasticsearch.
...
+This group defines attributes for Search.
+
+| Attribute | Type | Description | Examples | Stability |
+|---|---|---|---|---|
+| <a id="db-search-similarity-metric" href="#db-search-similarity-metric">`db.search.similarity_metric`</a> | string | The metric used in similarity search. | `cosine` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+---
 
-**[11] `db.elasticsearch.path_parts`:** Many Elasticsearch url paths allow dynamic values. These SHOULD be recorded in span attributes in the format `db.elasticsearch.path_parts.<key>`, where `<key>` is the url path part name. The implementation SHOULD reference the [elasticsearch schema](https://raw.githubusercontent.com/elastic/elasticsearch-specification/main/output/schema/schema.json) in order to map the path part values to their names.
+`db.search.similarity_metric` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
TaoChenOSU, line 250: The similarity metric doesn't need to be defined per search request. It can also be defined when the collection is created.

docs/attributes-registry/db.md
```diff
@@ -228,9 +233,38 @@ This group defines attributes for Elasticsearch.
...
+| `euclidean` | The euclidean distance metric. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `manhattan` | The Manhattan distance metric. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+## Db Vector Attributes
+
+This group defines attributes for vector databases.
+
+| Attribute | Type | Description | Examples | Stability |
+|---|---|---|---|---|
+| <a id="db-vector-dimension-count" href="#db-vector-dimension-count">`db.vector.dimension_count`</a> | int | The dimension of the vector. | `3` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
TaoChenOSU, line 265: nit: `db-vector-dimension-count` -> `db-vector-dimension`.

model/database/spans.yaml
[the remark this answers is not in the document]
lmolkova: > Sorry, I don't get that. We also have a generic [span.db.client](https://github.com/open-telemetry/semantic-conventions/blob/41053b507998b6dc4effa22f657cea96ab5f8fe8/model/database/spans.yaml#L111) (which then in addition has technology specific overrides). Wouldn't it be the same thing here, so having a generic description for vector db client spans + in addition having technology-specific overrides for the DBs listed above.

The problem is that none of the databases we have in the semconv should be considered a vector databases then. The  `span.db.vector.client` is an abstract thing nothing is going to use. MongoDB, CosmosDB, elastic, postres - all of them are used as vector databases - will they extend `span.db.vector.client`? Will generic instrumentations for them cover vector-related operations?

Also we're trying to introduce attributes without attempting to apply them to existing databases - this is the major blocker I see in this PR.

docs/database/vector.md
```diff
@@ -0,0 +1,197 @@
...
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`db.operation.name`](/docs/attributes-registry/db.md) | string | The operation  to be performed on the vector database (e.g. build an index/collection) [1] | `build`; `insert`; `search`; `delete` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`http.request.method`](/docs/attributes-registry/http.md) | string | HTTP request method. [2] | `GET`; `POST`; `HEAD` | `Required` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
lmolkova, line 37: `http.request.method` should not be used on generic database span

PR #1233, opened 2024-07-10
Editorial: mention that Redis `db.namespace` is a string

trask:
To avoid what could be a common question.

I don't think a change log is necessary, but happy to add one if anyone wishes.

### DIFF
--- a/docs/database/redis.md
+++ b/docs/database/redis.md
@@ -23,7 +23,7 @@ described on this page.
 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
-| [`db.namespace`](/docs/attributes-registry/db.md) | string | The index of the database being accessed as used in the [`SELECT` command](https://redis.io/commands/select). [1] | `0`; `1`; `15` | `Conditionally Required` If and only if it can be captured reliably. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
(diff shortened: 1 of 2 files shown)


docs/database/redis.md
```diff
@@ -23,7 +23,7 @@ described on this page.
 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
-| [`db.namespace`](/docs/attributes-registry/db.md) | string | The index of the database being accessed as used in the [`SELECT` command](https://redis.io/commands/select). [1] | `0`; `1`; `15` | `Conditionally Required` If and only if it can be captured reliably. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.namespace`](/docs/attributes-registry/db.md) | string | The index of the database being accessed as used in the [`SELECT` command](https://redis.io/commands/select) (captured as a string). [1] | `0`; `1`; `15` | `Conditionally Required` If and only if it can be captured reliably. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
heyams, line 26: second column has String type. does it need to mention "captured as a string" again? 
db.redis.database_index was an int, now it's replaced with db.namespace as a string. as long as we put something in the comments of the migration PR for `db.redis.database_inde`, i am good with it.

lmolkova (approved):

docs/database/redis.md
joaopgrassi: Agreed, the string type should be clear already, but I also don't mind having the extra hint.

joaopgrassi (approved):

PR #1234, opened 2024-07-10
Move http migration guide to supplementary-guidelines directory

trask:
My understanding is this is a better (non-normative) place for the migration guide to live.


lmolkova (approved):

joaopgrassi (approved):

joaopgrassi:
Hum now thinking, will this impact the page on the website? 🤔 https://opentelemetry.io/docs/specs/semconv/http/migration-guide/

CC @chalin

lmolkova:
Great catch!

Perhaps we can use the same trick as here https://github.com/open-telemetry/opentelemetry-specification/pull/3956/files?

```
linkTitle: Metric Requirement Levels
redirect: /docs/specs/semconv/general/metric-requirement-level/ 301!
--->
```

supplementary-guidelines/http-migration-guide.md
chalin, line 1: I'd suggest renaming the file to `http-migration.md` given that it is under `...-guidelines`. Just a thought.

chalin (commented):
Depending on your vote for or against #1240, we'll need to carefully choose which file to add the redirect rule to so that it actually gets picked up by Hugo.

chalin:
Btw, I'm OOO next week so if this doesn't get resolved this week, I'll get to offering feedback here when I'm back.

joaopgrassi (changes requested):
Blocking to avoid merging until we decide what to do with the website

Issue #1159, opened 2024-06-14
Database: add `db.query.text` as opt-in on metrics


lmolkova:
In some applications `db.query.text` (templatized, sanitized) could have low enough cardinality to be reported on metrics where it can be quite useful.
We should allow users to explicitly enable it.

Update: would we also consider using query text on span name?

PR #1243, opened 2024-07-12
db.query.text IN-clauses MAY be collapsed during the sanitization process

trask:
Addresses https://github.com/open-telemetry/semantic-conventions/issues/1053#issuecomment-2183051443

`db.query.text` IN-clauses MAY be collapsed during the sanitization process.


docs/database/database-spans.md
```diff
@@ -250,6 +250,10 @@ in which case the instrumentation MAY choose a different placeholder.
 
 Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id = $1` can be captured as is.
 
+[IN-clauses](https://en.wikipedia.org/wiki/Where_(SQL)#IN) MAY be collapsed during sanitization,
+e.g. from `IN (?, ?, ?, ?)` to `IN (?)`, as this can help with extremely long IN-clauses,
+and can help control cardinality for users who choose to (optionally) add `db.query.text` to their metric attributes.
```
lmolkova, line 255: Not quite related to this PR:

Remember we've got some push-back on span name changes for HTTP when we added `url.template`? 

If we consider that such opt-in for `db.query.text` would be also useful on the span name, we should try to do the #1159 now (pre-stability) to avoid friction in the future.

Updated #1159 and moved it to Todo to consider it

lmolkova (approved):

docs/database/database-spans.md
trask: I think the pushback was about adding `url.template` to span name by default, but in this case it would be opt-in?

lmolkova: I mean that span name is part of general stability guarantees. and yes, if it's opt-in then it'd not be a breaking change.

Still, documenting in advance would be nice, since it would make it clear for backends that there are alternative DB span name formats.

joaopgrassi (approved):

Issue #1164, opened 2024-06-19
Add k8s.volume.type attribute


ChrsMark:
### Area(s)

area:k8s

### Is your change request related to a problem? Please describe.

`kubeletstats` receiver provides support for adding the `k8s.volume.type` attribute along with the volume related metrics: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.102.0/receiver/kubeletstatsreceiver/internal/kubelet/conventions.go#L7.

### Describe the solution you'd like

We need to define a Semantic Convention for this attribute.

### Additional context

Related to https://github.com/open-telemetry/semantic-conventions/issues/1119 and https://github.com/open-telemetry/semantic-conventions/issues/1032.

/cc @open-telemetry/semconv-k8s-approvers

TylerHelmuth:
This attribute has been in use for a long time successfully, officially merging it here makes sense to me

joaopgrassi:
@ChrsMark will you take this? Should I move to the system semconv project board?

ChrsMark:
@joaopgrassi I can assign this to me and work on it in the near future unless there is anybody else willing to pick it up earlier.

This does not fall into the system semconv project. The System Metrics WG only focuses on host/system metrics and attributes.
This can be tracked by https://github.com/open-telemetry/semantic-conventions/issues/1032 until we form a K8s WG, hopefully some time soon :).

PR #1251, opened 2024-07-15
Add k8s.volume.{name,type} attributes

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1164

## Changes

Add k8s.volume.{name,type} attributes.

From https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.102.0/receiver/kubeletstatsreceiver/internal/kubelet/conventions.go#L7

### DIFF
--- a/model/registry/k8s.yaml
+++ b/model/registry/k8s.yaml
@@ -177,13 +177,13 @@ groups:
         brief: >
           The name of the CronJob.
         examples: ['opentelemetry']
-      - id: volume.name
+      - id: k8s.volume.name
         type: string
         stability: experimental
         brief: >
           The name of the K8s volume.
         examples: [ 'volume0' ]
-      - id: volume.type
+      - id: k8s.volume.type
         stability: experimental
         brief: >
           The type of the K8s volume.
--- a/model/registry/k8s.yaml
+++ b/model/registry/k8s.yaml
@@ -177,3 +177,53 @@ groups:
         brief: >
           The name of the CronJob.
         examples: ['opentelemetry']
+      - id: volume.name
+        type: string
+        stability: experimental
+        brief: >
+          The name of the K8s volume.
+        examples: [ 'volume0' ]
+      - id: volume.type
+        stability: experimental
+        brief: >
+          The type of the K8s volume.
+        examples: [ "emptyDir", "persistentVolumeClaim" ]
+        type:
+          allow_custom_values: true
+          members:
+            - id: persistentvolumeclaim
+              value: 'persistentVolumeClaim'
+              brief: "A [persistentVolumeClaim](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#persistentvolumeclaim) volume"
+              stability: experimental
+            - id: configmap
+              value: 'configMap'
+              brief: "A [configMap](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#configmap) volume"
+              stability: experimental
+            - id: downwardapi
+              value: 'downwardAPI'
+              brief: "A [downwardAPI](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#downwardapi) volume"
+              stability: experimental
+            - id: emptydir
+              value: 'emptyDir'
+              brief: "An [emptyDir](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#emptydir) volume"
+              stability: experimental
+            - id: secret
+              value: 'secret'
+              brief: "A [secret](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#secret) volume"
+              stability: experimental
+            - id: local
+              value: 'local'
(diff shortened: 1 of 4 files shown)


model/resource/k8s.yaml
```diff
@@ -100,3 +100,12 @@ groups:
     attributes:
       - ref: k8s.cronjob.uid
       - ref: k8s.cronjob.name
+
+  - id: k8s.volume
```
jsuereth, line 104: Let's avoid adding anything to resource until Entities SiG makes more progress.

I'd love to do more here, but right now Identifying + Descriptive attributes are blended in Resource and this is problematic for our long term success.

Let's remove the resource part, but we can reserve the other attributes with the meaning you suggest for general usage in OTEL.

jsuereth (changes requested):

model/resource/k8s.yaml
TylerHelmuth: @jsuereth does this mean the semantic conventions are not accepting any more resource attribute proposals until Entities is stabilized?

TylerHelmuth: @jsuereth based on the discussion around resource attributes during today's SIG meeting, I believe at least `k8s.volume.name` meets the criteria of "not changing over the lifetime of the object".  I believe that `k8s.volume.type` also meets the criteria, but @ChrsMark can you confirm that a volume's type cant change without recreating?

ChrsMark: From https://kubernetes.io/docs/concepts/workloads/pods/#pod-update-and-replacement:

> Pod updates may not change fields other than spec.containers[*].image, spec.initContainers[*].image, spec.activeDeadlineSeconds or spec.tolerations. For spec.tolerations, you can only add new entries.

However thinking of it again, my concern is if we should make these resource attributes or metric attributes.
In `kubeletstats` we are going to use them as metric attributes right (for the `k8s.volume.*` metrics) ?

It seems we are in need of metric attributes here. So it sounds correct if we only keep them in the registry for now with a purpose similar to the [`system.device`](https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/docs/attributes-registry/system.md?plain=1#L24).

ChrsMark:
@open-telemetry/semconv-k8s-approvers could you take a look?

TylerHelmuth (approved):

ChrsMark:
Thank's @lmolkova for the comments! Should be addressed now.

lmolkova (approved):

model/registry/k8s.yaml
```diff
@@ -177,3 +177,48 @@ groups:
...
+              brief: "A [local](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#local) volume"
+              stability: experimental
+            - id: gce_persistent_disk
+              value: 'gcePersistentDisk'
+              brief: "A [gcePersistentDisk](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#gcepersistentdisk) volume"
+              stability: experimental
+            - id: glusterfs
+              value: 'glusterfs'
+              brief: "A [glusterfs](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#glusterfs) volume"
+              stability: experimental
```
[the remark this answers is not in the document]
ChrsMark: done: [c511e27](https://github.com/open-telemetry/semantic-conventions/pull/1251/commits/c511e279344c506cb44979d887cc34cc4cab54e2)

model/registry/k8s.yaml
```diff
@@ -177,3 +177,48 @@ groups:
...
+              value: 'secret'
+              brief: "A [secret](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#secret) volume"
+              stability: experimental
+            - id: local
+              value: 'local'
+              brief: "A [local](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#local) volume"
+              stability: experimental
+            - id: gce_persistent_disk
+              value: 'gcePersistentDisk'
+              brief: "A [gcePersistentDisk](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#gcepersistentdisk) volume"
```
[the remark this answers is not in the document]
ChrsMark: done: [c511e27](https://github.com/open-telemetry/semantic-conventions/pull/1251/commits/c511e279344c506cb44979d887cc34cc4cab54e2)

ChrsMark:
Hey @jsuereth, could you take another looks please?

jsuereth (approved):

model/resource/k8s.yaml
jsuereth: See: https://github.com/open-telemetry/oteps/pull/264 for the path forward I'm proposing here.  Once we've agreed to what we're doing, we can unlock moving these attributes into Resource/Entity sections in Semconv.

ChrsMark: Thank's for the head's up!

ChrsMark:
The failure is unrelated. I have filed https://github.com/open-telemetry/semantic-conventions/pull/1332 to fix this.

PR #1259, opened 2024-07-16
add `nodejs.eventloop.time` metric

maryliag:
## Changes

Adding `nodejs.eventloop.time` metric to a counter.
Feedback came during prototype of the nodejs runtime metrics [here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2136/files#r1678358022).


maryliag:

docs/runtime/nodejs-metrics.md
```diff
@@ -313,10 +313,10 @@ This metric is [recommended][MetricRecommended].
 
 | Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
 | -------- | --------------- | ----------- | -------------- | --------- |
-| `nodejs.eventloop.utilization` | Gauge | `1` | Event loop utilization. [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `nodejs.eventloop.utilization` | Counter | `s` | Event loop utilization. [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
ben-childs-docusign, line 316: nodejs perf_hooks expose 2 cumulative counters to measure utilization:
https://nodejs.org/api/perf_hooks.html#performanceeventlooputilizationutilization1-utilization2

active seconds and idle seconds

the utilization value itself seems to somehow be computed based on the difference between two values and it is not clear if that could be represented as a counter directly.

my suggestion would be to add 
```
nodejs.eventloop.active.time
nodejs.eventloop.idle.time
```
and then keep utilization as a gauge if we think it is useful.

maryliag: Created a new metric for time, with the state as an attribute. Kept the utilization as is.

model/metrics/nodejs-metrics.yaml
```diff
@@ -73,9 +73,23 @@ groups:
     type: metric
     metric_name: nodejs.eventloop.utilization
     brief: "Event loop utilization."
+    instrument: gauge
+    unit: "1"
+    stability: experimental
+    note: >
+      The value range is [0.0,1.0] and can be retrieved from `utilization` from
```
lmolkova, line 80: nit 
```suggestion
      The value range is [0.0, 1.0] and can be retrieved from
```

model/metrics/nodejs-metrics.yaml
```diff
@@ -73,9 +73,23 @@ groups:
...
     instrument: counter
     unit: "s"
     stability: experimental
+    attributes:
+      - ref: nodejs.eventloop.type
+        requirement_level: required
     note: >
-      Value can be retrieved from
-      [`performance.eventLoopUtilization([utilization1[, utilization2]])`](https://nodejs.org/api/perf_hooks.html#performanceeventlooputilizationutilization1-utilization2)
+      Value can be retrieved from `idle` and `active` from
```
lmolkova, line 94: nit:

> Value can be retrieved from `idle` and `active` from
[...]()

this sentence seems to be broken

model/metrics/nodejs-metrics.yaml
```diff
@@ -73,9 +73,23 @@ groups:
...
+      [`performance.eventLoopUtilization([utilization1[, utilization2]])`](https://nodejs.org/api/perf_hooks.html#performanceeventlooputilizationutilization1-utilization2)
```
lmolkova:
```suggestion
      [`performance.eventLoopUtilization([utilization1[, utilization2]])`](https://nodejs.org/api/perf_hooks.html#performanceeventlooputilizationutilization1-utilization2)
```

model/metrics/nodejs-metrics.yaml
```diff
@@ -73,9 +73,23 @@ groups:
...
+
+  - id: metric.nodejs.eventloop.time
+    type: metric
+    metric_name: nodejs.eventloop.time
+    brief: "Cumulative duration of time the event loop has been on the state."
     instrument: counter
     unit: "s"
     stability: experimental
+    attributes:
+      - ref: nodejs.eventloop.type
```
lmolkova, line 91: I don't feeling strong about it, but please consider `nodejs.eventloop.state` which seems to work better with `active` and `idle` values. I'd expect something else under type (like specific implementation name).

model/registry/nodejs.yaml
```diff
@@ -0,0 +1,20 @@
...
+        type:
+          members:
+            - id: active
+              value: 'active'
+              brief: 'Active time.'
+              stability: experimental
+            - id: idle
+              value: 'idle'
+              brief: 'Idle time.'
+              stability: experimental
```
lmolkova:
```suggestion
              stability: experimental

```

model/metrics/nodejs-metrics.yaml
david-luna: +1 to the naming change. Looking at some refs within semconv I see the concept of `mode` for [cpu time](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemcputime) or `state` for [memory usage](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemcputime)

maryliag: Updated

model/metrics/nodejs-metrics.yaml
```diff
@@ -77,5 +77,19 @@ groups:
...
     stability: experimental
     note: >
-      The value range is [0.0,1.0] and can be retrieved from value
+      The value range is [0.0, 1.0] and can be retrieved from
+      [`performance.eventLoopUtilization([utilization1[, utilization2]])`](https://nodejs.org/api/perf_hooks.html#performanceeventlooputilizationutilization1-utilization2)
+
+  - id: metric.nodejs.eventloop.time
+    type: metric
+    metric_name: nodejs.eventloop.time
+    brief: "Cumulative duration of time the event loop has been on the state."
```
lmolkova, line 86: maybe

```suggestion
    brief: "Cumulative duration of time the event loop has been in each state."
```

?

lmolkova (approved):

jsuereth (approved):

model/registry/nodejs.yaml
```diff
@@ -0,0 +1,20 @@
+groups:
+  - id: registry.nodejs
+    type: attribute_group
+    brief: Describes Node.js related attributes.
+    display_name: Node.js Attributes
+    prefix: nodejs
```
lmolkova, line 6: we've got rid of prefix while you were working on this PR, so this should be removed and added to the attribute id

```suggestion
```

model/registry/nodejs.yaml
```diff
@@ -0,0 +1,20 @@
...
+    attributes:
+      - id: eventloop.state
```
lmolkova, line 8:
```suggestion
      - id: nodejs.eventloop.state
```

PR #1263, opened 2024-07-16
[chore] Add clarification on event attributes vs body fields

JamieDanielson:
## Changes

* Add clarification describing when to use attributes and when to use body fields for an event
  * This was discussed briefly during SemConv SIG meeting, with the goal of helping clarify the structure of events.
* Rename data to body 
  * The name of the `body` had gone through several iterations before deciding on the final term `body`. I'm not sure whether the word `payload` is still necessary here or if it's a relic of the previous term, so I've left it but could see renaming these to just `body` from payload (data). This can also be a separate PR if further discussion is warranted.

I cannot tell if this would require a changelog, happy to add if deemed appropriate.


JamieDanielson:
@open-telemetry/semconv-event-approvers please take a look!

joaopgrassi (approved):

trask (approved):

scheler (approved):

PR #1267, opened 2024-07-20
[chore] update scripts to detect macos or linux for GNU sed usage

adrielp:
## Changes

Adding some basic OS detection in scripts and `make` due to the difference between `sed` on MacOs and Linux. Namely, `gsed` (GNU Sed) will be used on MacOs. The contributing guide documents the install under pre-reqs and gives the `brew bundle` command as an optional way to install `gsed` on Mac. This is simply a quality of life improvement for those on Mac.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,7 +85,6 @@ environment configured:
 
   ```bash
   brew bundle
-
   ```
 
 ### 1. Modify the YAML model
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,6 +80,14 @@ environment configured:
   npm install
   ```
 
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -80,6 +80,14 @@ environment configured:
   npm install
   ```
 
+- If on MacOs, ensure you have `gsed` (GNU Sed) installed. If you have [HomeBrew](https://brew.sh)
+  installed, then you can run the following command to install GSED.
+
+  ```bash
+  brew bundle
+
```
joaopgrassi, line 88:
```suggestion
```

joaopgrassi (approved):
Thanks for improving this!

PR #1270, opened 2024-07-23
[chore] Fix outdated registry markdown

joaopgrassi:
Somehow https://github.com/open-telemetry/semantic-conventions/pull/1216 was merged even with the registry check failing 🤔 

This fixes it.


trisch-me (approved):
this is because of auto merge most probably

PR #1281, opened 2024-07-25
Add attributes request/response body size for rpc

crossoverJie:
Ref: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/11833#discussion_r1689747737

## Changes
Add attributes:
- rpc.client.request.body.size
- rpc.client.response.body.size
- rpc.server.request.body.size
- rpc.server.response.body.size


crossoverJie:
@trask @lmolkova Please take a look.

model/registry/rpc.yaml
```diff
@@ -264,3 +264,23 @@ groups:
         type: int
         stability: experimental
         brief: "Uncompressed size of the message in bytes."
+      - id: client.request.body.size
```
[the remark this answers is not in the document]
crossoverJie: Thank for your suggestion. Done.

model/registry/rpc.yaml
```diff
@@ -264,3 +264,23 @@ groups:
...
+        type: int
+        stability: experimental
+        brief: "The size of the client request payload body in bytes."
+        examples: 3495
+      - id: client.response.body.size
```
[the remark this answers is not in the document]
crossoverJie: I have added some details, please take a look again.

model/registry/rpc.yaml
```diff
@@ -264,23 +264,13 @@ groups:
         type: int
         stability: experimental
         brief: "Uncompressed size of the message in bytes."
-      - id: client.request.body.size
+      - id: request.body.size
         type: int
         stability: experimental
-        brief: "The size of the client request payload body in bytes."
+        brief: "The size of the request payload body in bytes. For gRPC, this value is calculated through protobuf (including unary and streaming calls)."
```
lmolkova, line 270: you can add grpc-specific note on grpc only by referencing that attribute and updating the brief in 

https://github.com/open-telemetry/semantic-conventions/blob/d182f9e66f48d854e8afcc74a0962bad89f9dd3c/model/trace/rpc.yaml#L61-74

Also, please provide the description of what the size is (not how it's calculated). If there is some protobuf documentation that's useful to link, let's link it.

crossoverJie: I add some information but only found link related to Java, I cound't find a description of the `body.size` on the protobuf [official website](https://protobuf.dev/).

model/registry/rpc.yaml
```diff
@@ -262,3 +262,19 @@ groups:
         type: int
         stability: experimental
         brief: "Uncompressed size of the message in bytes."
+      - id: rpc.request.body.size
+        type: int
+        stability: experimental
+        brief: "The size of the request payload body in bytes."
+        note: >
+          For gRPC, this value is calculated through [protobuf](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/MessageLite.html#getSerializedSize--)
```
lmolkova, line 270: Thank you for providing the link to the docs!

If I understand correctly, this is a size of the message.

I.e. for unary calls, that's the request size, but for streaming calls, it'll be the size of the individual message and it could be useful on events describing individual streaming messages.

It would be misleading to use request/response size attributes for streaming calls on such message events.  

So I think we have a couple of options:

1. Try using something like `rpc.message.body.size` - it'll look good on events, but won't work for rpc spans (we need to report request and response sizes)
2. Keeping `rpc.request|response.body.size` for the request/response part only.
3. Do both (1+2)

It seems we're doing option 2 which is fine with me, we just need to clarify it further - I'll leave some suggestions.

lmolkova: maybe:

```suggestion
          The request body size represents the serialized size of the message payload that is sent over the wire. In case of streaming calls, it accounts for the original message in the request and does not include size of the consequent message stream.
```

model/registry/rpc.yaml
```diff
@@ -262,3 +262,19 @@ groups:
...
+          (including unary and streaming calls).
+        examples: 3495
+      - id: rpc.response.body.size
+        type: int
+        stability: experimental
+        brief: "The size of the response payload body in bytes."
+        note: >
+          For gRPC, this value is calculated through [protobuf](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/MessageLite.html#getSerializedSize--)
```
lmolkova, line 278:
```suggestion
          The response body size represents the serialized size of the message payload that is received in the response over the wire. In case of streaming calls, it accounts for the first response payload and does not include size of the consequent message stream.
```

model/trace/rpc.yaml
```diff
@@ -71,6 +79,12 @@ groups:
       - ref: rpc.grpc.response.metadata
         tag: grpc-tech-specific
         requirement_level: opt_in
+      - ref: rpc.request.body.size
+        tag: grpc-tech-specific
+        requirement_level: recommended
+      - ref: rpc.response.body.size
+        tag: grpc-tech-specific
+        requirement_level: recommended
```
lmolkova, line 87:
```suggestion
        requirement_level: recommended
        note: >
          The response body size represents the serialized size of the message payload that is received in the response over the wire. In case of streaming calls, it accounts for the first response payload and does not include size of the consequent message stream.        
           This value can usually be obtained from protobuf API such as [AbstractMessage#getSerializedSize](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/AbstractMessage.html#getSerializedSize--).
```

model/trace/rpc.yaml
```diff
@@ -71,6 +79,12 @@ groups:
       - ref: rpc.grpc.response.metadata
         tag: grpc-tech-specific
         requirement_level: opt_in
+      - ref: rpc.request.body.size
+        tag: grpc-tech-specific
+        requirement_level: recommended
```
lmolkova, line 84:
```suggestion
        requirement_level: recommended
        note: >
            The request body size represents the serialized size of the message payload that is sent over the wire. In case of streaming calls, it accounts for the original message in the request and does not include size of the consequent message stream.
             This value can usually be obtained from protobuf API such as [AbstractMessage#getSerializedSize](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/AbstractMessage.html#getSerializedSize--).
```

model/registry/rpc.yaml
crossoverJie: Thank you very much for your addition👍.

crossoverJie:
Any updates?

docs/rpc/grpc.md
```diff
@@ -26,12 +26,18 @@ Below is a table of attributes that SHOULD be included on client and server gRPC
...
+| [`rpc.grpc.response.metadata.<key>`](/docs/attributes-registry/rpc.md) | string[] | gRPC response metadata, `<key>` being the normalized gRPC Metadata key (lowercase), the value being the metadata values. [4] | `rpc.grpc.response.metadata.my-custom-metadata-attribute=["attribute_value"]` | `Opt-In` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
-**[1]:** Instrumentations SHOULD require an explicit configuration of which metadata values are to be captured. Including all request metadata values can be a security risk - explicit configuration helps avoid leaking sensitive information.
+**[1]:** The request body size represents the serialized size of the message payload that is sent over the wire. In case of streaming calls, it accounts for the original message in the request and does not include size of the consequent message stream. This value can usually be obtained from protobuf API such as [AbstractMessage#getSerializedSize](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/AbstractMessage.html#getSerializedSize--).
```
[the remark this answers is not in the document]
lmolkova: Assuming we record metrics, we'd need to record direction as a part of the metric name - something like `rpc.sent.message.size`, `rpc.received.message.size` (histograms). 

We could use similar naming (`rpc.sent.message.size`, `rpc.received.message.size` ) for these attributes and replace existing ones (`rpc.message.*_size`) populated on the [events](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/rpc-spans.md#events).

Unary calls then could be recorded without events - there is nothing(?) useful on the unary messages besides size anyway (direction is now part of the name and index is always 0). Having everything on one span would be easy and convenient for the simple case.

Streaming calls would not have size attributes on spans - they would always be reported on events to avoid potential ambiguity/confusion.

Would this work? @jsuereth @crossoverJie @yashykt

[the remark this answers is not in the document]
yashykt: The wording here that says "message payload that is sent over the wire" is kinda confusing because after a payload is serialized, we could compress it and encrypt it as well. The part where it says "In case of streaming calls, it accounts for the original message in the request and does not include size of the consequent message stream." also seems like a big restriction.

I'm slightly confused whether this is meant for metrics or tracing, but I'll assume that this is for tracing. Refer to https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md, for the gRPC team's design. Essentially, we decided to always use events for recording message sizes irrespective of whether we are using unary or streaming. As a side-benefit, this also allows us to gain some insight on when a message was compressed/decompressed and how much time that took.

PR #1289, opened 2024-07-26
clarify pool name being unique

maryliag:
## Changes

Working on the prototype for DB Metrics, the recommendation for using just `host:port` as a fallback for pool name was not enough to identify a pool, so this PR adds example of how to make the name more unique.


maryliag:
cc @open-telemetry/semconv-db-approvers

model/registry/db.yaml
```diff
@@ -335,8 +335,9 @@ groups:
         brief: >
           The name of the connection pool; unique within the instrumented application.
           In case the connection pool implementation doesn't provide a name,
-          instrumentation should use a combination of `server.address` and `server.port` attributes
-          formatted as `server.address:server.port`.
+          instrumentation should use a combination of parameters that would make the name
+          unique, for example, combining attributes `server.address`, `server.port`, `username`
+          and `database`, formatted as `username@server.address:server.port/database`.
```
lmolkova, line 340: Capturing username could be problematic - we redact it on URLs and used to redact in database connection string (when it was recoded). We also don't currently have an attribute for it.

```suggestion
          instrumentation SHOULD use a combination of parameters that would make the name
          unique, for example, combining attributes `server.address`, `server.port`, 
          and `db.namespace`, formatted as `server.address:server.port/db.namespace`.
          
          Instrumentations that generate connection pool name following different patterns
          SHOULD document it.
```

joaopgrassi (approved):

PR #1290, opened 2024-07-26
Db metrics pending requests

maryliag:
## Changes

During the prototype for the DB Metrics, found that is not always possible to have a cumulative value. Having just a current value makes more sense. Existing implementations of this metric are already using the current or are using a wrong implementation of the cumulative.


maryliag:
cc @open-telemetry/semconv-db-approvers

lmolkova (approved):

.chloggen/1290.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
```
gregkalapos, line 1: Isn't this technically a breaking change from SemConv's point of view?
[Docs](https://opentelemetry.io/docs/specs/otel/versioning-and-stability/#semantic-conventions-stability) say:
> Semantic Conventions defines breaking changes as those that would break the common usage of tooling written against the telemetry it produces. That is, the portions of telemetry where specialized tooling (alerts, dashboards, e.g.) interact are expected to remain stable ...

Assuming there was a previous correct implementation (which I understand is probably a wrong assumption in reality :) - but from SemConv's point of view, I think it's fair to assume correct implementations) which changes from cumulative to current - then dashboards and alerts need to be adapted.

This is an experimental metric and based on the issue description, the proposed change totally makes sense - cumulative seems to be a bit unrealistic.

So overall, I think this makes sense and I'd assume it's ok to have a breaking change here.

lmolkova: I think "cumulative for the entire pool" meant not that it's cumulative counter, but it's a sum across the pool (but there is a pool attributes...)?

So @maryliag and me thought it's just confusing - it makes no sense to report cumulative counter (especially as UpDown one) and all instrumentations we checked reported the number of current requests.

So this change does not seem to be changing anything, but removing the confusing part.

gregkalapos: I see - indeed confusing :)

Thanks for clarifying.

gregkalapos (approved):

trask (approved):

PR #1291, opened 2024-07-26
Change `gen_ai.responmse.finish_reasons` type from array to `string`, use it on metrics

### LINKED ISSUE #1277: GenAI: do we need to support multiple finish reasons?
See https://github.com/open-telemetry/semantic-conventions/pull/980#discussion_r1586695157.

Context:
- some models return multiple choices in one response
- each choice has a finish reason that explains why model stopped generating content which contains values like `stop`, `length`, `filtered`... - this could be very useful to know, query and aggregate by

Having array attribute is problematic since it's harder to query and not really possible to use on metrics or per-choice events.

Multiple choices are supported by a limited set of models. Event when multiple choices are supported some SDKs (e.g. openai-dotnet) make a choice not to expose it on the convenience API level to simplify the design and provide much more friendly experience.
Most of examples and documentation assumes there is just one choice.

Given this, it seems that in most of the cases there will be just one choice and just one finish reason on each span.

The proposal is to 
- change type from string array to string
- rename the attribute to `gen_ai.response.finish_reason` 
- report one value (either one of the following)
  1. last reason
  2. comma-separated reasons like `stop,length`
- we can reuse this attribute on events in case we want to promote corresponding body field to attributes

lmolkova:
Fixes #1277

## Changes

While OpenAI supports returning multiple completions (and therefore could have multiple finish reasons), I did not find models outside of OpenAI that return multiple choices and many SDKs (e.g. [vercel](https://sdk.vercel.ai/docs/reference/ai-sdk-core/generate-text#returns)) don't support returning multiple choices.

So effectively there is usually one finish reason. 
Having an array for it:
- complicates querying
- makes usage of finish reason on metrics more complicated

This PR:
- turns finish reasons attribute into a string (comma-separated if there is more than one)
- adds it to metrics. 

Having comma-separated list has high-ish theoretical cardinality in edge cases, but results in best experience for common case.
I'm suggesting to start with it and consider making it opt-in on metrics if it's proven to be problematic.


docs/gen-ai/gen-ai-metrics.md
```diff
@@ -266,8 +274,9 @@ This metric SHOULD be specified with [ExplicitBucketBoundaries] of
 | [`gen_ai.system`](/docs/attributes-registry/gen-ai.md) | string | The Generative AI product as identified by the client or server instrumentation. [2] | `openai` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [3] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` if the operation ended in an error | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 | [`server.port`](/docs/attributes-registry/server.md) | int | Server port number. [4] | `80`; `8080`; `443` | `Conditionally Required` If `sever.address` is set. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.response.finish_reason`](/docs/attributes-registry/gen-ai.md) | string | The reason(s) the model stopped generating tokens. [5] | `stop`; `stop,length` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 277: @achandrasekar any thoughts/objections? on adding finish_reason attribute to LLM server metrics?

achandrasekar: This looks good to me. Adding a finish_reason attribute makes sense. It would be useful in a lot of cases including benchmarking where we can see if we generated to the max_length or stopped early due to the EOS token.

Issue #1264, opened 2024-07-18
Embed functionality v0.2


trisch-me:
After several discussions in comments and during our semconv and tooling meetings we came to conclusion to rewrite embed functionality to make embed more prominent and visible because it has potential to influence and change not only the schema but also other parts of Open Telemetry such as APIs etc.

So the proposal is to make an embed a part of the group as opposed to the part of attributes before. Both embedded namespaces and embed attributes will be visible in separate block, so the developer would know that those are treated differently

``` yaml
groups:
   - id: registry.client
     prefix: client
     type: attribute_group
     brief: "Describes client"

     embed:
     - attribute_groups: 
        - ref: geo
          as: location (if omitted, the name is equal as namespace prefix, in this example geo)
          requirement_level: Recommended
          brief: "My own embedded geo space"

          attributes: 
          - ref: geo.lat (override only those attributes from the group which needs to be overridden. 
                             Other attributes are added AS IS)
            requirement_level: Required
            brief: "This attribute is required"


      - attributes: 
         - ref: geo.lat
           as: my.latitude
           requirement_level: Required
           brief: "My own embedded latitude"
         - ref: geo.lon
```

### Embedded namespaces

- Embedded namespaces will be listed under `attribute_groups `. They might have an optional parameter `as`. If omitted it is equal to the embedded namespace, for example resolved schema will be `client.geo.*`. If provided, the name of the embedded namespace will be equal to the value of `as`, such as `client.location.*` where `*` has all `geo` attributes.
- Embedded namespace might have it's own requirements level and brief, if requirement level omitted - default value is `Recommended`
- If there is a need to override specific attributes within embedded namespace, it will be possible to do so by listing those attributes under `attributes` block within `attribute_groups`. The `ref` attribute is mandatory. The `requirement_level` and `brief` are optional. If omitted, the original values will be used. In example above all attributes from `geo` will be optional, but `geo.lat` will be required and have a different brief.
- Example of embedded namespaces where `as` is needed, resolved schema will be `process.parent.*`:
```yaml
groups:
   - id: registry.process
     prefix: process
     type: attribute_group
     brief: "Describes process"

     embed:
     - attribute_groups: 
        - ref: process
          as: parent
          requirement_level: Required
          brief: "Parent of the process is itself a process namespace"

          attributes: 
          - ref: process.id
            requirement_level: Required
            brief: "This attribute is required, all other process attributes are optional"

```

### Embedded singular attributes

[...]

PR #1293, opened 2024-07-26
Remove prefix in yaml, add policy check to block future usages (and minor cleanups)

### LINKED ISSUE #1292: Stop using prefix to define attributes
Attributes can be defined in two equivalent ways.

E.g. `foo.bar` can be defined with prefix `foo` and id `bar`

```yaml
group:
- id: foo
  prefix: foo
  attributes: 
  - id: bar
```

or

without prefix and id `foo.bar`

```yaml
group:
- id: foo
  attributes: 
  - id: foo.bar
```

Using prefix is inconvenient:
- when searching for attribute definition in yaml files
- when reading yaml definitons, it's not obvious what the full attribute name would be generated
- tooling needs to resolve full attribute name and we need to be aware of it in some cases when writing policy checks, writing jq filters, generating code

Removing prefix would keep all MD files intact, but would resolve these inconveniences without any obvious downsides.

lmolkova:
## Changes

Related to #1292 - this PR does not block anyone from using prefix, only changes the current pattern.

- Changes all attribute definitions to use fully qualified attribute names and removes all prefixes
- Add rego policy to detect and fail when prefix is used.
- Removes (ignored) `allow_custom_values` on enums


LikeTheSalad (approved):

joaopgrassi (approved):

lmolkova:
@trisch-me great point. This would be a blocker to #1292 and tooling changes https://github.com/open-telemetry/weaver/issues/263. 

I still want to proceed with this PR since it does not stop someone from using prefix now (when necessary) or tooling from introducing any changes that can still leverage the prefix.

trisch-me:
actually it does break it - as soon as prefix will be added because someone needs it the resolved registry with current weaver will double the prefix, so it will be `prefix.prefix.attribute`. I think this PR should go along together with weaver PR to NOT include prefix into resolved name anymore

trisch-me:
>  whoever adds it, would remove it from the attribute id to not break things 

but then it is inconsistent. If we have decided to be explicit and provide prefix always in the id we need to update instrumentation as well so in case prefix would be there it will not break. Therefore I propose to merge this PR only together with/after weaver PR. 

If you want I can create a PR in weaver which does this job but I'm against merging this PR until tooling is ready.

lmolkova:
> but then it is inconsistent. If we have decided to be explicit and provide prefix always in the id we need to update instrumentation as well so in case prefix would be there it will not break.

it's already inconsistent. Check out deprecated attributes in the registry. All the tooling is agnostic to where the prefix is today. Instrumentations use fully qualified name (prefix + id). 

I'm ok waiting for the tooling changes. I'm trying to explain that this PR does not change status quo.

trask (approved):
this seems like a good incremental progress to me

trisch-me:
> Instrumentations use fully qualified name (prefix + id) 
this is where I see things are breaking. Because if someone adds a prefix back because it's needed for other functionality, all id's in that group will be wrong or should be updated again to not have prefix in them. So better to remove dependency on prefix for name resolution

jsuereth (approved):

adrielp (approved):

lmolkova:
discussed at the tooling wg call 
- it's important to make a consistent choice and, if the decision is to stop using prefix, to stop using it.
- we can prohibit using prefix with rego policy - will add one in this PR
- schema and tooling updates can happen after migration to weaver.

@trisch-me wanted to double-check if you're ok with it.
The embed design does not seem depend on it anymore (#1264), but If it we end up needing it, we can always roll back this PR (partially or fully) - this is another benefit of enforcing it in the policy for now and not (yet) in the tooling.

trisch-me (approved):
thank you @lmolkova 
With concrete goal to deprecate prefix I approve this PR

PR #1312, opened 2024-08-02
Clarification to db.namespace when a query explicitly targets a specific database

gregkalapos:
## Changes

From today's DB Stability WG meeting.

We also discussed that the spec should say that instrumentation should best effort keep track of the currently selected database. I feel this is already very well covered by the current text above my new line, so I did not add anything to this. This is what we have and I think this is captures what we meant in the meeting:
>   The database name can usually be obtained with database driver API such as  [JDBC `Connection.getCatalog()`](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#getCatalog--) or [.NET `SqlConnection.Database`](https://learn.microsoft.com/dotnet/api/system.data.sqlclient.sqlconnection.database).

> Some database drivers don't detect when the current database is changed (for example, with SQL `USE database` statement). Instrumentations that parse SQL statements MAY use the database name provided in the connection string and keep track of the currently selected database name.

### DIFF
--- a/model/trace/database.yaml
+++ b/model/trace/database.yaml
@@ -331,6 +331,9 @@ groups:
 
           For commands that switch the database, this SHOULD be set to the target database (even if the command fails).
 
+          In case a specific query explicitly targets a database different from the one used by the connection,
+          the db.namespace should still be the one from the connection.
+
           If instrumentation cannot reliably determine the current database name, it SHOULD NOT set `db.namespace`.
   - id: db.cosmosdb
     type: span
--- a/.chloggen/dbnamespace_clarification.yaml
+++ b/.chloggen/dbnamespace_clarification.yaml
@@ -10,7 +10,7 @@ change_type: enhancement
 component: db
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note:  Clarify `{db.namespace}` when a query explicitly targets a specific database.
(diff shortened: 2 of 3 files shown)


model/trace/database.yaml
```diff
@@ -331,6 +331,9 @@ groups:
 
           For commands that switch the database, this SHOULD be set to the target database (even if the command fails).
 
+          In case a specific query explicitly targets a database different from the one used by the connection,
+          the db.namespace should still be the one from the connection.
+
           If instrumentation cannot reliably determine the current database name, it SHOULD NOT set `db.namespace`.
```
lmolkova, line 337: Since we're changing from "don't set if you're not sure" to "the best effort", I think we need to update text around it too.

E.g.

```suggestion
          Instrumentations, especially ones that don't support query parsing, have limited ability to determine the `db.namespace` query is performed against. For example, instrumentations may not be able to reliably detect when application switches active database with `USE {db}` query or provides database name inside fully-qualified table name.
          
          Instrumentations SHOULD do the best effort to determine actual `db.namespace` query is performed against and  SHOULD fallback to the database name associated with the connection or provided in the connection string when this information is available.
```

model/trace/database.yaml
lmolkova, line 1: There is a similar note on redis that need to be changed

> For commands that switch the database, this SHOULD be set to the target database (even if the command fails).

Could you please also check if there are other database extensions that might need to be updated?

PR #1320, opened 2024-08-06
Add k8s.{pod,node}.cpu.{time,usage} metrics

ChrsMark:
## Changes

This PR adds the `k8s.pod.cpu.time`, `k8s.pod.cpu.usage`, `k8s.node.cpu.time`, `k8s.node.cpu.usage` metrics.
The respective Collector's implementation can be found at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kubeletstatsreceiver/metadata.yaml#L84

Part of https://github.com/open-telemetry/semantic-conventions/issues/1032.


TylerHelmuth (approved):
These metrics have been in please for 4+ years (except usage, whose rename from utilization is in progress).

docs/system/k8s-metrics.md
```diff
@@ -0,0 +1,168 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Container
+--->
+
+# Semantic Conventions for K8s Metrics
```
mx-psi, line 5: We are not consistent in other docs using K8s vs Kubernetes but I think on the title it makes sense to use Kubernetes as we do here https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/k8s.md

mx-psi (approved):

ChrsMark:
@open-telemetry/specs-semconv-approvers could you also take a look?

docs/system/k8s-metrics.md
```diff
@@ -0,0 +1,168 @@
...
+linkTitle: Container
+--->
+
+# Semantic Conventions for Kubernetes Metrics
+
+**Status**: [Experimental][DocumentStatus]
+
+## K8s Metrics
+
+This document describes instruments and attributes for common K8s level
```
lmolkova, line 11: could we document which resources we expect to be used with these metrics?

model/metrics/k8s.yaml
```diff
@@ -0,0 +1,40 @@
+groups:
+  # k8s.pod.cpu.* metrics and attribute group
```
lmolkova, line 2: nit:

```suggestion
  # k8s.pod.cpu.* metrics
```

(seems there are no attributes)

lmolkova:
We probably discussed it somewhere, but could you remind me if we have cpu count available for pods/nods and if there are any plans to report it?

ChrsMark:
> We probably discussed it somewhere, but could you remind me if we have cpu count available for pods/nods and if there are any plans to report it?

The `k8sclusterreceiver` can report the [`k8s.node.allocatable_cpu`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/19ea8dd816400e98c5ada0a3f0a7801b18f16920/receiver/k8sclusterreceiver/metadata.yaml#L516-L517) so at some point this would be declared in SemConv as well.

The container's define the [cpu_limits/cpu_requests](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/19ea8dd816400e98c5ada0a3f0a7801b18f16920/receiver/k8sclusterreceiver/documentation.md#k8scontainercpu_limit), so the the limits/requests of a Pod can be calculated by summarizing its container's limits/requests.

docs/system/k8s-metrics.md
ChrsMark: Could you clarify what "resources" stands for here? Pods and Nodes are K8s resources, do you refer to this or sth else :) ?

lmolkova: I mean that these metrics does not make any sense without specific resources (nods and pods). There is nothing in this document that talks about it.

I'd suggest to say something similar to 

https://github.com/open-telemetry/semantic-conventions/blob/46b1dffba6187e0f4cc8d59a48f8544fb84f11f3/docs/system/hardware-metrics.md?plain=1#L49-L50

ChrsMark: That makes sense, I believe most (hopefully all) of the upcoming metrics will be related to some of the K8s resources.

docs/system/k8s-metrics.md
```diff
@@ -12,6 +12,9 @@ This document describes instruments and attributes for common K8s level
 metrics in OpenTelemetry. These metrics are collected from technology-specific,
 well-defined APIs (e.g. Kubelet's API).
 
+Metrics in `k8s.` instruments should be attached to a [K8s Resource](/docs/resource/k8s.md)
```
lmolkova, line 15:
```suggestion
Metrics in `k8s.` instruments SHOULD be attached to a [K8s Resource](/docs/resource/k8s.md)
```

lmolkova (approved):

TylerHelmuth:
@lmolkova does this repo have merge queue enabled?

lmolkova:
> @lmolkova does this repo have merge queue enabled?

nope, it's all manual. I enabled auto-merge for this one, but all open discussions need to be resolved to allow it to happen.

PR #1322, opened 2024-08-06
Genai user feedback evaluation

truptiparkar7:
## Changes

It provides details for user feedback event which can be used for evaluation purposes. 


model/trace/gen-ai.yaml
```diff
@@ -85,3 +85,17 @@ groups:
           conditionally_required: if and only if corresponding event is enabled
         note: >
           It's RECOMMENDED to format completions as JSON string matching [OpenAI messages format](https://platform.openai.com/docs/guides/text-generation)
+          @@ -106,3 +106,13 @@ groups:
```
trisch-me, line 88: something is wrong here

docs/gen-ai/genai-evaluation-events.md
```diff
@@ -0,0 +1,78 @@
...
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+The event name MUST be `gen_ai.evaluation.user_feedback`.
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`gen_ai.response.id`](/docs/attributes-registry/gen-ai.md) | string | The unique identifier for the completion. | `chatcmpl-123` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.system`](/docs/attributes-registry/gen-ai.md) | string | The Generative AI product as identified by the client or server instrumentation. [1] | `openai` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
nirga, line 43: I wonder why this would be needed specifically. And, more importantly - how can this be achieved if the evaluation happens asynchronously from the generation. I think it's better to use span links or something similar to connect this with the original generation span.

docs/gen-ai/gen-ai-evaluation-events.md
```diff
@@ -0,0 +1,50 @@
...
+<!-- semconv gen_ai.evaluation.user_feedback -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+The event name MUST be `gen_ai.evaluation.user_feedback`.
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
```
lmolkova, line 33: this probably should be in the common section and we should talk about user_feedback as an example.

docs/gen-ai/gen-ai-evaluation-events.md
```diff
@@ -0,0 +1,50 @@
...
+|---|---|---|---|---|---|
+| [`gen_ai.response.id`](/docs/attributes-registry/gen-ai.md) | string | The unique identifier for the completion. | `chatcmpl-123` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
marcklingen, line 35: From my point of view, user feedback often relates to the overall output of an LLM application (which used multiple LLM completions to produce a final response to the user). `gen_ai.response.id` targets LLM completions specifically which limits how user feedback can be used if response id is required. I'd suggest to allow for correlating user feedback with an id that can be set on any non-LLM-completion span, especially if this will define the schema for other evaluation metrics going forward.

Rutledge:
We at Scorecard.io use OTEL for tracing but have our user feedback and model graded (LLM as judge) evaluations stored separately and linked so would love if there is a natural way for this data model to be standardized. 

Kartik makes great points (and we have similar requirements) of evaluations needing to be supported asynchronously from span generation time.

karthikscale3:
> We at Scorecard.io use OTEL for tracing but have our user feedback and model graded (LLM as judge) evaluations stored separately and linked so would love if there is a natural way for this data model to be standardized.
> 
> Kartik makes great points (and we have similar requirements) of evaluations needing to be supported asynchronously from span generation time.

Yeah, thats the number 1 challenge. We are doing it exactly the same way at the moment. Evals are stored in a separate model and linked using the `span_id` as foreign key for referencing the original trace.

marcklingen:
> > We at Scorecard.io use OTEL for tracing but have our user feedback and model graded (LLM as judge) evaluations stored separately and linked so would love if there is a natural way for this data model to be standardized.
> > Kartik makes great points (and we have similar requirements) of evaluations needing to be supported asynchronously from span generation time.
> 
> Yeah, thats the number 1 challenge. We are doing it exactly the same way at the moment. Evals are stored in a separate model and linked using the `span_id` as foreign key for referencing the original trace.

+1, see my comment above. I think this also helps to correlate scores with non-llm calls which is useful

drewby:
> > > We at Scorecard.io use OTEL for tracing but have our user feedback and model graded (LLM as judge) evaluations stored separately and linked so would love if there is a natural way for this data model to be standardized.
> > > Kartik makes great points (and we have similar requirements) of evaluations needing to be supported asynchronously from span generation time.
> > 
> > 
> > Yeah, thats the number 1 challenge. We are doing it exactly the same way at the moment. Evals are stored in a separate model and linked using the `span_id` as foreign key for referencing the original trace.
> 
> +1, see my comment above. I think this also helps to correlate scores with non-llm calls which is useful

We need a correlation(s) that works also when span_id is not available. The trace context is not available in all situations where evaluation scores or feedback are captured.  There could also be other correlations in a system, response_id, session_id, turn_id, that are meaningful to a particular application or toolset. 

Is there a straightforward way to offer more than one option in the conventions? response_id, span_id, turn_id, etc. I'd think you want to `require` at least one be present.

docs/gen-ai/gen-ai-evaluation-events.md
```diff
@@ -0,0 +1,50 @@
...
+It SHOULD, when possible, be parented to the GenAI span describing such response.
+
+<!-- semconv gen_ai.evaluation.user_feedback -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+The event name MUST be `gen_ai.evaluation.user_feedback`.
```
lmolkova, line 31: discussing at GenAI call:
- metrics for score are potentially more useful

lmolkova: gen_ai.evaluation.relevance
dimensions:
- evaluator method
- ...

docs/gen-ai/gen-ai-evaluation-events.md
lmolkova, line 35: this should not be required and we should allow other (any) correlation ids.

axiomofjoy:
Hi all 👋 I’m a maintainer of Arize-Phoenix and OpenInference. Just want to share some experiences with evaluations as it pertains to instrumentation in case it is helpful.

Both the Arize SaaS platform and Phoenix open-source application accept human evaluations submitted via the UI or programmatic evaluations (LLM or code). Since these evaluations come after the span has ended, OpenInference doesn’t attach evaluations to the OTel span under scrutiny itself. Rather, we maintain separate evaluation tables with foreign key relations back to the spans table (in the case of Phoenix, which uses a relational DB) or keep the evaluations as columns in the spans table (in the case of Arize, which uses an OLAP DB). LLM evaluations are traced in the same way as LLM calls in the application, and OpenInference doesn’t currently have semantic conventions specifically related to evaluations.

We think of an evaluation as being comprised of:

- name (required)
- label (optional)
- score (optional)
- explanation (optional)

This attempts to capture that some evaluations are categorical, some numeric, some both, and some accompanied by a human- or LLM-generated explanation. Given the generic nature of the evaluations we ingest, we don’t place max/ min limitations on the score.

Much of what we evaluate is not just LLM calls, but chains, retrievals made via RAG, entire traces or “sessions” (groups of traces corresponding to a back-and-forth conversation between a user and the application). So we allow evaluations to be attached not just to LLM spans, but to any span kind defined in the OpenInference spec (typically, we attach evaluations for traces to the root span). The evaluation interface is pretty consistent no matter what we’re evaluating.

PR #1325, opened 2024-08-08
Fix array attribute examples 

lmolkova:
Examples of array attributes are ambiguous, this PR disambiguates them.

E.g. it's unclear what this example means:

```yaml
type: string[]
examples: [foo, bar]
```

Should it result in 2 examples (`[foo]` and `[bar]`) or one example (`[foo, bar]`)?

Thanks to @jerbly  for the context on how it should be interpreted:

>  The [syntax document in build-tools](https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/syntax.md#examples-for-examples) does call out specifically: "A single example value for an array of strings attribute" examples: ['first element of first array', 'second element of first array'] as valid.

But it goes against yaml array syntax - the above example is equivalent to 
```yaml
type: string[]
examples: 
  - foo # this should be an array, not a string
  - bar # this should be an array, not a string
```

which then fails type checks being introduced in https://github.com/open-telemetry/weaver/pull/300.

The proper syntax should be

```yaml
type: string[]
examples: 
  - [foo] 
  - [foo, bar]
```


lmolkova:
TIL it's possible to write

```yaml
type: string[]
examples: 
  -
    - foo
  - 
    - foo
    - bar
```

🙈

ChrsMark (approved):

mx-psi (approved):

MadVikingGod (approved):

lquerel (approved):

jsuereth (approved):

PR #1326, opened 2024-08-08
Fix yaml schema violations and formatting

lmolkova:
yaml cleanup:
- check against semconv schema, fix violations
- auto-format


model/trace/faas.yaml
```diff
@@ -33,18 +33,6 @@ groups:
...
-    type: span
-    brief: >
-      Semantic Convention for FaaS triggered as a response to some data
-      source operation such as a database or filesystem read/write.
-
-  - id: faas_span.pubsub
-    type: span
-    brief: >
-      Semantic Convention for FaaS set to be executed when messages are
-      sent to a messaging system.
```
MadVikingGod, line 46: All of the rest of the changes are formating.  This removes some spans.   Is there a reason for this?

jsuereth: We should notify the FAAS group, likely.

I agree on removing and adding it back in when they're documented.

jsuereth (approved):

model/trace/faas.yaml
tylerbenson: Thanks for the heads up.  I'm good with removing them and revisiting later.

joaopgrassi (approved):

PR #1329, opened 2024-08-09
process: Add build_id to executable

florianl:
This is a follow up from a discussion in https://github.com/open-telemetry/semantic-conventions/pull/1188#discussion_r1674113373.

**UPDATE:**
- 2024-08-27: created https://github.com/open-telemetry/opentelemetry-specification/pull/4197 to define the requirement of `build_id` for OTel Profiles

## Changes

In the OTel Profiling use case, symbolization is important to resolve numeric addresses into human readable (function-) names. For this use case it is important to uniquely identify executables. There are existing and well established identifiers for executables, like GNU Build ID and Go Build ID.
This proposed change adds these identifiers to the `process.executable` attribute.

For visibility: @open-telemetry/profiling-maintainers


model/registry/process.yaml
```diff
@@ -41,6 +41,18 @@ groups:
           The PID of the process's group leader. This is also the process group
           ID (PGID) of the process.
         examples: [23]
+      - id: process.executable.build_id.gnu
```
mjwolf, line 44:
```suggestion
      - id: process.executable.gnu.build_id
```

Would it make more sense to use the gnu namespace first, so other ELF gnu attributes can be added later?
Potentially `process.gnu.hash`, `process.gnu.version` could be added. (If changing to an ELF schema isn't done).

model/registry/process.yaml
```diff
@@ -41,6 +41,18 @@ groups:
...
+        stability: experimental
+        type: string
+        brief: >
+          The GNU build ID as can be found in the `.note.gnu.build-id` ELF section.
+        examples: ['c89b11207f6479603b0d49bf291c092c2b719293']
+      - id: process.executable.build_id.go
```
mjwolf, line 50: `
```suggestion
      - id: process.executable.go.build_id
```

reverse namespace here too

jsuereth:
If no further comments/updates happen before monday, I'd like to talk about this in the general semconv meeting.

From what I can tell - you may need something *stronger* then semantic conventions here for profiling to work, but I want to confirm.   Would you be able to attend the meeting Monday at 7am PDT to discuss in person?

If not, I'll give the gist of the discusison here:

- Semantic conventions are *optional* meaning, while instrumentaiton from the OpenTelemetry ecosystem has to abide by it, the general community does not.
- If attaching symbols to profiles is something that *must work* for Profiles to be usable, generally, you may need a stronger convention for recording this, e.g. a direct Proto field in OTLP.

I think the answer to this question would resolve my concerns: Will profiles be useful if we can't find symbol tables for executables?

If the answer is "not really": we *can* use semconv, but we may want to full up specify the attributes and force semconv to have them defined [like these attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/semantic-conventions.md#reserved-attributes)

jsuereth:
> If the answer is "not really": we can use semconv, but we may want to full up specify the attributes and force semconv to have them defined [like these attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/semantic-conventions.md#reserved-attributes)

Discussed in the meeting today - and I think the answer is *no really*, we need flexibility here, so these belong in semconv.

model/registry/process.yaml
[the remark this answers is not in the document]
florianl: > ECS has an [ELF schema](https://www.elastic.co/guide/en/ecs/current/ecs-elf.html); there's not currently the equivalent in semantic-conventions. Could you consider adding it, or waiting until the equivalent exists?

The idea is to be provider independent - which ECS is not. Reading the [linked ELF ECS schema](https://www.elastic.co/guide/en/ecs/current/ecs-elf.html), it reads like a mix of Go and general ELF attributes.

Also, the [linked embbeded functionality](https://github.com/open-telemetry/semantic-conventions/issues/1264) does not satisfy the profiling use case afaict.

florianl: The [`.gnu.version`](https://refspecs.linuxbase.org/LSB_3.1.1/LSB-Core-generic/LSB-Core-generic/symversion.html#SYMVERTBL) section of an ELF shall contain the Symbol Version Table. It is not used to uniquely identify an ELF, like `.note.gnu.build-id` - which is proposed with this change.

Adding attributes for all section headers is not the intended purpose of this proposed change.

model/registry/process.yaml
florianl: see https://github.com/open-telemetry/semantic-conventions/pull/1329#discussion_r1723527952

florianl: Switched namespaces as suggested with https://github.com/open-telemetry/semantic-conventions/pull/1329/commits/d779b0d81f1354f230bcd93b67b52f32dee6cfa0.

model/registry/process.yaml
florianl: Switched namespaces as suggested with https://github.com/open-telemetry/semantic-conventions/pull/1329/commits/d779b0d81f1354f230bcd93b67b52f32dee6cfa0.

mjwolf: > The [`.gnu.version`](https://refspecs.linuxbase.org/LSB_3.1.1/LSB-Core-generic/LSB-Core-generic/symversion.html#SYMVERTBL) section of an ELF shall contain the Symbol Version Table. It is not used to uniquely identify an ELF, like `.note.gnu.build-id` - which is proposed with this change.
> 
> Adding attributes for all section headers is not the intended purpose of this proposed change.

Understood, I was just suggesting that to support these fields in the future, no need to add them now

florianl:
Here is an example of how the build ID can be used with public available information, to get symbol/debug information:

```
$ DEBUGINFOD_URLS="https://debuginfod.elfutils.org/" debuginfod-find -vv debuginfo c89b11207f6479603b0d49bf291c092c2b719293
debuginfod_find_debuginfo c89b11207f6479603b0d49bf291c092c2b719293
server urls "https://debuginfod.elfutils.org/"
checking build-id
checking cache dir /home/user/.cache/debuginfod_client
using timeout 90
init server 0 https://debuginfod.elfutils.org/buildid
url 0 https://debuginfod.elfutils.org/buildid/c89b11207f6479603b0d49bf291c092c2b719293/debuginfo
query 1 urls in parallel
Progress 1 / 0
Progress 42 / 0
Progress 48 / 0
header HTTP/1.1 200 OK
header Date: Mon, 26 Aug 2024 11:37:05 GMT
header Server: Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fips mod_qos/11.70
header Cache-Control: public
header Last-Modified: Tue, 20 Sep 2022 15:27:27 GMT
header X-DEBUGINFOD-FILE: /usr/lib/debug/.build-id/c8/9b11207f6479603b0d49bf291c092c2b719293.debug
header X-DEBUGINFOD-ARCHIVE: /srv/mirror/debian-debug/main/c/coreutils/coreutils-dbgsym_9.1-1_amd64.deb
header X-DEBUGINFOD-SIZE: 143112
header Content-Type: application/octet-stream
header Vary: Accept-Encoding
header Content-Encoding: gzip
header Strict-Transport-Security: max-age=16070400
header Transfer-Encoding: chunked
header 
committed to url 0
Progress 8106 / -1
Progress 97162 / -1
server response No error
got file from server
found /home/user/.cache/debuginfod_client/c89b11207f6479603b0d49bf291c092c2b719293/debuginfo (fd=5)
Headers:
X-DEBUGINFOD-FILE: /usr/lib/debug/.build-id/c8/9b11207f6479603b0d49bf291c092c2b719293.debug
X-DEBUGINFOD-ARCHIVE: /srv/mirror/debian-debug/main/c/coreutils/coreutils-dbgsym_9.1-1_amd64.deb
X-DEBUGINFOD-SIZE: 143112
Downloaded from https://debuginfod.elfutils.org/buildid/c89b11207f6479603b0d49bf291c092c2b719293/debuginfo
/home/user/.cache/debuginfod_client/c89b11207f6479603b0d49bf291c092c2b719293/debuginfo
```

lmolkova (approved):

model/registry/process.yaml
```diff
@@ -41,6 +41,25 @@ groups:
...
+        stability: experimental
+        type: string
+        brief: >
+          The Go build ID as it can be retrieved by `go tool buildid <go executable>`.
+        examples: ['foh3mEXu7BLZjsN9pOwG/kATcXlYVCDEFouRMQed_/WwRFB1hPo9LBkekthSPG/x8hMC8emW2cCjXD0_1aY']
+      - id: process.executable.build_id.profiling
+        stability: experimental
+        type: string
+        brief: >
+          Profiling specific build ID for executables. See the OTel specification
```
florianl, line 60: Linking the specification is not possible at this moment, as https://github.com/open-telemetry/opentelemetry-specification/pull/4197 is pending.

model/registry/process.yaml
```diff
@@ -41,6 +41,25 @@ groups:
           The PID of the process's group leader. This is also the process group
           ID (PGID) of the process.
         examples: [23]
+      - id: process.executable.build_id.gnu
+        stability: experimental
+        type: string
+        brief: >
+          The GNU build ID as can be found in the `.note.gnu.build-id` ELF section.
```
christos68k, line 48:
```suggestion
          The GNU build ID as found in the `.note.gnu.build-id` ELF section (hex string).
```

model/registry/process.yaml
```diff
@@ -41,6 +41,25 @@ groups:
...
+        examples: ['c89b11207f6479603b0d49bf291c092c2b719293']
+      - id: process.executable.build_id.go
+        stability: experimental
+        type: string
+        brief: >
+          The Go build ID as it can be retrieved by `go tool buildid <go executable>`.
```
christos68k, line 54:
```suggestion
          The Go build ID as retrieved by `go tool buildid <go executable>`.
```

christos68k (approved):

lmolkova (approved):
It  can be done in a separate PR, but I think that https://github.com/open-telemetry/opentelemetry-specification/blob/a3248bfba81746acc1ba239cd98479cc1796b8e0/specification/profiles/mappings.md should exist in semantic conventions instead and be linked from the spec. 

Reasons:
- you describe attribute names and requirement levels there - semconv allow you do it formally + automate MD rendering and possibly code generation for the group(s) of profiling-related attribute.
- you describe hashing algorithm there and anyone looking into semconv will need to go look into the spec. Same is true for the spec - you need to look into semconv anyway

So having everything described in details and in a formal way in semconv + having a short version in the spec would be easier to understand and navigate. Plus will have higher chance of staying consistent if any changes happen.

The group definition and markdown generation could be done in the same way as you already do for `profile.frame`

jsuereth (approved):

PR #1330, opened 2024-08-09
CosmosDB: define applicable values for `db.operation.name` and add proper reference to `az.namespace`

lmolkova:
- List operations names for Cosmos DB clients. 
- Lowercase operation_type value.
- Add `az.namespace` attribute and reference it (previously undefined, but referenced in md)
- Clean up examples


docs/attributes-registry/db.md
```diff
@@ -170,23 +170,23 @@ This group defines attributes for Azure Cosmos DB.
...
+| -------------------- | ------------------ | ---------------------------------------------------------------- |
+| `batch`              | batch              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `create`             | create             | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `delete`             | delete             | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `execute`            | execute            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `execute_javascript` | execute_javascript | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `head`               | head               | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `head_feed`          | head_feed          | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `invalid`            | invalid            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `Patch`              | patch              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
jcocchi, line 183: nit: is there a reason this is upper case and the others are lower case?

docs/attributes-registry/db.md
```diff
@@ -170,23 +170,23 @@ This group defines attributes for Azure Cosmos DB.
...
+| `query`              | query              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `query_plan`         | query_plan         | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `read`               | read               | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `read_feed`          | read_feed          | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `Replace`            | replace            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
jcocchi, line 188: nit: is there a reason this is upper case and the others are lower case?

.chloggen/1330.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: az, db
+note: Define  `db.operation.name` values for CosmosDB, declare `az.namespace` attribute and add proper reference to it.
```
jcocchi, line 3: nit: officail branding has a space between "Cosmos DB"

docs/attributes-registry/db.md
```diff
@@ -155,7 +155,7 @@ This group defines attributes for Azure Cosmos DB.
 | ------------------------------------ | ------ | --------------------------------- | -------------------------------------- | ---------------------------------------------------------------- |
 | `db.cosmosdb.client_id`              | string | Unique Cosmos client instance id. | `3ba4827d-4422-483f-b59f-85b74211c11d` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `db.cosmosdb.connection_mode`        | string | Cosmos client connection mode.    | `gateway`; `direct`                    | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `db.cosmosdb.operation_type`         | string | CosmosDB Operation Type.          | `Invalid`; `Create`; `Patch`           | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `db.cosmosdb.operation_type`         | string | CosmosDB Operation Type.          | `batch`; `create`; `delete`            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
jcocchi, line 158: nit: official branding has a space between "Cosmos DB"

I owe a branding review on this repo as there may be other places. Can handle this separately if it's easier

model/trace/database.yaml
```diff
@@ -358,7 +358,7 @@ groups:
         examples: ['cosmos-netstandard-sdk/3.23.0\|3.23.1\|1\|X64\|Linux 5.4.0-1098-azure 104 18\|.NET Core 3.1.32\|S\|']
       - ref: db.cosmosdb.connection_mode
         requirement_level:
-          conditionally_required: if not `direct` (or pick gw as default)
+          conditionally_required: if not `direct` (the default value is assumed to be `direct`).
```
jcocchi, line 361: Direct is only default for .NET and Java, and direct mode doesn't exist in other languages. Should we say conditionally required if using .NET or Java?

docs/attributes-registry/db.md
lmolkova: great catch, thanks!

model/trace/database.yaml
lmolkova: good point. If it's only required for .NET and Java, then the default should be `gateway`. 

The only reason to have a default is to slightly reduce telemetry volume. So if we pick gateway as default, most of .NET and Java users will have `direct` populated.
If `direct` is default, then all python, js, go, etc will have the attribute.

wdyt?

docs/attributes-registry/db.md
lmolkova: Thanks! I believe I replaced all `CosmosDB` occurrences in the repo

templates/registry/markdown/weaver.yaml
```diff
@@ -12,7 +12,6 @@ acronyms:
   - CICD
   - CloudEvents
   - CPU
-  - CosmosDB
```
lmolkova, line 15: not used and anyway it should be `Cosmos DB`

model/trace/database.yaml
```diff
@@ -379,3 +379,124 @@ groups:
...
+
+          - `delete_conflict`
+          - `query_conflicts`
+          - `read_all_conflicts`
+          - `read_conflict`
+
+          Container operations:
+
+          - `create_container`
+          - `create_container_if_not_exists` # TODO or not needed?
```
lmolkova, line 410: TODO: folowing-up

FabianMeiswinkel (changes requested):
I think transactional batch but especially bulk needs some more thoughts. Bulk operations are done fundamentally differently in Java and .Net. In Bulk there is an explicit API for executing a set of operations as bulk - so something like `execute_bulk` would probably be good? In .Net whether point operations are executed as bulk or normally depends on some config on the client - which right now you result in `create` or `read` covering operations that are actually executed as part of a bulk operation and would have completely different non-functional characteristics like perf.

model/trace/database.yaml
```diff
@@ -379,3 +379,138 @@ groups:
...
+          - `read_user_defined_function`
+
+          If none of them applies, it's RECOMMENDED to use language-agnostic representation of
+          client method name in snake_case. Instrumentations SHOULD document
+          additional values when introducing new operations.
+
+        examples: ['create_item', 'query_items', 'read_item']
+      - ref: server.port
+        requirement_level:
+          recommended: If not default (443).
```
lmolkova, line 516:
```suggestion
          conditionally_required : If not default (443).
```

docs/database/cosmosdb.md
```diff
@@ -25,51 +25,174 @@ Cosmos DB instrumentation includes call-level (public API) surface spans and net
...
+| [`user_agent.original`](/docs/attributes-registry/user-agent.md) | string | Full user-agent string is generated by Cosmos DB SDK [11] | `cosmos-netstandard-sdk/3.23.0\|3.23.1\|1\|X64\|Linux 5.4.0-1098-azure 104 18\|.NET Core 3.1.32\|S\|` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`db.query.parameter.<key>`](/docs/attributes-registry/db.md) | string | A query parameter used in `db.query.text`, with `<key>` being the parameter name, and the attribute value being a string representation of the parameter value. [12] | `someval`; `55` | `Opt-In` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
 **[1]:** It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.
 If the collection name is parsed from the query text, it SHOULD be the first collection name found in the query and it SHOULD match the value provided in the query text including any schema and database name prefix.
 For batch operations, if the individual operations are known to have the same collection name then that collection name SHOULD be used, otherwise `db.collection.name` SHOULD NOT be captured.
```
jcocchi, line 47: Some of this default language doesn't apply to Cosmos DB because we can only have operations against one container. Can we remove it?

```suggestion
**[1]:** It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.
```

jcocchi (approved):
Looks good to me! Left one minor comment

docs/attributes-registry/db.md
```diff
@@ -170,23 +170,23 @@ This group defines attributes for Azure Cosmos DB.
...
-| `Patch`             | patch              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `Query`             | query              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `QueryPlan`         | query_plan         | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `Read`              | read               | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `ReadFeed`          | read_feed          | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `Replace`           | replace            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `Upsert`            | upsert             | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| Value                | Description        | Stability                                                        |
+| -------------------- | ------------------ | ---------------------------------------------------------------- |
+| `batch`              | batch              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 175:
```suggestion
| `batch`              | transactional batch    | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

docs/attributes-registry/db.md
```diff
@@ -170,23 +170,23 @@ This group defines attributes for Azure Cosmos DB.
...
+| `create`             | create             | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `delete`             | delete             | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `execute`            | execute            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `execute_javascript` | execute_javascript | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `head`               | head               | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `head_feed`          | head_feed          | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `invalid`            | invalid            | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 182: invalid should also be removed

docs/attributes-registry/db.md
```diff
@@ -170,23 +170,23 @@ This group defines attributes for Azure Cosmos DB.
...
+| `patch`              | patch              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `query`              | query              | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `query_plan`         | query_plan         | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `read`               | read               | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `read_feed`          | read_feed          | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 187: probably worth adding "change_feed" in addition to read_feed - because we have to different APIs in the SDK - for normal read feed vs. changefeed

FabianMeiswinkel: IN java there is also a dedicated read_all API (to read all documents of a container or all documents of a logical partition) - we can probably group it under read_feed - but read_many should definitely be its own operation name. SO, my proposal would be to add read_all and read_many and change_feed instead of read_feed

docs/database/cosmosdb.md
```diff
@@ -25,51 +25,174 @@ Cosmos DB instrumentation includes call-level (public API) surface spans and net
...
+Database operations:
 
-**[6]:** If using a port other than the default port for this DBMS and if `server.address` is set.
+- `create_database`
+- `create_database_if_not_exists`
+- `delete_database`
+- `query_databases`
+- `read_all_databases`
+- `read_database`
+- `read_database_throughput`
```
FabianMeiswinkel, line 99:
```suggestion
```

FabianMeiswinkel (approved):
LGTM - Thanks!

trask (approved):

PR #1337, opened 2024-08-16
container: add CSI attributes

### LINKED ISSUE #1119: k8s: new attributes: CSI driver and volume handle
### Area(s)

Semantic convention for otelcol's kubeletstatsreceiver, `area:k8s`.

In k8s, I would like to record CSI driver and volume handle information for relevant Persistent Volume Claims (PVCs) to be able to associate metrics to underlying disks.

Two new attributes, with each mapping 1:1 to the respective driver/volumeHandle fields in [CSIPersistentVolumeSource](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#csipersistentvolumesource-v1-core):

* `csi.driver`
* `csi.volume.handle`

And a new value for the `k8s.volume.type` attribute, `csiPersistentVolume`.

None

### Additional context

This grew from https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055

gdvalle:
Fixes #1119 

## Changes

Add CSI (Container Storage Interface) attributes: `container.csi.plugin.name` and `container.csi.volume.id`.

--- 
xref https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055#discussion_r1670696286


model/registry/k8s.yaml
```diff
@@ -214,3 +214,21 @@ groups:
               value: 'local'
               brief: "A [local](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#local) volume"
               stability: experimental
+      - id: csi.driver
+        type: string
+        stability: experimental
+        brief: >
+          The name of the CSI driver used by the volume.
```
jmacd, line 221:
```suggestion
          The name of the Container Storage Interface (CSI) driver used by the volume.
```

model/registry/k8s.yaml
```diff
@@ -214,3 +214,21 @@ groups:
...
+        note: >
+          This maps 1:1 to [CSIPersistentVolumeSource.driver](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#csipersistentvolumesource-v1-core).
+        examples:
+          - "pd.csi.storage.gke.io"
+      - id: csi.volume.handle
+        type: string
+        stability: experimental
+        brief: >
+          The unique volume name returned by the CSI volume plugin's CreateVolume.
```
jmacd, line 230:
```suggestion
          The unique volume name returned by the Container Storage Interface (CSI) volume plugin's CreateVolume.
```

jmacd (changes requested):

docs/attributes-registry/k8s.md
```diff
@@ -15,8 +15,10 @@ Kubernetes resource attributes.
...
 | --------------------------------------------- | ------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------- |
+| `csi.driver`                                  | string | The name of the CSI driver used by the volume. [1]                                                                                                               | `pd.csi.storage.gke.io`                                                                                                                   | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
joaopgrassi, line 18: Given CSI is a API specification that can be used by any container orchestrator and not only k8s, it's probably wrong that these attributes are part of the k8s file, no? 

I'm no expert, but to me it seems `csi` should be its own thing in the registry, and then `k8s` just refer to its attributes.

joaopgrassi (changes requested):

docs/attributes-registry/k8s.md
gdvalle: I think that's right. I think what's not clear to me is how these attrs should be named generically for all of CSI.

Kube calls this particular field a "driver". CSI seems to refer to the overall interface as a "plugin". In the spec's protobuf this field is returned as part of GetPluginInfo, a `name` field. So should this be called `csi.plugin.name`?

For the other field, `csi.volume.handle`, this also is a k8s-named thing. A handle in this case seems to map to a CreateVolumeResponse, which returns a [Volume](https://github.com/container-storage-interface/spec/blob/2f340622302a8d26950aba0f73f240ea1132a67b/csi.proto#L552), which has `volume_id`. Maybe `csi.volume.id`?

I was a little uneasy making these assumptions without some extra domain knowledge across the other CSI implementations, so I mapped 1:1 with k8s, but maybe we can decide on something reasonable.

joaopgrassi: Yeah I think we could go either way. But of course, the most feature proof way is that CSI is it's own "thing" in semconv, having attributes that map to their concepts (like you said, Plugins). And then each implementation like k8s just import them.

gdvalle: I did the bits above and moved to a new component in https://github.com/open-telemetry/semantic-conventions/pull/1337/commits/c352b3de8eee816867a2c653fa4b3ab551f28f8c, and set the k8s attrs to refer to it as "recommended"...I'm not sure I did that right, but PTAL!

model/registry/k8s.yaml
```diff
@@ -214,3 +214,7 @@ groups:
               value: 'local'
               brief: "A [local](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#local) volume"
               stability: experimental
+      - ref: csi.plugin.name
```
joaopgrassi, line 217: The registry is supposed to contain everything as a "stand alone" thing. So, here we don't refer to the CSI. We rather do it in the actual k8s file we use from the registry. https://github.com/open-telemetry/semantic-conventions/blob/main/model/resource/k8s.yaml

joaopgrassi (changes requested):
Make sure to run update things via the make targets:

```
make fix
make generate-gh-issue-templates
```

docs/attributes-registry/csi.md
```diff
@@ -0,0 +1,20 @@
+<!--- Hugo front matter used to generate the website version of this page:
+--->
+
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/attribute_namespace.md.j2 -->
+
+# Csi
```
joaopgrassi, line 7: You can add `CSI` to the list of acronyms here: https://github.com/open-telemetry/semantic-conventions/blob/main/templates/registry/markdown/weaver.yaml then I _think_ it should generate this with the proper case :)

joaopgrassi (approved):

gdvalle:
I ran into a couple little snags:
https://github.com/open-telemetry/semantic-conventions/pull/1386 fixes the volume UID/GID mappings for the makefile containers.
And then there's one more, a difference between BSD sort and GNU sort defaults. The `get-registry-areas.sh` script just calls `sort`, but apparently an OSX/BSD user committed up the current one, which puts `cloudevents` after `cloud`. With GNU sort, it's the other way.

docs/attributes-registry/csi.md
gdvalle: Thanks! I was wondering how that came about.

mx-psi (approved):

docs/attributes-registry/csi.md
```diff
@@ -0,0 +1,20 @@
...
+
+# CSI
+
+## Container Storage Interface (CSI) Attributes
+
+Attributes related to the Container Storage Interface (CSI).
+
+| Attribute         | Type   | Description                                          | Examples                                                      | Stability                                                        |
+| ----------------- | ------ | ---------------------------------------------------- | ------------------------------------------------------------- | ---------------------------------------------------------------- |
+| `csi.plugin.name` | string | The name of the CSI plugin used by the volume. [1]   | `pd.csi.storage.gke.io`                                       | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 15: Given that CSI is not immediately recognizable by someone not familiar with container storage, I wonder if it would make more sense to make it a sub-namespace in `container` ?

E.g. `container.csi.plugin.name` or `container.storage_interface.plugin.name`. This way domain experts and general population would both understand what it is just by looking at the attribute name.

I understand we have some examples in this repo where we could do the same (e.g. `oci`) which would probably could benefit from the same refactoring *`container.oci`), but not in scope of this PR.

Thoughts? @open-telemetry/semconv-container-approvers @open-telemetry/semconv-k8s-approvers ?

joaopgrassi: Yeah I see how that could be more appealing and easier to understand. By attaching it to the "container" namespace, we put it into context of where its used, rather than have `csi` as a top level thing which might not be obvious.

gdvalle: I'm a bit indifferent, but I'll put down some thoughts. Typing `container.storage_interface` is a little tedious compared to `container.csi`. Maybe it's worth considering what to do for other metrics in this category.

Should it be `jvm`, `java.jvm` or `java.virtual_machine`? Is `jvm` well known enough, or how would one decide that to provide future guidance? What about `oci`: `container.open_container_initiative`? A bit wordy. `container.oci`/`container.csi` feels maybe the least painful and recognizable to me. I think people often know the acronyms more readily than what they stand for.

How about `container.csi`?

lmolkova: yes, `jvm` is different because it's much more recognizable 

https://github.com/open-telemetry/semantic-conventions/blob/ab0e4b4b2ba037e2d678e31d5a10fb55563bc698/docs/general/attribute-naming.md?plain=1#L74-L82

`container.csi` looks good to me (and becomes recognizeable in the context of containers).

ChrsMark: Are the `csi` attributes going to be used plain with containers?
It seems to be k8s specific right? Then why not to call it `k8s.csi.*`?

joaopgrassi: `csi` is a spec that can be used across different container orchestrators, so adding under `k8s` defeats that. 

> container.csi looks good to me (and becomes recognizeable in the context of containers).

Same 👍

ChrsMark: > csi is a spec that can be used across different container orchestrators, so adding under k8s defeats that.


Yeap, I see it's more generic: https://github.com/container-storage-interface/spec. `container.csi` looks good then 👍🏽 .

joaopgrassi: @gdvalle can you please move the attributes to be inside the `container` namespace? Thank you!

gdvalle: Done, barring any placement/CI mistakes...

Thanks everyone for your help, and attention!

lmolkova (approved):
LGTM!

lmolkova:
@jmacd could you please take another look? I believe your feedback is addressed.

model/container/registry.yaml
```diff
@@ -99,3 +99,23 @@ groups:
         brief: >
           Container labels, `<key>` being the label name, the value being the label value.
         examples: [ 'container.label.app=nginx' ]
+      - id: container.csi.plugin.name
+        type: string
+        stability: experimental
+        brief: >
+          The name of the CSI (Container Storage Interface) plugin used by the volume.
```
ChrsMark, line 106: Shall we link to the CSI spec at https://github.com/container-storage-interface/spec for direct reference?

```suggestion
          The name of the CSI ([Container Storage Interface](https://github.com/container-storage-interface/spec)) plugin used by the volume.
```

docs/attributes-registry/csi.md
ChrsMark, line 15: Are these standalone attributes still valid 🤔 ? I thought these would only go under the `container.*` namespace, or I miss something here?

gdvalle: Good catch... I think there's a hole in the generator process that should be removing files prior to generate to catch stragglers like this. I removed it by hand for now.

ChrsMark: fyi @lmolkova @trisch-me

ChrsMark (approved):

docs/attributes-registry/csi.md
lmolkova: great catch! It seems like a merge issue after #1345 (a lot of files were moved around)

jmacd (approved):

ChrsMark:
Hey @gdvalle, is there a reason why those were only added in the registry and not as Resource Attributes as well?
In https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055/files#diff-3343de7bfda986546ce7cb166e641ae88c0b0aecadd016cb253cd5a0463ff464R72-R79 I see they are added as Resource Attributes but from the SemConv perspective that's not clear yet, right?

gdvalle:
> Hey @gdvalle, is there a reason why those were only added in the registry and not as Resource Attributes as well? In https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055/files#diff-3343de7bfda986546ce7cb166e641ae88c0b0aecadd016cb253cd5a0463ff464R72-R79 I see they are added as Resource Attributes but from the SemConv perspective that's not clear yet, right? So I wonder if we can actually use them in the implementations right now.

No, just my oversight. See https://github.com/open-telemetry/semantic-conventions/pull/1499 for adding to resource attributes.

Issue #1269, opened 2024-07-22
Semconv schema: formalize registry definition


lmolkova:
Today we require all attribute definitions to happen in the registry, i.e. :
- attribute definitions should be in `/model/registry` 
- the group id must start with `registry.`

We should consider having a more formal way to mark a group as one defining something.
E.g. 

```yaml
groups:
- id: foo.bar
  type: attribute_registry <-------- this is new
```

or 

```yaml
groups:
- id: new-guid
  type: metric_registry <-------- this is new
```

This would simplify filtering/massaging semconv in code-generation, we'd avoid relying on specific group naming. 
This would move post-factum policies enforcing attribute definitions into schema validation (`id` can only appear on `*_registry` groups and `ref` cannot appear there).

lmolkova:
It's a good question if we can reuse existing types (`attribute_group`, `metric`, etc) and I believe the answer is no:

- we use `attribute_group` a lot to define a common group of ref attributes that are reused between multiple spans or metrics. This helps to avoid writing repetitive notes and similar for different http/messaging/etc metrics. E.g. https://github.com/open-telemetry/semantic-conventions/blob/21b8f4620d2029fe8abb98a4a06655da3232d11d/model/metrics/http.yaml#L2
- we have precedents when [existing metrics are referenced](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/dotnet/dotnet-http-metrics.md#metric-httpclientrequestduration) with caveats (such as specific `error.type` and notes)

Issue #925, opened 2024-04-15
[Discussion] How should events be represented in semantic conventions (one per file or one per namespace)


MSNev:
### Area(s)

area:event

## Description

This issue is a discussion into how should events be described within semantic conventions, there are several options

- One event per event (the client / event preference), but will require additional naming
- One file per namespace, eg all "browser" events should be within the same file
- Leave it up to the event domain as it might depend.

We currently have a mixture with the following 2 PR's open which are showing the basic 2 options

- Single file per namespace: [Add browser events and describe their payload by scheler · Pull Request #838 · open-telemetry/semantic-conventions (github.com)](https://github.com/open-telemetry/semantic-conventions/pull/838/files)
- Single file for event: [Update the device.app.lifecycle event description and constraints by MSNev · Pull Request #794 · open-telemetry/semantic-conventions (github.com)](https://github.com/open-telemetry/semantic-conventions/pull/794)

## One per file

When using one event per file we should have some file / folder naming conventions to avoid accidentally creating duplicate events with the same name but located in different sections / files of semantic conventions.

### Proposals

Note: the `<domain>` is the "area" where the events is created / generated and will most likely be related to the namespace of the event.

- have a folder structure with the namespace
  -  `docs/<domain>/the/namespace/eventname.md`
  - `docs/<domain>/the.namespace/eventname.md`
  -  `docs/events/the/namespace/eventname.md`
  - `docs/events/the.namespace/eventname.md`

- have the filename be full qualified
  - `docs/<domain>/the,namespace.eventname.md`
   - `docs/events/the,namespace.eventname.md`

Questions

- Should the events be "grouped" together (a single `events` folder) or by domain `mobile` / `browser`

## One file per namespace

This enables all related events to be grouped together and leaves the naming so all events are in a single file `events.md`

### Proposals

Note: the `<domain>` is the "area" where the events is created / generated and will most likely be related to the namespace of the event. (`mobile`, `browser`)

- have a folder structure with the namespace
  -  `docs/<domain>/events.md`

One possible downside of this approach is that when / if events become generic they will completely change their location.

MSNev:
Basic Voting Emoji's

👍 - Require a single file per event
🚀 - Require Single file per domain (all events of a namespace / domain should be in a single file `events.md`)
👀 - Leave it up to the event domain as it will depend (defer the discussion)

MSNev:
Where should the events be located in 
[...]

Issue #592, opened 2023-12-07
Should we have metrics registry at some point?


lmolkova:
Metrics should be referenceable in the same way as attributes are, here're some examples:
  - #283 introduces a bunch of .NET-specific metrics, but also refers to standard HTTP ones
  - specific messaging system/db may want to specify which generic metrics it reports and add extra attributes, etc
  - things like connection-level metrics can be shared between other semconvs #454

Referencing mechanism for metrics would be useful to:
- specify differences in a formal manner (e.g. additional attributes, opt-in attributes that's never reported, etc)
- codegen that can generate code for specific metric

lmolkova:
we're getting there with weaver schema v2 that is going to introduce metric refinements - ability to document a specific flavor of the metric. 

Another good reason for metric registry - https://github.com/open-telemetry/semantic-conventions/pull/3267#discussion_r2679327850 - ability to show deprecated metrics in markdown.

PR #1345, opened 2024-08-18
Group yaml files by root namespace instead of signal

lmolkova:
Proposal to change code organization for yaml files.

Now:

```
├── model
│   ├── traces
│   │   ├── database.yaml
│   ├── metrics
│   │   ├── database.yaml
│   ├── registry
│   │   ├── database.yaml
│   │   ├── server.yaml
│   ├── resources
│   │   ├── database.yaml # if ever necessary
├─ database-common.yaml
```

proposal

```
├── model
│   ├── database
│   │   ├── database-common.yaml
│   │   ├── database-events.yaml       # if ever necessary
│   │   ├── database-metrics.yaml
│   │   ├── database-registry.yaml
│   │   ├── database-resources.yaml    # if ever necessary
│   │   ├── database-spans.yaml
│   ├── server
│   │   ├── server-registry.yaml
```

Pros:

- easier to navigate (e.g. open file by name), easier to update semconv when changing multiple signals
- matches MD structure

TODO:

- [x] update CODEOWNERS


MSNev:
This seems to be somewhat related to this discussion (which is limited to events) https://github.com/open-telemetry/semantic-conventions/issues/925

AlexanderWert:
+1 on the above proposal

-1 on the extended idea from the SIG meeting today on also including registry files into the domain directories.  IMHO, that would cause confusion again about mixing definition with usage (i.e. referencing) of attributes.

lmolkova:
> -1 on the extended idea from the SIG meeting today on also including registry files into the domain directories. IMHO, that would cause confusion again about mixing definition with usage (i.e. referencing) of attributes.

@AlexanderWert, I'd like to understand the concern better. 
I feel having registry yamls in the corresponding directory is not strong enough and we should make it more formal - https://github.com/open-telemetry/semantic-conventions/issues/1269. 

I don't see how this PR changes the status quo wrt registry:
1. we don't change markdown structure. It clearly documents what's in the registry and what's not
2. we have checks that ensure how attributes are defined (we name corresponding group as `registry.*`, it must be an attribute group. We don't have any checks for the location of the registry files.
3. If/once we have registry for other signals (metrics and events are good candidates - https://github.com/open-telemetry/semantic-conventions/issues/592), we'd have a problem with having everything in one folder
4. In mind yaml is still an implementation detail. We could move to XML and it won't change anything for the consumers of semantic conventions, so the location of the file is even smaller impl details.

Reasons why I want to have registry attributes in the domain folder: 
- having all yamls for given domain in one folder makes it MUCH easier to work on semantic conventions. 

WDYT?

AlexanderWert:
> @AlexanderWert, I'd like to understand the concern better.
> I feel having registry yamls in the corresponding directory is still not strong enough and we should make it more formal - #1269.

I think the main question is: Is `domain` the same as `namespace`? I would say: NO!
And I agree with you that it doesn't matter much for the end-users / consumers of the SemConv as docs do not change. BUT, for the contributors (especially the ones not contributing on a regular basis to SemConv) it can become very confusing if we start mixing attribute definition (by `namespace`) and attribute usage (by `domain`), again. That's basically what we had before we introduced the registry.

Let me try to illustrate that by a concrete example:
Let's take the domain `HTTP`. We would create a new directory `./model/http` which would contain the following current files:
- `./model/http-common.yaml`
- `./model/trace/http.yaml`
- `./model/metrics/http.yaml`

So far so good. And like that proposal, so far.
But now, we are proposing that we would also add the attribute definition files in there.
The `HTTP` domain covers attributes from many different namespaces:

- `http.*`
- `network.*`
- `url.*`
- `user_agent.*`
- `server.*`
- `error.*`

The proposal of including the registry files into the domains raises following confusion and inconsistencies:
-- So, if the `http.yaml` file from the registry would live under `./model/http/`, where would all the definitions for the other attributes (`network.*`, ...) live?

lmolkova:
@AlexanderWert there is a `general` folder where the common attributes are defined - they can't live in any domain - they are cross-domain by design. This PR does not change that those are grouped together. Some of them have attribute groups defined in addition to the attributes, they also live in the same folder.

I don't see how this PR changes anything around it - if there is a confusion between domain and namespace, it already exists.

> By keeping the attributes registry (also in the yaml) structure, we clearly do a separation of concerns (definition vs. usage of attributes) and avoid the confusion about namespaces and semantical domains.

We do keep the separation of concerns with this PR, I'd like to bring your attention back to the fact that separate folder is a very weak separation, and we have better means to enforce it.

Given that co-locating strongly related files significantly improves semconv development ergonomics and folder structure is internal implementation detail, I would still prefer to keep attribute definitions along with the span/metric/event definitions.

lmolkova:
@AlexanderWert is there a solution that would work for you that does not depend on the folder structure?

We have ~100 files in the registry folder now and will have much more over time. 
We'll have metrics/event registry in the future.

It creates some problems:
- when you modify messaging or databases, you're usually modifying attributes along with spans/metrics definitions. Having them in the same folder makes navigation easy. I can show you a demo.
- sometimes we forget to reference attribute but add it to the registry - e.g. https://github.com/open-telemetry/semantic-conventions/issues/1353

Could you please suggest some ways that would preserve the registry status AND allow to organize yaml folders in this repo in convenient way?

If the folder structure is the only way, then can you explain how it helps or makes anything different?

We have semconv tooling meeting on Wed at 7am PST (tomorrow), would you be able to make it? I'd be great to discuss over the call and there will be other people there.

lmolkova:
Based on the discussion in Semconv tooling WG (on Wed 9/11/2024), changing this PR to group by root-namespace instead of domain (with no exceptions for resources and/or common conventions)

CONTRIBUTING.md
```diff
@@ -93,6 +94,38 @@ Refer to the
...
+
+```
+├── docs
+│   ├── attribute_registry
+│   ├── {root-namespace}
+│   │   ├── README.md
+│   │   ├── ....md
+├── model
+│   ├── {root-namespace}
+│   │   ├── {root-namespace}-events.yaml
```
jsuereth, line 109: Do we need to force `{root-namespace}` in all the file names?  Should we instead make sure directory is easy to see in provenance annotations/error messages?

Makefile
```diff
@@ -210,7 +210,19 @@ chlog-update: $(CHLOGGEN)
...
+	mkdir -p $(TOOLS_DIR)/bin
+	docker run --rm \
+	-u $(id -u ${USER}):$(id -g ${USER}) \
+	--mount 'type=bind,source=$(PWD)/internal/tools/scripts,target=/home/weaver/templates,readonly' \
+	--mount 'type=bind,source=$(PWD)/model,target=/home/weaver/source,readonly' \
+	--mount 'type=bind,source=$(TOOLS_DIR)/bin,target=/home/weaver/target' \
+	$(WEAVER_CONTAINER) registry generate \
+		--registry=/home/weaver/source \
+		--templates=/home/weaver/templates \
+		--config=/home/weaver/templates/registry/areas-weaver.yaml \
```
jsuereth, line 222: Fun use of weaver !

.github/CODEOWNERS
```diff
@@ -16,85 +16,73 @@
...
-/model/registry/artifact.yaml @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-cicd-approvers
-/model/registry/cicd.yaml @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-cicd-approvers
-/model/registry/code.yaml @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-cicd-approvers
-/model/registry/deployment.yaml @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-cicd-approvers
-/model/registry/test.yaml @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-cicd-approvers
-/model/registry/vcs.yaml @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-cicd-approvers
+/model/cicd/                      @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-cicd-approvers
+/model/test/                      @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-cicd-approvers
+/model/vcs/                       @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-cicd-approvers
+/model/artifact/*-registry.yaml   @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-cicd-approvers
```
jsuereth, line 86: Why are thesee just the registry elements?

.github/CODEOWNERS
jsuereth, line 1: Overall - I really like the cleanup on this file.  It's obvious which groups are responsible for which portions.

.github/CODEOWNERS
```diff
@@ -16,85 +16,73 @@
 *   @open-telemetry/specs-semconv-maintainers @open-telemetry/specs-semconv-approvers
 
 # Schemas and schema file tooling
-/schemas/                       @open-telemetry/specs-semconv-approvers @tigrannajaryan
-/internal/tools/schema_check.sh @open-telemetry/specs-semconv-approvers @tigrannajaryan
+/schemas/                                 @open-telemetry/specs-semconv-approvers @tigrannajaryan
```
jsuereth, line 19: For later:  We should think about using the tooling SIG or a separate set of folks to own the schema definitions.

cc @lquerel @tigrannajaryan @open-telemetry/specs-semconv-maintainers

jsuereth (approved):
I really like the direction here, and appreciate the cleanup I'm seeing.

Approving the direction, have a few nits and of course, we may need to pause other merges for you to be able to get this in. Let's discuss in today's meeting!

MSNev (approved):

AlexanderWert (approved):

Makefile
lquerel: Indeed!

lquerel (approved):

Makefile
lmolkova: It's really cool to be able to get a list of actual namespaces instead of relying on the folder structure!

.github/CODEOWNERS
lmolkova: based on the discussion on the SemConv call I'm going to apply ownership to the whole directory

CONTRIBUTING.md
lmolkova: I'd like to enforce it and have the root namespace for two reasons:

- navigation - the way I navigate in the repo is go-to-file in vscode and on GH based on the name and having ~100 files called `spans` or `registry` would be hard.
- consistency (minor-ish)

I don't think it's super-critical, but also having it does not seem to hurt.

PS: if someone has a better way to navigate, please show me, the struggle is real

trask: > navigation - the way I navigate in the repo is go-to-file in vscode and on GH based on the file name and having ~100 files called spans.yaml or registry.yaml would be hard.

fwiw, in Intellij whatever you type in go-to-file is applied across both directories and files, so typing `databasemssql` will get you to `docs/database/mssql.md`

lmolkova: wow, thanks @trask, it's a game-changer!

.github/CODEOWNERS
lmolkova: created https://github.com/open-telemetry/semantic-conventions/issues/1410 to track it and some other areas we forgot to mention

PR #1347, opened 2024-08-19
chore: Keep docker containers up to date with latest patches.

jsuereth:
Prototype makefile and dockerfile depenency that Dependabot will automatically update. This should make sure that using docker as the key tool choice in semconv build allows us to keep on top of security vulnerabilities and fixes.  See #1346 for the other part of this.

FYI @adrielp


arminru (approved):
Renovate might be worth looking into but this workaround here should be good enough for a start.

adrielp (approved):

jsuereth:
@arminru Agreed!  I exhausted my "investigate stuff" budget this week, but I'll look into rennovate next.

I did configure semconv branch rules to allow rennovate if anyone else has time to look into it before I get to it.

lmolkova (approved):

PR #1355, opened 2024-08-20
genai: add system_fingerprint attribute

xrmx:
## Changes

Add `system_fingerprint` attribute to `genai` to track the value from the OpenAI api response as requested in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2759/files/6383978266d85eb058fd8030e8da2b83e6d90043#r1710313834


model/registry/gen-ai.yaml
```diff
@@ -148,3 +148,8 @@ groups:
           If one of the predefined values applies, but specific system uses a different name it's RECOMMENDED to document it in the semantic
           conventions for specific GenAI system and use system-specific name in the instrumentation.
           If a different name is not documented, instrumentation libraries SHOULD use applicable predefined value.
+      - id: gen_ai.system_fingerprint
+        stability: experimental
+        type: string
+        brief: A fingerprint to track any eventual change in the Generative AI environment.
```
xrmx, line 154: Not sure about this description, for reference the instrumentation that requires this field used `System fingerprint of the system that generated the response`

drewby (commented):
I have two questions:
1) Can we confirm that this is returned from more vendors/models than OpenAI?
2) How often does system_fingerprint change? If its low cardinality enough, I'd suggest adding this to the metric as well as a good way to correlate between fingerprint changes and degradations/changes in app performance.

xrmx:
> I have two questions:
> 
>     1. Can we confirm that this is returned from more vendors/models than OpenAI?

I think it's mostly OpenAI specific, not found something similar in Mistral or Anthropic APIs.

>     2. How often does system_fingerprint change? If its low cardinality enough, I'd suggest adding this to the metric as well as a good way to correlate between fingerprint changes and degradations/changes in app performance.

I don't know

drewby:
> > I have two questions:
> > ```
> > 1. Can we confirm that this is returned from more vendors/models than OpenAI?
> > ```
> 
> I think it's mostly OpenAI specific, not found something similar in Mistral or Anthropic APIs.
> 
> > ```
> > 2. How often does system_fingerprint change? If its low cardinality enough, I'd suggest adding this to the metric as well as a good way to correlate between fingerprint changes and degradations/changes in app performance.
> > ```
> 
> I don't know



> > I have two questions:
> > ```
> > 1. Can we confirm that this is returned from more vendors/models than OpenAI?
> > ```
> 
> I think it's mostly OpenAI specific, not found something similar in Mistral or Anthropic APIs.
> 
> > ```
> > 2. How often does system_fingerprint change? If its low cardinality enough, I'd suggest adding this to the metric as well as a good way to correlate between fingerprint changes and degradations/changes in app performance.
> > ```
> 
> I don't know

Can you add this to the metric 

> > I have two questions:
> > ```
> > 1. Can we confirm that this is returned from more vendors/models than OpenAI?
> > ```
> 
> I think it's mostly OpenAI specific, not found something similar in Mistral or Anthropic APIs.
> 
> > ```
> > 2. How often does system_fingerprint change? If its low cardinality enough, I'd suggest adding this to the metric as well as a good way to correlate between fingerprint changes and degradations/changes in app performance.
> > ```
> 
> I don't know

I think its worth adding this to the base set for metrics as Conditionally Required, If available : https://github.com/open-telemetry/semantic-conventions/blob/main/model/metrics/gen-ai.yaml

If it exists in Mistral and Anthropic, then it makes sense to keep in in the generic namespace vs a vendor namespace.

lmolkova (commented):
Thanks a lot for following up and adding this @xrmx!

Given that it seems to be OpenAI specific, it'd need to be `gen_ai.openai.response.system_fingerprint` (or similar) attribute.

But before we can add it, we'd need to add the whole openai specific conventions doc - I created https://github.com/open-telemetry/semantic-conventions/issues/1370 to track it and it's up for grabs for anyone interested.

drewby:
@xrmx OpenAI conventions are now merged. Can you rebase and apply your system_fingerprint attribute to the new openai attributes?

https://github.com/open-telemetry/semantic-conventions/pull/1385

xrmx:
@drewby @lmolkova I have rebased this, PTAL!

model/gen-ai/metrics.yaml
```diff
@@ -37,6 +37,8 @@ groups:
     attributes:
       - ref: gen_ai.openai.response.service_tier
         requirement_level: recommended
+      - ref: gen_ai.openai.response.system_fingerprint
```
lmolkova, line 40: we have a bug that prevents MD table generation - https://github.com/open-telemetry/semantic-conventions/pull/1535, let me get than one merged and then you'll need to re-generate tables again. Sorry for the inconvenience

lmolkova (approved):
Thank you for following up on this one! 
LGTM (but will need to regenerate tables once upstream heals)

nirga (approved):

nirga:
@xrmx can you rebase? I think it will fix the CI failure

xrmx:
Rebased and regenerated table

Issue #127, opened 2023-06-22
Add system metrics reporting total memory capacity or clarify how to recover existing ones


mx-psi:
Current system metrics cover usage of memory, paging/swap memory and filesystems, but we don't currently support the total capacity of any of these systems as a separate metric.

The following metric can't be recovered from the existing system metrics and would need to be added to support this:

| Name                      | Description | Units | Instrument Type | Value Type | Attribute Key | Attribute Values         |
| ------------------------- | ----------- | ----- | ------------------------------------------------- | ---------- | ------------- | ------------------------ |
| system.disk.limit       |   Total memory available in the disk.       | By    | UpDownCounter                                     | Int64      | device       |  (identifier) |

The following metric may be recovered from `system.memory.usage`, but the current description of the attribute values of `state` is insufficient to recover this.

| Name                      | Description | Units | Instrument Type | Value Type | Attribute Key | Attribute Values         |
| ------------------------- | ----------- | ----- | ------------------------------------------------- | ---------- | ------------- | ------------------------ |
| system.memory.limit       |   Total memory available in the machine. Does not include paging/swap memory.       | By    | UpDownCounter                                     | Int64      | n/a        |  n/a |------------ | ---------- | ------------- | ------------------------ |

The following metrics are available as the sum of `used` and `free` (and `reserved` for the file system one). They could be added as a convenience metric:

| Name                      | Description | Units | Instrument Type | Value Type | Attribute Key | Attribute Values         |
| ------------------------- | ----------- | ----- | ------------------------------------------------- | ---------- | ------------- | ------------------------ |
| system.filesystem.limit       |   Total memory available in the disk.       | By    | UpDownCounter                                     | Int64      | device       |  (identifier) |
|                               |             |       |                                                   |            | state         | used, free, reserved |
|                               |             |       |                                                   |            | type          | ext4, tmpfs, etc.    |
|                               |             |       |                                                   |            | mode          | rw, ro, etc.         |
|                               |             |       |                                                   |            | mountpoint    | (path)               |
| system.paging.limit       |   Total paging/swap memory available.       | By    | UpDownCounter                                     | Int64      | n/a       |  n/a |

Items for this issue:
- [ ] Add `system.disk.limit` to the specification
- [x] Clarify set of attribute values for `system.memory.usage` state and consider using `system.memory.total`
- [ ] Consider adding `system.filesystem.limit` and `system.paging.total` if convenience justifies it

[...]

PR #1356, opened 2024-08-20
Add the system.disk.limit metric

rogercoll:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/127

## Changes

Adds a new limit metric for system disks.


rogercoll:
Unrelated CI error: `ERROR: 1 dead links found in ./docs/system/system-metrics.md !
  [✖] https://web.archive.org/web/20180321091318/http://www.onlamp.com/pub/a/linux/2000/11/16/LinuxAdmin.html → Status: 503`

docs/system/system-metrics.md
```diff
@@ -916,6 +917,44 @@ This metric is [recommended][MetricRecommended].
...
 
 
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `system.disk.limit`
+
+This metric is [recommended][MetricRecommended].
```
mx-psi, line 927: Why is this recommended vs opt-in?

docs/system/system-metrics.md
```diff
@@ -916,6 +917,44 @@ This metric is [recommended][MetricRecommended].
...
+<!-- semconv metric.system.disk.limit(metric_table) -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `system.disk.limit` | UpDownCounter | `By` | The total memory of the disk | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
mx-psi, line 938: Is this metric equivalent to adding up another `system.disk` metric across a particular dimension? (I don't think so, but just double checking)

docs/system/system-metrics.md
rogercoll: Initially because the metric can not be derived from others, but seeing that all the "limit" metrics are opt-in I think it makes sense to change it here too: https://github.com/open-telemetry/semantic-conventions/pull/1356/commits/a6f673b1ef6b9b76771c75e998fa50b40ce5fe0a

docs/system/system-metrics.md
rogercoll: I don't think so. All the current `system.disk` metrics are related to the performance of the disk (read/write io and read/write operationd), not the actual usage.

mx-psi (approved):

AlexanderWert (approved):

docs/system/system-metrics.md
```diff
@@ -916,6 +917,44 @@ This metric is [recommended][MetricRecommended].
...
```
jsuereth, line 938: What does "memory" mean here? Like how much memory the disk is using or how much space the disk has available for storage?

rogercoll: Description changed in https://github.com/open-telemetry/semantic-conventions/pull/1356/commits/19271957bb5b1714badbdacb1f4f6e3f48b150f2

jsuereth (approved):

PR #1359, opened 2024-08-20
chore: No longer run weaver as root, or give full filesystem access

jsuereth:
The latest release of weaver docker image no longer runs as the root user.

This does the following:

- Explicitly mounts only the `docs` directory as read-write *only* in make targets that write files.
- Runs as the local UID/GID so files written by the docker image are owned by the local user.
- Bumps to weaver version 0.9.1

cc @lquerel


Makefile
```diff
@@ -113,33 +113,44 @@ yamllint:
...
+	docker run --rm \
+		-u $(id -u ${USER}):$(id -g ${USER}) \
+		--mount 'type=bind,source=$(PWD)/templates,target=/home/weaver/templates,readonly' \
+		--mount 'type=bind,source=$(PWD)/model,target=/home/weaver/source,readonly' \
+		--mount 'type=bind,source=$(PWD)/docs,target=/home/weaver/target' \
 		$(WEAVER_CONTAINER) registry generate \
-		  --registry=/source \
-		  --templates=/weaver/templates \
+		  --registry=/home/weaver/source \
+		  --templates=/home/weaver//templates \
```
xrmx, line 138:
```suggestion
		  --templates=/home/weaver/templates \
```

dependencies.Dockerfile
```diff
@@ -3,7 +3,7 @@
 # Dependabot can keep this file up to date with latest containers.
 
 # Weaver is used to generate markdown docs, and enforce policies on the model.
-FROM otel/weaver:0.8.0 AS weaver
+FROM otel/weaver:v0.9.0 AS weaver
```
arminru, line 6: I noticed that the image tagging scheme was indeed changed as can be seen at https://hub.docker.com/r/otel/weaver/tags but was this intentional or will this be reverted?

jsuereth: This is intentional.  The docker provided github actions provide a best-practice versioning scheme and we now match that.

- Versions now match the tag in github
- `latest` means latest release
- `main` matches the branch in github.

Features would do the same.

lmolkova (approved):

joaopgrassi (approved):

PR #1380, opened 2024-09-03
Add the system.filesystem.limit metric

rogercoll:
Partially fixes # https://github.com/open-telemetry/semantic-conventions/issues/127

## Changes

Adds a new (opt-in) limit metric for the filesystem.


rogercoll:
Unrelated CI error:
```
```

joaopgrassi (approved):

docs/system/system-metrics.md
```diff
@@ -977,7 +978,12 @@ This metric is [recommended][MetricRecommended].
 
 | Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
 | -------- | --------------- | ----------- | -------------- | --------- |
-| `system.filesystem.usage` | UpDownCounter | `By` |  | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `system.filesystem.usage` | UpDownCounter | `By` |  [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 981: brief is missing

model/metrics/system-metrics.yaml
```diff
@@ -248,6 +251,19 @@ groups:
...
 
+  - id: metric.system.filesystem.limit
+    type: metric
+    metric_name: system.filesystem.limit
+    stability: experimental
+    brief: "The total storage capacity of the filesystem"
+    instrument: updowncounter
+    unit: "By"
+    attributes:
+      - ref: system.device
```
lmolkova, line 262: I know it's been like this before this PR, but I don't really understand what this attribute represents. Can we give it a better description or at least a good example?

docs/system/system-metrics.md
rogercoll: Brief added in https://github.com/open-telemetry/semantic-conventions/pull/1380/commits/536fbd8dd01542c0d9c619acab2ecc05de16c54d

model/metrics/system-metrics.yaml
rogercoll: This attribute value depends on the signal which is being used, for example, in system network metrics it will be equal to the network's interface while in system file system metrics it will be set to a disk identification/name. Would a single example be enough?

docs/system/system-metrics.md
```diff
@@ -977,7 +978,12 @@ This metric is [recommended][MetricRecommended].
 
 | Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
 | -------- | --------------- | ----------- | -------------- | --------- |
-| `system.filesystem.usage` | UpDownCounter | `By` |  | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `system.filesystem.usage` | UpDownCounter | `By` | Reports a filesystem's space usage across different states. [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+
+**[1]:** The sum over all `system.filesystem.state` values SHOULD equal the total storage
```
heyams, line 984:
```suggestion
**[1]:** The sum over all `system.filesystem.usage` values SHOULD equal the total storage
```

rogercoll: I followed the same pattern as other's `usage` metrics description, e.g. https://github.com/open-telemetry/semantic-conventions/blob/1dc64eef5215dec963efcb66dfe91360e59ab532/docs/system/system-metrics.md#metric-systemmemoryusage

Should we create a follow-up PR to fix them all? What we want to explain in this note is that the sum of all `system.filesystem.usage` values over the different `system.filesystem.state` attributes should be equal to the limit.

heyams: > the sum of all system.filesystem.usage values over the different system.filesystem.state attributes

can you fix the description in this PR and then use a follow-up PR to fix the rest?

heyams: I did a quick search, didn't find elsewhere has the same wrong description. can you list them all?

model/metrics/system-metrics.yaml
lmolkova: You can provide new brief/description that adds context/specialization to the attribute when it's referenced. E.g. here it's referenced under `system.filesystem.limit` and we could do something like

```suggestion
      - ref: system.device
         brief: The name of the disk.
```

rogercoll: Great, I was not aware of this feature. Added in https://github.com/open-telemetry/semantic-conventions/pull/1380/commits/f9deb24349f9a09a3809adf3e116e5ff25544106

docs/system/system-metrics.md
rogercoll: Fixed in https://github.com/open-telemetry/semantic-conventions/pull/1380/commits/db0fd2242196e91258909c7bf958cdd3f44ad7d6

rogercoll: > I did a quick search, didn't find elsewhere has the same wrong description. can you list them all?


- https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemmemoryusage
- https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemmemorylimit

heyams (approved):

lmolkova (approved):

docs/system/system-metrics.md
lmolkova: I enabled auto-merge, so the PR will be merged once this discussion is resolved.

PR #1385, opened 2024-09-04
Add system specific conventions for OpenAI

### LINKED ISSUE #1370: GenAI: Add system-specific conventions
We have a pattern in other conventions where we define generic conventions along with system-specific ones. 
System-specific conventions describe how generic apply to them (e.g. we'd want to update some attribute descriptions for Cohere or Anthropic) and may define additional attributes.

This issue tracks adding docs for all known `gen_ai.system` (but it should probably be broken down into individual PRs)

- [ ] OpenAI 
- [ ] Cohere
- [ ] Anthropic
- [ ] Vertex AI
- [ ] Azure AI Inference is coming soon

drewby:
Fixes #1370 

## Changes

Adds system specific attributes for OpenAI.
*  gen_ai.openai.request.see
*  gen_ai.openai.request.response_format
* gen_ai.openai.request.json_schema.name
* gen_ai.openai.request.service_tier
* gen_ai.openai.response.service_tier

Adds an openai.md document to outline the attributes that should be added to spans in addition to the general attributes. It also outlines the attributes which should be added to generative AI metrics.


drewby:
I'm creating this as Draft since its the first vendor specific conventions we are adding to the gen_ai area.
 For spans, its a list of the new attributes for OpenAI and the general attributes combined.
For metrics, its only the OpenAI attributes and refers to the general metric document for other attributes.

* Should these be in one file or separate files for metrics and spans?
* Should it be a combined list of openai and general attributes, or only the openai attributes that get added?

docs/gen-ai/openai.md
```diff
@@ -0,0 +1,162 @@
...
+| [`gen_ai.request.model`](/docs/attributes-registry/gen-ai.md) | string | The name of the GenAI model a request is being made to. [2] | `gpt-4` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.system`](/docs/attributes-registry/gen-ai.md) | string | The Generative AI product as identified by the client or server instrumentation. [3] | `openai` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [4] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` if the operation ended in an error | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.openai.request.json_schema.name`](/docs/attributes-registry/gen-ai.md) | string | The name of the JSON schema used for the response. | `chat_response` | `Conditionally Required` [5] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 45: do we need this attribute? the only concern is potential bloating of telemetry - every time we add a new attribute we should try to balance the usefulness of it vs potential cost of populating it. 

If we don't know the scenario when it could be useful, let's not add it for now

docs/gen-ai/openai.md
```diff
@@ -0,0 +1,162 @@
...
+| [`gen_ai.openai.request.response_format`](/docs/attributes-registry/gen-ai.md) | string | The response format that is requested. | `json` | `Conditionally Required` if the request includes a response_format | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.openai.request.seed`](/docs/attributes-registry/gen-ai.md) | int | Requests with same seed value more likely to return same result. | `100` | `Conditionally Required` if the request includes a seed | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.openai.request.service_tier`](/docs/attributes-registry/gen-ai.md) | string | The service tier requested. May be a specific tier, detault, or auto. | `auto`; `default`; `standard` | `Conditionally Required` if the request includes a service_tier | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.openai.response.service_tier`](/docs/attributes-registry/gen-ai.md) | string | The service tier used for the response. | `standard` | `Conditionally Required` if the request includes a service_tier | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 49: do we need to capture both - request and response tiers?

docs/gen-ai/openai.md
```diff
@@ -0,0 +1,162 @@
...
+| [`gen_ai.request.max_tokens`](/docs/attributes-registry/gen-ai.md) | int | The maximum number of tokens the model generates for a request. | `100` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.request.presence_penalty`](/docs/attributes-registry/gen-ai.md) | double | The presence penalty setting for the GenAI request. | `0.1` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.request.stop_sequences`](/docs/attributes-registry/gen-ai.md) | string[] | List of sequences that the model will use to stop generating further tokens. | `["forest", "lived"]` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.request.temperature`](/docs/attributes-registry/gen-ai.md) | double | The temperature setting for the GenAI request. | `0.0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.request.top_k`](/docs/attributes-registry/gen-ai.md) | double | The top_k sampling setting for the GenAI request. | `1.0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 56: I don't think openai has this parameter - foes it?

docs/gen-ai/openai.md
```diff
@@ -0,0 +1,162 @@
...
+| [`gen_ai.response.id`](/docs/attributes-registry/gen-ai.md) | string | The unique identifier for the completion. | `chatcmpl-123` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.response.model`](/docs/attributes-registry/gen-ai.md) | string | The name of the model that generated the response. [7] | `gpt-4-0613` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.usage.input_tokens`](/docs/attributes-registry/gen-ai.md) | int | The number of tokens used in the GenAI input (prompt). | `100` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`gen_ai.usage.output_tokens`](/docs/attributes-registry/gen-ai.md) | int | The number of tokens used in the GenAI response (completion). | `180` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 62: might be useful to update the description for input/output tokens for openai and tell that they are coming as prompt/completion tokens from the response

docs/gen-ai/openai.md
```diff
@@ -0,0 +1,162 @@
...
+| `vertex_ai` | Vertex AI | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## OpenAI Metric attributes
```
lmolkova, line 141: I hope there will be a formal way to do it at some point, for now I think we should rather mention which metrics we would report for openai clients (probably all) and document how they are reported differently than generic ones.

You can check out an example from https://github.com/open-telemetry/semantic-conventions/blob/main/docs/dotnet/dotnet-http-metrics.md#metric-httpserverrequestduration

lmolkova (commented):
Thank you! Left some comments

lmolkova:
I've sent a similar one for Azure AI Inference - #1393 (without metrics though - no plans to report anything custom there yet).

> Should these be in one file or separate files for metrics and spans?

I think either way. I'd keep them in one file until it grows too big.

drewby:
> Yes, the point is to describe all attributes that are applicable (and how if it's not obvious). Otherwise it's not clear which generic ones apply to openai.

Thanks, I brought this up in the Azure inference list as well, but is there a way to highlight, sort, or otherwise make it clear which ones are specific to openai? One can scan the list and fine the namespaces, but it'd be easier if they stood out visually.

docs/gen-ai/openai.md
drewby: If "auto" is used it would be important to know which tier was chosen by the service to generate the response, if one wants to compare performance across tiers.

docs/gen-ai/openai.md
drewby: It is included because I extended the genai span which has it. Is there a way for mt override its inclusion, or do I need to remove the extends and copy the attribute list?

model/registry/gen-ai.yaml
```diff
@@ -148,3 +148,56 @@ groups:
           If one of the predefined values applies, but specific system uses a different name it's RECOMMENDED to document it in the semantic
           conventions for specific GenAI system and use system-specific name in the instrumentation.
           If a different name is not documented, instrumentation libraries SHOULD use applicable predefined value.
+  - id: registry.gen_ai.openai
```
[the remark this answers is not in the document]
drewby: I'm not able to pull the otel/weaver image locally now. Can't find the linux/arm64/v8 image. It worked a few days ago.

docs/gen-ai/openai.md
lmolkova: There is an example in https://github.com/open-telemetry/semantic-conventions/pull/1393 - effectively we add a new group with attributes used everywhere and then extend this group in system-specific conventions. 

I wonder if top_k is in fact a generic attribute though - it seems it's not applicable to a bunch of systems - but that's out of scope of this PR.

lmolkova: wait, it seems `top_logprobs` is the OpenAI parameter that's aligns with Anthropic and Cohere top_k

[OpenAI top_logprobs](https://platform.openai.com/docs/api-reference/chat/create#chat-create-top_logprobs):

> An integer between 0 and 20 specifying the number of most likely tokens to return at each token position, each with an associated log probability. logprobs must be set to true if this parameter is used.


[Cohere k](https://docs.cohere.com/reference/chat#request.body.k): 
> Ensures only the top k most likely tokens are considered for generation at each step. Defaults to 0, min value of 0, max value of 500.

[Anthropic top_k](https://docs.anthropic.com/en/api/messages#:~:text=top_k,integer):

> Only sample from the top K options for each subsequent token.
> Used to remove "long tail" low probability responses. [Learn more technical details here](https://towardsdatascience.com/how-to-sample-from-language-models-682bceb97277).

drewby: The closest parameter in OpenAI to the other two is *top_p*. Both seek to limit the number of likely tokens used in each prediction. *top_k* sorts by the probability and simply takes the top k number of tokens, then choses a next token.  *top_p* uses a distribution and then cuts off the tokens when CDF exceeds p. However, in these cases you still get one set of token completions. 

*top_logprobs* has a different behavior. This parameter combined with the boolean *logprobs* parameter, will return n number of tokens for each prediction, along with their probabilities for the client to do some processing with.  Each prediction has multiple tokens. 

I did add *top_p* to the openai set in my last set of commits. However, I think we may want to tackle *logprobs* in its own PR because there could also be design considerations on how the response is captured.

model/trace/gen-ai.yaml
```diff
@@ -85,3 +83,33 @@ groups:
           conditionally_required: if and only if corresponding event is enabled
         note: >
           It's RECOMMENDED to format completions as JSON string matching [OpenAI messages format](https://platform.openai.com/docs/guides/text-generation)
+
+  - id: trace.gen_ai.client
```
drewby, line 87: @lmolkova This is how I'm resolving the fact that *top_k* is not included in openai. Flagging it here for your review.

mattmccleary (approved):

karthikscale3 (approved):
Looks good to me. Thanks @drewby

docs/gen-ai/openai.md
```diff
@@ -0,0 +1,185 @@
...
+| `_OTHER` | A fallback error value to be used when the instrumentation doesn't define a custom value. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+
+
+`gen_ai.openai.request.response_format` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `json_object` | JSON object response format | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `json_schema` | JSON schema response format | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `text` | Text response format | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
codefromthecrypt, line 100: `verbose_json` is defined [here](https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L1356).. is that an error in the openai spec or an ommission here?

model/trace/gen-ai.yaml
```diff
@@ -85,3 +83,33 @@ groups:
...
+          conditionally_required: if the request includes a seed
+      - ref: gen_ai.openai.request.response_format
+        requirement_level:
+          conditionally_required: if the request includes a response_format
+      - ref: gen_ai.openai.request.service_tier
+        requirement_level:
+          conditionally_required: if the request includes a service_tier
+      - ref: gen_ai.openai.response.service_tier
+        requirement_level:
+          conditionally_required: if the request includes a service_tier
```
lmolkova, line 111:
```suggestion
          conditionally_required: if the response was received and includes a service_tier
```

model/trace/gen-ai.yaml
```diff
@@ -85,3 +83,33 @@ groups:
...
+    attributes:
+      - ref: gen_ai.openai.request.seed
+        requirement_level:
+          conditionally_required: if the request includes a seed
+      - ref: gen_ai.openai.request.response_format
+        requirement_level:
+          conditionally_required: if the request includes a response_format
+      - ref: gen_ai.openai.request.service_tier
+        requirement_level:
+          conditionally_required: if the request includes a service_tier
```
lmolkova, line 108: maybe we can skip it if it has a default value?  - it'll be set on the response anyway.

```suggestion
          conditionally_required: if the request includes a service_tier and the value is not `auto`
```

> [When not set, the default behavior is 'auto'.](https://platform.openai.com/docs/api-reference/chat/create)

(ironically they have a `default` value which is not a default value 😄 )

model/registry/gen-ai.yaml
```diff
@@ -148,3 +148,51 @@ groups:
...
+              value: "default"
+              brief: The system will utilize the default scale tier.
+              stability: experimental
+        brief: The service tier requested. May be a specific tier, detault, or auto.
+        examples: ['auto', 'default', 'standard']
+      - id: gen_ai.openai.response.service_tier
+        stability: experimental
+        type: string
+        brief: The service tier used for the response.
+        examples: ['standard']
```
codefromthecrypt, line 198: should this be 'scale' and 'default'? I don't see 'standard' defined https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L10079

model/registry/gen-ai.yaml
```diff
@@ -148,3 +148,51 @@ groups:
...
+            - id: auto
+              value: "auto"
+              brief: The system will utilize scale tier credits until they are exhausted.
+              stability: experimental
+            - id: default
+              value: "default"
+              brief: The system will utilize the default scale tier.
+              stability: experimental
+        brief: The service tier requested. May be a specific tier, detault, or auto.
+        examples: ['auto', 'default', 'standard']
```
codefromthecrypt, line 193: I don't see 'standard' defined here, but I do see 'auto' and 'default' https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L9924

codefromthecrypt (approved):
Only glitch is service tiers, and I wonder if there would be difference vs OpenAI the platform vs Azure OpenAI. Certainly emulators would have differences. Should we mention the openai api docs, but caution that platform differences apply in exact values?

docs/gen-ai/openai.md
```diff
@@ -0,0 +1,185 @@
...
+<!-- tocstop -->
+
+The Semantic Conventions for [OpenAI](https://openai.com/) extend and override the semantic conventions
+for [Gen AI Spans](gen-ai-spans.md) and [Gen AI Metrics](gen-ai-metrics.md).
+
+`gen_ai.system` MUST be set to `"openai"`.
+
+## OpenAI Span attributes
+
+These attributes track input data and metadata for a request to an OpenAI model. The attributes include general Generate AI
```
lmolkova, line 27:
```suggestion
These attributes track input data and metadata for a request to an OpenAI model. The attributes include general Generative AI
```

docs/gen-ai/openai.md
```diff
@@ -0,0 +1,185 @@
...
+
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## OpenAI Metric attributes
+
+OpenAI metrics follow metrics defined in [Generative AI metrics](gen-ai-metrics.md) with the noted additional attributes.
```
lmolkova, line 137: nit: 

```suggestion
OpenAI metrics follow [Generative AI metrics](gen-ai-metrics.md) with the noted additional attributes.
```

lmolkova, line 137: mind adding a small comment to generic GenAI metrics saying that individual systems may include additional system-specific attributes and it's best to check system-specific docs if available?

lmolkova (approved):
Left some minor comments, but LGTM overall!

model/metrics/gen-ai.yaml
lmolkova, line 1: this file was moved to model/gen-ai/metrics.yaml - please rebase

model/registry/gen-ai.yaml
lmolkova, line 1: this file was moved to model/gen-ai/registry.yaml - please rebase

model/trace/gen-ai.yaml
lmolkova, line 1: and this one was moved to model/gen-ai/spans.yaml

docs/gen-ai/openai.md
codefromthecrypt: double-checked, and verbose_json is also in the openai docs (as is vtt) https://platform.openai.com/docs/api-reference/audio/createTranscription#audio-createtranscription-response_format

model/registry/gen-ai.yaml
codefromthecrypt: https://github.com/openai/openai-openapi/pull/337 about "scale" being wrong. Should be 'auto' or 'default' afaict

model/graphgl/registry.yaml
lmolkova, line 1: oops

drewby: Yes, I oopsed and will fix it today. :(

drewby: @lmolkova, I think I have this all squared now.

model/registry/gen-ai.yaml
drewby: Thanks! removed 'standard'

model/registry/gen-ai.yaml
drewby: Thanks! Updated.

docs/gen-ai/openai.md
drewby: It seems 'verbose_json' is in the context of audio? We could add it in another PR?

joaopgrassi (approved):

PR #1386, opened 2024-09-04
ci: fix makefile docker ownership

gdvalle:
Fix the docker user flag to properly fetch group and user IDs.

Without this, on linux with GNU Make 4, a new registry component will end up throwing an error:
```
[error] Unable to write file "docs/attributes-registry/csi.md":
```
because it writes out as root rather than the current user.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,8 @@ WEAVER_CONTAINER=$(shell cat dependencies.Dockerfile | awk '$$4=="weaver" {print
 SEMCONVGEN_CONTAINER=$(shell cat dependencies.Dockerfile | awk '$$4=="semconvgen" {print $$2}')
 OPA_CONTAINER=$(shell cat dependencies.Dockerfile | awk '$$4=="opa" {print $$2}')
 
+DOCKER_USER=$(shell id -u):$(shell id -g)
+
 
 # TODO: add `yamllint` step to `all` after making sure it works on Mac.
 .PHONY: all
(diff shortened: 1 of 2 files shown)


jsuereth (approved):

lmolkova (approved):

lmolkova (changes requested):
it breaks policy check 

docker run --rm \
	-u 1001:127 \
	--mount 'type=bind,source=/home/runner/work/semantic-conventions/semantic-conventions/policies,target=/home/weaver/policies,readonly' \
	--mount 'type=bind,source=/home/runner/work/semantic-conventions/semantic-conventions/model,target=/home/weaver/source,readonly' \
	otel/weaver:v0.9.1 registry check \
	--registry=/home/weaver/source \
	--baseline-registry=https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.27.0.zip[model] \
	--policy=/home/weaver/policies
....

  × Cache directory not created: Permission denied (os error [13](https://github.com/open-telemetry/semantic-conventions/actions/runs/10708326730/job/29690368799?pr=1386#step:3:14))

gdvalle:
> it seems to break policy check :(

Doh. I don't think we need it here anyway, it's just a read-only mount, and unless we have world read off (not so common) this should work fine. Seems the weaver container is running [under its own UID/GID](https://github.com/open-telemetry/weaver/blob/2e74ab36e6bd6599bd9231b83e9002e40f199712/Dockerfile#L29).

Removed in https://github.com/open-telemetry/semantic-conventions/pull/1386/commits/33d04316ecea2833ccb44f3c4fb01aafccc33062

lmolkova (approved):

PR #1389, opened 2024-09-04
feat!: Update GraphQL span name convention

### LINKED ISSUE #1361: Concerns about high cardinality in GraphQL span names
### Area(s)

area:graphql

The Ruby SIG has some concerns about the cardinality of span names for the current GraphQL semantic convention.

@karmingc [opened a PR on the opentelemetry-ruby-contrib repo to update the GraphQL span names](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/966). 

This would update Ruby's existing name, `graphql.execute_query`, to [the current semantic convention's format, `<graphql.operation.type> <graphql.operation.name>`.](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/graphql/graphql-spans.md?plain=1#L12-L15)

During review, @robertlaurin raised a concern about unbounded cardinality for the span names if operation name and operation type are included:
> ...OpenTelemetry consistently discourages high cardinality span names, and this by definition is unbounded cardinality.
>
> GraphQL operation name is just a free form label supplied by the client caller. There's absolutely no constraints, if we don't put the raw path in an http server span name, I'm not at all convinced we should all operation name here.
>
> Fwiw query, mutation, subscription seems reasonable on its own.
[(source)](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/966/files#r1597133695)

The PR is currently blocked until we can come to a resolution about the appropriate span name.  

`<graphql.operation.type>` seems sufficient, since the `<graphql.operation.name>` is available in span attributes. 

The GraphQL semantic convention could also be updated to more closely match the DB span name, by adding a `SHOULD` instead of a `MUST` for the name and/or add the `low_cardinality` caveat about `db.operation.name`. 

> The span name SHOULD be {db.operation.name} {target} if there is a (low-cardinality) {db.operation.name} available (see below for the exact definition of the [{target}](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#target-placeholder) placeholder).
[[source]](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md?plain=1#L59-L60)

### Additional context

This may be similar to https://github.com/open-telemetry/semantic-conventions/issues/182, but since the problem focuses specifically on the span name, I thought it warranted a new issue.

kaylareopelle:
Fixes #1361

## Changes

Due to concerns about high cardinality with the current span name format, use only the `graphql.operation.type` attribute for the span name. The higher cardinality attribute, `graphql.operation.name` can still be used in queries, as it is an attribute on the span. The original fallback span name, `GraphQL Operation`, is preserved.

* [X] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions. **(I don't think this needs to be updated because it's a span name and not an attribute)**


.chloggen/update-graphql-span-name.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: breaking
```
kaylareopelle, line 1: Since the GraphQL spans page is experimental, I'm not sure if this should be a `breaking` change or an `enhancement`.

docs/graphql/graphql-spans.md
```diff
@@ -9,10 +9,9 @@ linkTitle: GraphQL Server
 This document defines semantic conventions to apply when instrumenting the GraphQL implementation. They map GraphQL
 operations to attributes on a Span.
 
-The **span name** MUST be of the format `<graphql.operation.type> <graphql.operation.name>` provided that
-`graphql.operation.type` and `graphql.operation.name` are available. If `graphql.operation.name` is not available, the
-span SHOULD be named `<graphql.operation.type>`. When `<graphql.operation.type>` is not available, `GraphQL Operation`
-MAY be used as span name.
+The **span name** MUST be of the format `<graphql.operation.type>` provided that
```
kaylareopelle, line 12: Most other references I could find for `**span name**` use `SHOULD`-level requirements rather than `MUST`-level requirements. I left this as `MUST` to remain closer to the previous convention, but I can also see benefits in making it a `SHOULD`-level requirement.

.chloggen/update-graphql-span-name.yaml
lmolkova: it's an enhancement and also a breaking change :)

Let's keep `breaking` - it's important to announce them.

docs/graphql/graphql-spans.md
lmolkova: I'd support SHOULD, especially because the next sentence implies that operation type is not always available.

lmolkova, line 12: it seems that the main reason for this change is avoiding high cardinality. 

What's the problem with operation name cardinality? Is it theoretically high or high in practice?

E.g. for HTTP server spans, you could have 100s of `http.route` values, but we use it in the span name given that usually cardinality is much lower and that route is such a useful grouping keu.

Or for databases we say the following:

https://github.com/open-telemetry/semantic-conventions/blob/1507791993b8893c10ca24796b74b5600ad42738/docs/database/database-spans.md?plain=1#L59-L63

would it be reasonable to follow the same pattern here?

kaylareopelle: Excellent. Updated in 3806796

joaopgrassi: If I understand it right, the problem is not so much of cardinality but because it could be unbounded, given it's user specified? https://graphql.org/learn/queries/#operation-name and I'm not sure how an instrumentation can know if a operation name is "low cardinality".

lmolkova: it seems query name is something user optionally provides

https://stackoverflow.com/questions/34011964/what-is-the-point-of-naming-queries-and-mutations-in-graphql

Specifically the second answer https://stackoverflow.com/a/52542928/3239947

> We use named queries so that they can be monitored consistently, and so that we can do persistent storage of a query. The duplication is there [for query variables](https://medium.com/the-graphqlhub/graphql-tour-variables-58c6abd10f56) to fill the gaps.

> The advantage is that the same query each time, not a different string because the query variables are the bit that differs. This means you can build tools on top of those queries because you can treat them as immutable.

If that's accurate, then it's intended for exactly this purpose and intended to have lowish/user-controlled cardinality. Seems like a perfect thing for the span name - for databases we actually would want if user provided us something like this so we could provide better diagnostics.

We should keep it.

robertlaurin: >If that's accurate, then it's intended for exactly this purpose and intended to have lowish/user-controlled cardinality. 

Unfortunately the intended use and reality are not aligned. If an API client decides that it wants to generate an operation name based off of an incrementing id, there isn't much the server can do as it is technically a valid label.

Recording that operation name as an attribute on the span can be quite useful. Setting your span name to something with unbounded cardinality is not.

There seems to be an incorrect assumption here the systems being instrumented here control both the client callers. 

This is not a hypothetical concern.

lmolkova: > If an API client decides that it wants to generate an operation name based off of an incrementing id,

Is it a client library or a an application that's using this client library?

If application user decides to use dynamic operation names, they would experience consequences of it in their telemetry systems. Primarily grouping by span name will be useless. It's unlikely that anything will break.

If they configured span->metrics, these metrics will reach cardinality limit and will be useless.

---------

If my assumptions are correct, then removing operation name would be bad for everyone who uses operation name properly in order to protect those who use it incorrectly - this is a weird design choice to make.

Please help me understand where (if) they are wrong.

robertlaurin: >Is it a client library or a an application that's using this client library?

The instrumentation in question here is for server side spans generated from a graphql query/mutation/subscription, it receives client requests.

>If application user decides to use dynamic operation names, they would experience consequences of it in their telemetry systems.

The server and client can be different entities. Imagine a public graphql api, anyone can issue queries, and you have no guarantee that operation name will conform to any pattern or bounded cardinality.

>*Primarily grouping by span name will be useless*.

I completely agree, the current specification is directing instrumentation authors to make the graphql execute span useless for any grouping functions.

>If they configured span->metrics, these metrics will reach cardinality limit and will be useless

Agree here as well, it means that this span is no longer a viable option metric generation.

In general it needlessly complicates any aggregate analysis of this span.

>If my assumptions are correct, then removing operation name would be bad for everyone who uses operation name properly in order to protect those who use it incorrectly - this is a weird design choice to make.

- Operation name can be of unbounded cardinality (unless you control all of your client requests)
- Operation name is recorded as an attribute
- Operation name is also denormalized into the span name in the form `"#{operation_type} #{operation_name}"`

By injecting the operation name into the span name it means for any graphql api that doesn't control all of the client consumers it is vulnerable to the issues that come with unbounded cardinality on a span name.

By not injecting the operation name into the span name, you are no longer vulnerable to those issue, and you can still optionally do any grouping or analysis based off of the span name and `graphql.operation.name` attribute.

joaopgrassi: Maybe we can reach a compromise here, that by default, operation name is not added, but maybe can be configured/added in a optional manner? This then solves the problem for both categories of grapql servers - those who control the clients and those who don't.

docs/graphql/graphql-spans.md
```diff
@@ -9,10 +9,9 @@ linkTitle: GraphQL Server
 This document defines semantic conventions to apply when instrumenting the GraphQL implementation. They map GraphQL
 operations to attributes on a Span.
 
-The **span name** MUST be of the format `<graphql.operation.type> <graphql.operation.name>` provided that
-`graphql.operation.type` and `graphql.operation.name` are available. If `graphql.operation.name` is not available, the
-span SHOULD be named `<graphql.operation.type>`. When `<graphql.operation.type>` is not available, `GraphQL Operation`
-MAY be used as span name.
+The **span name** SHOULD be of the format `<graphql.operation.type>` provided that
+`graphql.operation.type` is available. If `graphql.operation.type` is not available, the
+span SHOULD be named `GraphQL Operation`.
```
kaylareopelle, line 14: @lmolkova @joaopgrassi @robertlaurin - Based on your earlier discussion, how does this update sound?

```suggestion
The **span name** SHOULD be of the format `<graphql.operation.type>` provided
`graphql.operation.type` is available. If `graphql.operation.type` is not available, the
span SHOULD be named `GraphQL Operation`.

In addition, Instrumentation MAY provide a configuration option to add 
`<graphql.operation.name>` to the span name, making the name 
`<graphql.operation.type> <graphql.operation.name>`. In this case, if 
`<graphql.operation.name>` is not available, the span SHOULD be named 
`<graphql.operation.type>`.
```

lmolkova: looks good!

A small readability suggestion, feel free to ignore if you don't like it.

> [!NOTE] 
> The `graphql.operation.name` value is provided by the client and can have high cardinality. Using it in GraphQL server span name (by default) is NOT RECOMMENDED.
>
> Instrumentation MAY provide a configuration option to enable more descriptive span name following `{graphql.operation.type} {graphql.operation.name}` format when `graphql.operation.name` is available.

lmolkova: BTW, we use `{}` for templates instead of `<>`  in other documents.

kaylareopelle: I like those suggestions! How's this? 

```suggestion
The **span name** SHOULD be of the format `{graphql.operation.type}` provided
`graphql.operation.type` is available. If `graphql.operation.type` is not available, the
span SHOULD be named `GraphQL Operation`.

> **Warning**
> The `graphql.operation.name` value is provided by the client and can have high cardinality. Using it in the GraphQL server span name (by default) is NOT RECOMMENDED.
>
> Instrumentation MAY provide a configuration option to enable a more descriptive span name following `{graphql.operation.type} {graphql.operation.name}` format when `graphql.operation.name` is available.
```

lmolkova: thanks! looks good!

lmolkova (approved):
Thank you!

kaylareopelle:
It seems like the wayback machine is down at the moment, causing the markdown-link-check job to fail: https://github.com/open-telemetry/semantic-conventions/actions/runs/11263374842/job/31321102055?pr=1389#step:4:10

joaopgrassi (approved):

PR #1393, opened 2024-09-05
Add genai system-specific conventions for Azure AI Inference

lmolkova:
Adds 
- `az.ai.inference` system name.
- md file describing how generic conventions apply to Azure AI Inference

It's being implemented in https://github.com/Azure/azure-sdk-for-js/pull/30800 and https://github.com/Azure/azure-sdk-for-python/pull/36890, other languages to follow


model/trace/gen-ai.yaml
```diff
@@ -1,13 +1,12 @@
...
     type: span
     brief: >
       Describes GenAI operation span.
     attributes:
-      - ref: gen_ai.system
-        requirement_level: required
       - ref: gen_ai.request.model
-        requirement_level: required
+        requirement_level:
+          conditionally_required: If available.
```
lmolkova, line 9: Azure AI Inference allows to use specific model-as-a-service and does not always have a model name at request time.

Client interacts with the endpoint, but does not need to know the model. 

So, changing level to `conditionally_required` to accommodate this scenario. Other system-specific conventions may override it and say it's required

model/registry/gen-ai.yaml
```diff
@@ -25,6 +25,10 @@ groups:
               stability: experimental
               value: "cohere"
               brief: 'Cohere'
+            - id: az.ai_inference
```
lmolkova, line 28: TBD

docs/gen-ai/azure-ai-inference.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Semantic Conventions for Azure AI Inference
+
+**Status**: [Experimental][DocumentStatus]
+
+The Semantic Conventions for [Azure AI Inference](https://learn.microsoft.com/azure/ai-studio) extend and override the [GenAI Semantic Conventions](README.md).
+
+## Azure AI Inference
+
+`gen_ai.system` MUST be set to `"az.ai.inference"` and SHOULD be provided **at span creation time**.
```
lmolkova, line 13: @truptiparkar7 could you please bless this name? Thanks!

model/trace/gen-ai.yaml
drewby: Interesting. It used to be that the request would use the "deployment name" which then would have an underlying model configured in the service. Is that still the case?

docs/gen-ai/azure-ai-inference.md
```diff
@@ -0,0 +1,83 @@
...
+| [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [2] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` if the operation ended in an error | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.request.model`](/docs/attributes-registry/gen-ai.md) | string | The name of the GenAI model a request is being made to. [3] | `gpt-4` | `Conditionally Required` If available. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`server.port`](/docs/attributes-registry/server.md) | int | GenAI server port. [4] | `80`; `8080`; `443` | `Conditionally Required` If not default (443). | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`az.namespace`](/docs/attributes-registry/azure.md) | string | [Azure Resource Provider Namespace](https://learn.microsoft.com/azure/azure-resource-manager/management/azure-services-resource-providers) as recognized by the client. [5] | `Microsoft.CognitiveServices` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
drewby, line 30: IMO it's a bit difficult to parse what is specific to the vendor conventions and what is repeated from the general conventions. I have the same concern with the openai conventions. Is there a way to highlight, or sort, or otherwise call out what is the additional attributes?

model/trace/gen-ai.yaml
lmolkova: if you use [serverless models](https://learn.microsoft.com/en-us/azure/ai-studio/how-to/deploy-models-serverless?tabs=azure-ai-studio), after you deploy it, the only thing the client knows is the endpoint  🤷‍♀️

```js
const client = ModelClient("https://contoso.eastus2.inference.ai.azure.com", credential);
  const response = await client.path("/chat/completions").post({
    body: {
      messages: [
        ...
      ],
    }
  });
```

docs/gen-ai/azure-ai-inference.md
lmolkova: no way to highlight now, but it's a good feature to have.

My assumption that users who use some specific vendor would come to the corresponding page and learn what this vendor reports (doesn't matter if it's generic or not).

I'm curious to learn why you need to find what's different. An easy way to to go read yaml which includes all customizations:

https://github.com/open-telemetry/semantic-conventions/blob/c0f29fff3085b96f0be13f082388f15646b7b1e3/model/trace/gen-ai.yaml#L97-L116

docs/gen-ai/azure-ai-inference.md
TaoChenOSU: How about `azureai`?

docs/gen-ai/azure-ai-inference.md
```diff
@@ -0,0 +1,83 @@
...
+|---|---|---|
+| `_OTHER` | A fallback error value to be used when the instrumentation doesn't define a custom value. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+
+
+`gen_ai.operation.name` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `chat` | Chat completion operation such as [OpenAI Chat API](https://platform.openai.com/docs/api-reference/chat) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `text_completion` | Text completions operation such as [OpenAI Completions API (Legacy)](https://platform.openai.com/docs/api-reference/completions) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
TaoChenOSU, line 74: More of a general discussion: I have been thinking if we should have dedicated names for streaming. I found it useful to be able to tell if the telemetry comes from a streaming API or not in my experience.

docs/gen-ai/azure-ai-inference.md
lmolkova: we had some internal discussions on this and it seems we'll have other Azure AI things, so we need a way to distinguish. 

Also all the libs are called `azure-ai-inference` - https://learn.microsoft.com/azure/ai-studio/reference/reference-model-inference-api

So the motivation for this name was:
1. We have other services that will report `*.system` attributes. E.g. `az.search` could be reported in `db.system`. I.e. `az` or `azure` is a namespace
2. We already have `az.` attributes that are stable from Azure SDK side, so it should be `az`
3. we expect other things in `az.ai`, so we need `az.ai.inference` for this one

I'm totally open to suggestions from the product team.

truptiparkar7: Yes I think it will be good to be consistent with the libraries this will be used with. So the name looks good to me.

truptiparkar7: Overall this PR looks great and will be very helpful for us.

truptiparkar7 (approved):
This PR looks great, is aligned with our product strategy and is very helpful for us. Thanks!

docs/gen-ai/azure-ai-inference.md
lmolkova: Yep, we definitely need to add it - https://github.com/open-telemetry/semantic-conventions/issues/1170

I've been thinking we can add an attribute showing time-to-first token on the span which could be used as an indication of a streaming call or something else more informative than `is_streaming: true`. Let me know if you have any thoughts!

docs/gen-ai/gen-ai-metrics.md
lmolkova, line 1: request.model is still required on metrics

docs/gen-ai/gen-ai-spans.md
```diff
@@ -114,6 +114,7 @@ Instrumentations SHOULD document the list of errors they report.
 | Value  | Description | Stability |
 |---|---|---|
 | `anthropic` | Anthropic | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `az.ai.inference` | Azure AI Inference | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
codefromthecrypt, line 117: nit: is it possible to use lower snake case separation?

codefromthecrypt (commented):
not used azure much, but is deployment a primary attribute? 

(online answer: deployment is the same as the model name)

codefromthecrypt (approved):
from openai standpoint sound like the model is optional, and this is a good example of that. better to clarify indeed.

docs/gen-ai/gen-ai-spans.md
codefromthecrypt: understand there may be future ai products so this isn't valid to snakecase

docs/gen-ai/azure-ai-inference.md
TaoChenOSU: I never thought of time-to-first token. It may be a good idea and may carry other useful information implicitly. The benefit of a more straight-forward one is less confusion but also lower entropy.

TaoChenOSU: Will comment on this issue: https://github.com/open-telemetry/semantic-conventions/issues/1170

TaoChenOSU (approved):

docs/gen-ai/openai.md
```diff
@@ -11,7 +11,7 @@ linkTitle: OpenAI traces and metrics
 <!-- toc -->
 
 - [OpenAI Span attributes](#openai-span-attributes)
-- [OpenAI Metric attributes](#openai-metric-attributes)
+- [OpenAI Metrics](#openai-metrics)
```
drewby, line 14: Minor: Why the difference in titling between span and metrics? One includes the word `attributes`, the other does not?

drewby (approved):

docs/gen-ai/openai.md
lmolkova: good catch!

joaopgrassi (approved):

lzchen (approved):

Issue #929, opened 2024-04-17
Clarify "type" in Resource semantic conventions


lmolkova:
Resources have a "type" property (manually written, not part of yaml). E.g. 

https://github.com/open-telemetry/semantic-conventions/blob/4728f6374361ad82949eecdfccb67759f9f2535c/docs/resource/README.md?plain=1#L74-L78

It's not clear what it means - so we should either document it or remove it.

_Originally posted by @lmolkova in https://github.com/open-telemetry/semantic-conventions/pull/919#discussion_r1562678752_

lmolkova:
See the details on type in https://github.com/open-telemetry/oteps/pull/264. We should yamlify it.

lmolkova:
It was documented in https://github.com/open-telemetry/semantic-conventions/blob/main/docs/how-to-write-conventions/resource-and-entities.md originally added in https://github.com/open-telemetry/semantic-conventions/pull/2328, closing this issue as completed

PR #1400, opened 2024-09-10
Use YAML descriptions for Resources

### LINKED ISSUE #1399: Resource Semantic Convention Stabilization Issues
### Area(s)

area:other

## Description

Resource semantic convention stabilization is not in the model.  If you look at https://github.com/open-telemetry/semantic-conventions/blob/main/model/resource/service.yaml - It is not marked as stable.

However, if you look in the markdown: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/README.md#service it is marked as stable.

Additionally, we have an odd section added that somehow stabilizes version attributes without defining them: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/README.md#version-attributes

Given the progress in the Entities WG - I'd like to propose the following:

- We stabilize [`service`](https://github.com/open-telemetry/semantic-conventions/blob/main/model/resource/service.yaml) group as is.
- We move [`telemetry`](https://github.com/open-telemetry/semantic-conventions/blob/main/model/resource/telemetry.yaml) resource group to `telemetry.sdk` and stabilize the YAML group.
- Any actions/updates performed by Entities + Resource WG will need to pull along these stable groups.
- An updated to backwards compatibility requirements to guarantee stable attribtues are *not* removed from these stable resource groups.

Give the state of https://github.com/open-telemetry/oteps/pull/264 and direction of the group - I think we can agree we already have `telemetry.sdk` and `service` as stabilized "entities" or "resource groups".  What we do NOT have is programatic access to that stability for rule enforcement, break detection, etc.  Let's fix that.

v1.27.0

### Additional context

_No response_

jsuereth:
Fixes #1399

Sending for feedback on approach.

## Changes

- Update service + telemetry.sdk YAML to officially mark themselves as stable.
- Update snippet template to generate FULL resource semconv template, forcing all markdown to by in sync
- Add `name` to resource groups that denote the `type` field previously in markdown only.


joaopgrassi (approved):

model/resource/faas.yaml
```diff
@@ -1,6 +1,7 @@
 groups:
   - id: faas_resource
     type: resource
+    name: faas
```
tigrannajaryan, line 4: Do we need `id` to be different from `name` in this case? I was wondering if we could simplify and keep just the `name`.

docs/resource/README.md
```diff
@@ -73,19 +73,20 @@ as specified in the [Resource SDK specification](https://github.com/open-telemet
...
 <!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
 <!-- see templates/registry/markdown/snippet.md.j2 -->
 <!-- prettier-ignore-start -->
 <!-- markdownlint-capture -->
 <!-- markdownlint-disable -->
 
+
+**Status:** ![Stable](https://img.shields.io/badge/-stable-lightgreen)
+
+**type:** `service`
```
lmolkova, line 86: thanks to your otep I finally understand what it is, I almost tried to remove it - https://github.com/open-telemetry/semantic-conventions/issues/929

model/resource/process.yaml
```diff
@@ -1,6 +1,7 @@
 groups:
   - id: process
     type: resource
+    name: process
```
lmolkova, line 4: it's unfortunate that yaml type is `resource` and proto resource type is `name`.

How much would you hate `resource_type` instead of the `name` ?

I know we want a common identity property on all groups, but `name` is problematic on spans anyway

policies/compatibility.rego
```diff
@@ -331,6 +340,71 @@ deny contains back_comp_violation(description, group_id, "") if {
 }
 
 
+# Rule: Detect Removed Resources
+#
+# This rule checks for stable resources that existed in the baseline registry
+# but are no longer present in the current registry. Removing resoruces
```
lmolkova, line 346:
```suggestion
# but are no longer present in the current registry. Removing resources
```

docs/resource/k8s.md
```diff
@@ -93,17 +99,20 @@ conflict.
...
 <!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
 <!-- see templates/registry/markdown/snippet.md.j2 -->
 <!-- prettier-ignore-start -->
 <!-- markdownlint-capture -->
 <!-- markdownlint-disable -->
 
+
+**Status:** ![Experimental](https://img.shields.io/badge/-experimental-blue)
+
+**type:** `none`
```
lmolkova, line 112: oops?

model/resource/service.yaml
```diff
@@ -1,8 +1,10 @@
 groups:
   - id: service
     type: resource
+    name: 'service'
```
lmolkova, line 4: super-nit: can we use/not use quotes consistently ?

```suggestion
    name: service
```

policies/compatibility.rego
```diff
@@ -331,6 +340,71 @@ deny contains back_comp_violation(description, group_id, "") if {
...
+# is considered a backward compatibility violation.
+#
+# In other words, we do not allow the removal of a resource once added
+# to the registry. It must exist SOMEWHERE, but may be deprecated.
+deny contains back_comp_violation(description, group_id, "") if {
+    # Find data we need to enforce
+    some resource in baseline_resources
+    resource.stability == "stable"
```
lmolkova, line 354: we don't remove experimental attributes or metrics - should we allow removing experimental resources?

model/resource/service_experimental.yaml
```diff
@@ -1,6 +1,7 @@
 groups:
   - id: service_experimental
     type: resource
+    name: 'service'
```
lmolkova, line 4: it seems that resource name is not unique (e.g. we have the same name for experimental and stable). But we use it such for back-compat checks.

Would it create any problems when merging resources?

policies/yaml_schema.rego
```diff
@@ -48,6 +48,26 @@ deny[yaml_schema_violation(description, group.id, name)] {
     description := sprintf("Event name '%s' is invalid. Event name %s'", [name, invalid_name_helper])
 }
 
+# require resources have names
+deny[yaml_schema_violation(description, group.id, "")] {
+    group := input.groups[_]
+    group.type == "resource"
+    not group.name
+    description := sprintf("Resource id '%s' is invalid.  Resource must have name.", [group.id])
```
lmolkova, line 56:
```suggestion
    description := sprintf("Resource id '%s' is invalid. Resource must have name.", [group.id])
```

model/resource/process.yaml
jsuereth: There's a possibility we rename this to `entity` so I'd like to avoid major refactorings right now.

policies/compatibility.rego
jsuereth: We haven't made guarantees on resource yet.  I think we COULD prevent removal, but I took a "make progress" stance on this PR.  At a minimum, let's protect what we stabilized.  Going forward, we need to understand what constitutes breaking for resource.

model/resource/service_experimental.yaml
jsuereth: backcompat is only checking stable resources against each other.

Right now, you can/should only move stable attribues from experimental -> stable group, so it should be ok.

I agree that there's some oddities here, but I think we can make progress on better resource/entity modelling in parallel.  For now, let's protect the bits we stabilized.

docs/resource/k8s.md
jsuereth: fixed, policy check needed to updated.

lmolkova (approved):

templates/registry/markdown/resource_macros.j2
```diff
@@ -0,0 +1,9 @@
+{#- Macros for simplifying creating "Resource" documentation. -#}
+{% import 'stability.j2' as stability %}
+{% macro header(resource) %}
+**Status:** {{ stability.badge(resource.stability, resource.deprecated) }}
+
+**type:** `{{ resource.name }}`
```
lquerel, line 6: Maybe we should use `display_name` and fall back to `name` if the display name doesn’t exist. I'm not sure.

This also leads me to think that sometimes we have a `display_name` without the definition of `name` in the semconv (e.g. `registry/container.yaml`). Perhaps we should consider adding this kind of rule in the future.

lquerel (approved):

templates/registry/markdown/resource_macros.j2
jsuereth: I use `name` because it must be unique here.  Type, eventually, will be used formally in merge rules, see: https://github.com/open-telemetry/oteps/pull/264

ChrsMark (approved):

AlexanderWert (approved):

Issue #1407, opened 2024-09-16
Add system.device attribute to system paging metrics


rogercoll:
### Area(s)

area:system

### Is your change request related to a problem? Please describe.

Usage paging system metrics refer to the `system.paging.state` attribute but not the actual paging device. On the other hand, the hostmetrics receiver is currently adding the `device` attribute to `system.paging.usage` and `system.paging.utilization` metrics: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/pagingscraper/metadata.yaml#L8

### Describe the solution you'd like

Link usage paging metrics to the `system.device` attribute.

Issue #722, opened 2024-02-08
Why is adding new attributes to a metric a breaking change?


alanwest:
Is there a clear rationale documented some where for why adding new attributes to a metric is a breaking change?

[This PR comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2180#pullrequestreview-820513856) is one of the earliest assertions that I'm aware of that adding attributes is breaking, but it has remained unclear to me whether folks have a shared understanding of why.

> ADDITIVE changes to metrics can break user alerts by creating new timeseries and adjusting thresholds. I think we need to be a lot more careful around changes, particularly around metrics here.

I do not personally have experience with a backend where this is true, but if it is true for some backend, I think it would be useful to articulate this some where.

trask:
I agree it would be great to document the rationale since this has been surprising to many people.

In the meantime, I think the best we have is @jsuereth's document
https://docs.google.com/document/d/1Nvcf1wio7nDUVcrXxVUN_f8MNmcs0OzVAZLvlth1lYY/edit#heading=h.58b8u7y5tsmu, see the "Instability: Alerting w/ Metric Attribute changes" section

pyohannes:
It would be great to clarify that. If we would apply that rule strictly, wouldn't we need to say that instrumentations MUST NOT add any additional attributes to metrics that are not covered by the specification?

For example, if [ASP.NET Core 8 adds an additional attribute to `http.server.request.duration`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/dotnet/dotnet-http-metrics.md#metric-httpserverrequestduration), wouldn't that be breaking in the same sense, as it adds a new time series?

> * Additional attributes:
>
>   * The `aspnetcore.request.is_unhandled` boolean attribute is reported when the request was not handled by the application pipeline. It's skipped otherwise.

pyohannes:
That's the exact wording from the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#semantic-conventions-stability):

> In addition to the 3 types of changes described above there are certain types that are always allowed. Such changes do not need to be described (and are not described) by schema files. Here is the list of such changes:
> [...]
> * Adding new attributes to existing metrics that do not "break apart" existing timeseries, such that alert thresholds would break / need to change.

lmolkova:
@pyohannes 
[...]

PR #1408, opened 2024-09-16
Add system.device attribute to system paging metrics

rogercoll:
Fixes  https://github.com/open-telemetry/semantic-conventions/issues/1407

## Changes

Reference `system.device` in system paging usage metrics.


mx-psi (approved):

ChrsMark (approved):

rogercoll:
CI will be fixed with https://github.com/open-telemetry/semantic-conventions/pull/1412

docs/system/system-metrics.md
```diff
@@ -514,6 +514,7 @@ This metric is [recommended][MetricRecommended].
 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
+| [`system.device`](/docs/attributes-registry/system.md) | string | Unique identifier for the device responsible for managing paging operations. | `/dev/dm-0` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 517: It looks like duplicate of `device.id` attribute - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/device.md

model/system/metrics.yaml
```diff
@@ -110,6 +110,9 @@ groups:
     unit: "By"
     attributes:
       - ref: system.paging.state
+      - ref: system.device
```
lmolkova, line 113:
```suggestion
      - ref: device.id
```

docs/system/system-metrics.md
lmolkova: /cc @open-telemetry/semconv-mobile-approvers - is  there any reason not to reuse `device` namespace for non-mobile devices?

LikeTheSalad: It does look a lot like `device.id`. I don't see any reason why it couldn't be reused for non-mobile devices, although I just noticed that there's a note [here [1]](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/device.md) mentioning mobile specific rules around `device.id` for some reason, which I guess can be interpreted as it being a mobile-only attr. I'm up for reusing attrs as much as possible and I don't see an issue with reusing `device.id`, although we should probably update its note to avoid mentioning mobile-specific platforms (unless there is an important reason constraint it to mobile only, which I'm not aware of).

rogercoll: Makes sense to me, we should avoid having duplicated attributes. I had a quick check, and it looks like many system metrics are currently using the `system.device` attribute, opened this issue to track the replacement; https://github.com/open-telemetry/semantic-conventions/issues/1432

ChrsMark: @open-telemetry/semconv-system-approvers wdyt?

TBH it looks really weird to me mixing these 2. How a value of `/dev/dm-0` can be represented by the same attribute that  represents the mobile device `2ab2916d-a51f-4ac8-80ee-45ac31a28092`?
If I understand it correctly the `system.device` is used for network, storage, filesystem etc devices that are part of a wider system (i.e.), a host. For example I plug an extra network interface to a physical host and I get an additional `system.device` with its own traffic metrics.

But how this can be mixed with mobile devices? 

Additionally, on entity level that's also wrong: a mobile device can plug external storage devices like `/dev/disk0` so in that case what is the `device.id`? We can find lot's of such inconsistencies specially on Android which is linux based I guess.

LikeTheSalad: > How a value of `/dev/dm-0` can be represented by the same attribute that represents the mobile device `2ab2916d-a51f-4ac8-80ee-45ac31a28092`?

Viewing it from that perspective, it seems like the term `device` might be overloaded then. The idea I had in mind as "device" was as an independent "machine", e.g. mobile phone, laptop, watch, etc, that produces telemetry. However, if we're talking about the notion of identifying mounted devices in a host machine then it does sound quite strange to mix them up.

lmolkova: We have hardware attributes that would cover hardware components - 

https://github.com/open-telemetry/semantic-conventions/blob/b780f480e73ffc2d1da117689475b7b9c8af610b/model/hardware/registry.yaml#L7-L19

could we reuse them here?

ChrsMark: That won't work either I think. We could have cases in virtual machines where we have virtual network interfaces, disks etc. I believe we don't need to re-use anything here, `system.device` is quite accurate and represents how a system sees and handles any kind of device (systemically).

lmolkova: thanks for the explanation! I'm fine with a new attribute if there is a consensus within @open-telemetry/semconv-system-approvers that's necessary,

I'd still suggest to call it `system.device.id` or `system.device.name` to be specific and have a room to add other properties.

model/system/metrics.yaml
lmolkova: or 

```suggestion
      - ref: system.device.id|name
```

docs/system/system-metrics.md
joaopgrassi: Hum but `system.device` is not a new attribute, this PR simply references to it: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/system.md#general-system-attributes.

I guess it makes sense to add it then first, and later as a followup, deprecate it in favor of `system.device.id|name`.

lmolkova: Oh, you're right

.chloggen/add_paging_device_attr.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
```
lmolkova, line 7: nit:

```suggestion
change_type: 'breaking'
```

Adding a new  metric attribute is a breaking change - https://github.com/open-telemetry/semantic-conventions/issues/722, which is fine on experimental metrics, we just need to document it accordingly

lmolkova (approved):
LGTM!

(And sorry for the confusion - I created https://github.com/open-telemetry/semantic-conventions/issues/1474 to follow up on the `system.device`)

ChrsMark (approved):

PR #1414, opened 2024-09-17
Add paragraph about usage of parent-child relationship

### LINKED ISSUE #1282: Messaging: explain that parent could be used in addition to link
Extracted from https://github.com/open-telemetry/semantic-conventions/issues/1205

- add spec text allowing consumer span to parent to message
- add parenting to examples
- maybe explain why link as necessary

joaopgrassi:
Fixes #1282

## Changes

Adds a paragraph mentioning the possibility to use the message creation context as a parent of "Process" spans, to achieve a more traditional parent-child trace structure.


jsuereth (approved):

docs/messaging/messaging-spans.md
```diff
@@ -261,6 +261,23 @@ batch of messages, or for no message at all (if it is signalled that no
...
 "Receive" span SHOULD link to the message's creation context.
 
+> [!IMPORTANT]
+> These conventions use spans links as the default mechanism to correlate
+> producers and consumer(s) because:
+>
+> - It is the only consistent trace structure that can be guaranteed,
+> given the many different messaging systems models available.
+>
+> - It is the only option to correlate produce and consumer(s) in batch scenarios
```
lmolkova, line 271: nit

```suggestion
> - It is the only option to correlate producer and consumer(s) in batch scenarios
```

docs/messaging/messaging-spans.md
```diff
@@ -261,6 +261,23 @@ batch of messages, or for no message at all (if it is signalled that no
...
+> as a span can only have a single parent.
+>
+> - Offers a consistent experience for users analysing traces from different
```
lmolkova, line 274: It seems very similar to the first point, no?

> - It is the only consistent trace structure that can be guaranteed, given the many different messaging systems models available.

docs/messaging/messaging-spans.md
```diff
@@ -261,6 +261,23 @@ batch of messages, or for no message at all (if it is signalled that no
...
+> messaging systems.
+
+Exclusively for single messages scenarios, "Process" or "Receive" span MAY
```
lmolkova, line 277:
```suggestion
Exclusively for single messages scenarios, "Process" span MAY
```

Receive spans can't use message context as a parent because they start before message is received and the context is known.

docs/messaging/messaging-spans.md
```diff
@@ -261,6 +261,23 @@ batch of messages, or for no message at all (if it is signalled that no
...
+>
+> - It is the only consistent trace structure that can be guaranteed,
+> given the many different messaging systems models available.
+>
+> - It is the only option to correlate produce and consumer(s) in batch scenarios
+> as a span can only have a single parent.
+>
+> - Offers a consistent experience for users analysing traces from different
+> messaging systems.
+
```
lmolkova, line 276:
```suggestion
> - It is the only option to correlate produce and consumer(s) when message consumption can happen in the scope of another ambient context such as HTTP server span.
```

docs/messaging/messaging-spans.md
```diff
@@ -261,6 +261,23 @@ batch of messages, or for no message at all (if it is signalled that no
...
+Exclusively for single messages scenarios, "Process" or "Receive" span MAY
+use the message's creation context as its parent, thus achieving a direct
+parent-child relationship between producer and consumer(s).
```
lmolkova, line 279: I'm suggesting to add a bit of additional guidance around the ambient context and defaults.

```suggestion
parent-child relationship between producer and consumer(s). Instrumentations SHOULD document whether they use message creation context as a parent for "Process" spans and MAY provide configuration option allowing users to control this behavior. 

It's NOT RECOMMENDED to use message creation context as a parent for "Process" span (by default) if processing usually happens in the scope of another span.

If instrumentation uses message creation context as a parent on "Process" span in the scope of another valid ambient context, it SHOULD set the ambient context as a link on "Process" span to preserve correlation between message processing and that context. 

For example, a messaging broker pushes messages to consumers over HTTP and a consumer application has HTTP server and messaging instrumentations enabled. The messaging instrumentation would create "Process" span following one of these possible approaches:

- "Process" span is a child of HTTP server span context and has a link to message creation context. This is the default behavior for this instrumentation.
- "Process" span is a child of message creation context and has two links: one to message creation context and another one to HTTP server span context. This is opt-in behavior.
```

LMK what you think. 

Maybe we can also put it all into a new section like "Using message context as a parent for "Process" spans" so it appears in the TOC?

lmolkova (approved):
LGTM, suggesting to add a bit more guidance for parent case

docs/messaging/messaging-spans.md
joaopgrassi: Ah yeah of course!

docs/messaging/messaging-spans.md
joaopgrassi: Yeah good point. I tried to merge them together but didn't sound good. I will just remove it then :)

docs/messaging/messaging-spans.md
joaopgrassi: I was thinking the same, but to add new headers for `Receive` spans and `Process` inside the `Consumer spans` section. Right now they are all together and I find it a bit hard to follow.

lmolkova: Thank you! I like "Message creation context as parent of "Process" span"!

lmolkova (approved):

PR #1417, opened 2024-09-18
Escape square brackets that don’t represent a link

lquerel:
Square brackets are often used to specify a range in metrics. However, without escaping them, Markdown-based documentation interprets these as links. This PR addresses the issue by escaping such square brackets, ensuring they are displayed correctly as plain text. 

This change is similar to [PR #1416](https://github.com/open-telemetry/semantic-conventions/pull/1416).

### DIFF
--- a/model/go/metrics.yaml
+++ b/model/go/metrics.yaml
@@ -89,7 +89,7 @@ groups:
     metric_name: go.config.gogc
     brief: "Heap size target percentage configured by the user, otherwise 100."
     note: >
-      The value range is [0.0,100.0].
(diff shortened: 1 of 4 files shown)


lquerel:
@lmolkova pointed out that this part of the markdown "spec" is ambiguous, and most markdown linters accept these links (except for example `cargo doc`). Therefore, we decided to create a new configuration in Weaver to automatically correct these links when used in the context of code generation, such as in the Rust codegen. See this Weaver issue for more details https://github.com/open-telemetry/weaver/issues/374

PR #1422, opened 2024-09-20
Change messaging.operation.type=publish to send

### LINKED ISSUE #1285: Messaging: `publish` operation type is pub/sub/topic specific
Extracted from https://github.com/open-telemetry/semantic-conventions/issues/1205

`messaging.operation.type = publish` is specific to pub/sub/topic.

`send` would sound more neutral and apply to pub/sub and MQ. E.g. GCP pub/sub, Azure EventHubs and Kafka use `send`.

We should consider changing operation type and also metric name to use `send`.

### LINKED ISSUE #2185: Simpler events for recording (local) GenAI function tool calling as generalized function tracing events
### Area(s)

area:gen-ai

### What's missing?

Although there are existing events for tool calling (`gen_ai.assistant.message`, `gen_ai.tool.message`), these are more geared towards the Gen AI model's view of what tools should be invoked/called (rather than from the standpoint of instrumenting the tool that ultimately got invoked). Additionally, these are somewhat complicated to generate and may not always be available (if the tool calling is deeply nested inside the logic of the client).

Where the model invokes local tools, it would be useful to be able to implement recording the input/output of those tool calls by intercepting the tool registration and having a somewhat simpler interface for tracing those calls. I also think that some of this modelling could be more general than just Gen AI.

I propose that we have a more generalized representation of function tracing to events that could be used.

I'd like for there to be a more general set of events associated with function tracing:

## Function Start

Represents a function being invoked and provides a way to record the parameters:

  - Attributes:
    - `event.name`: "function_start"
    - `code.function.name`: the name of the function that was called
    - `positional_argument_count`: the count of positional args
    - `keyword_argument_count`: the count of keyword or optional arguments
   - Body:
     -  `positional_arguments`: an array of the positional arguments
     -  `keyword_arguments`: a dictionary of the keyword arguments

## Function End

Represents the result of a function being invoked.

  - Attributes:
    - `event.name`: "function_end"
    - `code.function.name`: the name of the function that was called
    - `error.type`: if there was an error
   - Body:
     -  `result`: JSON representation of the (succesful) result
     -  `error`: JSON representation of the (error) result

mamwl1:
Fixes #1285 

## Changes

Change messaging.operation.type=publish to send, as well as the related messaging.client.published.messages metric. Adjust text to use send where appropriate. Update example to use send where appropriate.


mamwl1:
Closing while CLA is being sorted which might take some time.

docs/messaging/messaging-spans.md
```diff
@@ -532,19 +534,19 @@ flowchart LR;
 | `messaging.system` | `"rabbitmq"` | `"rabbitmq"` | `"rabbitmq"` |
 | `messaging.destination.name` | `"T"` | `"T"` | `"T"` |
 | `messaging.operation.name` | `"publish"` | `"consume"` | `"consume"` |
-| `messaging.operation.type` | `"publish"` | `"process"` | `"process"` |
+| `messaging.operation.type` | `"send"` | `"process"` | `"process"` |
 | `messaging.message.id` | `"a"` | `"a"`| `"a"` |
```
joaopgrassi, line 539: In the section "Topic with multiple consumers" we say

> Given is a publisher that publishes a message to a topic exchange

I guess we need to change that to

> Given is a producer that sends a message to a topic exchange

joaopgrassi: Also the example table below it is still using "Publish" terminology
https://github.com/mamwl1/semantic-conventions/blob/mamwl1_branch/docs/messaging/messaging-spans.md#topic-with-multiple-consumers

docs/messaging/messaging-spans.md
```diff
@@ -558,26 +560,26 @@ flowchart LR;
...
 | `server.port` | `1234` | `1234` | `1234` |
 | `messaging.system` | `"kafka"` | `"kafka"` | `"kafka"` |
 | `messaging.destination.name` | `"Q"` | `"Q"` | `"Q"` |
 | `messaging.operation.name` | `"send"` | `"send"` | `"poll"` |
-| `messaging.operation.type` | `"publish"` | `"publish"` | `"receive"` |
+| `messaging.operation.type` | `"send"` | `"send"` | `"receive"` |
 | `messaging.message.id` | `"a1"` | `"a2"` | |
 | `messaging.batch.message_count` |  |  | 2 |
 
 ### Batch publishing
```
joaopgrassi, line 580:
```suggestion
### Batch sending
```

We also want to change here, right? The mermaid diagram + table are also using "publish" still.

joaopgrassi (changes requested):
Thank you! I think there are still some "publish" left-overs we need to tackle.

docs/messaging/kafka.md
```diff
@@ -113,10 +113,10 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
 | `process` | One or more messages are processed by a consumer. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `publish` | One or more messages are provided for publishing to an intermediary. If a single message is published, the context of the "Publish" span can be used as the creation context and no "Create" span needs to be created. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `receive` | One or more messages are requested by a consumer. This operation refers to pull-based scenarios, where consumers explicitly call methods of messaging SDKs to receive messages. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `send` | One or more messages are provided for sending to an intermediary. If a single message is sent, the context of the "Send" span can be used as the creation context and no "Create" span needs to be created. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `settle` | One or more messages are settled. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
mamwl1: I deliberately left a number of publishes in. A messaging.operation.type of "send" in span data allows us to have a more generic and inclusive approach that applies to more messaging products. However when describing scenarios with a specific messaging product we should be using the terminology that works best for that product, and the specific scenario. And for Kafka that is publish. NB that while the Kafka docs always talk about publishing to a topic, the actual Kafka API for that is "send", as shown in the table for the Quarkus example. 

For IBM MQ if pub/sub was used we would talk about publishing to a topic, and if point to point was used we would talk about sending to a queue. But in both cases we would use a messaging.operation.type of "send".

Does that make sense?

docs/messaging/messaging-spans.md
```diff
@@ -185,6 +186,7 @@ If a corresponding `{destination}` value is not available for a specific operati
 Examples:
 
 * `publish shop.orders`
```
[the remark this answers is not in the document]
mamwl1: Deliberately left as the publish here comes from messaging.operation.name which is messaging product specific.

docs/messaging/messaging-spans.md
mamwl1: Same comment as for Kafka. messaging.operation.name of publish is appropriate for RabbitMQ.

docs/messaging/gcp-pubsub.md
```diff
@@ -151,7 +151,7 @@ flowchart LR;
 | `messaging.batch.message_count` |  |  | 2 |
 | `messaging.destination.name` | `"T"` | `"T"` | `"T"` |
 | `messaging.operation.name` | `"create"` | `"create"` | `"send"` |
-| `messaging.operation.type` | `"create"` | `"create"` | `"publish"` |
+| `messaging.operation.type` | `"create"` | `"create"` | `"send"` |
 | `messaging.message.id` | `"a1"` | `"a2"` | |
 | `messaging.message.envelope.size` | `1` | `1` | |
 | `messaging.system` | `"gcp_pubsub"` | `"gcp_pubsub"` | `"gcp_pubsub"` |
```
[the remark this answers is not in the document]
mamwl1: Good catch.

docs/messaging/kafka.md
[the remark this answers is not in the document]
joaopgrassi: Yep, I was thinking that was the reason, I just wanted to confirm. All good!

docs/messaging/messaging-spans.md
joaopgrassi: Then should the diagram say `Span Publish A`? Because now it is confusing.. we say publish in the text + in the table but in the diagram is now Send.

docs/messaging/messaging-spans.md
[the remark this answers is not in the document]
joaopgrassi: Right. But then several Kafka examples are using `send` for `messaging.operation.name` also the mermaid diagrams say "Send"

docs/messaging/messaging-spans.md
mamwl1: Before I made any changes:
-the mermaid diagrams use the messaging.operation.type values, in the green boxes, and these values are also used in the table column names
-the descriptive text before the mermaid diagrams used values which are appropriate for the messaging product, and consistent with the span name and messaging.operation.name rows

I tried to stick with that approach when making the change from messaging.operation.type = publish -> send.
It looks like I didn't get this right for the second Kafka example as publish should be send, in the diagram and the column header. I have added that change.

I agree that there is a bit of a disconnect between the different verbs being used. But I think that is inevitable when dealing with many different products. At least we can try and get the approach used across all the examples consistent. 

It also looks like the example at the end of https://opentelemetry.io/docs/specs/semconv/messaging/kafka/ isn't consistent with the approach on the main messaging spans page. I can change that if you agree that the approach above is what we should be using.

And it looks like the "Links" row in the tables are also inconsistent. Sometimes the span name is used, sometimes the column name. Looks like the column name is the one that is generally used, so I can change the values used in https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#topic-with-multiple-consumers if you are happy with that.

joaopgrassi: Makes sense. We discussed this in the sig meeting and we will see if we need to address this later, in a follow up to not block this. Sorry for may be delaying this.

joaopgrassi:
Heads up: If https://github.com/open-telemetry/semantic-conventions/pull/1446 merges first, we will need to adapt the new example introduced there.

PR #1423, opened 2024-09-20
Merge experimental and stable attributes per resource type.

jsuereth:
Merge the experimental and stable portions of `resource` together where previously we had them split by group but still had them with the same `type`.

Marking this draft to get review on direction.

My own concern is around whether the notion of "stable resource type" but "unstable attributes" is clear or needs called out somewhere else.

## Changes

- Merge `service` experimental + stable attributes in the same group.
- Merge `telemetry.sdk` experimental + stable attributes in the same group.
- Modify markdown so that both stable + experimental attributes show up joined for a given resource type.


joaopgrassi (approved):

docs/resource/README.md
```diff
@@ -90,40 +88,13 @@ as specified in the [Resource SDK specification](https://github.com/open-telemet
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
 | [`service.name`](/docs/attributes-registry/service.md) | string | Logical name of the service. [1] | `shoppingcart` | `Required` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`service.instance.id`](/docs/attributes-registry/service.md) | string | The string ID of the service instance. [2] | `627cc493-f310-47de-96bd-71410b7dec09` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 91: If we combine it with https://github.com/open-telemetry/semantic-conventions/pull/1472, we have a problem:

- stable instrumentations can't emit experimental attributes (by default). 
- they may have a feature-flag that allows to enable experimental parts

but it means that experimental attributes are opt-in.

I think @joaopgrassi recently created (https://github.com/open-telemetry/weaver/pull/392) a loophole we can use here:

```yaml
- ref: service.instance.id
   requirement_level:
     opt_in:  on stable instrumentations, `recommended` otherwise.
``` 

Let's chat more about it (and if we can do better than that) on the tooling call tomorrow

joaopgrassi: > but it means that experimental attributes are opt-in.

Perhaps it is just a lack of orientation/spec on our side to properly make this clear/obvious:

> stable instrumentations can't emit experimental attributes (by default)

I wouldn't indicate such thing via the requirement level. I feel this should be a "rule" in semconv that applies to everyone. Then having them in the same table is not a problem. 

Perhaps as a UX improvement:

- Sort stable attributes first
- Add an empty line in the table to somewhat separate stable vs experimental ones

lmolkova: Created https://github.com/open-telemetry/semantic-conventions/issues/1519 to track stabilization of `service.instance.id`

Since #1472 is merged, the policies won't be happy about the `Recommended` level here (but as we discussed, we'd add exceptions for existing things).

I agree with @joaopgrassi we could do a bunch of UX improvements on how things are rendered to visually separate experimental from stable attributes

lmolkova (approved):

docs/resource/README.md
```diff
@@ -219,9 +193,9 @@ All custom identifiers SHOULD be stable across different versions of an implemen
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
-## Telemetry Distribution (Experimental)
+## Telemetry SDK
```
lmolkova, line 196: this one should probably be 

```suggestion
## Telemetry Distribution
```

there is another "Telemetry SDK" on line 145

lmolkova (approved):
🚀

PR #1424, opened 2024-09-20
Add `db.response.status_code` attribute, deprecate `db.cosmos.status_code`

lmolkova:
Introducing status code for database call based on the discussion in DB Semconv.

Most databases support a notion of a status code (usually representing an error, but also used for warnings and sometimes to differentiate success results).

E.g. SQL standard defines [`SQLSTATE`](https://wikipedia.org/wiki/SQLSTATE) - a string code. It, for example, appears on the [`java.sql.SqlException#getSQLState`](https://docs.oracle.com/javase/8/docs/api/java/sql/SQLException.html).

Databases that run on top of HTTP/gRPC/etc may also support status codes from that protocol. For example, Cosmos DB can run on top of HTTP or RNTBD (proprietary protocol), but needs one status code for both.

HBase supports HTTP, Thrift and gRPC and may report different codes depending on the protocol.

------------------
 
We need a place to capture DB response code that's not protocol-specific and **could be reported on common metrics** for all databases.

`error.type` does not work at least in some cases
- Capturing exception type may give human-readable result (e.g. [SQLSyntaxErrorException](https://docs.oracle.com/javase/8/docs/api/java/sql/SQLSyntaxErrorException.html) instead od SQLSTATE `"42000"`, or [`InvalidPreparedStatementDefinition`](https://www.psycopg.org/docs/errors.html)  instead of `"42P14"`, but there are no guarantees that exception type maps 1:1 to error code. So we want to capture both.
- Status code does not necessarily indicate an error:
   - someone may want to differentiate successful responses (e.g.  206 - partial success vs 200 - full one) 
   - unsuccessful response codes don't necessarily indicate a failure (e.g. database may report HTTP 409 or gRPC `ALREADY_EXISTS` for `createIfNotExist` call and it's not an error, but also useful to know if resource already existed).


joaopgrassi (approved):

model/database/common.yaml
```diff
@@ -14,10 +15,18 @@ groups:
       - ref: server.port
         requirement_level:
           conditionally_required: If using a port other than the default port for this DBMS and if `server.address` is set.
+      - ref: db.response.status_code
+        requirement_level:
+          conditionally_required: If the operation failed.
```
trask, line 20: and if available? (or recommended?)

.chloggen/1421.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: breaking
+component: database
+note: Add new `db.response.status_code` attribute, deprecate `db.cosmos.status_code`.
```
jsuereth, line 3: Are we able to express this in schema migrations or not?

lmolkova: unfortunately no, it went from int to string

model/database/spans.yaml
```diff
@@ -132,8 +145,15 @@ groups:
       - ref: network.peer.port
         requirement_level:
           recommended: if and only if `network.peer.address` is set.
+      - ref: db.response.status_code
+        brief: >
+          [Cassandra protocol error code](https://github.com/apache/cassandra/blob/cassandra-5.0/doc/native_protocol_v5.spec). If
+
+          represented as a string.
```
trask, line 152: it's not clear to me what "if represented as a string" means here

model/database/spans.yaml
```diff
@@ -244,9 +282,16 @@ groups:
         requirement_level:
           conditionally_required: If available.
         note: ""  # overriding the base note
+      - ref: db.response.status_code
+        brief: >
+          [MongoDB error code](https://www.mongodb.com/docs/manual/reference/error-codes/) represented as a string.
+        requirement_level:
+          conditionally_required: If the operation failed and status code is available.
```
trask, line 289:
```suggestion
          conditionally_required: If the operation failed and error code is available.
```

model/database/spans.yaml
```diff
@@ -179,9 +205,16 @@ groups:
         requirement_level:
           conditionally_required: If available.
         note: ""  # overriding the base note
+      - ref: db.response.status_code
+        brief: >
+          The HTTP response code returned by the Couch DB.
+        examples: ["200", "201", "429"]
+        requirement_level:
+          conditionally_required: If response was received.
```
trask, line 213: not sure if it's generally available to instrumentation, e.g. https://docs.couchbase.com/sdk-api/couchbase-java-client/com/couchbase/client/java/Collection.html#get(java.lang.String)

```suggestion
          conditionally_required: If response was received and the HTTP response code is available.
```

model/database/spans.yaml
```diff
@@ -332,8 +383,53 @@ groups:
...
+          known vendor-specific error code. If only SQLSTATE is available,
+          it SHOULD be used.
+
+          When multiple error codes are available and specificity is unclear,
+          instrumentation SHOULD set the `db.response.status_code` to the
+          concatenated string of all codes with '/' used as a separator.
+
+          For example, generic DB instrumentation that detected an error and has
+          SQLSTATE `"42000"` and vendor-specific `1071` should set
+          `db.response.status_code` to `"42000/1071"`."
```
trask, line 425: this is a great note 🌟

model/database/spans.yaml
```diff
@@ -132,8 +145,15 @@ groups:
       - ref: network.peer.port
         requirement_level:
           recommended: if and only if `network.peer.address` is set.
+      - ref: db.response.status_code
+        brief: >
+          [Cassandra protocol error code](https://github.com/apache/cassandra/blob/cassandra-5.0/doc/native_protocol_v5.spec). If
```
lmolkova, line 150:
```suggestion
          [Cassandra protocol error code](https://github.com/apache/cassandra/blob/cassandra-5.0/doc/native_protocol_v5.spec). 
```

model/database/spans.yaml
lmolkova, line 152:
```suggestion
          [Cassandra protocol error code](https://github.com/apache/cassandra/blob/cassandra-5.0/doc/native_protocol_v5.spec) represented as a string.
```

maryliag (approved):

Issue #308, opened 2023-09-11
Revisit system network metrics attributes


dmitryax:
## Context

This issue is a follow-up to https://github.com/open-telemetry/semantic-conventions/pull/89/files#r1299769832. We need to take the opportunity of the attribute renaming required for https://github.com/open-telemetry/semantic-conventions/issues/51 and choose names that inspire full confidence. The goal is to open a discussion and reach consensus on attribute names that minimize the risk of future changes.

## All Network Metrics Attributes (Old Name / Current Name):

- `device` -> `system.device`
- `direction` -> `system.network.direction`
- `protocol` -> `network.protocol`
- `state` -> `system.network.state`

We should revisit each of these attributes within the scope of this issue. If you have any ideas or suggestions, please comment. The following are the main concerns we should address:

### `protocol` / `network.protocol`

We picked `network.protocol` because it's generic and can be applicable to other metrics. It's listed in https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attributes.md#network-attributes. We should rename it to `network.protocol.name` then.

### `state` / `system.network.state`

While it currently sounds like a state of the network, in the OTel collector host metrics receiver, it represents connection status (e.g., "CLOSED," "LISTEN," ESTABLISHED," etc.). I think we should rename it to `network.connection.status` and add it to https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attributes.md#network-attributes.

# Sub-issues (added by @mx-psi):
- [x] Rename `device` to `system.device`
  - [Added `network.interface.name`](https://github.com/open-telemetry/semantic-conventions/pull/1492)
- [x] Rename `direction` to `system.network.direction`
  - [Renamed to `network.io.direction` with values `transmit`/`receive`](https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/model/registry/network.yaml#L223-L234)
- [x] Rename `protocol` to `network.protocol` or `network.protocol.name`
  - [Renamed to `network.protocol.name`](https://github.com/open-telemetry/semantic-conventions/blob/01c542ccacd869ee8a329aa33c65fb2c47bf2815/model/registry/network.yaml#L159-L164)
- [ ] Rename `state` to `system.network.state`
  - PR is open to fix this: #1632
- [x] Rename `system.cpu.logical_number` to `host.cpu.logical_number`
  - Closed by https://github.com/open-telemetry/semantic-conventions/issues/325#issuecomment-1721185169

[...]

PR #1427, opened 2024-09-24
Add k8s.{pod,node}.network.{io,errors} metrics

ChrsMark:

Fixes https://github.com/open-telemetry/semantic-conventions/issues/1487

## Changes

This PR adds the following metrics:
- `k8s.pod.network.io`
- `k8s.pod.network.errors`
- `k8s.node.network.io`
- `k8s.node.network.errors`

The above metrics are already used in the Collector in the [kubeletstats](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kubeletstatsreceiver/documentation.md) receiver and the naming is aligned with the [System metrics](https://github.com/open-telemetry/semantic-conventions/blob/main/model/system/metrics.yaml).

These metrics come with the following attributes `network.io.direction` and `system.device` which are already part of SemConv registry and are used in the [System metrics](https://github.com/open-telemetry/semantic-conventions/blob/main/model/system/metrics.yaml). The attributes in the Collector will need to be adjusted to follow the recent SemConv rules.


model/k8s/metrics.yaml
```diff
@@ -30,6 +30,28 @@ groups:
...
+  # k8s.pod.network.* metrics
+  - id: metric.k8s.pod.network.io
+    type: metric
+    metric_name: k8s.pod.network.io
+    stability: experimental
+    brief: "Pod network IO"
+    instrument: counter
+    unit: "By"
+    attributes:
+      - ref: system.device
```
TylerHelmuth, line 42: Is this the equivalent of `interface`?

docs/system/k8s-metrics.md
```diff
@@ -85,6 +85,74 @@ This metric is [recommended][MetricRecommended].
...
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `k8s.pod.network.io` | Counter | `By` | Pod network IO | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`network.io.direction`](/docs/attributes-registry/network.md) | string | The network IO operation direction. | `transmit` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`system.device`](/docs/attributes-registry/system.md) | string | The device identifier | `(identifier)` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
TylerHelmuth, line 111: To save ourselves work later and to start preparing end users, can we start a list somewhere, maybe in the readme, of the breaking changes that will be included in the stable versions of these semantics?

I know this particular change is additive in this repo, but considering that the kubeletstatsreceiver has been emitting this metric with the attributes `interface` and `direction` since its inception this change will be very impactful.

model/k8s/metrics.yaml
ChrsMark: Yes, that is designed based on the [hostmetricsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/networkscraper/documentation.md#systemnetworkio) which will also need to be updated to the latest SemConv once the System metrics are stable.

docs/system/k8s-metrics.md
ChrsMark: Good idea. I wonder where this information would fit best. @open-telemetry/specs-semconv-maintainers @open-telemetry/specs-semconv-approvers any suggestions?

jsuereth: I think, ideally, we should be able to understand breaking changes and document them in Schema URL eventually for this purpose.

We don't have a good mechanism right now to do that, but @trask and @lmolkova have the most experience having stabilzied HTTP and soon, DB.

ChrsMark: One thing we can do for the Collector specifically is to file a meta issue that will collect all this breaking changes.

Additionally we could update the k8s' components that will be affected to have their documentation explicitly mentioning this breaking change meta issue.

model/k8s/metrics.yaml
ChrsMark: However, I'm not really sure if that's a good and settled decision right now: https://github.com/open-telemetry/semantic-conventions/issues/308#issuecomment-2391638296

docs/system/k8s-metrics.md
lmolkova: we usually have a migration plan as a part of stabilization process - like this one 

https://github.com/open-telemetry/semantic-conventions/blob/cf4ce09650a7a8bafc41a13d062e6f8ac64ade84/docs/database/README.md?plain=1#L18-L39

In a few words 
- the instrumentation should keep doing what it's doing now (following semconv or not)
- it should have an option to enable new conventions (for collector I'd assume it not be env var, but it'd make sense to have consistent way of opting in across different receivers). 
- it can drop old conventions or just switch to a new ones by default with the next major version. 

We've been also creating migration guides like these https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/http-migration.md or https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/db-migration.md which outline the list of changes. 

You can probably do something similar and just mention undocumented (in semconv) old attributes.

model/k8s/metrics.yaml
```diff
@@ -60,3 +82,25 @@ groups:
       Total memory usage of the Node
     instrument: gauge
     unit: "By"
+
+  # k8s.node.network.* metrics
+  - id: metric.k8s.node.network.io
+    type: metric
+    metric_name: k8s.node.network.io
+    stability: experimental
+    brief: "Node network IO"
```
lmolkova, line 91: could you please elaborate? Is it a number of bytes transmitted by the node?

model/k8s/metrics.yaml
```diff
@@ -60,3 +82,25 @@ groups:
...
+    instrument: counter
+    unit: "By"
+    attributes:
+      - ref: system.device
+      - ref: network.io.direction
+  - id: metric.k8s.node.network.errors
+    type: metric
+    metric_name: k8s.node.network.errors
+    stability: experimental
+    brief: "Node network errors"
```
lmolkova, line 101: Could you elaborate on this one? Is there some external documentation we can link, what kinds of errors does it cover?

model/k8s/metrics.yaml
```diff
@@ -60,3 +82,25 @@ groups:
...
+    instrument: counter
+    unit: "{error}"
+    attributes:
+      - ref: system.device
+      - ref: network.io.direction
```
lmolkova, line 106: is there any information about the error we can include? We have a very vague `error.type` attribute intended for this purpose.

model/k8s/metrics.yaml
```diff
@@ -30,6 +30,28 @@ groups:
...
+    instrument: counter
+    unit: "By"
+    attributes:
+      - ref: system.device
+      - ref: network.io.direction
+  - id: metric.k8s.pod.network.errors
+    type: metric
+    metric_name: k8s.pod.network.errors
+    stability: experimental
+    brief: "Pod network errors"
```
lmolkova, line 48: similar to the node errors - could you please elaborate on what it means?

model/k8s/metrics.yaml
```diff
@@ -30,6 +30,28 @@ groups:
...
+    instrument: counter
+    unit: "{error}"
+    attributes:
+      - ref: system.device
+      - ref: network.io.direction
```
lmolkova, line 53: can we report any meaningful `error.type` here?

model/k8s/metrics.yaml
```diff
@@ -30,6 +30,28 @@ groups:
     instrument: gauge
     unit: "By"
 
+  # k8s.pod.network.* metrics
+  - id: metric.k8s.pod.network.io
+    type: metric
+    metric_name: k8s.pod.network.io
+    stability: experimental
+    brief: "Pod network IO"
```
lmolkova, line 38:
```suggestion
    brief: "Number of bytes transmitted by the pod"?
```

ChrsMark: It's not only transmitted but also received. That's why I guess we used `IO` in the kubeletstats' docs.

model/k8s/metrics.yaml
ChrsMark: The best external reference we can find is from the Kubelet's stats API: https://github.com/kubernetes/kubernetes/blob/v1.31.1/staging/src/k8s.io/kubelet/pkg/apis/stats/v1alpha1/types.go#L189-L204

model/k8s/metrics.yaml
ChrsMark: I don't think we have any information about the errors' types. It seems these are aggregated errors over a specific interface: https://github.com/kubernetes/kubernetes/blob/v1.31.1/staging/src/k8s.io/kubelet/pkg/apis/stats/v1alpha1/types.go#L189-L204

docs/system/k8s-metrics.md
ChrsMark: Thank's for the information @lmolkova! If I understand correctly could we create a similar migration plan file for K8s already and start collecting the diffs there?

If I understood @TylerHelmuth initial comment here, we would like to start informing users already about such changes/diffs even if we are far from stability. In this way users could already be prepared about what changes to expect.

model/k8s/metrics.yaml
lmolkova: Makes sense! Is there something more elaborate we can say than "Pod network IO" ?

ChrsMark: I made it `Network bytes for the Pod`. Guess it's more descriptive?

model/k8s/metrics.yaml
ChrsMark: If https://github.com/open-telemetry/semantic-conventions/pull/1492 makes it, we will use `network.interface.name` here.

ChrsMark: Changed to `network.interface.name`: [804d91a](https://github.com/open-telemetry/semantic-conventions/pull/1427/commits/804d91a851a0184d830b881beb47ac2b95e2d66a)

docs/system/k8s-metrics.md
TylerHelmuth: @joaopgrassi I love it, thats exactly what I want to do here

ChrsMark:
@TylerHelmuth @lmolkova anything blocking this one?

docs/system/k8s-metrics.md
TylerHelmuth: @ChrsMark where did we land with trying to track these breaking changes as we go?

ChrsMark: As mentioned at https://github.com/open-telemetry/semantic-conventions/pull/1427#issuecomment-2449304586, I have filed https://github.com/open-telemetry/semantic-conventions/issues/1537 for the migration plan (following the idea suggested above). I will start working on the migration plan PR soonish.

TylerHelmuth (approved):

model/k8s/metrics.yaml
ChrsMark: Seems, this is covered. I'm going to resolve this conversation.

docs/system/k8s-metrics.md
ChrsMark: Seems we are covered here. I'm going to resolve this conversation.

lmolkova (approved):

lmolkova:
@ChrsMark please regenerate markdown files (there was a small change on main) and it should be ready to go.

Issue #1433, opened 2024-09-27
Add User Agent OS attributes


rogercoll:
### Area(s)

area:os, area:user-agent

### Is your change request related to a problem? Please describe.

The current user-Agent attributes don't make any reference to the represented OS. We propose adding semantic convention attributes specifically aimed at extracting the operating system (OS) from user-agent strings. The OS information is already available in the current `user_agent.original` attribute, but it is embedded within a larger string, making it difficult to reliably extract and utilize. By introducing dedicated semantic convention attributes such as user_agent.os.name, user_agent.os.version, and user_agent.os.family, we can isolate this information for easier parsing and processing. 

### Describe the solution you'd like

Embed OS attribute under the `user_agent` namespace: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/os.md  

A similar solution is the Elastic ECS reuse fields: https://www.elastic.co/guide/en/ecs/current/ecs-user_agent.html#_field_reuse_30

PR #1434, opened 2024-09-27
add User-agent OS attributes

rogercoll:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1433

## Changes


rogercoll:
The PR is mostly a copy and paste from the [OS attributes current definition](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/os.md) to the user_agent's attributes, is there a way to reference them but applying the "user_agent" prefix?

docs/attributes-registry/user-agent.md
```diff
@@ -19,3 +22,31 @@ Describes user-agent attributes.
 **[1]:** [Example](https://www.whatsmyua.info) of extracting browser's name from original string. In the case of using a user-agent for non-browser products, such as microservices with multiple names/versions inside the `user_agent.original`, the most significant name SHOULD be selected. In such a scenario it should align with `user_agent.version`
 
 **[2]:** [Example](https://www.whatsmyua.info) of extracting browser's version from original string. In the case of using a user-agent for non-browser products, such as microservices with multiple names/versions inside the `user_agent.original`, the most significant version SHOULD be selected. In such a scenario it should align with `user_agent.name`
+
+## User-agent OS Attributes
+
+Describes the OS user-agent attributes.
```
lmolkova, line 28: Could you please describe the use-case for these attributes? is there any instrumentation or semantic convention that's supposed to populate it?  

We usually add attributes along with the convention (span/metric/resource/event) that uses them - this usually results in more specific attribute definition and more clarity for those who want to populate the attribute on other conventions.

So, could we define such convention and/or document the use-case in some other way?

rogercoll: > Could you please describe the use-case for these attributes?

For example, to understand which are the source OSes that are making requests to your system. This could be used from a Nginx Ingress Controller log, which records all requests UserAgent's.

> is there any instrumentation or semantic convention that's supposed to populate it?

There is an ongoing PR in the collector: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35458

rogercoll: @lmolkova How could we proceed with this PR? In which section could I add the use-case? brief field?

If a proof of concept is needed beforehand, we could unblock https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35886#discussion_r1829742221 by just referencing this PR for now.

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,83 @@ groups:
...
+              brief: "SunOS, Oracle Solaris"
+              stability: experimental
+            - id: z_os
+              value: 'z_os'
+              brief: "IBM z/OS"
+              stability: experimental
+        brief: >
+          The operating system type.
+        stability: experimental
+      - id: user_agent.os.description
```
trask, line 94: is this needed? I don't see it in ECS, and I'm not sure what you would populate it with via https://github.com/ua-parser/uap-core

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,83 @@ groups:
...
+        brief: 'Human readable operating system name.'
+        examples: ['iOS', 'Android', 'Ubuntu']
+      - id: user_agent.os.version
+        type: string
+        stability: experimental
+        brief: >
+          The version string of the operating system as defined in
+          [Version Attributes](/docs/resource/README.md#version-attributes).
+        examples: ['14.2.1', '18.04.1']
+      - id: user_agent.os.build_id
```
trask, line 113: same question for this attribute, thanks

model/user-agent/registry.yaml
rogercoll: I think the equivalent in ECS would be the `os.full` field, ECS embeds all the `os` fields in the user_agent: https://www.elastic.co/guide/en/ecs/current/ecs-user_agent.html#_field_reuse_30

Given that there is no direct mapping for this field in https://github.com/ua-parser/uap-core/blob/master/docs/specification.md#os_parsers and the collector's use case does not cover this attribute at the moment, it is ok for me to remove it until the use case https://github.com/open-telemetry/semantic-conventions/pull/1434/commits/1b46946a1d752f6e7fc157c22ec6670262eb7b11

model/user-agent/registry.yaml
rogercoll: Removing it until we have a use case https://github.com/open-telemetry/semantic-conventions/pull/1434/commits/1b46946a1d752f6e7fc157c22ec6670262eb7b11

wdyt? Thanks for the review!

docs/attributes-registry/user-agent.md
trask: I think the (Event) Log SIG may be far enough along that we can start looking into #982, I've left a couple of initial comments there, and added it to next week's Log SIG meeting because I think it's another good example for us to use when considering #1651

Another option for use case besides access log could be to add these attributes as opt-in on HTTP spans / metrics which would enable breaking down HTTP telemetry by Android/iOS or by Windows/Mac.

It would be helpful to understand how the attributes in this PR (`os.type`/`os.name`/`os.version`) map to a common user-agent parser e.g. https://github.com/ua-parser

rogercoll: > Another option for use case besides access log could be to add these attributes as opt-in on HTTP spans / metrics which would enable breaking down HTTP telemetry by Android/iOS or by Windows/Mac.

+1. From the collector's perspective, it does not matter if the `user_agent` attribute is within a metric, log or span. The idea behind this PR is to define/formalize some specific fields that are already available in the current unparsed `user_agent` attribute.   

> It would be helpful to understand how the attributes in this PR (os.type/os.name/os.version) map to a common user-agent parser e.g. https://github.com/ua-parser

Maybe this comment provides some hints https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35458#issuecomment-2395604776. The only one missing at the moment is `user_agent.os.type`, as the ua-parser package does not provide a 1:1 mapping, and we would need to extend the lookup. If preferred, we can start by just adding the fields that have a 1:1 mapping.

trask: thanks, I think leaving `user_agent.os.type` out from this initial PR would help

I also think it could help to reference the ua parser mapping in the attribute notes

rogercoll: Let's keep things simple for this initial PR then, `user_agent.os.type` removed and ua-parser reference added in https://github.com/open-telemetry/semantic-conventions/pull/1434/commits/a4183fa086eb6c88d9ee046d5e87bd6b70e3f731

.chloggen/user_agent_os.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: user-agent
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add user-agent OS attributes
```
trask, line 13:
```suggestion
note: Add `user_agent.os.name` and `user_agent.os.version` attributes
```

trask (approved):
this looks like a straightforward embedding of `os.name` and `os.version` under the `user_agent.*` namespace 👍

joaopgrassi (approved):

trisch-me (approved):
I think we might revisit these attributes when embediing will be there to re-use `os.*` attributes

trisch-me:
from my POV one of the goals of embedding - not to break semantics of fields itself, but adding additional meaning.
In this particular case `os.*` attributes will be attached to the user agent hence additional meaning. But this is what we already having so I don’t see any problems

trask:
@rogercoll sorry, one more thing, can you describe _how_ you are using these attributes? e.g. are you enriching existing spans that already have `user_agent.original`? or emitting events, or other spans using them? thanks!

based on how you're using them, we can consider if we would like to add them to one of those signals in semconv, e.g. as an opt-in attribute on http spans

rogercoll:
> @rogercoll sorry, one more thing, can you describe _how_ you are using these attributes? e.g. are you enriching existing spans that already have `user_agent.original`? or emitting events, or other spans using them? thanks!
> 
> based on how you're using them, we can consider if we would like to add them to one of those signals in semconv, e.g. as an opt-in attribute on http spans

Not enriching spans at the moment, but enriching logs that contain an unparsed user_agent. Specifically, the Nginx Ingress Controller [structured logs](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/log-format.md) in which we extract the corresponding fields using OpenTelemetry Collector's processors (OTTL). Although the [Nginx Ingress Controlled Log Event PR](https://github.com/open-telemetry/semantic-conventions/pull/982) is in a draft state, we would like to include the log attributes containing this `user_agent.os*` values.

trask:
> Although the [Nginx Ingress Controlled Log Event PR](https://github.com/open-telemetry/semantic-conventions/pull/982) is in a draft state, we would like to include the log attributes containing this `user_agent.os*` values.

perfect, thanks (cc @gregkalapos)

PR #1436, opened 2024-09-28
Fix markdown generation

trask:
Markdown generation seems to be failing currently due to

>   ⚠ The attribute `file.attributes` in the group `registry.file` contains an example that will be considered invalid in the future. All examples SHOULD be of type `string[]`. Provenance: "/home/weaver/source/file/registry.yaml"

but for some reason it's not failing CI and so an out-of-sync markdown was able to get merged.


trask:
closing, fixed by #1441

PR #1437, opened 2024-09-29
Recommend to capture `db.namespace` from initial connection over not capturing any, also specify `db.namespace` value for PostgreSQL, MySQL and MariaDB

trask:
Recommend to capture `db.namespace` from the initial connection over not capturing any value.

Specify `db.namespace` value for PostgreSQL, MySQL and MariaDB.


lmolkova (approved):

alanwest (approved):

maryliag (approved):

docs/database/postgresql.md
```diff
@@ -0,0 +1,108 @@
...
+
+The current schema may change during the lifetime of a connection, e.g. from executing `SET search_path TO myschema`.
+If the search path has multiple schemas, the first schema in the search path SHOULD be used.
+
+Instrumentation SHOULD set `db.namespace` using the schema provided at connection time if it is
+unable to capture the current schema without causing an additional query to be executed (e.g. `SELECT current_schema()`).
+
+If instrumentation is unable to capture the current schema without causing an additional query to be executed
+(e.g. `SELECT current_schema()`), then it is RECOMMENDED to set `db.namespace` using the schema provided at connection time
+instead of not capturing any schema for `db.namespace`.
```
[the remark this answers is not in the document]
trask: > What value does that provide at this point?

it represents a "best effort" at capturing `db.namespace`

and we guess that most (though certainly not all) applications don't change the schema after the initial connection

also, this is how Java JDBC Instrumentation has always worked, populating `db.name` from the initial connection string

> Was there some other discussion about this maybe?

the discussion took place in SIG meetings

joaopgrassi (approved):

model/database/spans.yaml
```diff
@@ -94,14 +94,103 @@ groups:
...
+          It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.
         examples: ["instance1.products", "customers"]
 
+  - id: db.postgresql
+    type: span
+    extends: db.sql
+    brief: >
+      Attributes for PostgreSQL
+    attributes:
+      - ref: db.namespace
```
lmolkova, line 121: now that we have sql-db-specific yamls, would you mind overwriting generic note/brief on `db.response.status_code` to be specific to that database?

https://github.com/open-telemetry/semantic-conventions/blob/87ec4e6e3b56effd895775f4f8f63e412bde7f75/model/database/spans.yaml#L385

I can also do it as a follow up - just let me know.

trask: I will add that 👍

trask:
@lmolkova @maryliag @alanwest @joaopgrassi @gregkalapos I took another pass at refining `db.namespace` definition factoring in #1312, ptal, thanks!

PR #1440, opened 2024-09-30
Feature Flag evaluation event

dyladan:
First draft of `feature_flag.evaluation` event from the feature flag semantic conventions working group. Based on proposal here: https://docs.google.com/document/d/1sBDD-uifx-Qy0NxtwRcrRAQ48sHQS9dTp6tpArQsoZg/edit

Summary of changes:
- Replace `feature_flag.provider_name` with ~~`feature_flag.provider.id`~~ `feature_flag.system`
- Rename `feature_flag` to `feature_flag.evaluation`
- Add attributes to `feature_flag.evaluation`
  - `feature_flag.context.id`
  - ~~`feature_flag.environment.id`~~ `feature_flag.flag_set.id`
  - `feature_flag.value`
  - `feature_flag.version`
- Remove feature flag span event
- Add `feature_flag.error.code` with values from https://openfeature.dev/specification/types/#resolution-details
- Add `feature_flag.reason` with values from https://openfeature.dev/specification/types/#resolution-details


docs/attributes-registry/feature-flag.md
```diff
@@ -6,15 +6,22 @@
...
+| `feature_flag.provider.id`    | string | The name of the service provider that performs the flag evaluation.                                                     | `Flag Manager`                         | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `feature_flag.value`          | string | The evaluated value of the feature flag.                                                                                | `#ff0000`; `1`; `true`                 | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `feature_flag.variant`        | string | SHOULD be a semantic identifier for a value. If one is unavailable, a stringified version of the value can be used. [1] | `red`; `true`; `on`                    | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `feature_flag.version`        | string | The version of the ruleset used during the evaluation. Could use a hash if a version is not available.                  | `1`; `01ABCDEF`                        | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
jimmyca15, line 24: > Could use a hash if a version is not available.

Why hash specifically? Can't it be any stable value representing the state of the ruleset?

model/feature-flag/registry.yaml
```diff
@@ -10,7 +10,7 @@ groups:
         stability: experimental
         brief: The unique identifier of the feature flag.
```
askpt, line 11: As discussed, I suggested to rename to "The lookup key of the feature flag."

model/feature-flag/registry.yaml
```diff
@@ -10,7 +10,7 @@ groups:
...
         examples: ["logo-color"]
-      - id: feature_flag.provider_name
```
askpt, line 13: I suggest leaving this open. Using an "id" means it is "blocked" only to use IDs, and we can potentially use it for names or other semantically valid values.
I would allow you to use the following examples:
["provider-1", "guid_here",  "provider-v1"]

askpt (commented):
I added a couple of comments; otherwise, it looks good.

model/feature-flag/registry.yaml
```diff
@@ -31,3 +31,27 @@ groups:
...
+        stability: experimental
+        examples: ["1", "01ABCDEF"]
+        brief: >
+          The version of the ruleset used during the evaluation. This may be any stable value which uniquely identifies the ruleset.
+      - id: feature_flag.environment.id
+        type: string
+        stability: experimental
+        examples: ["proj-1", "ab98sgs"]
+        brief: >
+          The identifier of the environment which the feature flag belongs to in a flag management system.
```
beeme1mr, line 57:
```suggestion
      - id: feature_flag.flag_set.id
        type: string
        stability: experimental
        examples: ["proj-1", "ab98sgs", "service1/dev"]
        brief: >
          The identifier of the flag set, which the feature flag belongs to in a flag management system.
```

A flag set is defined [here](https://openfeature.dev/specification/glossary#flag-set) as `A collection of related [flags](https://openfeature.dev/specification/glossary#flag). This grouping helps organize feature flags based on their intended use, facilitating easier management and deployment.

model/feature-flag/registry.yaml
dyladan: I'm not sure what you mean by "leaving this open"

model/feature-flag/registry.yaml
```diff
@@ -9,26 +9,45 @@ groups:
...
           A semantic identifier, commonly referred to as a variant, provides a means
           for referring to a value without including the value itself. This can
           provide additional context for understanding the meaning behind a value.
           For example, the variant `red` maybe be used for the value `#c05543`.
-
-          A stringified version of the value can be used in situations where a
-          semantic identifier is unavailable. String representation of the value
-          should be determined by the implementer.
+      - id: feature_flag.value
+        type: string
```
beeme1mr, line 31: Does SemCon support an `any` type?

model/feature-flag/registry.yaml
askpt: What I mean is to allow any value type to be used, such as a number, string, or GUID.

kinyoklion (approved):

model/feature-flag/registry.yaml
```diff
@@ -9,26 +9,45 @@ groups:
...
+        examples: ["5157782b-2203-4c80-a857-dbbd5e7761db"]
+        brief: >
+          The unique identifier for the flag evaluation context. For example, the targeting key.
+      - id: feature_flag.version
+        type: string
+        stability: experimental
+        examples: ["1", "01ABCDEF"]
+        brief: >
+          The version of the ruleset used during the evaluation. This may be any stable value which uniquely identifies the ruleset.
+      - id: feature_flag.flag_set.id
```
lmolkova, line 48: nit: consider `feature_flag.set.id` - I understand 'flag set' is existing terminology, but `feature_flag.set.id` is still pretty clear and close, but shorter

model/feature-flag/registry.yaml
```diff
@@ -9,26 +9,45 @@ groups:
...
         stability: experimental
-        brief: The unique identifier of the feature flag.
+        brief: The lookup key of the feature flag.
         examples: ["logo-color"]
-      - id: feature_flag.provider_name
+      - id: feature_flag.provider.id
         type: string
         stability: experimental
-        brief: The name of the service provider that performs the flag evaluation.
+        brief: Any string which identifies the feature flag provider being used such as its name.
```
lmolkova, line 17: maybe just 

```suggestion
        brief: Identifies the feature flag provider.
```
?

Does it map to flag management system name? If so, it's really similar to `db.system`. `messaging.system`, `gen_ai.system`. Would it make sense to have `feature_flag.system` for consistency?

model/feature-flag/logs.yaml
```diff
@@ -1,11 +1,26 @@
 groups:
-  - id: log-feature_flag
+  - id: feature_flag.evaluation
     type: event
-    stability: experimental
-    name: feature_flag
+    name: feature_flag.evaluation
     brief: >
-      This event describes feature flag evaluation representation on a Log Record.
+      This semantic convention defines the attributes used to represent a
```
lmolkova, line 6: can we describe an event here (or in the note) - e.g. who/how/when someone would report it?

model/feature-flag/registry.yaml
beeme1mr: I'm in favor of this change. 👍

model/feature-flag/registry.yaml
beeme1mr: Our use cases are exactly like `gen_ai.system`. For consistency, I'm fine switching to `feature_flag.system`.

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-events.md#common-attributes

Thanks @lmolkova!

beeme1mr: The gen ai example enumerates some well-known values and allows custom ones. This list could become quite large in the feature flagging space but it may be something worth considering.

model/feature-flag/logs.yaml
dyladan: Would you want that here or in logs.yaml?

model/feature-flag/registry.yaml
dyladan: updated to use `system`. I left it as an open-ended string. I think we should consider putting well-known values in a follow up PR.

model/feature-flag/logs.yaml
dyladan: nvm i see this comment is on logs.yaml. I was turned around.

model/feature-flag/registry.yaml
dyladan: We could consider moving the flag value to be a body field, which would allow us to use `unknown` for its type. I'm still not clear when it is appropriate to use a body field vs an attribute. Maybe @lmolkova can make a suggestion.

beeme1mr (approved):

docs/attributes-registry/feature-flag.md
```diff
@@ -6,21 +6,46 @@
...
+| `targeting_match` | The resolved value was the result of a dynamic evaluation, such as a rule or specific user-targeting. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `unknown` | The reason for the resolved value could not be determined. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+## Deprecated Feature Flag Attributes
+
+Describes deprecated Feature Flag attributes.
+
+| Attribute | Type | Description | Examples | Stability |
+|---|---|---|---|---|
+| <a id="feature-flag-provider-name" href="#feature-flag-provider-name">`feature_flag.provider_name`</a> | string | Deprecated, use `feature_flag.provider.id` instead. | `Flag Manager` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `feature_flag.provider.id`. |
```
beeme1mr, line 51:
```suggestion
| <a id="feature-flag-provider-name" href="#feature-flag-provider-name">`feature_flag.provider_name`</a> | string | Deprecated, use `feature_flag.provider.id` instead. | `Flag Manager` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `feature_flag.system`. |
```

model/feature-flag/deprecated/registry-deprecated.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - id: registry.feature_flag.deprecated
+    type: attribute_group
+    display_name: Deprecated Feature Flag Attributes
+    brief: "Describes deprecated Feature Flag attributes."
+    attributes:
+      - id: feature_flag.provider_name
+        type: string
+        brief: 'Deprecated, use `feature_flag.provider.id` instead.'
+        stability: experimental
+        deprecated: "Replaced by `feature_flag.provider.id`."
```
beeme1mr, line 11:
```suggestion
        deprecated: "Replaced by `feature_flag.system`."
```

model/feature-flag/registry.yaml
dyladan: ah I see what you mean. Unfortunately that isn't possible with the types available.

model/feature-flag/registry.yaml
```diff
@@ -9,26 +9,126 @@ groups:
...
+        examples: ["1", "01ABCDEF"]
+        brief: >
+          The version of the ruleset used during the evaluation. This may be any stable value which uniquely identifies the ruleset.
+      - id: feature_flag.set.id
+        type: string
+        stability: experimental
+        examples: ["proj-1", "ab98sgs", "service1/dev"]
+        brief: >
+          The identifier of the [flag set](https://openfeature.dev/specification/glossary/#flag-set) which the feature flag belongs to in a flag management system.
+      - id: feature_flag.reason
```
dyladan, line 48: @beeme1mr do you think `feature_flag.reason` is clear that it is talking specifically about the evaluation or should we do evaluation reason? In other semconv it looks like they are a bit more verbose about this.

```suggestion
      - id: feature_flag.evaluation.reason
```

docs/attributes-registry/feature-flag.md
```diff
@@ -6,21 +6,61 @@
...
-| <a id="feature-flag-provider-name" href="#feature-flag-provider-name">`feature_flag.provider_name`</a> | string | The name of the service provider that performs the flag evaluation. | `Flag Manager` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| <a id="feature-flag-variant" href="#feature-flag-variant">`feature_flag.variant`</a> | string | SHOULD be a semantic identifier for a value. If one is unavailable, a stringified version of the value can be used. [1] | `red`; `true`; `on` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="feature-flag-context-id" href="#feature-flag-context-id">`feature_flag.context.id`</a> | string | The unique identifier for the flag evaluation context. For example, the targeting key. | `5157782b-2203-4c80-a857-dbbd5e7761db` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="feature-flag-error-code" href="#feature-flag-error-code">`feature_flag.error.code`</a> | string | An error code which shows why a flag evaluation failed | `provider_not_ready`; `targeting_key_missing`; `provider_fatal`; `general` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
toddbaert, line 19:
```suggestion
| <a id="feature-flag-error-code" href="#feature-flag-error-code">`feature_flag.error.code`</a> | string | An error code which denotes why a flag evaluation failed | `provider_not_ready`; `targeting_key_missing`; `provider_fatal`; `general` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

docs/attributes-registry/feature-flag.md
```diff
@@ -6,21 +6,61 @@
...
+| <a id="feature-flag-error-message" href="#feature-flag-error-message">`feature_flag.error.message`</a> | string | A message explaining why an error occurred during flag evaluation. | `Flag `header-color` expected type `string` but found type `number`` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
toddbaert, line 20:
```suggestion
| <a id="feature-flag-error-message" href="#feature-flag-error-message">`feature_flag.error.message`</a> | string | A message explaining the nature of an error occurring during flag evaluation | `Flag `header-color` expected type `string` but found type `number`` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

A bit of a nit, but "explaining why an error occurred" seems slightly inaccurate. Messages describe errors, generally, not just "why" they occurred.

docs/attributes-registry/feature-flag.md
```diff
@@ -6,21 +6,61 @@
...
+| <a id="feature-flag-key" href="#feature-flag-key">`feature_flag.key`</a> | string | The lookup key of the feature flag. | `logo-color` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="feature-flag-reason" href="#feature-flag-reason">`feature_flag.reason`</a> | string | The reason code which shows how a feature flag value was determined. | `static`; `targeting_match`; `error`; `default` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="feature-flag-set-id" href="#feature-flag-set-id">`feature_flag.set.id`</a> | string | The identifier of the [flag set](https://openfeature.dev/specification/glossary/#flag-set) which the feature flag belongs to in a flag management system. | `proj-1`; `ab98sgs`; `service1/dev` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
toddbaert, line 23:
```suggestion
| <a id="feature-flag-set-id" href="#feature-flag-set-id">`feature_flag.set.id`</a> | string | The identifier of the [flag set](https://openfeature.dev/specification/glossary/#flag-set) to which the feature flag belongs. | `proj-1`; `ab98sgs`; `service1/dev` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

Nit: I think I would leave out the "management system" bit; "management system" implies a robustness that we should not assume, IMO.

model/feature-flag/registry.yaml
toddbaert: I think evaluation reason is certainly more clear. :thinking:

toddbaert (approved):
I left some small nits (perhaps not in the right places; I'm not sure if some portion of this is generated or not) but approving.

model/feature-flag/registry.yaml
beeme1mr: We called it `feature_flag.evaluation.reason` in the Google doc. @dyladan error code and message also contained the `feature_flag.evaluation` prefix.

askpt (approved):

model/feature-flag/logs.yaml
```diff
@@ -1,11 +1,49 @@
...
+        requirement_level: recommended
+      - ref: feature_flag.context.id
+        requirement_level: recommended
+      - ref: feature_flag.version
+        requirement_level: recommended
+      - ref: feature_flag.set.id
+        requirement_level: recommended
+      - ref: feature_flag.evaluation.reason
+        requirement_level: recommended
+      - ref: feature_flag.evaluation.error.code
```
lmolkova, line 30: any reason not to reuse generic `error.type`? It'd be great for consistency, especially if we'll ever have some FF metrics or spans.
It seems you'd still have `feature_flag.evaluation.error.message` for anything FF-specific.

model/feature-flag/logs.yaml
```diff
@@ -1,11 +1,49 @@
...
+        requirement_level:
+          conditionally_required: Required if an error occurred during flag evaluation and `feature_flag.evaluation.error.message` is not set, recommended otherwise.
```
lmolkova, line 32: not sure I follow, could we elaborate? I.e. (nerdy interpretation could be):

- error occurred and `feature_flag.evaluation.error.message` is not set - required
- error, but `feature_flag.evaluation.error.message` is not set - recommended (why not required?)
- no error and `feature_flag.evaluation.error.message` is set - recommended  (should not  be set, right?)
- no error and `feature_flag.evaluation.error.message` is not  set - recommended (should not  be set, right?)

It does not seem right.

I'd simplify it to:

```suggestion
          conditionally_required: if and only if an error occurred during flag evaluation.
```

model/feature-flag/logs.yaml
```diff
@@ -1,11 +1,49 @@
...
+      - ref: feature_flag.evaluation.error.message
+        requirement_level:
+          conditionally_required: Required if an error occurred during flag evaluation and `feature_flag.evaluation.error.code` is not set, recommended otherwise.
```
lmolkova, line 35:
```suggestion
          recommended: If and only if an error occurred. It's NOT RECOMMENDED to duplicate the value of `feature_flag.evaluation.error.code` in `feature_flag.evaluation.error.message`.
```

lmolkova:
Left a few mostly cosmetic comments. Please take a look and also, could you please update https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml (list renamed attributes there)?. Thanks!

model/feature-flag/logs.yaml
dyladan: I was trying to make it so that either error.message or error.code was required, but not both. In some cases error code might not be available, but if it is available it is likely sufficient on its own without message

jsuereth:
FYI - you need to clean up the event group's id:

```
Diagnostic report:

Violation: Event id 'feature_flag.evaluation' is invalid. Event id must follow 'event.{name}' pattern and match 'event.feature_flag.evaluation'
  - Category         : yaml_schema_violation
  - SemConv group    : feature_flag.evaluation
  - SemConv attribute: feature_flag.evaluation
  - Provenance: /home/weaver/source/feature-flag/logs.yaml  
```

dyladan:
@jsuereth like this? [43cdfd4](https://github.com/open-telemetry/semantic-conventions/pull/1440/commits/43cdfd4a7e234a817162a92dfd37c3af68c9d79e)

lmolkova (approved):
LGTM!

joaopgrassi (approved):
> Please take a look and also, could you please update [main/schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) (list renamed attributes there)?

This is pending still.

dyladan:
> > Please take a look and also, could you please update [main/schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) (list renamed attributes there)?
> 
> This is pending still.

Added `feature_flag.provider_name` -> `feature_flag.system` rename.

There are a couple changes I'm not sure how to make there (or if it's possible).

1. We removed the span semconv entirely in favor of events
2. We renamed the event from `feature_flag` to `event.feature_flag.evaluation`

dyladan:
> > > Please take a look and also, could you please update [main/schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) (list renamed attributes there)?
> > 
> > 
> > This is pending still.
> 
> Added `feature_flag.provider_name` -> `feature_flag.system` rename.
> 
> There are a couple changes I'm not sure how to make there (or if it's possible).
> 
> 1. We removed the span semconv entirely in favor of events
> 2. We renamed the event from `feature_flag` to `event.feature_flag.evaluation`

According to https://opentelemetry.io/docs/specs/otel/schemas/file_format_v1.1.0 it looks like you can rename _`span_events`_ but not top-level events? Additionally, it's still not clear at all to me how to handle the differences events/logs/span_events. Previously we had a file named log which had an event name and a file named span which had a span event. Now we only have an event. How do we map that in the schema? And if we want to keep span events, do we need multiple files or is it possible to have a single event definition that works for both?

.chloggen/1440.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
```
lmolkova, line 1:
```suggestion
change_type: breaking
```

.chloggen/1440.yaml
```diff
@@ -0,0 +1,4 @@
...
+component: feature_flag
+note: Define `feature_flag.evaluation` event.
```
lmolkova, line 3: Rename `feature_flag` event to `feature_flag.evaluation` event, define new feature flag attributes and provide body definition. Remove `feature_flag` span event definition in favor of log-based event.

lmolkova (approved):
> There are a couple changes I'm not sure how to make there (or if it's possible).

> We removed the span semconv entirely in favor of events
> We renamed the event from feature_flag to event.feature_flag.evaluation

changelog entry is the best we can do today. Added some suggestions

docs/attributes-registry/feature-flag.md
```diff
@@ -6,21 +6,47 @@
...
+| <a id="feature-flag-evaluation-reason" href="#feature-flag-evaluation-reason">`feature_flag.evaluation.reason`</a> | string | The reason code which shows how a feature flag value was determined. | `static`; `targeting_match`; `error`; `default` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="feature-flag-key" href="#feature-flag-key">`feature_flag.key`</a> | string | The lookup key of the feature flag. | `logo-color` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="feature-flag-set-id" href="#feature-flag-set-id">`feature_flag.set.id`</a> | string | The identifier of the [flag set](https://openfeature.dev/specification/glossary/#flag-set) to which the feature flag belongs. | `proj-1`; `ab98sgs`; `service1/dev` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
rossgrambo, line 22: Were people aligned on `feature_flag.set.id`? In our service, we use a "label" which a string that groups flags together but doesn't offer anything more. 

The `id` here makes it seem like `feature_flag.set` is a larger concept that could have other fields on it. I'm unfamiliar with what those would be. 

Did we consider something flatter? Like perhaps `| feature_flag.set | string |`

rossgrambo (approved):

docs/attributes-registry/feature-flag.md
rossgrambo: I suppose we also offer tags which some users use to group flags. An array here could offer more flexibility for grouping  `| feature_flag.tags | string[] |`. I know this isn't defined in OpenFeature but I also can't tell if OpenFeature has a limit for how many sets a flag can be in. This would better support multiple.

dyladan: I think it's just because of the history of renames. It was environment id when we thought we might multiple fields on environment, then we renamed environment to flag set. There wasn't really a conscious choice to keep the `id` part for flag set. I think we could definitely consider flattening to `feature_flag.set`. @beeme1mr do you think we're likely to have anything other than id in the future?

dyladan: It sounds to me like your `label` concept is a set id. Am I misunderstanding? I don't see any strong reason to flatten in this instance. We're using id in a lot of places and this is consistent with that, and it allows the potential for expansion in the future. One example of a future expansion might be `feature_flag.set.source` which could be `disk`, `network`, `git`, or something else. 

If your flag label is not a flag set ID can you please provide an example to help me understand?

dyladan:
> @dyladan This Looks good to merge, assuming the conversations are resolved. The yamllint error - let me know if we need to update the settings. We may be hitting limitations in YAML event definitions and "max line length" rules...

I wish the yaml lint would tell me which line is failing. I can't find where the error is actually caused. I think it's likely the table that I inlined for the error values? But that's only 197 characters

docs/attributes-registry/feature-flag.md
rossgrambo: > It sounds to me like your `label` concept is a set id.

Correct. In the systems I'm familiar with, 'set' is not an object that justifies a true 'id' or other additional fields. It's only a string that relates it to other flags.

In Azure, we offer a `string` or `string[]` to group flags together (or support prefixes). In Launch Darkly (I believe they use tags which is `string[]`). 

I'm trying to learn about how OpenFeature implements sets- but I don't see it in any of the SDKs I've inspected. I only see it mentioned in its glossary- which makes it sound like it's just a string. Do you or anyone else have a link to where sets are used or more defined in OpenFeature? I checked the .net, go, and node sdks and don't see anything with "set". It's not a field on their eval events either. 

> `feature_flag.set.source` which could be `disk`, `network`, `git`, or something else.

Thank you, this is valid justification. To me- this wouldn't belong on the evaluation event. And if it did- it would belong on the flag itself> `feature_flag.source` or possibly `feature_flag.system.source`.

dyladan: It _can_ map to environment, or to something else. A set of flags may be the `dev` flags, but it may also be `frontend` flags. It's really up to the user and different backends have different ideas of when to group flags together. That's why we decided to keep the grouping name fairly generic with `set`.

Issue #1172, opened 2024-06-20
When someone should use user.name/user.id ?


lmolkova:
## Context

Related to https://github.com/open-telemetry/semantic-conventions/issues/1142

We're considering adding `user.id/user.name` under `db` namespace and not sure which one (neither/both?) to use.
We'd like to stay consistent with root `user` namespace to leverage embedding mechanism (once it's implemented). 

When connecting to the database/messaging system/cloud service/etc it's common to talk about the identity rather than user:

E.g. [Azure has several kinds of identities](https://learn.microsoft.com/en-us/entra/workload-id/workload-identities-overview#workload-identities-other-machine-identities-and-human-identities)
AWS also has multiple - [users, groups, roles](https://catalog.workshops.aws/general-immersionday/en-US/basic-modules/30-iam/iam#identity-and-access-management-(iam)-overview)

## Problem 

The concept of user in semconv seem to map to the human identity only. It does not seem to apply in a generic case of `identity` used to access a resource. 

Even within a 'human identity', it's not clear how to use it:
- is name unique within the system? 
- if so, why there is an id? 
- the login terminology does not seem to apply in many cases (authentication could be a broader term). I believe it's related to https://github.com/open-telemetry/semantic-conventions/pull/1146

So, I think we need to decide and document:

- what does the `user` namespace describe? OS user? Human identity? Any identity?
- what's the appropriate namespace based on this definition?
- what's the relationship between name and id?

lmolkova:
cc @open-telemetry/semconv-security-approvers

lmolkova:
Based on the SemConv SIG discussion 6/24

- `user.name` is a login/human-readable name, e.g. `root` or user login, `user.id` is something else - e.g. an identifier the system uses internally. 
- user namespace is used for end-user (e.g. website user) or the OS user.
- it does not seem to be able to describe generic identity. Also there are multiple identities that act at the same time (end-user, OS user service runs with, client identities used to access resources)

Action items:
- @open-telemetry/semconv-security-approvers will discuss

mjwolf:
I have some thoughts on what a user should be in semantic-conventions, and how it relates to identities.

## Use Cases to Handle

A "user" (or multiple types of user resources) should handle these use cases. 

* Represent existing well-known usages of 'user' attributes.
    * many existing concepts/objects have "user" attributes, OTel should be able to generate telemetry with these attributes.
* Have a trackable scope/chain for each representation of "user".
[...]

PR #1456, opened 2024-10-07
Update enduser domain and add `enduser.pseudo.id`

### LINKED ISSUE #1104: User.id for authenticated user id

heyams:
Fixes #1104


model/enduser/registry.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - id: enduser.scope
+        type: string
+        stability: experimental
+        brief: >
+          Scopes or granted authorities the client currently possesses extracted from token
+          or application security context. The value would come from the scope associated
+          with an [OAuth 2.0 Access Token](https://tools.ietf.org/html/rfc6749#section-3.3)
+          or an attribute value in a [SAML 2.0 Assertion](http://docs.oasis-open.org/security/saml/Post2.0/sstc-saml-tech-overview-2.0.html).
+        examples: 'read:message, write:files'
+      - ref: authentication.id
```
heyams, line 29: @trisch-me need help with this. 

I have tried a few options, none works so far

```yaml
- id: enduser.authentication
- ref: registry.authentication
```
or 
```yaml
-id: enduser.authentication
-type:
    members:
        - ref: authenticaiton.id
```
current yaml allows me to update markdown but it breaks policy check. any ideas?

jsuereth: You would have:

```
- id: enduser.authentication.id
  ...
```

You do not need to make the `authenticaton` namespace if you want it under `enduser`

model/enduser/registry.yaml
```diff
@@ -0,0 +1,32 @@
...
+      Describes information about the end user, which can be used as a subdomain of browser, client, or user domains.
+    attributes:
+      - id: enduser.id
+        type: string
+        stability: experimental
+        brief: >
+          Identifier of an end user who interacts with a system.
+          This identifier may be unique only through best-effort means and does not imply that the user is authenticated to the system.
+        examples: ['QdH5CAWJgqVT4rOr0qtumf']
+      - id: enduser.role
```
lmolkova, line 15: is one role enough or should it be an array? Also, do you need this field right away? If not, maybe we can keep it deprecated for now until the use-case comes?

model/enduser/registry.yaml
```diff
@@ -0,0 +1,32 @@
...
+        type: string
+        stability: experimental
+        brief: 'Actual/assumed role the client is making the request under extracted from token or application security context.'
+        examples: 'admin'
+      - id: enduser.scope
```
lmolkova, line 20: same here, do we need it right now? Can we keep it deprecated for the time being until it's necessary?

model/enduser/registry.yaml
```diff
@@ -0,0 +1,32 @@
+groups:
+  - id: registry.enduser
+    type: attribute_group
+    display_name: End User Attributes
+    brief: >
+      Describes information about the end user, which can be used as a subdomain of browser, client, or user domains.
+    attributes:
+      - id: enduser.id
```
lmolkova, line 8: It seems confusing to set `enduser.id = "QdH5CAWJgqVT4rOr0qtumf"` and `enduser.authentication.id = "lmolkova"`, based on the 

https://github.com/open-telemetry/semantic-conventions/pull/1146/files#r1712997369 and https://github.com/open-telemetry/semantic-conventions/pull/1146/files#r1710187141

It'd be more clear if we called this one `enduser.tracking.id` or `enduser.anonymous.id` so that people would not put PII there.

model/enduser/registry.yaml
heyams: I can keep it as deprecated for now.

model/enduser/registry.yaml
heyams: will change it to `deprecated` for now.

model/enduser/registry.yaml
heyams: per earlier discussion, it seemed that `anonymous` was confusing to some. 
i'm good with `enduser.anonymous_id` or `enduser.tracking_id`. neither `tracking` nor `anonymous` a namespace, nesting here doesn't seem to follow the naming convention?

lmolkova: do you mean that `enduser.tracking.id` would not follow the naming convention? Why?

trask: I like `enduser.tracking.id`

heyams: some examples https://github.com/open-telemetry/semantic-conventions/blob/d5d2b9d2a27f5983a5930699e0d40d5aba5c06fe/schemas/1.19.0#L32

nesting is used for `.`. if not nesting, use `_`.

trask: I'd recommend `.`, check out https://github.com/open-telemetry/semantic-conventions/issues/1462#issue-2576611847

> we have some attributes that would benefit from renaming (e.g. `db.cosmosdb.client_id` should stay consistent with `messaging.client.id`).

heyams: we can discuss it in the next SIG. It seems inconsistent naming patterns used.

lmolkova: can you elaborate on what's inconsistent?

lmolkova: Semconv call: 

- Default: use `.` everywhere
- `_` is for old things and exceptional cases
- we need to document this guidance

MSNev: > I like `enduser.tracking.id`

I'm not a fan of `tracking` as it doesn't really "add" anything to the definition of what this field is really used for... And can have mixed meanings about what you are "tracking" about the enduser. eg. Is this used to identify that the `enduser` and opted in to enable tracking of their requests for support reasons and this is the id of the tracking session?

vs `anonymous` which is a more descriptive definition of the meaning of this property.

You could also use something `enduser.uaid` (Unique Anonymous Id), to be descriptive

lmolkova: we prefer to avoid acronyms https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attribute-naming.md#name-abbreviation-guidelines unless they are well-recognized

But I have no objections against `anonymous` or any other descriptive term.

MSNev: Yeah, sorry I didn't mean to "add" to and extend the discussion was just trying to clarify what the real meaning of this attribute will be for.

My choice from these would be `anonymous` as it makes it clearer (at least to me) about this "should" be used for (as someone will always do something unexpected)

MSNev: And then based on the Sem-Conv meeting this morning, I would suggest that it should be `enduser.anonymous.id` so that the `anonymous` namespace could be extended (if necessary) in the future.

trask: I'm good updating the PR to use `anonymous`, let's just confirm with @jsuereth before merging b/c I thought he had a concern with that term but don't recall specifics (and could be misremembering completely)

heyams: https://github.com/open-telemetry/semantic-conventions/issues/1104#issuecomment-2243437378 here is josh's concern. do i need to wait for next SIG to get it answered? @jsuereth can you take a look at @MSNev's comment above? 
are you ok with `enduser.anonymous.id`? I'm good with `enduser.tracking.id` too if we have consensus.

MSNev: Not that I really want to extend the discussion further, but based on @jsuereth comments and understanding from his perspective, I can see how `anonymous` could also be misunderstood and given his suggestions of

> "user.unknown_id", "anonymous_user.id", "user.unauth_id".

From these comments, I would suggest the `unauth_id` option, would identify the meaning better as it's "unknown" but rather than a random (anonymous) identifier for the current `enduser` which (from a RUM perspective) which persisted for some period of time. As its unlikely that `unauth` would represent some common (extended) namespace then either `_` or `.` would work.

It's also represents a possible "linkable" identifier, while on it's own it doesn't identify the user, but in conjunction with other attributes it (may) be possible to identify the `enduser` so from that perspective it's not "truely" anonymous, which I infer is Josh's point on the name.

jsuereth: My concerns around anonymous still apply.  I don't think the id is really anonymous and can't be considered as such.  It's a tracking id, and should be considered sensitive

heyams: how about `enduser.unauthenticated.id`?

heyams:
need some advice with this weaver error, it didn't point to any files that I have modified. it's persisting.
image

trask:
> need some advice with this weaver error, it didn't point to any files that I have modified. it's persisting. image

if you scroll up from there, you'll see:

> ℹ Validating: $/home/weaver/target/general/attributes.md
> ✖ Could not find: identity

which looks related to this file that was deleted in your PR: `model/enduser/deprecated/common.yaml`

model/enduser/registry.yaml
lmolkova: + `enduser.pseudo.id`

Let's create a list of options and add pros/cons.

heyams: @lmolkova @trisch-me  Please review and fill in anything that i may have missed.: 

`enduser.pseudo.id`
     - pros: it's clear to know this id is not authenticated
     - cons: it might lead to misinterpretation of this id, like it's not a real id, a testing id, a temporary id?
     
`enduser.tracking.id'
     - pros: it's clear that this id is used to track a particular user.
     - cons: `tracking` may raise privacy concerns, as it implies monitoring user behavior, which could lead to user distrust. it also lacks context of what exactly is being tracked (e.g. user actions, sessions, locations, etc)
 
`enduser.unauth.id`
     - pros: `unauth` is short
     - cons: `unauth` is ambiguous, as it can be `unauthenticated` or `unauthorized`. additionally, acronym is not a good naming practice and leads to more confusion.
 
`enduser.temp.id` or `enduser.transient.id`
     - pros: it suggests that this id is temporary and associated with user who has not been authenticated.
     - cons: it lacks context about the id is temporary for what context (e.g. session, authentication)
 
`enduser.unauthenticated.id`
     - pro: it's clear to indicate an authenticated user.
     - cons: it collides with `enduser.authentication.id`, which can be renamed to `enduser.authenticated.id`, then it would have been fine?  
 
`enduser.anonymous.id`
     - pros: it's clear that this id is anonymous.
     - cons: it can be confusing and lacks context. as long we have a clear documentation, this should be ok?

docs/attributes-registry/authentication.md
```diff
@@ -0,0 +1,15 @@
+<!--- Hugo front matter used to generate the website version of this page:
+--->
+
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/attribute_namespace.md.j2 -->
+
+# Authentication
+
+## Authentication Attributes
```
joaopgrassi, line 9: I think this should be inside `enduser.md`. If you add it like this, it will introduce a new "area" called `authentication` which is not what happens in reality. The authentication attributes are inside `enduser` which is the area. Please move it to `enduser.md`

joaopgrassi (dismissed):

model/enduser/registry.yaml
```diff
@@ -0,0 +1,42 @@
...
+        stability: experimental
+        examples: ['QdH5CAWJgqVT4rOr0qtumf']
+        brief: >
+          Deprecated, use `enduser.anonymous.id` instead.
+      - id: enduser.anonymous.id
+        type: string
+        stability: experimental
+        brief: >
+          Identifier of an anonymous end user who interacts with a system.
+          This identifier may be unique only through best-effort means and does not imply that the user is authenticated to the system.
```
trask, line 20: meeting notes from semantic convention SIG meeting today:

* is this something that OpenTelemetry instrumentation is planning populate, or is this only something that vendor-specific instrumentation is planning to populate?
* if it's something that OpenTelemetry instrumentation is planning to populate, what would the implementation look like, e.g. would this be stored in a persistent cookie? would it be stamped onto a specific event?
* there is a desire not to add attributes into the semconv repo when only a single vendor has expressed interest in them
* there is also a desire not to add attributes into the semconv repo without having any span/event/metric definitions that use them

the recommendation for next steps was to discuss this in the Client SIG

heyams:
image

it was a false-negative. that link works. how do i ignore this failure?

trask:
> it was a false-negative. that link works. how do i ignore this failure?

it's ok, it's not a required check, we're working on resolving the flakiness

model/enduser/registry.yaml
trisch-me: what were the pros for `enduser.unauthenticated.id`?

model/enduser/registry.yaml
trisch-me: @heyams were those questions addressed during client sig meeting?

model/enduser/registry.yaml
heyams: we have already discussed and agreed to have `enduser.authentication.id`. `authentication` as a sub-namespace under enduser. anything that is not under `authentication` is either unauthenticated or unauthorized or anonymous or some random id.

MSNev (approved):

model/enduser/registry.yaml
trask: can you post the answers here?

model/enduser/registry.yaml
```diff
@@ -0,0 +1,41 @@
+groups:
+  - id: registry.enduser
+    type: attribute_group
+    display_name: End User Attributes
+    brief: >
+      Describes information about the end user, which can be used as a subdomain of browser, client, or user domains.
+    attributes:
+      - id: enduser.id
+        type: string
+        deprecated: Replaced by `enduser.pseudo.id` attribute.
```
trisch-me, line 10: deprecated attributes should be moved from registry to the deprecated section

heyams: my understanding is that if the namespace is deprecated, then it will move everything to the deprecated folder. 
if some of the attributes are deprecated, then simply use `deprecated:`. please correct me if I'm wrong. an example will be helpful.

model/enduser/registry.yaml
MSNev: > is this something that OpenTelemetry instrumentation is planning populate

- Yes, this will be added by instrumentations either directly or indirectly (via a psuedo user manager  -- like the Session Manager)

> if it's something that OpenTelemetry instrumentation is planning to populate, what would the implementation look like, e.g. would this be stored in a persistent cookie? would it be stamped onto a specific event?

- It will need something like the SessionManager implementation to "manager" the lifecycle of this value, some environments would be just to create a simple random value for App start (like Android) while in a browser which is stateless then the "user manager" would most likely use cookies (but it could also use session storage)

> there is a desire not to add attributes into the semconv repo when only a single vendor has expressed interest in them

- This will not be single vender specific, but it is highly RUM specific

> there is also a desire not to add attributes into the semconv repo without having any span/event/metric definitions that use them

- It should be available both Spans and Logs, while it could be available for Metrics it's cardinality (because its a random value) should not necessarily be used for metrics...

trask: thanks! a couple of followup questions:

* if/when entity (or mutable resource) attributes are available, do you see `enduser.pseudo.id` as one of those?
* are there any specific span or event semantic conventions that we can add `enduser.pseudo.id` to? or is `enduser.pseudo.id` hopefully just an entity (or mutable resource) attribute?

heyams: I believe we discussed these in the client SIG. I will wait for @MSNev's response.

MSNev: - If/when mutable resources attributes (whether they are entities or [shorter lived lifespan resources](https://github.com/open-telemetry/opentelemetry-specification/pull/4316) ) then this (and session) could possibly by represented their.

> are there any specific span or event semantic conventions that we can add enduser.pseudo.id to?

No, not specifically as this really is just additional context details.

model/enduser/common.yaml
```diff
@@ -0,0 +1,22 @@
...
+        deprecated: "Removed."
+        requirement_level: recommended
+      - ref: enduser.scope
+        deprecated: "Removed."
+        requirement_level: recommended
+      - ref: enduser.authentication.id
+        requirement_level: required
+        note: >
+          The `enduser.authentication.id` attribute is intended to provide an unique identifier of an authenticated enduser.
+          The deprecated attributes `enduser.authentication.role` and `enduser.authentication.scope` are removed from the enduser registry.
```
trask, line 22: I was thinking that `enduser.id` would be the normal (authenticated) user id, do I remember that correctly? thanks

heyams: https://github.com/open-telemetry/semantic-conventions/pull/1456#discussion_r1880740782 
we decided to use `enduser.authentication.id` for the authenticated user id since the beginning of this discussion. 

renaming `enduser.id` to something specific (like `enduser.pseudo.id`) so that user don't put authenticated user id under this attribute.

model/enduser/registry.yaml
lmolkova: deprecated attributes should be moved to the deprecated folder. E.g. like here https://github.com/open-telemetry/semantic-conventions/blob/main/model/http/deprecated/registry-deprecated.yaml - you can find a bunch of other examples with https://github.com/search?q=repo%3Aopen-telemetry%2Fsemantic-conventions+%22deprecated%3A%22+language%3AYAML&type=code

model/enduser/registry.yaml
lmolkova: I believe having a document covering `enduser` convention as a whole was a critical feedback point.
We're trying to avoid adding attributes without also adding a signal that leverages it.

I'm surprised we don't have any signals to populate these on. Wouldn't we add them on session events? 
https://github.com/open-telemetry/semantic-conventions/blob/f7362c7066856ff8591ac461d4e3b31ad7af3a4b/docs/general/session.md

If we're going to just stamp them on all telemetry items, let's document it as an attribute group - [this would be the place](https://github.com/open-telemetry/semantic-conventions/blob/f7362c7066856ff8591ac461d4e3b31ad7af3a4b/docs/general/attributes.md#general-identity-attributes) to describe any additional guidance. This doc has to be updated anyway.

I believe it should cover:
- who populates those attributes - would an HTTP instrumentation do it? Distro? End user apps? 
- when they are populated? Whenever they are known? Is it opt-in? etc
- how to populate them - where would I get this information from? There are some mappings in existing docs - is it still relevant
- is user info propagated to other services (currently docs say it should be using baggage)

model/enduser/common.yaml
trask: can you summarize the reason for using `enduser.authentication.id` instead of `enduser.id`?

does this mean we will prohibit future "embedding" of `user.id` into the `enduser.*` namespace? or will we have 3 "id" attributes including `enduser.id`?

docs/general/attributes.md
```diff
@@ -414,34 +414,7 @@ system. It is expected this information would be propagated unchanged from node-
...
-[OAuth 2.0 Client Credentials Grant]: https://tools.ietf.org/html/rfc6749#section-4.4
-[OpenID Connect 1.0 IDToken]: https://openid.net/specs/openid-connect-core-1_0.html#IDToken
-[Kerberos]: https://tools.ietf.org/html/rfc4120
-[JavaEE/JakartaEE Servlet]: https://jakarta.ee/specifications/platform/8/apidocs/javax/servlet/http/HttpServletRequest.html
-[Windows Communication Foundation]: https://docs.microsoft.com/dotnet/api/system.servicemodel.servicesecuritycontext?view=netframework-4.8
-
-Given the sensitive nature of this information, SDKs and exporters SHOULD drop these attributes by
-default and then provide a configuration parameter to turn on retention for use cases where the
-information is required and would not violate any policies or regulations.
+`enduser.pseudo.id` attribute can be set by a specific client component, e.g. through a cookie out of the Span's HTTP request headers. Client side application should be able to stamp this attribute on any telemetry item emitted by the application whenever this cookie is available.
```
heyams, line 417: @lmolkova please let me know your thoughts on this statement.

model/enduser/common.yaml
heyams: we decided to have a sub-namespace called authentication so that we can add other attributes to do with authentication in the future. `enduser.id` is renamed to `enduser.pseudo.id` for unauthenticated/unauthorized/anonymous or any other kind of tracking id as long as it's not authenticated.

model/enduser/registry.yaml
heyams: done

model/enduser/registry.yaml
heyams: addressed [here](https://github.com/open-telemetry/semantic-conventions/pull/1456#discussion_r1913854373)

model/enduser/common.yaml
lmolkova: > we decided to have a sub-namespace called authentication so that we can add other attributes to do with authentication in the future.

I don't see any discussion around it or some motivation to introduce `enduser.authentication.id`.  Is there a link? Where did it come from?
If we need to add `authentication` namespace with some other properties in the future, we can always do it. 

The `enduser.id`  seems general and concise. I'd prefer to use it unless there are strong reasons not to

lmolkova, line 22:
```suggestion
```

I don't think that' accurate and also there is no need to mention changes in yaml that have no effect on markdown.

model/enduser/deprecated/registry-deprecated.yaml
```diff
@@ -2,23 +2,27 @@ groups:
   - id: registry.enduser.deprecated
     type: attribute_group
     display_name: Deprecated End User Attributes
-    brief: Describes deprecated enduser attributes. Complete enduser namespace has been deprecated
+    brief: "Describes deprecated end user attributes."
     attributes:
       - id: enduser.id
         type: string
+        brief: 'Deprecated, use `enduser.pseudo.id` instead.'
```
lmolkova, line 9: I believe `enuser.id` should stay as it is. Even if we call it `enduser.authentication.id`, the old `enduser.id` matches authenticated one.

model/enduser/registry.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - id: enduser.pseudo.id
+        type: string
+        stability: experimental
+        brief: >
+          Pseudonymous identifier of an end user. This identifier is unique to the user but does not reveal their actual identity.
+        examples: ['QdH5CAWJgqVT4rOr0qtumf']
+      - id: enduser.authentication.id
+        type: string
+        brief: "Unique identifier of an authenticated user in the system."
+        examples: [ 'S-1-5-21-202424912787-2692429404-2351956786-1000' ]
```
lmolkova, line 17: it seems to be an example of windows OS user ([SID](https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers)). Given that this attribute is intended to represent human identity, we should use a realistic example (login? email? guid?)

Also, we should add a note that it's likely a PII. I have a proposal on how to capture it (until we have a formal way in #1707)

https://github.com/open-telemetry/semantic-conventions/blob/d09cce343377962be7d6afe72d71d39081df3630/docs/general/how-to-define-semantic-conventions.md?plain=1#L70-L76

model/enduser/registry.yaml
```diff
@@ -0,0 +1,18 @@
+groups:
+  - id: registry.enduser
+    type: attribute_group
+    display_name: End User Attributes
+    brief: >
+      Describes information about the end user, which can be used as a subdomain of browser, client, or user domains.
+    attributes:
+      - id: enduser.pseudo.id
+        type: string
+        stability: experimental
```
lmolkova, line 10: we use development now, let's change it here and in other non-deprecated attributes

```suggestion
        stability: development 
```

model/enduser/registry.yaml
```diff
@@ -0,0 +1,18 @@
+groups:
+  - id: registry.enduser
+    type: attribute_group
+    display_name: End User Attributes
+    brief: >
+      Describes information about the end user, which can be used as a subdomain of browser, client, or user domains.
```
lmolkova, line 6:
```suggestion
      Describes the end user.
```


I don't think the other part is correct:

> which can be used as a subdomain of browser, client, or user domains.

model/enduser/deprecated/registry-deprecated.yaml
```diff
@@ -2,23 +2,27 @@ groups:
...
+        brief: 'Actual/assumed role the client is making the request under extracted from token or application security context.'
         examples: 'admin'
       - id: enduser.scope
         type: string
+        deprecated: "Removed."
         stability: experimental
-        deprecated: Removed.
-        brief: "Deprecated, no replacement at this time."
+        brief: >
+          Scopes or granted authorities the client currently possesses extracted from token
```
lmolkova, line 24: is there a need to change it? This attribute is deprecated.

model/enduser/deprecated/registry-deprecated.yaml
```diff
@@ -2,23 +2,27 @@ groups:
...
     attributes:
       - id: enduser.id
         type: string
+        brief: 'Deprecated, use `enduser.pseudo.id` instead.'
         stability: experimental
-        deprecated: Replaced by `user.id` attribute.
-        brief: "Deprecated, use `user.id` instead."
-        examples: 'username'
+        deprecated: "Replaced by `enduser.pseudo.id`."
+        examples: ['QdH5CAWJgqVT4rOr0qtumf']
```
lmolkova, line 12: please bring the old example back

model/enduser/common.yaml
```diff
@@ -0,0 +1,22 @@
...
+    type: attribute_group
+    brief: >
+      These attributes may be used for any operation with an authenticated and/or authorized enduser.
+    attributes:
+      - ref: enduser.id
+        deprecated: Replaced by `enduser.pseudo.id` attribute.
+        requirement_level: recommended
+      - ref: enduser.pseudo.id
+        requirement_level: recommended
+      - ref: enduser.role
```
lmolkova, line 12: please remove it from this group - it's listed as deprecated attribute and is not necessary here

model/enduser/common.yaml
```diff
@@ -0,0 +1,22 @@
...
+        deprecated: "Removed."
+        requirement_level: recommended
+      - ref: enduser.scope
```
lmolkova, line 15: same here, please remove the reference

model/enduser/common.yaml
```diff
@@ -0,0 +1,22 @@
+groups:
+  - id: identity
+    type: attribute_group
+    brief: >
+      These attributes may be used for any operation with an authenticated and/or authorized enduser.
```
lmolkova, line 5:
```suggestion
      Describes end user identity. 
```

model/enduser/common.yaml
```diff
@@ -0,0 +1,22 @@
...
+      - ref: enduser.pseudo.id
+        requirement_level: recommended
+      - ref: enduser.role
+        deprecated: "Removed."
+        requirement_level: recommended
+      - ref: enduser.scope
+        deprecated: "Removed."
+        requirement_level: recommended
+      - ref: enduser.authentication.id
+        requirement_level: required
```
lmolkova, line 19: I don't think it can be required since it's likely to contain PII

docs/general/attributes.md
lmolkova: I don't know if my suggestion is factually correct - please adjust it, but I'd phrase it differently - something along the following lines:

> Enduser attributes capture end user identity. They are likely to contain PII and should be populated, processed, and stored with caution. 
Information about the end user is usually available on the client side (in a mobile application of browser). Enduser attributes are populated by the user application in coordination with OpenTelemetry SDK. Some OpenTelemetry distributions auto-collect this information from HTTP cookies.
When user information is available, it's RECOMMENDED to add it to all spans and events emitted in the scope of operation initiated by this user. 

> Application in coordination with OpenTelemetry SDK and Distro MAY propagate user information from the client application and across different services using custom HTTP cookies or [Baggage](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.40.0/specification/baggage/api.md). 

> Enduser information is collected and populated by specialized components, 
> other instrumentations such as HTTP or RPC are not expected to populate these attributes by default. 

> \<!-- TODO: add link to the implementation in otel-js-contirb --\>


I would also put it above the table.

I think it should also replace the content of lines 413-415 above

> It is expected this information would be propagated unchanged from node-to-node within the system
using the Baggage mechanism. These attributes should not be used to record system-to-system
authentication attributes.

docs/general/attributes.md
```diff
@@ -410,34 +414,7 @@ system. It is expected this information would be propagated unchanged from node-
 using the Baggage mechanism. These attributes should not be used to record system-to-system
 authentication attributes.
 
-Examples of where the `enduser.id` value is extracted from:
-
-| Authentication protocol | Field or description            |
```
lmolkova, line 415: I would prefer to keep this table - it actually explains what the `enduser.id` is - we don't provide an explanation like this anymore. Is this information still accurate and can it be used to capture `enduser.id` ?

heyams: `enduser.id` has been replaced with 'enduser.pseudo.id`. i don't think it makes sense to keep it.

model/enduser/common.yaml
heyams: yes, `enduse.pseudo.id` can be unauthenticated, unauthorized, anonymous, and etc.. as long as it's not authenticated, which is tracked under `enduser.authentication.id`.

model/enduser/common.yaml
heyams: will remove it

model/enduser/deprecated/registry-deprecated.yaml
heyams: the main reason we came up with `enduser.pseudo.id` or a list of other naming options because we don't want customers to put authenticated user id to `enduser.id`. authenticated one is using `enduser.authentication.id`.

model/enduser/common.yaml
heyams: https://github.com/open-telemetry/semantic-conventions/pull/1146

heyams: that was the 1st thing i did on this long-discussed thread.

trask: #1146 was never approved

model/enduser/deprecated/registry-deprecated.yaml
lmolkova: the ask is to keep `enuser.id` instead of adding `enuser.authentication.id`. This PR would add `enuser.pseudo.id` and un-deprecate `enuser.id`

model/enduser/common.yaml
heyams: but this PR is based on that PR.

model/enduser/deprecated/registry-deprecated.yaml
heyams: that wasn't what we discussed though. what about the sub-namespace authentication? i thought that is what we want to do so that any other attributes associated with `authentication` can be added later. 

if you want to change it now, that will be different. we would need to re-discuss it. i thought the consensus is to have authentication sub namespace.

heyams: https://github.com/open-telemetry/semantic-conventions/issues/1104#issuecomment-2259426406 it was from you @lmolkova

lmolkova: it's been under the assumption that we use `user.id` for pseudo/tracking/anonymous id - as you can see we've reconsidered it after that.

> if you want to change it now, that will be different. we would need to re-discuss it.

It seems at least me and @trask are on the same page to use `enduser.id` instead of `enduser.authentication.id`. I believe I saw some support from @trisch-me and @jsuereth on it too. 

Please make the change and if anyone has objections they will be able to comment and provide reasoning.

heyams:
For people who are flip-flopping on what we discussed about sub namespace `authentication`, a friendly reminder to revisit these links:

https://github.com/open-telemetry/semantic-conventions/issues/1104#issuecomment-2259426406
https://github.com/open-telemetry/semantic-conventions/issues/1104#issuecomment-2260106290

https://github.com/open-telemetry/semantic-conventions/issues/1172
image

trask:
> For people who are flip-flopping

hi @heyams, I don't see any arguments in those links for specifically preferring `enduser.authenticated.id` over `enduser.id`?

in order to help this PR move forward, can you write out the reasons you have for specifically preferring `enduser.authenticated.id` over `enduser.id`? (other than that you thought it was decided already)

trask:
@heyams I still don't see any reasons in your last comment for specifically preferring `enduser.authenticated.id` over `enduser.id`? (other than that you thought it was decided already)

heyams:
> hey @heyams
> 
> it might be a long thread but I think we all want to see it merged. We have discussed this issue on semconv meeting this Monday and as @trask mentioned above (and to my personal preference) I am in favor of having just `id` without additional subdomain. And I don’t see clear preference for the other option.
> 
> I remember we have discussed in the beginning that it might be good to have additional sub namespace for `authentication` attributes and put there `id` and other attributes, but I think we can start simple without specific subdomain until we have a use case for it.
> 
> please let me know if it would work for you also everyone else please chime in if you think this isn't working for some cases

I was planning to attend the SIG to get everyone's feedback. Monday was a public holiday here. If that was the consensus from everyone from the SIG to reuse enduser.id for the authenticated user id, I'm good with it. I can go ahead to make a change this week and then discuss it further in the next SIG if anyone else objects.

i appreciate that you reach out because this sub-namespace was your original idea.

trisch-me:
hey @heyams - sure, back then we wanted to find a solution on how to manage both auth and not auth id, and that was one of ideas, but now we have already `user.pseudo.id`, so I believe we can proceed with just `user.id`

model/enduser/deprecated/registry-deprecated.yaml
```diff
@@ -2,23 +2,17 @@ groups:
...
-        deprecated: Replaced by `user.id` attribute.
-        brief: "Deprecated, use `user.id` instead."
-        examples: 'username'
       - id: enduser.role
         type: string
+        deprecated: "Removed."
         stability: experimental
-        deprecated: Replaced by `user.roles` attribute.
-        brief: "Deprecated, use `user.roles` instead."
+        brief: 'Actual/assumed role the client is making the request under extracted from token or application security context.'
```
trisch-me, line 11: nit: we usually write about deprecation in brief as well
```suggestion
        brief: "Deprecated, no replacement at this time"
```

trisch-me (approved):

trisch-me:
@heyams there are still conflicts not resolved

MSNev (approved):
Provided comments offline about possible rewording of the description for `enduser.pseudo.id` specifically to remove the "reveal" wording. But this would be  a refinement and not a blocker.

heyams:
@joaopgrassi i don't have permission to dismiss your "request for change". Your comment no longer applied. 
please review it again and unblock this PR.

model/enduser/registry.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - id: registry.enduser
+    type: attribute_group
+    display_name: End User Attributes
+    brief: >
+      Describes the end user.
+    attributes:
+      - id: enduser.id
+        type: string
+        brief: "Unique identifier of an authenticated user in the system."
+        examples: [ 'username' ]
```
lmolkova, line 11: please add a note that it usually contains PII. E.g. like this
```suggestion
        note: |
          > [!Warning]
          > This field contains sensitive (PII) information.
        examples: [ 'username' ]
```

model/enduser/registry.yaml
```diff
@@ -0,0 +1,18 @@
+groups:
+  - id: registry.enduser
+    type: attribute_group
+    display_name: End User Attributes
+    brief: >
+      Describes the end user.
+    attributes:
+      - id: enduser.id
+        type: string
+        brief: "Unique identifier of an authenticated user in the system."
```
lmolkova, line 10: Do we have any recommendations on what this id should be ? login? email? anything user app wants to use?

Please adjust this note based on the answer

```suggestion
        brief: "Uniquely identifies end user in the system. It may be a login, email address, or another id used for authentication"
```

model/enduser/registry.yaml
```diff
@@ -0,0 +1,18 @@
...
+        examples: [ 'username' ]
+        stability: development
+      - id: enduser.pseudo.id
+        type: string
+        stability: development
+        brief: >
+          Pseudonymous identifier of an end user. This identifier should be a random value that is not directly linked or associated with the end user's actual identity.
+        examples: ['QdH5CAWJgqVT4rOr0qtumf']
```
lmolkova, line 18: please add a note on PII
```suggestion
        note: |
          > [!Warning]
          > This field contains sensitive (linkable PII) information.
        examples: ['QdH5CAWJgqVT4rOr0qtumf']
```

docs/general/attributes.md
```diff
@@ -439,6 +438,19 @@ Given the sensitive nature of this information, SDKs and exporters SHOULD drop t
 default and then provide a configuration parameter to turn on retention for use cases where the
 information is required and would not violate any policies or regulations.
```
lmolkova, line 440: Please review and update unchanged section (lines 407-440). I think it has at least a couple of out-of-date points

> It is expected this information would be propagated unchanged from node-to-node within the system
using the Baggage mechanism.

It is? Do we expect user apps to propagate those? I don't think we could recommend it since we can't stop `baggage` from leaking everywhere.

> Given the sensitive nature of this information, SDKs and exporters SHOULD drop these attributes by
default and then provide a configuration parameter to turn on retention for use cases where the
information is required and would not violate any policies or regulations.

Is it what client SDK / distros are going to do? Should we say that it's user app who enables user tracking and they should consult with local policies and regulations before enabling corresponding instrumentation/client SDK?

heyams: i am not sure. it was there before. if you prefer a different text, please suggest what should i update it to. we can delete the whole thing if it is no longer needed.

model/enduser/deprecated/registry-deprecated.yaml
```diff
@@ -2,23 +2,17 @@ groups:
...
-        deprecated: Replaced by `user.id` attribute.
-        brief: "Deprecated, use `user.id` instead."
-        examples: 'username'
       - id: enduser.role
         type: string
+        deprecated: "Removed."
         stability: experimental
-        deprecated: Replaced by `user.roles` attribute.
-        brief: "Deprecated, use `user.roles` instead."
+        brief: 'Deprecated, no replacement at this time.'
```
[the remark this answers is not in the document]
heyams: reverted. it was suggested by one of the reviewers.

.chloggen/add_enduser_pseudo_id.yaml
```diff
@@ -0,0 +1,23 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: enduser
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: introduce new attribute `enduser.pseudo.id`
```
trask, line 13:
```suggestion
note: Introduce new attribute `enduser.pseudo.id`
```

docs/general/attributes.md
```diff
@@ -396,9 +396,14 @@ These attributes may be used for any operation with an authenticated and/or auth
...
-| [`enduser.scope`](/docs/attributes-registry/enduser.md) | string | Deprecated, no replacement at this time. | `read:message, write:files` | `Recommended` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Removed. |
+| [`enduser.id`](/docs/attributes-registry/enduser.md) | string | Unique identifier of an end user in the system. It maybe a username, email address, or other identifier. [1] | `username` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`enduser.pseudo.id`](/docs/attributes-registry/enduser.md) | string | Pseudonymous identifier of an end user. This identifier should be a random value that is not directly linked or associated with the end user's actual identity. [2] | `QdH5CAWJgqVT4rOr0qtumf` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1] `enduser.id`:** > [!Warning]
+> This field contains sensitive (PII) information.
+
+**[2] `enduser.pseudo.id`:** > [!Warning]
+> This field contains sensitive (linkable PII) information.
```
trask, line 406: this isn't rendering well:

image

model/enduser/registry.yaml
```diff
@@ -0,0 +1,24 @@
...
+    display_name: End User Attributes
+    brief: >
+      Describes the end user.
+    attributes:
+      - id: enduser.id
+        type: string
+        brief: "Unique identifier of an end user in the system. It maybe a username, email address, or other identifier."
+        note: |
+          > [!Warning]
+          > This field contains sensitive (PII) information.
```
trask, line 13: i'm not sure, but try adding a newline here to fix the rendering issue:

```suggestion
        note: |

          > [!Warning]
          > This field contains sensitive (PII) information.
```

docs/general/attributes.md
heyams: Updated.

docs/general/attributes.md
```diff
@@ -412,9 +412,9 @@ These attributes may be used for any operation with an authenticated and/or auth
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
-These attributes describe the authenticated user driving the user agent making requests to the instrumented
+`enduser.id` describe the authenticated user driving the user agent making requests to the instrumented
 system. It is expected this information would be propagated unchanged from node-to-node within the system
-using the Baggage mechanism. These attributes should not be used to record system-to-system
+using the Baggage mechanism. This attribute should not be used to record system-to-system
```
lmolkova, line 417: is this expected? would client SDK populate them on the baggage? if not, let's just remove it. This was deprecated before, we should document the new behavior.

heyams: removed.

model/enduser/deprecated/registry-deprecated.yaml
```diff
@@ -2,23 +2,17 @@ groups:
...
     attributes:
-      - id: enduser.id
-        type: string
-        stability: experimental
-        deprecated: Replaced by `user.id` attribute.
-        brief: "Deprecated, use `user.id` instead."
-        examples: 'username'
       - id: enduser.role
         type: string
+        deprecated: "Removed."
```
lmolkova, line 9: Any reason to change this one?

```suggestion
        deprecated: Replaced by `user.roles` attribute.
```

lmolkova (approved):

model/enduser/deprecated/registry-deprecated.yaml
heyams: i can revert it.

docs/general/attributes.md
```diff
@@ -394,19 +394,26 @@ These attributes may be used for any operation with an authenticated and/or auth
...
 <!-- endsemconv -->
 
-These attributes describe the authenticated user driving the user agent making requests to the instrumented
-system. It is expected this information would be propagated unchanged from node-to-node within the system
-using the Baggage mechanism. These attributes should not be used to record system-to-system
-authentication attributes.
+`enduser.id` describe the authenticated user driving the user agent making requests to the instrumented
+system.
 
 Examples of where the `enduser.id` value is extracted from:
```
trask, line 419: I wonder if generic instrumentation should be capturing `user.id` instead of `enduser.id` in the examples below (since generic instrumentation wouldn't know whether it's an "end user" vs some other kind of identity)?

heyams: existing General identity attributes never mentioned `user` namespace. 
i think it might require more discussion like wording, and it doesn't need to block this PR. This is for authenticated `enduser.id`. We can add a sub-section for user attributes later once wording is decided.

trask: I think I'd be in favor of removing this entire section: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attributes.md#general-identity-attributes

I think better to remove than to have likely incorrect language (now that enduser is being undeprecated)

heyams: done

model/enduser/common.yaml
trask, line 1: I think this file can be deleted, was only being used to generate the table in the just deleted section

docs/general/attributes.md
```diff
@@ -381,62 +380,6 @@ Examples of `peer.service` that users may specify:
...
-
-<!-- markdownlint-restore -->
-<!-- prettier-ignore-end -->
-<!-- END AUTOGENERATED TEXT -->
-<!-- endsemconv -->
-
-These attributes describe the authenticated user driving the user agent making requests to the instrumented
-system. It is expected this information would be propagated unchanged from node-to-node within the system
-using the Baggage mechanism. These attributes should not be used to record system-to-system
-authentication attributes.
```
trask, line 409: @MSNev is this important to retain somewhere? if so we could move it to the attribute definitions, or maybe we need an `docs/enduser/README.md` file

docs/general/attributes.md
```diff
@@ -381,62 +380,6 @@ Examples of `peer.service` that users may specify:
...
-[OAuth 2.0 Client Identifier]: https://tools.ietf.org/html/rfc6749#section-2.2
-[OAuth 2.0 Client Credentials Grant]: https://tools.ietf.org/html/rfc6749#section-4.4
-[OpenID Connect 1.0 IDToken]: https://openid.net/specs/openid-connect-core-1_0.html#IDToken
-[Kerberos]: https://tools.ietf.org/html/rfc4120
-[JavaEE/JakartaEE Servlet]: https://jakarta.ee/specifications/platform/8/apidocs/javax/servlet/http/HttpServletRequest.html
-[Windows Communication Foundation]: https://docs.microsoft.com/dotnet/api/system.servicemodel.servicesecuritycontext?view=netframework-4.8
-
-Given the sensitive nature of this information, SDKs and exporters SHOULD drop these attributes by
-default and then provide a configuration parameter to turn on retention for use cases where the
-information is required and would not violate any policies or regulations.
```
trask, line 438: @MSNev do we need to move this also somewhere? was it referring to the client-side SDK, or referring to server-side SDKs that are reading this out of baggage?

docs/general/attributes.md
MSNev: Note that I'm aware of the "user agent" reference here doesn't seem to reference anything to do with the browser environment.

docs/general/attributes.md
MSNev: Likewise don't know, these are unrelated to anything specifically for "clients", this probably got moved from the previous renaming / reusing of the enduser for the authenticated user.

.chloggen/add_enduser_pseudo_id.yaml
```diff
@@ -0,0 +1,23 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: enduser
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Introduce new attribute `enduser.pseudo.id`
```
trask, line 13:
```suggestion
note: Undeprecate `enduser.id` and introduce new attribute `enduser.pseudo.id`
```

.chloggen/add_enduser_pseudo_id.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [1104]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The new attribute `enduser.pseudo.id` is intended to provide an unique identifier of a pseudonymous enduser.
```
trask, line 23:
```suggestion
  The new attribute `enduser.pseudo.id` is intended to provide a unique identifier of a pseudonymous enduser.
```

trask (approved):

MSNev (approved):

jsuereth (approved):
Thanks for updating this! I realize it took a long time to get to it.

I do agree with both of trask's rewording suggestions, but otherwise this looks good as is.

docs/general/attributes.md
```diff
@@ -381,62 +380,6 @@ Examples of `peer.service` that users may specify:
 - A Redis cache of auth tokens as `peer.service="AuthTokenCache"`.
 - A gRPC service `rpc.service="io.opentelemetry.AuthService"` may be hosted in both a gateway, `peer.service="ExternalApiService"` and a backend, `peer.service="AuthService"`.
 
-## General identity attributes
```
chalin, line 384: With this link-target gone, what should be used instead?

PR #1461, opened 2024-10-09
[chore] uncomment device.app.lifecycle event definition

MSNev:
Now that Semantic Conventions supports `event` yaml definitions, this PR is uncommenting the `device.app.lifecycle` event definition so that it is defined in yaml and ready for markdown generation.

Note: Once committed this will block anyone consuming the updated SemConv version that has not yet migrated to using weaver.


lmolkova (approved):

LikeTheSalad (approved):

joaopgrassi:
> Note: Once committed this will block anyone consuming the updated SemConv version that has not yet migrated to using weaver.

Then this cannot be merged, right? Because there are still SIGs that haven't yet migrated to Weaver: https://github.com/open-telemetry/weaver/issues/227

lmolkova:
@joaopgrassi that's a great point! I agree we should not add features that break build-tools until the migration is finished

I just tried if build-tools will fail on this change and it doesn't - https://github.com/lmolkova/opentelemetry-python/pull/3/

Build-tools has a mode to ignore validation errors and warns with `[Warning] /source/device/events.yaml: Invalid keys: ['body'] - @12:5 ('device.app.lifecycle')`.

As a result, nobody can generate event bodies with build-tools, but that's a new feature and nobody generated them in the past (and won't be able to).

BTW we still have a CI check - `semantic-conventions-compatibility`  that runs build-tools in this mode and we can use it to see if new feature introduces any unrecoverable problems.

PR #1464, opened 2024-10-10
Render events with bodies

lmolkova:
- Add templates to render event bodies 
- Updates to existing events:
  - Add missing stability (`stable`) on exception span event
  - Update RPC, device lifecycle, and feature flags events - move brief/notes from markdown to yaml.
  - Uncomment azure resource log definition


templates/registry/markdown/snippet.md.j2
```diff
@@ -4,7 +4,7 @@
 <!-- markdownlint-capture -->
 <!-- markdownlint-disable -->
 {% import 'attribute_table.j2' as at %}{% import 'metric_table.j2' as mt %}{% import 'event_macros.j2' as event %}{% import 'resource_macros.j2' as resource %}
-{% if group.type == "event" %}{{ event.header(group) }}{% elif group.type == "resource" %}{{ resource.header(group) }}{% elif group.type=="metric" %}{{ mt.generate(group) }}{% endif %}{{ at.generate(group.attributes, tag_filter, attribute_registry_base_url, group.lineage.attributes) }}
+{% if group.type == "event" %}{{ event.header(group) }}{% elif group.type == "resource" %}{{ resource.header(group) }}{% elif group.type=="metric" %}{{ mt.generate(group) }}{% endif %}{{ at.generate(group.attributes, tag_filter, attribute_registry_base_url, group.lineage.attributes) }}{% if group.type == "event" %}{{ event.body(group.body) }}{% endif %}
```
jsuereth, line 7: Nit: Should we move to a structure like:
```
{% if group.type == "X"}x_snippet(group, ...){% elif group.type == "Y"}y_snippet(group,...){% else %} attribute_table(group,...){% endif %}
```

And then define local macros for `x_snippet`, `y_snippet` etc?

templates/registry/markdown/event_macros.j2
```diff
@@ -1,4 +1,19 @@
 {#- Macros for simplifying creating "Event" documentation. -#}
-{% macro header(event) %}{% if event.name %}The event name MUST be `{{ event.name }}`.
+{% import 'stability.j2' as stability %}
+{% import 'body_field_table.j2' as body_table %}
+{% macro header(event) %}
 
+{% if event.name %}The event name MUST be `{{ event.name }}`.
+
+**Status:** {{ stability.badge(event.stability, event.deprecated) }}
```
jsuereth, line 8: Nit: Similar to resource - Are we ok having attributes with different stability in a stable event?

Personally I think it's ok, but wonder if we should officially decide that as a group.

templates/registry/markdown/examples_macros.j2
```diff
@@ -0,0 +1,17 @@
+{% macro print_examples(examples) %}{%- for e in examples %}{%if loop.first == false %}; {% endif %}`{{ e | trim }}`{%- endfor %}{% endmacro %}
+
+{% macro format(item) %}{%- if item.examples %}
```
jsuereth, line 3: Nice cleanup!!!!!

templates/registry/markdown/body_field_table.j2
```diff
@@ -0,0 +1,15 @@
+{% import 'requirement.j2' as requirement %}
+{% import 'stability.j2' as stability %}
+{% import 'notes.j2' as notes %}
+{% import 'enum_macros.j2' as enums %}
+{% import 'examples_macros.j2' as examples %}
+{% macro flatten(fields, ns, depth) %}{% if fields %}{% for f in fields | sort(attribute="id") %}
+{% set ns.flat = [ns.flat, [{'field':f,'depth':depth}]] | flatten %}{% if f.fields %}{% set _= flatten(f.fields, ns, depth + 1) %}{% endif %}
```
jsuereth, line 7: This makes me smile.

Can you open a ticket against Weaver to provide this as a built-in filter we can use on fields?

E.g. something like:

``` 
{% for field, depth in fields | sort_by_body_field_depth %}
```

docs/rpc/rpc-spans.md
```diff
@@ -254,9 +255,7 @@ different processes could be listening on TCP port 12345 and UDP port 12345.
 
 ### Events
 
-In the lifetime of an RPC stream, an event for each message sent/received on
-client and server spans SHOULD be created. In case of unary calls only one sent
-and one received message will be recorded for both client and server spans.
+### Message event
```
lmolkova, line 258:
```suggestion
#### Message event
```

templates/registry/markdown/body_field_table.j2
lmolkova: great point! https://github.com/open-telemetry/weaver/issues/410

model/azure/logs.yaml
```diff
@@ -13,90 +13,79 @@ groups:
         brief: The [Fully Qualified Azure Resource ID](https://docs.microsoft.com/rest/api/resources/resources/get-by-id) the log is emitted for.
         note: ""
       - ref: event.name
```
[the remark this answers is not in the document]
MSNev: As part of the definition of an `event` the top level `name` defines the expected value for the `event.name` attribute (for log based events).

This is defined in yaml like this [per the `device.app.lifecycle`](https://github.com/MSNev/semantic-conventions/blob/b7b512928ee2faf64e1b128d5093aa197174c1a1/model/device/events.yaml#L5)

And enforced in weaver via [this "check" in weaver](https://github.com/open-telemetry/weaver/blob/402583b906e05c75dc12979e537dde9923a4e7d8/crates/weaver_semconv/src/group.rs#L123-L139) which enforces that when a `body` is present you must specify the `name` property for the `event`.

We should consider a policy that blocks adding this to an `event` definition. And I guess the MarkDown generation should reuse this property to automatically populate the text [like this](https://github.com/MSNev/semantic-conventions/blob/b7b512928ee2faf64e1b128d5093aa197174c1a1/docs/mobile/events.md#event-details)
```markdown
The event name MUST be `devide.app.lifecycle`.
```

docs/azure/events.md
```diff
@@ -28,30 +30,36 @@ The event name MUST be `az.resource.log`.
...
+**[1]:** Typically, this field includes the authorization and claims or JWT token from Active Directory.
+
+> [!Warning]
+> This field contains sensitive (PII) information.
+
+**[2]:** if the event is tied to an Active Directory tenant.
+
+
+
+
```
joaopgrassi, line 58: Oh no white spaces are back 😅

templates/registry/markdown/event_macros.j2
lmolkova: let's keep discussing it in the https://github.com/open-telemetry/semantic-conventions/pull/1472 - this pr does not change the status quo.

model/azure/logs.yaml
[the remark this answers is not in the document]
lmolkova: removed `event.name` from this event and added a policy check

docs/azure/events.md
lmolkova: fixed! 🙈

PR #1469, opened 2024-10-11
GenAI events - yamlify body definitions and cosmetic examples improvements

lmolkova:
- Adds yaml definitions for GenAI events in preparation for https://github.com/open-telemetry/semantic-conventions/pull/1464
- Updates examples to be more concise and readable, fixes small inconsistencies, adds diagrams.


model/gen-ai/events.yaml
```diff
@@ -46,3 +180,111 @@ groups:
...
+              fields:
+                - id: name
+                  type: string
+                  stability: experimental
+                  brief: >
+                    The name of the function.
+                  examples: ["get_weather"]
+                  requirement_level: required
+                - id: arguments
+                  type: undefined
```
lmolkova, line 286: let's clarify if these should be deserialized or provided as string. 
The perf cost of deserialization compared to all other things necessary to create this event is small.

trentm: A concern: In general the "arguments" string returned by (at least OpenAI -- my experience is limited) may not be valid JSON. In the (likely rare) case that it isn't this means that "arguments" would then, presumably, fallback to be a string. That could be a rare surprise (i.e. possible bug) to whatever system is processing this telemetry.

I suppose that is fine if the type allows both string and object (or perhaps `AnyValue` would be used) -- then implementations need to cope.

---

Another possible concern: https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format includes the (surprising to me) warning:

> Important: when using JSON mode, you must also instruct the model to produce JSON yourself via a system or user message. Without this, the model may generate an unending stream of whitespace until the generation reaches the token limit, resulting in a long-running and seemingly "stuck" request.

Not having the raw "arguments" string in the telemetry could make it hard to debug why a parsed JSON response of, say `{"foo": "bar"}`, took 1024 (or however many) tokens.

---

What are reasonable use cases for having the parsed arguments in telemetry?

lmolkova: if max token is reached in the middle of the arguments, it's truncated, so let's keep it a string, but we might change it for some systems.

model/gen-ai/events.yaml
```diff
@@ -30,6 +70,73 @@ groups:
     brief: >
       This event describes the assistant message passed to GenAI system or received from it.
     extends: gen_ai.common.event.attributes
+    body:
```
lmolkova, line 73: TODO: add generated md table into the description

model/gen-ai/events.yaml
xrmx: Then we should also update the examples above

lmolkova: Thank you both, added a note and updated examples.

xrmx (approved):

model/gen-ai/events.yaml
```diff
@@ -30,6 +70,79 @@ groups:
...
+        - id: role
+          type: string
+          stability: experimental
+          brief: >
+            The actual role of the message author as passed in the message.
+          examples: ["assistant", "bot"]
+          requirement_level:
+            conditionally_required: if available and not equal to `assistant`.
+        - id: tool_calls
+          type: map  # TODO: it's an array
```
lmolkova, line 94: TODO: this is blocker - need to release new weaver that supports it

nirga (approved):

model/gen-ai/events.yaml
aabmass: Is this still blocked and the PR ready for review?

lmolkova: the PR is ready for review, but the tooling is not ready for us to define array of map type. So even if you review, we can't merge it yet

lmolkova: fixed now that tooling supports it

model/gen-ai/events.yaml
```diff
@@ -14,6 +14,26 @@ groups:
...
+          examples: ["You're a helpful bot"]
+          requirement_level: opt_in
+        - id: role
+          type: string
+          stability: experimental
+          brief: >
+            The actual role of the message author as passed in the message.
+          examples: ["system", "instruction"]
+          requirement_level:
+            conditionally_required: if available and not equal to `system`.`
```
codefromthecrypt, line 36:
```suggestion
            conditionally_required: if available and not equal to `system`.
```

codefromthecrypt (approved):
thanks

model/gen-ai/events.yaml
```diff
@@ -14,6 +14,26 @@ groups:
     brief: >
       This event describes the instructions passed to the GenAI system inside the prompt.
```
drewby, line 15:
```suggestion
      This event describes the system instructions passed inside the prompt to the GenAI model.
```
Reading through the markdown beginning to end, I realized the three message descriptions could be more precise.

model/gen-ai/events.yaml
```diff
@@ -22,6 +42,26 @@ groups:
     brief: >
       This event describes the prompt message specified by the user.
```
drewby, line 43:
```suggestion
      This event describes the user message passed inside the prompt to the GenAI model.
```
Per my previous comment.

model/gen-ai/events.yaml
```diff
@@ -30,6 +70,79 @@ groups:
     brief: >
       This event describes the assistant message passed to GenAI system or received from it.
```
drewby, line 71:
```suggestion
      This event describes the assistant instructions passed inside the prompt to the GenAI model.
```
On this one, I thought the expectation is not that the response from the model is added as a `gen_ai.assistant.message` event. Rather, this is a previous `gen_ai.choice` added into a new prompt to create a multi-turn interaction in the stateless communication.  The "or received from it" at the end of the sentence could be confusing.

model/gen-ai/events.yaml
```diff
@@ -38,6 +151,33 @@ groups:
     brief: >
       This event describes the tool or function response message.
```
drewby, line 152:
```suggestion
      This event describes the response from a tool or function call passed inside the prompt to the GenAI model.
```
Clear that this is passed to the model, not received from it.

drewby (changes requested):
Overall looks good. While the examples at the end of the markdown help clarify, I think the short descriptions of the input messages by role can be clearer.

jsuereth (approved):
Reviewing for general semconv.

Overall, glad we can now put the ToolCall structure/semantics in the YAML.  I think this is a good first step.

I do think *seeing* the hierarchy with the spaces in markdown still leaves a bit wanting.  I'm not sure what options we have with markdown here, unfortunately.  

I like the look of what prometheus docs have done [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md) but this requires dropping into HTML table land, and it ruins the readabiltiy of the markdown by itself.  That *may* be ok if it's autogenerated.

lmolkova:
> I do think seeing the hierarchy with the spaces in markdown still leaves a bit wanting. I'm not sure what options we have with markdown here, unfortunately.

We could switch to (pseudo) code or some form of JSON, I've heard you like designing languages, @jsuereth  :)

model/gen-ai/events.yaml
lmolkova: thanks! I want to avoid "passed inside the prompt" since we should be able to reuse the same messages for something like OpenAI assistants, which can take instructions at the assistant creation time. I'm taking the rest though

model/gen-ai/events.yaml
lmolkova: I want to avoid "passed in the prompt" here too, since we should be able to reuse message events for assistants

model/gen-ai/events.yaml
lmolkova: fixed in https://github.com/open-telemetry/semantic-conventions/pull/1469/commits/557b99dc09eb08565c501460a1fada716ed050e4 with the same caveat about "passed inside the prompt"

model/gen-ai/events.yaml
lmolkova: fixed in https://github.com/open-telemetry/semantic-conventions/pull/1469/commits/a25e181f025d7f195bddb3691930f757e8c1f4bf with the same caveat about "passed inside the prompt"

instructions can be misleading -it's a synonym to system message

drewby (approved):
Looks good!

PR #768, opened 2024-02-23
Database: clarify network attributes usage on common semconv

lmolkova:
## Changes

Addresses DB portion of https://github.com/open-telemetry/semantic-conventions/issues/690:

- Removes `network.transport|type` attributes from general db semantic conventions.
- Clarifies when specific systems should set `network.server.address|port` and keeps them on the cassandra, redis and elasticsearch extensions

Based on the findings from Java instrumentation libraries, following instrumentations populate network attributes:
- cassandra
- redis (Redisson, jedis, lettuce, vertx, redis)
- elastic
- opensearch

See https://github.com/open-telemetry/semantic-conventions/pull/698#issuecomment-1958869759 for more details.

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* ~~[schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.~~

Issue #724, opened 2024-02-09
Is `db.connection_string` necessary?


lmolkova:
Connection strings can have secrets and have very specific format which makes them not that useful for visualization tooling (or user queries). 
They can also be long and redundant on every span.

They can (secrets redacted) be useful to log to help investigate configuration issues, but probably should not appear on spans/metrics.

Proposal is to remove the attribute or make it opt-in.

trask:
@open-telemetry/java-instrumentation-approvers do you know of anything important being done with `db.connection_string` that we are capturing in Java instrumentation?

lmolkova:
Looking into https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-java-instrumentation+db_connection_string&type=code&p=1, it seems database instrumentation either  don't set connection string or set driver:schema://host:port. 

host and port are covered with `server.address|port` and we can record schema with `url.scheme` (maybe for specific systems)?
 We can also introduce attribute for driver (when applicable)

joaopgrassi:
Maybe we should ping other maintainers here to get a final opinion?

trask:
cc @open-telemetry/go-approvers @open-telemetry/javascript-approvers @open-telemetry/python-approvers @open-telemetry/opentelemetry-python-contrib-approvers @open-telemetry/dotnet-approvers @open-telemetry/dotnet-contrib-approvers

pellared:
Connection strings can have passwords. 

Example from Go:  https://github.com/go-sql-driver/mysql/?tab=readme-ov-file#dsn-data-source-name

Example from .NET: https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/connection-string-syntax

**Sometimes** the libraries have connection string parsing function that could allow redacting. I think it should be **at least** opt-in.

breedx-splk:
We think that changing the default behavior to omit/exclude the `db.connection_string` attribute will cause immediate degradation of some of our product features.  My preference would be to have the default include the attribute but somehow scrub passwords.

trask:
@breedx-splk can you provide a bit more details about how you are using `db.connection_string`? thanks

trask:
also, is this only for Java customers, or includes other languages as well?

pichlermarc:
I'd also prefer to drop `db.connection_string` or make it opt-in.

[...]

Issue #690, opened 2024-02-05
DB/messaging/rpc(?): should logical operations include `network.*` attributes


lmolkova:
DB, messaging (maybe rpc and other conventions) arguably define logical operations.

E.g. database call could be a complex operation involving multiple underlying network calls (auth, retries, check-if-exists, etc).

- **Assuming the underlying protocol is instrumented, it introduces duplication and confusion**

- **Some systems can operate on top of different interchangeable protocols and network information can be missing/too verbose/irrelevant:**
  * GCP frequently operates on HTTP and gRPC 
  * Azure messaging clients use AMQP, but to some extent can operate on top of HTTP
  * ORM systems don't  even have to know the network details of underlying DB driver.

- **Retries are ambiguous**
  * retries can have different `network.peer|local.address|port` if they reach different nodes of the service

- **Network calls within one logical call can use different protocols, e.g. in case of Azure CosmosDB:**
  * Auth can be HTTP
  * Some other rare operations are HTTP
  * Things on the hot path are done over proprietary TCP-based protocol

**Proposal:**

Remove `network.*` attributes from generic db/messaging logical operations:
- this would make `network.*` attributes opt-in
- logical operations should still use 'logical' attributes such as `server.address|port`
- individual DB/messaging extensions can still reference/require/recommend `network.*` when it makes sense for them

Related: https://github.com/open-telemetry/semantic-conventions/issues/674 (extracting db/messaging blocking part from that issue)

pyohannes:
For messaging, this would affect the following attributes:
| Attribute | Example | Requirement Level |
|--|--|--|
| [`network.peer.address`](../attributes-registry/network.md) | `10.1.2.80`; `/tmp/my.sock` | Recommended |
| [`network.peer.port`](../attributes-registry/network.md) | `65123` | Recommended: If `network.peer.address` is set. |
| [`network.protocol.name`](../attributes-registry/network.md) | `amqp`; `mqtt` | Conditionally Required: [15] |
| [`network.protocol.version`](../attributes-registry/network.md) | `3.1.1` | Recommended |
| [`network.transport`](../attributes-registry/network.md)  | `tcp`; `udp` | Recommended |
| [`network.type`](../attributes-registry/network.md)  | `ipv4`; `ipv6` | Recommended |

It makes sense to make all the recommended ones Opt-In. For `network.protocol.name`, we currently have it as conditionally required for messaging systems that support more than one protocol.

[...]

PR #1470, opened 2024-10-11
db: OTEL_SEMCONV_STABILITY_OPT_IN is important instrumentation libraries supporting v1.25.0 and prior

pellared:
Addresses https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6172#issuecomment-2397429335

Most likely it was a "copy-paste" error from messaging spans and metrics or someone has forgotten to bump the version.

Do you want a changelog entry for this?

CC @trask

### DIFF
--- a/docs/database/README.md
+++ b/docs/database/README.md
@@ -15,7 +15,7 @@ database metrics and logs.
 > **Warning**
 >
 > Existing database instrumentations that are using
-> [v1.24.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/database/database-spans.md)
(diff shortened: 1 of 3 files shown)


trask:
I'm not sure this is right, the pause was put into place in #733, which was prior to the v1.25.0 release.

and there were [breaking database changes in v1.25.0](https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md#v1250):

- db: Remove network.transport and network.type attributes from database semantic conventions, clarify when network.peer.address|port should be populated. (#690, #768)
- db: Deprecate db.connection_string attribute in favor of server.address and server.port (#724, #769)
- db: Deprecate db.jdbc.driver_classname attribute (#796)

pellared:
@trask, you are right. Closing. Sorry for inconvenience and thanks for your help.

PR #1479, opened 2024-10-15
Specify which span kind to use for messaging settle operations

### LINKED ISSUE #1478: Specify which span kind to use for `settle` operation type
### Area(s)

area:messaging

We have the tables showing the mapping between operation type and span kind. But the `settle` operation type is not mentioned, so it's  unclear which span kind to use when dealing with settle operations.

That the conventions specify which span kind to use for `settle` operation types.

N/A

### Additional context

Related to #1162

joaopgrassi:
Fixes #1478

## Changes

Defines which Span kind to use on spans modeling `settle` operations.


docs/messaging/messaging-spans.md
```diff
@@ -216,6 +216,7 @@ Span kind SHOULD be set according to the following table, based on the operation
 | `send`         | `PRODUCER` if the context of the "Send" span is used as creation context, otherwise `CLIENT`. |
 | `receive`      | `CLIENT`    |
 | `process`      | `CONSUMER`  |
+| `settle`       | `CLIENT` if the span models a synchronous call to the intermediary. Otherwise `INTERNAL` |
```
lmolkova, line 219: we use SHOULD right above this table

 https://github.com/open-telemetry/semantic-conventions/blob/d5d2b9d2a27f5983a5930699e0d40d5aba5c06fe/docs/messaging/messaging-spans.md?plain=1#L211

So I wonder if we really need to explicitly call out the `INTERNAL` case.

Also the "to the intermediary" is not always the case. E.g. Azure EventHubs settles using some external service (such as Azure Blob Storage, Redis or something else). 

So maybe just

```suggestion
| `settle`       | `CLIENT` |
```

?

joaopgrassi: Hum that makes sense. I guess just having client is really fine.

lmolkova (approved):

Issue #805, opened 2024-03-10
How to record multi-operation/table/dbs operations on DB metrics


lmolkova:
For  a simple database queries (such as `SELECT * from foo where bar="baz"`), we'd like to capture the following attributes when possible (on spans):

- `db.operation.name = SELECT`
- `db.collection.name = foo` (aka `db.sql.table` and other similar system-specific attributes)
- `db.collection.namespace = mydb` (aka `db.name` along with `db.instance.id` and similar)
- `db.query.text = SELECT * from foo where bar=?` (aka `db.statement`)
- `db.query.parameter.bar=baz`

(attribute names are being discussed and are not final).

This simple case is supported by current version of DB semantic conventions. It's also a common one in the NoSQL world if we exclude bulk operations (or non-homogeneous batch operations).
Attributes (except `db.query.*` ones) have reasonable cardinality and can be used on traces and metrics. 

-----

More complicated queries involve multiple operations, tables, or even databases. E.g. in `SELECT * from foo JOIN bar ON baz`
we have two operations (`SELECT` and `JOIN`), two tables (`foo` and ``bar`), and just one database.

- `db.query.text` and `db.query.parameter.*` for such queries are still relevant and make sense on spans (still cardinality is a problem for metrics)
- Operation and collection names become problematic to record

DB WG is considering multiple options:

------------
**Option 1: always capture `db.operation.names`, `db.collection.names`, `db.collection.namespaces` as  arrays**

Pros:  consistent understandable model

Cons: 
- array attribute on metrics
- simple case (one operation) becomes hard to use
- hard to query and not quite useful on spans (if we already capture templatized query text) 
- the combined cardinality of operation name, collection names, database names is the same as in the `db.query.text`

------------
**Option 2: capture both `db.operation.name` and  `db.operation.names`  (same for collections and namespaces).**

The array attributes are only captured when more than one operation is performed.
In this case we can entertain different options for `db.operation.name` - it may contain the first operation, operations joined as string, or shouldn't be reported at all.

Pros: 
- simple case is easy
- complex case is slightly better (e.g. when `db.operation.name` contains joined list `SELECT JOIN`)
 
Cons: 
- array attribute on metrics
- hard to query  and not quite useful on spans (if we already capture templatized query text) 
- the combined cardinality of operation name, collection names, database names is the same as in the `db.query.text`

------------
**Option 3: don't capture multiple operation names, collection names, namespaces**

Pros: simple case is easy
[...]

Issue #521, opened 2023-11-14
Database general-purpose target attribute


alanwest:
## Proposal for new `db.target` attribute

The type of targets available for a database operation is dependent on a database system. For example, the target of a SQL database operation may be a table, function, or stored procedure. Other database systems may refer to their targets as collections or indexes, etc.

Currently, there exists database system specific attributes for describing this. Examples include:

* [`db.sql.table`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/sql.md#call-level-attributes)
* [`db.redis.database_index`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/redis.md#call-level-attributes)
* [`db.mongodb.collection`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/mongodb.md#call-level-attributes)
* [`db.cassandra.table`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/cassandra.md#call-level-attributes)

Introducing a general-purpose attribute would enable backends to provide a generic experience for faceting database operations by their target without requiring the backend have knowledge of all the database system specific attributes. It would also enable end users to more easily explore their data at a high-level without having to know the specific database systems in use.

I propose a new [call-level](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#call-level-attributes) attribute `db.target`  for describing the target of a database operation. The term "target" may not be the greatest, but I'm using it for lack of a better term for the moment. I considered `db.operation.target` but of course `db.operation` is already an attribute. Maybe `db.operation` could become `db.operation.type` thereby allowing `db.operation.target`.

## Issues to consider

1. Similar to `db.operation`, the value of the `db.target` may not be readily available without parsing the value of `db.statement`, e.g. `SELECT * FROM Users` the `db.operation` is `SELECT` and the `db.target` is `Users`. The specification advises against parsing `db.statement` to derive other attributes. I consider this a limitation particularly once we have a database client operation metric (see #512). The `db.statement` attribute is fine for spans but because of its high-cardinality will not work for metrics. Therefore, having `db.operation` and `db.target` attributes on metrics will be useful.
2. There may be many targets for a single database operation, e.g. `SELECT * FROM Users U JOIN Company C ON ...`.

Perhaps parsing `db.statement` can be an opt-in feature?

Various backends have prior art for addressing these issues. For example:

[...]

PR #1482, opened 2024-10-15
DB: fix how operation and collection names are recorded for complex queries (opt-in `db.query.text` on metrics, new `db.query.summary` recommended attribute)

### LINKED ISSUE #1159: Database: add `db.query.text` as opt-in on metrics

lmolkova:
Fixes #1159, #805

Related: #521

Based on the discussion on slack https://cloud-native.slack.com/archives/C06HH05KNMC/p1728891505239829 and spec discussion on 10/25/24:
- having a single operation/collection name on metrics is not enough 
- it provides a baseline that works for nosql and sql with simple queries
- it can work better if/once we provide a way for users to supply their own operation alias/other context
- it's misleading to record one operation/collection when in fact there are multiple

We can however:
- allow users to enable capturing sanitized value of a `db.query.text` as opt-in to mitigate shortcomings of single operation/collection name.
- create a synthetic query text from a combination of operations and collections (like `SELECT foo, UPDATE bar, DELETE bar` - it'd have low-ish cardinality, would be shorter than a query, won't have any sensitive information that needs sanitization and would be readable enough. This allows to make it recommended on metrics.

This PR implements both options as discussed in https://github.com/open-telemetry/semantic-conventions/issues/805#issuecomment-2414781903


docs/database/database-spans.md
```diff
@@ -268,6 +287,84 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
 e.g. from `IN (?, ?, ?, ?)` to `IN (?)`, as this can help with extremely long IN-clauses,
 and can help control cardinality for users who choose to (optionally) add `db.query.text` to their metric attributes.
 
+## Synthetic query text
```
lmolkova, line 290: TODO: use it on spans

docs/database/database-spans.md
```diff
@@ -268,6 +287,84 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+
+When query text is parsed, the parser SHOULD extract a list of operations and corresponding
+collection names performed in this query.
+
+The `db.query.synthetic` attribute value SHOULD contain a semicolon separated list
+of operation and corresponding collection names formatted in the following way:
+
+```
+{operation1} {collection1}; {operation2} {collection2}; ...
+````
```
trask, line 305: wdyt of defining `db.query.synthetic` as a super sanitized version of `db.query.text` where the only things that are kept are operation names and collection names?

so that

> SELECT * from TABLE1, TABLE2 where ...

would become

> SELECT TABLE1 TABLE2

docs/database/database-spans.md
```diff
@@ -268,6 +287,84 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+   INSERT INTO shipping_details
+               (order_id,
+               address)
+   SELECT order_id,
+         address
+   FROM   orders
+   WHERE  order_id = ?
+   ```
+
+   the corresponding `db.query.synthetic` is `INSERT shipping_details; SELECT orders`.
```
trask, line 341: just seeing what it would look like elsewhere:

> INSERT shipping_details SELECT orders

docs/database/database-spans.md
```diff
@@ -268,6 +287,84 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+- Query that performs an operation that's applied to multiple collections:
+
+   ```sql
+   SELECT *
+   FROM   songs,
+         artists
+   WHERE  songs.artist_id == artists.id
+   ```
+
+  the corresponding `db.query.synthetic` is `SELECT songs,artists`.
```
trask, line 352: and

> SELECT songs artists

docs/database/database-spans.md
```diff
@@ -268,6 +287,84 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+- Query that performs an operation on an anonymous table:
+   ```sql
+   SELECT order_date
+   FROM   (SELECT *
+           FROM   orders o
+                  JOIN customers c
+                    ON o.customer_id = c.customer_id)
+   ```
+
+  the corresponding `db.query.synthetic` is `SELECT; SELECT orders; JOIN orders,customers`.
```
trask, line 363: and

> SELECT orders customers

docs/database/database-spans.md
```diff
@@ -268,6 +287,84 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+collection names performed in this query.
+
+The `db.query.synthetic` attribute value SHOULD contain a semicolon separated list
+of operation and corresponding collection names formatted in the following way:
+
+```
+{operation1} {collection1}; {operation2} {collection2}; ...
+````
+
+For example, a
```
lmolkova, line 307: oops

trask:
some possibilities for naming:

* `db.query.synthetic`
* `db.query.short`
* `db.query.summary`
* `db.query.description`

docs/database/database-spans.md
lmolkova: what about the first `SELECT` ? or should it be `SELECT SELECT orders customers` ? 

also it seems you dropped JOIN - I wanted to clarify that you don't consider it to be an operation, correct?

trask: oh, yes, I didn't mean to skip the extra `SELECT` (though potentially we could consider removing consecutive duplicates...)

and yeah, I wasn't considering `JOIN` to be an operation, though it could be...

lmolkova:
I really like the `db.query.summary` - it's accurate, it can be collected by other means (e.g. via a comment or a contextual attribute like in https://github.com/open-telemetry/oteps/pull/207). the `db.query.synthetic` would have limited opportunity to evolve into something more usable.

docs/database/database-spans.md
XSAM: Using empty space as a separator may cause problems, as databases may allow table names to have empty spaces.

For instance, in MySQL, this statement is allowed:

```sql
CREATE TABLE `USER TABLE` (
    UserID INT
);
```

For a query like this

```sql
SELECT * from `USER TABLE`, A;
```

would become

```sql
SELECT USER TABLE A
```

Which seems confusing, as we might think it queries three tables.

---

If we use a comma as a separator, we got 

```sql
SELECT USER TABLE, A
```

This seems clear.

lmolkova: It's a good point! Though I don't think anyone could make much sense out of this attribute alone or parse it in any meaningful way.

It's just a grouping key that is also sometimes readable for human beings.  I expect that the main use case for it is to get top N longest queries and then for each of them look into `db.query.text` to see what was this query about (via span attributes or metric exemplars). 

So I'm not sure if we need to worry about spaces in the table names on this attribute.

docs/database/database-spans.md
lmolkova: I'd rather not remove duplicates to preserve a bit more of the original query. I added some wording to explain that things like SQL statements is a good start and clauses can be added on case-by-case basis, but there are tradeoffs. PTAL

docs/database/database-spans.md
XSAM: > Though I don't think anyone could make much sense out of this attribute alone or parse it in any meaningful way.

It sounds like it is a hashed-alike string with a human-readable context. There may be a use case to see the duration among different `db.query.synthetic`, so we can quickly find the slowest `db.query.synthetic` (a pattern) from a table perspective instead of a single query perspective.

This would require us to determine the exact database tables from `db.query.synthetic`. With `SELECT USER TABLE A`. We might think there are three tables: `USER`, `TABLE, ' and `A`. Or, two tables: `USER`, `TABLE A`, etc.

I know using empty spaces should be fine in most cases, but I think we should keep our design more tolerant of rare cases. Just like what we did for `db.collection.name`.

XSAM: Another option: Users can add `"` to the database table for this special case. So, it can look like

```sql
SELECT "USER TABLE" "A"
```

wdyt? Would it sound too complicated?

trask: would it work if we only quote tables that have spaces?

docs/database/database-spans.md
```diff
@@ -268,6 +291,91 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
 e.g. from `IN (?, ?, ?, ?)` to `IN (?)`, as this can help with extremely long IN-clauses,
 and can help control cardinality for users who choose to (optionally) add `db.query.text` to their metric attributes.
 
+## Generating a summary of the query text
+
+The `db.query.summary` attribute captures a shortened representation of a query text
+which SHOULD have low-cardinality and SHOULD NOT contain any dynamic or sensitive data.
+
+> [!NOTE]
+> The `db.query.text` is intended to identify individual queries. Even though
```
trask, line 300: maybe
* The `db.query.text` attribute is ...
* The `db.query.text` value is ...
* `db.query.text` is ...

docs/database/database-spans.md
```diff
@@ -268,6 +291,91 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+> is it sanitized if captured by default, it still could have high cardinality and
+> might reach hundreds of lines of code.
+>
+> The `db.query.summary` is intended to provide less granular grouping key that
```
trask, line 304:
```suggestion
> The `db.query.summary` is intended to provide a less granular grouping key that
```

docs/database/database-spans.md
```diff
@@ -268,6 +291,91 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+> can be used as a span name or a metric attribute in a common case. It SHOULD
```
trask, line 305:
```suggestion
> can be used as a span name or a metric attribute in common cases. It SHOULD
```

docs/database/database-spans.md
```diff
@@ -268,6 +291,91 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+> only contain information that has a significant impact on the query, database,
+> or application performance.
+
+Instrumentations that support query parsing SHOULD generate query summary when
+one is not readily available from other sources.
+
+The summary SHOULD preserve the following parts of query in the order they were provided:
+- operations such as SQL SELECT, INSERT, UPDATE, DELETE or other statements
```
trask, line 313: is "statement" the right word here?

docs/database/database-spans.md
```diff
@@ -268,6 +291,91 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+   the corresponding `db.query.summary` is `SELECT wuser_table`.
+
+- Query that performs multiple operations:
+
+   ```sql
+   INSERT INTO shipping_details
+               (order_id,
+               address)
+   SELECT order_id,
+         address
```
trask, line 347:
```suggestion
          address
```

trask (commented):
I'm 👍 on this

docs/database/database-spans.md
lmolkova: what would we expect the `db.collection.name` be in the case of `` `USER TABLE` `` ? I'd expect it to contain all the characters that were provided in the query including quotes, double-quotes, backticks, non-ascii chars. etc.

That aligns with the language we have for collection name:
> to capture the value as provided by the application without attempting to do any case normalization

Then it should be represented in the same way in the `db.query.summary` - I can add some wording that targets there should follow the same as-is principle as on the attributes

docs/database/database-spans.md
lmolkova: yeah, not right. I changed to "command", but it's not a precise referenceable technical term either - I don't think there is a precise term defined in the language :( 

The problem is that we want a small subset of clauses - only the top-level ones. Open to any suggestions on how we can describe it better.
The only alternative I can think of is literally document all the commands that we'd support for SQL.

E.g. here's what ChatGPT  classifies as 'commands'

- SELECT
- INSERT
- UPDATE
- DELETE
- MERGE
- CREATE
- ALTER
- DROP
- TRUNCATE
- RENAME
- GRANT
- REVOKE
- COMMIT
- ROLLBACK
- SAVEPOINT
- SET TRANSACTION
- CREATE INDEX
- DROP INDEX
- ALTER INDEX
- START TRANSACTION
- SET

which seems to align with what we want it to be, so just saying "command" might be good enough.

trask: > "command" might be good enough

👍

model/database/registry.yaml
```diff
@@ -44,13 +50,18 @@ groups:
...
         note: >
-          It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.
+          It is RECOMMENDED to capture the value as provided by the application
+          without attempting to do any case normalization.
 
-          If the operation name is parsed from the query text, it SHOULD be the first operation name found in the query.
+          A single database query may involve multiple operations. If the operation
+          name is parsed from the query text, it SHOULD only be captured for queries that
+          contain a single operation or when the operation name describing the
+          whole query is available by other means (such as SQL comments).
```
trask, line 59: it may be better not to specifically mention SQL comments yet since we will probably want to define a standard format(?)

```suggestion
          contain a single operation or when the operation name describing the
          whole query is available by other means.
```

model/database/registry.yaml
```diff
@@ -90,6 +101,29 @@ groups:
...
+        stability: experimental  # RELEASE CANDIDATE
+        brief: >
+          Low cardinality representation of a database query text.
+        note: >
+          `db.query.summary` provides static summary of the query text. It describes
+          a class of database queries and is useful as a grouping key,
+          especially when analyzing telemetry for database calls involving complex queries.
+
+          Summary may be available to the instrumentation through SQL comment,
+          instrumentation hooks, or other means. If it is not available, instrumentations
```
trask, line 115: (similar thought here)
```suggestion
          Summary may be available to the instrumentation through
          instrumentation hooks or other means. If it is not available, instrumentations
```

model/database/common.yaml
```diff
@@ -7,8 +7,9 @@ groups:
       - ref: db.operation.name
         requirement_level:
           conditionally_required: >
-            If readily available. The operation name MAY be parsed from the query text,
-            in which case it SHOULD be the first operation name found in the query.
+            If readily available and if there is a single operation name that describes the
```
maryliag, line 10: what we do on the prepared statements case? 
For example, imagine a prepared statement that uses several operations, but since that has a name, that name could be used, which will make it easier to identify the operation. On the Postgres-JS prototype, if a statement has a name, it is using that as the operation name, otherwise tries to grab the first operation name.
My suggestion is to add a comment here, that if the query has a name, it can be used instead, no matter how many operations it has, otherwise follow the rules that you added here

model/database/registry.yaml
maryliag: I was also curious if any format or guidance on the sql comments would be added here

model/database/common.yaml
lmolkova: That was my goal with the first sentence

> If readily available and if there is a single operation name that describes the database call.

I believe when you execute a prepared statement or stored procedure, you'd write something like `CALL procedure-name` or `EXECUTE stmt_name` or `EXEC  procedure-name`

In this case the would still be one command and one target name and this description seems to work.

But it made me think that  `CALL procedure-name` should result in 

```
db.operation.name = CALL
db.collection.name = procedure-name
```

wonder if you and other folks think about it (we don't have it documented)

maryliag: That was indeed what I wanted to clarify, the result of _operation_ and _collection_ for the store procedure case.
Agree that the operation should be `CALL`, `EXEC` etc, and the collection should be the procedure name. So is worth making that clear on the collection name section

lmolkova: Agree! 

If you don't mind, I'd prefer to do it in a separate PR since it's a slightly different concern - created https://github.com/open-telemetry/semantic-conventions/issues/1491

maryliag: I don't mind at all 😄

docs/database/database-spans.md
```diff
@@ -268,6 +302,98 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
 
+## Generating a summary of the query text
+
+The `db.query.summary` attribute captures a shortened representation of a query text
+which SHOULD have low-cardinality and SHOULD NOT contain any dynamic or sensitive data.
+
+> [!NOTE]
+> The `db.query.text` attribute is intended to identify individual queries. Even though
+> it is sanitized if captured by default, it could still have high cardinality and
+> might reach hundreds of lines of code.
```
trask, line 313:
```suggestion
> might reach hundreds of lines.
```

docs/database/database-spans.md
```diff
@@ -268,6 +302,98 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+>
+> The `db.query.summary` is intended to provide a less granular grouping key that
+> can be used as a span name or a metric attribute in common cases. It SHOULD
+> only contain information that has a significant impact on the query, database,
+> or application performance.
+
+Instrumentations that support query parsing SHOULD generate query summary when
```
trask, line 320:
```suggestion
Instrumentations that support query parsing SHOULD generate a query summary when
```

docs/database/database-spans.md
```diff
@@ -268,6 +302,98 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+one is not readily available from other sources.
+
+The summary SHOULD preserve the following parts of query in the order they were provided:
+
+- operations such as SQL SELECT, INSERT, UPDATE, DELETE or other commands
+- operation targets such as collection or database names
```
trask, line 326:
```suggestion
- operations such as SQL SELECT, INSERT, UPDATE, DELETE, and other commands
- operation targets such as collections and database names
```

docs/database/database-spans.md
```diff
@@ -268,6 +302,98 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+
+Instrumentation MAY include additional details such as specific SQL clauses as long
+as summary remains relatively short and its cardinality remains low comparing to
+the `db.query.text`.
```
trask, line 330: any particular SQL clauses you have in mind?

docs/database/database-spans.md
```diff
@@ -268,6 +302,98 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+
+The instrumentation SHOULD parse the query and extract a list of operations and
+corresponding targets from the query. It SHOULD set `db.query.summary` attribute
```
trask, line 333: also, should we qualify "if instrumentation parses" here?

```suggestion
The instrumentation SHOULD parse the query and extract a list of operations and
targets from the query. It SHOULD set `db.query.summary` attribute
```

docs/database/database-spans.md
```diff
@@ -268,6 +302,98 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+to the value formatted in the following way:
+
+```
+{operation1} {target1} {operation2} {target2} {target3} ...
+````
+
+Instrumentations SHOULD capture the values of operations and targets as provided
+by the application without attempting to do any case normalization. If the operation
+and target value is populated on `db.operation.name`, `db.collection.name`, `db.namespace`
+or other attributes, it SHOULD match the value used in the `db.query.summary`.
```
trask, line 343: I think we've defined `db.namespace` in terms of the connection now, so not sure if it would ever be extracted from the query

trask (approved):

docs/database/database-spans.md
```diff
@@ -268,6 +302,98 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+
+   ```sql
+   SELECT *
+   FROM   songs,
+          artists
+   WHERE  songs.artist_id == artists.id
+   ```
+
+  the corresponding `db.query.summary` is `SELECT songs artists`.
+
```
XSAM, line 381:
```suggestion

- Query that performs an operation that's applied to multiple collections with double-quotes or other punctuation:

    ```sql
    SELECT *
    FROM   "song list",
           'artists'
    ```

    the corresponding `db.query.summary` is `SELECT "songs list" 'artists'`.

```

Could we include this in examples? So it could guide future readers to deal with this situation.

Related comments: https://github.com/open-telemetry/semantic-conventions/pull/1482#discussion_r1803546355

docs/database/database-spans.md
lmolkova: I think we still allow instrumentations to keep track if the current connection

> A connection's currently associated database may change during its lifetime, e.g. from executing `USE <database>`.

> If instrumentation is unable to capture the connection's currently associated database on each query without triggering an additional query to be executed (e.g. `SELECT DATABASE()`),
then it is RECOMMENDED to fallback and use the database provided when the connection was established.

>  Instrumentation SHOULD document if `db.namespace` reflects the database provided when the connection was established.

So it's technically possible if they match

docs/database/database-spans.md
```diff
@@ -268,6 +302,98 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+The summary SHOULD preserve the following parts of query in the order they were provided:
+
+- operations such as SQL SELECT, INSERT, UPDATE, DELETE, and other commands
+- operation targets such as collections and database names
+
+Instrumentation MAY include additional details such as specific SQL clauses as long
+as summary remains relatively short and its cardinality remains low comparing to
+the `db.query.text`.
+
+The instrumentation SHOULD parse the query and extract a list of operations and
```
trask, line 332:
```suggestion
Instrumentations that support query parsing SHOULD parse the query and extract a list of operations and
```

docs/database/database-spans.md
lmolkova: removed `db.namespace` from this list, it's not necessary there anyway

docs/database/database-spans.md
lmolkova: added!

docs/database/database-spans.md
lmolkova: I had `JOIN` in mind, but nothing else. I'm considering to remove this sentence - it doesn't seem to convey anything new.

docs/database/database-spans.md
trask: I think I see...

so someone is using database `MyDB1` and then executes `USE MyDB2`:

* `db.query.text=USE MyDB`
* `db.query.summary=USE MyDB`
* `db.operation.name=USE`
* `db.collection.name` is not captured
* `db.namespace=MyDB1` (I think, b/c that was the namespace associated with the connection at the time?)

and internally if the instrumentation is tracking namespace changes, then on the next query `SELECT * FROM table5`:

* `db.query.text=SELECT * FROM table5`
* `db.query.summary=SELECT table5`
* `db.operation.name=SELECT`
* `db.collection.name=table5`
* `db.namespace=MyDB2`

docs/database/database-spans.md
lmolkova: moved to the end of the section:

> Semantic conventions for individual database systems or specialized instrumentations

XSAM (approved):
👍

docs/database/database-spans.md
```diff
@@ -268,6 +302,105 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+
+> [!NOTE]
+> The `db.query.text` attribute is intended to identify individual queries. Even though
+> it is sanitized if captured by default, it could still have high cardinality and
+> might reach hundreds of lines.
+>
+> The `db.query.summary` is intended to provide a less granular grouping key that
+> can be used as a span name or a metric attribute in common cases. It SHOULD
+> only contain information that has a significant impact on the query, database,
+> or application performance.
```
trask, line 318: i'm not sure this part helps

I think I understand, that it groups things where queries in each bucket will have somewhat similar performance characteristics to each other, but worried that people would use this language to pull in more stuff into the summary than we want (increasing cardinality and reducing brevity), since there are many things that can have a significant impact (e.g. where clauses)

```suggestion
> can be used as a span name or a metric attribute in common cases.
```

docs/database/database-spans.md
```diff
@@ -268,6 +302,105 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+
+Instrumentations that support query parsing SHOULD generate a query summary when
+one is not readily available from other sources.
+
+The summary SHOULD preserve the following parts of query in the order they were provided:
```
trask, line 323:
```suggestion
The summary SHOULD preserve the following parts of a query in the order they were provided:
```

docs/database/database-spans.md
trask: did you mean to mark this resolved?

docs/database/database-spans.md
lmolkova: I think so.

Until recently we used to say that even the first span should point to the new database: https://github.com/open-telemetry/semantic-conventions/commit/25d438c3bf7234d27ba7a55e4520943e601f3cb9

> For commands that switch the keyspace, this SHOULD be set to the target keyspace (even if the command fails).

But the old version goes against of the current attribute definition.

docs/database/database-spans.md
```diff
@@ -268,6 +302,105 @@ Placeholders in a parameterized query SHOULD not be sanitized. E.g. `where id =
...
+    SELECT *
+    FROM   "song list",
+           'artists'
+    ```
+
+    the corresponding `db.query.summary` is `SELECT "songs list" 'artists'`.
+
+Semantic conventions for individual database systems or specialized instrumentations
+MAY specify a different `db.query.summary` format as long as produced summary remains
+relatively short and its cardinality remains low comparing to the `db.query.text`.
```
trask, line 402:
```suggestion
MAY specify a different `db.query.summary` format as long as the produced summary remains
relatively short and its cardinality remains low compared to `db.query.text`.
```

trask (approved):
🌟

joaopgrassi (approved):

PR #1492, opened 2024-10-18
Add network.interface.name attribute

ChrsMark:
## Changes

ref: https://github.com/open-telemetry/semantic-conventions/issues/308#issuecomment-2421604996

As discussed in the above referenced issue, this PR introduces the `network.interface.name` attribute and replace with it the `system.device` attribute of system and container network metrics.

Note: I'm not sure if this change can be reflected in `schema-next.yaml` 🤔 .


joaopgrassi (commented):
Please also add the old attribute to the deprecated file, like https://github.com/open-telemetry/semantic-conventions/blob/main/model/database/deprecated/registry-deprecated.yaml

ChrsMark:
Thank's @joaopgrassi . My concern here is that `system.device` is not entirely removed, it's still used as an attribute in other metrics. Here we just replace its usage in specific metrics. I guess this can be reflected in the `schema-next` as you mentioned, but should this also be added in the deprecated file(s)?

joaopgrassi:
> but should this also be added in the deprecated file(s)?

I believe yes, each "area" now has its own deprecated file, so within that usage, the new attribute is considered deprecated. There then it's also mentioned which one to use instead. 

@lmolkova @jsuereth is this correct in your view? Not sure if we had such case before 🤔

model/container/deprecated/registry-deprecated.yaml
```diff
@@ -29,3 +29,17 @@ groups:
...
+    brief: >
+      "Describes deprecated container metrics attributes."
+    stability: experimental
+    attributes:
+      - id: system.device
+        type: string
+        stability: experimental
+        brief: "Deprecated, use `network.interface.name` instead."
+        deprecated: "Replaced by `network.interface.name`."
+        examples: [ "(identifier)" ]
```
lmolkova, line 45:
```suggestion
```

since you're declaring it deprecated in system/deprecated/registry-deprecated.yaml, no need to add anything here.

the file structure is only important for humans, tooling does not care about file names.

model/system/deprecated/registry-deprecated.yaml
```diff
@@ -51,3 +51,17 @@ groups:
...
         deprecated: 'Replaced by `cpu.mode`'
         examples: [ "idle", "interrupt" ]
+
+  - id: registry.system.metrics.deprecated
+    type: attribute_group
+    display_name: Deprecated System Metrics
+    brief: >
+      "Describes deprecated system metrics attributes."
+    stability: experimental
+    attributes:
```
lmolkova, line 61:
```suggestion
```

just add it  to the existing group - there is in special meaning in the group name and properties here -  it does not matter which `attribute_group` attribute is declared in for tooling

schema-next.yaml
```diff
@@ -2,6 +2,18 @@ file_format: 1.1.0
 schema_url: https://opentelemetry.io/schemas/next
 versions:
   next:
+    metrics:
```
lmolkova, line 5:
```suggestion
    all:
```

it does not matter where the attribute was used: spans or metrics or resources, it should be renamed everywhere (and yes, this file schema is not perfect)

lmolkova (approved):
LGTM (with some minor suggestions to make CI happy)

model/container/deprecated/registry-deprecated.yaml
lmolkova: Wait, from https://github.com/open-telemetry/semantic-conventions/pull/1492#issuecomment-2425987628 I jus realized that the attribute is still used. Let's not deprecate it at all then, just use a new one on the metrics it's applicable to.

lmolkova:
> Appears that we can't model this right now?

There is no need to deprecate attributes if it's still used.

The only thing you need to do is to write the proper section in the schema-next. I think what we need is https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/schemas/file_format_v1.1.0.md#rename_attributes-transformation-2

```
   metrics:
      changes:
        - rename_attributes:
            attribute_map:
              # map of key/values. The keys are the old attribute name used
              # in the previous version, the values are the new attribute name
              # starting from this version.

            apply_to_metrics:
              # Optional. If it is missing the transformation is applied
              # to all metrics. If it is present the transformation is applied
              # only to the metrics with the name that is found in the sequence
              # specified below.
```

model/system/deprecated/registry-deprecated.yaml
```diff
@@ -51,3 +51,17 @@ groups:
...
+      - id: system.device
+        type: string
+        stability: experimental
+        brief: "Deprecated, use `network.interface.name` instead."
+        deprecated: "Replaced by `network.interface.name`."
+        examples: [ "(identifier)" ]
```
lmolkova, line 67:
```suggestion
```

schema-next.yaml
```diff
@@ -2,6 +2,18 @@ file_format: 1.1.0
...
+      changes:
+        # https://github.com/open-telemetry/semantic-conventions/pull/1492
+        - rename_attributes:
+            attribute_map:
+              system.device: network.interface.name
+            apply_to_metrics:
```
lmolkova, line 11: 👍

model/system/deprecated/registry-deprecated.yaml
ChrsMark: Fixed.

ChrsMark:
@open-telemetry/semconv-system-approvers could you review this one please ?

rogercoll (approved):

mx-psi (approved):

AlexanderWert (approved):

PR #1496, opened 2024-10-21
[chore] Improve markdown gen for registry pages

joaopgrassi:
## Changes

I noticed that the registry attribute tables have these spaces in the table headings, which are added by `npm run fix:format`. I believe this was introduced because Jinja spaces that are left over and the format fixes them while also adding spaces for the table headings/cells.

I improved the Jinja templates a bit so now it's not necessary to run the npm format task anymore. This aligns with the generation for non-registry attribute tables as well.


jsuereth (approved):
Absolute hero...

ChrsMark (approved):

Issue #1485, opened 2024-10-17
[k8s] Define semantic conventions for k8s volume metrics


ChrsMark:
### Area(s)

area:k8s

### Is your change request related to a problem? Please describe.

As part of the K8s SemConv stability work we need to define k8s volume metrics that are already in use by the [Opentelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.111.0/receiver/kubeletstatsreceiver/documentation.md#k8svolumeavailable).

Related to https://github.com/open-telemetry/semantic-conventions/issues/1032.

### Describe the solution you'd like

The existing metrics in use that we need to define as semantic conventions are the following:

```
k8s.volume.available
k8s.volume.capacity
k8s.volume.inodes
k8s.volume.inodes.free
k8s.volume.inodes.used 
```

PR #1499, opened 2024-10-21
k8s: k8s.persistent_volume: new resource, assign CSI attrs

gdvalle:
This follows up from https://github.com/open-telemetry/semantic-conventions/pull/1337, and adds the CSI attributes as resource attributes for kubernetes volumes for use in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055.

xref https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055/files#r1805969619


model/container/resources.yaml
```diff
@@ -20,3 +20,5 @@ groups:
         requirement_level: opt_in
       - ref: container.label
       - ref: oci.manifest.digest
+      - ref: container.csi.plugin.name
+      - ref: container.csi.volume.id
```
lmolkova, line 24: quick question: could these change during container lifetime?

gdvalle: I suppose that would depend on the container runtime. I could see something like hot-plugging for container volumes, but I'm not aware of anything that does that.

Is that a heuristic you're using to decide if this is a resource attribute?

lmolkova: yes, we currently have a problem with mutable resource attributes - there is no way to update them once they are set when otel is configured, see https://github.com/open-telemetry/semantic-conventions/issues/1181#issuecomment-2229168814 for the details.

Having said this, it seems in (at least) most of the cases, it's not possible, so we can add them as resource attributes with a following clause 

```suggestion
      - ref: container.csi.plugin.name
        requirement_level:
          recommended: if and only if container runtime does not support changing CSI plugin during container lifetime. The attribute MUST NOT be set if its value may change during OpenTelemetry SDK instance lifetime.
      - ref: container.csi.volume.id
        requirement_level:
          recommended: if and only if container runtime does not support changing CSI plugin during container lifetime. The attribute MUST NOT be set if its value may change during OpenTelemetry SDK instance lifetime.      
```

unless @jsuereth has any problem with it

ChrsMark: I'm not sure if this attribute has to do with the container, I think this is an attribute of the Persistent Volume based on the implementation at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055/files#diff-ff719bd25c1ddf9644468b5e873d43ad7a8e95d8ff5a59b685cf4b6ecb6c1f77R54-R115?

@open-telemetry/semconv-k8s-approvers any thoughts on this?

gdvalle: > I'm not sure if this attribute has to do with the container

I came to say the same thing. It's not a container resource attribute. One clear tell is it's not 1:1 to a container, rather 1:N.

It seems there's no `k8s.volume` section, so I think that'll have to go in first. I assume this can map to [core Volume v1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#volume-v1-core)?

gdvalle: I've updated the PR to reflect that change in https://github.com/open-telemetry/semantic-conventions/pull/1499/commits/b85d7be0d7d56c129037cdada66df7eccf1ca6d4.

ChrsMark: > I assume this can map to [core Volume v1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#volume-v1-core)?

I'm not sure about how we should model this. It appears that `csi` can be an attribute of more than 1 K8s resource types. So the question is which one we should start with for that specific use case.

I see that `csi` also appears in the [PersistentVolumeSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#persistentvolumespec-v1-core) which seems to reflect what we do in the [implementation](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055/files#diff-ff719bd25c1ddf9644468b5e873d43ad7a8e95d8ff5a59b685cf4b6ecb6c1f77R54-R115?)?

gdvalle: A k8s [Volume](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#volume-v1-core) can have a [CSIVolumeSource](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#csivolumesource-v1-core). This has fields for `driver`, but not `handle` (i.e. `container.csi.volume.id`). So for Volume, it seems only `plugin.name` would be valid as an optional attribute.

A [PersistentVolume](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#persistentvolume-v1-core) can have [CSIPersistentVolumeSource](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#csipersistentvolumesource-v1-core), which has `driver` (`container.csi.plugin.name`) and `volumeHandle` (`container.csi.volume.id`), so I think this would have both attrs as optional.

I think I was proposing that we just collapse these into one k8s.volume representation, marking them both optional, but that does mean the non-persistent-volumes contain a schema with an attribute that isn't possible. I feel this is a slightly more desirable tradeoff as opposed to having another bag of attrs under say `k8s.persistent_volume`, with the thinking that the end-user doesn't care about the distinction here.

gdvalle:
not stale

model/container/resources.yaml
ChrsMark: > I think I was proposing that we just collapse these into one k8s.volume representation, marking them both optional, but that does mean the non-persistent-volumes contain a schema with an attribute that isn't possible.

I find this a bit weird tbh specially when it will be part of the schema. I think the schema definition should be precise. Could we only define exactly what we are going to use in the [implementation](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055/files#diff-ff719bd25c1ddf9644468b5e873d43ad7a8e95d8ff5a59b685cf4b6ecb6c1f77R101) for now?

gdvalle: > I think the schema definition should be precise. Could we only define exactly what we are going to use in the [implementation](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055/files#diff-ff719bd25c1ddf9644468b5e873d43ad7a8e95d8ff5a59b685cf4b6ecb6c1f77R101) for now?

I think I crossed some wires with:
> I think I was proposing that we just collapse these into one k8s.volume representation

We don't really need that as this categorization isn't exactly showing up anywhere the user interacts, so perhaps just calling this `k8s.persistent_volume` is what we're after? I did just that now.

model/k8s/resources.yaml
```diff
@@ -111,3 +111,13 @@ groups:
...
       - ref: k8s.cronjob.uid
       - ref: k8s.cronjob.name
+
+  - id: resource.k8s.persistent_volume
+    type: resource
+    stability: experimental
+    name: k8s.persistent_volume
+    brief: >
+      A Kubernetes Persistent Volume object.
+    attributes:
```
ChrsMark, line 121: I guess this would at least need to define the `k8s.persistent_volume.name` and `k8s.persistent_volume.uid` attributes.

ChrsMark (commented):
I think the reason we are stuck here is that there is no clear way right now to link metrics to resources in the spec: https://github.com/open-telemetry/semantic-conventions/pull/1649#discussion_r1872258043

The volume metrics that we'd want to define are listed in https://github.com/open-telemetry/semantic-conventions/issues/1485. It would help if we start with defining them first in SemConv since the Collector's [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32055) actually tries to add (resource) attributes to them when they are not yet defined in SemConv. This complicates things.

In addition we have already defined `k8s.volume.name` and `k8s.volume.type` in the [registry](https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/registry.yaml#L180-L186) but not yet as part of a resource.

Following the [Collector's code flow](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/dc74f03898b8fb3185dfa6c330a88957bea717ff/receiver/kubeletstatsreceiver/scraper.go#L139-L158) I still think that the `csi.*` attributes look to be related to the Persistent Volume resource primarily. We first get the PVC, then we get the PV that it references and then out of this PV we get the `csi.*` info.

In general this touches the following open questions/problems:
1. Figure out how volume, persistent_volume etc are related as resources.
1. Figure out in which resource(s) `container.csi.*` attributes should become members.
2. Figure out how volume metrics are related to the resources mentioned above.

ps: I would be fine if we proceed with the persistent volume definition in case @open-telemetry/specs-semconv-maintainers and @open-telemetry/semconv-k8s-approvers agree, I'm just raising my concerns here since we are kind of moving in uncharted waters here.

PR #1507, opened 2024-10-22
Add system uptime metric

### LINKED ISSUE #648: Add system uptime metric

kevinnoel-be:
Fixes #648

## Changes

Add `system.uptime` metric


docs/system/system-metrics.md
```diff
@@ -1159,3 +1161,24 @@ See also the [Slab allocator](https://blogs.oracle.com/linux/post/understanding-
...
+<!-- semconv metric.system.uptime -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `system.uptime` | Counter | `s` | The time the system has been running | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 1180: can we add some language on the precision? I.e. the unit is seconds, but the counter should be double and instrumentation should measure it with the highest precision possible. 
Accuracy, however, would depend on the instrumentation/OS/etc.

trask:
hi @kevinnoel-be! it may be worth looking over https://github.com/open-telemetry/oteps/pull/185

docs/system/system-metrics.md
kevinnoel-be: Added a note around the precision and what should be expected

model/system/metrics.yaml
```diff
@@ -394,3 +394,16 @@ groups:
     unit: "By"
     attributes:
       - ref: linux.memory.slab.state
+
+  # system.uptime metric
+  - id: metric.system.uptime
+    type: metric
+    metric_name: system.uptime
+    stability: experimental
+    brief: "The time the system has been running"
```
braydonk, line 403: @open-telemetry/specs-semconv-maintainers My instinct is to say that the brief should mention that this is in seconds in particular; different OS APIs may provide a value in seconds or milliseconds so it feels like a good idea to be specific in the description on top of `s`. But I'm not sure if there's a stance against being redundant about units in descriptions.

braydonk (commented):
LGTM, this was pretty much what I was going to do. I do have an open question to semconv maintainers, depending on their answer may ask for a change or approve.

model/system/metrics.yaml
```diff
@@ -394,3 +394,13 @@ groups:
     unit: "By"
     attributes:
       - ref: linux.memory.slab.state
+
+  # system.uptime metric
```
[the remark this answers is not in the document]
braydonk: disregard, making a different suggestion

model/system/metrics.yaml
```diff
@@ -394,3 +394,16 @@ groups:
...
+  - id: metric.system.uptime
```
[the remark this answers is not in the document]
kevinnoel-be: Makes sense, moved it on top, same for the docs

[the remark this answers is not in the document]
braydonk: Looks good thanks!

model/system/metrics.yaml
```diff
@@ -1,4 +1,17 @@
...
+    type: metric
+    metric_name: system.uptime
+    stability: experimental
+    brief: "The time the system has been running"
+    note: |
+      Instrumentations SHOULD use counter with type `double` and measure uptime with the highest precision available.
+      The actual accuracy would depend on the instrumentation and operating system.
+    instrument: counter
+    unit: "s"
+    attributes: []
```
lmolkova, line 13: super-nit:

```suggestion
```

I know the schema requires it which results in the VSCode showing violation, but it's a bug and they are not required by anything.

lmolkova (approved):

model/system/metrics.yaml
lmolkova: No stance, it's ok to repeat something if it's important.

If we happen to mention unit in the brief it'd be nice to also mention that it's floating point.

model/system/metrics.yaml
```diff
@@ -1,4 +1,17 @@
...
+  # system.* metrics
+  - id: metric.system.uptime
+    type: metric
+    metric_name: system.uptime
+    stability: experimental
+    brief: "The time the system has been running"
+    note: |
+      Instrumentations SHOULD use counter with type `double` and measure uptime with the highest precision available.
+      The actual accuracy would depend on the instrumentation and operating system.
+    instrument: counter
```
lmolkova, line 11: it seems @jsuereth has good argument to make it a gauge in 
https://github.com/open-telemetry/oteps/blob/05ff3aed8d6d9b9625e4756b4c279048d46e3b28/text/metrics/0185-uptime-monitoring-semantic-conventions.md#uptime

jsuereth:
Just want to call out that we have some previous thoughts and rationale here: https://github.com/open-telemetry/oteps/pull/185

model/system/metrics.yaml
braydonk: Yeah based on that OTEP, I think this should be changed to `gauge`. I'm going to take that OTEP and work with @jsuereth and the other folks who originally authored the guidance to turn this into official System Semconv guidance. So while it's not yet, we'll soon attempt to make it a proper rule that uptime metrics should be a gauge, and will add references to that guidance in the note for this and any future uptime metrics once it's ready.

model/system/metrics.yaml
```diff
@@ -1,4 +1,17 @@
 groups:
+  # system.* metrics
+  - id: metric.system.uptime
+    type: metric
+    metric_name: system.uptime
+    stability: experimental
+    brief: "The time the system has been running"
```
braydonk, line 7: Based on the response to my earlier question to semconv maintainers, I think this brief should be adjusted that the time is a floating point number of seconds.

braydonk (commented):
Based on https://github.com/open-telemetry/oteps/pull/185 and the response from the semconv maintainers, I'm requesting the following changes.

docs/system/system-metrics.md
```diff
@@ -73,6 +75,31 @@ Resource attributes related to a host, SHOULD be reported under the `host.*` nam
...
+<!-- semconv metric.system.uptime -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `system.uptime` | Counter | `s` | The time the system has been running [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
jsuereth, line 93: I believe this is stated elsewhere, but this should be a gauge.

Copying rationale from the uptime monitoring OTEP:

> uptime is reported as a gauge with the value of the number of seconds that the process has been up. This is written as a gauge because users want to know the actual value of the number of seconds since the last restart to satisfy the use cases above. Sums are not a good fit for these use cases because most metric backends tend to default cumulative monotonic sums to rate-calculations, and have overflow handling that is undesired for this use case.
>
> Sums report a total value that has accumulated over a time window; it is valid, for instance, to subtract the current value of a cumulative sum and restart the start timestamp to now. (OpenTelemetry's Prometheus receiver does this, for instance.)
An intended use case of a sum is to produce a meaningful value when aggregating away labels using sum. Such aggregations are not meaningful in the above use cases.

jsuereth (commented):
As braydonk@ stated, I think the `system.uptime` should be a gauge, which DOES deviate from other time-related values for specific reasons, which is why we wrote the OTEP oh so long ago.

Otherwise looks good.

model/system/metrics.yaml
kevinnoel-be: updated

model/system/metrics.yaml
kevinnoel-be: updated

model/system/metrics.yaml
kevinnoel-be: removed

docs/system/system-metrics.md
kevinnoel-be: updated

kevinnoel-be:
Adressed all the comments, PTAL again, thanks!

jsuereth (approved):

trask (approved):

lmolkova (approved):

PR #1512, opened 2024-10-23
Makes all span/metric/event/resource groups id meaningful 

### LINKED ISSUE #1324: Yaml schema: unify semconv identity property
- attributes have `id`
- events have `name`
- metrics have `metric_name`
- resource have `name`

resources and spans don't have identity except the group-id.

We should consider unifying a property that describes a full name of a thing, especially if it's meaningful outside of semconv repo, e.g. for code generation and instrumentations.

lmolkova:
Fixes #1324

Part of https://github.com/open-telemetry/semantic-conventions/pull/1472

In order to provide stability guarantees, we need to make semconv groups identifiable.

We currently have different identifying properties - `name` on event, `metric_name` on metric, `name`  on resource, but 
- nothing for spans
- these properties are called differently and it's hard to use them in generic way leading to issues like https://github.com/open-telemetry/weaver/issues/289
- these properties are unique within signal, but could be the same across signals (e.g. `dns.lookup.duration` could be a metric and an attribute)

The group `id` is the common identity we can have, but it does not have any meaning now. This PR enforces the following pattern for `id`
- `metric.{metric_name}` for metrics
- `event.{name}` for events
- `resource.{name}` for resources
- `span.\.+.{span_kind}` for spans - spans don't have identifying property, so there is nothing to check at the moment. We should be able to change this format if we introduce something better.


docs/resource/cloud-provider/heroku.md
```diff
@@ -2,11 +2,7 @@
 
 **Status**: [Experimental][DocumentStatus]
 
-**type:** `heroku`
-
-**Description:** [Heroku dyno metadata]
```
lmolkova, line 7: it's autogenerated below, so removing duplication

model/aws/sdk-spans.yml
```diff
@@ -28,53 +29,22 @@ groups:
       - ref: aws.request_id
         requirement_level: recommended
 
-  - id: dynamodb.shared
```
lmolkova, line 31: this group is not used

model/cloudevents/spans.yaml
```diff
@@ -1,6 +1,6 @@
 groups:
   - id: cloudevents
-    type: span
+    type: attribute_group
```
lmolkova, line 3: @joaopgrassi it seems we're saying that cloudevents are just attributes on messaging semconv, do you think we can change the group type?

model/feature-flag/logs.yaml
```diff
@@ -1,11 +1,14 @@
 groups:
-  - id: log-feature_flag
```
lmolkova, line 2: that's effectively the same change as made in the https://github.com/open-telemetry/semantic-conventions/pull/1440/files#diff-1db9b9bcfb009755e349125b55d3c945e2b5636723e91ee73ac4cd9184108036

model/feature-flag/events.yaml
```diff
@@ -1,14 +0,0 @@
-groups:
-  - id: feature_flag
```
lmolkova, line 2: one of identical feature_flag events has to go so that we can have one group id that follows the pattern.

Removing it also partially unblocks  https://github.com/open-telemetry/weaver/issues/306

model/cloudevents/spans.yaml
joaopgrassi: Yes, I think that is fine and should be like that anyway!

policies/yaml_schema.rego
```diff
@@ -80,6 +104,22 @@ deny[yaml_schema_violation(description, group.id, name)] {
     description := sprintf("Resource name '%s' is invalid. Resource name %s'", [name, invalid_name_helper])
 }
 
+# checks that resource group id matches resource.{name}
+deny[yaml_schema_violation(description, group.id, name)] {
+    group := input.groups[_]
+    group.type == "resource"
+    name := group.name
+
+    # TODO: remove once TODO is fixed
```
joaopgrassi, line 113: Which TODO is that?

policies/yaml_schema.rego
```diff
@@ -48,6 +60,18 @@ deny[yaml_schema_violation(description, group.id, name)] {
     description := sprintf("Event name '%s' is invalid. Event name %s'", [name, invalid_name_helper])
 }
```
joaopgrassi, line 62: Should (or do we?) have a test that ensures now that resource `id` is in the form of `resource.{name}`?

docs/feature-flags/feature-flags-logs.md
```diff
@@ -48,11 +48,11 @@ The table below indicates which attributes should be added to the
 
 The event name MUST be `feature_flag`.
 
-This event describes feature flag evaluation representation on a Log Record.
+This event describes feature flag evaluation.
 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
-| [`feature_flag.key`](/docs/attributes-registry/feature-flag.md) | string | The unique identifier of the feature flag. | `logo-color` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`feature_flag.key`](/docs/attributes-registry/feature-flag.md) | string | The unique identifier of the feature flag. | `logo-color` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
joaopgrassi, line 55: Is the change from recommended to required here intentional? 🤔

policies/yaml_schema.rego
lmolkova: great catch! I had tests for spans, metrics and events, but not for resources

docs/feature-flags/feature-flags-logs.md
lmolkova: It actually is.

We currently have two almost identical feature flag definitions (one for events, one for logs) and they are only different in the requirement level on this attribute. I stole the fix from https://github.com/open-telemetry/semantic-conventions/pull/1440 


@dyladan and @open-telemetry/semconv-feature-flag-approvers to confirm

policies/attribute_name_collisions_test.rego
```diff
@@ -1,19 +0,0 @@
-package after_resolution
```
joaopgrassi, line 1: Are these not needed anymore?

dyladan (approved):
Approving from feature_flag perspective. What is the reason you removed `event.yaml` and kept `log.yaml` instead of the other way around as it appears in most other semconvs? Is that just to mimic what I did in my other PR?

docs/feature-flags/feature-flags-logs.md
dyladan: I wasn't sure when to have an events.yaml and when to have a logs.yaml. Is there anywhere documentation which describes when to use which one?

lmolkova: I *think* it should be event.yaml, but really the file name is not enforced or regulated.

The convention itself should be for events because of

> * A semantic convention needs to be defined. We do not define semantic
  conventions for LogRecords that are not Events.

https://github.com/open-telemetry/opentelemetry-specification/blob/8548dfc199a76504d573880c4a3fe5a0e483357b/specification/logs/event-api.md?plain=1#L75

Let me create an issue so we document our stance on span events vs events vs logs in semantic conventions

lmolkova:
> Approving from feature_flag perspective. What is the reason you removed event.yaml and kept log.yaml instead of the other way around as it appears in most other semconvs? Is that just to mimic what I did in my other PR?

Thanks  @dyladan ! Yes, just to mimic what you've done in the other PR.

MSNev:
General Observation, while I (believe) that I understand the desire to make this change it seems a little redundant to basically have the group `id` prefixed with the `type`.

Would it not be better to just enforce this via weaver so that the `id` (when loaded) is always prefixed with its explicit `type` (so that all `ids` become unique based on the type. So all `ids` and then unique across all `types` and it would avoid someone trying to bypass the logic rules of "naming" a span some other type....

And then we should enforce that any references / extend to the existing `id` **must** use the fully constructed `type.id`, so we still enforce the isolation of the types?

lmolkova:
> OK. For consistency with other areas of semconv should we use events.yaml instead? Looks like most are using that and the id you used is using the `event` prefix.

Agreed, I'll rename to event.yaml.

lmolkova:
@MSNev I'd prefer to keep things explicit and self-contained in semconv. It'd be surprising to write `http.client` when defining something and then reference it as a `span.http.client`.

docs/feature-flags/feature-flags-logs.md
lmolkova: Created https://github.com/open-telemetry/semantic-conventions/issues/1526 and renamed file to `events.yaml`

trask (approved):

PR #1520, opened 2024-10-28
process.executable.build_id: rename attribute

florianl:
## Changes

With https://github.com/open-telemetry/opentelemetry-specification/pull/4197 it was decided to rename the attribute from `process.executable.build_id.profiling` to `process.executable.build_id.htlhash` (Head-Tail-Length Hash).

FYI @open-telemetry/profiling-maintainers


felixge (approved):
LGTM.

Is this level of change tracking needed while we're still experimental?

docs/attributes-registry/process.md
```diff
@@ -75,6 +75,7 @@ Deprecated process attributes.
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
 | <a id="process-cpu-state" href="#process-cpu-state">`process.cpu.state`</a> | string | Deprecated, use `cpu.mode` instead. | `system`; `user`; `wait` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `cpu.mode` |
+| <a id="process-executable-build-id-profiling" href="#process-executable-build-id-profiling">`process.executable.build_id.profiling`</a> | string | "Deprecated, use `process.executable.build_id.htlhash` instead." |  | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `process.executable.build_id.htlhash` |
```
christos68k, line 78: Since AFAIK we haven't released a protocol version that includes `build_id.profiling`, maybe we shouldn't list it as deprecated to avoid carrying unnecessary baggage forward.

christos68k (approved):

docs/attributes-registry/process.md
florianl: Other attributes, that got renamed and are marked as `experimental`, also requires this approach - e.g. https://github.com/open-telemetry/semantic-conventions/pull/1422 

maybe @open-telemetry/specs-semconv-maintainers can provide more feedback and guidance

lmolkova: we never remove attributes that were released. The reason is that once version is released, language SIGs generate artifacts based on it like  https://github.com/open-telemetry/semantic-conventions-java and at this point it's a public API that users might take a dependency on.

We could still break things (since artifacts with experimental attributes should be experimental too), but it'd create a lot of pain due to dependency version conflicts.

TL;DR: we could at some point drop a bunch of experimental deprecated attributes, but we don't allow removing them now.

lmolkova (approved):

trask (approved):

PR #1521, opened 2024-10-28
Add the concept of reference attributes to the spec

### LINKED ISSUE #1428: Seeking input on proposal for "reference"-type attribute values.
### Area(s)

area:new

Some attributes may may be very large such as:

 - [ECS span attributes for HTTP](https://www.elastic.co/guide/en/ecs/current/ecs-http.html):
   - `http.request.body.content`
   - `http.response.body.content`
 - [GenAI span event attributes](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#events):
   - `gen_ai.prompt`
   - `gen_ai.completion`
 - [Logs exceptions](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/#attributes):
   -  `exception.stacktrace`
- [Span event exceptions:](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/):
   -  `exception.stacktrace`

In order to provide a very tight SLO and to provide consistent performance, an operations backend might provide very tight limits on the size of payloads/content/attributes. However, it may still be desirable when this is the case to be able to index most of the properties/attributes as well as to cross-link to an alternative storage solution (with a looser SLO) for full, large data.

In the OTel Collector Contrib repository, there is an open issue for a new connector component ("[New component: Blob Attribute Uploader Connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33737)") which attempts to resolve the above issue by proposing a connector that will, in the instrumentation/collector, do the following:

1. Upload the full attribute data to a blob storage system (such as Google Cloud Storage, Amazon S3, Azure Blob storage, or -- in the future -- any other blob storage backend supported by the CDK or that chooses to contribute to the connector).

2. Remove the original attribute that would otherwise be too large.

3. Inject attribute(s) that contain references to where the data was uploaded.

It is with respect to this last one, that we need an established standard / data model for how to represent the uploaded data. The "[SetInMap](https://github.com/michaelsafyan/open-telemetry.opentelemetry-collector-contrib/blob/blob_writer_span_processor/connector/blobattributeuploadconnector/internal/foreignattr/factory.go)" function in the "[foreign attribute](https://github.com/michaelsafyan/open-telemetry.opentelemetry-collector-contrib/tree/blob_writer_span_processor/connector/blobattributeuploadconnector/internal/foreignattr)" internal library of the [draft connector](https://github.com/michaelsafyan/open-telemetry.opentelemetry-collector-contrib/tree/blob_writer_span_processor/connector/blobattributeuploadconnector) shows the solution currently being implemented/pursued. However, I'd like to get more community feedback/input in order to ensure that there is agreement on the approach.

## Summary:

| **Before uploading** | **After uploading** |
| --- | --- |
| `somekey` | `somekey.ref.uri` |
|                    | `somekey.ref.content_type` _\[OPTIONAL\]_ |

## Formally:

A backend system that sees an attribute matching the pattern "${prefix}.ref.uri" should assume that "${prefix}.ref.uri" contains a URI that reports that location of the original, full value of "${prefix}". If a key named "${prefix}" also is present, it is likely that "${prefix}" contains a truncated and/or redacted copy of the original value, with "${prefix}.ref.uri" pointing to the location where the original, unadulterated, full version of the value has been recorded. If "${prefix}.ref.uri" exists, then "${prefix}.ref.content_type" may also optionally exist, containing a MIME type describing the data stored at the URI in "${prefix}.ref.uri".

## Prerequisites / Interactions:

If this course of action were to be taken, we would need two changes to [the "Attribute Naming" specification in OTel:](https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/)

1. Reserve ".ref." so that it can only be used for this purpose ("*.ref.*" to be used only for information about reference-typed values and their metadata).

2. Exempt this usage from the rule that "Names SHOULD NOT coincide with namespaces"

With respect to the latter exemption, our understanding is that this rule exists to allow coaelescing the flat attributes into a structured object. We believe that ".ref." fits the spirit of this even while violating the literal rule as it exists now, because the ".ref." does not introduce a conceptually new attribute; rather, it replaces the existing attribute with a different representation. Thus if a backend were to implement a coalescing system to make attributes non-flat, they could combine all of the ".ref." attributes into a single "ReferenceValue" type as the value for the top-level attribute that does not contain any ".ref." value in it.

## Use a compound data type

For example, we could use a "KeyValueList kvlist_value" to represent the reference not as a bunch of separate attributes but as a single complex attribute value. However, this would go contrary to the [attribute specification](https://opentelemetry.io/docs/specs/otel/common/) as well as existing OTel libraries which do not provide direct access to the underlying data model in OTLP following this convention restriction.

## Introduce a new data type

We could attempt to introduce into OTLP (and into other parts of the specification) a new "ReferenceValue" data type. However, this would require significant effort across multiple languages, libraries, backend systems, etc. to support and thus seems like a non-starter.

## Just move the large attributes to be events/logs

This assumes that the event/logs backend can, itself, accept arbitrarily large data. This assumption may not hold true for every event/log backend system. There is an inherent tradeoff between reliabilty/latency and the amount of data that a backend can allow; to provide a very tight latency SLO that caps 99%-ile latency, it's important to guarantee low variance in the size of requests which, in turn, may limit the amount of data that an event/log backend can accept (while being able to provide such a tight time bound). When using such a system for indexing the metadata about the events and making them quickly available for searching, it would still be useful to have a client-side mechanism to route larger content that is not critical for indexing to a blob storage system more suitable for large object storage and to be able to make it easy to interlink and navigate between backend systems by referencing the storage location.

## Replace/upload just the entire body of events/logs

While this might serve the purpose of "gen_ai.prompt" / "gen_ai.completion" if/when it moves from a span event to a more general event, it would still be necessary to provide a data model for representing that the event body had been uploaded/replaced with a reference. In addition, this would not cover many other cases (e.g. "http.request.body.content"), and -- beyond this -- it is desirable to upload a much more targeted/limited portion of the data in order to allow indexing / searching the other content that does fit within backend subsystem limits.

### Example 1: `http.response.body.content` _\[Span Attribute\]_

**Before**

```
# TracesData [Before]
resource_spans: {
  resource: { … }
  scope_spans: {
     scope: { … }
     spans: {
         trace_id: …
         span_id: …
         …
         attributes: {
            key: "http.response.body.content"
            value: { string_value: "{ \"results\": [ … ] " }   # very long API JSON repsonse
          }
         }
         …
     }
  }
}
```

**After**

```
# TracesData [After]
resource_spans: {
  resource: { … }
  scope_spans: {
     scope: { … }
     spans: {
         trace_id: …
         span_id: …
         …
         attributes: {
            key: "http.response.body.content.ref.uri"
            value: { string_value: "gs://some-bucket/traceAttachments/12345/7890/response.json" }
          }
          attributes: {
             key: "http.response.body.content.ref.content_type"
             value: { string_value: "application/json" }
            }
         }
         …
     }
  }
}
```

### Example 2: `gen_ai.prompt` / ` gen_ai.completion` _\[Span Event Attribute\]_

**Before**

```
# TracesData [Before]
resource_spans: {
  resource: { … }
  scope_spans: {
     scope: { … }
     spans: {
         trace_id: …
         span_id: …
         …
         events: {
           # Note that this precise event name is mandatory per:
           # https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/
           name: "gen_ai.content.prompt"
           # …
           attributes: {
              key: "gen_ai.prompt"
              value: { string_value: "Imagine that there is a very long text prompt here…." }
            }
           }
         }
        events: {
             # Note that this precise event name is mandatory per:
             # https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/
             name: "gen_ai.content.completion"
             attributes: {
                 key: "gen_ai.completion"
                 value: { string_value: "{ \"completions\": [...]}"  }  # very long completion JSON
              }
           }
        }
         …
     }
  }
}

```

**After**

```
# TracesData [After]
resource_spans: {
  resource: { … }
  scope_spans: {
     scope: { … }
     spans: {
         trace_id: …
         span_id: …
         …
         events: {
           # Note that this precise event name is mandatory per:
           # https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/
           name: "gen_ai.content.prompt"
           # …
           attributes: {
              key: "gen_ai.prompt.ref.uri"
              value: { string_value: "s3://bucket/some/path/to/the/prompt.txt" }
            }
            attributes: {
               key: "gen_ai.prompt.ref.content_type"
               value: { string_value: "text/plain" }
             }
           }
         }
        events: {
             # Note that this precise event name is mandatory per:
             # https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/
             name: "gen_ai.content.completion"
             attributes: {
                 key: "gen_ai.completion.ref.uri"
                 value: { string_value: "azblob://account/container/path/to/response.json" }
              }
              attributes: {
                  key: "gen_ai.completion.ref.content_type"
                  value: { string_value: "application/json" }
              }
           }
        }
         …
     }
  }
}
```

michaelsafyan:
Fixes #1428

## Changes

Adds the concept of "reference attributes" / "reference fields" to the specification.

### DIFF
--- a/docs/general/events.md
+++ b/docs/general/events.md
@@ -98,6 +98,14 @@ Recommendations on using attributes vs. body fields:
   * The _fields_ are unique to the named event (`event.name`) and different events
     may use the same _field_ name to represent different data, due to the unique
     nature of the event.
+  * The _fields_ SHOULD NOT use the `*.ref.*` name, except when applying the
+    same principle/data model as "attribute references" but in relation to
+    event fields (e.g. using `somefield.ref.uri` to relay the value of the
+    field `somefield` via indirection to an external storage system).
+  * For any _field_ name `somefield`, there exist a set of corresponding
+    fields `somefield.ref.*` that may be used to relay the value of `somefield`
+    via reference to an external storage system, following the same rules as
+    "attribute references" as outlined in the attribute naming conventions.
 
 ## External event compatibility
 
--- a/.chloggen/attribute_references.yaml
+++ b/.chloggen/attribute_references.yaml
@@ -1,0 +1,4 @@
+change_type: enhancement
+component: specification
+note: Defines "reference attribute" concept
+issues: [1428]
--- a/docs/general/attribute-naming.md
+++ b/docs/general/attribute-naming.md
@@ -46,17 +46,20 @@ Names SHOULD follow these rules:
   indicate entity hierarchies. This purpose should primarily drive the decision
   about forming nested namespaces.
 
+- The special namespace `*.ref.*` is reserved for "reference attributes".
+
 - For each multi-word dot-delimited component of the attribute name separate the
   words by underscores (i.e. use snake_case). For example
   `http.response.status_code` denotes the status code in the http namespace.
 
 - Names SHOULD NOT coincide with namespaces. For example if
-  `service.instance.id` is an attribute name then it is no longer valid to have
-  an attribute named `service.instance` because `service.instance` is already a
+  `service.instance.id` is an attribute name, then it is no longer valid to have
+  an attribute named `service.instance`, because `service.instance` is already a
   namespace. Because of this rule be careful when choosing names: every existing
   name prohibits existence of an equally named namespace in the future, and vice
   versa: any existing namespace prohibits existence of an equally named
-  attribute key in the future.
+  attribute key in the future. Note that "reference attributes" are exempt
+  from this rule, because they are/replace the original attribute.
 
 ## Name Pluralization Guidelines
 
@@ -172,6 +175,32 @@ and protocols.
 
 Any additions to the `otel.*` namespace MUST be approved as part of
 OpenTelemetry specification.
+
+## Reference Attributes
+
+A "reference attribute" is a set of derived attribute names that are used to
+provide the value of an attribute indirectly via a URI reference to a storage
+system where the value of the attribute may be retrieved.
+
+In general, if there exists an attribute `somekey`, then there is implicitly
+defined another attribute `somekey.ref.uri` which may be used to provide
+the value of the attribute `somekey` by reference to an external storage
+system from which the value of `somekey` may be fetched.
+
+Additional `*.ref.*` attributes, beyond `.ref.uri`, may be used to provide
+additional information concerning the external reference, including:
+
+  - `*.ref.content_type`: the MIME type of the data (e.g. `text/plain`, `application/json`, `application/octet-stream`)
+  - `*.ref.size`: the size of the attribute value in bytes
+  - `*.ref.hash_value`: a hash of the data for validation
+  - `*.ref.hash_algorithm`: the algorithm used to compute the hash
+
+If both a reference attribute and its non-reference variant appear together
+within a signal (e.g. both `somekey` and `somekey.ref.uri` are present),
+it should be assumed that only the reference attribute references the full,
+complete, original value of the data; the non-reference variant may be used
+to preview/summarize the data but should be assumed to potentially contain
+a modified, truncated, redacted, or otherwise non-original value within it.
 
 [DocumentStatus]:
   https://opentelemetry.io/docs/specs/otel/document-status


michaelsafyan:
Given the existence of the use of `ref` in the VCS attributes, I will update this PR with an alternative segment.

michaelsafyan:
Here are some possible naming alternatives to `ref`:

 - `external`: given that it will be in external storage
 - `ext`: shorter way of using the name "external"
 - `foreign`: given that it is in a foreign storage location
 - `remote`: given that it may be in another storage location
 - `ext_storage`: similar mash up of the above
 - `remote_storage`: similar mash up of the above
 - `blob_ref`: another possible name

Any other thoughts? Any opinions on what is best?

michaelsafyan:
I've updated the pull request to use `*.blob_ref.*` instead of `*.ref.*` but am open to alternative naming suggestions.

christophe-kamphaus-jemmic:
I think `blob_ref` is a good choice.

Should there be some tooling to make sure we don't accidentally define anything `blob_ref.*` in the future?
Eg. could we add a policy in `policies/` to flag a violation if `blob_ref.*` is detected?

docs/general/events.md
```diff
@@ -98,6 +98,15 @@ Recommendations on using attributes vs. body fields:
   * The _fields_ are unique to the named event (`event.name`) and different events
     may use the same _field_ name to represent different data, due to the unique
     nature of the event.
+  * The _fields_ SHOULD NOT use the `*.blob_ref.*` name, except when applying the
```
jsuereth, line 101: NIt:  You can provide a policy to enforce this in the `/policies` directory.  Would like to see this happen after this naming and feature is approved.

michaelsafyan: Sounds good. I'm happy to either add this enforcement in a separate PR or to add to the existing PR. Would it be better to do that separately or in a single PR?

docs/general/events.md
```diff
@@ -98,6 +98,15 @@ Recommendations on using attributes vs. body fields:
...
+    same principle/data model as "attribute references" but in relation to
+    event fields (e.g. using `somefield.blob_ref.uri` to relay the value of the
+    field `somefield` via indirection to an external storage system).
+  * For any _field_ name `somefield`, there exist a set of corresponding
```
aabmass, line 105: How would you represent when the entire body/payload is being referenced in blob storage? For example if the original payload was a large string/bytes/ArrayValue or you wanted to reference an entire KeyValueList and not subfields

michaelsafyan: Thanks. My thought here is that a subfield named `blob_ref.uri` being present (no prefix before `blob_ref`) would indicate that the entire `AnyValue` (rather than a specific field of it) was written-aside.

I will amend this proposal to make that explicit...

docs/general/events.md
michaelsafyan: I've since added `*.rego` policy files for this. OK to resolve?

docs/general/events.md
michaelsafyan: I've amended the PR to explicitly cover this case. OK to resolve?

policies/blob_reference_property_reserved_segment.rego
```diff
@@ -0,0 +1,48 @@
...
+
+import rego.v1
+
+reserved_segment_name = "blob_ref"
+
+deny contains violation if {
+	group := input.groups[_]
+	group.prefix != null
+	contains_segment(group.prefix, reserved_segment_name)
+	description := sprintf("Registry group '%s' with prefix '%s' contains illegal segment '%s'; reserved for Blob Reference Properties.", [group.id, group.prefix, reserved_segment_name])
```
jsuereth, line 11: nit: Can you add a URL to link to your blob-reference-properties.md file?

jsuereth (commented):
Policy looks great!

Looking for more input from @open-telemetry/specs-semconv-maintainers

docs/general/events.md
aabmass: Yes thanks!

policies/blob_reference_property_reserved_segment.rego
michaelsafyan: Done.

michaelsafyan:
Received feedback in [Nov 18 SemConv Meeting](https://docs.google.com/document/d/10xG7DNKWRhxNmFGt3yYd3980a9uwS8lMl2LvQL3VNK8/edit?tab=t.0#bookmark=id.dm08tj1baisu) that this PR needs to be amended to remove "attributes" from the scope, because of the possibility of encouraging large attributes. 

Was also suggested to narrow scope to define a specific event type that contains its values as a reference.

PR #1523, opened 2024-10-28
add `http.request.synthetic` attribute to server spans and metrics

### LINKED ISSUE #1127: Introduction of a Synthetic Attribute for Server Span Telemetry
### Area(s)

area:browser

I would like to be able to identify telemetry created by synthetic sources such as bots or crawlers. This issue looks to work on defining conventions surrounding marking spans as originating from a synthetic source. 

I would like to introduce an attribute to HTTP server span semantic conventions, as well as metrics and logs that represents a low-cardinality string such as the below:

synthetic -> `"not set" | "bot" | "synthetic test"`

Where the synthetic attribute being set to `"not set"` represents telemetry that is not generated from a synthetic source. This convention will be helpful for scenarios where a user may want to mark telemetry generated from frequent synthetic tests or web crawlers separately from direct human engagement. 

The determination of which of the three options a span falls into could be made by maintaining a list of known synthetic sources or allowing this decision to be user configurable.

While we could consider setting the `synthetic` attribute to a Boolean value, I believe the extra granularity of the low-cardinality string would be valuable.

### Additional context

_No response_

JacksonWeber:
Fixes #1127

## Changes

Adds the `http.request.synthetic` attribute to server spans and metrics to track if telemetry is generated from a synthetic testing source, some kind of bot, or a real user. This will help users filter out telemetry generated by these kinds of synthetic sources that they may not want to consider. 


model/http/registry.yaml
```diff
@@ -170,3 +170,21 @@ groups:
...
+        brief: >
+          A flag indicating that the request was synthetically created and did not originate from genuine client traffic.
+        note: >
+          This flag can primarily be determined by the contents of the `user_agent.original` attribute. Instrumentations should determine what they consider synthetic or bot traffic,
+          and set this attribute accordingly. This attribute is useful for distinguishing between genuine client traffic and synthetic traffic generated by bots or tests.
+        type:
+          members:
+            - id: bot
+              value: "bot"
+              brief: 'bot source.'
```
trask, line 184:
```suggestion
              brief: 'Bot source.'
```

model/http/registry.yaml
```diff
@@ -170,3 +170,21 @@ groups:
...
+              stability: experimental
+            - id: test
+              value: "synthetic test"
+              brief: 'synthetic test source.'
```
trask, line 188:
```suggestion
              brief: 'Synthetic test source.'
```

model/http/registry.yaml
```diff
@@ -170,3 +170,21 @@ groups:
         brief: State of the HTTP connection in the HTTP connection pool.
         stability: experimental
         examples: ["active", "idle"]
+      - id: http.request.synthetic
+        stability: experimental
+        brief: >
+          A flag indicating that the request was synthetically created and did not originate from genuine client traffic.
+        note: >
+          This flag can primarily be determined by the contents of the `user_agent.original` attribute. Instrumentations should determine what they consider synthetic or bot traffic,
```
trask, line 178: if the primary use case is related to `user_agent.original`, maybe renaming `http.request.synthetic` to `user_agent.synthetic` could be considered?

JacksonWeber: Basing primarily on user-agent seems to be consistent across approaches to identifying synthetic telemetry. I'd expect any instrumentation's solution to populating this attribute to be based (at least in-part) on the `user_agent.orginal` field. My original thinking was that we're marking the `request` as synthetic telemetry, not the `user_agent`, but like @lmolkova pointed out on #1127 it's not a guarantee that that this attribute will only apply to `http` so I agree that this may be more appropriate as a `user_agent` field.

I'll adjust this PR accordingly.

.chloggen/add-synthetic-source.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: http
```
lmolkova, line 10:
```suggestion
component: user_agent
```

.chloggen/add-synthetic-source.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add the http.request.synthetic attribute to track if spans and metrics are the result of real users, synthetic testing, or bots.
```
lmolkova, line 13: please update the attribute name

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,20 @@ groups:
...
+          This flag can primarily be determined by the contents of the `user_agent.original` attribute. Instrumentations should determine what they consider synthetic or bot traffic,
+          and set this attribute accordingly. This attribute is useful for distinguishing between genuine client traffic and synthetic traffic generated by bots or tests.
+        type:
+          members:
+            - id: bot
+              value: "bot"
+              brief: 'Bot source.'
+              stability: experimental
+            - id: test
+              value: "synthetic test"
```
lmolkova, line 52:
```suggestion
              value: "test"
```

is `synthetic test` necessary? why not just `test`?

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,20 @@ groups:
           using a user-agent for non-browser products, such as microservices with multiple names/versions inside the
           `user_agent.original`, the most significant version SHOULD be selected. In such a scenario it should align
           with `user_agent.name`
+      - id: user_agent.synthetic
```
lmolkova, line 38: the `user_agent.synthetic` would work great for a boolean flag, but  we have an enum that'll probably grow (we might wat to differentiate `crawler` from `bot` or have a `healthcheck`, `monitoring` in addition to `test`)

So I'm suggesting to find a more precise name.

E.g.  `user_agent.type` that could cover all possible types, synthetic or not - it don't need to be populated for the typical traffic. This might be a bit too broad and people might use it to store something like `mobile` or `browser`- I'm not sure if that's what we want.

Alternative could be `user_agent.synthetic.type` that would only apply to the synthetic requests.

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,20 @@ groups:
...
+        stability: experimental
+        brief: >
+          A flag indicating that the user agent represents a synthetic source and did not originate from genuine client traffic.
```
lmolkova, line 41: Most of the traffic probably happens between services. Can we clarify what the synthetic means?

My interpretation is:

```suggestion
         Specifies the category of synthetic traffic, such as monitoring, crawler, bot, or another automation.
```

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,20 @@ groups:
...
+        note: >
+          This flag can primarily be determined by the contents of the `user_agent.original` attribute. Instrumentations should determine what they consider synthetic or bot traffic,
```
lmolkova, line 43: It there any prior art we can refer to? E.g. a well-known database of crawlers/bots?

model/user-agent/registry.yaml
JacksonWeber: I really like the `user_agent.synthetic.type` suggestion. I can see how it might spark confusion to have something like `user_agent.synthetic` since it implies from the naming that this would be boolean. Updating.

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,20 @@ groups:
           using a user-agent for non-browser products, such as microservices with multiple names/versions inside the
           `user_agent.original`, the most significant version SHOULD be selected. In such a scenario it should align
           with `user_agent.name`
+      - id: user_agent.synthetic.type
```
jsuereth, line 38: Asked this in chat, but also asking here.

Should this (also) be added to client spans so synthetic agents can self-identify in a trace?

e.g. https://opentelemetry.io/blog/2023/synthetic-testing/

JacksonWeber: Makes sense to me to allow agents to self-identify and allow for propagation of `user_agent.synthetic.type` to any server spans created in response to the remote client span from the synthetic agent.

model/user-agent/registry.yaml
JacksonWeber: There are a number sites that maintain lists of the most popular. So far, the best I've found is Data Dome's list of those most popular in 2024 https://datadome.co/bot-management-protection/crawlers-list/.

model/user-agent/registry.yaml
trask: these sound like two different (though potentially both useful) things:

> allow agents to self-identify

and

> allow for propagation of `user_agent.synthetic.type` to any server spans created in response to the remote client span from the synthetic agent

I'd suggest sticking to just the first in this PR

JacksonWeber: I agree that both are useful. I'd be happy to keep this PR focused for specific review of the `synthetic.type` on server spans. @jsuereth does limiting the scope here work for you?

JacksonWeber: @jsuereth per our discussion in the semantic convention WG, I've added further information regarding the usefulness of this attribute for client self-identification as synthetic.

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,21 @@ groups:
...
```
trask, line 38: often we have tried to differentiate between a default value (e.g. in this case the user agent is not synthetic) and a missing value (e.g. in this case not supported by instrumentation)

I think it may be ok in this case to have the ambiguity, but just wanted to make sure we discuss / agree on that

model/user-agent/registry.yaml
trask: cc @open-telemetry/javascript-approvers

trask: > I'd expect the attribute to be populated by a processor (in the SDK or collector) - this way it's easier to update and modify configuration (e.g. add new user-agents to the list of bots).

this is a good point, though I'm not sure if it would work for capturing `user_agent.synthetic.type` on metrics?

lmolkova: oh, right. It's hard to imagine though that we'd add a static regex/dictionary to individual instrumentation libraries in all languages to populate something as dynamic as list of bots. E.g. this repo https://github.com/omrilotan/isbot updated patterns ~20 times during this year alone https://github.com/omrilotan/isbot/commits/main/src/patterns.json

Also it's not the only thing people would want to extract from the user-agent.
It seems to be a common thing to parse user_agent and build analytics on top of its components (e.g. https://github.com/open-telemetry/semantic-conventions/pull/1434 is about adding OS component). ECS defines a few other properties to extract  - https://www.elastic.co/guide/en/ecs/current/ecs-user_agent.html

We could try making the whole `user_agent.original` opt-in on metrics with expectation that processor would extract individual parts and remove the original string, but the local memory consumption could be too high due to cardinality. 

Another approach is that instrumentations might have a hook accepting external UserAgent parser that should be able to populate multiple attributes from the string.

model/user-agent/registry.yaml
JacksonWeber: That's a good point; we did discuss this on the original issue here: https://github.com/open-telemetry/semantic-conventions/issues/1127#issuecomment-2397444676. And I'm personally in favor of simply not populating unless there are serious concerns regarding differentiation.

model/user-agent/registry.yaml
JacksonWeber: I like the greater scope of your proposed wording so we don't rule out any relevant components which might want to set this value.

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,21 @@ groups:
...
+        stability: experimental
+        brief: >
+         Specifies the category of synthetic traffic, such as monitoring, crawler, bot, or another automation.
+        note: >
+          This flag can primarily be determined by the contents of the `user_agent.original` attribute. Instrumentations should determine what they consider synthetic or bot traffic,
+          and set this attribute accordingly. This attribute can either be set on client spans for self-identification purposes, or on server spans detected to be generated as a result
```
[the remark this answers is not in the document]
JacksonWeber: The scope of adding this attribute for this PR was discussed here: https://github.com/open-telemetry/semantic-conventions/pull/1523/#discussion_r1828270167. I'm happy to remove this specific mention of client spans and just describe that the attribute can be used for self-identification reasons.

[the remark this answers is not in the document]
trask: right now, the PR only adds the new attribute under HTTP server spans. I think @lmolkova is suggesting to also add it under HTTP client spans.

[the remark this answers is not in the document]
JacksonWeber: I see, I've added the synthetic attribute to client spans so it can be used for self-identification purposes as was discussed. Let me know if that satisfies your concerns @lmolkova

docs/http/http-metrics.md
```diff
@@ -143,6 +144,8 @@ SHOULD include the [application root](/docs/http/http-spans.md#http-server-defin
 > Since this attribute is based on HTTP headers, opting in to it may allow an attacker
 > to trigger cardinality limits, degrading the usefulness of the metric.
 
+**[10] `user_agent.synthetic.type`:** This flag can primarily be determined by the contents of the `user_agent.original` attribute. Components that populate the attribute should determine what they consider synthetic or bot traffic, and set this attribute accordingly. This attribute can either be set for self-identification purposes, or on spans detected to be generated as a result of a synthetic request. This attribute is useful for distinguishing between genuine client traffic and synthetic traffic generated by bots or tests.
```
lmolkova, line 147: nit: we should not mention spans on attribute definition since it's used on different signals

lmolkova (approved):
LGTM!

jsuereth (approved):

.chloggen/add-synthetic-source.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: user_agent
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add the user_agent.synthetic.type attribute to track if spans and metrics are the result of real users, testing, or bots.
```
trask, line 13:
```suggestion
note: Add the user_agent.synthetic.type attribute to specify the category of synthetic traffic, such as tests or bots.
```

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,21 @@ groups:
           using a user-agent for non-browser products, such as microservices with multiple names/versions inside the
           `user_agent.original`, the most significant version SHOULD be selected. In such a scenario it should align
           with `user_agent.name`
+      - id: user_agent.synthetic.type
+        stability: experimental
+        brief: >
+         Specifies the category of synthetic traffic, such as monitoring, crawler, bot, or another automation.
```
trask, line 41: when reading this, I was expecting the enum to have values for `monitoring`, `crawler`, and maybe `other`, maybe could keep it simple
```suggestion
         Specifies the category of synthetic traffic, such as tests or bots.
```

.chloggen/add-synthetic-source.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: user_agent
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add the user_agent.synthetic.type attribute to specify the category of synthetic traffic, such as tests or bots.
```
trask, line 13:
```suggestion
note: Add the `user_agent.synthetic.type` attribute to specify the category of synthetic traffic, such as tests or bots.
```

model/http/spans.yaml
```diff
@@ -45,6 +45,9 @@ groups:
         requirement_level: opt_in
       - ref: http.response.body.size
         requirement_level: opt_in
+      - ref: user_agent.synthetic.type
+        requirement_level:
+          recommended: When the client knows it is sending synthetic or bot traffic.
```
trask, line 50:
```suggestion
          recommended: When the client knows it is sending synthetic traffic, such as tests or bots.
```

model/user-agent/registry.yaml
```diff
@@ -35,3 +35,21 @@ groups:
...
           `user_agent.original`, the most significant version SHOULD be selected. In such a scenario it should align
           with `user_agent.name`
+      - id: user_agent.synthetic.type
+        stability: experimental
+        brief: >
+         Specifies the category of synthetic traffic, such as tests or bots.
+        note: >
+          This flag can primarily be determined by the contents of the `user_agent.original` attribute. Components that populate the attribute should determine what they consider synthetic or bot traffic,
+          and set this attribute accordingly. This attribute can either be set for self-identification purposes, or on telemetry detected to be generated as a result
+          of a synthetic request. This attribute is useful for distinguishing between genuine client traffic and synthetic traffic generated by bots or tests.
```
trask, line 45: (removed the word "should", or could change it to "SHOULD")
```suggestion
          This attribute MAY be derived from the contents of the `user_agent.original` attribute.
          Components that populate the attribute are responsible for determining what they consider to be synthetic bot or test traffic.
          This attribute can either be set for self-identification purposes, or on telemetry detected to be generated as a result
          of a synthetic request. This attribute is useful for distinguishing between genuine client traffic and synthetic traffic generated by bots or tests.
```

trask (approved):

model/http/spans.yaml
```diff
@@ -45,6 +45,8 @@ groups:
         requirement_level: opt_in
       - ref: http.response.body.size
         requirement_level: opt_in
+      - ref: user_agent.synthetic.type
+        requirement_level: opt_in
```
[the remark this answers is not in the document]
lmolkova: oh, it has to be `opt_in` since it's an experimental attribute on a stable convention. We should be able to add comment to opt_in with the next weaver release.

model/user-agent/registry.yaml
JacksonWeber: I think MAY makes sense here as it's not a requirement, but simply a possible choice.

trask:
@JacksonWeber can you check the CI failures? feel free to ping me if you need any help with them

PR #1438, opened 2024-09-30
Cosmos DB: Operation Level Metrics

sourabh1007:
Introducing operation-level metrics in the .NET Cosmos DB SDK.
This PR includes the registration of semantic conventions in the OpenTelemetry community, ensuring full alignment with OpenTelemetry's established guidelines.

**Dimensions**
Tag/dimension name | Sample value
-- | --
db.system | cosmodb
db.collection.name | myCollectionName
db.namespace | myDatabaseName
server.address | myaccountname.documents.azure.com
server.port | 443
db.operation.name | query_items
~~db.cosmosdb.operation_type~~ | ~~`batch`; `create`; `delete`~~
db.response.status_code | 200 or 429 etc.
db.cosmos.sub_status_code | 1002 etc.
~~db.cosmos.client_id~~ | ~~Guid~~
db.cosmosdb.consistency_level | Eventual, ConsistentPrefix, BoundedStaleness, Strong or Session
~~db.cosmosdb.machine_id~~ | ~~Unique Id for a machine~~ (Dependent of otel generated machine details)

**Metrics**
Name | Unit | Type | Description
-- | -- | -- | --
db.cosmosdb.operation.request_charge | `{request_units}` | histogram | Total   request units per operation (sum of RUs for all requested needed when   processing an operation)   This metric SHOULD be specified with `[ExplicitBucketBoundaries] of [ 1, 5, 10, 25, 50, 100, 250, 500, 1000]`
db.client.operation.duration | `s` | histogram | Total   end-to-end duration of the operation   This metric SHOULD be specified with `[ExplicitBucketBoundaries] of [0.001, 0.005, 0.010,  0.050, 0.100, 0.200, 0.500, 1.000, 2.000, 5.000]`
~~db.query.limit~~ | `{count}` | histogram | ~~For feed operations (query, readAll, readMany, change feed) and batch operations, this meter capture the requested maxItemCount per page/request. This metric SHOULD be specified with `[ExplicitBucketBoundaries] of [50, 100, 250, 500, 1000, 2000, 5000]`~~
db.client.response.row_count | `{count}` | histogram | For   feed operations (query, readAll, readMany, change feed) batch operations this   meter capture the actual item count in responses from the service. This metric SHOULD be specified with `[ExplicitBucketBoundaries] of [10, 50, 100, 250, 500, 1000, 2000, 5000, 10000]`.
[...]

PR #1525, opened 2024-10-29
Cosmos DB: Added regions_contacted as new dimension

sourabh1007:
**Dimensions**
Tag/dimension name | Sample value | Description
-- | -- | --
db.cosmosdb.regions_contacted | "North Central US,Central US", "East US 2,North Central US" | It would help us to identify if operation call was cross regional or single regional calls

**Metrics**
Name | Unit | Type | Description
-- | -- | -- | --
~~db.cosmosdb.operation.count~~ | `{operations}` | counter | Total number of executed operations. This metric will help assess the success and failure rates of operations and identify patterns in case of failures.

ref. https://github.com/open-telemetry/semantic-conventions/pull/1438
 and https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/docs/Metrics.md


.chloggen/users_sourabhjain_moreoperationmetrics.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Added `db.client.operation.count` metric to track the number of operations performed by the client.
```
trask, line 13: do we need `db.client.operation.count`? e.g. we don't (currently) have `http.(client|server).request.count`, under the assumption that you can get count from `http.(client|server).request.duration`

docs/database/cosmosdb.md
```diff
@@ -42,29 +42,32 @@ Cosmos DB instrumentation includes call-level (public API) surface spans and net
...
+| [`user_agent.original`](/docs/attributes-registry/user-agent.md) | string | Full user-agent string is generated by Cosmos DB SDK [16] | `cosmos-netstandard-sdk/3.23.0\|3.23.1\|1\|X64\|Linux 5.4.0-1098-azure 104 18\|.NET Core 3.1.32\|S\|` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`db.query.parameter.<key>`](/docs/attributes-registry/db.md) | string | A query parameter used in `db.query.text`, with `<key>` being the parameter name, and the attribute value being a string representation of the parameter value. [17] | `someval`; `55` | `Opt-In` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
 **[1]:** It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.
 
 **[2]:** if not `gateway` (the default value is assumed to be `gateway`).
 
-**[3]:** The `db.operation.name` has the following list of well-known values.
+**[3]:** When the `db.cosmosdb.regions_contacted` attribute is populated, it indicates that the operation was performed on multiple regions i.e. cross-regional call.
```
jcocchi, line 68: This statement is misleading because it will be populated for single region or multi region operations. Let's modify slightly

```suggestion
**[3]:** The `db.cosmosdb.regions_contacted` attribute is populated with all regions that were contacted during the operation. If there is more than one region listed, it indicates that the operation was performed on multiple regions i.e. cross-regional call.
```

model/database/registry.yaml
```diff
@@ -543,7 +543,11 @@ groups:
         stability: experimental
         brief: Account or request [consistency level](https://learn.microsoft.com/azure/cosmos-db/consistency-levels).
         examples: ["Eventual", "ConsistentPrefix", "BoundedStaleness", "Strong", "Session"]
-
+      - id: db.cosmosdb.regions_contacted
+        type: string
```
trask, line 547: this feels like it should be a `string[]`

I'm not sure if we've put any `string[]` attributes on metrics yet though...

I've added this to Monday's semconv SIG agenda to get more feedback in case we don't get more feedback here (and I'm off the rest of this week)

docs/database/cosmosdb.md
```diff
@@ -338,6 +341,7 @@ Explaining bucket configuration:
 |---|---|---|---|---|---|
 | [`db.collection.name`](/docs/attributes-registry/db.md) | string | Cosmos DB container name. [1] | `public.users`; `customers` | `Conditionally Required` If available | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`db.cosmosdb.consistency_level`](/docs/attributes-registry/db.md) | string | Account or request [consistency level](https://learn.microsoft.com/azure/cosmos-db/consistency-levels). | `Eventual`; `ConsistentPrefix`; `BoundedStaleness`; `Strong`; `Session` | `Conditionally Required` If available. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.cosmosdb.regions_contacted`](/docs/attributes-registry/db.md) | string | List of regions contacted by the client during a given operation. | `North Central US,Central US`; `East US 2,North Central US` | `Conditionally Required` If available | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 344: how would someone use this? would you filter by a certain combination? Check which combinations result in the highest latency? Does the order matter?

.chloggen/users_sourabhjain_moreoperationmetrics.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Added `db.cosmosdb.regions_contacted` attribute as dimension for cosmos db metrics.
```
lmolkova, line 13: It seems it's not set on metrics, is this note accurate?

sourabh1007: https://github.com/open-telemetry/semantic-conventions/blob/9db67a5b37104dfa06beb9efe7e0974799ffb83c/model/database/common.yaml#L63 added it as common dimension

docs/database/cosmosdb.md
sourabh1007: In a metric, we can compare latencies for records where the dimension is set to either 'Region1' alone or 'Region1, Region2.' If the metric shows a latency value for 'Region1, Region2,' it indicates that the SDK is making cross-regional calls, which results in higher latency for the customer."
No order doesn't matter there.

model/database/registry.yaml
sourabh1007: how does it matter? if its string or string[]

trask: since the underlying data is a list, it should ideally be modeled that way (instead of encoded into a comma-separated string)

e.g. see `http.request.header.<key>`

model/database/registry.yaml
```diff
@@ -543,7 +543,12 @@ groups:
         stability: experimental
         brief: Account or request [consistency level](https://learn.microsoft.com/azure/cosmos-db/consistency-levels).
         examples: ["Eventual", "ConsistentPrefix", "BoundedStaleness", "Strong", "Session"]
-
+      - id: db.cosmosdb.regions_contacted
+        type: string[]
+        stability: experimental
+        brief: Comma separated list of regions contacted by the Azure Cosmos DB operation.
```
trask, line 549: does order matter?

does case-sensitivity matter?

sourabh1007: order doesn't matter and case-sensitivity, ideally should not matter as I would never see _West US 2_ and _WEST US 2_ together. and both would be representing same region.

trask: does it make sense to normalize these in any way, so that e.g. different orders don't create different time series (different attribute sets)?

.chloggen/users_sourabhjain_moreoperationmetrics.yaml
lmolkova, line 13:
```suggestion
note: Added `db.cosmosdb.regions_contacted` attribute to Cosmos DB metrics and traces.
```

model/database/registry.yaml
lmolkova: also can we normalize the values in the list and list all of the supported values and/or how to normalize it?

The goal is to have cross-language consistency and alignment with other areas (e.g. REST API values, some existing Azure location list, etc).

model/database/cosmosdb-metrics.yaml
```diff
@@ -7,6 +7,12 @@ groups:
     unit: "{request_unit}"
     stability: experimental
     extends: attributes.db.cosmosdb.minimal
+    attributes:
+      - ref: db.cosmosdb.regions_contacted
```
lmolkova, line 11: creating an array per each call could be expensive - did we consider this? 

One way to approach it could be to add some config option to turn the attribute off in the future,

For now I think it's important to leave a room for it - in particular change `requirement_level` to recommended  instead of `conditionally_required`

sourabh1007: creating an array per each call could be expensive - did we consider performance aspect of this attribute? => we already have it in `Set`, we really don't have to do much to convert it to `array` or `string` and I am fine with anything.

@kirankumarkolli @jcocchi what do you think about making it opt-in? if customer is on multi-region set up, only then this dimension would be required.

model/database/registry.yaml
jcocchi: @sourabh1007 is the list ordered in the same sequence that the regions were actually contacted by the SDK? If that's the case I think we should leave order as is, if the order can be "random" then we should alphabetize or some other standarization

model/database/cosmosdb-metrics.yaml
jcocchi: I don't have a preference on string vs. array type, but I think having an opt in for `regions_contacted` might be overly complicated. What scenarios would we recommend vs. not recommend it? What would the default be? 

For cases we would recommend not having it, the impact is low. If a customer only has one region then the list would always be a single value. For two region set ups which is common, there would only be 4 total possible values here. For high impact cases with a lot of permutations, the attribute is very useful and we would recommend they capture it

model/database/registry.yaml
sourabh1007: it is random order list. we can sort it.

model/database/cosmosdb-metrics.yaml
sourabh1007: I agree, we can have it always by default.

model/database/registry.yaml
lmolkova: Let's document the format of the region. As we discussed in the https://github.com/open-telemetry/semantic-conventions/pull/1543#discussion_r1831606621 is should be the `nae` or `displayName` from https://learn.microsoft.com/en-us/rest/api/subscription/subscriptions/list-locations?view=rest-subscription-2021-10-01&tabs=HTTP#location.

BTW I don't understand why `displayName` is preferred - it has spaces and case-sensitive things are always problematic in user queries. I'd use `name`

model/database/registry.yaml
```diff
@@ -543,7 +543,15 @@ groups:
...
-
+      - id: db.cosmosdb.regions_contacted
+        type: string[]
+        stability: experimental
+        brief: >
+          List of regions contacted by the Azure Cosmos DB operation.
+          Format of region name would be similar to the "displayname"
+          mentioned [here](https://learn.microsoft.com/en-us/rest/api/subscription/subscriptions/list-locations?view=rest-subscription-2021-10-01&tabs=HTTP#location)
+        examples:
+          - ['North Central US']
```
lmolkova, line 554:
```suggestion
        brief: List of regions contacted during operation in the order specific region was contacted.
        note: >
          Region name matches the format of `displayName` in [Azure Location API](https://learn.microsoft.com/rest/api/subscription/subscriptions/list-locations?view=rest-subscription-2021-10-01&tabs=HTTP#location)
        examples:
          - ['North Central US']
          - ['Australia East, 'Australia Southeast]
```

model/database/registry.yaml
lmolkova: Unresolving this, as discussed it should be an array.

model/database/cosmosdb-metrics.yaml
lmolkova: unresolving this. It's still conditionally_required. The recommendation is to make it `recommended: if available` to make it future-proof (allow opting out)

model/database/cosmosdb-metrics.yaml
```diff
@@ -7,6 +7,12 @@ groups:
...
+        brief: >
+          List of regions contacted by the client during a given operation.
```
trask, line 13: is overriding the brief needed here, or is it ok to inherit the brief in the registry?

model/database/spans.yaml
```diff
@@ -737,3 +737,9 @@ groups:
       - ref: db.cosmosdb.consistency_level
         requirement_level:
           conditionally_required: If available.
+      - ref: db.cosmosdb.regions_contacted
+        note: >
+            The `db.cosmosdb.regions_contacted` attribute is populated with all regions that were contacted during the operation.
+            If there is more than one region listed, it indicates that the operation was performed on multiple regions i.e. cross-regional call.
```
trask, line 743: is it need to override the note here, or is it ok to inherit the note from the registry?

model/database/cosmosdb-metrics.yaml
trask: @sourabh1007 did you intend to resolve this discussion?

sourabh1007: yes i have made the changes.

trask (approved):

lmolkova (approved):

PR #1527, opened 2024-10-29
Rename `db.collection.name` to `db.target.name`

### LINKED ISSUE #1491: [database] reporting stored procedures/prepared statements
See https://github.com/open-telemetry/semantic-conventions/pull/1482#discussion_r1805142038

E.g. query like `CALL procedure-name` should result in the following attributes:

`db.query.text = "CALL procedure-name"`
`db.operation.name = "CALL"`
`db.collection.name = "procedure-name"`

We should document that collection name can contain stored procedure/prepared statement name

lmolkova:
Fixes #1491, #521

## Changes

`db.collection.name` does not work well for all possible database objects such as stored procedures, user defined functions, triggers, users (e.g. when performing `GRANT` queries).

This PR renames it to `db.target.name` to cover all these cases


schema-next.yaml
```diff
@@ -8,6 +8,13 @@ versions:
         # https://github.com/open-telemetry/semantic-conventions/pull/1422
         - rename_metrics:
             messaging.client.published.messages: messaging.client.sent.messages
+    all:
+      changes:
+        # TODO
```
trask, line 13:
```suggestion
        # https://github.com/open-telemetry/semantic-conventions/pull/1527
```

model/database/spans.yaml
```diff
@@ -248,13 +250,13 @@ groups:
         requirement_level:
           conditionally_required: If applicable.
         examples: ['mynamespace']
-      - ref: db.collection.name
+      - ref: db.target.name
         sampling_relevant: true
-        brief: The HBase table name.
+        brief: The HBase table or namespace name.
```
trask, line 255:
```suggestion
        brief: The HBase table or another entity that the operation is acting upon.
```

model/database/common.yaml
```diff
@@ -72,10 +72,10 @@ groups:
...
           recommended: if readily available or if instrumentation supports query summarization.
-      - ref: db.collection.name
+      - ref: db.target.name
         sampling_relevant: true
         requirement_level:
           conditionally_required: >
-            If readily available and if a database call is performed on a single collection.
-            The collection name MAY be parsed from the query text,
-            in which case it SHOULD be the single collection name in the query.
+            If readily available and if a database call is performed on a single target.
```
trask, line 79:
```suggestion
            If readily available and if a database operation is performed on a single target.
```

model/database/common.yaml
```diff
@@ -42,7 +42,7 @@ groups:
       # - ref: db.system
       #   requirement_level:
       #     conditionally_required: if available
-      - ref: db.collection.name
+      - ref: db.target.name
         brief: >
           Cosmos DB container name.
```
trask, line 47:
```suggestion
          Cosmos DB container or another entity that the operation is acting upon.
```

trask (approved):

XSAM (approved):

bogdandrutu:
Few comments here:
1. There may be multiple tables/functions (only one store procedure). Because of that I think we should make it a list of strings this argument.
2. Ideally I would separate "storage" targets from "executions" targets. The reason is that the "storage" targets like tables will not have traces, but "executions" targets will have.

If you comment that if traces exists for other executions then the spans will be available, that is not the case in some systems, since the function/procedure are user-code and traces may be in a different system than the platform traces.

model/database/registry.yaml
```diff
@@ -5,23 +5,31 @@ groups:
...
+          For data manipulation operations, target usually matches database table,
+          view, stored procedure or a user-defined function name.
+          For data definition or administrative operations, target would match the
+          database, index, user, or some other object type.
+
+          It is RECOMMENDED to capture the value as provided by the application
+          without attempting to do any case normalization.
 
-          A single database query may involve multiple collections.
+          A single database query may involve multiple targets.
```
bogdandrutu, line 23: Understand, and what should you do then? I think this statement is not useful, keep only line 32.

cbandy:
> 1. There may be multiple tables/functions (only one store procedure). Because of that I think we should make it a list of strings this argument.

The same can be said of `db.namespace`, perhaps.

My opinion is that deeper insights belong in more specific fields. If a database (engine/system) can or wants to indicate that the operation involves multiple tables, indices, tablets, pages, locks, what-have-you; then it can go in a tech-specific field.

lmolkova:
@bogdandrutu 

> There may be multiple tables/functions (only one store procedure). Because of that I think we should make it a list of strings this argument.

Array will be problematic on metrics and will be significantly harder to use for noSQL databases which primarily operate on one collection. It also has much higher cardinality.

Having array of operations and array of collections will be even worse for usability.

That's why we introduced `db.query.summary` - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#generating-a-summary-of-the-query-text which would contain something like `INSERT shipping_details SELECT orders` for query like 

```sql
INSERT INTO shipping_details
            (order_id,
            address)
SELECT order_id,
       address
FROM   orders
WHERE  order_id = ?
```

and will be a better representation of a query than arrays of operations and tables.

> Ideally I would separate "storage" targets from "executions" targets. The reason is that the "storage" targets like tables will not have traces, but "executions" targets will have.

I'm not sure I understand what the difference between storage and execution is. `CREATE TABLE foo ...` or `CREATE USER user_name` is traced in the same way as `SELECT * from foo where ...`.  Instrumentations are not even required to parse query text and will create spans in the same way for them.

> If you comment that if traces exists for other executions then the spans will be available, that is not the case in some systems, since the function/procedure are user-code and traces may be in a different system than the platform traces.

We're only defining client DB conventions now.

lmolkova:
@cbandy 

> The same can be said of db.namespace, perhaps.
> My opinion is that deeper insights belong in more specific fields. If a database (engine/system) can or wants to indicate that the operation involves multiple tables, indices, tablets, pages, locks, what-have-you; then it can go in a tech-specific field.

Deeper insights are available on `db.query.text`. Query text is not available on metrics (by default) due to high0ish cardinality, so we introduced `db.query.summary` that contains perations and targets - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#generating-a-summary-of-the-query-text

`db.namespace` (for SQL database) is defined as `The database associated with the connection` - you should not have more than one at the same time. 

Cross-DB queries in common case would have fully-qualified table names, i.e. multiple namespaces will be captured as a part of each target name and inside `db.query.text|summary`.  Generic instrumentation cannot really say if something in the table name represents an database or a schema or, maybe that database supports `.` in the table name itself.

model/database/registry.yaml
lmolkova: makes sense, updated the wording

jsuereth (approved):

trask (changes requested):
putting a temporary block, as @lmolkova and I are reconsidering this PR

lmolkova:
Closing this since we're pursuing a different option to represent stored procedures/other db object name.

PR #1528, opened 2024-10-29
[chore] Add messaging stabilization notice to all pages

joaopgrassi:
## Changes

Adds the full stabilization notice to all messaging-related pages.


lmolkova (approved):
Thank you!

docs/messaging/README.md
```diff
@@ -11,6 +11,34 @@ path_base_for_github_subdir:
 
 This document defines semantic conventions for messaging systems spans, metrics and logs.
 
+> [!Warning]
+>
+> Existing messaging instrumentations that are using
+> [v1.24.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/messaging/messaging-spans.md)
```
trask, line 17: just confirming that this version choice is intentional (and not a copy-paste from db semconv which happened to use the same version)

docs/messaging/README.md
```diff
@@ -11,6 +11,34 @@ path_base_for_github_subdir:
...
+> (or prior):
+>
+> * SHOULD NOT change the version of the messaging conventions that they emit by default
+>   until the messaging semantic conventions are marked stable.
+>   Conventions include, but are not limited to, attributes,
+>   metric and span names, span kind and unit of measure.
+> * SHOULD introduce an environment variable `OTEL_SEMCONV_STABILITY_OPT_IN`
+>   in the existing major version which is a comma-separated list of values.
+>   If the list of values includes:
```
trask, line 26: looks like a typo in db also, I can send PR to fix that
```suggestion
>   The list of values includes:
```

trask (approved):

docs/messaging/README.md
joaopgrassi: Yes it was a copy from db. I'm not sure to be honest what criteria we use to define the version here. Is it by the amount of instrumentations using x version? Or is it the "last version" before major changes happened? Because the latter is waaaay back so I'm not sure 🤔 

I saw the Java instrumentation most likely is using 1.25.

docs/messaging/README.md
joaopgrassi: Ah good catch. I fixed in all files here.

arminru (approved):

docs/messaging/README.md
trask: one criteria we could use is the current version in place when the initial placeholder was added, which looks like v1.24 (#734)

what were the breaking changes between v1.24 and v1.25 that Java instrumentation implemented (probably incorrectly)?

trask: > I think keeping as 1.24 (when it was added) is the best we can do

sounds good to me, thanks

PR #1531, opened 2024-10-30
RabbitMQ: Specify how to populate `messaging.destination.name`

### LINKED ISSUE #1529: [Messaging RabbitMQ] What to use for `destination.name` when the default exchange is used (empty string)
### Area(s)

area:messaging

In the RabbitMQ conventions we say: 

> In RabbitMQ, the destination is defined by an exchange and a routing key. messaging.destination.name MUST be set to the name of the exchange. This will be an empty string if the default exchange is used.

In Java, the RabbitMQ instrumentation uses `<default>` in case the exchange name is empty (default one is used). 

It is also "confusing" that we say a destination is composed of exchange and routing key, but we don't say how or if even the routing key is supposed to be part of the destination name attribute. 

Given there are clients for many languages, the conventions should specify what to use when the default exchange is used. Right now, each language can do their own thing, resulting in inconsistent behavior/value for the `destination.name` attribute.

_No response_

### Additional context

Java example:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/RabbitDeliveryAttributesGetter.java#L41

joaopgrassi:
Fixes #1529

## Changes

Specify in more detail how to populate `messaging.destination.name` for RabbitMQ. I based it on the [Default exchange](https://www.rabbitmq.com/tutorials/amqp-concepts#exchange-default) docs. My understanding is that a `routing key` is always present, even when using the default exchange (empty in this case). The routing key then is the queue name. 

For all other cases, I left it as `amq.default` which is what RabbitMQ uses. I saw reference for it for example here: https://www.rabbitmq.com/docs/access-control

> For convenience RabbitMQ maps AMQP 0-9-1's default exchange's blank name to 'amq.default' when performing permission checks.


docs/messaging/rabbitmq.md
```diff
@@ -12,8 +12,15 @@ The Semantic Conventions for [RabbitMQ](https://www.rabbitmq.com/) extend and ov
...
+### Destination name
+
 In RabbitMQ, the destination is defined by an *exchange* and a *routing key*.
-`messaging.destination.name` MUST be set to the name of the exchange. This will be an empty string if the default exchange is used.
+`messaging.destination.name` MUST be set according to the following rules:
+
+- Both *exchange* and *routing key* present and non-empty: `{exchange}:{routing key}`
+- Only *exchange* present and non-empty: `{exchange}`
+- Only *routing key* present and non-empty: `{routing key}`
+- Otherwise: `amp.default`
```
lmolkova, line 23: typo?

```suggestion
- Otherwise: `amq.default`
```

also, do we need to set it? maybe just keep it empty if nothing was provided?

docs/messaging/rabbitmq.md
```diff
@@ -12,8 +12,15 @@ The Semantic Conventions for [RabbitMQ](https://www.rabbitmq.com/) extend and ov
 
 ## RabbitMQ attributes
 
+### Destination name
+
 In RabbitMQ, the destination is defined by an *exchange* and a *routing key*.
-`messaging.destination.name` MUST be set to the name of the exchange. This will be an empty string if the default exchange is used.
+`messaging.destination.name` MUST be set according to the following rules:
+
+- Both *exchange* and *routing key* present and non-empty: `{exchange}:{routing key}`
```
lmolkova, line 20: I think we have a problem on the consumer (even before this PR). 

image

If someone publishes a message to exchange with a routing key and then there are bindings that route messages with this key to a queue (or queues) with a different name, then we have 3 things on the consumer side:

- we still have access to exchange and routing key - https://www.rabbitmq.com/docs/consumers#message-properties
- plus we have a queue name from which we got the message

and we have just two attributes (`mesasging.destination.name` and `messaging.rabbitmq.routing_key`) to store them

joaopgrassi: Yes, I also stumbled on the same thing, where does the queue fits into these two attributes.  To make matters worse, for example in the Java instrumentation the [Queue name is what is used for the Span name](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/ReceiveRequest.java#L28), not the exchange nor the routing key.

docs/messaging/rabbitmq.md
joaopgrassi: Yep, it's a typo. I'm not sure of leaving it empty, I saw that in the RabbitMQ management UI it also shows `amq.default` so maybe it's nice to also have that very explicitly as part of the destination/span name.

docs/messaging/rabbitmq.md
joaopgrassi: I pushed a change, let me know what you think. [b02ff2a](https://github.com/open-telemetry/semantic-conventions/pull/1531/commits/b02ff2ae3bb8d7054e82ab7504167a2d35801204)

docs/messaging/rabbitmq.md
```diff
@@ -14,13 +14,19 @@ The Semantic Conventions for [RabbitMQ](https://www.rabbitmq.com/) extend and ov
...
+for consumers, a *queue*.
+`messaging.destination.name` MUST be set to:
 
-- Both *exchange* and *routing key* present and non-empty: `{exchange}:{routing key}`
-- Only *exchange* present and non-empty: `{exchange}`
-- Only *routing key* present and non-empty: `{routing key}`
-- Otherwise: `amp.default`
+**Producers**
+
+- `{exchange}:{routing key}` when both values are present and non-empty.
```
lmolkova, line 51: let's consider `.` to match what we do in DBs

docs/messaging/rabbitmq.md
```diff
@@ -14,13 +14,19 @@ The Semantic Conventions for [RabbitMQ](https://www.rabbitmq.com/) extend and ov
...
+If any has an empty value (e.g., the default exchange is used) it SHOULD be omitted.
+- Otherwise: `amq.default`
+
+**Consumers**
+- `{exchange}:{routing key}:{queue}` when all values are present and non-empty.
```
lmolkova, line 57: maybe check if routing key and queue are the same and populate just one

joaopgrassi: Yeah I looked and it seems when receiving/delivering, only the routing key is present https://www.rabbitmq.com/docs/consumers#message-properties. Also, if I understand it right from some research, the routing key can be then also the same as queue name https://stackoverflow.com/questions/22655735/retrieve-name-of-rabbitmq-queue-where-message-was-consumed-from

So I think we should be good to keep only the exchange and routing key. PTAL

lmolkova: wait, the queue is available on the client level -https://www.rabbitmq.com/client-libraries/java-api-guide#consuming

```java
boolean autoAck = false;
channel.basicConsume(queueName, autoAck, "myConsumerTag",
     new DefaultConsumer(channel) {
         @Override
         public void handleDelivery(String consumerTag,
                                    Envelope envelope,
                                    AMQP.BasicProperties properties,
                                    byte[] body)
             throws IOException
         {
             String routingKey = envelope.getRoutingKey();
             String contentType = properties.getContentType();
             long deliveryTag = envelope.getDeliveryTag();
             <i>// (process the message components here ...)</i>
             channel.basicAck(deliveryTag, false);
         }
     });
```

and most of the instrumentations should be able to populate it. We need to capture it somewhere

joaopgrassi: Yes, I checked and it's correct. I added it back, only now stating when both routing key and queue name are the same, only one should be used.

docs/messaging/rabbitmq.md
```diff
@@ -40,8 +40,26 @@ The Semantic Conventions for [RabbitMQ](https://www.rabbitmq.com/) extend and ov
...
 ## RabbitMQ attributes
 
-In RabbitMQ, the destination is defined by an *exchange* and a *routing key*.
-`messaging.destination.name` MUST be set to the name of the exchange. This will be an empty string if the default exchange is used.
+### Destination name
+
+n RabbitMQ, the destination is defined by an *exchange*, a *routing key* and
+for consumers, a *queue*.
+
+`messaging.destination.name` MUST be set to:
```
lmolkova, line 48: given how nuanced it is, maybe we should relax to `SHOULD`?

docs/messaging/rabbitmq.md
```diff
@@ -40,8 +40,26 @@ The Semantic Conventions for [RabbitMQ](https://www.rabbitmq.com/) extend and ov
...
+
+#### Producers
+
+- `{exchange}:{routing key}` when both values are present and non-empty.
+If any has an empty value (e.g., the default exchange is used) it SHOULD be omitted.
+- Otherwise: `amq.default`
```
lmolkova, line 54: nit:

```suggestion
- `amq.default` when default exchange is used and no routing key is provided
```

(just to make it a bit easier to read)

lmolkova (approved):
LGTM with some small suggestions

trask (commented):
should this text instead be part of rabbitmq yaml for `messaging.destination.name`?

joaopgrassi:
> should this text instead be part of rabbitmq yaml for `messaging.destination.name`?

Good point, I didn't consider it but maybe it makes sense 🤔

joaopgrassi:
@trask I moved the text to the yaml file instead in [f48427f](https://github.com/open-telemetry/semantic-conventions/pull/1531/commits/f48427fa52ee06c5fa8f5b957c128fe4abc20e4b). I think it makes more sense. 

@lmolkova what do you think?

model/messaging/spans.yaml
```diff
@@ -118,6 +118,22 @@ groups:
...
+          - **Producers**: `{exchange}:{routing key}` when both values are present and non-empty.
+          Otherwise: `amq.default` when the default exchange is used and no routing key is provided
+
+          - **Consumers**: `{exchange}:{routing key}:{queue}` when all values are present and non-empty.
+          If any has an empty value (e.g., the default exchange is used) it SHOULD be omitted.
+          For cases when `{routing key}` and `{queue}` are equal, only one of them SHOULD
+          be used, e.g., `{exchange}:{routing key}`.
+        examples: ['direct_logs:warning', 'logs']
+        requirement_level:
+          conditionally_required: If span describes operation on a single message or if the value applies to all messages in the batch.
```
lmolkova, line 136: Does rabbit support batching at all? I don't think it does. Wonder if we can just say that it's `required`  (or `conditionally_required: if available`)

joaopgrassi: Yes, true, there's pre-fetch but we explicitly say to not create receive/process span for those so I think we can remove this from Rabbit. 

I'm leaning towards just marking it as required, as I don't see a case when it's not available.. _something_ is always available, (like even when using the default exchange, or anonymous queues). [bde99d0](https://github.com/open-telemetry/semantic-conventions/pull/1531/commits/bde99d03a24315db9777f533fe7d0832accb53d5)

lmolkova (approved):
Thank you!

model/messaging/spans.yaml
```diff
@@ -118,6 +118,21 @@ groups:
           Message [correlation Id](https://www.rabbitmq.com/tutorials/tutorial-six-java#correlation-id) property.
       - ref: messaging.message.body.size
         requirement_level: opt_in
+      - ref: messaging.destination.name
+        note: |
+          In RabbitMQ, the destination is defined by an *exchange*, a *routing key* and for consumers, a *queue*.
+
+          `messaging.destination.name` SHOULD be set to:
+
+          - **Producers**: `{exchange}:{routing key}` when both values are present and non-empty.
```
trask, line 127:
```suggestion
          - **On the producer side**: `{exchange}:{routing key}` when both values are present and non-empty.
```

model/messaging/spans.yaml
```diff
@@ -118,6 +118,21 @@ groups:
...
+          Otherwise: `amq.default` when the default exchange is used and no routing key is provided
```
trask, line 128: can there be an `{exchange}` but not a `{routing key}`, or can there be a `{routing key}` but not a `{exchange}`? it may be helpful to be explicit about fallback options (if any), in particular, do you drop the `:` or not when only one of them is present, here's an example where we cover these fallbacks: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#name

model/messaging/spans.yaml
```diff
@@ -118,6 +118,21 @@ groups:
...
+
+          - **Consumers**: `{exchange}:{routing key}:{queue}` when all values are present and non-empty.
```
trask, line 130:
```suggestion
          - **On the consumer side**: `{exchange}:{routing key}:{queue}` when all values are present and non-empty.
```

model/messaging/spans.yaml
joaopgrassi: Yeah they can be empty like that AFAIK. I adapted it a bit, hopefully it is clearer now.

carlosalberto:
Overall LGTM but I see that in the consumer side, you don't ever use `amq.default` _ever_ - is it because it's not even possible to use the default exchange _without_ a routing key?

carlosalberto (approved):

joaopgrassi:
> Overall LGTM but I see that in the consumer side, you don't ever use `amq.default` _ever_ - is it because it's not even possible to use the default exchange _without_ a routing key?

Yes, AFAIK for consumers there's always a queue, which also can be the same as the routing key in the 'default exchange' case. https://www.rabbitmq.com/tutorials/amqp-concepts#consumers

model/messaging/spans.yaml
```diff
@@ -118,6 +118,22 @@ groups:
...
         requirement_level: opt_in
+      - ref: messaging.destination.name
+        note: |
+          In RabbitMQ, the destination is defined by an *exchange*, a *routing key* and for consumers, a *queue*.
+
+          `messaging.destination.name` SHOULD be set to:
+
+          - **On the producer side**: `{exchange}:{routing key}` when both values are present and non-empty.
+          When only one is available, only that value SHOULD be used. E.g., `{exchange}` or `{routing key}`.
+          Otherwise: `amq.default` when the default exchange is used and no routing key is provided
```
trask, line 129: what does the `amq` part signify? thanks

joaopgrassi: It's how RabbitMQ "calls" the default (empty) exchange name:

> For convenience RabbitMQ maps AMQP 0-9-1's default exchange's blank name to 'amq.default' when performing permission checks.

https://www.rabbitmq.com/docs/access-control#authorisation
https://www.rabbitmq.com/tutorials/amqp-concepts#exchange-default
https://github.com/search?q=repo%3Arabbitmq%2Frabbitmq-server%20amq.default&type=code

trask (approved):

Issue #534, opened 2023-11-16
[http] Publish migration plan for HTTP semconv changes.


TylerHelmuth:
I was looking through https://github.com/open-telemetry/semantic-conventions/tree/main/docs/http#semantic-conventions-for-http and I am worried that users looking to understand how the breaking changes of the semantic convention will not have a clear path to understanding. I don't spend a lot of time in the spec/semconv, but I spend a lot of time in the project and I feel confused about certain things - I am worried about users who are purely consumers of OTEL.

My current questions:
- Is the warning in https://github.com/open-telemetry/semantic-conventions/tree/main/docs/http#semantic-conventions-for-http still accurate?
- Is the Oct 1st 2023 deadline still applicable (meaning Instrumentation Libraries are free to drop the env var support with the next stable release?

Ultimately I would love to see a page dedicated to outlining the migration steps. As a user I want to see exactly what values changed, what was dropped, what was added, etc. I'd also want to see which languages support the new values, and any important deadlines.

I feel the [blog post](https://opentelemetry.io/blog/2023/http-conventions-declared-stable/) was really good and I know a lot of details went into the changelog, but both of those locations aren't great for discoverability.  Important details in release notes, blog posts, and changelogs get moved downward as time goes on, making them harder to find.  I really want a permanent, dedicated, discoverable space for us to communicate this changes to our end users.

Hopefully any patterns we do now can help with any future semconv breaking changes.

TylerHelmuth:
> yes (was there something that led you to think it isn't accurate, so we can fix that?)

My general confusion of the doc is that as a reader I'm not sure if it is applicable to me.  It gives a warning about using `v1.20.0` or prior, but to be able to answer that I have to know what my instrumentation is using.  That's all possible to research, but it isn't something I expect the average OTel consumers to know. Ultimately it feels like a message for the SDK maintainers, and not something for an end user to consume (which is fine). But if an end user finds it I think it creates confusion. As an end user all I want to see is how the SDK I'm using is handling the breaking change.

https://opentelemetry.io/blog/2023/http-conventions-declared-stable/#summary-of-changes is what I'm looking for, but in a more permanent/discoverable place.  My worry is that down the road when some enterprise bumps their java/dotnet instrumentation from 1.0 to 2.0 they need to be able to find these changes or they'll feel pain. Of course we'd hope that they'd take advantage of the env var while in 1.0, but who knows.

[...]

PR #1533, opened 2024-10-30
[chore] Mention JavaAgent v2.0.0 default behaviour

devurandom:
## Changes

HTTP semantic convention stability migration docs [^1] do not mention that the JavaAgent v2.0.0 already defaults to the new, stable behaviour  [^2], which could lead to people believing they have to set `OTEL_SEMCONV_STABILITY_OPT_IN=http` when they don't.

[^1]: https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/
[^2]: https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.0.0

This does not change an actual convention, thus I believe it does not need a changelog entry.

### DIFF
--- a/docs/non-normative/http-migration.md
+++ b/docs/non-normative/http-migration.md
@@ -29,7 +29,7 @@ updated to the stable HTTP semantic conventions, they:
 - May drop the environment variable in their next major version and emit only
   the stable HTTP and networking conventions.
 
-The OpenTelemetry Java Agent v2.0.0 and later already defaults to the stable
+The OpenTelemetry Java Agent [v2.0.0](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.0.0) and later already defaults to the stable
 conventions -- you do not have to opt into them.
 
 ## Summary of changes
--- a/docs/non-normative/http-migration.md
+++ b/docs/non-normative/http-migration.md
@@ -29,6 +29,9 @@ updated to the stable HTTP semantic conventions, they:
 - May drop the environment variable in their next major version and emit only
   the stable HTTP and networking conventions.
 
+The OpenTelemetry Java Agent v2.0.0 and later already defaults to the stable
+conventions -- you do not have to opt into them.
+
 ## Summary of changes
 
 This section summarizes the changes made to the HTTP semantic conventions


docs/non-normative/http-migration.md
```diff
@@ -29,6 +29,9 @@ updated to the stable HTTP semantic conventions, they:
 - May drop the environment variable in their next major version and emit only
   the stable HTTP and networking conventions.
 
+The OpenTelemetry Java Agent v2.0.0 and later already defaults to the stable
```
arminru, line 32: perhaps refer to
```suggestion
The OpenTelemetry Java Agent [v2.0.0](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.0.0) and later already defaults to the stable
```
for clarity

arminru:
@open-telemetry/java-instrumentation-maintainers FYI

trask:
@devurandom a related discussion has been taking place in #534, can you review that issue? thanks

devurandom:
> a related discussion has been taking place in #534, can you review that issue?

Thanks for the hint.  I was not aware of that issue.

My PR is (to a limited degree) addressing his point in https://github.com/open-telemetry/semantic-conventions/issues/534#issuecomment-1816832645: I point out that, if you're using Java agent v2.0.0 or later, the migration does not apply to you.

I do not feel qualified to address the wider-scope suggestions he points out, like https://github.com/open-telemetry/semantic-conventions/issues/534#issuecomment-1834604634, that fully address the issue and could close it.

trask:
> I point out that, if you're using Java agent v2.0.0 or later, the migration does not apply to you.

if you are upgrading from Java agent 1.x to 2.x, I think it does apply to you

docs/non-normative/http-migration.md
```diff
@@ -29,6 +29,9 @@ updated to the stable HTTP semantic conventions, they:
 - May drop the environment variable in their next major version and emit only
   the stable HTTP and networking conventions.
 
+The OpenTelemetry Java Agent [v2.0.0](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.0.0) and later already defaults to the stable
```
lmolkova, line 32: thanks for the clarification!

should this text belong in the javaagent docs? I believe this doc is intended for instrumentation authors and telemetry backends rather than the end-users of specific instrumentation. 

I don't mind having it here either, but then  I think we need a section for instrumentation-specific behavior that would initially contain Java instrumentation agent.

Curious what @trask thinks.

trask: > should this text belong in the javaagent docs?

I think that would probably be a better place, until we solve the more general questions in #534

docs/non-normative/http-migration.md
```diff
@@ -29,6 +29,9 @@ updated to the stable HTTP semantic conventions, they:
 - May drop the environment variable in their next major version and emit only
   the stable HTTP and networking conventions.
```
trask, line 30: I wonder if this whole section above is adding to the confusion. It was a good part of the blog post, but I wonder if the document here should just focus on the summary of changes?

PR #1535, opened 2024-10-30
Make table generation and check targets strict and fix violations

lmolkova:
Mitigates https://github.com/open-telemetry/weaver/issues/434:

- `make table-generation` and `make table-check` are now strict 
- fixes recently introduced violation that didn't raise an issue (but prevented table generation from completing its job)

### DIFF
--- a/model/process/deprecated/registry-deprecated.yaml
+++ b/model/process/deprecated/registry-deprecated.yaml
@@ -25,3 +25,4 @@ groups:
         deprecated: 'Replaced by `process.executable.build_id.htlhash`'
         brief: >
           "Deprecated, use `process.executable.build_id.htlhash` instead."
+        examples: ['600DCAFE4A110000F2BF38C493F5FB92']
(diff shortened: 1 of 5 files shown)


joaopgrassi (approved):

model/process/deprecated/registry-deprecated.yaml
```diff
@@ -25,3 +25,4 @@ groups:
         deprecated: 'Replaced by `process.executable.build_id.htlhash`'
         brief: >
           "Deprecated, use `process.executable.build_id.htlhash` instead."
+        examples: ['600DCAFE4A110000F2BF38C493F5FB92']
```
trask, line 28: are examples required now?

trask (approved):

model/process/deprecated/registry-deprecated.yaml
lmolkova: they are (and have been required) except for enums

Issue #1495, opened 2024-10-21
Cosmos DB:  Network level Metrics


sourabh1007:
### Area(s)

area:db

### Is your change request related to a problem? Please describe.

In the Cosmos DB SDK, a single operation involves several network calls. Currently, if something goes wrong (e.g., high latency) with these network calls, customers rely solely on the logs they have implemented in their applications. When investigating such issues, we are dependent on the information provided by the customer and backend telemetry. To improve monitoring and make it more aligned with potential errors, I am proposing a set of metrics that the SDK should collect to enhance observability.

### Describe the solution you'd like

Proposing below list of metrics for network calls (only TCP, for Http calls we are dependent on default metrics list provided by http):

**Dimensions**
Tag/dimension name | Sample value
-- | --
db.system | cosmodb
db.collection.name | myCollectionName
db.namespace | myDatabaseName
server.address | myaccountname.documents.azure.com
server.port | 443
db.operation.name | query_items
db.response.status_code | 200 or 429 etc.
db.cosmos.sub_status_code | 1002 etc.
db.cosmosdb.consistency_level | Eventual, ConsistentPrefix, BoundedStaleness, Strong or Session
network.protocol.name | `rntbd`
tcp.response.status_code | 200 or 429 etc.
tcp.server.address | host  from `rntbd://<host> : <port>`
tcp.server.port | port from `rntbd://<host>:<port>`
db.cosmosdb.tcp.response.sub_status_code | 1002 etc.
db.cosmosdb.tcp.request.region | region name, where request was sent
db.cosmosdb.tcp.server.partitionId | cosmosdb partititon id
db.cosmosdb.tcp.server.replicaid | cosmosdb replica id

**Metrics**
Name | Unit | Type | Description
-- | -- | -- | --
tcp.client.request.count| `{requests}` | Histogram | Number of TCP requests made
tcp.client.request.duration | `{seconds}` | Histogram | Duration of TCP client requests.
tcp.client.active_requests | `{request}` | UpDownCounter | Number of active TCP client requests (inflightrequest information we can put here)
tcp.client.request.body.size | `By` | Histogram | Size of TCP client request bodies.
tcp.client.response.body.size | `By` | Histogram | Size of TCP client response bodies.
tcp.client.connection.duration | `s` | Histogram | The duration of the successfully established outbound TCP connections. i.e. Channel Aquisition Time
tcp.server.request.duration | `{seconds}` | Histogram | Backend Latency
[...]

PR #1543, opened 2024-11-01
Cosmos DB: Network Level Metrics

sourabh1007:
## Changes

Ref. https://github.com/open-telemetry/semantic-conventions/issues/1495


docs/attributes-registry/db.md
```diff
@@ -216,6 +217,19 @@ This group defines attributes for Azure Cosmos DB.
...
 
+## Azure Cosmos DB Network level Attributes
+
+This group defines attributes for Azure Cosmos DB Network call.
+
+| Attribute | Type | Description | Examples | Stability |
+|---|---|---|---|---|
+| <a id="db-cosmosdb-network-host" href="#db-cosmosdb-network-host">`db.cosmosdb.network.host`</a> | string | Host from Gateway or Backend endpoint address | `rntbd://cdb-east-us.azure.document.com` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="db-cosmosdb-network-port" href="#db-cosmosdb-network-port">`db.cosmosdb.network.port`</a> | int | Port from Gateway or Backend endpoint address | `443` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="db-cosmosdb-network-region" href="#db-cosmosdb-network-region">`db.cosmosdb.network.region`</a> | string | Azure Cosmos DB Region where the network call was made | `WEST US 2` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 228: Please add explanation whether it is the source or target region. Also not sure whether we would need both. For rntbd host when cx uses private endpoint the host would never indicate the target region.

docs/attributes-registry/db.md
```diff
@@ -216,6 +217,19 @@ This group defines attributes for Azure Cosmos DB.
...
+| <a id="db-cosmosdb-network-resource-id" href="#db-cosmosdb-network-resource-id">`db.cosmosdb.network.resource_id`</a> | string | Azure Cosmos DB Partition Key Range Id for a Gateway call and PartitionId/RepilcaId for a Direct call | `45678765678987/2345678906789s` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 229: typo in ReplicaId

docs/database/cosmosdb.md
```diff
@@ -440,5 +448,743 @@ It captures the number of active instances at any given time. Best practices dic
...
+<!-- semconv metric.db.client.cosmosdb.request.duration -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `db.client.cosmosdb.request.duration` | Histogram | `s` | End to end Latency of Azure Cosmos DB Backend network call | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 472: unit shoudl be ms not s

FabianMeiswinkel: Backend network call - IMO Backend is confusing here - maybe "service network call"

docs/database/cosmosdb.md
```diff
@@ -440,5 +448,743 @@ It captures the number of active instances at any given time. Best practices dic
...
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`db.collection.name`](/docs/attributes-registry/db.md) | string | Cosmos DB container name. [1] | `public.users`; `customers` | `Conditionally Required` If available | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 476: Please be specific whether it is CollectionName or RID. Also Comsos DB does not use term collection but Container for NoSQL Accounts. I assume you do this to be consistent with OpenTelemetry - but given that this is in the COsmos namespace shouldn't you use consistent terminology with Cosmos DB?

docs/database/cosmosdb.md
```diff
@@ -440,5 +448,743 @@ It captures the number of active instances at any given time. Best practices dic
...
+| [`db.cosmosdb.consistency_level`](/docs/attributes-registry/db.md) | string | Account or request [consistency level](https://learn.microsoft.com/azure/cosmos-db/consistency-levels). | `Eventual`; `ConsistentPrefix`; `BoundedStaleness`; `Strong`; `Session` | `Conditionally Required` If available. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 477: Please be specific what this value is - is it null unless specified on request or is it the effective consistency level (request specified, clinet specified, account default) whatever comes first?

docs/database/cosmosdb.md
```diff
@@ -440,5 +448,743 @@ It captures the number of active instances at any given time. Best practices dic
...
+| [`db.cosmosdb.network.region`](/docs/attributes-registry/db.md) | string | Azure Cosmos DB Region where the network call was made | `WEST US 2` | `Conditionally Required` If available | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.cosmosdb.network.response.status_code`](/docs/attributes-registry/db.md) | int | HTTP Status code returned for network call | `200` | `Conditionally Required` If available | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.cosmosdb.network.response.sub_status_code`](/docs/attributes-registry/db.md) | int | Azure Cosmos DB Sub Status code returned for network call | `1002` | `Conditionally Required` If available | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.cosmosdb.sub_status_code`](/docs/attributes-registry/db.md) | int | Cosmos DB sub status code. | `1000`; `1002` | `Conditionally Required` when response was received and contained sub-code. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 483: Probably needs an explanation why sub_status_code exists in two namespaces - I get it - but I imagine most customer's won't unless explained

docs/database/cosmosdb.md
```diff
@@ -440,5 +448,743 @@ It captures the number of active instances at any given time. Best practices dic
...
+| [`db.namespace`](/docs/attributes-registry/db.md) | string | The name of the database, fully qualified within the server address and port. | `customers`; `test.users` | `Conditionally Required` If available. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 484: I do not get the comment about server address and port?

docs/database/cosmosdb.md
```diff
@@ -440,5 +448,743 @@ It captures the number of active instances at any given time. Best practices dic
...
+| [`db.operation.name`](/docs/attributes-registry/db.md) | string | The name of the operation or command being executed. [2] | `findAndModify`; `HMSET`; `SELECT` | `Conditionally Required` [3] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 485: Please use samples that exist in Cosmos DB

docs/database/cosmosdb.md
```diff
@@ -440,5 +448,743 @@ It captures the number of active instances at any given time. Best practices dic
...
+| [`db.response.status_code`](/docs/attributes-registry/db.md) | string | Database response status code. [4] | `102`; `ORA-17002`; `08P01`; `404` | `Conditionally Required` [5] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 486: Please use samples that exist in Cosmos DB

docs/database/cosmosdb.md
```diff
@@ -440,5 +448,743 @@ It captures the number of active instances at any given time. Best practices dic
...
+| [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [6] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` If and only if the operation failed. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
FabianMeiswinkel, line 487: I am a bit confused - pretty much all exceptions thrown form the SDK to the application would be a CosmosException. Maybe one of teh sub-classes in Java. The sample here should never be thrown to applicatin - unless the intent is to use not the exception thrown to app but the inner most inner exception? Which is the desired exception to be used?

docs/database/cosmosdb.md
```diff
@@ -440,5 +448,743 @@ It captures the number of active instances at any given time. Best practices dic
...
+| [`network.protocol.name`](/docs/attributes-registry/network.md) | string | [OSI application layer](https://osi-model.com/application-layer/) or non-OSI equivalent. [7] | `amqp`; `http`; `mqtt` | `Conditionally Required` If available | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`server.port`](/docs/attributes-registry/server.md) | int | Server port number. [8] | `80`; `8080`; `443` | `Conditionally Required` [9] | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`server.address`](/docs/attributes-registry/server.md) | string | Name of the database host. [10] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`db.cosmosdb.network.resource_id`](/docs/attributes-registry/db.md) | string | Azure Cosmos DB Partition Key Range Id for a Gateway call and PartitionId/RepilcaId for a Direct call | `45678765678987/2345678906789s` | `Opt-In` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 491: Typo in ReplicaId

FabianMeiswinkel (changes requested):
Left few comments

docs/database/cosmosdb.md
trask: check out https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#instrument-units:

> When instruments are measuring durations, seconds (i.e. `s`) SHOULD be used.

docs/database/cosmosdb.md
sourabh1007: Idea here is to keep things consistent with otel conventions. Description about the attribute would tell us what does it really mean, in terms of cosmosdb.

docs/database/cosmosdb.md
sourabh1007: I tried to rewrite it.

docs/database/cosmosdb.md
sourabh1007: in our case, account gateway endpoint is `server.address` and `443` is port. So `db.namespace` represents the database for that account, mentioned in these attributes.

docs/database/cosmosdb.md
sourabh1007: Put the appropriate examples.

model/database/registry.yaml
```diff
@@ -541,9 +542,50 @@ groups:
...
+        type: int
+        stability: experimental
+        brief: The port used for the Gateway or Backend Service endpoint.
+        examples: 443
+      - id: db.cosmosdb.network.response.status_code
+        type: int
+        stability: experimental
+        brief: The HTTP status code returned from the network request.
+        examples: 200
+      - id: db.cosmosdb.network.response.sub_status_code
```
lmolkova, line 572: is there any difference between `db.cosmosdb.network.response.sub_status_code` and `db.cosmosdb.sub_status_code` ?  Can we reuse the existing one instead of introducing a new thing?

model/database/registry.yaml
```diff
@@ -541,9 +542,50 @@ groups:
...
+        examples: ["rntbd://cdb-east-us.azure.document.com"]
+      - id: db.cosmosdb.network.port
+        type: int
+        stability: experimental
+        brief: The port used for the Gateway or Backend Service endpoint.
+        examples: 443
+      - id: db.cosmosdb.network.response.status_code
+        type: int
+        stability: experimental
+        brief: The HTTP status code returned from the network request.
```
lmolkova, line 570: I think HTTP is misleading here. If I understand correctly, the protocol happens to reuse HTTP status codes, but it's not necessarily HTTP, right?

Consider something like:

```suggestion
        brief: The status code returned in response.
        note: The protocol reuses HTTP status codes.
```

model/database/registry.yaml
```diff
@@ -541,9 +542,50 @@ groups:
...
+        type: string
+        stability: experimental
+        brief: Specifies the host address from the Gateway or Backend Service endpoint.
+        examples: ["rntbd://cdb-east-us.azure.document.com"]
+      - id: db.cosmosdb.network.port
+        type: int
+        stability: experimental
+        brief: The port used for the Gateway or Backend Service endpoint.
+        examples: 443
+      - id: db.cosmosdb.network.response.status_code
```
lmolkova, line 567: Do we need to introduce new attribute for this? And should it be in `db.cosmosdb.network` ?

If it belongs in DB domain, let's reuse `db.response.status_code` instead.
If it's a protocol-specific thing, let's move it to the `rntbd.*` namespace

model/database/registry.yaml
```diff
@@ -541,9 +542,50 @@ groups:
...
+        type: int
+        stability: experimental
+        brief: The Azure Cosmos DB sub-status code for a network request made during an operation.
+        examples: 1002
+      - id: db.cosmosdb.network.region
+        type: string
+        stability: experimental
+        brief: Specifies the Azure Cosmos DB target region for the network request.
+        examples: "WEST US 2"
+      - id: db.cosmosdb.network.resource_id
```
lmolkova, line 582: resource_id could be confused with Azure resource id, could we find some other name that would be descriptive?

Also `Azure Cosmos DB Partition Key Range ID for a Gateway call, or Partition ID/Replica ID for a Direct call` does not sound like something network-specific - why put it into `db.cosmosdb.network` namespace?

model/database/cosmosdb-metrics.yaml
```diff
@@ -26,3 +26,68 @@ groups:
...
+    brief: "End to end Latency of Azure Cosmos DB Backend Service network call"
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    extends: attributes.db.cosmosdb.network.minimal
+
+  - id: metric.db.client.cosmosdb.request.body.size
+    type: metric
+    metric_name: db.client.cosmosdb.request.body.size
+    brief: "Size of request object sent to Azure Cosmos DB backend"
```
lmolkova, line 42: could we be more specific? What does request object mean? Does it include headers/metadata? Is it the size transported over the wire (i.e. compressed or not compressed depending on the circumstances)?

docs/database/cosmosdb.md
```diff
@@ -440,5 +448,743 @@ It captures the number of active instances at any given time. Best practices dic
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Network Level Metrics
+
+### Metric: `db.client.cosmosdb.request.duration`
```
lmolkova, line 453: would you report network-level metrics for both direct and gateway mode?

model/database/cosmosdb-metrics.yaml
```diff
@@ -26,3 +26,68 @@ groups:
       - ref: server.port
         requirement_level:
           conditionally_required: If using a port other than the default port for this DBMS and if `server.address` is set.
+
+  - id: metric.db.client.cosmosdb.request.duration
+    type: metric
+    metric_name: db.client.cosmosdb.request.duration
+    brief: "End to end Latency of Azure Cosmos DB Backend Service network call"
```
lmolkova, line 33: I assume we measure the latency on the client, but the brief gives an impression it's measured on the server

```suggestion
    brief: "End to end latency of a network call made to Azure Cosmos DB service"
```

model/database/cosmosdb-metrics.yaml
```diff
@@ -26,3 +26,68 @@ groups:
...
+    brief: "Time taken in creating connection, It is only applicable in Direct Mode."
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    extends: attributes.db.cosmosdb.network.minimal
+
+  - id: metric.db.server.cosmosdb.request.duration
+    type: metric
+    metric_name: db.server.cosmosdb.request.duration
+    brief: "Backend Server Latency, i.e. time taken by Azure Cosmos DB server to process the request. It is only applicable in Direct Mode."
```
lmolkova, line 69: See my other comment - we're measuring the client-side latency, not the service side (they are different)

model/database/cosmosdb-metrics.yaml
```diff
@@ -26,3 +26,68 @@ groups:
...
+    metric_name: db.client.cosmosdb.request.channel_aquisition.duration
+    brief: "Time taken in creating connection, It is only applicable in Direct Mode."
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    extends: attributes.db.cosmosdb.network.minimal
+
+  - id: metric.db.server.cosmosdb.request.duration
+    type: metric
+    metric_name: db.server.cosmosdb.request.duration
```
lmolkova, line 68: Are we going to report it on the client side?

model/database/cosmosdb-metrics.yaml
```diff
@@ -26,3 +26,68 @@ groups:
...
+    brief: "Backend Server Latency, i.e. time taken by Azure Cosmos DB server to process the request. It is only applicable in Direct Mode."
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    extends: attributes.db.cosmosdb.network.minimal
+
+  - id: metric.db.client.cosmosdb.request.transit.duration
+    type: metric
+    metric_name: db.client.cosmosdb.request.transit.duration
```
lmolkova, line 77: How is this different from `db.client.cosmosdb.request.duration`?

model/database/cosmosdb-metrics.yaml
```diff
@@ -26,3 +26,68 @@ groups:
...
+    brief: >
+      "Transit time refers to the duration taken by a request to travel across the network—from the moment it leaves the client to
+      when the response is received back at the client. It is only applicable in Direct Mode."
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    extends: attributes.db.cosmosdb.network.minimal
+
+  - id: metric.db.client.cosmosdb.request.received.duration
```
lmolkova, line 86: you can get time-in-SDK by subtracting `db.client.operation.duration` from `metric.db.client.cosmosdb.request.duration`. Why is this metric necessary? 

Also, have you tried implementing it? It'd be really hard to know when request is "fully processed by the SDK" given layers of convenience APIs, paging and other things.

model/database/common.yaml
```diff
@@ -60,6 +64,35 @@ groups:
       - ref: db.cosmosdb.consistency_level
         requirement_level:
           conditionally_required: If available.
+      - ref: error.type
+        examples: ["CosmosException", "CosmosOperationCanceledException", "CosmosObjectDisposedException", "CosmosNullReferenceException"]
```
lmolkova, line 68: it should be fully qualified exception name `Microsoft.Azure.Cosmos.CosmosOperationCanceledException`

model/database/cosmosdb-metrics.yaml
```diff
@@ -26,3 +26,68 @@ groups:
...
+  - id: metric.db.client.cosmosdb.response.body.size
+    type: metric
+    metric_name: db.client.cosmosdb.response.body.size
+    brief: "Size of response received from Azure Cosmos DB backend"
+    instrument: histogram
+    unit: "{bytes}"
+    stability: experimental
+    extends: attributes.db.cosmosdb.network.minimal
+
+  - id: metric.db.client.cosmosdb.request.channel_aquisition.duration
```
lmolkova, line 57: how is this different from https://github.com/open-telemetry/semantic-conventions/blob/2ce39aa87d5b773a85e5ac28b6b05ce4766ab8f7/model/database/metrics.yaml#L102C16-L102C47 ?

model/database/cosmosdb-metrics.yaml
```diff
@@ -26,3 +26,68 @@ groups:
...
+    unit: "{bytes}"
+    stability: experimental
+    extends: attributes.db.cosmosdb.network.minimal
+
+  - id: metric.db.client.cosmosdb.response.body.size
+    type: metric
+    metric_name: db.client.cosmosdb.response.body.size
+    brief: "Size of response received from Azure Cosmos DB backend"
+    instrument: histogram
+    unit: "{bytes}"
```
lmolkova, line 53:
```suggestion
    unit: "By"
```

(and in other places)

model/database/registry.yaml
sourabh1007: We want to have both, operation level statuscode and network level status code as a dimension

model/database/registry.yaml
sourabh1007: It is DB network call specific, which can be `http` or `rntbd` call. Basically, for an operation, SDK makes multiple network calls (`rntbd` for Direct Mode , `http` for Gateway Mode)

model/database/registry.yaml
sourabh1007: It is network specific information, because SDK always makes a network call to a particular partition id or pkrangeid. what about `cosmos_routing_id`

cc @FabianMeiswinkel @kirankumarkolli

model/database/cosmosdb-metrics.yaml
sourabh1007: Yes, it is the request body size transported over the wire and it is not compressed

docs/database/cosmosdb.md
sourabh1007: yes..this metrics would be available for gateway and direct mode both

model/database/cosmosdb-metrics.yaml
sourabh1007: This is the server-side latency (it is different from client-side latency) in SDK. we get the Backend server-side latency. which indicates, how much time it spends on the cosmos db sever to execute the particular query.

model/database/registry.yaml
lmolkova: why we can't use `db.response.status_code` for it?

model/database/registry.yaml
lmolkova: I don't see `db.cosmosdb.network.response.sub_status_code` used along with `db.cosmosdb.sub_status_code` on metrics. And if they were, how can they be different on the same telemetry item? What's the difference?

model/database/cosmosdb-metrics.yaml
lmolkova: could you add it to the brief or note?

docs/database/cosmosdb.md
lmolkova: why not rely on HTTP metrics for HTTP calls?

model/database/cosmosdb-metrics.yaml
sourabh1007: yes, we are going to report, backend latency to customer.

lmolkova: Let's document it then - that this is the latency returned in response and reported from the client. It will not be reported if response was not received

model/database/registry.yaml
sourabh1007: its already there as a dimension to represent operation level statuscode.

docs/database/cosmosdb.md
sourabh1007: We will loose the cosmos db specific dimensions and in HTTP we can't inject custom dimensions

model/database/cosmosdb-metrics.yaml
sourabh1007: a `rntbd` request has this timeline, ref https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/troubleshoot-dotnet-sdk-slow-request?tabs=cpu-new#requesttimeline

So we decided to capture, `ChannelAcquisitionStarted`,` Transit time`, `Received` as separate metrics.

model/database/registry.yaml
sourabh1007: for a same telemetry item, operation level substatuscode i.e. `db.cosmosdb.sub_status_code` would indicate if network call belongs to successful/failed operation (which also indicates that it was transient/non-transient issue at network layer)

model/database/registry.yaml
```diff
@@ -541,9 +542,50 @@ groups:
...
+        stability: experimental
+        brief: The HTTP status code returned from the network request.
+        examples: 200
+      - id: db.cosmosdb.network.response.sub_status_code
+        type: int
+        stability: experimental
+        brief: The Azure Cosmos DB sub-status code for a network request made during an operation.
+        examples: 1002
+      - id: db.cosmosdb.network.region
+        type: string
```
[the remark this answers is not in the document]
sourabh1007: I really don't want to list the region names here because we keep adding new region and it would be very hard to maintain that, along with it, i don't think (at least I am not aware) if we document region list anywhere.

model/database/cosmosdb-metrics.yaml
sourabh1007: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/troubleshoot-dotnet-sdk-slow-request?tabs=cpu-new#requesttimeline we already have this information for `rntbd` calls and for we are not going to capture it for `HTTP` calls

model/database/cosmosdb-metrics.yaml
sourabh1007: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/troubleshoot-dotnet-sdk-slow-request?tabs=cpu-new#requesttimeline

It is connection _get or create_ time and the above one is connection create time.

model/database/registry.yaml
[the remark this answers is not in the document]
lmolkova: how can you stay consistent across languages without have a list documented somewhere?

model/database/registry.yaml
lmolkova: when the network call ends you don't know how overall operation will go. Are you going to wait for it to complete before reporting telemetry?

model/database/registry.yaml
lmolkova: same comment as in https://github.com/open-telemetry/semantic-conventions/pull/1543#discussion_r1833661536 - I don't understand how you can report overall operation status on a specific network call. 

It'd mean that you need to accumulate all telemetry and report it only when overall operation ends. It'd lead to all sorts of implementation issues and edge cases, would increase memory consumption, and would keep metrics less real-time (potentially minutes). 

In case there is a huge amount of timing out operations, you'd have a hard time with providing observability into that's going on. 

In case of multi-request operations it becomes even more fun since you'd have to wait for all of them to complete and the status of the overall operation you report on a specific network request will have nothing to do with this request.

model/database/cosmosdb-metrics.yaml
lmolkova: How is this different? If it is, then the brief should reflect this. Currently it's

> Time taken in creating connection, It is only applicable in Direct Mode.

model/database/registry.yaml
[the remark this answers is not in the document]
sourabh1007: consistency already maintained across SDKs. As soon as we get a new region, we add it in this list https://github.com/Azure/azure-sdk-for-java/blob/ab2fdccc08946c9571424558700fce253118f4da/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/routing/RegionNameToRegionIdMap.java#L27 similar list is maintained in .Net SDK also.

Why do you want to put region list here, customer/user should not be concerned about the value, as soon as they are getting a value here.

model/database/registry.yaml
sourabh1007: Yes, we are currently collecting network-level telemetry at the completion or failure of each operation. The SDK is designed to capture this information, so we already have it available. I just need to emit it.

model/database/registry.yaml
sourabh1007: SDK is already designed to have this information available after the operation is completed. you can refer this PR https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4872

model/database/registry.yaml
FabianMeiswinkel: I like cosmos_routing_id

model/database/registry.yaml
FabianMeiswinkel: Maybe to step back and provide some context. For debuggability in the Cosmos SDKs with direct mode we already carry a state bag CosmosDiagnostics through all layers and add information needed for debuggability. Each operation/API call allows getting the CosmosDiagnostics from CosmosException or the Response in success cases. We provide an option to specify diagnostics thresholds to control when to emit additional debug info (whether to include the string/json representation of CosmosDiagnostics in logs/traces, whether to emit network lever metrics etc.). Why is this needed? We promise customers in the COsmos DB SLA not only availability SLAs from the end-client's perspective (vs. most other Azure services only looking at server-side availability) but the same for latency under certain conditions (10ms end-to-end latency). This means customers sometimes come and expect RCA/explanation why a (small) subset of operations misses SLA. And collectiong the Comos Diagnostics info in logs/traces network metrics all the time would be way to noisy and expensive/inefficient. And the decision whether diagnostics are needed can only be made on the operation level (an operation could very well have one slow network call - but due to various retry policies succeed successfully quickly - at that point we would not want to consider it a violation of diagnostics thresholds. So, we will have to emit logs, network metrics and even traces after the fact for network calls based on the CosmosDiagnostics collected anyway. Java is already doing this - it works well and also helps encapsulating the different diagnostics processing logic. So, I think Sourabh's approach for .Net makes sense as well.

model/database/registry.yaml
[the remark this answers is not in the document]
lmolkova: Here's the list of values https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/AzureLocation.cs or https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/AzureLocation.cs

You'd also get it if you run `az account list-locations -o table`

image

Here's the REST API behind it - `displayName` and `name` are part of this API - https://learn.microsoft.com/en-us/rest/api/subscription/subscriptions/list-locations?view=rest-subscription-2021-10-01&tabs=HTTP#location

As you can see
- there are at least three ways to record region. We need to pick one of them.
- this is not cosmos only concept. there is already an attribute for it - `cloud.region`

https://github.com/open-telemetry/semantic-conventions/blob/43f9e308ada81c5ef6851cd665d95f294b157a0f/model/cloud/registry.yaml#L48

So the proposal is to avoid adding a new attribute and instead 
- reuse `cloud.region`
- we can go ahead and update `cloud.region` description for azure and say that it should match either `displayName` or `name`. I think we should use `name` since it's one-word, lower-case and seems to be more query friendly.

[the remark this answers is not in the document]
sourabh1007: I am fine with `cloud.region` as an attribute name but value would be same as `displayName` in above list.

model/database/registry.yaml
sourabh1007: making it `routing_id`, cosmos is already there in the prefix.

model/database/registry.yaml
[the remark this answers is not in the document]
sourabh1007: I don't think we should put the whole list of Azure regions here. Because we allow new region without SDK upgrade and above list is mostly applicable for Management SDKs. and everytime, we do a region upgrade, we need to make sure this list is upto date for future support)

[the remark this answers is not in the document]
FabianMeiswinkel: @lmolkova - I agree the frustration that every Azure service defined region names more or less independently and there are different variations. It started because there was no central API/repository/guidance around this i guess, but the fact is that region names are independent for different services and there is no way to change it. You could of course extend the AzureLocation API to contain mappings to the service-specific region names - and especially for configuration that could be useful. But the drawback of using strict enums for region names is that it requires SDK upgrades for every single new Azure region we add. We were in this situation in the V2 .Net SDK (luckily now out of support) and for sure will not go down that direction again - it should be possible to start using a new Azure Region without forcing SDK upgrades. For the telemetry - using a strict enum is a bad idea for the same reason - adding any new value to an enum would require a new version of telemetry. How about making this a string - and referencing a document with the region names/mappings form the doucmentation. The consistency in region names within Cosmos DB across languages comes from the fact that the service requires certain region names to be used - so, the service normalizes there. And the AzureLocation above is by the way also missing regions (UsSec for example or New Zealand north). IMO any reasonable implementation in Azure.Core should get the metadata for regions based off of a service call - a region provider service - we could have all the mappings across sevrices there and provide one central place where new region configuration needs to be added and new regions would become usable without having to just through the hoops of upgrading SDKs.  If there is anyone in Azure Central willing to drive such an API please let me know and I am happy to contribute.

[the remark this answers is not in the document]
sourabh1007: As @FabianMeiswinkel  suggested here, we can keep it as of now and once we have the list supported region available publicly..I will update the link here.

docs/database/cosmosdb.md
```diff
@@ -14,6 +14,14 @@ linkTitle: Cosmos DB
     - [Metric: `db.client.response.returned_rows`](#metric-dbclientresponsereturned_rows)
     - [Metric: `db.client.cosmosdb.operation.request_charge`](#metric-dbclientcosmosdboperationrequest_charge)
     - [Metric: `db.client.cosmosdb.active_instance.count`](#metric-dbclientcosmosdbactive_instancecount)
+  - [Network Level Metrics](#network-level-metrics)
+    - [Metric: `db.client.cosmosdb.request.duration`](#metric-dbclientcosmosdbrequestduration)
+    - [Metric: `db.client.cosmosdb.request.body.size`](#metric-dbclientcosmosdbrequestbodysize)
+    - [Metric: `db.client.cosmosdb.response.body.size`](#metric-dbclientcosmosdbresponsebodysize)
+    - [Metric: `db.client.cosmosdb.request.channel_aquisition.duration`](#metric-dbclientcosmosdbrequestchannel_aquisitionduration)
```
FabianMeiswinkel, line 21: Please remove this metric from semantic profile. It will go away soon anyway (ThinClient) - so, fine to emit it for direct mode - but adding it to semantic profile to then removing it in a couple of months is bad.

docs/database/cosmosdb.md
```diff
@@ -14,6 +14,14 @@ linkTitle: Cosmos DB
...
+    - [Metric: `db.server.cosmosdb.request.duration`](#metric-dbservercosmosdbrequestduration)
+    - [Metric: `db.client.cosmosdb.request.transit.duration`](#metric-dbclientcosmosdbrequesttransitduration)
```
FabianMeiswinkel, line 23: Dito

docs/database/cosmosdb.md
```diff
@@ -14,6 +14,14 @@ linkTitle: Cosmos DB
...
+    - [Metric: `db.client.cosmosdb.request.received.duration`](#metric-dbclientcosmosdbrequestreceivedduration)
```
FabianMeiswinkel, line 24: Dito

docs/database/cosmosdb.md
```diff
@@ -40,10 +48,10 @@ Cosmos DB instrumentation includes call-level (public API) surface spans and net
...
-| [`db.cosmosdb.consistency_level`](/docs/attributes-registry/db.md) | string | Account or request [consistency level](https://learn.microsoft.com/azure/cosmos-db/consistency-levels). | `Eventual`; `ConsistentPrefix`; `BoundedStaleness`; `Strong`; `Session` | `Conditionally Required` If available. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.cosmosdb.connection_mode`](/docs/attributes-registry/db.md) | string | Azure Cosmos DB client connection mode. | `gateway`; `direct` | `Conditionally Required` [2] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.cosmosdb.consistency_level`](/docs/attributes-registry/db.md) | string | Effective Request [consistency level](https://learn.microsoft.com/azure/cosmos-db/consistency-levels). This represents the first applicable consistency level among those specified by the request or the client. | `Eventual`; `ConsistentPrefix`; `BoundedStaleness`; `Strong`; `Session` | `Conditionally Required` If available. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
FabianMeiswinkel, line 52:
```suggestion
| [`db.cosmosdb.consistency_level`](/docs/attributes-registry/db.md) | string | Effective Request [consistency level](https://learn.microsoft.com/azure/cosmos-db/consistency-levels). This represents the first applicable consistency level among those specified by the request, the client or the account's default consistency level. | `Eventual`; `ConsistentPrefix`; `BoundedStaleness`; `Strong`; `Session` | `Conditionally Required` If available. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

docs/database/cosmosdb.md
FabianMeiswinkel: It is not true That the Gateway endpoint always expresses the Cosmso DBAccount - I think you need a way to speicfy both - in most cases the accountname is simply the prefix for the Gateway address hostname - but they can be idfferent (and are for few customers who use their own DNS mapping)

FabianMeiswinkel (changes requested):
One blocking comment left

docs/database/cosmosdb.md
```diff
@@ -451,5 +456,448 @@ It captures the number of active instances at any given time. Best practices dic
...
+| [`db.cosmosdb.network.response.sub_status_code`](/docs/attributes-registry/db.md) | int | The Azure Cosmos DB sub-status code for a network request made during an operation. | `1002` | `Conditionally Required` If available | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.cosmosdb.sub_status_code`](/docs/attributes-registry/db.md) | int | The Azure Cosmos DB sub-status code for an operation represents the final sub-status code after multiple network interactions have been completed during the operation. | `1000`; `1002` | `Conditionally Required` when response was received and contained sub-code. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.namespace`](/docs/attributes-registry/db.md) | string | The name of the database, fully qualified within the server address and port. | `customers`; `test.users` | `Conditionally Required` If available. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
sourabh1007: Let's take it separately..if global account should be included or not..I will create a PR/issue with details. It is becoming very hard to keep this branch updated :(

model/database/registry.yaml
```diff
@@ -542,7 +543,9 @@ groups:
               value: "ConsistentPrefix"
               stability: experimental
         stability: experimental
-        brief: Account or request [consistency level](https://learn.microsoft.com/azure/cosmos-db/consistency-levels).
+        brief: >
+          Effective Request [consistency level](https://learn.microsoft.com/azure/cosmos-db/consistency-levels).
+          This represents the first applicable consistency level among those specified by the request, the client or the account's default consistency level.
```
trask, line 548:
```suggestion
          This represents the first applicable consistency level among those specified by the request, the client's default consistency level, or the account's default consistency level.
```

model/database/registry.yaml
```diff
@@ -555,6 +558,32 @@ groups:
...
+    display_name: Azure Cosmos DB Network level Attributes
+    stability: experimental
+    brief: >
+      This group defines attributes for Azure Cosmos DB Network call.
+    attributes:
+      - id: db.cosmosdb.network.response.status_code
+        type: int
+        stability: experimental
+        brief: The status code returned in response.
+        note: The protocol reuses HTTP status codes.
```
trask, line 573: I think it would be good to explain the difference between `db.cosmosdb.network.response.status_code` and `db.response.status_code` here.

model/database/registry.yaml
```diff
@@ -555,6 +558,32 @@ groups:
...
+        examples: 200
+      - id: db.cosmosdb.network.response.sub_status_code
+        type: int
+        stability: experimental
+        brief: The Azure Cosmos DB sub-status code for a network request made during an operation.
```
trask, line 578: I think it would be good to explain the difference between `db.cosmosdb.network.response.sub_status_code` and `db.cosmosdb.sub_status_code` here.

model/database/registry.yaml
sourabh1007: it is not client default consistency, consistency is defined at client level also

model/database/registry.yaml
sourabh1007: Improved category level desc.

model/database/registry.yaml
trask: I found the sentence confusing, is this correct then?

```suggestion
          This represents the first applicable consistency level among those specified by the request, the client's consistency level or the account's default consistency level.
```

sourabh1007: updated to this "This represents the applicable consistency level determined based on the hierarchy of settings: request-level, client instance-level, and account-level default consistency levels."

FabianMeiswinkel (approved):
LGTM

docs/database/cosmosdb.md
```diff
@@ -451,5 +456,448 @@ It captures the number of active instances at any given time. Best practices dic
...
+| [`db.operation.name`](/docs/attributes-registry/db.md) | string | The name of the operation or command being executed. [3] | `create_item`; `query_items`; `read_item` | `Conditionally Required` [4] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.response.status_code`](/docs/attributes-registry/db.md) | string | Database response status code. [5] | `200`; `201` | `Conditionally Required` [6] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 604: I still think it's highly problematic to request every instrumentation to:
- measure latency (with all attributes) at each network call
- propagate it back to the public API surface
- report these metrics there

This behavior is very difficult to implement is error prone. It would be problematic to implement in SDKs that don't get that much love such as Python, Go, JS, etc. Also I believe Java does not report metrics in this way now - https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/docs/Metrics.md

It also skews analysis since metrics will be reported at a different time than when response is received. If there are some locks or cases when multiple logical operations are in flight, there could be periods of time when no metrics are reported (since everything is waiting).

Given this, even if you do this now, you should have an escape path that allows you to change this behavior in the future.

The suggestion is to either
- report `db.response.status_code` and `db.cosmosdb.sub_status_code` only and describe only the network side of things in each measurement. I'd prefer this path
OR
- make  `db.response.status_code` and `db.cosmosdb.sub_status_code` `recommended` so that SDKs like Python or Go won't be required to implement all this logic and .NET would be able to  change this behavior.

model/database/common.yaml
```diff
@@ -59,6 +63,53 @@ groups:
...
+        examples: ["https", "rntbd"]
+        requirement_level:
+          conditionally_required: If available
+        note: >
+          The protocol name SHOULD be one of the following values: "https" or "rntbd".
+          The protocol name SHOULD be reported as "https" when the request is made over HTTPS i.e Azure Cosmos DB Gateway Service.
+          The protocol name SHOULD be reported as "rntbd" when the request is made over the Azure Cosmos DB transport protocol.
+      - ref: network.peer.address
+        stability: experimental
+        brief: Specifies the host address from the Gateway or Backend Service endpoint.
```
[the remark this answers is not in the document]
sourabh1007: FQDN

model/database/common.yaml
```diff
@@ -59,6 +63,53 @@ groups:
...
+  - id: attributes.db.cosmosdb.network.minimal
+    type: attribute_group
+    brief: 'Azure Cosmos DB Client network request level attributes'
+    stability: experimental
+    extends: attributes.db.cosmosdb.minimal
+    attributes:
+      - ref: network.protocol.name
+        stability: experimental
+        brief: Specifies the host protocol.
+        examples: ["https", "rntbd"]
```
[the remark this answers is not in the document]
sourabh1007: connection mode term would be confusing here because it is client side property and we make few `http` calls also in direct mode. IMO its ok, if customer doesn't undertand, what `rntbd` really is, attribute key clearly says that its a network protocol. OR we can use `tcp` term instead of `rntbd`, thats how we talk to the customers about `rntbd` calls.

docs/database/cosmosdb.md
sourabh1007: I understand your concern, but we are leveraging existing design to generate the metrics and avoiding any new code in the hot path.
**This behavior is very difficult to implement is error prone** => We have diagnostic engine implemented in SDK which collects all the information for a request and can be used to generate diagnostic JSON for that request (on demand). So, this engine will provide the data point for the metrics.
**Java does not report metrics in this way** => Java report metrics in this way only. There also we use diagnostic engine to generate these metrics. @FabianMeiswinkel can you please confirm it.

I would go with 2nd option; it would help us to correlate/analyze the network level behavior of successful operations.

@kirankumarkolli @FabianMeiswinkel thoughts?

FabianMeiswinkel: My 2 cents - keep all transport-level metrics out-of semantic preofile. There simply is no need to standardize it - we intentionally disable transport-level metrics by default anyway - and the whole purpose of them is debugging certain situations. When we have the new connection mode thin client proxy that will be supported across all SDKs, we can reconsider this - because then we will have one set of transport metrics that is applicable to all SDKs
Right now - let .Net and Java do whatever fits their needs for transport metrics - trying to standardize this metric is simply over-engineering.

model/database/common.yaml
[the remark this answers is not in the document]
lmolkova: Please clarify it in the description so that others who don't read discussions on the PR would understand it too

docs/database/cosmosdb.md
lmolkova: > My 2 cents - keep all transport-level metrics out-of semantic preofile

If metrics are not formally documented, they would be implemented in different ways in different languages. Also, users won't be able to understand what these metrics report or build shareable dashboards.

> Java report metrics in this way only. There also we use diagnostic engine to generate these metrics. 

It does not list two different sets of status attributes (global and local)  though

If you feel strong that it's necessary to report global operation status on all network calls, please apply my second suggestion and make it recommended and not required: 

> make db.response.status_code and db.cosmosdb.sub_status_code recommended so that SDKs like Python or Go won't be required to implement all this logic and .NET would be able to change this behavior.

sourabh1007:
We have decided, not to move forward with documentation of semantic conventions because Our transport layer is going to change in near future. Hence, we will revisit it once that is done.
We will document these metrics in our github repo as of now.

PR #1551, opened 2024-11-06
Remove "sampling relevant" from metric sections

trask:
"sampling relevant" is not applicable to metric attributes


lmolkova (approved):
Great catch! We should have some better checks/schema preventing metrics attributes from having `sampling_relevant` property

jsuereth (approved):

PR #1559, opened 2024-11-08
Generalize `db.query.parameter.<key>` to `db.operation.parameter.<key>`

trask:
Generalize `db.query.parameter.<key>` to `db.operation.parameter.<key>`, so that it can be used also for stored procedure parameters as well (and maybe other kinds of operations?)

Note: I don't think(?) schema transforms support attribute type `template[string]`


lmolkova:
> Note: I don't think(?) schema transforms support attribute type template[string]

@lquerel this is another data point on schema transformation limitations. Do we have an existing tracking issue I can update to include it?

lmolkova (approved):

lquerel:
@lmolkova, we have two GitHub issues related to this topic:
1. Why we need to change the way the OTEL Schema is specified: [Limitations and evolution of the OpenTelemetry Schema file](https://github.com/open-telemetry/weaver/issues/450).
2. The diff format I’m proposing, which could serve as the basis for a migration guide, schema update, etc.: [Diff between two versions of a semconv registry - Proposal](https://github.com/open-telemetry/weaver/issues/432).

It would be great to get your feedback on one or both of these issues, especially in light of the discussions in the current PR.

> > Note: I don't think(?) schema transforms support attribute type template[string]
> 
> @lquerel this is another data point on schema transformation limitations. Do we have an existing tracking issue I can update to include it?

jsuereth (approved):
We need to (eventually) support this in Schema URL.  For now, marking this breaking in changelog seems the only thing we can really do.

IIRC - the plan is to have a "2.0" like migration period for DB instrumentation right?

maryliag (approved):

Issue #1536, opened 2024-10-31
Document generic approach for span status (code + description) and exception event when instrumented code throws


lmolkova:
The common approach seems to be:
- do nothing if the exception is handled by the instrumented library (retries, etc)
- for unhandled (by client lib) exceptions:
  - set span status to error
  - set span status description to exception message
  - set `error.type` attribute on spans/metrics based on the exception type or more specific low-cardinality error code
  - DO NOT record exception event (by default) unless recording local root span or when the instrumentation knows that user code didn't handle the exception. Reason: exceptions are huge and expensive. Users decide if/how to record them when they catch them. If exception stays unhandled, the server instrumentation will record it.

We should document it and link from DB/messaging/gen-ai and other conventions.

alanwest:
Another aspect of this to consider is whether information should be captured from the outermost exception or innermost in the case of nested exceptions.

In capturing `error.type` and span status description, my intuition is that the innermost exception type/message is the most useful because it most closely describes the root of the problem.

On the other hand, when a user opts in to recording exception events, my intuition is that the outermost exception may be the most useful because the stack trace captured usually contains information from all nested exceptions.

Interestingly, the [DB semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/docs/database/database-spans.md#common-attributes) seems to suggest capturing details from the innermost exception:

> [9]: The error.type SHOULD match the db.response.status_code returned by the database or the client library, or the canonical name of exception that occurred. **When using canonical exception type name, instrumentation SHOULD do the best effort to report the most relevant type. For example, if the original exception is wrapped into a generic one, the original exception SHOULD be preferred.** Instrumentations SHOULD document how error.type is populated.

The HTTP semantic conventions seem less opinionated and does not have a similar statement.

trask:
> * DO NOT record exception event (by default) unless recording SERVER/CONSUMER span when the instrumentation knows that user code didn't handle the exception.

a slight variation on this is to replace SERVER/CONSUMER span above with "local root" span

trask:
since it appears that github discussions don't get backreferenced, linking here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/12125

cheempz:
[...]

Issue #1450, opened 2024-10-04
Define how to work with configuration from semconv


lmolkova:
We need to start creating config options in https://github.com/open-telemetry/opentelemetry-configuration and also start to documenting them in semconv

PR #1560, opened 2024-11-08
Add section on span status for databases

lmolkova:
Addresses the DB part of https://github.com/open-telemetry/semantic-conventions/issues/1536.

Let's polish it for databases and then expand to other conventions (and potentially have a generic guidance).

The intention is to keep exception event part experimental since 
- there doesn't seem to be a consensus or a common approach for it
- span events is going away

### DIFF
--- a/.chloggen/1560.yaml
+++ b/.chloggen/1560.yaml
@@ -1,0 +1,4 @@
+change_type: enhancement
+component: db
+note: Specify how to set span kind for database operations.
+issues: [1536, 1560]
--- a/.chloggen/1560.yaml
+++ b/.chloggen/1560.yaml
@@ -1,4 +1,4 @@
 change_type: enhancement
 component: db
-note: Specify how to set span kind for database operations.
+note: Specify how to set span status for database operations.
 issues: [1536, 1560]
--- a/docs/database/database-spans.md
+++ b/docs/database/database-spans.md
@@ -96,6 +96,7 @@ Instrumentation SHOULD consider the operation as failed if any of the following 
 - the `db.response.status_code` value indicates an error
 
   > [!NOTE]
+  >
   > The classification of status code as an error depends on the context.
   > For example, a SQL STATE `02000` (`no_data`) indicates an error when the application
   > expected the data to be available. However, it is not an error when the
@@ -127,15 +128,19 @@ When the operation ends with an error, instrumentation:
 **Status**: [Experimental][DocumentStatus]
 
 When the operation fails with an exception, instrumentation SHOULD record
-an [exception event](../exceptions/exceptions-spans.md) by default if and only if the
-span being recorded is a local root span (does not have a local parent).
+an [exception event](../exceptions/exceptions-spans.md) by default if, and only if,
+the span being recorded is a local root span (does not have a local parent).
 
 > [!NOTE]
+>
 > Exception stack traces could be very long and are expensive to capture and store.
 > Exceptions which are not handled by instrumented libraries are likely to be handled
 > and logged by the caller.
 > Exceptions that are not handled will be recorded by the outermost (local root)
 > instrumentation such as HTTP or gRPC server.
+
+Instrumentation MAY provide a configuration option to record exceptions that
+escape the surface of the instrumented API.
 
 ## Common attributes
 
@@ -369,6 +374,7 @@ The `db.query.summary` attribute captures a shortened representation of a query 
 which SHOULD have low-cardinality and SHOULD NOT contain any dynamic or sensitive data.
 
 > [!NOTE]
+>
 > The `db.query.text` attribute is intended to identify individual queries. Even though
 > it is sanitized if captured by default, it could still have high cardinality and
 > might reach hundreds of lines.
--- a/docs/database/database-spans.md
+++ b/docs/database/database-spans.md
@@ -89,16 +89,16 @@ For example, for an operation describing SQL query on an anonymous table like `S
 
 ## Status
 
-[Span Status Code][SpanStatus] MUST be left unset if operation has ended without any errors.
+[Span Status Code][SpanStatus] MUST be left unset if the operation has ended without any errors.
 
-Instrumentation SHOULD consider operation as failed if:
+Instrumentation SHOULD consider the operation as failed if any of the following is true:
 
 - the `db.response.status_code` value indicates an error
 
   > [!NOTE]
   >
   > The classification of status code as an error depends on the context.
-  > For example, a SQL STATE `no_data` status code indicates an error if the application
+  > For example, a SQL STATE `no_data` status code indicates an error when the application
   > expected the data to be available. However, it is not an error when the
   > application is simply checking whether the data exists.
   >
@@ -107,33 +107,33 @@ Instrumentation SHOULD consider operation as failed if:
   > Instrumentations that don't have any additional context MUST follow the
   > guidelines in this section.
 
-- exception is thrown by the instrumented method call
-- instrumented method returns error in another way
+- an exception is thrown by the instrumented method call
+- the instrumented method returns error in another way
 
-When operation ends with an error, instrumentation:
+When the operation ends with an error, instrumentation:
 
-- SHOULD set span status code to `Error`
-- SHOULD set `error.type` attribute
-- SHOULD set span status description when it has additional information
+- SHOULD set the span status code to `Error`
+- SHOULD set the `error.type` attribute
+- SHOULD set the span status description when it has additional information
   about the error that aligns with [Span Status Description][SpanStatus] and
   can't be inferred from `db.response.status_code` or `error.type` attributes.
 
-  When operation fails with exception, the span status description SHOULD be set to
-  exception message.
+  When the operation fails with an exception, the span status description SHOULD be set to
+  the exception message.
 
 ### Recording exception events
 
 **Status**: [Experimental][DocumentStatus]
 
-When operation fails with exception, instrumentation MAY record
-[exception event](../exceptions/exceptions-spans.md) by default if and only if the
+When the operation fails with an exception, instrumentation MAY record
+an [exception event](../exceptions/exceptions-spans.md) by default if and only if the
 span being recorded is a local root span (does not have a local parent).
 
 > [!NOTE]
 > Exception stack traces could be very long and are expensive to capture and store.
 > Exceptions which are not handled by instrumented libraries are likely to be caught,
 > logged, and handled by the caller.
-> Exceptions that are not handled are recorded by they outermost instrumentation
+> Exceptions that are not handled will be recorded by the outermost (local root) instrumentation
 > such as HTTP or gRPC server.
 
 Instrumentations SHOULD NOT record exception events (by default) on spans that
--- a/docs/database/database-spans.md
+++ b/docs/database/database-spans.md
@@ -98,7 +98,7 @@ Instrumentation SHOULD consider the operation as failed if any of the following 
   > [!NOTE]
   >
   > The classification of status code as an error depends on the context.
-  > For example, a SQL STATE `no_data` status code indicates an error when the application
+  > For example, a SQL STATE `02000` (`no_data`) indicates an error when the application
   > expected the data to be available. However, it is not an error when the
   > application is simply checking whether the data exists.
   >
@@ -108,15 +108,17 @@ Instrumentation SHOULD consider the operation as failed if any of the following 
   > guidelines in this section.
 
 - an exception is thrown by the instrumented method call
-- the instrumented method returns error in another way
+- the instrumented method returns an error in another way
 
 When the operation ends with an error, instrumentation:
 
 - SHOULD set the span status code to `Error`
 - SHOULD set the `error.type` attribute
 - SHOULD set the span status description when it has additional information
-  about the error that aligns with [Span Status Description][SpanStatus] and
-  can't be inferred from `db.response.status_code` or `error.type` attributes.
+  about the error that aligns with [Span Status Description][SpanStatus].
+
+  It's NOT RECOMMENDED to duplicate `db.response.status_code` or `error.type`
+  in span status description.
 
   When the operation fails with an exception, the span status description SHOULD be set to
   the exception message.
@@ -125,7 +127,7 @@ When the operation ends with an error, instrumentation:
 
 **Status**: [Experimental][DocumentStatus]
 
-When the operation fails with an exception, instrumentation MAY record
+When the operation fails with an exception, instrumentation SHOULD record
 an [exception event](../exceptions/exceptions-spans.md) by default if and only if the
 span being recorded is a local root span (does not have a local parent).
 
@@ -135,9 +137,6 @@ span being recorded is a local root span (does not have a local parent).
 > logged, and handled by the caller.
 > Exceptions that are not handled will be recorded by the outermost (local root) instrumentation
 > such as HTTP or gRPC server.
-
-Instrumentations SHOULD NOT record exception events (by default) on spans that
-have local parents.
 
 ## Common attributes
 
--- a/docs/database/database-spans.md
+++ b/docs/database/database-spans.md
@@ -96,7 +96,6 @@ Instrumentation SHOULD consider the operation as failed if any of the following 
 - the `db.response.status_code` value indicates an error
 
   > [!NOTE]
-  >
   > The classification of status code as an error depends on the context.
   > For example, a SQL STATE `02000` (`no_data`) indicates an error when the application
   > expected the data to be available. However, it is not an error when the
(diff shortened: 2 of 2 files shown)


docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
 If a corresponding `{target}` value is not available for a specific operation, the instrumentation SHOULD omit the `{target}`.
 For example, for an operation describing SQL query on an anonymous table like `SELECT * FROM (SELECT * FROM table) t`, span name should be `SELECT`.
 
+## Status
+
+[Span Status Code][SpanStatus] MUST be left unset if operation has ended without any errors.
```
trask, line 92:
```suggestion
[Span Status][SpanStatus] MUST be left unset if the operation has ended without any errors.
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+
+Instrumentation SHOULD consider operation as failed if:
```
trask, line 94:
```suggestion
Instrumentation SHOULD consider the operation as failed if any of the following is true:
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+
+- the `db.response.status_code` value indicates an error
+
+  > [!NOTE]
+  >
+  > The classification of status code as an error depends on the context.
+  > For example, a SQL STATE `no_data` status code indicates an error if the application
+  > expected the data to be available. However, it is not an error when the
```
trask, line 102: (just to match the `when` in the next sentence)
```suggestion
  > For example, a SQL STATE `no_data` status code indicates an error when the application
  > expected the data to be available. However, it is not an error when the
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+  > application is simply checking whether the data exists.
+  >
+  > Instrumentations that have additional context about a specific operation MAY use
+  > this context to set the span status more precisely.
+  > Instrumentations that don't have any additional context MUST follow the
+  > guidelines in this section.
+
+- exception is thrown by the instrumented method call
```
trask, line 110:
```suggestion
- an exception is thrown by the instrumented method call
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+- instrumented method returns error in another way
```
trask, line 111:
```suggestion
- the instrumented method returns error in another way
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+
+When operation ends with an error, instrumentation:
```
trask, line 113:
```suggestion
When the operation ends with an error, instrumentation:
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+
+- SHOULD set span status code to `Error`
+- SHOULD set `error.type` attribute
```
trask, line 116:
```suggestion
- SHOULD set the `error.type` attribute
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+- SHOULD set span status description when it has additional information
```
trask, line 117:
```suggestion
- SHOULD set the span status description when it has additional information
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+  > this context to set the span status more precisely.
+  > Instrumentations that don't have any additional context MUST follow the
+  > guidelines in this section.
+
+- exception is thrown by the instrumented method call
+- instrumented method returns error in another way
+
+When operation ends with an error, instrumentation:
+
+- SHOULD set span status code to `Error`
```
trask, line 115:
```suggestion
- SHOULD set the span status code to `Error`
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+- SHOULD set `error.type` attribute
+- SHOULD set span status description when it has additional information
+  about the error that aligns with [Span Status Description][SpanStatus] and
+  can't be inferred from `db.response.status_code` or `error.type` attributes.
+
+  When operation fails with exception, the span status description SHOULD be set to
+  exception message.
```
trask, line 122:
```suggestion
  When the operation fails with an exception, the span status description SHOULD be set to
  the exception message.
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+
+### Recording exception events
+
+**Status**: [Experimental][DocumentStatus]
+
+When operation fails with exception, instrumentation MAY record
+[exception event](../exceptions/exceptions-spans.md) by default if and only if the
```
trask, line 129:
```suggestion
When the operation fails with an exception, instrumentation MAY record
an [exception event](../exceptions/exceptions-spans.md) by default if and only if the
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+span being recorded is a local root span (does not have a local parent).
+
+> [!NOTE]
+> Exception stack traces could be very long and are expensive to capture and store.
+> Exceptions which are not handled by instrumented libraries are likely to be caught,
+> logged, and handled by the caller.
+> Exceptions that are not handled are recorded by they outermost instrumentation
```
trask, line 136:
```suggestion
> Exceptions that are not handled will be recorded by the outermost (local root) instrumentation
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+[Span Status Code][SpanStatus] MUST be left unset if the operation has ended without any errors.
+
+Instrumentation SHOULD consider the operation as failed if any of the following is true:
+
+- the `db.response.status_code` value indicates an error
+
+  > [!NOTE]
+  >
+  > The classification of status code as an error depends on the context.
+  > For example, a SQL STATE `no_data` status code indicates an error when the application
```
lmolkova, line 101:
```suggestion
  > For example, a SQL STATE `02000` (`no_data`) status code indicates an error when the application
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+  > expected the data to be available. However, it is not an error when the
+  > application is simply checking whether the data exists.
+  >
+  > Instrumentations that have additional context about a specific operation MAY use
+  > this context to set the span status more precisely.
+  > Instrumentations that don't have any additional context MUST follow the
+  > guidelines in this section.
+
+- an exception is thrown by the instrumented method call
+- the instrumented method returns error in another way
```
trask, line 111:
```suggestion
- the instrumented method returns an error in another way
```

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+
+When the operation ends with an error, instrumentation:
+
+- SHOULD set the span status code to `Error`
+- SHOULD set the `error.type` attribute
+- SHOULD set the span status description when it has additional information
+  about the error that aligns with [Span Status Description][SpanStatus] and
+  can't be inferred from `db.response.status_code` or `error.type` attributes.
```
trask, line 119: I didn't follow why this part? thanks

> and can't be inferred from `db.response.status_code` or `error.type` attributes

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+
+> [!NOTE]
+> Exception stack traces could be very long and are expensive to capture and store.
+> Exceptions which are not handled by instrumented libraries are likely to be caught,
+> logged, and handled by the caller.
+> Exceptions that are not handled will be recorded by the outermost (local root) instrumentation
+> such as HTTP or gRPC server.
+
+Instrumentations SHOULD NOT record exception events (by default) on spans that
+have local parents.
```
trask, line 140: this seems covered by the "if and only if" above?

docs/database/database-spans.md
```diff
@@ -85,6 +87,58 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+  can't be inferred from `db.response.status_code` or `error.type` attributes.
+
+  When the operation fails with an exception, the span status description SHOULD be set to
+  the exception message.
+
+### Recording exception events
+
+**Status**: [Experimental][DocumentStatus]
+
+When the operation fails with an exception, instrumentation MAY record
```
trask, line 128: what are your thoughts?
```suggestion
When the operation fails with an exception, instrumentation SHOULD record
```

docs/database/database-spans.md
lmolkova: I'm trying to avoid duplicating things between `db.response.status_code` / `error.type` and span status description.

We have a similar wording in HTTP 

https://github.com/open-telemetry/semantic-conventions/blob/43f9e308ada81c5ef6851cd665d95f294b157a0f/docs/http/http-spans.md?plain=1#L112

Let me see if I can explain it more clearly

trask: > We have a similar wording in HTTP

👍

I was thinking it might be useful for backends which display error status + description in a general way across all spans, but definitely understand the duplication concern

trask: I'm good resolving, makes sense to stay consistent with HTTP

docs/database/database-spans.md
```diff
@@ -85,6 +87,56 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+When the operation fails with an exception, instrumentation SHOULD record
+an [exception event](../exceptions/exceptions-spans.md) by default if and only if the
+span being recorded is a local root span (does not have a local parent).
+
+> [!NOTE]
+> Exception stack traces could be very long and are expensive to capture and store.
+> Exceptions which are not handled by instrumented libraries are likely to be handled
+> and logged by the caller.
+> Exceptions that are not handled will be recorded by the outermost (local root)
+> instrumentation such as HTTP or gRPC server.
```
trask, line 138: let's ask for feedback on this point from the Specification SIG before we merge

trask (approved):

docs/database/database-spans.md
jsuereth: I think this is a directional shift in logging exceptions.

I see two things here:

1. We still can't determine "layers" of instrumentation, i.e. we *can't* know if someone else will log the exception or not.
2. This violates the previously recorded wisdom of how to handle exceptions and stack traces.


My opinion:

This is the right direction to go.  We *should* assume that there is "outer" instrumentation that logs/catches exceptions, particularly for database "client" spans.  I.e. we may be able to blanketly state that "client" spans should never log unhandled exceptions and that we encourage "outer" instrumentation that will (e.g. server spans).

docs/database/database-spans.md
```diff
@@ -85,6 +87,56 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+  > guidelines in this section.
+
+- an exception is thrown by the instrumented method call
+- the instrumented method returns an error in another way
+
+When the operation ends with an error, instrumentation:
+
+- SHOULD set the span status code to `Error`
+- SHOULD set the `error.type` attribute
+- SHOULD set the span status description when it has additional information
```
maryliag, line 116: should this have any requirements about not having any confidential data? Similar to how we have on metrics errors where we don't add if it shows actual values of queries, etc

docs/database/database-spans.md
trask: I added to next week's spec SIG agenda

docs/database/database-spans.md
```diff
@@ -85,6 +87,56 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+### Recording exception events
+
+**Status**: [Experimental][DocumentStatus]
+
+When the operation fails with an exception, instrumentation SHOULD record
+an [exception event](../exceptions/exceptions-spans.md) by default if and only if the
+span being recorded is a local root span (does not have a local parent).
+
+> [!NOTE]
+> Exception stack traces could be very long and are expensive to capture and store.
```
joaopgrassi, line 134:
```suggestion
>
> Exception stack traces could be very long and are expensive to capture and store.
```
I think you need an empty line here so the icon shows up.

docs/database/database-spans.md
```diff
@@ -85,6 +87,56 @@ and SHOULD adhere to one of the following values, provided they are accessible:
...
+
+  When the operation fails with an exception, the span status description SHOULD be set to
+  the exception message.
+
+### Recording exception events
+
+**Status**: [Experimental][DocumentStatus]
+
+When the operation fails with an exception, instrumentation SHOULD record
+an [exception event](../exceptions/exceptions-spans.md) by default if and only if the
```
joaopgrassi, line 130:
```suggestion
an [exception event](../exceptions/exceptions-spans.md) by default if, and only if, the
```

docs/database/database-spans.md
trask: Feedback from the spec SIG was that this sounds like a good default behavior, but we should make it configurable.

Some options for making it configurable:

(A) handle this all in the SDK:
* new configuration added to [TracerConfig](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#tracerconfig)
* [RecordException](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception) behavior could depend on this new configuration in TracerConfig

but I don't think we can change the default behavior of the SDK.

(B) Another option could be to introduce "trace advice" (similar to metric advice) that would allow a given instrumentation the ability to opt-in to this new behavior.

(C) Another option could be:
* new configuration added to [TracerConfig](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#tracerconfig)
* instead of changing the (existing) behavior of RecordException, ask instrumentations to check this configuration value before calling RecordException

(D) Another option could be:
* Introduce a standard "database" instrumentation configuration property (similar to the standard "http" configuration properties that we already have: https://github.com/open-telemetry/opentelemetry-configuration/blob/e1f89ef021f3c8e000d05e6b6e53bc4a2372fe52/examples/kitchen-sink.yaml#L396-L400)
* ask instrumentations to check this configuration value before calling RecordException
* we may also want to be able to override this at individual instrumentation level somehow
 
(E) Or even:
* Introduce a global instrumentation configuration property that covers all instrumentations
* ask instrumentations to check this configuration value before calling RecordException
* we may also want to be able to override this at individual instrumentation level somehow

Probably we should spin this out into a separate issue.

cc @jack-berg

docs/database/database-spans.md
lmolkova: I don't see how database instrumentation can determine if there is any sensitive or confidential data in the arbitrary exception message.  Someone can put sensitive data into database name (bluesky has DB per user and they might call them accordingly), but how can we provide basic observability in the common case if we assume the worst? 

Happy to consider any specific cases when we can do something on the instrumentation side.

docs/database/database-spans.md
lmolkova: Thanks for the notes!

Since we don't usually specify how instrumentations should configure things, I'd prefer to tackle it separately. 
It's definitely a problem that we should fix in the https://github.com/open-telemetry/semantic-conventions/issues/1536 and https://github.com/open-telemetry/semantic-conventions/issues/1450

.chloggen/1560.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: db
+note: Specify how to set span kind for database operations.
```
lmolkova, line 3:
```suggestion
note: Specify how to set span status for database operations.
```

docs/database/database-spans.md
maryliag: Sorry if I wasn't clear, I didn't mean that is the job of the instrumentation to determine this, I meant more as a recommendation, that this sentence could include something on the lines of 

```
SHOULD set the span status description when it has additional information about the error that aligns with [Span Status Description][SpanStatus] and does not contain sensitive data.
```

maryliag (dismissed):

maryliag (approved):

jsuereth (approved):

PR #1562, opened 2024-11-08
Use project version of markdown-link-check package

joaopgrassi:
## Changes

The `changelog.yaml` workflow is installing the latest version of the `markdown-link-check` package on each run, and the latest version seems to have a issue (See #1552).

Adding a make target to run this task, so then it uses the version defined in `package.json` for the repo.


.github/workflows/changelog.yml
```diff
@@ -76,13 +76,6 @@ jobs:
 
       # In order to validate any links in the yaml file, render the config to markdown
       - name: Render .chloggen changelog entries
-        run: make chlog-preview > changelog_preview.md
-      - name: Install markdown-link-check
-        run: npm install -g markdown-link-check
+        run: make chlog-preview 2> changelog_preview.md
```
joaopgrassi, line 79: I noticed while running this `make chlog-preview > changelog_preview.md` on Ubuntu that the file created does not contain the changelog preview, but only the actual make code.

joaopgrassi: Now it works :) 

image

christophe-kamphaus-jemmic:
Is there a commit on which the changelog workflow succeeded and wasn't skipped?

joaopgrassi:
> Is there a commit on which the changelog workflow succeeded and wasn't skipped?

No. The problem is that I added the new make target, but I can't use it in the workflow yet, since it fails because the new target is not there. So I guess we will only know when it merges. I only tested it locally and all worked.

jsuereth (approved):

trask (approved):

arminru (approved):

joaopgrassi:
Let's see if it works 😅

PR #1566, opened 2024-11-10
Don't capture `db.operation.name` and `db.collection.name` from query formats that support multiples

trask:
Don't capture `db.operation.name` and `db.collection.name` from query formats that support multiples.

The reason behind this change is that it could be confusing to have `db.operation.name` and `db.collection.name` captured for _some_ SQL operations but not for other SQL operations, also leading to an inconsistent experience in SQL metrics.

And these parsed values (single or multiple) are now present in `db.query.summary`, so the loss of information seems minimal.

Specifically, `db.operation.name` is changed from:

> A single database query may involve multiple operations. If the operation name is parsed from the query text, it SHOULD only be captured for queries that contain a single operation or when the operation name describing the whole query is available by other means.

to:

> The operation name SHOULD NOT be extracted from `db.query.text`, unless the query format is known to only ever have a single operation name present.

and `db.collection.name` is changed from:

> If the collection name is parsed from the query text, it SHOULD only be captured for queries that contain a single collection and it SHOULD match the value provided in the query text including any schema and database name prefix.

to:

> The collection name SHOULD NOT be extracted from `db.query.text`, unless the query format is known to only ever have a single collection name present and the collection name is known to always be under the captured `db.namespace`.


docs/attributes-registry/db.md
```diff
@@ -34,18 +34,13 @@ This group defines the attributes used to describe telemetry in the context of d
...
 **[1]:** It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.
 
-A single database query may involve multiple collections.
-
-If the collection name is parsed from the query text, it SHOULD only be captured for queries that
-contain a single collection and it SHOULD match the value provided in
-the query text including any schema and database name prefix.
+The collection name SHOULD NOT be extracted from `db.query.text`,
+unless the query format is known to only ever have a single collection name present
+and the collection name is known to always be under the captured `db.namespace`.
```
lmolkova, line 39: > collection name is known to always be under the captured `db.namespace`

curious - why?

Also maybe 

```suggestion
and the collection name is known to always belong to the captured `db.namespace`.
```

(I've read it first as that collection name should be captured as `db.namespace`)

trask: > > collection name is known to always be under the captured `db.namespace`
> 
> curious - why?

I was thinking in cases where there's only ever a single collection name that it would probably(?) be easy to ensure (and would be a nice normalization), but I've removed this part from the PR since it's tangential anyways and I'll open separate issue to explore.

docs/database/hbase.md
```diff
@@ -40,10 +40,8 @@ This attribute has stability level RELEASE CANDIDATE.
 **[3]:** It is RECOMMENDED to capture the value as provided by the application
 without attempting to do any case normalization.
 
-A single database query may involve multiple operations. If the operation
-name is parsed from the query text, it SHOULD only be captured for queries that
-contain a single operation or when the operation name describing the
-whole query is available by other means.
+The operation name SHOULD NOT be extracted from `db.query.text`,
```
lmolkova, line 43: I don't think it's applicable to HBASE (but as in other cases, this is not a new problem introduced in this PR)

docs/database/redis.md
```diff
@@ -44,10 +44,8 @@ Instrumentation SHOULD document if `db.namespace` reflects the database index pr
 **[2]:** It is RECOMMENDED to capture the value as provided by the application
 without attempting to do any case normalization.
 
-A single database query may involve multiple operations. If the operation
-name is parsed from the query text, it SHOULD only be captured for queries that
-contain a single operation or when the operation name describing the
-whole query is available by other means.
+The operation name SHOULD NOT be extracted from `db.query.text`,
```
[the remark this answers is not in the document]
matt-hensley: Depending on how you define query, Redis has MULTI/EXEC to wrap multiple commands in a single transaction:
https://redis.io/docs/latest/develop/interact/transactions/

The response will be an array with the output of each command.

This is used commonly to implement rate limiting.

docs/database/hbase.md
trask: Tracking in #1573

docs/database/cosmosdb.md
```diff
@@ -348,13 +348,20 @@ Explaining bucket configuration:
...
+This attribute has stability level RELEASE CANDIDATE.
+
 **[2]:** It is RECOMMENDED to capture the value as provided by the application
 without attempting to do any case normalization.
 
-A single database query may involve multiple operations. If the operation
-name is parsed from the query text, it SHOULD only be captured for queries that
-contain a single operation or when the operation name describing the
-whole query is available by other means.
+The operation name SHOULD NOT be extracted from `db.query.text`,
```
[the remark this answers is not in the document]
trask: Tracking in #1573

docs/database/cassandra.md
```diff
@@ -68,10 +63,8 @@ This attribute has stability level RELEASE CANDIDATE.
 **[4]:** It is RECOMMENDED to capture the value as provided by the application
 without attempting to do any case normalization.
 
-A single database query may involve multiple operations. If the operation
-name is parsed from the query text, it SHOULD only be captured for queries that
-contain a single operation or when the operation name describing the
-whole query is available by other means.
+The operation name SHOULD NOT be extracted from `db.query.text`,
```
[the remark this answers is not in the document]
trask: Tracking in #1573

docs/database/cassandra.md
```diff
@@ -40,22 +40,16 @@ The Semantic Conventions for [Cassandra](https://cassandra.apache.org/) extend a
...
 | [`network.peer.port`](/docs/attributes-registry/network.md) | int | Peer port number of the network connection. | `65123` | `Recommended` if and only if `network.peer.address` is set. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 | [`server.address`](/docs/attributes-registry/server.md) | string | Name of the database host. [17] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
-| [`db.operation.parameter.<key>`](/docs/attributes-registry/db.md) | string | A database operation parameter, with `<key>` being the parameter name, and the attribute value being a string representation of the parameter value. [18] | `someval`; `55` | `Opt-In` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`db.query.parameter.<key>`](/docs/attributes-registry/db.md) | string | A query parameter used in `db.query.text`, with `<key>` being the parameter name, and the attribute value being a string representation of the parameter value. | `someval`; `55` | `Opt-In` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `db.operation.parameter`. |
```
lmolkova, line 43: oops - need to update attribute reference to the new one

lmolkova (approved):

PR #1576, opened 2024-11-13
Add `device.crash` event semantic convention

bidetofevil:
Adding the definition for an `Event` that models a crash in the mobile app. There are still some open questions to be worked out regarding naming and the specific event attribute definitions, but I want to put the draft changes in a PR to get further feedback from a larger group, having done so already via a doc with some other folks in the mobile community.

The PR is not ready to merging right now. I will make it a non-draft when it is.


model/device/events.yaml
```diff
@@ -82,3 +82,131 @@ groups:
...
+            ],
+            \"proguard_file_id\": \"<UUID>\"
+          }" ]
+        - id: data_content_type
+          stability: experimental
+          requirement_level: recommended
+          brief: >
+            The format of the `data` field as defined by [RFC 2046](https://datatracker.ietf.org/doc/html/rfc2046).
+          note: >
+            This, combined with a priori knowledge of the structure of the blob, will allow Collectors to parse and process the `data` field.
```
LikeTheSalad, line 192: I'd like to check this with an example just to make sure I got it properly.

Considering an event with `source = jvm_exception`. Let's say that this `data_content_type` field has `application/json` as its value. The collector in this case will be able to parse it as json, however, how can it know what fields to look for in that json? My question is mostly about ensuring that the UI can later display this data properly.

LikeTheSalad (commented):
Thank you 🙌 I think it's a great start.

jzwc (commented):
Thank you for your initiative.

~I apologize for not attending the SIG calls where this may have been discussed, but I am curious if there is a proposal to include an additional attribute that would help identify the crash report format beyond just the MIME type. Specifically, on iOS, there are several well-known libraries that generate crash reports in their own, somewhat distinct formats. For instance, knowing that a crash report was generated using PLCrashReporter could enhance interoperability. While this information could be incorporated into the MIME type, it would require parsing it out, which introduces another potential source of incompatibility.~

Never mind, this is indeed covered by the `source` attribute.

model/device/events.yaml
```diff
@@ -82,3 +82,131 @@ groups:
...
+    body:
+      id: device_crash_state
+      type: map
+      requirement_level: required
+      fields:
+        - id: id
+          stability: experimental
+          requirement_level: required
+          brief: >
+            An ID that uniquely identifies the crash instance obtained from a specific `source`.
```
tonzhan2, line 118: How would this be generated? From say, an Android device crash?

tonzhan2:
This looks like a great start. I'm curious as to what some examples of crashes could look like in this format. Such as an Android jvm crash. Obviously most of it will be in the body, so I guess of that format is still tbd. Could be useful nonetheless to have some examples of crashes that use this formatting

model/device/events.yaml
jzwc: Never mind the previous comment. This is indeed covered by the `source` attribute.

jzwc:
Just another quick thought. It would be generally beneficial is app state / resource attributes like `device.app.lifecycle` would be allowed as the crash event attributes to pass the respective application and device state at the moment of the crash.

model/device/events.yaml
```diff
@@ -82,3 +82,131 @@ groups:
...
+        - id: data_content_type
+          stability: experimental
+          requirement_level: recommended
+          brief: >
+            The format of the `data` field as defined by [RFC 2046](https://datatracker.ietf.org/doc/html/rfc2046).
+          note: >
+            This, combined with a priori knowledge of the structure of the blob, will allow Collectors to parse and process the `data` field.
+          examples: [ "application/json" ]
+          type: string
+        - id: crashed_service_version
```
thompson-tomo, line 195: Use attribute field instead and use the corresponding pre-existing semconv attribute.

model/device/events.yaml
```diff
@@ -82,3 +82,131 @@ groups:
...
+          stability: experimental
+          requirement_level: recommended
+          brief: >
+            The version of the app when the crash happened, which may be different than the `service.version` resource attribute.
+          note: >
+            This is required so crashes can be aggregated by the version in which it occurred, not the one that emitted the event.
+          examples: [ "7.5.0" ]
+          type: string
+        - id: crashed_os_version
```
thompson-tomo, line 204: Use attribute field instead and use the corresponding pre-existing semconv attribute.

model/device/events.yaml
```diff
@@ -82,3 +82,131 @@ groups:
...
+  - id: event.device.crash
+    stability: experimental
+    type: event
+    name: device.crash
+    brief: >
+      A crash event represents the termination of an application instance due to an unhandled error or exception. It can be detected and
+      recorded as it is happening (e.g. through an UncaughtExceptionHandler), or after the fact, when a tombstone is detected
+      containing information about a previously terminated app instance that was caused by an unhandled error or exception.
+    note: >
+      The body fields of this event contain data and metadata about the crash tht can be used to classify and aggregate it with similar
```
KalleOlaviNiemitalo, line 95: Should be "about the crash **that** can be used".

breedx-splk:
@bidetofevil Do you want to circle back on this? It's kinda stale and needs a rebase at the least. I'd love to see this work out. Thanks!

bidetofevil:
@breedx-splk Yeah, I'm looking to get back on this soon. I put it aside when I got busy and, well, it's finall time to look at this again. Starting with getting it building and reviewing the comments.

LikeTheSalad:
Just for the sake of using semantic conventions in OTel Android instrumentations, I've created [this PR](https://github.com/open-telemetry/semantic-conventions/pull/3441) to add the `device.crash` event with minimal requirements (open to future enhancements).

PR #1583, opened 2024-11-15
[chore] Render attribute name on foot notes

joaopgrassi:
## Changes

Renders the attribute name on the generated table foot notes. In my opinion this makes looking at the notes easier, as one usually needs to go back and forth to remember what the [x] index refers to. 


arminru (approved):

mx-psi (approved):

docs/http/http-metrics.md
```diff
@@ -124,21 +124,21 @@ additional filters are applied.
...
 If the request has completed successfully, instrumentations SHOULD NOT set `error.type`.
 
-**[4]:** MUST NOT be populated when this is not supported by the HTTP server framework as the route attribute should have low-cardinality and the URI path can NOT substitute it.
+**[4] `http.route`:** MUST NOT be populated when this is not supported by the HTTP server framework as the route attribute should have low-cardinality and the URI path can NOT substitute it.
 SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
 
-**[5]:** The value SHOULD be normalized to lowercase.
+**[5] `network.protocol.name`:** The value SHOULD be normalized to lowercase.
 
 **[6]:** If not `http` and `network.protocol.version` is set.
```
lmolkova, line 132: wonder if we should also render the attribute name on requirement level notes?

I'd actually love to merge them eventually to one note in markdown with more structure (not for this PR)

lmolkova (approved):
Thank you! This is a great step forward

jsuereth (approved):

docs/http/http-metrics.md
joaopgrassi: > wonder if we should also render the attribute name on requirement level notes?

Yes, I was wondering the same! But this will require some more changes, so if it's OK I'd like to do that in a follow up. I will create an issue for it and work on it.

joaopgrassi: Here it is: #1587

PR #1586, opened 2024-11-16
[chore] Add lightweight make check target

braydonk:
## Changes

The link check is super slow, but the rest of the `all` checks are useful while iterating on a change. This PR adds a new target that depends on all the same targets minus the link check.


lmolkova (approved):

Makefile
```diff
@@ -41,6 +41,10 @@ DOCKER_USER=$(shell id -u):$(shell id -g)
 all: install-tools markdownlint markdown-link-check misspell table-check compatibility-check schema-check \
 		 check-file-and-folder-names-in-docs
 
+.PHONY: all-skip-link-check
+all: install-tools markdownlint misspell table-check compatibility-check schema-check \
```
codeboten, line 45: This should be updated to all-skip-link-check

braydonk: Fixed when I renamed the target in [8ccd778](https://github.com/open-telemetry/semantic-conventions/pull/1586/commits/8ccd778a05529eca1bd3598e220ccff5f1cfd558)

arminru:
> The Docker docs site is having some manner of outage (DNS? certs?)

@braydonk seems to be back now, I'll run the build again

trask (approved):

Issue #864, opened 2024-03-29
Process semconv: define common attributes and revisit requirement levels


lmolkova:
### Area(s)

area:process

### Is your change request related to a problem? Please describe.

We require any of the following attributes to be present

https://github.com/open-telemetry/semantic-conventions/blob/a04a1f4ba413d0a1f137c223fc243ea14fdcd89b/docs/resource/process.md?plain=1#L39-L45

As a result, users would get inconsistent data from different systems and instrumentations. 
Vendors will need to know how to parse and extract specific details (such as process name).

### Describe the solution you'd like

The approach we follow in HTTP and other semconv:
- Find 1-2 required attributes (e.g. `process.executable.name`) which are available everywhere
- Require instrumentations to do the parsing/extraction of this common information so that users and consumers would get consistent data
- Set appropriate requirement levels on other attributes

ChrsMark:
cc: @open-telemetry/semconv-system-approvers

mx-psi:
Discussed on 2024-12-12 System SemConv WG meeting. We can take care of this as part of defining processes as Entities. The place to put this information down as YAML is not ready yet but we can work on this in parallel.

@dmitryax has a doc where we are working on this. He will create an issue for defining processes as an entitity and mark this as blocked on that issue.

mx-psi:
> @/dmitryax has a doc where we are working on this. He will create an issue for defining processes as an entitity and mark this as blocked on that issue.

See #1751, marking this as blocked

ChrsMark:
I think that's also related to https://github.com/open-telemetry/semantic-conventions/issues/626 which was discussed in the past.

osullivandonal:
Just looking at this, correct me if I am wrong, we need to update our process common attributes and make some required. For example, this would mean making `process.executable.name` required instead of conditionally_required. The idea is we want to make a few attributes required that are consistently available across all platforms, for example, `process.executable.name` should be available across, Linux, Windows, and Darwin.

[...]

PR #1588, opened 2024-11-18
Add conventions for CLI Spans

### LINKED ISSUE #1577: Semconv for cli command executions
### Area(s)

area:new

Would be nice to formalize semantic conventions for cli command execution regarding span name, status and attributes.

For the attributes we can reuse something from [Process](https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/).

_No response_

### Additional context

I'm implementing an instrumentation for the click library:

```python
# hello.py
import click

@click.command()
def hello():
    click.echo('Hello World!')

if __name__ == '__main__':
    hello()
```
and you call them as `python hello.py`

This is also the underlying library used by flask where you can run the server itself, e.g. `flask run` or implement your own command, e.g. `flask createuser myusername`.

xrmx:
Fixes #1577

## Changes

Start implements semantic conventions for CLI apps.


model/cli/spans.yaml
```diff
@@ -0,0 +1,14 @@
+groups:
+  - id: trace.cli.common_attributes
+    type: attribute_group
+    stability: experimental
+    brief: >
+      Describes CLI (Command Line Interfaces) programs span.
+    attributes:
+      - ref: process.command_line
```
xrmx, line 8: I guess it's not fine to reference other namespaces?

```
  × The semantic convention spec is invalid (path_or_url: "/home/weaver/
  │ source/cli/spans.yaml"). groups[0].attributes: data did not match any
  │ variant of untagged enum AttributeSpec at line 8 column 7
```

model/cli/spans.yaml
```diff
@@ -0,0 +1,14 @@
+groups:
+  - id: span.cli
+    type: attribute_group
+    stability: experimental
+    brief: >
+      Describes CLI (Command Line Interfaces) programs span.
+    attributes:
+      - ref: process.command_line
```
xrmx, line 8: Should I reference all attributes that are considered alternative in the process semconv?

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Status**: [Experimental][DocumentStatus]
+
+This document defines semantic conventions to apply when instrumenting CLI programs.
+operations to attributes on a Span.
+
+The **span name** SHOULD be of the format `{process.executable.name}`. In case this is
+not informative enough because of interpreters calling other files, e.g. `python script.py` or in
+cases of subcommands, e.g. `flask run mycommand` instrumentation MAY provide a more
+descriptive span name of the format `process.command_line` if appropriate.
```
xrmx, line 15:
```suggestion
descriptive span name of the format `{process.command_line}` if appropriate.
```

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,38 @@
...
+
+> **Warning**
+> The `process.command_line` value can have high cardinality.
```
xrmx, line 18:
```suggestion
> The `process.command_line` value can have high cardinality in case of parameters passed by the user. It is recommended to strip parameters if there is enough domain knowledge for being able to do so.
```

tammy-baylis-swi: Good call, I think this is important. Could this be added to table below and as an update to the original `process` [registry](https://github.com/open-telemetry/semantic-conventions/blob/25e0baecd6393d10c1b87a7074520985f0da2fac/model/process/registry.yaml#L87-L95)?

model/cli/spans.yaml
tammy-baylis-swi: I'd prefer references to all of them.

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,38 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: CLI
+--->
+
+# Semantic Conventions for CLI (Command Line Interface) programs
+
+**Status**: [Experimental][DocumentStatus]
+
+This document defines semantic conventions to apply when instrumenting CLI programs.
+operations to attributes on a Span.
```
tammy-baylis-swi, line 10: Nit typo

```suggestion
This document defines semantic conventions to apply when instrumenting CLI programs'
operations to attributes on a Span.
```

docs/cli/cli-spans.md
codefromthecrypt: should this doc link to the process registry or is it assumed? personally, I didn't know there was one until you mentioned it..

codefromthecrypt: actually it is mentioned below, possibly handy to say up front that "a CLI span is defined in terms of process attributes" with a link to that?

codefromthecrypt (approved):
looks good to me. Only minor point is if it is worthwhile recording `process.executable.path` as sometimes that can disambiguate what's being run.

model/cli/spans.yaml
```diff
@@ -0,0 +1,14 @@
+groups:
+  - id: span.cli
+    type: attribute_group
+    stability: experimental
```
lmolkova, line 4: please specify span_kind

model/cli/spans.yaml
lmolkova, line 8: what about executable name? pid? owner?

model/cli/spans.yaml
```diff
@@ -0,0 +1,14 @@
...
+    brief: >
+      Describes CLI (Command Line Interfaces) programs span.
+    attributes:
+      - ref: process.command_line
+        requirement_level: recommended
+      - ref: process.command_args
+        requirement_level: recommended
+      - ref: process.exit.code
```
lmolkova, line 12: we also need `error.type` if exist code is not 0.

model/cli/spans.yaml
lmolkova: We should pick a reasonable set - what we do in process attributes today is really an anti-pattern  - https://github.com/open-telemetry/semantic-conventions/issues/864.

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,38 @@
...
+|---|---|---|---|---|---|
+| [`process.command_args`](/docs/attributes-registry/process.md) | string[] | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `["cmd/otecol", "--config=config.yaml"]` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`process.command_line`](/docs/attributes-registry/process.md) | string | The full command used to launch the process as a single string representing the full command. On Windows, can be set to the result of `GetCommandLineW`. Do not set this if you have to assemble it just for monitoring; use `process.command_args` instead. | `C:\cmd\otecol --config="my directory\config.yaml"` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 30: the command line and command args are the same, but in a different format. Why specify both?

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,38 @@
...
+
+# Semantic Conventions for CLI (Command Line Interface) programs
+
+**Status**: [Experimental][DocumentStatus]
+
+This document defines semantic conventions to apply when instrumenting CLI programs.
+operations to attributes on a Span.
+
+The **span name** SHOULD be of the format `{process.executable.name}`. In case this is
+not informative enough because of interpreters calling other files, e.g. `python script.py` or in
```
lmolkova, line 13: how would generic instrumentation know whether an arbitrary string is informative or not?

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,38 @@
...
+cases of subcommands, e.g. `flask run mycommand` instrumentation MAY provide a more
+descriptive span name of the format `{process.command_line}` if appropriate.
+
+> **Warning**
+> The `process.command_line` value can have high cardinality.
+
```
lmolkova, line 19: let's mention that span status should be set to error when the exit code is not `0`

docs/cli/cli-spans.md
codefromthecrypt: The history of this was not generic, rather in context of click which would have a better value I think. How might we clarify this better? https://github.com/open-telemetry/semantic-conventions/issues/1577

model/cli/spans.yaml
lzchen: @lmolkova 

When you say reasonable set, do you mean find a set of attributes and mark them as required?

lzchen: How many attributes do we need to include before CLI semconv becomes equivalent to process semconv? I was under the assumption we would only extract the "command" related attributes.

lmolkova: process semconv describes resource attributes - not spans. Here I believe we're trying to specify which attributes would be populated on spans, what's the span name and kind is.

And why would someone not want process pid on the corresponding span? 

Also, maybe the answer is that process attributes are set as resource attributes on that span and then here we'd just add a link to process resource attributes saying that they should be set?

lmolkova: I mean that we have an extensive list of all `process` attributes defined in the registry - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/process.md

We need to pick attributes that sufficiently describe the process span and set requirement levels. 

If we want to start small - I have no objections. On the first sight, I'd add the following attributes to the span:
- `process.executable.name` as required - it could be in the command line, but it's something would want to query on and group by
- `process.executable.path` - required or recommended
- `process.pid` - required (if it's always possible to obtain it)
- `process.exit.code` - required (or conditionally_required if not 0)
- `process.title` - recommended this should probably go the the span name
- `process.command_args` (array) or another one that describes command line

There are other applicable attributes, but they can be added later if anyone  needs them.

docs/cli/cli-spans.md
xrmx: I think the second line was a copy and paste error, reworked the text a bit.

docs/cli/cli-spans.md
xrmx: Removed the stringified one, most languages follow C argv like interface.

docs/cli/cli-spans.md
xrmx: Done

docs/cli/cli-spans.md
xrmx: Yep, the context of this is building an instrumentation for a library (click) used for building CLIs. So there are cases where we have more clue :)

model/cli/spans.yaml
xrmx: Updated the list of attributes

xrmx: Added this but `process.title` which I haven't find in the registry.

model/cli/spans.yaml
```diff
@@ -0,0 +1,21 @@
+groups:
+  - id: span.cli
```
xrmx, line 2: I need help with this for pleasing policies-check

lmolkova: it  should be `span.cli.internal`

```suggestion
  - id: span.cli.internal
```

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,71 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: CLI
+--->
+
+# Semantic Conventions for CLI (Command Line Interface) programs
+
+**Status**: [Experimental][DocumentStatus]
+
+This document defines semantic conventions to apply when instrumenting CLI programs.
+A CLI span is defined in terms of [process attributes][ProcessRegistry].
```
lmolkova, line 10: I don't understand what this sentence means. Is it necessary?

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,71 @@
...
+
+The **span name** SHOULD be of the format `{process.executable.name}`. In case this is
+not informative enough because of interpreters calling other files, e.g. `python script.py` or in
+cases of subcommands, e.g. `flask run mycommand` instrumentation MAY provide a more
+descriptive span name providing a stringified version of `{process.command_args}` if appropriate.
+
+> **Warning**
+> The `process.command_args` value can have high cardinality in case of parameters passed by the user.
+> It is recommended to strip parameters if there is enough domain knowledge for being able to do so.
```
lmolkova, line 19: I'd rephrase it to

```suggestion
The span name SHOULD, when possible, match low-cardinality representation of command line. All variable parameters SHOULD be replaced with a `?` placeholder.

If instrumentation does not parse and sanitize command line, it SHOULD use set span status to `{process.executable.name}`. 

Instrumentations that have additional context about executed commands MAY use different span name format and SHOULD document it.
```

This would be a bit more prescriptive, would almost match what we do in databases + we span name should have low-ish cardinality according to the spec.

lmolkova (approved):
Left a few minor comments, LGTM otherwise!

docs/cli/cli-spans.md
xrmx: I don't feel ? would as natural as in db here.  Something like this below?

```
The span name SHOULD be set to `{process.executable.name}`. 
Instrumentations that have additional context about executed commands MAY use a different low-cardinality span name format and SHOULD document it.
```

docs/cli/cli-spans.md
xrmx: I'll drop it since all the attributes in the table link the the process registry already.

docs/cli/cli-spans.md
lmolkova: Looks good!

lzchen (approved):

docs/cli/cli-spans.md
christophe-kamphaus-jemmic, line 1: Is this related to the following OTEP?
* https://github.com/open-telemetry/oteps/pull/258
  Following this OTEP the [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification) will be updated to define span propagation when launching CLI programs.

CC @adrielp

xrmx: Nope

lmolkova (approved):

docs/cli/cli-spans.md
xrmx: > @xrmx could you please create an issue to track this future work and maybe add a TODO comment in the markdown saying that context propagation will be added in the future? Thanks!

I'm not involved with the OTEP linked above, not sure I'm the right person to track the followups :)

model/cli/spans.yaml
```diff
@@ -0,0 +1,21 @@
+groups:
+  - id: span.cli.internal
+    type: span
+    span_kind: internal
+    stability: experimental
+    brief: >
+      Describes CLI (Command Line Interfaces) programs span.
```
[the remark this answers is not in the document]
xrmx: What do you mean with both sides? A caller of a CLI program and a the CLI program itself? Had a chat around this  topics with @AlexanderWert today and given we added `exit.code` as required field probably must say this is not for daemon apps, just for synchronous / foreground ones.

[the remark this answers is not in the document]
lmolkova: > What do you mean with both sides? A caller of a CLI program and a the CLI program itself?

yep, this.

> I think this can work for both cases.

Then perhaps it's worth mentioning this., but also we'd need a some way to differentiate these two spans in some way. I'd argue that the caller span should probably be `client` instead of `internal` .

> probably must say this is not for daemon apps, just for synchronous / foreground ones.

we can say that it's intended for short-lived apps focused on executing one tasks (I don't think days/weeks long spans are useful no matter if they describe daemon or a regular process)

docs/cli/cli-spans.md
xrmx: Added

model/cli/spans.yaml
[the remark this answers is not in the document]
xrmx: > > What do you mean with both sides? A caller of a CLI program and a the CLI program itself?
> 
> yep, this.
> 
> > I think this can work for both cases.
> 
> Then perhaps it's worth mentioning this., but also we'd need a some way to differentiate these two spans in some way. I'd argue that the caller span should probably be `client` instead of `internal` .
> 
> > probably must say this is not for daemon apps, just for synchronous / foreground ones.
> 
> we can say that it's intended for short-lived apps focused on executing one task (I don't think days/weeks long spans are useful no matter if they describe daemon or a regular process)

Updated PR, not sure that duplicating the attributes for the spans kind was the right call.

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,118 @@
...
+| [`process.pid`](/docs/attributes-registry/process.md) | int | Process identifier (PID). | `1234` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [1] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` if and only if process.exit.code is not 0 | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`process.command_args`](/docs/attributes-registry/process.md) | string[] | All the command arguments (including the command/executable itself) as received by the process. On Linux-based systems (and some other Unixoid systems supporting procfs), can be set according to the list of null-delimited strings extracted from `proc/[pid]/cmdline`. For libc-based executables, this would be the full argv vector passed to `main`. | `["cmd/otecol", "--config=config.yaml"]` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
kuisathaverat, line 33: Do you think we should comment on anything about not including/redacting sensible content?

xrmx: This is a sensible concern, but since this text comes from the process model I think we could revise the text in another PR.

jsuereth (approved):

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,118 @@
...
+
+| Value  | Description | Stability |
+|---|---|---|
+| `_OTHER` | A fallback error value to be used when the instrumentation doesn't define a custom value. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
```
lmolkova, line 68:
```suggestion

# Client (caller) spans

```

docs/cli/cli-spans.md
```diff
@@ -0,0 +1,118 @@
...
+Span kind SHOULD be `INTERNAL` when the traced program is the callee or `CLIENT` when the caller is tracing another program.
+
+The span name SHOULD be set to `{process.executable.name}`.
+Instrumentations that have additional context about executed commands MAY use a different low-cardinality span name format and SHOULD document it.
+
+Span status SHOULD be set to `Error` if `{process.exit.code}` is not 0.
+
+<!-- TODO: context propagation https://github.com/open-telemetry/semantic-conventions/issues/1612 -->
+
```
lmolkova, line 19:
```suggestion

# Execution (callee) spans

```

maybe?

model/cli/spans.yaml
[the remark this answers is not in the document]
lmolkova: I like it! There are ways to avoid copy-paste, but it's pretty minimal here and it has no effect on produced markdown.

I just left a couple of suggestions to add headings before tables, otherwise looks great!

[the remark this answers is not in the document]
xrmx: Cool, applied both suggestions, thanks!

docs/cli/cli-spans.md
xrmx: Opened a followup issues here https://github.com/open-telemetry/semantic-conventions/issues/1620

PR #1591, opened 2024-11-19
Relax `server.port` requirement level on HTTP server spans

### LINKED ISSUE #1387: Clarify `server.address` and `server.port` in forwarded situations
### Area(s)

area:server

In https://github.com/open-telemetry/semantic-conventions/blob/bf0a2c1134f206f034408b201dbec37960ed60ec/docs/http/http-spans.md#clientserver-example-with-reverse-proxy the forwarded header may not contain enough information to determine the port if it contains `host` but not `proto`. The http semconv states that port is required if address exists.

Please clarify what to do in this situation:

1. Leave port undefined
2. Do not set port or address
3. Set port to a default value like 80 or 443
4. look for port in further fallback options like `x-forwarded-host` or `host` headers, or use port of proxied request
5. Capture the full `Forwarded` header
6. Something else

_No response_

### Additional context

While the `Forwarded` _should_ contain enough information to determine the port, it may not in every situation. In these types of unexpected situations, telemetry is more important than ever.

trask:
Fixes #1387

## Changes

Relax `server.port` requirement level on HTTP server spans from

> If `server.address` is set.

to

> If available and `server.address` is set.


lmolkova (approved):
Thanks!

jsuereth (approved):

dyladan (approved):

maryliag (approved):

PR #1592, opened 2024-11-19
GenAI: Relax `server.port` requirement level on HTTP server spans

trask:
Follow-up to #1591

Will mark ready-to-review if/when #1591 lands


trask:
Nevermind, these look like they're on client spans, so I don't think #1591 is applicable.

Issue #1537, opened 2024-10-31
K8s: write migration guide for (to-be) stable semconv


ChrsMark:
Similar to [docs/non-normative/http-migration.md](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/http-migration.md) and [docs/non-normative/db-migration.md](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/db-migration.md), and also https://github.com/open-telemetry/semantic-conventions/pull/1198.

ChrsMark:
@joaopgrassi @lmolkova would that make sense to create and merge a document already instead of keeping a long running draft PR like the https://github.com/open-telemetry/semantic-conventions/pull/1198. 

We can mark the document as non-final somehow.

This will allow us to iterate on the document and allow multiple to contribute to it.

Also we could already start referencing to it by the Collector's components, even if it's state is not marked as finalized yet.

TylerHelmuth:
I like the idea of incremental changes to a document as it will make it easier for end-users to follow along.

ChrsMark:
For Collector migration a similar plan to the `hostmetrics` receiver can be followed: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35325#issuecomment-2491669773

PR #1597, opened 2024-11-20
add k8s semconv migration non-normative doc

ChrsMark:
Part of https://github.com/open-telemetry/semantic-conventions/issues/1537

## Changes

This PR adds the non-normative k8s semconv migration docs described by https://github.com/open-telemetry/semantic-conventions/issues/1537.

So far it only includes changes from https://github.com/open-telemetry/semantic-conventions/pull/1427 but the goal is to keep adding content in this file as we go.

This PR is blocked until https://github.com/open-telemetry/semantic-conventions/pull/1427 is merged.

/cc @TylerHelmuth

### DIFF
--- a/docs/non-normative/k8s-migration.md
+++ b/docs/non-normative/k8s-migration.md
@@ -1,0 +1,56 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: K8s migration
+--->
+
+# K8s semantic conventions stability migration
+
+Due to the significant number of modifications and the extensive user base
+affected by them, existing K8s instrumentations published by
+OpenTelemetry are required to implement a migration plan that will assist users in
+transitioning to the stable K8s semantic conventions.
+
+Specifically, when existing K8s instrumentations published by OpenTelemetry are
+updated to the stable K8s semantic conventions, they:
+
+- SHOULD introduce an environment variable `OTEL_SEMCONV_STABILITY_OPT_IN` in
+  their existing major version, which accepts:
+  - `k8s` - emit the stable k8s conventions, and stop emitting
+    the old k8s conventions that the instrumentation emitted previously.
+  - `k8s/dup` - emit both the old and the stable k8s conventions,
+    allowing for a phased rollout of the stable semantic conventions.
+  - The default behavior (in the absence of one of these values) is to continue
+    emitting whatever version of the old k8s conventions the
+    instrumentation was emitting previously.
+- Need to maintain (security patching at a minimum) their existing major version
+  for at least six months after it starts emitting both sets of conventions.
+- May drop the environment variable in their next major version and emit only
+  the stable k8s conventions.
+
+<!-- toc -->
+
+- [Summary of changes](#summary-of-changes)
+  - [K8s network metrics](#k8s-network-metrics)
+
+<!-- tocstop -->
+
+## Summary of changes
+
+This section summarizes the changes made to the K8s semantic conventions
+from a range of versions. Each starting version shows all the changes required
+to bring the conventions to
+[v1.TODO (stable)](https://github.com/open-telemetry/semantic-conventions/blob/v1.TODO/docs/k8s/README.md).
+
+### K8s network metrics
+
+The K8s network metrics implemented by the Collector and specifically the
+[kubeletstats](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.112.0/receiver/kubeletstatsreceiver/documentation.md)
+receiver were introduced as semantic conventions in [v1.TODO](https://github.com/open-telemetry/semantic-conventions/blob/v1.TODO/docs/k8s/README.md).
+
+The changes in their attributes are the following:
+
+<!-- prettier-ignore-start -->
+| Old (Collector) ![changed](https://img.shields.io/badge/changed-orange?style=flat) | New                       |
+|------------------------------------------------------------------------------------|---------------------------|
+| `interface`                                                                        | `network.interface.name`  |
+| `direction`                                                                        | `network.io.direction` |
+<!-- prettier-ignore-end -->
--- a/docs/non-normative/k8s-migration.md
+++ b/docs/non-normative/k8s-migration.md
@@ -9,7 +9,7 @@ affected by them, existing K8s instrumentations published by
 OpenTelemetry are required to implement a migration plan that will assist users in
 transitioning to the stable K8s semantic conventions.
 
-Specifically, when existing K8s instrumentations published by OpenTelemetry are
+When existing K8s instrumentations published by OpenTelemetry are
 updated to the stable K8s semantic conventions, they:
 
 - SHOULD introduce an environment variable `OTEL_SEMCONV_STABILITY_OPT_IN` in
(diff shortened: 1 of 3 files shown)


docs/non-normative/k8s-migration.md
```diff
@@ -0,0 +1,56 @@
...
+- SHOULD introduce an environment variable `OTEL_SEMCONV_STABILITY_OPT_IN` in
+  their existing major version, which accepts:
+  - `k8s` - emit the stable k8s conventions, and stop emitting
+    the old k8s conventions that the instrumentation emitted previously.
+  - `k8s/dup` - emit both the old and the stable k8s conventions,
+    allowing for a phased rollout of the stable semantic conventions.
+  - The default behavior (in the absence of one of these values) is to continue
+    emitting whatever version of the old k8s conventions the
+    instrumentation was emitting previously.
+- Need to maintain (security patching at a minimum) their existing major version
```
ChrsMark, line 24: We need to figure out how this will be applied on the Collector.

docs/non-normative/k8s-migration.md
```diff
@@ -0,0 +1,64 @@
...
+  the stable k8s conventions.
+
+Specifically for the Opentelemetry Collector:
+
+The transition will happen through two different feature gates.
+One for enabling the new schema, and one for disabling the old schema. Then:
+
+- On alpha the old schema is enabled, while the new schema is disabled
+- On beta/stable the old schema is disabled, while the new is enabled
+- It is an error to disable both
```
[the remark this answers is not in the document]
ChrsMark: That would be nice! One open question here is if these should be per component/receiver or the same across all the components.

I would go with sth like `semconv.k8s.stable` and `semconv.k8s.legacy` but would love to get more suggestions here. (or `<receiver_name>.semconv.k8s.stable` if we want to include the receiver name)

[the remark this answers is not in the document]
mx-psi: I feel like they should apply to all components at the same time (at least that's what I would expect as an end user). Those names sound reasonable to me.

dmitryax (approved):

docs/non-normative/k8s-migration.md
```diff
@@ -0,0 +1,65 @@
...
+
+Specifically for the Opentelemetry Collector:
+
+The transition will happen through two different feature gates.
+One for enabling the new schema called `semconv.k8s.stable`,
+and one for disabling the old schema called `semconv.k8s.legacy`. Then:
+
+- On alpha the old schema is enabled, while the new schema is disabled
+- On beta/stable the old schema is disabled, while the new is enabled
+- It is an error to disable both
```
TylerHelmuth, line 37: similar to `k8s.dup`, will users be allows to set both to true to emit both semantics?

ChrsMark: Yes, I applied a change to the feature gates' names and enhanced the description to make it clear.

The idea is:

```
- On alpha the old schema is enabled by default (`semconv.k8s.disableLegacy` defaults to false), 
  while the new schema is disabled (`semconv.k8s.enableStable` defaults to false).
- On beta/stable the old schema is disabled (`semconv.k8s.disableLegacy` defaults to true),
  while the new is enabled (`semconv.k8s.enableStable` defaults to true).
```

To have both schemas (`k8s/dup`) we need to configure:
- On alpha: `otelcol --config=config.yaml --feature-gates=semconv.k8s.enableStable`
- On Beta:  `otelcol --config=config.yaml --feature-gates=semconv.k8s.disableLegacy`

@TylerHelmuth @mx-psi would that make sense? (I'm always confused with the feature gates' life-cycle, please correct me here if I'm wrong :))

mx-psi: To enable both the old and new set of metrics you would have to pass 
```
--config=config.yaml --feature-gates=-semconv.disableLegacy,+semconv.k8s.enableStable
```

Both feature gates would graduate from alpha to beta on the same release, and from beta to stable at the same time as well.

ChrsMark: Clarified that in place.

ChrsMark:
[v1.29.0](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.29.0) release is out (this includes the changes this current doc references).

@TylerHelmuth @mx-psi shall we proceed with this or is there anything else missing here?

docs/non-normative/k8s-migration.md
```diff
@@ -0,0 +1,68 @@
...
+The transition will happen through two different feature gates.
+One for enabling the new schema called `semconv.k8s.enableStable`,
+and one for disabling the old schema called `semconv.k8s.disableLegacy`. Then:
+
+- On alpha the old schema is enabled by default (`semconv.k8s.disableLegacy` defaults to false),
+  while the new schema is disabled by default (`semconv.k8s.enableStable` defaults to false).
+- On beta/stable the old schema is disabled by default (`semconv.k8s.disableLegacy` defaults to true),
+  while the new is enabled by default (`semconv.k8s.enableStable` defaults to true).
+- It is an error to disable both schemas
+- Both schemas can be enabled with `--feature-gates=-semconv.disableLegacy,+semconv.k8s.enableStable`.
```
TylerHelmuth, line 40: Is there any way to make the gate names both "positive"? I think it will be clearer to users to do `--feature-gates=+semconv.k8s.enableLegacy,+semconv.k8s.enableStable` if you want both.

TylerHelmuth, line 40:
```suggestion
- Both schemas can be enabled with `--feature-gates=-semconv.k8s.disableLegacy,+semconv.k8s.enableStable`.
```

ChrsMark: Won't the `semconv.k8s.enableLegacy` default to false/disabled on `alpha` 🤔 ?

TylerHelmuth: It would have to start at beta, and then be moved to alpha following these [lifecycle steps](https://github.com/open-telemetry/opentelemetry-collector/tree/main/featuregate#feature-lifecycle). It wouldn't be the standard way for us to use the gate.

If we want to go from alpha -> beta -> stable we have to use the negative name.

ChrsMark: Ah I see now! I would be fine by this.

If we conclude with this approach it should be the same in the `hostmetrics` receiver and System SemConv as well (with the respective feature gates pair).
@mx-psi @dmitryax @braydonk @roger what do think?

docs/non-normative/k8s-migration.md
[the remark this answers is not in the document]
ChrsMark: I think we have moved on here. Resolving this.

docs/non-normative/k8s-migration.md
ChrsMark: Resolving this for now. We can continue at https://github.com/open-telemetry/semantic-conventions/pull/1597#discussion_r1866107784

ChrsMark:
It would be helpful to have this soon so we can use it as base for listing the upcoming changes, i.e. in https://github.com/open-telemetry/semantic-conventions/pull/1636.

The only leftover here seems to be the names of the feature gates: https://github.com/open-telemetry/semantic-conventions/pull/1597#discussion_r1866107784

docs/non-normative/k8s-migration.md
TylerHelmuth: Thinking about this more following the standard feature-gate processor is probably better.  As long as we document how to emit both users will have a solution.

TylerHelmuth (approved):

AlexanderWert (approved):

mx-psi (approved):

rogercoll (approved):

ChrsMark:
FYI @open-telemetry/specs-semconv-maintainers we have a lot of approvals already here.

PR #1603, opened 2024-11-21
GenAI: define conventions for embeddings operations

trentm:
Refs: #1174

---

Many LLMs support an Embeddings API, for example:

- [OpenAI](https://platform.openai.com/docs/api-reference/embeddings)
- [Cohere](https://docs.cohere.com/reference/embed)
- [Voyage AI](https://docs.anthropic.com/en/docs/build-with-claude/embeddings#voyage-http-api)
- [Amazon Bedrock Titan](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-titan-embed-text.html)

This proposal defines OpenTelemetry semantic conventions to use for instrumenting Embeddings API client usage.

# Overview

A *span* will be created for Embeddings API calls. The only differences from existing chat spans are:

- A new `embeddings` value is added for attribute `gen_ai.operation.name`.
- A new `gen_ai.request.encoding_formats` attribute is defined. It is only relevant for embeddings operations.
- Many of the "recommended" span attributes do not apply to embeddings operations. Perhaps that is fine as currently defined. (**Open question:** should a `note: ...` be added or `if applicable.` added to the `brief` for those span attributes?)

No new *events* are proposed. See "Notes" below.

For *metrics*:

- The existing `gen_ai.client.token.usage` metric applies, with the note that *output* tokens do not apply for embeddings operations, so the metric would only be recorded with the `gen_ai.token.type: 'input'` attribute (as already specified).
- The existing `gen_ai.client.operation.duration` metric applies as currently specified.

# Example span

This an example embeddings operation span using the `openai` client library and the OTel JS ConsoleSpanExporter:

```
{
  traceId: '1dea787b9cd7bb895aee5bb74090610d',
  parentId: undefined,
  traceState: undefined,
  name: 'embeddings text-embedding-ada-002',
  id: '65aafaf145cf535a',
  kind: 2,
  timestamp: 1732229356238000,
  duration: 551555.417,
  attributes: {
    'gen_ai.operation.name': 'embeddings',
    'gen_ai.request.model': 'text-embedding-ada-002',
    'gen_ai.system': 'openai',
    'gen_ai.request.encoding_formats': [ 'float' ],
    'gen_ai.response.model': 'text-embedding-ada-002',
    'gen_ai.usage.input_tokens': 9
  },
  status: { code: 0 },
  events: [],
  links: []
}
```

# Notes

This section contains supporting notes and reasoning for some of the proposed values.

- A span attribute for the `dimensions` parameter (in the OpenAI API) was considered, but dropped as likely not being useful. Happy to revisit that if others know of a reasonable use case.

- A (log) event to record the *input strings* to the Embeddings API call is **not** being proposed. The reasoning is that API calls for embeddings are expected to be higher volume, and the input strings less valuable to application observability than *chat* content, so the cost-benefit ratio is much less. If a good use case is presented for optionally recording Embeddings input strings, then this can be revisited.

- Recording Embeddings API response *vectors* in telemetry is **not** being proposed. Vectors are large and would not be useful for observability.

- The operation name `embeddings` was selected. Other possible options considered:

    1. `embeddings`
    2. `embedding`
    3. `embed`

    My inclination is `embeddings` to match the (OpenAI) API name: `embeddings.create`.
    [Cohere's API](https://docs.cohere.com/reference/embed) name is `embed`.
    LangTrace uses `embed`.
    Current semconv values are `chat` (e.g. for `openai.chat.completions.create()`) and `text_completion` (for the deprecated `openai.chat.create()`).

- The Cohere and Anthropic APIs have a request attribute `input_type`, for creating embeddings for inputs other than text. This might be worth considering adding as well. I have not currently proposed this because I have only prototyped with OpenAI.


model/gen-ai/registry.yaml
```diff
@@ -117,6 +127,8 @@ groups:
         type: int
         brief: The number of tokens used in the GenAI response (completion).
         examples: [180]
+        note: >
+          If applicable. For example, embeddings operations to do use output tokens.
```
trentm, line 131: Reviewer note: In the PR description I added an "Open question" about whether it would be good/helpful to add "If applicable" notes (or additions to the `brief`) for the span attributes that do not apply to embeddings operations. This is one such note to feel it out. Thoughts?

xrmx (approved):

model/gen-ai/registry.yaml
xrmx: Does this "to do" should be "do not"?

model/gen-ai/registry.yaml
```diff
@@ -90,6 +90,16 @@ groups:
...
         examples: [0.1]
+      - id: gen_ai.request.encoding_formats
+        stability: experimental
+        type: string[]
+        brief: The encoding formats requested in an embeddings operation, if specified.
+        examples:
+          - ['base64']
+          - ['float', 'binary']
+        note: >
+          In some APIs the encoding formats are called embedding types.
```
xrmx, line 101: Maybe these `APIs` should be `GenAI systems` to match the attribute?

trentm: Thanks. commit 2350ab0

docs/attributes-registry/gen-ai.md
```diff
@@ -45,13 +48,16 @@ is set to `openai` based on the instrumentation's best knowledge.
 For custom model, a custom friendly name SHOULD be used.
 If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
 
+**[4] `gen_ai.usage.output_tokens`:** If applicable. For example, embeddings operations to do use output tokens.
```
xrmx, line 51:
```suggestion
**[4] `gen_ai.usage.output_tokens`:** If applicable. For example, embeddings operations do not use output tokens.
```

CONTRIBUTING.md
```diff
@@ -114,7 +114,7 @@ The YAML (model definition) and Markdown (documentation) files are organized in
 ```
 
 All attributes must be defined in the folder matching their root namespace under
-`/{root-namespace}/*registry.yaml` file.
+`/model/{root-namespace}/registry.yaml` file.
```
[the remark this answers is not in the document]
lmolkova: thanks, let's keep it!

model/gen-ai/registry.yaml
lmolkova: "if applicable" usually goes on the `requirement_level" when referencing an attribute. No need to add it on the registry level - all attributes are used when they are appliable

lmolkova (approved):
Left a few minor comments, looks great otherwise!

model/gen-ai/registry.yaml
trentm: @lmolkova 

1. Do you mean like this?

https://github.com/open-telemetry/semantic-conventions/blob/5517079850f4a1bedfbb376354e2a2dfe74edc86/model/process/registry.yaml#L107-L118

That's the only example I've seen of `requirement_level` with both some explanatory text and a key other than `conditionally_required`.

See commit d79bf33.

2. Do you think there is value adding this kind of "if applicable for the particular GenAI operation" to most of the attributes? Or does that end up being too much table node spam?

docs/attributes-registry/gen-ai.md
trentm: commit b053fa1. Thanks.

model/gen-ai/spans.yaml
```diff
@@ -26,6 +26,8 @@ groups:
         requirement_level: recommended
       - ref: gen_ai.request.presence_penalty
         requirement_level: recommended
```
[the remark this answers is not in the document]
trentm: I would be fine with either option. I went with the latter in commit cdb9e09.

lmolkova (approved):
Looks great!

trentm:
^^ the check failure is:

```
make: *** [Makefile:72: markdown-link-check] Error 1
```

I'm guessing it was a fluke server error response from that site. The URL exists for me.  The check passes for me locally as well:

```
% make markdown-link-check
semantic-conventions2@ /Users/trentm/tm/semantic-conventions2
└── markdown-link-check@3.11.2
```

I don't have permissions to re-run that workflow run.

lmolkova:
the link check is not required and oracle blog is the usual offender. We need another approval to merge though. @open-telemetry/semconv-genai-approvers ptal!

karthikscale3 (approved):
Thanks for starting this

alizenhom (approved):

AlexanderWert (approved):

axiomofjoy:
Just wanted to add some thoughts on observability workflows for embeddings since they are important to our product at Arize. We use both embedding vectors and their associated text for debugging and troubleshooting workflows. Because embedding vectors map semantically similar content to nearby vectors, you can use them for observability workflows such as:

- clustering to find semantically meaningful groups of user queries, and then measuring performance/ computing evaluation metrics on those clusters to gain granular understanding of your application performance on a meaningful subset of data
- embedding-based similarity search to find semantically similar examples to a chosen example, e.g., for debugging or curating datasets for fine-tuning
- understanding how query and corpus distributions differ (e.g., to identify out-of-distribution queries that likely do not have answers in the embedded corpus)

These workflows require both embedding vectors and associated content (text, image, etc.), which we include as semantic conventions in the OpenInference spec.

I've included a few resources below for additional context.

https://github.com/user-attachments/assets/1125b701-c04e-4672-b308-a54029860e88

- https://www.youtube.com/watch?v=K532ZClP-xQ
- https://www.youtube.com/watch?v=JaFt1U5rxs4

Issue #1453, opened 2024-10-05
Discuss adding vcs.repository.name as an additional attribute


adrielp:

christophe-kamphaus-jemmic:
From discussion https://github.com/open-telemetry/semantic-conventions/pull/1383#discussion_r1742690825

adrielp:
Resolved through #1608

PR #1608, opened 2024-11-25
[vcs] Adds `vcs.repository.name` attribute and updates `vcs.repository.url.full` to be represented consistently.

### LINKED ISSUE #1254: Adjust vcs.repository.url.full to be represented consistently
# Overview

During the pull request feedback, there was a [mention of a need for having consistency in the `url.full`](https://github.com/open-telemetry/semantic-conventions/pull/1075#discussion_r1676436870) so that end users can consistently query across telemetry.

We should provide additional direction, and (or) enable vendor extension to address this.

adrielp:
Fixes #1254 and #1453

## Changes

Adds `vcs.repository.name` attribute and updates `vcs.repository.url.full` to be represented consistently.

### DIFF
--- a/.chloggen/vcs-1254.yaml
+++ b/.chloggen/vcs-1254.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: vcs
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Adds `vcs.repository.name` attribute to registry and update `vcs.repository.url.full` description for consistent representation.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [1254, 1453]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/model/vcs/registry.yaml
+++ b/model/vcs/registry.yaml
@@ -12,10 +12,31 @@ groups:
           The [URL](https://wikipedia.org/wiki/URL) of the repository
           providing the complete address in order to locate and identify the
           repository.
+        note: |
+          The URL SHOULD only be an HTTPS absolute URL and SHOULD NOT include
+          `.git` appended on the end nor be an SSH URL.
         examples:
           [
             "https://github.com/opentelemetry/open-telemetry-collector-contrib",
             "https://gitlab.com/my-org/my-project/my-projects-project/repo",
+          ]
+      - id: vcs.repository.name
+        type: string
+        stability: experimental
+        brief: >
+          The human readable name of the repository. It SHOULD NOT include any
+          additional identifier like Group/SubGroup in GitLab or organization
+          in GitHub.
+        note: |
+          Due to it only being the name, it can clash with forks of the same
+          repository if collecting telemetry across multiple orgs or groups in
+          the same backends. It is recommended to include
+          `vcs.repository.url.full` alongside this attribute to prevent
+          clashing.
+        examples:
+          [
+            "semantic-conventios",
+            "my-cool-repo",
           ]
       - id: vcs.ref.base.name
         type: string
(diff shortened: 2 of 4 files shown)


model/vcs/registry.yaml
```diff
@@ -12,11 +12,32 @@ groups:
           The [URL](https://wikipedia.org/wiki/URL) of the repository
           providing the complete address in order to locate and identify the
           repository.
+        note: |
+          The URL SHOULD only be an HTTPS absolute URL and SHOULD NOT include
+          `.git` appended on the end nor be an SSH URL.
```
christophe-kamphaus-jemmic, line 17: Should we add an exception for ssh if HTTPS access is not available?
Are there repo names ending with `.git`? Should we specify that the `.git` refers only to the optional suffix?
```suggestion
        note: |
          The URL SHOULD only be an HTTPS absolute URL and SHOULD NOT include
          any optional `.git` appended on the end nor be an SSH URL (unless ssh is the only protocol to access the repository).
```
Maybe it's better to split this into two separate sentences for simplicity.

model/vcs/registry.yaml
```diff
@@ -12,11 +12,32 @@ groups:
...
         examples:
           [
             "https://github.com/opentelemetry/open-telemetry-collector-contrib",
             "https://gitlab.com/my-org/my-project/my-projects-project/repo",
           ]
+      - id: vcs.repository.name
```
christophe-kamphaus-jemmic, line 23: Should we add this attribute as an optional attribute for any VCS metric which have a `vcs.repository.url.full` attribute?

model/vcs/registry.yaml
christophe-kamphaus-jemmic: Remark from Josh:
we should specify that this note refers to git repositories.
Other VCS types might use other protocols.

adrielp: Yes, this URL is only for accessing the repository through a browser. This isn't the URL to clone the repository over any type of CLI utility. Regardless of the vendor, or technology (git / svn / mercurial) AFAIK, all systems have URLs of which can be hit through the browser that take you to the repository.

model/vcs/registry.yaml
adrielp: Yup, I think so. Wasn't planning on changing the metrics though here. Was only tackling the registry to keep things small.

model/vcs/registry.yaml
```diff
@@ -12,11 +12,32 @@ groups:
           The [URL](https://wikipedia.org/wiki/URL) of the repository
           providing the complete address in order to locate and identify the
           repository.
+        note: |
+          The URL SHOULD only be an HTTPS absolute URL and SHOULD NOT include
```
lmolkova, line 16: super-nit: 

```suggestion
          The URL SHOULD only be an HTTP(S) absolute URL and SHOULD NOT include
```

model/vcs/registry.yaml
lmolkova: It's ok to follow up, but also referencing the attribute in this PR would be quite a small change

model/vcs/registry.yaml
```diff
@@ -12,11 +12,32 @@ groups:
...
+          in GitHub.
+        note: |
+          Due to it only being the name, it can clash with forks of the same
+          repository if collecting telemetry across multiple orgs or groups in
+          the same backends. It is recommended to include
+          `vcs.repository.url.full` alongside this attribute to prevent
+          clashing.
+        examples:
+          [
+            "semantic-conventios",
```
lmolkova, line 38:
```suggestion
            "semantic-conventions",
```

lmolkova (approved):

model/vcs/registry.yaml
adrielp: I've went ahead and adjusted the metrics to include it as recommended.

model/vcs/registry.yaml
```diff
@@ -9,12 +9,9 @@ groups:
...
         brief: >
-          The [URL](https://wikipedia.org/wiki/URL) of the repository
-          providing the complete address in order to locate and identify the
-          repository.
-        note: |
-          The URL SHOULD only be an HTTPS absolute URL and SHOULD NOT include
-          `.git` appended on the end nor be an SSH URL.
+          The [canonical URL](https://support.google.com/webmasters/answer/10347851?hl=en#:~:text=A%20canonical%20URL%20is%20the,Google%20chooses%20one%20as%20canonical.)
+          of the repository providing the complete HTTP(S) address in order to
+          locate and identify the repository through a browser.
```
christophe-kamphaus-jemmic, line 14: We can still give guidance on what we consider the canonical URL. Ie. without the optional `.git` suffix.

christophe-kamphaus-jemmic (approved):

carlosalberto (approved):

.chloggen/vcs-1254.yaml
```diff
@@ -0,0 +1,25 @@
...
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: vcs
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: |
+  Adds `vcs.repository.name` attribute to registry and update
+  `vcs.repository.url.full` description for consistent representation. Updates
+  the VCS metrics to include `vcs.repository.name` as an optional attribute.
```
trask, line 16:
```suggestion
  the VCS metrics to include `vcs.repository.name` as a recommended attribute.
```

trask (approved):

Issue #1602, opened 2024-11-21
Mark .NET runtime metrics as stable


lmolkova:
.NET runtime metrics are part of stable (as of last week) .NET 9. 
https://github.com/dotnet/runtime/blob/v9.0.0/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/RuntimeMetrics.cs

We should mark them as stable.

_Originally posted by @lmolkova in https://github.com/open-telemetry/semantic-conventions/pull/1364#discussion_r1852771509_

lmolkova:
@open-telemetry/semconv-dotnet-approver @stevejgordon would someone be able to work on it?

stevejgordon:
@lmolkova, I can pick this up.

PR #1609, opened 2024-11-25
Mark .NET runtime metrics as stable

stevejgordon:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1602

NOTE: `dotnet.process.cpu.time` remains `experimental` due to its dependency on `cpu.time`.

* ~[ ] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.~


stevejgordon:
cc @lmolkova

model/dotnet/runtime-metrics.yaml
```diff
@@ -9,7 +9,7 @@ groups:
       This metric reports the same values as accessing [`Environment.ProcessorCount`](https://learn.microsoft.com/dotnet/api/system.environment.processorcount).
     instrument: updowncounter
     unit: "{cpu}"
-    stability: experimental
+    stability: stable
 
   - id: metric.dotnet.process.cpu.time
     type: metric
```
lmolkova, line 15: I think we should mark this metric stable too since it's de-facto stable and .NET treats it as such (along with `cpu.mode` attribute.

There are a couple of tricks on how we can do it:
1. add an exception to the policy similarly to https://github.com/open-telemetry/semantic-conventions/blob/3fa52acca3ec1530f9b88a38a29bfd876ec2205d/policies/group_stability.rego#L11-L12
2. don't define corresponding yaml. just define things in markdown similarly to https://github.com/open-telemetry/semantic-conventions/blob/ab1362c44ea6f523398aa4d8b47120580365cc81/docs/dotnet/dotnet-dns-metrics.md?plain=1#L26

I'd prefer p1 since it's limited to one attribute that I hope we'll stabilize soon along with `process` and `system` metrics.

docs/runtime/dotnet-metrics.md
```diff
@@ -4,7 +4,7 @@ linkTitle: .NET
 
 # Semantic Conventions for .NET Common Language Runtime (CLR) Metrics
 
-**Status**: [Experimental][DocumentStatus]
+**Status**: [Stable][DocumentStatus]
```
lmolkova, line 7: If we decide to keep `dotnet.process.cpu.time` experimental in this PR, we should change it to 
```suggestion
**Status**: [Stable][DocumentStatus], Unless otherwise specified.
```

lmolkova (approved):
Thank you! Left a couple of small comments on `cpu.mode` stability. 

I can follow up on the https://github.com/open-telemetry/semantic-conventions/pull/1609/files#r1857409091 if you want.

model/dotnet/runtime-metrics.yaml
stevejgordon: I've applied this as suggested. Thanks for the hint.

lmolkova (approved):
LGTM!

tarekgh (approved):

arminru (approved):

PR #1618, opened 2024-11-26
[chore] System Semantic Conventions Non-Normative Guidance

braydonk:
## Changes

This PR adds non-normative guidance from the System Semantic Conventions Working Group. This is added in a new `groups` folder in `non-normative`, and a `system` subfolder in `groups`. The docs written here were already discussed in a Google doc where we were originally collaborating on this, a link to which can be shared directly if needed.


docs/non-normative/groups/system/use-cases.md
```diff
@@ -0,0 +1,77 @@
...
+
+* Physical Memory usage and/or utilization  
+* Virtual Memory usage  
+* CPU usage and/or utilization  
+* Disk throughput  
+* Network throughput
+
+### Alert
+
+* Process stops
```
lmolkova, line 70: nit

```suggestion
* Process stops unexpectedly
```

docs/non-normative/groups/system/use-cases.md
```diff
@@ -0,0 +1,77 @@
...
+### Alert
+
+* Process stops  
+* Memory/CPU usage/utilization goes above a threshold  
+* Memory exclusively rises over a period of time (memory leak detection)
+
+### Notes
+
+* Unless the OS provides the utilization data directly, the utilization requires calculation. Process instrumentation would need to be associated with a host entity that contains data about its memory capacity for utilization metrics to be calculated.  
+* Process instrumentation can also be used as data for benchmark evaluations, collecting the data for a period of time and evaluating the timeseries to get benchmarking/overhead insights about the process
```
lmolkova, line 77: is the intent here to say that process resource utilization metrics are useful for all sorts of perf testing?

docs/non-normative/groups/system/use-cases.md
```diff
@@ -0,0 +1,77 @@
...
+
+`General Information` \= The information that should be discoverable either through the entity, metrics, or metric attributes.
+
+`Dashboard` \= The information that should be attainable through metrics to create a comprehensive dashboard.
+
+`Alerts` \= Some examples of common alerts that should be creatable with the available information.
+
+## **Host**
+
+A user should be able to monitor the health of a host, including monitoring resource consumption, unexpected errors due to resource exhaustion or malfunction of core components of a host or fleet of hosts (network stack, memory, CPU…).
```
lmolkova, line 16: > unexpected errors due to resource exhaustion

not sure if we have anything defined today and if there is anything general we can provide, but it'd be nice to have some OS network/hw/etc errors and have them on the dashboards/alerts

docs/non-normative/groups/system/use-cases.md
```diff
@@ -0,0 +1,77 @@
...
+
+* Memory utilization  
+* CPU utilization  
+* Disk utilization  
+* Disk throughput  
+* Network traffic
+
+### Alerts
+
+* VM is up
```
lmolkova, line 36:
```suggestion
* VM is down unexpectedly
```

?

docs/non-normative/groups/system/use-cases.md
```diff
@@ -0,0 +1,77 @@
...
+* Machine name  
+* ID (relevant to its context, could be a cloud provider ID or just base machine ID)  
+* OS information (platform, version, architecture, etc)  
+* Number of CPU cores  
+* Memory Capacity
+
+### Dashboard
+
+* Memory utilization  
+* CPU utilization
```
lmolkova, line 29: do we support or track GPU metrics anywhere? (totally fine to leave it out of scope of this doc)

docs/non-normative/groups/system/use-cases.md
```diff
@@ -0,0 +1,77 @@
...
+
+* The alerts in particular should be capable of being uniformly applied to an entire fleet of hosts  
+* The user may be monitoring a virtualization host i.e. VMWare or Proxmox, and the instrumentation to monitor the health of the root host and the virtual machines it's spawned can be largely the same
+
+## **Process**
+
+A user should be able to monitor the health of an arbitrary process using data provided by the OS.  
+Reasons a user may want this:
+
+1. The process they want to monitor isn't covered by more specific semconv instrumentation such as language runtime metrics, db, http, etc.
```
lmolkova, line 50: maybe something along these lines?

```suggestion
1. The process they want to monitor doesn't have in-process runtime-specific instrumentation enabled or is not instrumentable at all such as antivirus or another background process.  
```

i.e. what matters seems to be whether you have in-proc instrumentation and if you can apply it at all.

lmolkova (approved):
I really like this doc!

I don't think we have similar precedents of "why we designed it in this way" documented (the closest analogy is OTEP), but I wish we had more of these. 
We might find a better place for it within the repo over time if we'll have more docs like this.

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
+# **System Semantic Conventions: Instrumentation Design Philosophy**
+
+The System Semantic Conventions are caught in a strange dichotomy that is unique among other semconv groups. While we want to make sure we cover obvious generic use cases, monitoring system health is a very old practice with lots of different existing strategies. While we can cover the basic use cases in cross platform ways, we want to make sure that users who specialize in certain platforms aren't left in the lurch; if users aren't given recommendations for particular types of data that isn't cross-platform and universal,they may come up with their own disparate ideas for how that instrumentation should look, leading to the kind of fracturing that the semantic conventions should be in place to avoid.
```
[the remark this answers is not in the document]
braydonk: [2c335e7](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/2c335e77ab4d2e815d424af45e8081035053f196)

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
...
+The main differences in how we handle Speciialist Class instrumentation are:
+
+1. The names and resulting values will map directly to what a user would expect hunting down the information themselves. We will rarely be prescriptive in how the information should be used or how it should be broken down. For example, a metric to represent a process's cgroup would have the resulting value match exactly to what the result would be if the user called `cat /proc/PID/cgroup`.  
+2. If a metric is specific to a particular operating system, the operating system will be in the name. See Operating System in names (TODO: Link section) for more information. For example, a metric for a process's cgroup would be `process.linux.cgroup`, given that cgroups are a specific Linux kernel feature.
```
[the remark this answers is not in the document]
braydonk: [2c335e7](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/2c335e77ab4d2e815d424af45e8081035053f196)

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
...
+Some General Class examples:
+
+* Memory/CPU usage and utilization metrics  
+* General disk and network metrics  
+* Universal system/process information (names, identifiers, basic specs)
+
+Some Specialist Class examples:
+
+* Particular Linux features like special process/system information in procfs (see things like [/proc/meminfo](https://man7.org/linux/man-pages/man5/proc_meminfo.5.html) or [cgroups](https://man7.org/linux/man-pages/man7/cgroups.7.html))  
+* Particular Windows features like special process information (see things like [Windows Handles](https://learn.microsoft.com/en-us/windows/win32/sysinfo/about-handles-and-objects), [Process Working Set](https://learn.microsoft.com/en-us/windows/win32/procthread/process-working-set))
```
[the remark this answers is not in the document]
braydonk: [2c335e7](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/2c335e77ab4d2e815d424af45e8081035053f196)

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
...
+
+Thus we have decided that any instrumentation that is:
+
+1. Specific to a particular operating system  
+2. Not meant to be part of what we consider our most important general use cases
+
+Will have the Operating System name as part of the namespace. For example, there may be `process.linux`, `process.windows`, or `process.posix` names for metrics and attributes. We will not have root `linux.*`,  `windows.*`, or `posix.*` namespaces. This is because of the principle we’re trying to uphold from the [Namespaces section](https://docs.google.com/document/d/1fCHZQemLun7qh5y--seBagPQZRuy_SuQJ2SOb2n2ZzU/edit?resourcekey=0-AZdnzcIOietd-cq6sGy-IA&tab=t.uq19eerhwz7#bookmark=id.2rkukcwjxprh); we still want the instrumentation source to be represented by the root namespace of the attribute/metric. If we had OS root namespaces, different sources like `system`, `process`, etc. could get very tangled within each OS namespace, defeating the intended design philosophy.
```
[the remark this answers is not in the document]
braydonk: Yep this was leftover by accident, meant to replace all of the section links but forgot this one.

[2c335e7](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/2c335e77ab4d2e815d424af45e8081035053f196)

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
+# **System Semantic Conventions: Instrumentation Design Philosophy**
```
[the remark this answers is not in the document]
braydonk: An artifact of the Google Docs -> Markdown conversion thing I used. I think I removed all the bolds now.

[2c335e7](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/2c335e77ab4d2e815d424af45e8081035053f196)

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
...
+The System Semantic Conventions generally cover the following namespaces:
+
+* `system`  
+* `process`  
+* `host`  
+* `memory`  
+* `network`  
+* `disk`  
+* `memory`  
+* `os`
```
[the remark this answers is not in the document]
braydonk: #1623

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
...
+
+In some cases this is simply a matter of making the instrumentation's purpose as clear as possible, but there are cases where attempts to share definitions across distinct instrumentation sources poses the potential for a clash. The concrete example of a time we accepted this consequence is with `cpu.mode`; the decision was to [unify all separate instances of `*.cpu.state` attributes into one shared `cpu.mode` attributed](https://github.com/open-telemetry/semantic-conventions/issues/1139). The consequence of this is that `cpu.mode` needs to have a broad enum in its root definition, with special exemptions in each different `ref` of `cpu.mode`, since `cpu.mode` used in `process.cpu.time` vs `container.cpu.time` vs `system.cpu.time` etc. has different subsets of the overall enum values. We decided as a group to accept the consequence in this case, however it isn't something we're keen on dealing with all over system semconv, as the instrumentation ends up polluted with so many edge cases in each namespace that it defeats the purpose of sharing the attribute in the first place.
```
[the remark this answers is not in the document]
braydonk: I've read through it a bunch of times and am for some reason not seeing the diff this is suggesting :see_no_evil:

braydonk:
Did a first pass of easy comments to address, will make some time soon to go through the comments that require more thought!

docs/non-normative/groups/system/design-philosophy.md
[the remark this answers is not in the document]
braydonk: I wasn't sure an OTEP was the right place for this. [This section for the oteps README](https://github.com/open-telemetry/oteps?tab=readme-ov-file#what-changes-require-an-otep) doesn't cover semantic conventions stuff super well, but since this document only nominally affects what is owned by System Semantic Conventions WG it seems to fall under `only affects one implementation`. (I am interpreting the doc pretty liberally though)

I had brought up before that I wasn't sure where this type of doc should live, and `non-normative` was what we thought seemed like the best place. I'm open to putting it in a better place, but I wonder if this is too centralized to one group to be a full OTEP.

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Examples
+
+Some General Class examples:
+
+* Memory/CPU usage and utilization metrics  
+* General disk and network metrics  
+* Universal system/process information (names, identifiers, basic specs)
+
+Some Specialist Class examples:
```
ChrsMark, line 56: While the whole description of the rationale here is exactly how it should be, I think we miss the part of having a set of rules/guidelines/sanity-checks that would help somebody in the future to decide into which directory a metric or attribute fall into. This might not be quite easy to define because of the nature of this problem but maybe it would worth adding a section in the bottom suggesting how this kind of situations should be handled in the future.

ChrsMark (approved):
LGTM with a question/suggestion.

docs/non-normative/groups/system/design-philosophy.md
braydonk: I do have a case study below for `process.linux.cgroup`; perhaps I can adapt this to more general rules?

docs/non-normative/groups/system/design-philosophy.md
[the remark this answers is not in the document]
joaopgrassi: That's a good reason to always keep the lines at 80 maximum, so in PRs reviews the suggestions can target a single line :)

docs/non-normative/groups/system/design-philosophy.md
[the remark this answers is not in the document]
joaopgrassi: Yeah I also think this would be great as an OTEP. We did several in past for stuff like messaging so there is precedent for semconv in there, but I'm also not opposed in having it here. 

One thing we need to verify is about the website. If such documents should be included there and if so, where.

docs/non-normative/groups/system/use-cases.md
mx-psi: The `hw.` metrics have some of it https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/hardware-metrics.md#hwgpu---gpu-metrics. We haven't really addressed these much.

docs/non-normative/groups/system/use-cases.md
mx-psi: We have the [`system.network.errors`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemnetworkerrors) metric, I don't think we have anything else (I don't know if there is a way to retrieve this, libraries like psutil don't provide this for other stuff like memory or disk AFAIK). Still, I think the existing metrics cover the case of troubleshooting resource exhaustion/malfunction

docs/non-normative/groups/system/use-cases.md
```diff
@@ -0,0 +1,77 @@
...
+## **Host**
+
+A user should be able to monitor the health of a host, including monitoring resource consumption, unexpected errors due to resource exhaustion or malfunction of core components of a host or fleet of hosts (network stack, memory, CPU…).
+
+### General Information
+
+* Machine name  
+* ID (relevant to its context, could be a cloud provider ID or just base machine ID)  
+* OS information (platform, version, architecture, etc)  
+* Number of CPU cores
```
mx-psi, line 23: Maybe this can be "CPU information" instead? We have a bunch of those [here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/host.md)

mx-psi (approved):
Approving, I left a few non-blocking comments above :)

AlexanderWert (approved):

docs/non-normative/groups/system/design-philosophy.md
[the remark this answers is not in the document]
lmolkova: Ooops :) I don't remember what the problem was, so let's assume there is no problem

jsuereth (approved):
I love writing this down.

The categorization of "Two Class Design Strategy" I think we should move to general non-normative guidance for all semantic conventions to follow.

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
...
+* `host`  
+* `memory`  
+* `network`  
+* `disk`  
+* `memory`  
+* `os`
+
+Deciding on the namespace of a metric/attribute is generally informed by the following belief:
+
+**The namespace of a metric/attribute should logically map to the Operating System concept being considered the instrumentation source.**
```
trask, line 24:
```suggestion
**The namespace of a metric/attribute should logically map to the Operating System concept being considered as the instrumentation source.**
```

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
...
+
+In some cases this is simply a matter of making the instrumentation's purpose as clear as possible, but there are cases where attempts to share definitions across distinct instrumentation sources poses the potential for a clash. The concrete example of a time we accepted this consequence is with `cpu.mode`; the decision was to [unify all separate instances of `*.cpu.state` attributes into one shared `cpu.mode` attributed](https://github.com/open-telemetry/semantic-conventions/issues/1139). The consequence of this is that `cpu.mode` needs to have a broad enum in its root definition, with special exemptions in each different `ref` of `cpu.mode`, since `cpu.mode` used in `process.cpu.time` vs `container.cpu.time` vs `system.cpu.time` etc. has different subsets of the overall enum values. We decided as a group to accept the consequence in this case, however it isn't something we're keen on dealing with all over system semconv, as the instrumentation ends up polluted with so many edge cases in each namespace that it defeats the purpose of sharing the attribute in the first place.
```
trask, line 28:
```suggestion
In some cases this is simply a matter of making the instrumentation's purpose as clear as possible, but there are cases where attempts to share definitions across distinct instrumentation sources poses the potential for a clash. The concrete example of a time we accepted this consequence is with `cpu.mode`; the decision was to [unify all separate instances of `*.cpu.state` attributes into one shared `cpu.mode` attribute](https://github.com/open-telemetry/semantic-conventions/issues/1139). The consequence of this is that `cpu.mode` needs to have a broad enum in its root definition, with special exemptions in each different `ref` of `cpu.mode`, since `cpu.mode` used in `process.cpu.time` vs `container.cpu.time` vs `system.cpu.time` etc. has different subsets of the overall enum values. We decided as a group to accept the consequence in this case, however it isn't something we're keen on dealing with all over system semconv, as the instrumentation ends up polluted with so many edge cases in each namespace that it defeats the purpose of sharing the attribute in the first place.
```

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
...
+
+Monitoring operating systems is an old practice, and there are lots of ways to skin the cat within different platforms. There are lots of metrics, even in basic stuff like memory usage, where there are platform specific pieces of information that are valuable to those who really specialize in that platform.
+
+Thus we have decided that any instrumentation that is:
+
+1. Specific to a particular operating system  
+2. Not meant to be part of what we consider our most important general use cases
+
+Will have the Operating System name as part of the namespace. For example, there may be `process.linux`, `process.windows`, or `process.posix` names for metrics and attributes. We will not have root `linux.*`,  `windows.*`, or `posix.*` namespaces. This is because of the principle we’re trying to uphold from the [Namespaces section](#namespaces); we still want the instrumentation source to be represented by the root namespace of the attribute/metric. If we had OS root namespaces, different sources like `system`, `process`, etc. could get very tangled within each OS namespace, defeating the intended design philosophy.
```
trask, line 73:
```suggestion
will have the Operating System name as part of the namespace.

For example, there may be `process.linux`, `process.windows`, or `process.posix` names for metrics and attributes. We will not have root `linux.*`,  `windows.*`, or `posix.*` namespaces. This is because of the principle we’re trying to uphold from the [Namespaces section](#namespaces); we still want the instrumentation source to be represented by the root namespace of the attribute/metric. If we had OS root namespaces, different sources like `system`, `process`, etc. could get very tangled within each OS namespace, defeating the intended design philosophy.
```

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,94 @@
...
+
+### What should it be named?
+
+Since this metric falls under Specialist Class, and is a Linux-specific feature, it follows that this attribute should be named `process.linux.cgroup`.
+
+### What should the value be?
+
+Since this metric falls under Specialist Class, we don't want to be too prescriptive. A user who needs to know the `cgroup` of a process likely already has a pretty good idea of how to interpret it and use it further, and it would not be worth it for this Working Group to try and come up with every possible edge case for how it might be used. It is much simpler for this attribute as it falls under our purview to simply reflect the value from the OS, i.e. the same thing as `cat /proc/PID/cgroup`. With cgroups in particular, there is high likelihood that more specialized semconv instrumentation could be developed, particularly in support of more specialized container runtime or systemd instrumentation. It's more useful for a working group that would develop special instrumentation that leverages cgroups would be more prescriptive about how the cgroup information should be interpreted and broken down with more specificity.
```
trask, line 94:
```suggestion
Since this metric falls under Specialist Class, we don't want to be too prescriptive. A user who needs to know the `cgroup` of a process likely already has a pretty good idea of how to interpret it and use it further, and it would not be worth it for this Working Group to try and come up with every possible edge case for how it might be used. It is much simpler for this attribute as it falls under our purview to simply reflect the value from the OS, i.e. the same thing as `cat /proc/PID/cgroup`. With cgroups in particular, there is high likelihood that more specialized semconv instrumentation could be developed, particularly in support of more specialized container runtime or systemd instrumentation. It's more useful for a working group developing special instrumentation that leverages cgroups to be more prescriptive about how the cgroup information should be interpreted and broken down with more specificity.
```

docs/non-normative/groups/system/design-philosophy.md
trask: ugh, github diff is useless on these long lines 😢

all that I changed above is removing a single character: `cpu.mode` attributed -> `cpu.mode` attribute

trask (approved):

docs/non-normative/groups/system/design-philosophy.md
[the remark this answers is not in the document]
lmolkova: Let's keep it here - it's always easy to move if necessary. Spec now have OTEPs inside the repo, let's keep `non-normative` as our local alias for semconv-scoped oteps.

docs/non-normative/groups/system/design-philosophy.md
joaopgrassi: I hate these things, but I'm more and more thinking we should introduce some markdown lint style enforcement, such as long lines.

mx-psi:
What is missing for this to be merged?

braydonk:
I'm finishing up edits for the remaining open comments, will be pushing this morning.

braydonk:
I've pushed up two new commits:

[487af83](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/487af83f99a9ab8eeeb937e6cfac28638114cd8b): Addresses review comments. I will re-request review from those who still had open comments.

[01f43e9](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/01f43e943767259ebb72f637b8fe5b428c8a16b0): To address the issue with the markdown files having really long lines, I have set up Prettier to apply to these markdown files and wrap them at 80 characters. Did this in a separate commit so it wasn't too difficult to see exactly how I addressed open comments.

docs/non-normative/groups/system/use-cases.md
braydonk: Reworded in [487af83](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/487af83f99a9ab8eeeb937e6cfac28638114cd8b)

docs/non-normative/groups/system/use-cases.md
braydonk: Done in [487af83](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/487af83f99a9ab8eeeb937e6cfac28638114cd8b)

docs/non-normative/groups/system/use-cases.md
braydonk: Done in [487af83](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/487af83f99a9ab8eeeb937e6cfac28638114cd8b)

docs/non-normative/groups/system/design-philosophy.md
braydonk: Done in [487af83](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/487af83f99a9ab8eeeb937e6cfac28638114cd8b)

docs/non-normative/groups/system/design-philosophy.md
braydonk: Done in [487af83](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/487af83f99a9ab8eeeb937e6cfac28638114cd8b)

docs/non-normative/groups/system/design-philosophy.md
braydonk: Done in [487af83](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/487af83f99a9ab8eeeb937e6cfac28638114cd8b)

docs/non-normative/groups/system/design-philosophy.md
braydonk: Done in [487af83](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/487af83f99a9ab8eeeb937e6cfac28638114cd8b)

docs/non-normative/groups/system/design-philosophy.md
braydonk: Made the fix in [487af83](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/487af83f99a9ab8eeeb937e6cfac28638114cd8b)

Applying line length enforcement in [01f43e9](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/01f43e943767259ebb72f637b8fe5b428c8a16b0)

trask (approved):

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,259 @@
...
+will have the Operating System name as part of the namespace.
+
+For example, there may be `process.linux`, `process.windows`, or `process.posix`
+names for metrics and attributes. We will not have root `linux.*`, `windows.*`,
+or `posix.*` namespaces. This is because of the principle we’re trying to uphold
+from the [Namespaces section](#namespaces); we still want the instrumentation
+source to be represented by the root namespace of the attribute/metric. If we
+had OS root namespaces, different sources like `system`, `process`, etc. could
+get very tangled within each OS namespace, defeating the intended design
+philosophy.
```
lmolkova, line 257: I'm curious what would be specific problems if we gave up on the prefix and use OS name as a root?

I'm trying to document naming patterns we have in https://github.com/open-telemetry/semantic-conventions/pull/1708

and I'm actually struggling to understand what benefit the `domain` prefix brings.

lmolkova: E.g. what should I do if I want to describe a property of OS that's not about process or system? which namespace would I use?

lmolkova: a few points:

1. The thing being reported is more important than instrumentation source. I hope that current instrumentations are temporary and we'll see more and more native ones. When you think about  native instrumentations, things change. E.g. 
    - windows itself emits tons of metrics/events. Should they all be reported under `os` ? Some of them describe system performance, some describe user behavior. Do we group all of them under `os`? It seems redundant. Or just some of them? How do we decides?
    - any database has management/control plane operations that has nothing to do with DB features (auth, permissions, connection management, scaling, etc). Do we report them as `db.{mydb}.*` attributes/metrics. E.g. [`db.cassandra.paxos.prepare.duration`](https://cassandra.apache.org/doc/4.1/cassandra/operating/metrics.html#:~:text=row%20cache%20misses.-,CasPrepare,-Latency) - protocol is orthogonal to the DB features of cassandra, but the metric is still about cassandra.  Now, do we report `db.cassandra.compaction.something` (because it's about database) and `cassandra.paxos.prepare.duration` because it's about protocol?

    TL;DR: any specific system/client lib has features that belong to more than one root namespace. How instrumentation is done may change (from specific collector component to native one), but metrics we define should survive it.

2. Having common root namespace for the "General Class" makes perfect sense to me: everything common about OS goes under `system`, everything common about databases goes under `db`. 

3. I'm challenging the "Specialist Class" naming: I'm reporting different metrics related to `jvm`, everyone who cares knows that `jvm` is a runtime, `runtime` in front of it is redundant. If I care about cassandra-specific metrics, I'm no longer in DB domain - `cassandra` is the root namespace


I.e. How strong do we feel about 

https://github.com/open-telemetry/semantic-conventions/blob/01f43e943767259ebb72f637b8fe5b428c8a16b0/docs/non-normative/groups/system/design-philosophy.md?plain=1#L107-L108

>  For example, a metric for a process's cgroup would be `process.linux.cgroup`,
   given that cgroups are a specific Linux kernel feature.

Vs `linux.cgroup` ?

lmolkova: [UPDATE]:

The `process.linux.cgroup` seems to be tricky since `process` is meaningful there. cgroup is a property of the (linux) process.  I believe it's less tricky in most of other semconv cases. Let me see.

braydonk: I didn't cover well enough the point you brought up, which is that there may be a case where the platform/domain name becomes so specific and disparate in itself that it warrants being considered a unique instrumentation source/root namespace/category.

Taking the `db.cassandra` example, I imagine it would follow a process like this hypothetical scenario:
1. `db.*` may have a large number of attributes that can be used cross platform, and our guidance largely expects that they are used as much as possible. For stuff specific to Cassandra some `db.cassandra.*` stuff starts to get sprinkled into the namespace when necessary.
2. Overtime more Cassandra specialists join in and start adding more and more `db.cassandra` attributes. It begins to pollute the `db` category, such that the instrumenting Cassandra in particular has a far greater variety of things very specific to it. The fact that so many `db.cassandra.*` things are being added implies that these things are Cassandra specific already, since they should have been using generic attributes when possible.
3. A decision is made that since Cassandra-specific instrumentation has become so rich and disparate from everything else in the `db.*` namespace, it makes the most sense to split `cassandra` out into its own root namespace to represent the richness of the instrumentation available for it and how much it diverts.

Applying this thought process to the questions in your above comment:

> windows itself emits tons of metrics/events. Should they all be reported under os ?

I think the steps from above could still occur in the same order. We'd try to use as many generic attributes as possible, introducing `os.windows.*` where it is specifically necessary. If the `os.windows` category becomes so deep and disparate from everything else, then it might make sense to move it into its own namespace.

> How strong do we feel about `process.linux.cgroup` vs `linux.cgroup`

It is essentially for the reason you stated in your update, but this is a bit of a different case; whether Linux exclusive or not, the `process` will always be the instrumentation source. And where something like `os` and `db` may be more like categories than an explicit object being instrumented, `process` will always be the most important root namespace and `linux.cgroup` on its own wouldn't carry the same semantic meaning.

That being said, there is a case to be made for `cgroup` here, as `cgroup` is itself a potentially instrumentable source. It's (hopefully) only a matter of time before semantic conventions come along to instrument `cgroup`s in more detail. In that case, I wonder if that would start in `linux.cgroup` or if it would just be `cgroup` off the bat. That I'm not sure about.

---

The problem I foresee with my own ideas here is it assumes things can be moved around easier. I think the problem with my thought process is that it sort of requires foresight to make sure that there isn't the potential for us to want to extract a platform namespace before reaching the point of stability for another namespace. So all of this probably needs some more thought and discussion. Maybe when I come back from the holidays I'll have the answer (probably not but I can dream :smiley:).

lmolkova: Given that some of the things we're discussing (`db.cosmosdb`) is reasonably close to stability, I think we should try to envision as many things as we can - we might not have another chance in the next few years 🤞 

I don't think we need a rigid naming policy - i.e. `process.linux.cgroup` can be whatever makes the most sense for it.
I'll bring up the general naming to Semonv SIG after the break.

Happy holidays!

lmolkova:
@braydonk it'd be awesome to get it merged. Is there anything controversial that we can't follow up on later?

braydonk:
@lmolkova if the latest conversation is resolved, then I don't think there's anything controversial left here. I think much of that system prefix naming discussion has moved to #1711 and related PRs, so if any of that marks a meaningful shift in guidance I can follow up and update the guidance. Other than that, I'm all set for merge!

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,259 @@
...
+or through existing OS tools that they want to model as OTLP.
+
+### Case study: `process.cgroup`
+
+Relevant discussions:
+[\#1357](https://github.com/open-telemetry/semantic-conventions/issues/1357),
+[\#1364 (particular thread)](https://github.com/open-telemetry/semantic-conventions/pull/1364#discussion_r1730743509)
+
+In the `hostmetricsreceiver`, there is a Resource Attribute called
+`process.cgroup`. How should this metric be adopted in System Semantic
```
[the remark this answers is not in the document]
braydonk: Fixed in [d68c8ba](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/d68c8ba6d9e4904ced7e4245af090694c1bfa610)

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -0,0 +1,259 @@
...
+down with more specificity.
+
+## Operating System in names
+
+Relevant discussions:
+[\#1255](https://github.com/open-telemetry/semantic-conventions/issues/1255),
+[\#1364](https://github.com/open-telemetry/semantic-conventions/pull/1364#discussion_r1852465994)
+
+Monitoring operating systems is an old practice, and there are lots of ways to
+skin the cat within different platforms. There are lots of metrics, even in
```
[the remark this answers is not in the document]
braydonk: Good point, old habit of mine to use phrases like that

Adjusted that as well as some other grammar in that paragraph in [d68c8ba](https://github.com/open-telemetry/semantic-conventions/pull/1618/commits/d68c8ba6d9e4904ced7e4245af090694c1bfa610)

joaopgrassi (approved):

PR #1622, opened 2024-11-27
rpc: add aws_api to well known rpc.system values

### LINKED ISSUE #1545: Inconsistent values used for `rpc.system` by aws-sdk spans
### Area(s)

area:aws, area:rpc

## Description

AWS SDK spans tell the value to use for `rpc.system` is `aws-api` (see [here](https://github.com/open-telemetry/semantic-conventions/blob/520702f1557743fa05588999757106007fb21ba3/model/aws/sdk-spans.yml#L14))

`rpc.system` does not contain this value (see [here](https://github.com/open-telemetry/semantic-conventions/blob/520702f1557743fa05588999757106007fb21ba3/model/rpc/registry.yaml#L216-L240)).

Even worse the generated [md file](https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/docs/cloud-providers/aws-sdk.md) holds a table for the well known `rpc.system` values which adds more confusion.

## Expected Result

Consistent specification which values to use for `rpc.system` in case of AWS SDK.

v1.28.0

### Additional context

_No response_

Flarna:
Fixes #1545

## Changes

AWS SDK uses RPC conventions therefore add the value used for `rpc.system`.

Adapt the name for `aws-api` to `aws_api` to match existing conventions.


.chloggen/rpc_system_aws_api.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: rpc
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add aws_api to well known rcp.system valuse
```
dyladan, line 13:
```suggestion
note: Add aws_api to well known rcp.system values
```

dyladan (approved):

lmolkova (changes requested):
we've had this discussion on some other PR, I don't think `aws_api` fits into RPC system. Looking at its usage, it's not a remote procedure calling protocol, but a logical operation kind that describes AWS SDK API call. 
We don't use `rpc.system` to describe database or other logical operations.

I suggest to 
- try to understand the scenarios where it's important to know that span is describing AWS SDK call
- if it's really necessary, suggest a different attribute that can capture this information. Ideally it should not be specific to AWS SDK and can cover some generic cases

Issue #1377, opened 2024-09-02
proposal to make code.stacktrace (or code.* namespace) stable


SylvainJuge:
### Area(s)

area:code

### Is your change request related to a problem? Please describe.

The `code.stacktrace` is currently experimental, and we think it would make sense to make it stable.

Current known usages:

- [inferred-spans module in Java contrib](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/4eb047394f008889f7b2a0ada96677f46102576c/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/semconv/Attributes.java#L15)
- [span-stacktrace module in Java contrib](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/4eb047394f008889f7b2a0ada96677f46102576c/span-stacktrace/src/main/java/io/opentelemetry/contrib/stacktrace/StackTraceSpanProcessor.java#L23)

Making this attribute stable would allow making features that rely on it stable, relying on an experimental/incubating attribute transitively forces the feature as experimental/incubating.

### Describe the solution you'd like

Making the `code.stacktrace` attribute stable, or making the whole `code.*` namespace stable.

### Describe alternatives you've considered

Making the whole `code.*` attributes stable would be an alternative as most of those fields are not ambiguous even if some of their values are platform-specific (the `code.stacktrace` being a good example).

For now keeping the scope to `code.stacktrace` is a way to reduce the scope as an attempt to making this change easier.

### Additional context

Making this attribute stable was discussed in August 29th Java SIG Meeting where the topic was "how to make a component in java contrib stable".

### Steps and progress
- [x] renaming some attributes: https://github.com/open-telemetry/semantic-conventions/pull/1624
- [x] clarify some attributes values https://github.com/open-telemetry/semantic-conventions/issues/1677
- [x] promotion to release candidate: https://github.com/open-telemetry/semantic-conventions/pull/1599
- [ ] promotion to `stable`

SylvainJuge:
After a bit of research on trying to find usages of fields in the `code.*` namespace, I found that we have:

- OpenTelemetry PHP instrumentation already uses `code.function`, `code.namespace`, `code.filepath` and `code.lineno`: https://github.com/open-telemetry/opentelemetry-php/blob/068086347b41c6d31c545cf9b37e3a9a6c1ff59f/src/API/Instrumentation/WithSpanHandler.php#L39
[...]

PR #1624, opened 2024-11-27
rename some code attributes

SylvainJuge:
Relates to #1377

In #1599 PR discussion, it was suggested we should probably rename a few `code.*` attributes before attempting to make them stable (which is the goal of #1599).
- https://github.com/open-telemetry/semantic-conventions/pull/1599#discussion_r1854499495
- https://github.com/open-telemetry/semantic-conventions/pull/1599#discussion_r1854501036

## Changes

- rename `code.function` to `code.function.name`: it allows to add in the future other function related attributes later like `code.function.visibility` or `code.function.signature`.
- rename `code.lineno` to `code.line.number` this follows the [general recommendation to avoid abbreviations](https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/#name-abbreviation-guidelines) while the `lineno` term is common, avoiding ambiguity is probably better here, also the `code.line` might be ambiguous as it could be interpreted as "the line of code".
- rename `code.column` to `code.column.number` for consistency/symmetry with `code.line.number`.
- EDIT: rename `code.filepath` to `code.file.path` for consistency with [`file.path`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/file.md#file-path) ([comment](https://github.com/open-telemetry/semantic-conventions/pull/1624#issuecomment-2533191759))

Existing usages of those attributes would be impacted (from a GH search, which means some might have been missed due to indirections from the respective generated semconv files):

- opentelemetry-php 
  - `code.lineno` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-php%20%22code.lineno%22&type=code)
  - `code.function` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-php+%22code.function%22&type=code)

- opentelemetry-rust
  - `code.lineno` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-rust+%22code.lineno%22&type=code)

- opentelemetry-java-instrumentation
  - `code.lineno` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-java-instrumentation%20%22code.lineno%22&type=code)
  - `code.function` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-java-instrumentation+%22code.function%22&type=code)

- opentelemetry-js-contrib
  - `code.lineno` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-js-contrib%20%22code.lineno%22&type=code)
  - `code.function` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-js-contrib%20%22code.lineno%22&type=code)

- opentelemetry-python
  - `code.lineno` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python%20%22code.lineno%22&type=code)
  - `code.function` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python+%22code.function%22&type=code)

- opentelemetry-cpp-contrib
  - `code.function` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-cpp-contrib+%22code.function%22&type=code)
  - `code.lineno` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-cpp-contrib+%22code.lineno%22&type=code)

- opentelemetry-php-contrib
  - `code.lineno` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-php-contrib%20%22code.lineno%22&type=code)
  - `code.function` [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-php-contrib%20%22code.function%22&type=code)

I haven't found any usage of `code.column` which means renaming here should not have any expected impact.

EDIT Renaming `code.filepath` to `code.file.path` has been added later in the discussion and potential impacts haven't been evaluated.


jsuereth:
> One problem is that renaming with a prefix is not allowed by policy, as a result I get the following error when trying to validate policy:
> 
> ```
> Violation: Attribute 'code.column' name is used as a namespace in the following attributes '["code.column.number"]'.
>   - Category         : naming_collision
>   - Type             : semconv_attribute
>   - SemConv group    :
>   - SemConv attribute: code.column
>   - Provenance: /home/weaver/source
> 
> Violation: Attribute 'code.function' name is used as a namespace in the following attributes '["code.function.name"]'.
>   - Category         : naming_collision
>   - Type             : semconv_attribute
>   - SemConv group    :
>   - SemConv attribute: code.function
>   - Provenance: /home/weaver/source
> ```
> 
> I don't know where does this restriction comes from, is it related to the generated per-language bindings ? From what I've seen so far they are mostly used as strings, so the limitations would be more on the processing/storage side.

We can add exceptions here: https://github.com/open-telemetry/semantic-conventions/blob/main/policies/attribute_name_collisions.rego#L83

This rule is in place to avoid conflicts if we ever "un-flatten" attributes.  E.g. if an attribute of `x` can have an "AnyValue" with field `y` and this means the same as `x.y`.

We should ensure we never send the attribute as a namespace and the new attribute in the same signal.

@lmolkova will update policy to avoid this warning *if* the namespace attribute becomes deprecated.

SylvainJuge:
Thanks to #1642 this is now unblocked and all the checks are now passing, it should now be ready for review.

adrielp (approved):

schema-next.yaml
```diff
@@ -5,6 +5,12 @@ versions:
   1.29.0:
     all:
       changes:
+        # https://github.com/open-telemetry/semantic-conventions/pull/1624
+        - rename_attributes:
+            attribute_map:
+              code.function: code.function.name
+              code.lineno: code.line.number
+              code.column: code.column.number
```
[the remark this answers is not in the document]
SylvainJuge: Good catch, done in 22c0413

lmolkova (approved):

.chloggen/code-rename.yaml
trask, line 1: I'd recommend leaving this PR open until December 12

I'd like to give enough community notice, especially since we aren't planning to provide any migration plan

I will post today in #otel-semantic-conventions, #otel-specification, and #otel-maintainers slack channels

and I'll add it to next week's meeting agendas for those SIGs to make sure everyone has a chance to voice any concerns before we merge it

trask (approved):

Kielek (approved):

.chloggen/code-rename.yaml
SylvainJuge: Is there a general strategy to provide a migration plan when renaming attributes beyond communication ?

Once this is merged and we know it will be included in the next semconv release, I think that I can open issues in the known impacted repositories that have been listed in the description. Please :+1: or reply to this comment if you think it's the right approach and I'll create issues.

Here we deprecate existing attributes, so I think that only the repositories where deprecation warnings are not permitted opentelemetry-java-instrumentation would require to deal with that in the semconv update PR.

SylvainJuge: From the related announcement on slack I got what "migration plan" here means: using `OTEL_SEMCONV_STABILITY_OPT_IN` configuration and keeping it compatible for a while, likely waiting for the next major release of each component before finally doing the change.

AlexanderWert (approved):

docs/attributes-registry/code.md
```diff
@@ -6,15 +6,28 @@
...
-| <a id="code-function" href="#code-function">`code.function`</a> | string | The method or function name, or equivalent (usually rightmost part of the code unit's name). | `serveRequest` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| <a id="code-lineno" href="#code-lineno">`code.lineno`</a> | int | The line number in `code.filepath` best representing the operation. It SHOULD point within the code unit named in `code.function`. | `42` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| <a id="code-namespace" href="#code-namespace">`code.namespace`</a> | string | The "namespace" within which `code.function` is defined. Usually the qualified class or module name, such that `code.namespace` + some separator + `code.function` form a unique identifier for the code unit. | `com.example.MyHttpService` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="code-function-name" href="#code-function-name">`code.function.name`</a> | string | The method or function name, or equivalent (usually rightmost part of the code unit's name). | `serveRequest` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
pellared, line 20: The description says that it can be also something "equivalent".

Therefore, [`otelslog`](https://pkg.go.dev/go.opentelemetry.io/contrib/bridges/otelslog) and [`otelzap`](https://pkg.go.dev/go.opentelemetry.io/contrib/bridges/otelzap) logging bridges set the value to a fully-qualified name e.g. `go.opentelemetry.io/contrib/bridges/otelslog.TestSLogHandler`. Given the attribute was also called `code.function` if felt pretty acceptable and compliant (maybe we were wrong).

We could set `code.function` to `TestSLogHandler` and add `code.namespace` with `go.opentelemetry.io/contrib/bridges/otelslog`. However, it would increase the performance overhead (as the Go APIs return and operate on fully-qualified names) and felt unnecessary. Should we split it anyway? Should there be a separate attribute for a fully qualified name? Or is it OK to use `code.function.name` for fully-qualified names?

I think we should have a clarification for such use case if we are planning to make breaking changes in these attributes.

pellared (changes requested):

docs/attributes-registry/code.md
pellared: CC @Jesse0Michael @suniastar @dmathieu @MrAlias @scorpionknifes

SylvainJuge: We should probably better clarify with examples for each platform/language what are the expected values for `code.namespace`, `code.function.name`. Leaving this for interpretation is not a great way to ensure consistency in the future. The downside of doing this though is that we will need input from lots of people.

I _think_ the original intent (this was contributed from ECS) is being able to search/group things by attributes, so it means that having things split by default is easier. Without this, it means having to do a prefix/suffix search and grouping which might not be available for all backends, also it means having some per-platform/language knowledge on the structure of a string attribute, in other words it's stringly-typed.

What orders of performance overhead do you expect to see here ? I'm not familiar with Go, but in Java it would mostly be splitting a string.

trask: this reminds me of our choice around capturing `url.full` on http client (https://github.com/open-telemetry/opentelemetry-specification/pull/2469). we decided it was very helpful for consumers of the data to have consistency instead of having some http clients reporting `url.full` and some reporting `url.scheme`, `url.path`, `url.query`, even though this can incur overhead in http client instrumentations that don't have direct access to `url.full` and have to construct it from `url.scheme`, `url.path`, `url.query`.

I think here it would be better for consumers of the data if we capture `code.namespace` and `code.function` consistently (either always splitting them or always combining them).

are folks aware of other instrumentation besides the Go logging bridges that capture `code.function` in this combined form?

if we find it's a limited set of instrumentation, and we decide to recommend splitting, there are still a couple of options if the Go logging bridges can't accept the splitting overhead:

* document in the Go logging bridges that they don't follow the `code.*` semantic conventions and that for performance reasons they report the combined value in `code.function.name` (or `code.namespace`?), and maybe give users an option if they prefer for the logging bridge to do the splitting
* add a new `code.*` semantic attribute that represents the combined value, though my initial thought is this may be less desirable for similar reasons to [open-telemetry/opentelemetry-specification#2469](https://github.com/open-telemetry/opentelemetry-specification/pull/2469) in that it multiplies the combinations backends need to deal with

pellared: I will do some benchmarks for Go. The overhead may be acceptable for us especially that it is an opt-in feature) but I wanted to call it out as the same concern could be valid for other languages.

pellared: > if we find it's a limited set of instrumentation,

It might be applicable for all Go instrumentations and logging bridges.

For Go we can simply decide to not set `code.namespace` and set the `code.function.name` with the "full function name" with the argument that this is how the Go runtime returns the function name:
- https://pkg.go.dev/runtime#Func.Name
- https://go.dev/play/p/As0a5RYy0sJ

This is also how existing Go logging libraries (e.g. https://pkg.go.dev/go.uber.org/zap and https://pkg.go.dev/log/slog from the standard library) report the function names.

I think this is how many Go end-users would like the logging bridges to report the "function names".

However, I am worried that setting it as  `code.function.name` may be confusing.

What is more to add, the [Go Specification](https://go.dev/ref/spec#Function_declarations) also defines a function name which is what you would expect from a function name.

I want also to call out the doc comment from https://pkg.go.dev/log/slog#Source:

```go
type Source struct {
	// Function is the package path-qualified function name containing the
	// source line. If non-empty, this string uniquely identifies a single
	// function in the program. This may be the empty string if not known.
	Function [string](https://pkg.go.dev/builtin#string) `json:"function"`
```



> add a new code.* semantic attribute that represents the combined value

This is what I would propose. Notice that this is just an "attribute registry" document and not actual instrumentation library semantic convention. The logging bridges may be configurable whether they emit `code.function.name` + `code.registry` or `code.function.full-name`.

pellared (approved):

docs/attributes-registry/code.md
SylvainJuge: Thanks for doing go benchmarks @pellared, for once it seems that consistency does not cost too much here.

As the initial goal of this PR is just the naming change, I suggest to follow-up with an issue/PR for providing per-language examples for those attributes in order to avoid any ambiguity in the future.

trask:
Probably we should also rename `code.filepath` to `code.file.path`, for the reasons above and also to match [`file.path`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/file.md#file-path)

pellared:
Today, I encountered another unclarity related to `code.function.name` and `code.namepsace`. How anonymous functions (lambdas) should be handled?

For instance, in Go we can have an anonymous function in `go.opentelemetry.io/contrib/bridges/otelslog.TestSLogHandler` function. The Go runtime will report its name e.g. as `go.opentelemetry.io/contrib/bridges/otelslog.TestSLogHandler.func7`. I decided to go with `code.function.name=func7` and `code.namespace=go.opentelemetry.io/contrib/bridges/otelslog.TestSLogHandler`.

SylvainJuge:
> Probably we should also rename `code.filepath` to `code.file.path`, for the reasons above and also to match [`file.path`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/file.md#file-path)

Done in 9303339, I have also updated this PR description.

.chloggen/code-rename.yaml
lmolkova: I believe the consensus is that `code` namespace is not used wide enough to justify the migration plan (see notes https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit?tab=t.0), so we should be good to go.

docs/attributes-registry/code.md
SylvainJuge: I have opened an issue #1677 to keep track of the per-language definition and examples for each.

.chloggen/code-rename.yaml
brunobat: It is being used. 
If people have to start properly migrating these breaking changes maybe they will be less trigger happy in creating them.

trask: @brunobat are you asking to include code attribute breaking changes under a `OTEL_SEMCONV_STABILITY_OPT_IN` plan similar to http attribute breaking changes? thanks

brunobat: @trask, I'm ok with your suggestion.

@SylvainJuge, I've only used those properties in PoCs, it's not critical if they change. 
This is a matter of principle. 
In a public project like this, people shouldn't assume it's not used just because it doesn't show up in a search.
Many projects and libraries outside the OpenTelemetry repository are starting to rely on semantic conventions. Breaking them frequently will put a break on adoption.

SylvainJuge:
As a reminder, the current state of this PR renames the following fields, there has been some approvals before the `code.filepath` was also renamed, so it's probably safer to be explicit here:

- `code.function` renamed to `code.function.name`
- `code.lineno` renamed to `code.line.number`
- `code.column` renamed to `code.column.number`
- `code.filepath` renamed to `code.file.path`

There is only one unresolved conversation: https://github.com/open-telemetry/semantic-conventions/pull/1624#discussion_r1868412697, which is mostly about introducing a breaking change, @trask @lmolkova what do you think are the best next steps ? Do you think it's worth doing a proper migration plan ?

lmolkova:
@SylvainJuge let's follow @trask's proposal in https://github.com/open-telemetry/semantic-conventions/pull/1624#discussion_r1882702128 to add a warning - this would reduce any additional churn (if any). 

Instrumentations that emit existing `code` attributes may still provide migration plan if necessary, without us requiring everyone to do it.

SylvainJuge:
> @SylvainJuge let's follow @trask's proposal in [#1624 (comment)](https://github.com/open-telemetry/semantic-conventions/pull/1624#discussion_r1882702128) to add a warning - this would reduce any additional churn (if any).

I've added a warning (and a dedicated `README.md`) in c108ab2f0fb706585975222fee5ce9a07f13237e, let me know if that's enough for now.

model/code/registry.yaml
```diff
@@ -15,26 +15,26 @@ groups:
...
         examples: com.example.MyHttpService
-      - id: code.filepath
+      - id: code.file.path
         type: string
         stability: experimental
         brief: >
           The source code file name that identifies the code unit as uniquely as possible (preferably an absolute file path).
         examples: /usr/local/MyApplication/content_root/app/index.php
-      - id: code.lineno
+      - id: code.line.number
```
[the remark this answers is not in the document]
lmolkova: there is absolutely no harm in introducing a namespace, and we've already been bitten more than once by having `_` in the name. We're now using `.` whenever it makes sense.

PS: I can easily imagine `code.line.content`

docs/code/README.md
```diff
@@ -0,0 +1,23 @@
...
+# Semantic Conventions for Code
+
+**Status**: [Experimental][DocumentStatus]
+
+This document defines semantic conventions for source code.
+
+> **Warning**
+>
+> Existing code instrumentations that are using
+> [v1.29.0 of code conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.29.0/docs/attributes-registry/code.md)
```
lmolkova, line 17: maybe 

```suggestion
> [v1.29.0 of `code` conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.29.0/docs/attributes-registry/code.md)
```

? 

We don't do it in other warning, but "code conventions" seems to be too generic.

docs/code/README.md
```diff
@@ -0,0 +1,23 @@
...
+> (or prior):
+>
+> * SHOULD NOT change the version of the code attribute conventions that they emit by default
```
lmolkova, line 20:
```suggestion
> * SHOULD NOT change the version of the `code` conventions that they emit by default
```

docs/code/README.md
```diff
@@ -0,0 +1,23 @@
...
+    >   until the code semantic conventions are marked stable.
```
lmolkova, line 21:
```suggestion
    >   until the `code` semantic conventions are marked stable.
```

docs/database/dynamodb.md
```diff
@@ -30,7 +30,7 @@ The Semantic Conventions for [AWS DynamoDB](https://aws.amazon.com/dynamodb/) ex
 | [`rpc.method`](/docs/attributes-registry/rpc.md) | string | The name of the operation corresponding to the request, as returned by the AWS SDK [1] | `GetItem`; `PutItem` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | [`rpc.service`](/docs/attributes-registry/rpc.md) | string | The name of the service to which a request is made, as returned by the AWS SDK. [2] | `DynamoDB`; `S3` | `Recommended` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
-**[1] `rpc.method`:** This is the logical name of the method from the RPC interface perspective, which can be different from the name of any implementing method/function. The `code.function` attribute may be used to store the latter (e.g., method actually executing the call on the server side, RPC client stub method on the client side).
+**[1] `rpc.method`:** This is the logical name of the method from the RPC interface perspective, which can be different from the name of any implementing method/function. The `code.function.name` attribute may be used to store the latter (e.g., method actually executing the call on the server side, RPC client stub method on the client side).
```
lmolkova, line 33: Unrelated this this PR - we should probably change AWS SDK semconv to use `code.function.name` instead of `rpc.method`

lmolkova (approved):

docs/database/dynamodb.md
lmolkova: left a comment on relevant issue so we don't lose track of it https://github.com/open-telemetry/semantic-conventions/issues/1545

brunobat (approved):

PR #1632, opened 2024-11-29
Replace system.network.state with network.connection.state

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/308

## Changes

ref: https://github.com/open-telemetry/semantic-conventions/issues/308#issuecomment-2504357654


mx-psi (approved):

model/network/registry.yaml
```diff
@@ -236,3 +236,45 @@ groups:
...
+              value: 'syn_recv'
+              stability: experimental
+            - id: syn_sent
+              value: 'syn_sent'
+              stability: experimental
+            - id: time_wait
+              value: 'time_wait'
+              stability: experimental
+        stability: experimental
+        brief: "A stateless protocol MUST NOT set this attribute"
```
AlexanderWert, line 279: can we move that statement into `note` and in `brief` describe what that attribute stands for?

model/system/registry.yaml
```diff
@@ -152,6 +152,7 @@ groups:
     brief: "Describes Network attributes"
     attributes:
       - id: system.network.state
+        deprecated: 'Removed, report network connection state with `network.connection.state` attribute'
```
AlexanderWert, line 155: Since this attribute is being deprecated here, it should be moved to this file: `/model/system/deprecated/registry-deprecated.yaml`

braydonk (approved):

dmitryax (approved):

ChrsMark:
@AlexanderWert @open-telemetry/specs-semconv-maintainers that should be good to go?

schema-next.yaml
```diff
@@ -2,6 +2,14 @@ file_format: 1.1.0
 schema_url: https://opentelemetry.io/schemas/next
 versions:
   next:
+    metrics:
```
lmolkova, line 5: since the attribute is renamed and the old one is deprecated, we should apply this change to all signals, not just specific metric

model/network/registry.yaml
```diff
@@ -236,3 +236,46 @@ groups:
         type: string
         brief: 'The network interface name.'
         examples: [ 'lo', 'eth0' ]
+      - id: network.connection.state
+        type:
+          members:
+            - id: close
```
lmolkova, line 242: I think this is coming from https://datatracker.ietf.org/doc/html/rfc9293, can we 
- review values and adjust them to match the RFC? E.g. this one should be `closed`
- add link to the RFC in the brief?

model/network/registry.yaml
```diff
@@ -236,3 +236,46 @@ groups:
...
+            - id: fin_wait_2
+              value: 'fin_wait_2'
+              stability: experimental
+            - id: last_ack
+              value: 'last_ack'
+              stability: experimental
+            - id: listen
+              value: 'listen'
+              stability: experimental
+            - id: syn_recv
```
lmolkova, line 269: RFC uses `SYN-RECEIVED`

model/network/registry.yaml
```diff
@@ -236,3 +236,46 @@ groups:
...
+            - id: close
+              value: 'close'
+              stability: experimental
+            - id: close_wait
+              value: 'close_wait'
+              stability: experimental
+            - id: closing
+              value: 'closing'
+              stability: experimental
+            - id: delete
```
lmolkova, line 251: I don't think there is a delete state in the RFC - do we need to add it?

model/network/registry.yaml
```diff
@@ -236,3 +236,46 @@ groups:
...
+              stability: experimental
+            - id: syn_sent
+              value: 'syn_sent'
+              stability: experimental
+            - id: time_wait
+              value: 'time_wait'
+              stability: experimental
+        stability: experimental
+        brief: "The state of network connection"
+        note: "A stateless protocol MUST NOT set this attribute"
```
lmolkova, line 280: I suggest to remove this note. Telemetry describing stateless protocols won't use this attribute and it seems obvious that stateless thing doesn't need a state.

lmolkova (commented):
Looks good, left a few suggestions to clean up old enum members

model/network/registry.yaml
ChrsMark: I did some digging, looking into where those are used/defined. It seems that they are defined in `hostmetrics` receiver:
1. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6f3ff4a37efe4591f728250d4ea9e4ad6672fb2b/receiver/hostmetricsreceiver/internal/scraper/networkscraper/network_others.go#L9-L20 
2. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c4bee406f5db233a2bc7f0c69185764fea9fb033/receiver/hostmetricsreceiver/internal/scraper/networkscraper/network_linux.go#L18-L29

Used at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c4bee406f5db233a2bc7f0c69185764fea9fb033/receiver/hostmetricsreceiver/internal/scraper/networkscraper/network_scraper.go#L178-L188

These match with what `gopsutil` provides:
1. https://github.com/shirou/gopsutil/blob/62a52f910bd9ad7d48a7416f5c2899386d50cb37/net/net_linux.go#L308
2. https://github.com/shirou/gopsutil/blob/62a52f910bd9ad7d48a7416f5c2899386d50cb37/net/net_windows.go#L565

Moreover, these seem to link to the OS specific definitions:
1. https://github.com/torvalds/linux/blob/cdd30ebb1b9f36159d66f088b61aee264e649d7a/include/net/tcp_states.h#L19
2. https://learn.microsoft.com/en-us/windows/win32/api/tcpmib/ns-tcpmib-mib_tcprow2?redirectedfrom=MSDN#members

So it seems that it is `CLOSED` for Windows while `CLOSE` for Linux. Should we allow both or should we fix the receiver's implementation to correctly map to RFC's values?

@braydonk @dmitryax what are your thoughts since you are the code-owners of the receiver?

model/network/registry.yaml
ChrsMark: Same as https://github.com/open-telemetry/semantic-conventions/pull/1632#discussion_r1867330961. 

It is `SYN_RECV` for Linux while `SYN_RECEIVED` for Windows.

model/network/registry.yaml
lmolkova: I defer to system semconv group to discuss and decide. I have a preference to stick to RFC definition (which is cross-platform) and allow OS-specific values. In case of `close`/`closed`, the method that translates linux enum number into otel string value can take care of reporting the standard RFC state name.

model/network/registry.yaml
lmolkova: replied there, my suggestion is to stick to RFC and it should be easy to translate enum const to otel attribute value string anyway.

model/network/registry.yaml
lmolkova: can we postpone providing it? It's not in the RFC and it's always easy to extend enum in semconv.

model/network/registry.yaml
mx-psi: Makes sense to me to remove it

model/network/registry.yaml
ChrsMark: Discussed in System Metrics WG today. We agree on having the SemConv following what the (cross-platform) RFC defines and then the receiver to properly do the mappings accordingly for each platform/implementation.

I will update this PR accordingly.

model/network/registry.yaml
ChrsMark: Removed.

model/network/registry.yaml
ChrsMark: Removed that.

model/network/registry.yaml
ChrsMark: Tuned the PR according to the rfc9293.

ChrsMark:
@lmolkova @open-telemetry/semconv-system-approvers I have tuned the PR so as the state values to be aligned with [rfc9293](https://datatracker.ietf.org/doc/html/rfc9293#section-3.3.2).

I'm not sure if we should explicitly deprecate the enum values that were changed or removed entirely.

mx-psi (approved):

lmolkova (approved):
LGTM and thank you!

lmolkova:
tagging @open-telemetry/semconv-security-approvers in case they want to take a look

schema-next.yaml
```diff
@@ -2,6 +2,12 @@ file_format: 1.1.0
 schema_url: https://opentelemetry.io/schemas/next
 versions:
   next:
+    all:
```
trisch-me, line 5: should here be also noted changes to the metrics?

model/system/deprecated/registry-deprecated.yaml
```diff
@@ -51,3 +51,46 @@ groups:
...
+              value: 'syn_recv'
+              stability: experimental
+            - id: syn_sent
+              value: 'syn_sent'
+              stability: experimental
+            - id: time_wait
+              value: 'time_wait'
+              stability: experimental
+        stability: experimental
+        brief: "A stateless protocol MUST NOT set this attribute"
```
trisch-me, line 95:
```suggestion
        brief: "Deprecated, use `network.connection.state` instead."
```

schema-next.yaml
ChrsMark: Initially it was applied only to metrics but I made it to apply to all as per https://github.com/open-telemetry/semantic-conventions/pull/1632#discussion_r1867013714

trisch-me (approved):

trisch-me:
@ChrsMark I think one test is failing, otherwise good to go!

PR #1635, opened 2024-11-29
Messaging: Clarify `{destination}` value used on span name

### LINKED ISSUE #1532: [Messaging] What to use for `{destination}` on span name when destination is temporary or anonymous
### Area(s)

area:messaging

In the current conventions for messaging Span name we have:

> The {destination} SHOULD describe the entity that the operation is performed against and SHOULD adhere to one of the following values, provided they are accessible:
> 
> 1. messaging.destination.template SHOULD be used when it is available.
> 2. messaging.destination.name SHOULD be used when the destination is known to be neither [temporary nor anonymous](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md#temporary-and-anonymous-destinations).
>  3. server.address:server.port SHOULD be used only for operations not targeting any specific destination(s).

But it is unclear what to use for `{destination}` if the destination is temporary or anonymous. Should it be used at all?

In the RabbitMQ Java instrumentation, when the destination is temporary, the `messaging.destination.name` attribute [is populated with the value `(temporary)`.](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api-incubator/src/main/java/io/opentelemetry/instrumentation/api/incubator/semconv/messaging/MessagingAttributesExtractor.java#L97), which raises the question if the same should be done for the span name. 

- We should define what value to use for `messaging.destination.name` when the destination is temporary or anonymous to avoid inconsistency across instrumentation
- We should define what the span name should look like when the destination is temporary or anonymous.

- We can mention that when the destination is temporary or anonymous,  `messaging.destination.name` should be set to something like `(temporary)` or `(anonymous)`
- For the span name we CAN:
  - Say if the destination is either temporary or anonymous, and no other attributes are available `{destination}` should be omitted on the span name OR
  - Say if the destination is either temporary or anonymous, a placeholder of `temporary` or `anonymous` should be used for `{destination}` on the span name

### Additional context

_No response_

joaopgrassi:
Fixes #1532

## Changes

- Make it explicit that `{destination}` should not be used in high cardinality scenarios (e.g., when the destination is temporary or anonymous)


model/messaging/registry.yaml
```diff
@@ -33,9 +33,12 @@ groups:
         type: string
         stability: experimental
         brief: 'The message destination name'
-        note: |
+        note: >
           Destination name SHOULD uniquely identify a specific queue, topic or other entity within the broker. If
           the broker doesn't have such notion, the destination name SHOULD uniquely identify the broker.
+
+          If `messaging.destination.anonymous` or `messaging.destination.temporary` is set to true,
```
lmolkova, line 40: we still want to capture the actual destination name, don't we? 

Temporary queues (e.g. https://www.rabbitmq.com/docs/queues#temporary-queues) are not different from regular queues, but they are short lived and their names might have high cardinality.

E.g. rabbit broker can generate random(?) name - https://www.rabbitmq.com/docs/queues#server-named-queues or user might assign it.

In any case, it seems important to capture the actual queue name regardless of its life expectancy on spans, but protect span name and metrics from high-cardinality.

carlosalberto: Do we have alternatives if we want to capture the destination name, other than "hiding" it behind a config option? (default would be using "(temporary)")

joaopgrassi: > In any case, it seems important to capture the actual queue name regardless of its life expectancy on spans, but protect span name and metrics from high-cardinality.

Right I think that makes sense on the `destination.name` here. So, what is really missing is define what to put on the span name when the destination is temporary/anonymous. 

I removed this now and tweaked the "destination" concept for span name. PTAL

docs/messaging/messaging-spans.md
```diff
@@ -191,17 +191,24 @@ Messaging spans SHOULD follow the overall [guidelines for span names](https://gi
...
 
-The <span id="destination-placeholder">`{destination}`</span> SHOULD describe the entity that the operation is performed against
+The <span id="destination-placeholder">`{destination}`</span>
+SHOULD describe the entity that the operation is performed against
 and SHOULD adhere to one of the following values, provided they are accessible:
 
 1. `messaging.destination.template` SHOULD be used when it is available.
 2. `messaging.destination.name` SHOULD be used when the destination is known to be neither [temporary nor anonymous](#temporary-and-anonymous-destinations).
 3. `server.address:server.port` SHOULD be used only for operations not targeting any specific destination(s).
+4. The placeholder `(temporary)` SHOULD be used when the destination is [temporary or anonymous](#temporary-and-anonymous-destinations)
```
lmolkova, line 210: we can just say that destination should not be provided if there is no low-cardinality destination. 

E.g. this is how DB span name is phrased

https://github.com/open-telemetry/semantic-conventions/blob/f0974a892c308511bcfd21fc13dab2d3e598a9fa/docs/database/database-spans.md?plain=1#L68-L73

It's not blocking. 
If you believe that having `(temporary)` marker is helpful, let's go with it.

lmolkova (approved):

docs/messaging/messaging-spans.md
joaopgrassi: Yeah given how the outcome now of the other PR https://github.com/open-telemetry/semantic-conventions/pull/1531 - we will have both the flag indicating it is temporary/anonymous + the actual name in the `messaging.destination.name` I don't see value in using the placeholder here.

trask (approved):

trask:
looks like needs a change log entry

carlosalberto (approved):

PR #1641, opened 2024-12-02
Fix event policy check broken after 1.29.0 release

lmolkova:
Back-compat policy check for events had a bug (events in new version were always empty). 

It was dormant because we never marked any event group as stable (until new release 1.29.0). 
image

This PR:
- fixes policy issue
- removes `stable` condition - we should not allow removing any events
- adds policy tests

### DIFF
--- a/policies/compatibility.rego
+++ b/policies/compatibility.rego
@@ -48,7 +48,7 @@ baseline_events := [ g |
 ]
 registry_events := [g |
     some g in data.semconv.groups
-    g.type == "events"
+    g.type == "event"
 ]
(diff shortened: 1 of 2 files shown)


policies/compatibility.rego
```diff
@@ -48,7 +48,7 @@ baseline_events := [ g |
 ]
 registry_events := [g |
     some g in data.semconv.groups
-    g.type == "events"
+    g.type == "event"
```
jsuereth, line 51: ouch! good catch.

jsuereth (approved):

trask (approved):

PR #1642, opened 2024-12-02
Relax collision detection policy to allow collisions between deprecated namespaces and attributes (and vice-versa)

lmolkova:
Related to https://github.com/open-telemetry/semantic-conventions/pull/1624, https://github.com/open-telemetry/semantic-conventions/pull/1638

Preventing collisions between namespace and attribute name is a protection in case we want to support nested properties in the attributes or add any semantic meaning to namespaces. E.g. having non-deprecated attribute `foo` along with non-deprecated attribute `foo.bar` would be problematic. 

It seems fine however to have deprecated `foo` along with non-deprecated `foo.bar`. 

This PR relaxes the policy and unblocks PRs like #1624 and #1638 (if they are approved).

### DIFF
--- a/policies/attribute_name_collisions.rego
+++ b/policies/attribute_name_collisions.rego
@@ -6,7 +6,7 @@ import rego.v1
 attribute_names := { obj |
     group := input.groups[_];
     attr := group.attributes[_];
-    obj := { "name": attr.name, "const_name": to_const_name(attr.name), "namespace_prefix": to_namespace_prefix(attr.name), "deprecated": get_or_null(attr, "deprecated") }
+    obj := { "name": attr.name, "const_name": to_const_name(attr.name), "namespace_prefix": to_namespace_prefix(attr.name), "deprecated": is_property_set(attr, "deprecated") }
 }
 
 # check that attribute constant names do not collide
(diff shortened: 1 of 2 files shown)


policies/attribute_name_collisions.rego
```diff
@@ -72,13 +78,17 @@ to_const_name(name) = const_name if {
...
 }
 
-# These lists contain exceptions for existing collisions that were introduced unintentionally.
+get_or_null(obj, key) = value if {
+    value := obj[key]
+} else = null if {
+    not obj[key]
+}
+
+# This list contain exceptions for existing collisions that were introduced unintentionally.
```
trask, line 87:
```suggestion
# This list contains exceptions for existing collisions that were introduced unintentionally.
```

trask (approved):

policies/attribute_name_collisions.rego
lmolkova: thank you!

policies/attribute_name_collisions.rego
```diff
@@ -72,13 +78,17 @@ to_const_name(name) = const_name if {
     const_name := replace(name, ".", "_")
 }
 
-# These lists contain exceptions for existing collisions that were introduced unintentionally.
+get_or_null(obj, key) = value if {
```
jsuereth, line 81: nit: Should this be: "has_or_null"?

I.e. we check existence and turn it into a true or false of sorts.

jsuereth (approved):
Looks good, just one minor nit.

PR #1646, opened 2024-12-03
Deprecate `event.name` attribute in favor of EventName on the log record

lmolkova:
Part of https://github.com/open-telemetry/opentelemetry-specification/issues/4260

Logs SIG is working on migrating from `event.name` attribute to `EventName` field on LogRecord. This is a WIP that depends on the 
- https://github.com/open-telemetry/opentelemetry-specification/pull/4320
- https://github.com/open-telemetry/opentelemetry-proto/pull/600

Creating it to socialize the proposal and gather high-level feedback.

## Semantic conventions that depend on `event.name` (over-the-wire):

- [gen-ai events](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-events.md) and corresponding python [instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai/opentelemetry-instrumentation-openai-v2)
- [`az.resource.log`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/azure/events.md?plain=1) and the corresponding [collector receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/azurelogs/resourcelogs_to_logs.go)
- [`device.app.lifecycle`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/mobile/events.md?plain=1)  event - did not find any instrumentations in otel org.
- [`feature_flag.evaluation` aka `feature_flag` event](https://github.com/open-telemetry/semantic-conventions/blame/main/docs/feature-flags/feature-flags-logs.md) - no instrumentations in otel org
- `exception` and `rpc.message` are span events - they should not populate `event.name` and should not be affected at all

## Transition

`event.name` is an experimental attribute and de-facto is used by new-ish conventions and instrumentation libraries. 

So we (Logs SIG) believe no common transition/migration plan is necessary, but individual instrumentation libraries can provide back-compat story.

- Instrumentations that produce events using explicit `event.name` attribute can do a breaking change and start using `LogRecord.EventName` property instead of the attribute. It's up to each instrumentation component to provide (or not provide) transition story. 
- Telemetry consumers should expect to receive deprecated `event.name` attribute, new `LogRecord.EventName` property or both. When both are present, the `LogRecord.EventName` should be used as a source of truth.

## Future 

Instead of deprecating `event.name`, we might want to reserve it for bridges similarly to `otel.status_code`. E.g. when users want to report event using logging bridge and the logging facade does not have a notion of event name, the `event.name` attribute maybe used  by the end users and translated to top-level `LogRecord` property by the adapter.

Deprecating the attribute is reversible, so I'd prefer to do it now and tackle compatibility with external logging libs separately.


breedx-splk (commented):
I know it's still just draft, but this seems fine.

docs/general/events.md
```diff
@@ -32,29 +32,9 @@ field semantics, and stability and requirement levels. Other events may be user-
...
-<!-- see templates/registry/markdown/snippet.md.j2 -->
-<!-- prettier-ignore-start -->
-<!-- markdownlint-capture -->
-<!-- markdownlint-disable -->
-
-| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
-|---|---|---|---|---|---|
-| [`event.name`](/docs/attributes-registry/event.md) | string | Identifies the class / type of event. [1] | `browser.mouse.click`; `device.app.lifecycle` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-
-**[1] `event.name`:** Event names are subject to the same rules as [attribute names](/docs/general/attribute-naming.md). Notably, event names are namespaced to avoid collisions and provide a clean separation of semantics for events in separate domains like browser, mobile, and kubernetes.
```
MSNev, line 48: This seems like a good place to keep this definition as it's part of the qualification for achieving the statement of `that uniquely identifies the event` (this went through many previous discussions around how to we ensure that the name is unique.

docs/general/events.md
```diff
@@ -87,15 +67,15 @@ Recommendations on using attributes vs. body fields:
 
 * If the field should be comparable across every type of record, it should be an attribute.
 * If the field is specific to the event itself, then it should be a body field.
-* Unless the same `event.name` exists on two events, anything in two event bodies is not comparable to each other.
+* Unless two events share the same `EventName` property value, anything in two event bodies is not comparable to each other.
```
MSNev, line 70: I think we should reword the "two events" definition, as now it reads (to me) more specific than the previous trailing "two events"... Perhaps something like
(really just removing the "two" qualification as someone is likely to ask the question of what about if three, four, five events etc.)

"Unless the `EventName` property value matches, any fields or values within event bodies that share the same name or path should not be considered comparable."

With maybe a qualifier "Unless explicitly called out by the specific event definition" (or we can leave this out as edge cases (which they are).

And maybe a general follow up qualification about case-sensitivity (although I'm assuming that by keeping the above comments about `Event names are subject to the same rules as [attribute names](/docs/general/attribute-naming.md)` this "should" avoid this question... (Hopefully)

docs/attributes-registry/event.md
```diff
@@ -12,6 +12,4 @@ Attributes for Events represented using Log Records.
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
-| <a id="event-name" href="#event-name">`event.name`</a> | string | Identifies the class / type of event. [1] | `browser.mouse.click`; `device.app.lifecycle` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-
-**[1] `event.name`:** Event names are subject to the same rules as [attribute names](/docs/general/attribute-naming.md). Notably, event names are namespaced to avoid collisions and provide a clean separation of semantics for events in separate domains like browser, mobile, and kubernetes.
```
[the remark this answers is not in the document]
lmolkova: We no longer have attribute to add a note to, we also document the uniqueness of event name in multiple other places:
- the spec ["This name SHOULD uniquely identify the event structure (both attributes and body)"](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-eventname)
- semconv [" An event MUST have an Event name property that uniquely identifies the event."](https://github.com/open-telemetry/semantic-conventions/blob/adf6a98ea2fe884a03e67b5062dcb1833bd44dd8/docs/general/events.md?plain=1)

docs/general/events.md
lmolkova: we still have this language in the semantic conventions, but not in the attribute definitions. - https://github.com/open-telemetry/semantic-conventions/pull/1646#discussion_r1876373344

docs/general/events.md
lmolkova: I've updated the text, ptal

docs/general/events.md
```diff
@@ -32,30 +28,11 @@ field semantics, and stability and requirement levels. Other events may be user-
...
-<!-- markdownlint-restore -->
-<!-- prettier-ignore-end -->
-<!-- END AUTOGENERATED TEXT -->
-<!-- endsemconv -->
-
-### General event semantics
-
-* An event MUST have an `event.name` attribute that uniquely identifies the event.
-* It MAY have [standard](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.39.0/specification/common#attribute)
+* An event MUST have an [Event name property](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.40.0/specification/logs/data-model.md#field-eventname)
```
lmolkova, line 33: spec 1.40.0 release is pending, let's wait for it to land

breedx-splk (approved):
Onward! 🚀

trask (approved):

Issue #332, opened 2023-09-19
Markdown link check runs on entire repo


breedx-splk:
When the required build checks for a PR run, one of these checks is a markdown link checker. This is useful to help ensure that we don't amass a bunch of broken links in the docs. However, the markdown link checker looks like it runs across the entire repo. This can result in PRs failing their checks even when the broken link exists in an unrelated pice of markdown. This can cause small PRs to appear broken even when they are not, and it can extend time-to-merge.

One example of this has been #215, where a link in the http docs caused the client `session.id` PR to fail verifications.

Suggestions:
* Only run the markdown link check against markdown files that have changed in the PR
* Instead of running on PRs, only run on a schedule across the whole repo and open an issue when the checker fails

AlexanderWert:
> Only run the markdown link check against markdown files that have changed in the PR

What if a file `A.md` was moved to `A-new.md` in that PR, and another _untouched_ markdown file `B.md` referenced the old file `A.md` but has not been updated in that PR? By running the check only on changed files would miss this broken link, right?

trask:
In the Java repos, we retry failures (file-by-file) a few times to help mitigate sporadic failures: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/scripts/markdown-link-check-with-retry.sh

breedx-splk:
> > Only run the markdown link check against markdown files that have changed in the PR
> 
> What if a file `A.md` was moved to `A-new.md` in that PR, and another _untouched_ markdown file `B.md` referenced the old file `A.md` but has not been updated in that PR? By running the check only on changed files would miss this broken link, right?

Definitely not a use case I had considered (repo-internal breakage)....I've only ever seen it break on external outbound links 🤣 .

I have no idea how to do this, but if the tools somehow support internal/external distinction this could be useful. Always check/validate internal links but only external links in changed files. I'm sure this is an unreasonable request. 😁

breedx-splk:
Oh, I see that the [github action version](https://github.com/gaurav-nelson/github-action-markdown-link-check) (not npm) has this `check-modified-files-only` option. 

How about this compromise (and I'm happy to PR this if that helps):
* PR checks are relaxed to only check the modified files
* Entire repo is checked nightly and issue created on failure

Thoughts?

wojtekLs:
@breedx-splk 
LGTM :)

joaopgrassi:
[...]

PR #1655, opened 2024-12-06
Extend GenAI System to include OpenAI compatible platforms

codefromthecrypt:
This extends GenAI System to include OpenAI compatible platforms. These are usable by OpenAI client libraries:

* [Azure OpenAI](https://github.com/openai/openai-python?tab=readme-ov-file#microsoft-azure-openai)
* [Gemini](https://developers.googleblog.com/en/gemini-is-now-accessible-from-the-openai-library/) 
* [Perplexity](https://docs.perplexity.ai/api-reference/chat-completions)
* [xAI](https://x.ai/api) (note they also say they are compatible with anthropic!)
* [DeepSeek](https://api-docs.deepseek.com/zh-cn/)
* [Groq](https://console.groq.com/docs/openai)

At the moment, there is no way to attribute signals to Azure OpenAI, so people use "openai". At Elastic, we have different
dashboards for OpenAI (the platform) vs Azure OpenAI, as these are independent services. Lacking a GenAI system name, we have to resort to "openai" which is ambiguous between these two.

Using "server.address" isn't a great option due to subdomains used in the Azure OpenAI service depending on the application name in Azure. While we could use cloud semantics, they aren't guaranteed to be present, and aren't integrated in current code. Even if it were, this it is less ideal to navigate or aggregate on two attributes vs one.

This also adds a value for Google Gemini which has exactly the same concern as it can be accessed either via its native API or via OpenAI libraries.

## Changes

Added the following, clarifying only when needed:
* "az.ai.openai" to follow conventions of "az.ai.inference"
* "gemini"", not "google.gemini" as "vertexai" is not prefixed either.
* "perplexity"
* "xai"
* "deepseek"
* "groq"


codefromthecrypt:
PS I used the Azure AI Inference client client per [instructions](https://github.com/elastic/genai-instrumentation/pull/318#:~:text=configuration%20per%20their-,README,-%2C%20including%20ENV%20variables) to access Azure OpenAI service, and I guess what's happening here more generally, is that `gen_ai.system` might be conflated with the client itself.

image

Specifically, the Azure AI Inference client can access three different products:
* GitHub models
* Azure AI Studio models
* Azure OpenAI Studio models

In each case the services are different (different subdomains, for example). However, status quo, all would set `gen_ai.system=az.ai.inference`

So, status quo `gen_ai.system` might be treated more like the SDK name than the actual system name. Should I update the text like that, as this applies even beyond openai compat?

karthikscale3:
This is definitely an issue that we have faced as well. Right now auto instrumentation default sets genai.system to openai even if the base_url used is different from the openai one. More broadly it's only looking at the client library being used as opposed to the inference server being used, which is the correct representation. In Langtrace's SDK, we capture the underlying server being used from the [base url using regex](https://github.com/Scale3-Labs/langtrace-python-sdk/blob/main/src/langtrace_python_sdk/instrumentation/openai/patch.py#L247) and set a langtrace attribute called `service-name`. I think we can move this behavior up to the spec with `genai.system`. We also need to capture this for all the other LLM providers that support openai client library

1. xai
2. deepseek
3. groq
4. perplexity

docs/attributes-registry/gen-ai.md
```diff
@@ -67,7 +69,9 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
 | `anthropic` | Anthropic | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `aws.bedrock` | AWS Bedrock | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `az.ai.inference` | Azure AI Inference | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `az.ai.openai` | Azure OpenAI | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `cohere` | Cohere | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `gemini` | Gemini | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `ibm.watsonx.ai` | IBM Watsonx AI | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `openai` | OpenAI | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `vertex_ai` | Vertex AI | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
karthikscale3, line 77: We can also add perplexity, xai, deepseek and groq to this list

codefromthecrypt: yeah I was thinking about xai, just didn't add it yet. happy to.

I didn't know that perplexity deepseek present openai compatible services. Can you link me? Otherwise, I'd prefer this PR to only include OpenAI compatible ones

lmolkova:
Base URL regex is not reliable either - you can run smthing behind the proxy, self-host model in vllm, your provider can host different models behind one API. Instrumentation can do the best effort, but cannot guarantee anything beyond "a system this library was designed for" on the `gen_ai.system`.

lmolkova (approved):

docs/attributes-registry/gen-ai.md
codefromthecrypt: Found links, so will update in a sec

* https://docs.perplexity.ai/api-reference/chat-completions
* https://x.ai/api (note they also say they are compatible with anthropic!)
* https://api-docs.deepseek.com/zh-cn/
* https://console.groq.com/docs/openai

codefromthecrypt:
@lmolkova want me to fix the "dead link" on https://blogs.oracle.com/linux/post/understanding-linux-kernel-memory-statistics? it is result code zero which I've seen elsewhere when a link takes too long to load or otherwise

codefromthecrypt:
The broken links might not be a new problem. We can work around this by retrying externally, in the make file, but I imagine for now, folks can just manually merge when MLC is flakey on a good link. https://github.com/open-telemetry/semantic-conventions/issues/332#issuecomment-2530588430

xrmx (approved):

xrmx:
> The broken links might not be a new problem. We can work around this by retrying externally, in the make file, but I imagine for now, folks can just manually merge when MLC is flakey on a good link. [#332 (comment)](https://github.com/open-telemetry/semantic-conventions/issues/332#issuecomment-2530588430)

It has been here since a few and has not been a problem to have things merged :)

lmolkova (approved):

codefromthecrypt:
This has been approved for a while now, anything missing needed for merge?

lmolkova:
@codefromthecrypt we require at least 2 approving reviews.

karthikscale3 (approved):

codefromthecrypt:
Thanks for merging this.

Last night, I noticed I missed a spot as mistral (system/platform/llm provider) is also compatible with openai (chat completions api). I was testing our [chatbot-rag-app](https://github.com/elastic/elasticsearch-labs/tree/main/example-apps/chatbot-rag-app), used by solutions architects for client demonstrations. It supports mistral as an llm implementation for LangChain. While LangChain uses raw HTTP to access mistral, it does so with the openai completions endpoint. I also noticed [spring-ai squarely classifies mistral as openai (api) compatible](https://docs.spring.io/spring-ai/reference/api/chat/mistralai-chat.html#_openai_api_compatibility). Basically as a cloud platform, mistral is accessed similar to the others out there. cc @ThomasVitale from spring-ai and I'll raise a follow-up PR with these notes.

cc @jack-berg on a sidebar as this PR puts modeling nuance in a very specific and concrete context. Others can ignore
---

mistral is a system/platform/llm provider e.g. https://docs.spring.io/spring-ai/reference/api/chat/mistralai-chat.html#_openai_api_compatibility shows its cloud endpoint
it is *also* a family of models, e.g. you can use mistral-nemo with ollama on your laptop, or on other cloud platforms.

depending on context, what we say is a "system" is a platform, api, model family or multiple of these. What's identifiable from the client is limited, but from the server you also get the fun of whether you report "ollama" or "vllm" as a "system" or not, regardless of it traffic came in on the "openai" compatible endpoint vs another inference one.

codefromthecrypt:
https://github.com/open-telemetry/semantic-conventions/pull/1719 on the follow-up for Mistral AI

Issue #1530, opened 2024-10-29
AI Agent framework Semantic Convetion


gyliu513:
### Area(s)

area:gen-ai

### Is your change request related to a problem? Please describe.

There are now many agent frameworks for GenAI, including crewai, autogen etc. Those agents will include many different components, like agent, tools, tasks etc, we need a semantic convention for those resources as well for agents.

@lmolkova @nirga @drewby @karthikscale3 ^^

### Describe the solution you'd like

Extend the GenAI Semantic Convention to cover agents

lmolkova:
FYI: I'm baking some early prototype here https://github.com/microsoft/opentelemetry-semantic-conventions/pull/3 I'll re-send the PR against this repo

gyliu513:
I did some test with crewai and Instana observability at https://gyliu513.github.io/jekyll/update/2024/10/22/crewai-observability.html, we may need some agent framework level semantic convention.

![image](https://github.com/user-attachments/assets/5e8616dd-cac1-486f-a0e2-7c50e84a1893)

Also did some test with langtrace, it can also capture some tracing and metrics when I was using crewai based on langchain.

![Screenshot 2024-10-29 at 3 18 22 PM](https://github.com/user-attachments/assets/02c27c09-459d-4415-a5b0-2455c28955a4)

@karthikscale3 has some demo for agent observability with langtrace as well.

@lmolkova it is great you have a PR under-going, will take a look, thanks!

lmolkova:
wow, this looks awesome!

I'm struggling with a couple of things and hope to gets your thoughts on a few big things:
- what's the scope: client framework, client to service-side agent, multi-agent story. How may layers do we need?
- the level of unification - e.g. task execution is the same as openai assistant run and will be the same as azure AI agent run - how far do we need to go in attempts to unify

Essentially my main worry is that I don't fully understand what exactly we want to unify in semconv or what we want the conventions for.

There is always an alternative that the LLM client level is unified but frameworks do some extra stuff which does not need much consistency. LMK what you think.

karthikscale3:
Thanks for starting the thread @gyliu513 . Shown below the span graph on Langtrace for CrewAI. Generally speaking, here's a general pattern that's emerging across agentic frameworks:

[1] Sessions - Each session can have multiple agents working independently or together to perform a bunch of tasks.
[2] Agents - Different agents that can do different tasks
[3] Agent Config - Sequential, Hierarchical, Networked
[...]

PR #1657, opened 2024-12-06
AI Agent Span Semantic Convention

gyliu513:
Fixed part of https://github.com/open-telemetry/semantic-conventions/issues/1530

## Changes


gyliu513:
@lmolkova @lzchen @nirga @karthikscale3 @drewby this is the very draft version, we may need a long discussion for this, hope we can start from here. 

Please share your comments here, actually, I do not know if we want to put the ai agent semantic convention to same folder as `gen-ai` or we need a new folder for ai-agent. Thanks!

docs/ai-agent/ai-agent-spans.md
```diff
@@ -0,0 +1,185 @@
...
+| `ai_agent.workflow.end_reason` | string | Reason for the workflow's end, if applicable.             | `Completed all tasks`             | Recommended       | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
+| `ai_agent.workflow.system`     | string | System or environment where the workflow is executed.     | `CrewAI`, `LangGraph`, `AutoGen`  | Recommended       | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
+
+### Agent Attributes
+
+| Attribute                      | Type   | Description                                | Example                          | Requirement Level | Stability    |
+| ------------------------------ | ------ | ------------------------------------------ | -------------------------------- | ----------------- | --- |
+| `ai_agent.agent.name`          | string | Name of the agent.                         | `Researcher Bot`                 | Required          | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
```
lmolkova, line 68: Those are still genai agents, so I think this out should be `gen_ai.agent.name`

docs/ai-agent/ai-agent-spans.md
```diff
@@ -0,0 +1,185 @@
...
+| `ai_agent.agent.role`          | string | Role assigned to the agent.                | `Data Collector`                 | Required          | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
+| `ai_agent.agent.backstory`     | string | Background story or context for the agent. | `Specializes in web data mining` | Required          | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
+| `ai_agent.agent.workflow_name` | string | Name of the workflow the agent is part of. | `Data Processing Pipeline`       | Required          | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
+| `ai_agent.agent.model`         | string | Underlying model powering the agent.       | `gpt-4`                          | Required          | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
```
lmolkova, line 72: how's agent model is different from `gen_ai.request.model`?

docs/ai-agent/ai-agent-spans.md
```diff
@@ -0,0 +1,185 @@
...
+| `ai_agent.agent.tools`         | array  | List of tools available to the agent.      | `["Web Scraper", "Analyzer"]`    | Recommended       | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
```
lmolkova, line 73: how'd it be different from generic gen_ai tool?

lmolkova (changes requested):
A few general points:
- we should not create attributes that would be the same as existing gen_ai attributes. We should use those instead of defining agent ones by default
- we need to define everything in yaml and stay compatible with the schema

I have a draft here - https://github.com/microsoft/opentelemetry-semantic-conventions/pull/3 for OpenAI assistant-like API which covers a lot of similar things, PTAL

svrnm:
hey there, 

I think this is a great starting point for a very hot topic right now. There is one major comment I have on that, and it's around the `workflow` and `task`. If I understand it correctly there is a `workflow` to `task` relationship and the `task`, I would assume they are modelled in a parent span, span relationship? So a workflow is a "parent span" and then tasks are "child spans" to that, is this a correct assumption? 

If that's correct and if I look at your examples, a workflow (and maybe even a task?) can be very long running, which is a currently [unsolved](https://github.com/open-telemetry/opentelemetry-specification/issues/373) piece of the otel specification, so maybe this need for AI Agents being modelled could help to be a driven force behind providing a beter specification for that, because I additionally see `workflow` and `task` not being unique to AI agents, see [CICD pipeline attributes](https://opentelemetry.io/docs/specs/semconv/attributes-registry/cicd/#cicd-pipeline-attributes) for example.

gyliu513:
> If that's correct and if I look at your examples, a workflow (and maybe even a task?) can be very long running, which is a currently https://github.com/open-telemetry/opentelemetry-specification/issues/373 piece of the otel specification, so maybe this need for AI Agents being modelled could help to be a driven force behind providing a beter specification for that, because I additionally see workflow and task not being unique to AI agents, see [CICD pipeline attributes](https://opentelemetry.io/docs/specs/semconv/attributes-registry/cicd/#cicd-pipeline-attributes) for example.

@svrnm Yes, this is the case, at least from my point of view, the `workflow` and `task` relationship is very similar as the [CICD pipeline attributes](https://opentelemetry.io/docs/specs/semconv/attributes-registry/cicd/#cicd-pipeline-attributes). 

Let me review https://github.com/microsoft/opentelemetry-semantic-conventions/pull/3 from @lmolkova first, and I will try to update my PR soon after some discussion on https://github.com/microsoft/opentelemetry-semantic-conventions/pull/3

docs/ai-agent/ai-agent-spans.md
karthikscale3: +1

karthikscale3:
> hey there,
> 
> I think this is a great starting point for a very hot topic right now. There is one major comment I have on that, and it's around the `workflow` and `task`. If I understand it correctly there is a `workflow` to `task` relationship and the `task`, I would assume they are modelled in a parent span, span relationship? So a workflow is a "parent span" and then tasks are "child spans" to that, is this a correct assumption?
> 
> If that's correct and if I look at your examples, a workflow (and maybe even a task?) can be very long running, which is a currently [unsolved](https://github.com/open-telemetry/opentelemetry-specification/issues/373) piece of the otel specification, so maybe this need for AI Agents being modelled could help to be a driven force behind providing a beter specification for that, because I additionally see `workflow` and `task` not being unique to AI agents, see [CICD pipeline attributes](https://opentelemetry.io/docs/specs/semconv/attributes-registry/cicd/#cicd-pipeline-attributes) for example.

@svrnm / @gyliu513 - Just wanted to add my 2 cents here. This is very much an issue and thanks for bringing it up. The OTEL instrumentation we have(at Langtrace) for frameworks like CrewAI, DSPy etc. runs into this issue from time to time where traces have, in occasions 100s of spans as part of the same trace. An option to flush spans in progress will be ideal for these scenarios so the user can see realtime feedback on the UI for ongoing agentic sessions. Having said that, a vast majority of the agents we are seeing(from our perspective) still work well with the existing capabilities. But, we definitely need to think about this sooner than later.

docs/ai-agent/ai-agent-spans.md
```diff
@@ -0,0 +1,185 @@
...
+| `ai_agent.tool.function` | string | Specific function or capability of the tool. | `Data Extraction` | Recommended       | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
+| `ai_agent.tool.output`   | object | Output produced by the tool.                 | `{"data": [...]}` | Required          | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
+
+### Task Attributes
+
+| Attribute                   | Type    | Description                                                              | Example                            | Requirement Level | Stability    |
+| --------------------------- | ------- | ------------------------------------------------------------------------ | ---------------------------------- | ----------------- | --- |
+| `ai_agent.task.name`        | string  | Name of the task.                                                        | `Data Collection`                  | Required          | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
```
karthikscale3, line 87: do we also need a `ai_agent.task.input` in addition to these?

docs/ai-agent/ai-agent-spans.md
```diff
@@ -0,0 +1,185 @@
...
+| `ai_agent.agent.model`         | string | Underlying model powering the agent.       | `gpt-4`                          | Required          | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
+| `ai_agent.agent.tools`         | array  | List of tools available to the agent.      | `["Web Scraper", "Analyzer"]`    | Recommended       | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
+
+### Tools Attributes
+
+| Attribute                | Type   | Description                                  | Example           | Requirement Level | Stability    |
+| ------------------------ | ------ | -------------------------------------------- | ----------------- | ----------------- | --- |
+| `ai_agent.tool.name`     | string | Name of the tool utilized by the agent.      | `Web Scraper`     | Required          | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
+| `ai_agent.tool.function` | string | Specific function or capability of the tool. | `Data Extraction` | Recommended       | ![Experimental](https://img.shields.io/badge/-experimental-blue)    |
```
karthikscale3, line 80: do we also need a `ai_agent.tool.input` in addition to these?

karthikscale3:
Nice first draft @gyliu513 . thanks for starting this.

docs/ai-agent/ai-agent-spans.md
lmolkova: I'm curious if we can consolidate it under `gen_ai.system|user|tool|assistant.message` events rather than attirbutes

codefromthecrypt:
@gyliu513 I'm giving you unsolicited advice and being intentionally not specific to a change here, because I think more research would lead you to your own changes. That's always the best (in my mind). Hope it helps.

One of the main gains we had in re-organizing the llm now genai sig to have a space in otel-contrib python was to be able to practice specs before committing to it. I have seen this in practice done in java and it helps quite a bit.

* Are you keen on instrumenting a draft PR on some open source agent library you believe is valid for this semconv
* How about an example PII washed feed from agentic cloud provider data, which would translate

Another thing to guide is especially bookend timestamps sounds like a discussion that would have happened here in another domain (start_xx end_xx). Certainly, it happened way back in zipkin days with "cs" "cr" though these were separate events. Can you research some prior work in otel where a spec like this was accepted or denied?

gyliu513:
@codefromthecrypt good comment, thanks and happy holidays!

> How about an example PII washed feed from agentic cloud provider data, which would translate

Can you please share more detail for your comment here? 

I was now reviewing https://github.com/microsoft/opentelemetry-semantic-conventions/pull/3 and this PR really helped a lot, I will probably update my PR soon after new year based on https://github.com/microsoft/opentelemetry-semantic-conventions/pull/3.

codefromthecrypt:
@gyliu513 for this comment I made "How about an example PII washed feed from agentic cloud provider data, which would translate"

What I mean is that we most of the time assume the data is coming from the application. Like we instrument langchain or something and spans and metrics are collected directly from the app.

While I don't know what services exist, another way is cloud integration, where a platform is generating the signals. One example is AWS Bedrock, where you can get data regardless of what the developers do https://aws.amazon.com/blogs/mt/monitoring-generative-ai-applications-using-amazon-bedrock-and-amazon-cloudwatch-integration/

So, for this PR, I mean that if its scope is only for application instrumentation, then we should look at which frameworks we are considering and maybe a draft/experiment/proof of concept that exercises the specs you are making.

Beyond that, if you are thinking about a specific cloud integration (I don't know if you are), some sample data or documentation on what that agentic feed looks like could help us translate if the semantic conventions here are valid for it or not.

Does that help? If not you can also quiz me on slack, but anyway happy holidays!

model/gen-ai/spans.yaml
```diff
@@ -122,3 +122,135 @@ groups:
...
+      - ref: gen_ai.agent.tool.name
+        requirement_level: required
+  - id: trace.gen_ai.client.run_tool
+    type: span
+    stability: experimental
+    brief: >
+      Describes GenAI external tool creation and is usually applicable when working
+      with remote external services.
+    note: |
+      The `gen_ai.operation.name` MUST be `create_tool`.
```
Cirilla-zmh, line 208: I suspect there is a clerical error. Shouldn't the operation name be `run_tool`?

model/gen-ai/spans.yaml
```diff
@@ -122,3 +122,135 @@ groups:
...
+    extends: trace.gen_ai.client.create_task
+    attributes:
+      - ref: gen_ai.agent.tool.name
+        requirement_level: required
+  - id: trace.gen_ai.client.create_agent
+    type: span
+    stability: experimental
+    brief: >
+      Describes GenAI agent creation and is usually applicable when working
+      with remote agent services.
```
Cirilla-zmh, line 144: Honestly I’m not entirely sure what 'create' means in this line. Is it related to the dynamic creation process of the agent by some tools?

gyliu513 (commented):
@Cirilla-zmh Thanks for the comments, I think there will be some change for this PR, I will update this PR soon, as we have just merged Agent Semantic Convention https://github.com/open-telemetry/semantic-conventions/pull/1739 , will update this Agent Framework based on the agent semantic convention, thanks!

PR #1660, opened 2024-12-09
Add k8s metrics for jobs and cronjobs

ChrsMark:

## Changes

This PR adds metrics for k8s jobs and cronjobs that are already in use by the Opentelemetry Collector ([k8scluster](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/documentation.md) receiver).

- [`k8s.job.active_pods`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/documentation.md#k8sjobactive_pods) ([collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a49008324b95f428d518cb4372e2002adf9564b1/receiver/k8sclusterreceiver/internal/jobs/jobs.go#L16))
- [`k8s.job.failed_pods`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/documentation.md#k8sjobfailed_pods) ([collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a49008324b95f428d518cb4372e2002adf9564b1/receiver/k8sclusterreceiver/internal/jobs/jobs.go#L17))
- [`k8s.job.successful_pods`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/documentation.md#k8sjobsuccessful_pods) ([collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a49008324b95f428d518cb4372e2002adf9564b1/receiver/k8sclusterreceiver/internal/jobs/jobs.go#L18))
- [`k8s.job.desired_successful_pods`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/documentation.md#k8sjobdesired_successful_pods) ([collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a49008324b95f428d518cb4372e2002adf9564b1/receiver/k8sclusterreceiver/internal/jobs/jobs.go#L21))
- [`k8s.job.max_parallel_pods`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/documentation.md#k8sjobmax_parallel_pods) ([collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a49008324b95f428d518cb4372e2002adf9564b1/receiver/k8sclusterreceiver/internal/jobs/jobs.go#L24))
- [`k8s.cronjob.active_jobs`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/documentation.md#k8scronjobactive_jobs) ([collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a49008324b95f428d518cb4372e2002adf9564b1/receiver/k8sclusterreceiver/internal/cronjob/cronjobs.go#L22))


dashpole (commented):
It seems like these should all be UpDownCounters, rather than gauges.

ChrsMark:
Good catch @dashpole! My only concern is about the [k8s.job.desired_successful_pods](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/documentation.md#k8sjobdesired_successful_pods) and
[k8s.job.max_parallel_pods](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/documentation.md#k8sjobmax_parallel_pods). Should those be also UpDownCounters since they are set by users and are mostly desired-state values?

dashpole:
The main question is how they should be aggregated.  UpDownCounters aggregate by summing, gauges aggregate by averaging.  If I was aggregating `k8s.job.desired_successful_pods` k8s.job.max_parallel_pods` across a cluster, I would probably want to see the sum of all desired_successful_pods or the sum of max_parallel_pods, rather than the average of either.  So IMO those should both be UpDownCounters.

jinja2:
This is a very interesting point. These metrics being additive makes sense to me. But the data point type from the receiver is [gauge](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/metadata.yaml#L366). I think most instrumentation libraries export UpDownCounter as non-monotonic sums but I am not sure if this a standard. Would this cause some confusion for users? Should there be a way to clarify export and instrument type in cases as this?

dashpole:
> I think most instrumentation libraries export UpDownCounter as non-monotonic sums but I am not sure if this a standard.

Yes, that is correct.  It is a bit odd to specify an instrument type for metrics that are not recorded using an instrument, but that seems mostly editorial here.

IMO we should update the data point type in the receiver to be a non-monotonic sum.

trask:
> I think most instrumentation libraries export UpDownCounter as non-monotonic sums but I am not sure if this a standard.

I think this is the closest we have:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/otlp.md#additional-environment-variable-configuration

which specifies exporting UpDownCounter with Cumulative aggregation in all cases

lmolkova:
Do @semconv-system-approvers want to take a look?

docs/system/k8s-metrics.md
```diff
@@ -329,5 +329,161 @@ This metric is [recommended][MetricRecommended].
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+### Metric: `k8s.job.active_pods`
```
lmolkova, line 332: nit and can totally be done as a follow-up. 
There are around ~20 metrics in this file and it might be hard to navigate around. It's also useful to see all metrics in the same place, so I wonder if it's possible to add a TOC here

Ptal at [messaging](https://github.com/open-telemetry/semantic-conventions/blob/b1691ad003f25a7b6ee58cee3de4ef4805bff8f6/docs/messaging/messaging-metrics.md#semantic-conventions-for-messaging-client-metrics) or [db](https://github.com/open-telemetry/semantic-conventions/blob/b1691ad003f25a7b6ee58cee3de4ef4805bff8f6/docs/database/database-metrics.md) to see how convenient it could be.

ChrsMark: Yeap, it was already in place from previous PRs. After rebasing it's included here as well.

AlexanderWert (approved):

TylerHelmuth (approved):

lmolkova (approved):

Issue #1554, opened 2024-11-07
Create guideline for modeling state and phase as metrics


ChrsMark:
That was discussed today in [SemConv SIG meeting](https://docs.google.com/document/d/10xG7DNKWRhxNmFGt3yYd3980a9uwS8lMl2LvQL3VNK8/edit?tab=t.0 ) (Nov 4, 2024). It seems that this modeling could fit well https://github.com/open-telemetry/semantic-conventions/issues/1032 and in https://github.com/open-telemetry/semantic-conventions/pull/1212. @braydonk will prepare a proposal to put this as generic guidance in Semantic Conventions (thank you Braydon :)). Based on the outcome of this we can unblock the related PRs.

_Originally posted by @ChrsMark in https://github.com/open-telemetry/semantic-conventions/issues/1032#issuecomment-2455161611_

ref: https://github.com/open-telemetry/semantic-conventions/issues/1032#issuecomment-2420399295

ChrsMark:
I have filed https://github.com/open-telemetry/semantic-conventions/pull/1668 which follows the discussed guideline. We can use this this to verify the modeling's proposal.

PR #1668, opened 2024-12-10
Add k8s.namespace.phase metric and attribute

ChrsMark:

## Changes

This PR adds the `k8s.namespace.phase` as metric along with its respective attribute `k8s.namespace.phase`.

This metric is already in use by the Collector and specifically the [`k8scluster` receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/k8sclusterreceiver/documentation.md#k8snamespacephase).

The introduced attribute is new in order to follow the modeling that is already in use by `hw` metrics and `jmx` metrics mentioned at https://github.com/open-telemetry/semantic-conventions/issues/1032#issuecomment-2420399295 (see https://github.com/open-telemetry/semantic-conventions/issues/1554).

Collector implementation: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9dfa2f7813b11500d001622f3d8c1dd8d9ac58fd/receiver/k8sclusterreceiver/internal/namespace/namespaces.go#L14

K8s API ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#namespacestatus-v1-core


docs/attributes-registry/k8s.md
```diff
@@ -65,6 +66,19 @@ Which states:
...
+
+---
+
+`k8s.namespace.phase` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `active` | Active namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `terminating` | Terminating namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `unknown` | Unknow phase is set if phase is blank for some reason | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
dashpole, line 80:
```suggestion
| `unknown` | Unknown phase is set if phase is blank for some reason | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

docs/system/k8s-metrics.md
```diff
@@ -329,5 +329,42 @@ This metric is [recommended][MetricRecommended].
...
+
+---
+
+`k8s.namespace.phase` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `active` | Active namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `terminating` | Terminating namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `unknown` | Unknow phase is set if phase is blank for some reason | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
dashpole, line 362:
```suggestion
| `unknown` | Unknown phase is set if phase is blank for some reason | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

model/k8s/registry.yaml
```diff
@@ -214,3 +214,25 @@ groups:
...
+              value: 'active'
+              brief: "Active namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase)"
+              stability: experimental
+            - id: terminating
+              value: 'terminating'
+              brief: "Terminating namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase)"
+              stability: experimental
+            - id: unknown
+              value: "unknown"
+              brief: "Unknow phase is set if phase is blank for some reason"
```
dashpole, line 237:
```suggestion
              brief: "Unknown phase is set if phase is blank for some reason"
```

model/k8s/metrics.yaml
```diff
@@ -126,3 +126,15 @@ groups:
...
       - ref: network.io.direction
+
+  # k8s.namespace.* metrics
+  - id: metric.k8s.namespace.phase
+    type: metric
+    metric_name: k8s.namespace.phase
+    stability: experimental
+    brief: "Operational status: `1` (true) or `0` (false) for each of the possible phases"
+    instrument: updowncounter
+    unit: "1"
```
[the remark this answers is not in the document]
ChrsMark: I followed what https://github.com/open-telemetry/semantic-conventions/blob/a596f197d51c1fc373e2efa6153456123093eef5/model/hardware/common-metrics.yaml#L49-L65 defines :thinking: .

@open-telemetry/specs-semconv-maintainers any suggestion here?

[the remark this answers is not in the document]
ChrsMark: Thank's for the context here! 

If I remove the `unit` completely then `make fix` fails:

```
  × Invalid metric definition in "/home/weaver/source/k8s/metrics.yaml".
  │ group_id=`metric.k8s.namespace.phase`. This group contains a metric type
  │ but the unit is not set.
```

Should that be an empty string instead (`unit: ""`), or something else?

ping @open-telemetry/specs-semconv-approvers for validation here.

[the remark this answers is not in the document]
dashpole: Shouldn't it be {namespace}, since it represents a count of namespaces with the given phase?  If I aggregated it across a cluster, it would give me a count of namespaces broken down by phase.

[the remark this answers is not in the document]
lmolkova: oh, you're absolutely right @dashpole, I reacted to https://github.com/open-telemetry/semantic-conventions/pull/1668#discussion_r1888643369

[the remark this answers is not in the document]
ChrsMark: Thank you all for the feedback here! I changed the unit to `{namespace}`.

dashpole (approved):

model/k8s/metrics.yaml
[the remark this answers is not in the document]
ChrsMark: > It's not really a count of namespaces in given phase, it's rather a boolean value

Without being super confident on the why, I think it's kind of both. My perspective is that it can also by seen as how many namespaces with the name `default` are in `active` phase, 0 to 1.
In this, the instrument's type is `updowncounter` which makes sense. 

@lmolkova @dashpole feel free to provide any additional information here.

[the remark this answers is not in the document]
dashpole: If it isn't aggregated, then it can be viewed as either a t/f boolean (is it in this phase or not) or as a count of namespaces in that phase.  Both interpretations are valid. If it is aggregated, then it clearly can't be viewed as a boolean anymore (3 doesn't map to true or false).  Given that, I think `{namespace}` is more correct. In general, I don't think we should frame timeseries as having "boolean" value, as they become less coherent when aggregated.

jinja2 (approved):

AlexanderWert (approved):

TylerHelmuth (approved):

trask (commented):
is "phase" a common term in the context of k8s namespaces? (i.e. vs [status or state](https://github.com/open-telemetry/semantic-conventions/issues/308#issuecomment-2462999844))

ChrsMark:
> is "phase" a common term in the context of k8s namespaces? (i.e. vs [status or state](https://github.com/open-telemetry/semantic-conventions/issues/308#issuecomment-2462999844))

I believe so. It's what the [API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#namespacestatus-v1-core) defines and I see it's used in the [docs description](https://kubernetes.io/docs/tasks/administer-cluster/namespaces/) as well:

```
A namespace can be in one of two phases:

- Active the namespace is in use
- Terminating the namespace is being deleted, and can not be used for new objects
```

model/k8s/registry.yaml
```diff
@@ -214,3 +214,25 @@ groups:
...
+              value: 'active'
+              brief: "Active namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase)"
+              stability: experimental
+            - id: terminating
+              value: 'terminating'
+              brief: "Terminating namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase)"
+              stability: experimental
+            - id: unknown
+              value: "unknown"
+              brief: "Unknown phase is set if phase is blank for some reason"
```
lmolkova, line 237: is this necessary to fill it at all? Let's say k8s introduces another phase - then `unknown` becomes ambiguous.  

We can capture the phase as is (including blank value)

model/k8s/metrics.yaml
```diff
@@ -442,3 +442,18 @@ groups:
 
       This metric SHOULD, at a minimum, be reported against a
       [`k8s.cronjob`](../resource/k8s.md#cronjob) resource.
+
+  # k8s.namespace.* metrics
+  - id: metric.k8s.namespace.phase
```
lmolkova, line 447: Were any alternative metric names considered? It sounds a bit confusing. 
Suggesting: `k8s.namespace.status`, `k8s.namespace.count`

model/k8s/metrics.yaml
```diff
@@ -442,3 +442,18 @@ groups:
...
+    type: metric
+    metric_name: k8s.namespace.phase
+    stability: experimental
+    brief: "Operational status: `1` (true) or `0` (false) for each of the possible phases"
```
lmolkova, line 451: maybe

```suggestion
    brief: "Describes number of K8s namespaces that are currently in a given phase."
```

?

model/k8s/registry.yaml
```diff
@@ -214,3 +214,25 @@ groups:
               value: 'local'
               brief: "A [local](https://v1-29.docs.kubernetes.io/docs/concepts/storage/volumes/#local) volume"
```
lmolkova, line 215: nit:

it seems we use 1.30 version below, can we reference the same version consistently?

lmolkova (approved):

model/k8s/metrics.yaml
ChrsMark: Phase is the one that we had already in the Collector and seems to be the familiar term for this in K8s: https://github.com/open-telemetry/semantic-conventions/pull/1668#issuecomment-2598710823.

Kube-state-metrics also uses this term (`kube_namespace_status_phase`): https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/cluster/namespace-metrics.md

model/k8s/metrics.yaml
ChrsMark: Got this from https://github.com/open-telemetry/semantic-conventions/blob/294d2dbf00dc679fba70ad48c89bb8e3dce72f23/model/hardware/common-metrics.yaml#L53. I guess what you suggest is also accurate.

lmolkova (approved):

Issue #1184, opened 2024-06-26
[cicd] embed system attributes under cicd runners


adrielp:
do I understand that here we basically need to embed other attributes from os namespace?
there is an [open PR](https://github.com/open-telemetry/weaver/pull/212) to do this, maybe you can create a dedicated issue out of it and postpone until it will be used into semconv?

_Originally posted by @trisch-me in https://github.com/open-telemetry/semantic-conventions/pull/1075#discussion_r1654395856_

christophe-kamphaus-jemmic:
The goal of this issue is to have the capability to link Otel signals emitted by a runner (eg. host metrics, logs, events) to any jobs using a given runner.

Would this mean embedding system metrics under the cicd namespace or to add some cicd attributes (eg. `cicd.pipeline.run.id`) to the system metrics ?

christophe-kamphaus-jemmic:
Related to #1111

---

#### SemConv meeting notes 2024-09-16

> How could we make the link between a cicd run and metrics emitted by the runner of that run?

This could be a question related to the Entity Group.
How would this link be expressed?
* It could be a resource attribute attached to the metrics stating the `cicd.pipeline.run.id`
* It could be a resource attribute of the event that states which runner the pipeline run will execute on

Using resource attribute will work currently:
"Okay, the idea behind resource. One of the ideas is that all the telemetry generated from a particular, you know thing component would have the same set of attributes in it. So you can use that to tie, tie the knot and understand. This is the same source."

Using another way to link metrics and cicd run (eg. in an event) is not currently possible.
The Entity SIG will be working on answering that question.

> Would this mean embedding system metrics under the cicd namespace or to add some cicd attributes (eg. cicd.pipeline.run.id) to the system metrics ?

In general, it's better to just use the existing metrics without copying / embedding them into the cicd namespace.

The exception would be if we wanted to measure something that could not be expressed using the existing metrics and that relates to cicd, then we could think of embedding a metric in cicd namespace.

---

Using resource attributes is how I have implemented the link between cicd run and runner metrics in https://github.com/jenkinsci/opentelemetry-agent-metrics-plugin.

Downsides to this approach are that
* the otel generated is limited by the duration of the cicd run (eg. run/runner setup/tear down might not be covered)
* if a runner is executing multiple cicd runs in parallel, then otel must be emitted for each run separately

christophe-kamphaus-jemmic:
We should define `cicd` resource semconv.
[...]

PR #1681, opened 2024-12-13
Add CICD metrics

### LINKED ISSUE #1600: Add metrics for CICD job queues

christophe-kamphaus-jemmic:
Fixes #1600

## Changes

This PR adds metrics for CICD systems and related attributes.


.vscode/settings.json
```diff
@@ -14,5 +14,6 @@
             "model/**/*.yaml"
         ]
     },
-    "json.schemaDownload.enable": true
+    "json.schemaDownload.enable": true,
+    "markdown.extension.toc.levels": "2..6"
```
christophe-kamphaus-jemmic, line 18: This fixes an autoformatting issue on save in vscode with the markdown extension.
By default the autoformat includes the h1 heading leading to changed md files (eg to include `# Semantic Conventions for CICD Metrics` in the TOC).
This change aligns the autoformatting with the `markdown-toc --no-first-h1` command.
With this change autoformatting no longer causes changed md files.

docs/attributes-registry/cicd.md
```diff
@@ -13,11 +13,26 @@ This group describes attributes specific to pipelines within a Continuous Integr
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
 | <a id="cicd-pipeline-name" href="#cicd-pipeline-name">`cicd.pipeline.name`</a> | string | The human readable name of the pipeline within a CI/CD system. | `Build and Test`; `Lint`; `Deploy Go Project`; `deploy_to_environment` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="cicd-pipeline-result" href="#cicd-pipeline-result">`cicd.pipeline.result`</a> | string | The result of a pipeline run. | `success`; `failure`; `timeout` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
kuisathaverat, line 16: we could add skipped to the examples

```suggestion
| <a id="cicd-pipeline-result" href="#cicd-pipeline-result">`cicd.pipeline.result`</a> | string | The result of a pipeline run. | `success`; `failure`; `timeout`; `skipped` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```

docs/attributes-registry/cicd.md
```diff
@@ -13,11 +13,26 @@ This group describes attributes specific to pipelines within a Continuous Integr
...
 | <a id="cicd-pipeline-task-run-url-full" href="#cicd-pipeline-task-run-url-full">`cicd.pipeline.task.run.url.full`</a> | string | The [URL](https://wikipedia.org/wiki/URL) of the pipeline run providing the complete address in order to locate and identify the pipeline run. | `https://github.com/open-telemetry/semantic-conventions/actions/runs/9753949763/job/26920038674?pr=1075` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | <a id="cicd-pipeline-task-type" href="#cicd-pipeline-task-type">`cicd.pipeline.task.type`</a> | string | The type of the task within a pipeline. | `build`; `test`; `deploy` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="cicd-worker-state" href="#cicd-worker-state">`cicd.worker.state`</a> | string | The state of a CICD worker / agent. | `idle`; `busy` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="cicd-worker-type" href="#cicd-worker-type">`cicd.worker.type`</a> | string | The type of worker / agent used by the CICD system. | `vm`; `pod` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
kuisathaverat, line 23: It defines how you provision the worker/agent, instead of `type`, could we use `provider` or `provisioner`?

docs/attributes-registry/cicd.md
```diff
@@ -13,11 +13,26 @@ This group describes attributes specific to pipelines within a Continuous Integr
...
+
+---
+
+`cicd.pipeline.result` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
```
kuisathaverat, line 30: I think we should add `skipped` too

docs/attributes-registry/cicd.md
```diff
@@ -28,3 +43,21 @@ This group describes attributes specific to pipelines within a Continuous Integr
 | `build` | build | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `deploy` | deploy | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `test` | test | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+---
+
+`cicd.worker.state` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
kuisathaverat, line 49: I think we should add `down` or `disconnected`

docs/attributes-registry/cicd.md
```diff
@@ -28,3 +43,21 @@ This group describes attributes specific to pipelines within a Continuous Integr
...
+
+| Value  | Description | Stability |
+|---|---|---|
+| `busy` | The worker is performing work for the CICD system. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `idle` | The worker is not performing work for the CICD system. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+---
+
+`cicd.worker.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
kuisathaverat, line 58: I think we should add `container`

docs/cicd/cicd-metrics.md
```diff
@@ -23,6 +29,193 @@ linkTitle: CICD metrics
...
+| [`cicd.pipeline.name`](/docs/attributes-registry/cicd.md) | string | The human readable name of the pipeline within a CI/CD system. | `Build and Test`; `Lint`; `Deploy Go Project`; `deploy_to_environment` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| [`cicd.pipeline.result`](/docs/attributes-registry/cicd.md) | string | The result of a pipeline run. | `success`; `failure`; `timeout` | `Required` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+---
+
+`cicd.pipeline.result` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `cancelled` | The pipeline run was cancelled, eg. by a user manually cancelling the pipeline run. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
kuisathaverat, line 65: ditto `skipped`

docs/cicd/cicd-metrics.md
```diff
@@ -23,6 +29,193 @@ linkTitle: CICD metrics
...
+<!-- semconv metric.cicd.queue.length -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `cicd.queue.length` | UpDownCounter | `{pipeline_run}` | The number of pipeline runs waiting for their start of execution. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
kuisathaverat, line 113: Do we need a metric to collect the number of pipelines in execution?

docs/attributes-registry/cicd.md
christophe-kamphaus-jemmic: Added in a0443412ee94ab6c220c1a22247736b39e3e8fde

docs/cicd/cicd-metrics.md
christophe-kamphaus-jemmic: Added in a0443412ee94ab6c220c1a22247736b39e3e8fde

docs/attributes-registry/cicd.md
christophe-kamphaus-jemmic: Added in 7330043f03834f396be3a0d8800d1473ec63784a

docs/attributes-registry/cicd.md
```diff
@@ -13,11 +13,26 @@ This group describes attributes specific to pipelines within a Continuous Integr
...
 | <a id="cicd-pipeline-task-run-id" href="#cicd-pipeline-task-run-id">`cicd.pipeline.task.run.id`</a> | string | The unique identifier of a task run within a pipeline. | `12097` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | <a id="cicd-pipeline-task-run-url-full" href="#cicd-pipeline-task-run-url-full">`cicd.pipeline.task.run.url.full`</a> | string | The [URL](https://wikipedia.org/wiki/URL) of the pipeline run providing the complete address in order to locate and identify the pipeline run. | `https://github.com/open-telemetry/semantic-conventions/actions/runs/9753949763/job/26920038674?pr=1075` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | <a id="cicd-pipeline-task-type" href="#cicd-pipeline-task-type">`cicd.pipeline.task.type`</a> | string | The type of the task within a pipeline. | `build`; `test`; `deploy` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="cicd-worker-state" href="#cicd-worker-state">`cicd.worker.state`</a> | string | The state of a CICD worker / agent. | `idle`; `busy` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
christophe-kamphaus-jemmic: Added in 7330043f03834f396be3a0d8800d1473ec63784a

docs/attributes-registry/cicd.md
christophe-kamphaus-jemmic: Added in a0443412ee94ab6c220c1a22247736b39e3e8fde

docs/attributes-registry/cicd.md
christophe-kamphaus-jemmic: Does it make sense to distinguish between pod and container?

docs/cicd/cicd-metrics.md
christophe-kamphaus-jemmic: Since `cicd.pipeline.run.duration` only counts finished pipeline runs and runs might take a long time to finish it makes sense to have a metric for in progress builds.
Added in 3a31dd93497adeed0413c1cdebed1015856d69bd.

docs/attributes-registry/cicd.md
kuisathaverat: A pod can have several containers so a pod is a wrapper around containers.

Joibel: Argo Workflows can run a DAG inside a [Container Set](https://argo-workflows.readthedocs.io/en/latest/container-set-template/) so in this case the number of containers and number of pods would not match each other.
These containers can be `busy` and `idle` and switch state, so this has meaning here.
(Not: the conventional ("one pod for one job") pod spun up by Argo would actually have two containers (the `main` doing the work and `wait` which is the controller), although I'm not sure it would be useful to track that in the metrics.

docs/cicd/cicd-metrics.md
```diff
@@ -23,6 +30,221 @@ linkTitle: CICD metrics
...
+| `timeout` | A timeout caused the pipeline run to be interrupted. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `cicd.pipeline.run.executing`
+
+This metric is [required][MetricRecommended].
```
Joibel, line 80: `required` or `MetricRecommended`?

This mismatch occurs in the other locations too.

christophe-kamphaus-jemmic: For `cicd.pipeline.run.duration` I made it required since I thought it was one of the most important metrics for CICD systems. However I'm open to keep it recommended for now and increase it to required in a later PR after we get more feedback.

The other `[required][MetricRecommended]` were a copy/paste mistake. Fixed in 2e8ab020d76f30a871608564af94dc1852ba6675.

docs/attributes-registry/cicd.md
christophe-kamphaus-jemmic: I added `container` as an additional value in 66732c5f627a60e5ea1d0f337b0914ce33f1ebeb.

docs/cicd/cicd-metrics.md
Joibel: I was only pointing out the mismatch between the words and the link. I think your suggestion required vs recommended is fine 👍

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,65 @@
...
+    instrument: updowncounter
+    unit: "{pipeline_run}"
+    stability: experimental
+    attributes:
+      - ref: cicd.pipeline.name
+        requirement_level: recommended
+  - id: metric.cicd.worker.count
+    type: metric
+    metric_name: cicd.worker.count
+    brief: 'The number of workers available to the CICD system and/or busy.'
```
pablochacin, line 47: as the workers may be down (neither available nor busy) and we can potentially add more status later, I would use a more generic definition of the metrics
```suggestion
    brief: 'The number of workers on the CI/CD system by type and status.'
```

docs/attributes-registry/cicd.md
```diff
@@ -28,3 +44,25 @@ This group describes attributes specific to pipelines within a Continuous Integr
...
+|---|---|---|
+| `busy` | The worker is performing work for the CICD system. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `down` | The worker is not available to the CICD system (disconnected / down). | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `idle` | The worker is not performing work for the CICD system. It is available to the CICD system to perform work on. [1] | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1]:** Pipelines might have conditions on which workers they are able to run so not every worker might be available to every pipeline.
+
+---
+
+`cicd.worker.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
pablochacin, line 62: In many complex CI/CDs the workers are categorized by other attributes, such as the platform they run (`ubuntu`, `windows`) the tooling they have installed  (`java`, `go`) or other attributes. 

I lean towards leaving this attribute open (without pre-defined values). Also consider `class` as an alternative name to reflect its intention to group the workers into categories.

pablochacin:
Here we are introducing the concept of worker, but it seem disconnected from other elements of the domain.

For instance, pipelines are executed in workers. Shouldn't the metrics related to pipelines be somehow reference the workers? I'm not quite sure how this could be done. Just adding the worker type to the pipeline seems insufficient.

Maybe we should consider a `worker id` attribute. It seems useful to, for example, identify all errors in a pipeline comes from a faulty worker.

christophe-kamphaus-jemmic:
> For instance, pipelines are executed in workers. Shouldn't the metrics related to pipelines be somehow reference the workers? I'm not quite sure how this could be done. Just adding the worker type to the pipeline seems insufficient.
> 
> Maybe we should consider a `worker id` attribute. It seems useful to, for example, identify all errors in a pipeline comes from a faulty worker.

I did consider this, however we could easily run into cardinality issues if we add such an attribute to metrics (eg. if dynamic workers are used). If we add a `cicd.worker.id` attribute to metrics it will have to be opt-in and with a warning about the cardinatity.

For [CICD span conventions](https://github.com/open-telemetry/semantic-conventions/issues/1713) we should definitely keep this in mind. For example https://github.com/jenkinsci/opentelemetry-plugin emits span with an attribute `jenkins.computer.name`.

Also at some point I would like to be able to link host or container metrics to a pipeline run despite any cardinality issues as an opt-in feature: https://github.com/open-telemetry/semantic-conventions/issues/1184 For this it makes sense to use a cicd.worker.id attribute as well.

I think we should discuss this further on a separate Github issue.

docs/attributes-registry/cicd.md
christophe-kamphaus-jemmic: The worker might not be provisioned by the CICD system. It could just be connected to it as a static worker.

Here we should cover both dynamic provisioned as well as statically connected workers.
As suggested in https://github.com/open-telemetry/semantic-conventions/pull/1681#discussion_r1888239103 we should also leave the possibility open for CICD systems to to further subdivide workers.
I have rename this attribute to `cicd.worker.class` in 26d3064808bb0ead5fcd0db81af2fa04820f696a.

docs/attributes-registry/cicd.md
christophe-kamphaus-jemmic: I have rename this attribute to `cicd.worker.class` in 26d3064808bb0ead5fcd0db81af2fa04820f696a.

model/cicd/metrics.yaml
christophe-kamphaus-jemmic: I have updated the brief in dc09536748e52ae2edc25ab56b1335e171c53bbb.

christophe-kamphaus-jemmic: In a2202cd19505740dc7aa10bf2a6579c1d45a143e I adapted the brief to the rename of the attribute `type` to `class`.

kuisathaverat (approved):

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,65 @@
...
+    unit: "s"
+    stability: experimental
+    attributes:
+      - ref: cicd.pipeline.name
+        requirement_level: recommended
+  - id: metric.cicd.queue.length
+    type: metric
+    metric_name: cicd.queue.length
+    brief: 'The number of pipeline runs waiting for their start of execution.'
+    instrument: updowncounter
```
joaopgrassi, line 38: Thinking if an UpDownCounter is the best instrument type for this. Given for a updowncounter, instrumentations will have to decrement once the job leaves the queue and goes to execution. Is this how you thought this metric would be used? Otherwise a Gauge may also work. 

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#consistent-updowncounter-timeseries

christophe-kamphaus-jemmic: > Given for a updowncounter, instrumentations will have to decrement once the job leaves the queue and goes to execution. Is this how you thought this metric would be used? 

Yes, I indeed mean for it to work like that.
The `cicd.queue.length` metric is meant to count the pipeline runs that are about to start but have not yet started execution. Once they start executing they are meant to be decremented from this metric.

I was following https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/ to determine which instrument type to choose.

For the `cicd.queue.length` metric I thought to measure the queue length and that the values were additive over the `cicd.pipeline.name` attribute. As such an asynchronous UpDownCounter would be used.
Of course counting the jobs added/removed with a synchronous UpDownCounter would work as well.

I did shortly wonder if the `cicd.queue.length` might be redundant since I did define `cicd.queue.latency` as a histogram which would also contain a count of queued pipeline runs. However the histogram would only record the values once they have left the queue and thus the `cicd.queue.length` gives complementary insight to the histogram metric since it counts the pipeline runs which have not yet been recorded by the histogram.

joaopgrassi: All good, makes more sense now!

joaopgrassi (approved):

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,65 @@
...
+    unit: "{pipeline_run}"
```
adrielp, line 39:
```suggestion
    unit: "{pipeline_run}"
```

 should this be the unit for queue.length?

model/cicd/registry.yaml
```diff
@@ -75,3 +75,78 @@ groups:
         brief: >
           The type of the task within a pipeline.
         examples: ["build", "test", "deploy"]
+      - id: cicd.pipeline.result
+        type:
+          members:
```
adrielp, line 80: I'm wondering if we should be consistent with names here.

Like:

* succeeded
* failed
* errored
* timed-out
* cancelled
* skipped

Or 

* success
* failure
* error
* timeout
* skip
* cancel

similar verbiage to what is in CDEvents for parity https://github.com/cdevents/spec/blob/v0.4.1/core.md

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,65 @@
...
+    brief: 'The number of workers on the CICD system by class and status.'
+    instrument: updowncounter
+    unit: "{count}"
+    stability: experimental
+    attributes:
+      - ref: cicd.worker.class
+        requirement_level: required
+      - ref: cicd.worker.state
+        requirement_level: required
+  - id: metric.cicd.errors
```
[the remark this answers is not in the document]
lmolkova: It's ok to pluralize the last part on counters.

Here's the doc

https://github.com/open-telemetry/semantic-conventions/blob/a8fdb03c68d50a0ac67852f5e5a3df9ef32df90a/docs/general/naming.md?plain=1#L233-L237

> https://github.com/open-telemetry/semantic-conventions/blob/a8fdb03c68d50a0ac67852f5e5a3df9ef32df90a/docs/general/naming.md?plain=1#L248

It took quite a few attempts for me to understand this guidance - I think we need to make it easier to understand

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,65 @@
...
+    type: metric
+    metric_name: cicd.errors
+    brief: 'The number of errors in the controller of the CICD system.'
```
lmolkova, line 59: The `cicd.errors` name seems too broad - it seems it would cover all kinds of errors in all CI/CD system parts and then it stops being practical.

The first time I read it, I assumed it to count pipeline run errors, but it's probably something else.

Wonder if we could limit the scope of this metric to something reasonable. E.g. should it be `cicd.controller.errors`? I could envision we'd want to differentiate per component, e.g. have a `cicd.schedule.errors`, etc

model/cicd/registry.yaml
```diff
@@ -75,3 +75,78 @@ groups:
...
+              stability: experimental
+            - id: skipped
+              value: skipped
+              brief: "The pipeline run was skipped, eg. due to a precondition not being met."
+              stability: experimental
+        stability: experimental
+        brief: >
+          The result of a pipeline run.
+        examples: ["success", "failure", "timeout", "skipped"]
+      - id: cicd.worker.class
```
lmolkova, line 112: is there an example of when this attribute is necessary? 

It seems to introduce some classification where members are not mutually exclusive (container vs pod), but also don't describe worker type enough (e.g. would we need to differentiate container runtime).

I wonder if we can capture useful information in some other form.

E.g. if we have worker-level metrics, we'd associate them with a set of resource attributes covering all you need to know about workers.

model/cicd/metrics.yaml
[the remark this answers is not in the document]
adrielp: @lmolkova - thank you for finding this! i heavily appreciate the check here!

model/cicd/registry.yaml
christophe-kamphaus-jemmic: I will go with the 2nd option and rename
skipped → skip
cancelled → cancel

model/cicd/metrics.yaml
christophe-kamphaus-jemmic: What do you mean?

model/cicd/metrics.yaml
christophe-kamphaus-jemmic: I intended to count the errors from the CICD system (controller, scheduler, agent) in this metric and not errors from the pipeline run.

It is indeed good to be able to distinguish the different system components as well as errors from the pipeline runs.
Adriel also made a good point that it would be good to be able to count pipeline run errors separately from just the cicd.pipeline.run.duration with result=error because a single pipeline run might encounter several different errors.

Most likely this will be a derived metric, eg. by using a [span metrics](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector/) connector on the pipeline run spans or [count](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/countconnector/) connector on the controller logs.

I will define the following metrics to cover these:
`cicd.pipeline.run.errors` for errors encountered as part of the pipeline run execution
`cicd.system.errors` for errrors encountered in CICD system components with an attribute component (eg. scheduler, agent, controller, …)

model/cicd/registry.yaml
christophe-kamphaus-jemmic: I thought that it would be good to be able to subdivide the `cicd.worker.count`.

I agree that vm/pod/container might not be the most useful subdivision here and that this kind of information is better left to the worker-level metrics.

I will remove it for now to not block this PR.
When we have a more concrete use cases for worker.count subdivision we can revisit this discussion.

model/cicd/metrics.yaml
christophe-kamphaus-jemmic: Similar metrics:
`kestrel.queued_connections` unit: `{connection}`
`kestrel.queued_requests` unit: `{request}`
`dotnet.thread_pool.queue.length` unit: `{work_item}`

Since I will change the unit for `cicd.pipeline.run.executing` from `{pipeline_run}` to `{run}` I will apply this unit name change here for consistency.

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,65 @@
...
+    brief: 'The duration a pipeline run takes from being triggered to the start of execution.'
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    attributes:
+      - ref: cicd.pipeline.name
+        requirement_level: recommended
+  - id: metric.cicd.queue.length
+    type: metric
+    metric_name: cicd.queue.length
```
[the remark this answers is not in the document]
christophe-kamphaus-jemmic: Alternatively we can also call it `cicd.pipeline.run.queued`.
This will be more similar to the `cicd.pipeline.run.active` metric.

model/cicd/registry.yaml
mrueg: The second option seems to mix style (verbs, nouns).

Since this describes an event that has already happened, in my opinion the past tense makes more sense.

model/cicd/metrics.yaml
christophe-kamphaus-jemmic: Done in 89f6a8f2b7ea9253495c96293b2c7a1a154d348e

model/cicd/registry.yaml
christophe-kamphaus-jemmic: Rename done in 81c440745a18b5a0865535bee182e6d5168a376e

Good point @mrueg that we should have either all verbs or all nouns.

I think we should go with nouns since it most directly attributes a value to the result attribute. For example:
"The pipeline result is a success."
"The pipeline result is a timeout."

The CDEvents spec also went with nouns.

Skip is both a verb and a noun.
The only item that needs to be renamed to become a noun is
cancel → cancellation

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,65 @@
...
+    metric_name: cicd.pipeline.run.duration
+    brief: 'Duration of a pipeline run grouped by pipeline and result.'
+    instrument: histogram
+    unit: "s"
+    stability: experimental
+    attributes:
+      - ref: cicd.pipeline.name
+        requirement_level: required
+      - ref: cicd.pipeline.result
+        requirement_level: required
```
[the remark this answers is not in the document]
christophe-kamphaus-jemmic: Done in c1bc664e38ce765d6138c254094c39459f369108

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - id: metric.cicd.pipeline.run.executing
+    type: metric
+    metric_name: cicd.pipeline.run.executing
+    brief: 'The number of pipeline runs currently executing.'
+    instrument: updowncounter
+    unit: "{pipeline_run}"
```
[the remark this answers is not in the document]
christophe-kamphaus-jemmic: Done in 89f6a8f2b7ea9253495c96293b2c7a1a154d348e

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,65 @@
...
+    unit: "s"
+    stability: experimental
+    attributes:
+      - ref: cicd.pipeline.name
+        requirement_level: required
+      - ref: cicd.pipeline.result
+        requirement_level: required
+  - id: metric.cicd.pipeline.run.executing
+    type: metric
+    metric_name: cicd.pipeline.run.executing
```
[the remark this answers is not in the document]
christophe-kamphaus-jemmic: Renamed in 226b835b3f835985f7e222862cd9fd5bf0d9fa67

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,65 @@
...
+    instrument: updowncounter
+    unit: "{pipeline_run}"
+    stability: experimental
+    attributes:
+      - ref: cicd.pipeline.name
+        requirement_level: required
+  - id: metric.cicd.queue.latency
+    type: metric
+    metric_name: cicd.queue.latency
+    brief: 'The duration a pipeline run takes from being triggered to the start of execution.'
```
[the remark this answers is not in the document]
christophe-kamphaus-jemmic: Renamed in 55414eee861b1ac2690b6cd1679f82739c980304

model/cicd/metrics.yaml
christophe-kamphaus-jemmic: Done in fc48dd2337c83bbe02b0dcacd0202d97efe80ee1, 0be8d4d34f4a039e8e582dede6f6d78aa8a8b79b and 996e8096857d163d8981dae887af41641bafe7be.

model/cicd/registry.yaml
christophe-kamphaus-jemmic: Done in b3326385c794f476f1c02036b1486cf6049a4c47 and cf67fd2d746574a599e7c1c86fbf7c6b7fecc696

docs/attributes-registry/cicd.md
christophe-kamphaus-jemmic: I have now actually removed `cicd.worker.class` in b3326385c794f476f1c02036b1486cf6049a4c47.
See the discussion https://github.com/open-telemetry/semantic-conventions/pull/1681#discussion_r1907598237 for reference.

model/cicd/metrics.yaml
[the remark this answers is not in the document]
christophe-kamphaus-jemmic: Done in 702ca16e7208f66586f8e3fc9e4c36537acc30ce

model/cicd/registry.yaml
christophe-kamphaus-jemmic: Renamed in adf29b48f1fdbfd30aa0d77429c2e0a8ae6f3c35

carlosalberto (approved):

carlosalberto:
Overall LGTM. A small (non-blocking) question: were `cicd.pipeline.run.queued` and `cicd.pipeline.run.active` considered as a single metric, with a `state` or similar label? Not proposing that, just curious on whether that was considered, as they are very similar.

christophe-kamphaus-jemmic:
> Overall LGTM. A small (non-blocking) question: were `cicd.pipeline.run.queued` and `cicd.pipeline.run.active` considered as a single metric, with a `state` or similar label? Not proposing that, just curious on whether that was considered, as they are very similar.

No I did not consider it until now. It's only with the latest changes that the similarity between these two metrics became apparent.
Also the `cicd.pipeline.run.duration` and `cicd.pipeline.run.time_in_queue` seem very similar.

I'm open to combine these two metric pairs and distinguishing them with `phase` attribute if additional reviewers are in favor of this change.

model/cicd/registry.yaml
adrielp: Thanks for working through that!

model/cicd/registry.yaml
```diff
@@ -75,3 +75,69 @@ groups:
...
+            - id: error
+              value: error
+              brief: >-
+                The pipeline run failed due to an error in the CICD system, eg. due to the worker being killed.
+              stability: experimental
+            - id: timeout
+              value: timeout
+              brief: "A timeout caused the pipeline run to be interrupted."
+              stability: experimental
+            - id: cancellation
```
adrielp, line 100: cicd-metrics.md needs to be regenerated. Not seeing these there.

christophe-kamphaus-jemmic: > cicd-metrics.md needs to be regenerated. Not seeing these there.

What are you referring to?

I have just run `make fix` and no changes were shown. Just in case I have updated my branch with the latest changes from main.

If you were referring to me combining the 2 metric pairs as discussed in https://github.com/open-telemetry/semantic-conventions/pull/1681#issuecomment-2585169909 I was waiting for additional feedback before making that change.

adrielp: hmm, when I was looking last night I was seeing the old enums listed in the markdown. Maybe I had a cached version as it looks good now.

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,81 @@
...
+    brief: 'The number of workers on the CICD system by state.'
+    instrument: updowncounter
+    unit: "{count}"
+    stability: experimental
+    attributes:
+      - ref: cicd.worker.state
+        requirement_level: required
+  - id: metric.cicd.pipeline.run.errors
+    type: metric
+    metric_name: cicd.pipeline.run.errors
```
lmolkova, line 59: Could it be derived from `metric.cicd.pipeline.run.duration`? It could be if there is just one error per run, but I assume it's not the case, right?

lmolkova (approved):

lmolkova:
> Overall LGTM. A small (non-blocking) question: were `cicd.pipeline.run.queued` and `cicd.pipeline.run.active` considered as a single metric, with a `state` or similar label? Not proposing that, just curious on whether that was considered, as they are very similar.

Not a strong opinion, but I like the idea of merging `cicd.pipeline.run.queued` and `cicd.pipeline.run.active`  to something like `cicd.pipeline.run.count` with `cicd.pipeline.run.state` attribute.  

> Also the cicd.pipeline.run.duration and cicd.pipeline.run.time_in_queue seem very similar.

- these metrics could easily have different range (different order of magnitude) of values (e.g. time in queue is hopefully tiny and pipeline duration is measured in minutes or hours)
- you'd configure different alerts for them and different people would pay attention to their values (in my BigCo anecdotal experience, workers are managed by a separate team/org and I have little control over them, while pipeline and its duration is controlled by dev teams)

christophe-kamphaus-jemmic:
I have combined `cicd.pipeline.run.active` and `cicd.pipeline.run.queued` into `cicd.pipeline.run.active` distinguishing them with the `cicd.pipeline.run.state` attribute in 1d88f59ecece69ded793431160ba20b7445f38f3.

model/cicd/metrics.yaml
```diff
@@ -0,0 +1,73 @@
...
+    brief: 'The number of pipeline runs currently active in the system by state.'
+    instrument: updowncounter
+    unit: "{run}"
+    stability: experimental
+    attributes:
+      - ref: cicd.pipeline.name
+        requirement_level: required
+      - ref: cicd.pipeline.run.state
+        requirement_level: required
+  - id: metric.cicd.pipeline.run.time_in_queue
```
christophe-kamphaus-jemmic, line 29: I thought about combining `cicd.pipeline.run.duration` and `cicd.pipeline.run.time_in_queue` into `cicd.pipeline.run.duration` and distinguishing with the `cicd.pipeline.run.state` attribute.

The advantages of combining the metrics are that
* the metric centers around the pipeline run
* is vendor-neutral (not supposing the existing of a queue)
* extensible by using additional states

The disadvantages are
* the issue Liudmila mentioned of the different orders of magnitude of time_in_queue vs duration spent executing.
This could be worked around by filtering on the `cicd.pipeline.run.state` attribute for display on chart and alert query, ie having separate charts and alerts for both cases.
* both metrics differ in the `cicd.pipeline.result`: either `cicd.pipeline.result` would need to be made conditionally required (ie only being set for the `cicd.pipeline.run.state` in which the `cicd.pipeline.result` becomes known) or a pending result would need to be added.

model/cicd/metrics.yaml
christophe-kamphaus-jemmic: There might be errors in a pipeline run that are non fatal, ie they are suppressed or in a parallel stage multiple stages could have an error.
Ie this error count might not be the same as the count of `metric.cicd.pipeline.run.duration` with run result `failure`.

We did think that this metric might either be derived as a span metric from a run trace or reported directly by the CICD system controller.

model/cicd/metrics.yaml
kuisathaverat: If I understand correctly, an execution of a pipeline would be something like this:

```json
{
    "cicd": {
        "run": {
            "duration": 60,
            "state": "pending"
        }
    }
}
```

```json
{
    "cicd": {
        "run": {
            "duration": 60,
            "state": "starting"
        }
    }
}
```

```json
{
    "cicd": {
        "run": {
            "duration": 60,
            "state": "executing"
        }
    }
}
```

```json
{
    "cicd": {
        "run": {
            "duration": 60,
            "state": "finalizing"
        }
    }
}
```

christophe-kamphaus-jemmic: I was thinking that the duration would record only the time spent in that state.
So we could have the following records being recorded in the `cicd.pipeline.run.duration` metric:
### Success
```
attributes:
  cicd.pipeline:
    name: Example1
    run.state: pending
value: 3
```
```
attributes:
  cicd.pipeline:
    name: Example1
    run.state: executing
    result: success
value: 60
```
```
attributes:
  cicd.pipeline:
    name: Example1
    run.state: finalizing
value: 2
```
### Failure & no clean up
```
attributes:
  cicd.pipeline:
    name: Example1
    run.state: pending
value: 2
```
```
attributes:
  cicd.pipeline:
    name: Example1
    run.state: executing
    result: failure
  error.type: Task non-zero exit code
value: 50
```
### Cancellation during initialization & clean up (no execution)
```
attributes:
  cicd.pipeline:
    name: Example1
    run.state: pending
    result: cancellation
  error.type: User cancellation
value: 1
```
```
attributes:
  cicd.pipeline:
    name: Example1
    run.state: finalizing
value: 4
```

christophe-kamphaus-jemmic: Feedback from SemConv meeting:

Either combine both the `cicd.pipeline.run.active` and `cicd.pipeline.run.duration` or neither.

The issue of different orders of magnitude can be worked around by defining buckets carefully or fixed by using exponential histograms.

---

I have made the metric merge into `cicd.pipeline.run.duration` in f244911800b6b5772b02f057f253551615ade49c

Please take a look at how I made `cicd.pipeline.result` **conditionally_required**. Does that look ok?

model/cicd/metrics.yaml
christophe-kamphaus-jemmic: I have added this as a note in 02e044a5f38bc91ce80a14d267e79ee16c7cb7ca.

model/cicd/registry.yaml
```diff
@@ -75,3 +95,69 @@ groups:
...
+              stability: experimental
+        stability: experimental
+        brief: >
+          The result of a pipeline run.
+        examples: ["success", "failure", "timeout", "skipped"]
+      - id: cicd.worker.state
+        type:
+          members:
+            - id: idle
+              value: idle
```
adrielp, line 136: Does `idle` also connote `available`?

christophe-kamphaus-jemmic: image
`Idle` seems appropriate to me as a state name.
Should `down` be renamed to `offline`?

adrielp: Available is a common name in semconv https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemlinuxmemoryavailable 

Also, https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/k8sclusterreceiver/metadata.yaml - common in k8s when talking about nodes/etc.  (though to be clear, I couldn't find these metrics ANYWHERE in semconv -- they are widely used in the industry though).

I think `idle` may not mean `available`. It would have to be idle & ready to be available. But we can say available means idle enough to accept new work.

👍🏼 to offline

christophe-kamphaus-jemmic: Rename done in 88e1c25990b74605848cfbc57c2a88a799d30273.

adrielp (approved):

PR #1574, opened 2024-11-13
Extend GenAI system to support IBM Watsonx AI and AWS Bedrock

gyliu513:
Added IBM Watsonx AI and AWS Bedrock to the list of possible values that the `gen_ai.system` attribute can take.

## Changes

The following are added:

- IBM Watsonx AI
- AWS Bedrock

Please provide a brief description of the changes here.

Note: if the PR is touching an area that is not listed in the [existing areas](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/README.md), or the area does not have sufficient [domain experts coverage](https://github.com/open-telemetry/semantic-conventions/blob/main/.github/CODEOWNERS), the PR might be tagged as [experts needed](https://github.com/open-telemetry/semantic-conventions/labels/experts%20needed) and move slowly until experts are identified.

## Merge requirement checklist

* [X] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [x] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

PR #1683, opened 2024-12-14
[chore] Consolidate genai.request.seed

gyliu513:
Fixes #

This is a follow up for https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3081/files#r1884746832

/cc @lmolkova @lzchen 

## Changes

Please provide a brief description of the changes here.

Note: if the PR is touching an area that is not listed in the [existing areas](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/README.md), or the area does not have sufficient [domain experts coverage](https://github.com/open-telemetry/semantic-conventions/blob/main/.github/CODEOWNERS), the PR might be tagged as [experts needed](https://github.com/open-telemetry/semantic-conventions/labels/experts%20needed) and move slowly until experts are identified.

## Merge requirement checklist

* [ ] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [ ] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

Issue #608, opened 2023-12-13
Guidance on product/project name inside attribute/metric name


lmolkova:
We should provide a guidance on product/project name being fully qualified (or not). We should keep the same pattern in different semconvs. 
Examples of inconsistencies:
- `db.cosmosdb` (Azure),   `db.dynamodb` (AWS), `db.couchdb` (under Apache umbrella),  `db.spanner` (GCP), `db2` (IBM), `hanadb` (SAP HANA)  etc and corresponding values in the `db.system` enum
- we also have `aws_sqs`, `gcp_pubsub`, `azure_servicebus` in the `messaging.system` enum

We should provide a guidance on how to represent multiple words:
- lowercase product name (`couchdb`) 
- multiple namespaces vs single namespace (e.g. `messaging.aws_sqs.destination.custom_attr` vs `messaging.aws.sqs.destination.custom_attr`)

Misc discrepancies: 
- `mssqlcompact` (Microsoft SQL Server Compact) should probably become `mssql_compact` 

Related: #602, #599

alanwest:
With regards to representing multiple words, can a product's branding provide a guide? For example, use `dynamodb`, `couchdb`, and `cosmos_db` because they are respectively branded DynamoDB, CouchDB, and Cosmos DB.

trask:
It would be nice if whatever the enum is, e.g. `cosmosdb`, that is also the namespace, e.g. `db.cosmosdb.*` for product-specific attributes

trask:
do we think we need `mssqlcompact`? that doesn't seem like something we'd necessarily know from the client side, and on the server side it could potentially be a resource attribute describing the "edition"

lmolkova:
Based on the messaging SIG discussion on 6/13, none of the controversial systems are part of the initial stability (kafka + RabbitMQ), so removing the stability blocker label.

lmolkova:
Been discussing it in the scope of #1734. There are competing consistency goals when it comes to project names. Let's explore them:

## 1. Stay consistent with external project/product/system/etc name whenever possible

The guidance would be:
- Use a registered trademark (wordmark) or another 'official' name
- When it's ambiguous (e.g. caché), it needs to be disambiguated, for example, prefixed with a company name (intersystems_cache)
 
Non-controversial examples:
- `mongodb`
- `postgresql`
- `cassandra`
- `ibm.mq`
- `oracle.db` (`oracledb`, `oracle.database`  and other possible variations)

Controversial examples
[...]

PR #1684, opened 2024-12-14
[chore] Update vertex_ai to vertex.ai

gyliu513:
Fixes #

This is a follow up for https://github.com/open-telemetry/semantic-conventions/pull/1574#issuecomment-2539951918

/cc @lmolkova

## Changes


lmolkova:
Thanks @gyliu513 !

Please make sure to add changelog entry.

/cc @aabmass - wdyt about this rename?

model/gen-ai/registry.yaml
```diff
@@ -13,9 +13,9 @@ groups:
               stability: experimental
               value: "openai"
               brief: 'OpenAI'
-            - id: vertex_ai
```
gyliu513, line 16: @lmolkova do you know if we can also need to mark `vertex_ai` as deprecated?  if so, how to make it? This is a value and not an attribute, did not find any example for this, thanks!

gyliu513:
I will update https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml after https://github.com/open-telemetry/semantic-conventions/pull/1683 got merged.

model/gen-ai/registry.yaml
```diff
@@ -13,9 +13,9 @@ groups:
...
+            - id: vertex.ai
               stability: experimental
-              value: "vertex_ai"
+              value: "vertex.ai"
```
aabmass, line 18: I can't find any evidence of just "Vertex" without "AI" in GCP's docs, so `vertex_ai` actually seems appropriate to me. I can't imagine any other `vertex.*` values.

If that reasoning holds water to keep it as `vertex_ai`, I can verify with the team.

aabmass, line 18: While we're here, should we prefix with `gcp.` like we have for some other vendors?

lmolkova: we're really inconsistent on prefixing things with cloud name (not intentionally).
E.g. we have [`spanner`](https://github.com/open-telemetry/semantic-conventions/blob/0d843e4cb842c816559072d89ede1fd5544f1657/docs/attributes-registry/db.md?plain=1#L149) and it's ok since there is just one spanner DB.

We have [`gcp_pubsub`](https://github.com/open-telemetry/semantic-conventions/blob/0d843e4cb842c816559072d89ede1fd5544f1657/docs/messaging/gcp-pubsub.md) for messaging because `pubsub` is too generic..

We also have some `gcp` [attributes](https://github.com/open-telemetry/semantic-conventions/blob/0d843e4cb842c816559072d89ede1fd5544f1657/docs/attributes-registry/gcp.md) and [resources](https://github.com/open-telemetry/semantic-conventions/tree/0d843e4cb842c816559072d89ede1fd5544f1657/docs/resource/cloud-provider/gcp)

Both `vertex.ai` or `vertext_ai` are unique enough - they don't **have to** have a prefix similarly to `spanner`. If it does, it should be `gcp`.

So what should we do:
1. `spanner`, `vertex_ai`, `gcp_pubsub`
2. `spanner`, `vertex_ai`, `gcp.pubsub`
3. `gcp.spanner`, `gcp.vertex_ai`, `gcp.pubsub`
4. `gcp_spanner`, `gcp_vertex_ai`, `gcp_pubsub`
5. ...
?

Speaking from Azure side, we're likely to stay where we are: `cosmosdb`, `eventhubs`, `az.ai.inference` (use product name when it's unique enough, use `az.` namespace when it's not). Reasons: it's nice to be consistent, but some of these things are de-facto stable and are used wide enough to make breaking really problematic.

General issue - https://github.com/open-telemetry/semantic-conventions/issues/608

cc @jsuereth if he has any thoughts

aabmass: There's also the [`cloud.platform` values](https://github.com/open-telemetry/semantic-conventions/blob/0d843e4cb842c816559072d89ede1fd5544f1657/docs/attributes-registry/cloud.md#:~:text=cloud.platform%20has%20the%20following%20list%20of%20well%2Dknown%20values.%20If%20one%20of%20them%20applies%2C%20then%20the%20respective%20value%20MUST%20be%20used%3B%20otherwise%2C%20a%20custom%20value%20MAY%20be%20used.) which look like `azure_functions`, `gcp_cloud_functions `.

lmolkova: I'm like 80% confident we can rename `azure_functions` to `az.functions`. Let me try

joaopgrassi: I also believe we should not change this to be a namespace. Vertex AI seems to be the product name.

joaopgrassi (changes requested):
I'd leave the final to the experts and product folks, but to my understanding, Vertex AI is a trademark https://trademarks.justia.com/907/18/vertex-90718213.html, so making it `vertex.ai` is not correct. 

I also looked and I think the addition of `watson.ai` in this PR was wrong https://github.com/open-telemetry/semantic-conventions/pull/1574. The same applies, that is the product name, not a namespace. 

Blocking it for now to avoid things getting merged.

gyliu513:
@joaopgrassi watsonx is actually right, please check the following links:

- watsonx.ai https://www.ibm.com/products/watsonx-ai
- watsonx.data https://www.ibm.com/products/watsonx-data
- watsonx.governance https://www.ibm.com/products/watsonx-governance

joaopgrassi:
Yes, I also checked those. The problem here is that I personally see the dot in the name as part of the brand name. In semconv a dot is to be used for namespacing things, so using the dot just because the brand name uses it is wrong IMO because it clashes with the deep namespace meaning the dot has in semconv.

model/gen-ai/registry.yaml
joaopgrassi: OK I re-read and thought about it, giving this is only changing the enum value all if fine.

joaopgrassi (commented):
Removing my blocker.

I got caught up and passed through me that this is just renaming the enum values of `gen_ai.system`. So this is fine. 

What triggered my concern is that it can get us into iffy territory if we start using the product name "dot" in semconv, as we won't know its meaning. 

It is of course fine if the dot is used for namespacing purposes (to add more attributes under it, or sub-namespaces).

model/gen-ai/registry.yaml
lmolkova: I'd leave things as they are for now. We need a more consistent strategy across all semconv (we have `gcp_pubsub` and `spanner` today) - it's tracked in https://github.com/open-telemetry/semantic-conventions/issues/608

Issue #1669, opened 2024-12-10
Support complex attribute values on logs/events in semantic conventions


lmolkova:
LogRecord attributes are special - [they support complex attribute types](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.39.0/specification/logs/data-model.md#field-attributes) unlike spans/metrics/resources that support [standard attributes only](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.39.0/specification/common/README.md#attribute) (primitives and arrays of primitives).  Note that the latter is the API limitation (not an [OTLP](https://github.com/open-telemetry/opentelemetry-proto/blob/653d80276d98bdf6cd3218dadc03ee9a30dac815/opentelemetry/proto/common/v1/common.proto#L64) one)

In semconv we assume that all attributes are cross-signal and therefore are limited to 'standard' attribute types. 

The proposal is to lift this limitation and allow defining complex attributes in semantic conventions with the following limitations:
- they cannot be referenced on spans, metrics, or resources/entities. TBD: check with profiles
  - this can be solved via the semconv policy (given that attribute types are limited on the OTel API surface and not the protocol, we might not want to hardcode it in weaver)
 - we should define if/how such attributes would interact with the standard attributes. E.g. are `geo.location.lat=x` and  `geo.location.lon=y` attributes set on a span equivalent to `geo.location={"lat": x, "lon": y}` set on the log record?

This would help with the following problems:
- https://github.com/open-telemetry/semantic-conventions/issues/1651 - we can have one property bag (attributes) to put all the event details into. Standard attributes can be used for correlation across all signals. Complex attributes could be reused across logs preserving semantic meaning  (which we don't have today). 
- it'd introduce reusable types discussed in https://github.com/open-telemetry/semantic-conventions/issues/1264 

Related: https://github.com/open-telemetry/opentelemetry-specification/pull/4334, https://github.com/open-telemetry/opentelemetry-specification/pull/2888

lmolkova:
Discussed at the spec call on 12/17

Some feedback/notes:
- Backends might leverage complex attributes if they start to popup, but it does not seem like there is a burning need for them
- Prometheus would have a problem. Overall, complex attributes on metrics do not seem to be useful
- We can keep the door of complex attributes on spans open (and special-case metrics or support flattening for them)
- We can make the first step with complex attributes on logs (as a testing ground)

The first step:
- we should define anything that should be set across signals as a standard attribute
[...]

PR #1689, opened 2024-12-16
Add value_type to feature flag event

dyladan:
This adds a `value_type` which references the body field `value` in feature flag evaluation events. The `value` field has an undefined type, allowing any type to be included. This means processing pipelines have to sniff the type before applying any logic. In some cases this may even result in ambiguous cases. For example if a processing pipeline stringifies the value, it may be impossible to tell the difference between boolean `false` and string `"false"`.


docs/attributes-registry/feature-flag.md
```diff
@@ -41,3 +42,19 @@ For example, the variant `red` maybe be used for the value `#c05543`.
 | `static` | The resolved value is static (no dynamic evaluation). | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `targeting_match` | The resolved value was the result of a dynamic evaluation, such as a rule or specific user-targeting. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `unknown` | The reason for the resolved value could not be determined. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+---
+
+`feature_flag.evaluation.value_type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
toddbaert, line 48: `feature_flag.evaluation.value_type` makes sense, but I could also see `feature_flag.evaluation.value.type`, based on some of the other attribute keys. Is there some basic heuristic you applied in this case?

docs/feature-flags/feature-flags-logs.md
```diff
@@ -83,14 +84,16 @@ A `feature_flag.evaluation` event SHOULD be emitted whenever a feature flag valu
 
 **[2] `error.type`:** If and only if an error occurred during flag evaluation.
 
-**[3] `feature_flag.variant`:** A semantic identifier, commonly referred to as a variant, provides a means
+**[3] `feature_flag.evaluation.value_type`:** If and only if `value` field is supplied in the event body
```
toddbaert, line 87: `value_type` could certainly be valuable independently of a `value`.

Making this required when the `value` is defined seems like the minimum, but this field also seems independently valuable. I suppose we could consider making it generally required, though I'm leaning against that:

We've emphasized `variant` over `value` for understanding the result of the evaluation, the implicit reasoning here being that you can reliably use the `variant` to understand the `value` that was returned. The same argument could be made for the `value_type`, I suppose.

dyladan: What I have is consistent in that it still emphasizes variant over value and value_type. I'm not sure what you could learn from value_type that isn't implied by variant and version? We could make it "required if [...] else optional" if you think it provides value on its own.

toddbaert: My comment might have been unclear; I was agreeing with the proposal as written, but just pointing out the alternative for consideration.

toddbaert:
Non-binding approval from me :heavy_check_mark:

docs/attributes-registry/feature-flag.md
dyladan: I might actually like `feature_flag.evaluation.value.type` better because that leaves things like `feature_flag.evaluation.value.<size|format>` et cetera open for future consideration.

docs/attributes-registry/feature-flag.md
```diff
@@ -41,3 +42,19 @@ For example, the variant `red` maybe be used for the value `#c05543`.
 | `static` | The resolved value is static (no dynamic evaluation). | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `targeting_match` | The resolved value was the result of a dynamic evaluation, such as a rule or specific user-targeting. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `unknown` | The reason for the resolved value could not be determined. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+---
+
+`feature_flag.evaluation.value.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
lmolkova, line 48: It's not for this PR, but we're seriously considering allowing complex attribute types **instead of body**

https://github.com/open-telemetry/semantic-conventions/issues/1669 https://github.com/open-telemetry/opentelemetry-specification/pull/4334.

Wonder what your thoughts are

lmolkova:
@dyladan could you please elaborate on 

> This means processing pipelines have to sniff the type before applying any logic.

the type information is a part of OTLP (e.g. here's a map attribute [json example](https://github.com/open-telemetry/opentelemetry-proto/blob/2bd940b2b77c1ab57c27166af21384906da7bb2b/examples/logs.json#L43)) 

```json
{
  "key": "map.attribute",
  "value": {
    "kvlistValue": {  # type info
      "values": [
        {
          "key": "some.map.key",
          "value": {
            "stringValue": "some value"   # type info in stringValue
          }
        }
      ]
    }
  }
```

body should be represented in the same way. So I wonder why is it a problem for the pipeline - it should know the type, no?

kinyoklion (approved):

beeme1mr (approved):

lukas-reining (approved):

dyladan:
I think for now we can leave this off and rely on the type info carried by OTLP. Easier to add later than it is to remove.

PR #1691, opened 2024-12-17
[chore] Add x-amzn-requestid as source for aws.request_id

Flarna:
## Changes

The response headers used by AWS SDK are `x-amzn-requestid`, `x-amzn-request-id` or `x-amz-request-id`.

see e.g. [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-logging.html) and [here](https://docs.aws.amazon.com/AmazonS3/latest/API/get-request-ids.html) or [here](https://github.com/aws/aws-sdk-js-v3/blob/7fee20d761655ebcdae7ec103fb61b1b695c7d02/clients/client-s3/src/protocols/Aws_restXml.ts#L9961C59-L9961C76).


model/aws/registry.yaml
```diff
@@ -8,7 +8,7 @@ groups:
       - id: aws.request_id
         type: string
         stability: experimental
-        brief: "The AWS request ID as returned in the response headers `x-amz-request-id` or `x-amz-requestid`."
+        brief: "The AWS request ID as returned in the response headers `x-amz-request-id` or `x-amzn-requestid`."
```
lmolkova, line 11: it seems both forms are supported

https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/API_dataplane_GetMedia.html

> x-amz-RequestId HTTP header – if you want to report an issue to AWS, the support team can better diagnose the problem if given the Request Id.

So maybe 
```suggestion
        brief: "The AWS request ID as returned in the response headers `x-amz-request-id`, `x-amz-requestid`, or `x-amzn-requestid`."
```

?

Flarna: Seems so yes, AWS SDK source code confirms. Updated and corrected commit message + PR description.

lmolkova (approved):
Thanks!

trask (approved):

PR #1697, opened 2024-12-18
Add cross-file anchor link check (use lychee for markdown link checking)

### LINKED ISSUE #332: Markdown link check runs on entire repo

### LINKED ISSUE #1009: Add cross-file anchor link check
See https://github.com/open-telemetry/semantic-conventions/issues/1008#issuecomment-2097036448
            
With docfx v3 disappearing into thin air, we have no tooling that validates that relative links to files in the repo with anchors exist (e.g. `/docs/http/http-spans.md#setting-serveraddress-and-serverport-attributes`)

- markdownlint validates anchors within the same file (`#setting-serveraddress-and-serverport-attributes`). Same as docfx v2
- markdown-link-check validates presence of file, but not the anchor there

https://github.com/theoludwig/markdownlint-rule-relative-links seems to be able to validate cross-file anchors, but only with relative path (`./docs/http/http-spans.md#setting-serveraddress-and-serverport-attributes`) which we can't do because of md generation.

trask:
Fixes #1009

I think... Fixes #332 because lychee has better retry behavior


.github/workflows/checks.yml
```diff
@@ -43,6 +43,13 @@ jobs:
     - name: install dependencies
       run: npm install
 
+    - name: Restore lychee cache
+      uses: actions/cache@v4
+      with:
+        path: .lycheecache
+        key: cache-lychee-${{ github.sha }}
+        restore-keys: cache-lychee-
+
```
trask, line 52: oops didn't mean to commit this
```suggestion
```

lmolkova (approved):

joaopgrassi (approved):

joaopgrassi:
Going to leave it open, so folks have the chance to look at it after the holidays.

Issue #1692, opened 2024-12-17
make check-format failed


gyliu513:
### Area(s)

area:docs

## Description

`check-format` is a sub-command for `check`, and `check-format` always failed.

```console
root@gyliu-d61:~/github.com/open-telemetry/semantic-conventions# make check-format
npm run check:format

> check:format
> npm run _check:format || (echo '[help] Run: npm run format'; exit 1)


> _check:format
> npx prettier --check .

Checking formatting...
[warn] docs/attributes-registry/android.md
[warn] docs/attributes-registry/artifact.md
[warn] docs/attributes-registry/aspnetcore.md
[warn] docs/attributes-registry/aws.md
[warn] docs/attributes-registry/azure.md
[warn] docs/attributes-registry/browser.md
[warn] docs/attributes-registry/cicd.md
[warn] docs/attributes-registry/client.md
[warn] docs/attributes-registry/cloud.md
[warn] docs/attributes-registry/cloudevents.md
[warn] docs/attributes-registry/cloudfoundry.md
[warn] docs/attributes-registry/code.md
[warn] docs/attributes-registry/container.md
[warn] docs/attributes-registry/cpu.md
[warn] docs/attributes-registry/db.md
[warn] docs/attributes-registry/deployment.md
[warn] docs/attributes-registry/destination.md
[warn] docs/attributes-registry/device.md
[warn] docs/attributes-registry/disk.md
[warn] docs/attributes-registry/dns.md
[warn] docs/attributes-registry/dotnet.md
[warn] docs/attributes-registry/enduser.md
[warn] docs/attributes-registry/error.md
[warn] docs/attributes-registry/event.md
[warn] docs/attributes-registry/exception.md
[warn] docs/attributes-registry/faas.md
[warn] docs/attributes-registry/feature-flag.md
[warn] docs/attributes-registry/file.md
[warn] docs/attributes-registry/gcp.md
[warn] docs/attributes-registry/gen-ai.md
[warn] docs/attributes-registry/geo.md
[warn] docs/attributes-registry/go.md
[warn] docs/attributes-registry/graphql.md
[warn] docs/attributes-registry/hardware.md
[warn] docs/attributes-registry/heroku.md
[warn] docs/attributes-registry/host.md
[warn] docs/attributes-registry/http.md
[warn] docs/attributes-registry/ios.md
[warn] docs/attributes-registry/jvm.md
[warn] docs/attributes-registry/k8s.md
[...]

PR #1699, opened 2024-12-18
[chore] Enbled make fix can fix table format

gyliu513:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1692

## Changes


docs/attributes-registry/android.md
```diff
@@ -13,16 +13,16 @@
 
 The Android platform on which the Android application is running.
 
-| Attribute | Type | Description | Examples | Stability |
-|---|---|---|---|---|
+| Attribute                                                                            | Type   | Description                                                                                                                                                                                                                               | Examples   | Stability                                                        |
```
lmolkova, line 16: these files are auto-generated, we should not modify them (and I think whitespaces in jinja are too complicated to format them in this way)

lmolkova (commented):
I believe we don't use format checks in the workflow and can probably remove this check from the makefile.

Wonder what @open-telemetry/specs-semconv-maintainers think

docs/attributes-registry/android.md
gyliu513: @lmolkova `make fix-format` will update the tables

gyliu513:
> I believe we don't use format checks in the workflows/CI and can probably remove this check from the makefile.

Yes, we can discuss if we need `make fix-format` or not, I have added `fix-format` to CI process in this PR.

lmolkova:
> Yes, we can discuss if we need `make fix-format` or not, I have added `fix-format` to CI process in this PR.

as you can see it breaks other checks (registry validation). Personally, I don't see a benefit in format check in MD + yaml repo with MD and markdown lint checks.

gyliu513:
Thanks @lmolkova , with that, shall we remove both `check-format` and `fix-format` from makefile? It is really annoying when running `make check`.

lmolkova:
> Thanks @lmolkova , with that, shall we remove both `check-format` and `fix-format` from makefile? It is really annoying when running `make check`.

I don't use `make check` since it takes forever, so I wonder if other @open-telemetry/specs-semconv-maintainers have any thoughts.

trask:
> I don't use `make check` since it takes forever, so I wonder if other @open-telemetry/specs-semconv-maintainers have any thoughts.

fwiw, the only thing I ever use is `make table-generation attribute-registry-generation`

gyliu513:
Hi @lmolkova and @trask , I removed check-format and fix-format in Makefile, hope this is ok, thanks!

joaopgrassi:
I think we discussed in some other PR/Issue that it would be good to start enforcing the line limit in markdown, but that's probably a bigger endeavor. For ex we would have to avoid modifying any part of the auto-generated content (tables etc). 

For now removing it sounds good to me. I also don't use it.

lmolkova (approved):

trask (approved):

PR #193, opened 2023-07-13
[editorial] markdownlint in less then 1.5 sec

chalin:
- Adds a Gulp task named `lint-md` to run `markdownlint`
  - The **task runs in less than 1.5 seconds** over the same files as the `markdownlint` make target, which takes much longer, when run from the command line
- Fixes the errors reported by `markdownlint` Gulp task. I'm unsure why, with the same config, the CLI misses those errors
- Replaces the call to the `markdownlint` CLI with the Gulp task in the `markdownlint` job

PR #1706, opened 2024-12-20
Update markdownlint

trask:
This is an alternative to #193

And closes #1703

I think the reason it was so slow before #193 is because it was running markdownlint in a loop over the files instead of passing the files into a single invocation of markdownlint

`time make markdownlint` on my machine reports:
* `main`: 8.8 seconds
* this PR: 8.8 seconds

I also verified that the errors CLI was missing prior to #193 are not missed anymore.

I think this is an easier path forward (at least for me) compared to figuring out how to fix #1703 (see also https://github.com/open-telemetry/opentelemetry.io/issues/5767)

cc @chalin


lmolkova (approved):
Nice!

joaopgrassi (approved):

joaopgrassi:
There's a new PR out now by renovate https://github.com/open-telemetry/semantic-conventions/pull/1717, probably can also close it by this one

lmolkova:
> There's a new PR out now by renovate #1717, probably can also close it by this one

this PR fixes the check (there is a breaking change), so merging it.

chalin (commented):
Very nice simplification! 🙌🏻

PR #1707, opened 2024-12-20
First stab at documenting the art of defining semantic conventions

lmolkova:
A lot of todos, which I'd like to tackle in follow up PRs (hopefully with the help of others)


docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+<!-- tocstop -->
+
+This document describes requirements, recommendations, and best practices on how to define conventions
+for the new areas or make substantial changes to the existing ones.
+
+## Defining new conventions
+
+- New conventions MUST have a group of codeowners. See [project management](https://github.com/open-telemetry/community/blob/main/project-management.md) for more details.
+  <!-- TODO: add CI check for CODEOWNERS file (when a new area is added) -->
+- New conventions SHOULD be defined in YAML files. See [YAML Model for Semantic Conventions](/model/README.md) for the details.
```
joaopgrassi, line 31: With this SHOULD I assume your intention is that things should be added to yaml files where possible, but also it's OK to add conventions as just text in markdown files? For example how to record span names? 

If this is right, maybe we can say something like this, to make it more clear?

> - New conventions MUST be defined in YAML files. Conventions that cannot be recorded as part of the YAML model MAY be added in the corresponding markdown file (e.g., conventions for span name)

docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Best practices
+
+#### Defining attributes
+
+Reuse existing attributes when possible. Look through [existing conventions](/docs/attributes-registry/) for similar areas,
+check out [common attributes](/docs/general/attributes.md).
+Semantic conventions authors are encouraged to use attributes from different namespaces.
+
+Introduce new attributes when there is a clear use-case for them. Consider adding
```
joaopgrassi, line 43: I guess we can make this shorter with just: 

```suggestion
Consider adding new attributes when the following apply:
```

docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+them if most of the following apply:
+
+- They provide a clear benefit to end users by enhancing telemetry.
+- You plan to use the attribute in spans, metrics, events, resources, or other telemetry signals.
```
joaopgrassi, line 47:
```suggestion
- There is a clear plan to use the attribute in spans, metrics, events, resources, or other telemetry signals.
```
I'm never sure if we should avoid using "you" in conventions 🤔

docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+- The attribute will be utilized in instrumentations.
```
joaopgrassi, line 48:
```suggestion
- The attribute will be utilized in instrumentations.
```
If the above is true, isn't this already somewhat implicit? (that they are used in spans/metrics/logs etc? 

I think what we have been trying to push when folks want to add new things, is that we'd like to see implementations of them, so maybe we can also say:

> - There is a clear plan on how these attributes will be used by instrumentations

docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+
+Postpone adding new attributes if their benefit to telemetry is not yet clear.
```
joaopgrassi, line 50: Maybe here we can say something like this instead? 

> Semantic convention mantainers may reject the addition of new attributes if their benefit/usecase is not yet clear.

joaopgrassi (commented):
This is a great start!

CONTRIBUTING.md
```diff
@@ -62,6 +63,18 @@ key, but non-obvious, aspects:
 
 Please make sure all Pull Requests are compliant with these rules!
 
+### Suggesting conventions for a new areas
+
+Defining semantic conventions involves a group of people who are familiar with the domain,
+are involved into instrumentation efforts, and are committed to be the point of contact for
```
braydonk, line 69:
```suggestion
are involved with instrumentation efforts, and are committed to be the point of contact for
```

CONTRIBUTING.md
```diff
@@ -62,6 +63,18 @@ key, but non-obvious, aspects:
...
+pull requests, issues, and questions in this area.
+
+Check out [project management](https://github.com/open-telemetry/community/blob/main/project-management.md)
+for the details on how to start.
+
+Refer to the [How to define new conventions](/docs/general/how-to-define-semantic-conventions.md) for
```
braydonk, line 75:
```suggestion
Refer to the [How to define new conventions](/docs/general/how-to-define-semantic-conventions.md) document for
```

braydonk (commented):
Minor grammar feedback

docs/general/how-to-define-semantic-conventions.md
lmolkova: I want to put MUST here, but there are plenty of cases where we can't. Span names is a good example, but we have other things we can't yet express: references to metrics, complex attribute types, resources that are expected to be associated with telemetry, etc.

I use SHOULD for new conventions and MUST as a prereq to stabilization. 
I feel REQUIRING yaml for experimental stuff is not necessary, but also, it feels wrong to require something none of the conventions fully implement. 

Maybe once the tooling covers 90% of scenarios we can revisit it and be more strict?

docs/general/how-to-define-semantic-conventions.md
lmolkova: > If the above is true, isn't this already somewhat implicit? (that they are used in spans/metrics/logs etc?

I'd like to avoid someone defining conventions without planning to actually implement it. Changed the wording slightly to make it more clear.

Also, some of the attributes are not really part of any specific telemetry signal. E.g. `code`, `thread`, `enduser`, `session` - it's something you stamp on any span/log/etc to enrich it. For those we'd probably not require referencing them on a span/metric/etc, but would ask about instrumentations and some plaintext description of how they are used.

> There is a clear plan on how these attributes will be used by instrumentations

I like this, thanks!

joaopgrassi (approved):

trisch-me:
While I’m not opposed to this PR and appreciate the idea of having a clear path for defining semantic conventions, I would like to raise a concern about [donation of ECS into Otel](https://github.com/open-telemetry/oteps/blob/main/text/0199-support-elastic-common-schema-in-opentelemetry.md), where we still have a significant number of new namespaces / new fields to be added to the Otel and this could potentially slow down the merger. Should we consider establishing a short-circuit path to streamline the donation process?

docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+
+Reuse existing attributes when possible. Look through [existing conventions](/docs/attributes-registry/) for similar areas,
+check out [common attributes](/docs/general/attributes.md).
+Semantic conventions authors are encouraged to use attributes from different namespaces.
+
+Consider adding a new attribute when the following apply:
+
+- It provides a clear benefit to end users by enhancing telemetry.
+- There is a clear plan to use the attributes when defining spans, metrics, events, resources, or other telemetry signals in semantic conventions.
+- There is a clear plan on how these attributes will be used by instrumentations
```
trisch-me, line 47: are these criteria connected with `AND`, should they all apply? I would recommend to specify it explicitly to avoid confusion

lmolkova:
@trisch-me This PR outlines path to success and best practices. Most of the language used there is non-normative.
And it's something we already practice when reviewing the code - we already ask to define conventions for signals along with the attributes.

CONTRIBUTING.md
```diff
@@ -82,6 +83,18 @@ and helps to keep conventions consistent and backward compatible.
 Want to define your own conventions outside this repo while building on OTel’s?
 Come help us [decentralize semantic conventions](https://github.com/open-telemetry/weaver/issues/215).
 
+### Suggesting conventions for a new areas
```
trask, line 86:
```suggestion
### Suggesting conventions for a new area
```

or

```suggestion
### Suggesting conventions for new areas
```

CONTRIBUTING.md
```diff
@@ -82,6 +83,18 @@ and helps to keep conventions consistent and backward compatible.
...
+
+Defining semantic conventions involves a group of people who are familiar with the domain,
```
trask, line 88:
```suggestion
Defining semantic conventions requires a group of people who are familiar with the domain,
```

docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+
+Semantic convention maintainers may reject the addition of a new attribute if its benefits
+and use-cases are not yet clear.
+
+When defining a new attribute:
+
+- Follow the [naming guidance](/docs/general/naming.md)
+- Provide descriptive `brief` and `note` sections to clearly explain what the attribute represents.
+  - If the attribute represents a common concept documented externally, include relevant links.
+    For example, always link to attributes related to concepts defined in RFCs or other standards.
```
trask, line 57: (?)
```suggestion
    For example, always link to concepts defined in RFCs or other standards.
```

docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+          > [!WARNING]
+          >
+          > This attribute contains sensitive (PII) information.
+    ```
+
+- Use the appropriate [attribute type](https://github.com/open-telemetry/weaver/blob/main/schemas/semconv-syntax.md#type)
+  - If the value has a reasonably short (open or closed) set of possible values, define it as an enum.
+  - If the value is a timestamp, record it as a string in ISO 8601 format.
+  - For arrays of primitives, use the array type. Avoid recording arrays as a single string.
+  - Use the template type to define attributes with variable names (only the last segment of the name should be dynamic).
```
trask, line 78: maybe
```suggestion
  - Use the template type to define attributes with dynamic names (only the last segment of the name should be dynamic).
```

docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,151 @@
...
+- Avoid defining attributes with potentially unbounded values, such as strings longer than
+  1 KB or arrays with more than 1,000 elements. Such values should be recorded in the log or event body instead.
+
+Consider the scope of the attribute and how it may evolve in the future:
+
+- When defining an attribute for a narrow use case, think about potential broader use cases.
+  For example, if creating a system-specific attribute, evaluate whether other systems
+  in the same domain might need a similar attribute in the future.
+
+  Similarly, instead of defining a simple boolean flag like `foo.is_error`, consider a
```
trask, line 92: is the `is_*` convention used for boolean flags?

trask (approved):

docs/general/how-to-define-semantic-conventions.md
lmolkova: removed - I don't think we have a convention - it's an anti-example anyway

AlexanderWert (approved):

xrmx (approved):

docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,155 @@
...
+  - If the value has a reasonably short (open or closed) set of possible values, define it as an enum.
+  - If the value is a timestamp, record it as a string in ISO 8601 format.
+  - For arrays of primitives, use the array type. Avoid recording arrays as a single string.
+  - Use the template type to define attributes with dynamic names (only the last segment of the name should be dynamic).
+    This is useful for capturing user-defined key-value pairs, such as HTTP headers.
+  - Represent complex values as a set of flat attributes. <!-- This may change, check out https://github.com/open-telemetry/semantic-conventions/issues/1669 to monitor the progress -->
+- Define new attributes with `development` stability.
+- Provide realistic examples
+- Avoid defining attributes with potentially unbounded values, such as strings longer than
+  1 KB or arrays with more than 1,000 elements. Such values should be recorded in the log or event body instead.
```
lmolkova, line 88:
```suggestion
  1 KB or arrays with more than 1,000 elements. Such values should be recorded in the log or event body instead. <!-- This may change, check out https://github.com/open-telemetry/semantic-conventions/issues/1651 to monitor the progress -->
```

docs/general/how-to-define-semantic-conventions.md
```diff
@@ -0,0 +1,155 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: How To Define New Semantic Conventions
+aliases: [how-to-define-new-semantic-conventions]
```
lmolkova, line 3:
```suggestion
linkTitle: How to define new semantic conventions
```

PR #1698, opened 2024-12-18
Change `azure_` and `az.` to `azure.`  across all conventions

lmolkova:
Azure-specific attributes/signals use `azure.` prefix to denote they belong to `azure` namespace.

We should 

- Consolidate `cloud.platform` enum values:
    - `azure_aks` -> `azure.aks`
    - `azure_app_service` -> `azure.app_service`
    - `azure_container_apps` -> `azure.container_apps`
    - `azure_container_instances` -> `azure.container_instances`
    - `azure_functions` -> `azure.functions`
    - `azure_openshift` -> `azure.openshift`
    - `azure_vm` -> `azure.vm`
- Consolidate `gen_ai.system` enum values:
   - `az.ai.inference` -> `azure.ai.inference`
   - `az.ai.openai` ->  -> `azure.ai.openai`
- Azure SDK attributes
  - `az.namespace` -> `azure.resource_provider.namespace` 
  - `az.service_request_id`  -> `azure.service.request.id` 
- Azure log event
   - `az.resource.log` -> `azure.resource.log`

Related to https://github.com/open-telemetry/semantic-conventions/pull/1684#discussion_r1889500899, #608, #1708, #1613

PS: We should do similar renames for AWS and GCP, `cloud.provider` -> `cloud.provider.name`, `cloud.platform` -> ?

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [ ] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

Issue #1403, opened 2024-09-12
Clarify OS specific system attributes/metrics namespace


rogercoll:
### Area(s)

area:system

### Is your change request related to a problem? Please describe.

Should system semantic conventions attributes and/or metrics specific to an OS include it as a namespace? At the moment, there are example of both options:

## With {os} namespace

- [linux.memory.slab.state](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/linux.md#linux-memory-attributes) attribute
- [system.{os} metrics](https://github.com/open-telemetry/semantic-conventions/blob/3146dc01dc4ab477549f51ceed3b8f8cb98b5829/specification/metrics/semantic_conventions/system-metrics.md#systemos---os-specific-system-metrics); an example would be `system.linux.memory.available`

## Without the {os} namespace

- [process.executable.build_id.gnu](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/process.md#process-attributes) attribute; gnu specific
- `process.owner`; Windows specific
- `process.open_file_descriptor.count`; Linux specific metric

### Describe the solution you'd like

Whether all attributes/metrics specific to an OS contain the OS name in its namespace or not.

This concern was raised during a @open-telemetry/semconv-system-approvers SIG while considering the implications of using the `linux` prefix for the `process.cgroup` attribute.

ChrsMark:
Looking into this again, I find it weird that we encode the `linux` information within the metric name.
If a metric is supported by an OS or not is an implementation detail that might change in the future.

We hit the same in some host's resource attributes, like at the `host.cpu.model.id` which the collector skips: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/5601dfb28d905d0802f2e450cad6a2cc80029b6b/processor/resourcedetectionprocessor/internal/system/system.go#L202-L207. Should we encode the os information in these too?

@open-telemetry/semconv-system-approvers check me here but wouldn't be just fine if we document in which operating systems a metric is supported and keep the name unified?

braydonk:
# Semantic Conventions Meeting Sept 23, 2024 Summary

This was discussed in the semantic conventions meeting on Monday Sept 23 2024. The following is the summary.

## Dual precedence

We started by discussing the current split of precedence between the OS being present in the name or not. There are multiple system-exclusive attributes and metrics that do and don't have OS in the name.

* linux.memory.slab.state - linux only
[...]

Issue #169, opened 2023-07-06
Consistent page title form and capitalization


chalin:
This is what the side nav looks like for the semconv pages, as (soon to be) published on the OTel website:

> <img width="408" alt="image" src="https://github.com/open-telemetry/semantic-conventions/assets/4140793/dc7eb52a-2160-48ab-a6d6-09027d839b82">

~(1) I'd like to suggest that all section titles (at this level and recurisively too) be of the form:~

- ~Xxx Semantic Conventions~

~instead of "Semantic Conventions for Yyy".~

(2) There should be consistency in capitalization. I'm ok with whichever convention you use:Title Case or Sentence case -- the latter being the Google dev docs recommendation).

(3) Some titles are "Semantic Convention", singular. This should be fixed.

WDYT @jsuereth @joaopgrassi

/cc @svrnm @cartermp

chalin:
Here is what it would look like if the titles were fixed:

> <img width="260" alt="image" src="https://github.com/open-telemetry/semantic-conventions/assets/4140793/0a7c7ba9-6999-4826-a3d3-70d5a0c3fdaf">

This is taken from the preview https://deploy-preview-2982--opentelemetry.netlify.app/docs/specs/semconv/ of

- https://github.com/open-telemetry/opentelemetry.io/pull/2982

PR #1708, opened 2024-12-21
Add system-specific naming guidance

### LINKED ISSUE #1494: Document `*.client.{system}.*` ordering in the metric naming
We need to decide on a pattern for system-specific client/server metrics.

The pattern for client/server is `{domain}.client.metric_name` and `{domain}.server.metric_name` (e.g. `http.client.request.duration`  and `http.server.request.duration`).

For db/messaging/rpc/genai/etc, we'll need to take into account system-specific metrics and have a consistent place for the system identifier.

E.g.
1. `db.cosmosdb.client.*` and `db.cosmosdb.server.*` 
vs
2.  `db.client.cosmosdb.*` and `db.server.cosmosdb.*`

Option 1 (`*.{system}.client.*`) is aligned better with **attribute naming pattern** (`{domain}.{system}.attribute_name`). 
Option 2 (`*.client.{system}.*`) is aligned better with **metric naming pattern** (`{domain}.client.{system}.metric_name`). 

There is no obvious advantage of one options versus the other, but keeping things consistent is important. Let's pick one option and document it as a guidance.

lmolkova:
See https://github.com/open-telemetry/semantic-conventions/pull/1734 which implements this guidance for databases.

Fixes #1494, #608 

Related: 
- #1698
- #1711

Documents:
- Systems-specific attribute names follow `{system_name}.{thing}.{property}` pattern 
- Systems-specific metric names follow `{system_name}.client|server.{metric_name}`
- Metrics for remote call operations should include some indication if they are `client` or `server`
- System names should not be ambiguous and match the project name or, in most cases, are prefixed with company/org name


docs/general/naming.md
```diff
@@ -309,4 +338,37 @@ freely. For example, `system.paging.faults` and `system.network.packets`.
...
+- `kestrel.connection.duration` - here, `kestrel` is the name of the web server,
+  so no additional indication is necessary.
+
+### System-specific metrics
+
+When a metric is specific to a system (e.g., project, provider, product) in a given domain,
+the system name should be included in the instrument name using the pattern:
+`{domain}.{client|server}.{system}.*.{property}` pattern.
+
+For example, `db.client.cosmosdb.operation.request_charge`
```
lmolkova, line 363: the difference with `signalr.server.connection.duration` below makes me think if we should simplify it and do `cosmosdb.client.operation.request_charge`. What's the benefit of having `db` in front of it?

docs/general/naming.md
```diff
@@ -222,9 +225,37 @@ Examples:
...
+
+**Status**: [Development][DocumentStatus]
+
+When an attribute is specific to a particular system (e.g., project, provider, product),
+the system name should be used in the attribute name, following the pattern:
+`{optional domain}.{system}.*.{property}` pattern.
+
+Examples:
+
+- `db.cassandra.consistency_level` - Describes the consistency level property
```
lmolkova, line 238: Similar to https://github.com/open-telemetry/semantic-conventions/pull/1708/files#r1894523030

what's the benefit of having `db` in front of `cassandra` here? why not `cassandra.consistency_level`?

lmolkova: Let's say we have some redis-specific attributes (and redis can be used as a DB and as a messaging system), would we define `db.redis` and `messaging.redis` ? probably not, it should be just `redis`.

docs/general/naming.md
lmolkova: summoning @open-telemetry/semconv-system-approvers 

It's the same question as for `system.linux` vs `linux` metrics in https://github.com/open-telemetry/semantic-conventions/issues/1403 or https://github.com/open-telemetry/semantic-conventions/pull/1618

docs/general/naming.md
joaopgrassi: Yeah in the particular case of `redis` it is clear that the prefix becomes kinda irrelevant.. but then again it is nice to have the well-known prefixes for the "main" semconv areas.. 🤔  I'm torn on this one.

lmolkova: > but then again it is nice to have the well-known prefixes for the "main" semconv areas

could you elaborate? Is it nice to have `db.cassandra.*` ? Can you think of cases when it'd be useful?

docs/general/naming.md
```diff
@@ -309,4 +315,106 @@ freely. For example, `system.paging.faults` and `system.network.packets`.
...
+
+When adding new system to the semantic conventions, follow these principles in descending order of priority:
+
+1. The system name SHOULD adhere to the general attribute naming guidelines outlined in this document,
+   as it will be used as a namespace in [system-specific attribute names](#system-specific-attributes).
+
+2. The system name SHOULD unambiguously identify this specific product or project.
+
+   For example, use `gcp.pubsub` or `oracle.db`. Avoid generic names like `pubsub`,
+   which could refer to multiple messaging products, or `oracle` which lacks specificity.
```
trask, line 365:
```suggestion
   which could refer to multiple messaging products, or `oracle` which could refer to
   multiple Oracle products.
```

docs/general/naming.md
```diff
@@ -309,4 +315,106 @@ freely. For example, `system.paging.faults` and `system.network.packets`.
...
+   - Independent projects that do not belong to a specific company, such as Apache Foundation projects like
+     `kafka` or `cassandra`.
+   - Products with names similar to the owning company, such as `mongodb` or `elasticsearch`
+   - Widely recognized products that are popular outside their company's ecosystem.
+     These products often have trademarks without the company name and have
+     their own top-level domain (e.g. `spring` or `mysql`).
+
+3. In other cases, the system name SHOULD be prefixed with the company (organization,
+   division, or group) name. For cloud services, the name SHOULD use the
+   corresponding cloud provider name. For example, use `ibm.db2` or `aws.dynamodb`.
```
trask, line 378: `ibm.db2` might not be the best example since some (me) don't think of db2 primarily as a cloud service (i'm probably wrong at this point...), and it's not clear to me whether ibm cloud will have its own namespace (`ibm.cloud`?)

```suggestion
   corresponding cloud provider name. For example, use `aws.dynamodb` or `azure.cosmosdb`.
```

docs/general/naming.md
```diff
@@ -309,4 +315,106 @@ freely. For example, `system.paging.faults` and `system.network.packets`.
...
+
+For example, `azure.cosmosdb.client.operation.request_charge`
+
+<!-- todo not yet - https://github.com/open-telemetry/semantic-conventions/pull/1711 -->
+<!-- update when https://github.com/open-telemetry/semantic-conventions/pull/1613 is merged -->
+
+The value of the [`*.system.name`](#attributes-that-describe-system-name) (or similar)
+attribute MUST match system specific metric namespace.
+
+For example, both `azure.cosmosdb.client.operation.request_charge` metric and `db.system.name=azure.cosmosdb`
```
trask, line 417:
```suggestion
For example, both the `azure.cosmosdb.client.operation.request_charge` metric and the `db.system.name=azure.cosmosdb`
```

trask (approved):
really appreciate all of the guidance documentation!

docs/general/naming.md
```diff
@@ -309,4 +315,106 @@ freely. For example, `system.paging.faults` and `system.network.packets`.
...
+For example, both `cassandra.consistency.level` attribute name and `db.system.name=cassandra`
+ use the same system name (`cassandra`).
+
+### System-specific Metrics
+
+When a metric is specific to a system (project, provider, product) in a given area,
+the system name should be included in the instrument name following the
+`{system_name}.*.{metric_name}` pattern.
+
+For example, `azure.cosmosdb.client.operation.request_charge`
```
lmolkova, line 409:
```suggestion
For example, `azure.cosmosdb.client.operation.request.charge`
```

docs/general/naming.md
ChrsMark: The difference with `process.linux.cgroup` is that the name describes effectively a cgroup that belongs to a linux process. While `cosmosdb.client.operation.request_charge` can stand on its own without the need of `db` in front.

Maybe we can try and standardize this difference as part of this doc?

docs/general/naming.md
ChrsMark: From my perspective it's only 2 possible cases here:

1. the metric is cassandra specific so just the `cassandra.*` namespace is fine
2. the metric can be generalised across different database technologies and hence it's just `db.consistency_level`.

For databases, cloud providers, messaging systems this approach looks valid.
I'm not sure if it's totally equivalent for `system`, `process` and `linux` areas (see also https://github.com/open-telemetry/semantic-conventions/pull/1708#discussion_r1931790425).

docs/general/naming.md
```diff
@@ -309,4 +315,104 @@ freely. For example, `system.paging.faults` and `system.network.packets`.
...
+
+### System-specific Attributes
+
+When an attribute is specific to a particular system (project, provider, product),
+the corresponding attribute name SHOULD start with the system name following the
+`{system_name}.*.{property}` pattern.
+
+Examples:
+
+- `cassandra.consistency.level` - Describes the consistency level property
```
ChrsMark, line 392: It'd be nice to explain here that if a metric/attribute can be generalized across different implementations/technologies then `db.consistency.level` could be used.

In general if we could collect more examples here to illustrate the different areas/use-cases that would be great.

docs/general/naming.md
trask: what about `linux.process.cgroup`?

docs/general/naming.md
lmolkova: absolutely, this is what I'm trying to do in https://github.com/open-telemetry/semantic-conventions/pull/1707 and would be interested in your input

docs/general/naming.md
lmolkova: Examples:

- [Imaginary] `linux.distro.name` (`linux.distro.version`, `linux.kernel.version`) instead of `os.linux.distro.name`
- [Real] [`jvm.system.cpu.load_1m`](https://github.com/open-telemetry/semantic-conventions/blob/a72c5eea291b947bcd4916b70feb3b2e6b4eae9a/model/jvm/metrics-experimental.yaml#L24) instead of `system.jvm..cpu.load_1m``
- [Real] [`dotnet.process.cpu.count`](https://github.com/open-telemetry/semantic-conventions/blob/a72c5eea291b947bcd4916b70feb3b2e6b4eae9a/model/dotnet/runtime-metrics.yaml#L4) instead of `process.dotnet.cpu.count`
 -  [Real, but different pattern] `linux.memory.slab.usage` (or `linux.system.memory.slab.usage`) instead of [`system.linux.memory.slab.usage`](https://github.com/open-telemetry/semantic-conventions/blob/a72c5eea291b947bcd4916b70feb3b2e6b4eae9a/model/system/metrics.yaml#L396)


Justification for `linux.process.cgroup` - cgroup is a property of a process, not linux. E.g. we have `process.pid`. What if there was a linux-specific process id, would we do `linux.process.id` or `process.linux.id` ? I think former is more friendly.

And since `linux`-specific metrics fall into 'specialist case' bucket, they don't necessarily need to be discoverable when you type `process` and let autocomplete guide you

docs/general/naming.md
lmolkova: let's continue the discussion there. We should not drop `process` or `system` when they are meaningful, but we should do `linux` instead of `os.linux`

lmolkova:
Discussed at the SemConv SIG call on 5/12 - let's document the exception for system semconv along with the reasoning (related to https://github.com/open-telemetry/semantic-conventions/issues/2245)

docs/general/naming.md
lmolkova: I added `Known exceptions` section and mentioned system.{os} and process.{os} there. As discussed with @braydonk during semconv call, we should follow-up with the PR explaining the reasoning for it.

docs/general/naming.md
```diff
@@ -258,18 +294,20 @@ For example if we have a namespace `system.process` which contains all metrics r
...
 
 Counters SHOULD NOT append `_total` either because then their meaning will
 be confusing in delta backends.
 
-### Instrument Naming
+### Instrument naming
 
 **Status**: [Development][DocumentStatus]
 
+**Status**: [Experimental][DocumentStatus]
```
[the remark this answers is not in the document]
lmolkova: great catch, thanks!

docs/general/naming.md
```diff
@@ -258,15 +294,15 @@ For example if we have a namespace `system.process` which contains all metrics r
...
+#### Do not use `total`
 
 UpDownCounters SHOULD NOT use `_total` because then they will look like
 monotonic sums.
 
 Counters SHOULD NOT append `_total` either because then their meaning will
 be confusing in delta backends.
 
-### Instrument Naming
+### Instrument naming
```
jsuereth, line 305: Why are we changing title convention everywhere?

jsuereth (approved):
Overall I'm on board, not clear on why title capitalization is changing

docs/general/naming.md
lmolkova: it's what otel.io does for link titles - https://github.com/open-telemetry/semantic-conventions/issues/169, pretty much all headings we have are in Sentence case now.

joaopgrassi (approved):

PR #1719, opened 2025-01-03
Adds `mistral_ai` as a `gen_ai.system` attribute value

codefromthecrypt:
This extends GenAI System to include Mistral AI. This accessed either via its native 'mistralai' API or via OpenAI libraries.

This is a follow-up to #1655

## Changes

Added the following:
* "mistral_ai", not "mistralai" to adhere to conventions suggested by @trask and @lmolkova 


drewby (approved):
This addition looks good to me.

model/gen-ai/registry.yaml
```diff
@@ -61,6 +61,10 @@ groups:
               stability: experimental
               value: "groq"
               brief: 'Groq'
+            - id: mistralai
```
lmolkova, line 64: since it's spelled as two words in the docs and on the [trademark](https://tsdr.uspto.gov/#caseNumber=79397874&caseSearchType=US_APPLICATION&caseType=DEFAULT&searchType=statusSearch) I believe it should be

```suggestion
            - id: mistral_ai
```

codefromthecrypt: so in the README I wrote this:

> "mistralai", not "mistral" as their client library is named "mistralai", despite some of their models named "mistral-XXX"

do we want to use a convention different than what they are using? ok by me, but worth the question.

codefromthecrypt: @Bam4d since you are from mistral, maybe you can answer this question. OpenTelemetry will make signals like metrics, and a dimension of the metrics is where chat requests are coming from. For Mistral AI, should the string be "mistralai" or "mistral_ai" (or something else). We just want to use a coherent value and there are multiple seemingly correct choices.

Thanks for letting us randomly ping you.

lmolkova (approved):

model/gen-ai/registry.yaml
trask: looking at https://mistral.ai, `mistral_ai` seems most consistent with [documented semconv naming](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#general-naming-considerations) to me, where words are separated with underscores

codefromthecrypt: ok renaming

gyliu513: This may related to https://github.com/open-telemetry/semantic-conventions/pull/1684 , I was trying to update `vertex_ai` to `vertex.ai`, do we need to keep consistent? Like use `mistral.ai`?

trask: hi @gyliu513, initial feedback in the semconv SIG meeting on Monday was that `vertex_ai` is likely more appropriate because we don't think(?) `vertex` is anything on its own and so `vertex.ai` wouldn't be following our [namespacing guidelines](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#general-naming-considerations)

gyliu513: Fair enough, let me had some discussion at https://github.com/open-telemetry/semantic-conventions/pull/1684

trask (approved):

model/gen-ai/registry.yaml
codefromthecrypt: fyi if you look at all the system names (not just what's visible in this diff), we basically have one of every kind of naming convention that has 'ai' in it.

trisch-me: We have `openai` written together just above in the list of members.
I think the naming of enum is bit different from regular guidelines for attributes, because enum can’t have additional sub-attributes in it

We might create a rule for it, but I do believe both names with or without dot should depend on the meaning and usage for the product itself, right now I see all options in the enum

`ibm.watsonx.ai`
`az.ai.openai`
`vertex_ai `

PR #1723, opened 2025-01-07
Use lychee for changelog links

lmolkova:
A tiny fix after https://github.com/open-telemetry/semantic-conventions/pull/1697 - apparently we used makrdownlint to check links in changelog in a separate step, and now the CI is broken - updating


.chloggen/deleteme.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: fake
+note: This is a fake changelog record to test CI. It also has a [non-existing link](https://dk2342342342423lfgjdflkgjlkdfgjkldjdflk.com)
```
lmolkova, line 3: it should fail, but it doesn't 😞

.github/workflows/changelog.yml
```diff
@@ -77,7 +77,13 @@ jobs:
       # In order to validate any links in the yaml file, render the config to markdown
       - name: Render .chloggen changelog entries
         run: |
-          make chlog-preview 2> changelog_preview.md
-          cat changelog_preview.md
-      - name: Run markdown-link-check
-        run: make markdown-link-check-changelog-preview
+          mkdir -p ${{ github.workspace }}/tmp
+          make chlog-preview &> ${{ github.workspace }}/tmp/changelog_preview.md
```
trask, line 81: is there problem with using `/tmp/`?
```suggestion
          make chlog-preview &> /tmp/changelog_preview.md
```

.github/workflows/changelog.yml
```diff
@@ -77,7 +77,13 @@ jobs:
...
+          cat ${{ github.workspace }}/tmp/changelog_preview.md
```
trask, line 82: (unless you want to keep it for debugging)
```suggestion
```

trask (approved):

.github/workflows/changelog.yml
lmolkova: not really, just more complicated docker command with multiple targets (to config and to temp). I came back to relative path - I changed it while going into a different rabbit hole

lmolkova:
lychee is super-fast! Thanks @trask for switching everything!

joaopgrassi (approved):

PR #1727, opened 2025-01-08
Update http-spans.md example POST

MrAlias:
Correct the `http.request.method` to be POST (matching the span name).


lmolkova (approved):
Thank you!

trask (approved):

PR #1734, opened 2025-01-10
Consistent naming: `db.system` to `db.system.name`, namespace constants, remove `db` from system-specific names

### LINKED ISSUE #1581: Should db.system be db.system.name?

lmolkova:
Fix #1581

Addresses DB-related portion of #608, replaces db-related things from #1711 and #1613. 

1. Rename `db.system` to `db.system.name`. It gives us a good opportunity to clean up constants, batching multiple breaking changes together (see below)
2. Removes deprecated constants from new enum and updates existing constants to follow consistent naming
3. Removes `db` prefix from system-specific attributes following #1708:
    - some systems don't fit into one domain (e.g. redis is a cache and database and also a messaging system)
    - not all operations that are done on the database are related to DB (control plane, auth, scaling, node communication)
    - we're quite inconsistent in naming
4. Adjusts `_` -> `.` where it makes sense


docs/database/database-spans.md
```diff
@@ -244,64 +244,59 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
 * [`server.port`](/docs/attributes-registry/server.md)
 
 ---
 
-`db.system` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+`db.system.name` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
 
 | Value  | Description | Stability |
 |---|---|---|
+| `actian.ingres` | [Actian Ingres](https://www.actian.com/databases/ingres/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 257: `ingres` seems to be too generic - easy to mix with a word "ingress", but it's [a trademark](https://tsdr.uspto.gov/#caseNumber=74288104&caseSearchType=US_APPLICATION&caseType=DEFAULT&searchType=statusSearch). I'm totally open to change it back to `ingres`

docs/database/database-spans.md
```diff
@@ -244,64 +244,59 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
-| `hsqldb` | HyperSQL DataBase | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `influxdb` | InfluxDB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `informix` | Informix | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `ingres` | Ingres | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `instantdb` | InstantDB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `interbase` | InterBase | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `intersystems_cache` | InterSystems Caché | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `mariadb` | MariaDB (This value has stability level RELEASE CANDIDATE) | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
-| `maxdb` | SAP MaxDB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `aws.dynamodb` | [Amazon DynamoDB](https://aws.amazon.com/pm/dynamodb/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 259: [Dynamodb](https://tsdr.uspto.gov/#caseNumber=85524361&caseSearchType=US_APPLICATION&caseType=DEFAULT&searchType=statusSearch) is a trademark, but also ["amazon redshift"](https://tsdr.uspto.gov/#caseNumber=85938060&caseSearchType=US_APPLICATION&caseType=DEFAULT&searchType=statusSearch) is a trademark.

Also, [s3](https://tsdr.uspto.gov/#caseNumber=98324800&caseSearchType=US_APPLICATION&caseType=DEFAULT&searchType=statusSearch) is a trademark, but we use `aws.s3` in [other places](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/object-stores/s3.md).

I'm suggesting to keep all things AWS consistent and use `aws` prefix regardless of what trademark is issued for

docs/database/database-spans.md
```diff
@@ -244,64 +244,59 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
-| `cockroachdb` | CockroachDB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `cosmosdb` | Microsoft Azure Cosmos DB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `couchbase` | Couchbase | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `couchdb` | CouchDB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `db2` | IBM Db2 | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `derby` | Apache Derby | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `dynamodb` | Amazon DynamoDB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `edb` | EnterpriseDB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `elasticsearch` | Elasticsearch | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `filemaker` | FileMaker | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 269: I'm not completely sure it's a database - https://www.claris.com/filemaker/.

I'm suggesting to remove it from the new enum and bring it back if necessary

docs/database/database-spans.md
```diff
@@ -244,64 +244,59 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+| `edb` | [EnterpriseDB](https://www.enterprisedb.com/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `elasticsearch` | [Elasticsearch](https://www.elastic.co/elasticsearch) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `firebird` | [Firebird](https://www.firebirdsql.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `geode` | [Apache Geode](https://geode.apache.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `h2` | [H2 Database](https://h2database.com/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `hbase` | [Apache HBase](https://hbase.apache.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `hive` | [Apache Hive](https://hive.apache.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `hsqldb` | [HyperSQL Database](https://hsqldb.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `ibm.db2` | [IBM Db2](https://www.ibm.com/db2) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 276: All IBM databases have independent trademarks. E.g. [db2](https://tsdr.uspto.gov/#caseNumber=73811020&caseSearchType=US_APPLICATION&caseType=DEFAULT&searchType=statusSearch)

Still, given  that other cloud services come with provider as a namespace, I'm suggesting `ibm.*`. 

Maybe @gyliu513 or @mamwl1 would have some input from the IBM side? (it aligns with [`ibm.watsonx.az`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md)) and we'd use `ibm.mq` for as a messaging system name.

docs/database/database-spans.md
```diff
@@ -244,64 +244,59 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+| `influxdb` | [InfluxDB](https://www.influxdata.com/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `instantdb` | [InstantDB](https://www.instantdb.com/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `intersystems.cache` | [InterSystems Caché](https://www.intersystems.com/products/cache/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `mariadb` | MariaDB | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
 | `memcached` | Memcached | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `mongodb` | MongoDB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `mssql` | Microsoft SQL Server (This value has stability level RELEASE CANDIDATE) | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
-| `mysql` | MySQL (This value has stability level RELEASE CANDIDATE) | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
+| `mssql` | Microsoft SQL Server | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
```
lmolkova, line 285: should we rename it to ``ms.sql` or `ms.sql.server` ? wdyt @trask @jcocchi ?

docs/database/database-spans.md
```diff
@@ -244,64 +244,59 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
 | `pervasive` | Pervasive PSQL | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `pointbase` | PointBase | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `postgresql` | PostgreSQL (This value has stability level RELEASE CANDIDATE) | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
-| `progress` | Progress Database | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `postgresql` | PostgreSQL | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
 | `redis` | Redis | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `redshift` | Amazon Redshift | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `sap.hanadb` | [SAP HANA](https://www.sap.com/products/technology-platform/hana/what-is-sap-hana.html) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `sap.maxdb` | SAP MaxDB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `spanner` | Cloud Spanner | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 296: [cloud spanner](https://tsdr.uspto.gov/#caseNumber=97086852&caseSearchType=US_APPLICATION&caseType=DEFAULT&searchType=statusSearch) is a trademark

docs/database/database-spans.md
```diff
@@ -244,64 +244,59 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
-| `postgresql` | PostgreSQL (This value has stability level RELEASE CANDIDATE) | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
-| `progress` | Progress Database | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `postgresql` | PostgreSQL | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
 | `redis` | Redis | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `redshift` | Amazon Redshift | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `spanner` | Cloud Spanner | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `sap.hana` | [SAP HANA](https://www.sap.com/products/technology-platform/hana/what-is-sap-hana.html) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `sap.maxdb` | SAP MaxDB | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `sqlite` | SQLite | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `sybase` | Sybase | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 302: it no longer exists - https://en.wikipedia.org/wiki/Sybase, missed it when deprecating dead databases in the past

docs/database/database-spans.md
```diff
@@ -244,64 +244,59 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+| `mysql` | MySQL | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
 | `neo4j` | Neo4j | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `netezza` | Netezza | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `opensearch` | OpenSearch | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `oracle` | Oracle Database | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `opensearch` | [OpenSearch](https://opensearch.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `oracle` | [Oracle Database](https://www.oracle.com/database/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `other_sql` | Some other SQL database. Fallback only. See notes. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `pervasive` | Pervasive PSQL | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `pointbase` | PointBase | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 295: https://docs.oracle.com/cd/E13218_01/wlp/docs92/db/pointbase.html


> Note:	PointBase is supported only for the design, development, and verification of applications; it is not supported for enterprise-quality deployment. The evaluation license of PointBase has a database size limit of 30 MB.


Suggesting to remove it and bring back if it ever comes up

docs/database/database-spans.md
```diff
@@ -244,64 +244,59 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
-| `postgresql` | PostgreSQL (This value has stability level RELEASE CANDIDATE) | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
-| `progress` | Progress Database | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 297: There are multiple database-related products under progress software - https://www.progress.com/legal/trademarks, so I'm not sure which one is meant here and what's could be a more specific name for it. 

Suggesting to remove until someone comes to bring it back.

docs/database/database-spans.md
```diff
@@ -244,65 +244,58 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+| `memcached` | [Memcached](https://memcached.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `mongodb` | [MongoDB](https://www.mongodb.com/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `mssql` | [Microsoft SQL Server](https://www.microsoft.com/sql-server) | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
+| `mysql` | [MySQL](https://www.mysql.com/) | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
+| `neo4j` | [Neo4j](https://neo4j.com/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `opensearch` | [OpenSearch](https://opensearch.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `oracle` | [Oracle Database](https://www.oracle.com/database/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | `other_sql` | Some other SQL database. Fallback only. See notes. | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `pervasive` | Pervasive PSQL | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 294: the company [no longer exists](https://en.wikipedia.org/wiki/Pervasive_Software) (since 2013), the database is now called [Actian Zen](https://www.actian.com/databases/zen). Suggesting to remove the old name and add the new once somebody needs it.


https://github.com/open-telemetry/opentelemetry-specification/pull/928/

docs/database/database-spans.md
```diff
@@ -244,65 +244,58 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
-| `postgresql` | PostgreSQL (This value has stability level RELEASE CANDIDATE) | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
-| `progress` | Progress Database | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `redis` | Redis | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `redshift` | Amazon Redshift | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `spanner` | Cloud Spanner | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `sqlite` | SQLite | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `sybase` | Sybase | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `teradata` | Teradata | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `trino` | Trino | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `vertica` | Vertica | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 305: I don't think this is an accurate name,  it's called [OpenText Analytics Database (Vertica)](https://www.opentext.com/products/analytics-database).

It was added in https://github.com/open-telemetry/opentelemetry-specification/pull/928 in a bulk-update, so I'm suggesting to drop it and again, add under a proper name when necessary

model/database/registry.yaml
```diff
@@ -127,238 +127,185 @@ groups:
...
-              brief: "Amazon DynamoDB"
-              stability: experimental
+              value: "derby"  # trademark on 'derby'
+              brief: "[Apache Derby](https://db.apache.org/derby/)"
+              stability: development
             - id: edb
-              value: "edb"
-              brief: "EnterpriseDB"
-              stability: experimental
+              value: "edb"  # trademark on 'enterprisedb' and 'edb'
```
trask, line 194: it looks like enterprisedb is based on postgresql, so probably clients won't know the difference and could remove

model/database/registry.yaml
```diff
@@ -127,238 +127,185 @@ groups:
...
             - id: firebird
-              value: "firebird"
-              brief: "Firebird"
-              stability: experimental
-            - id: firstsql
-              value: "firstsql"
-              deprecated: "Replaced by `other_sql`."
-              brief: "Deprecated, use `other_sql` instead."
-              stability: experimental
+              value: "firebird"  # no trademark?
```
trask, line 202: maybe `firebirdsql`, using their domain name, since these enums now are essentially taking top-level namespace slots(?)

docs/database/database-spans.md
```diff
@@ -244,65 +244,58 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+| `gcp.spanner` | [Google Cloud Spanner](https://cloud.google.com/spanner) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `geode` | [Apache Geode](https://geode.apache.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `h2` | [H2 Database](https://h2database.com/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `hbase` | [Apache HBase](https://hbase.apache.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `hive` | [Apache Hive](https://hive.apache.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `hsqldb` | [HyperSQL Database](https://hsqldb.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `ibm.db2` | [IBM Db2](https://www.ibm.com/db2) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `ibm.informix` | [IBM Informix](https://www.ibm.com/products/informix) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `ibm.netezza` | [IBM Netezza](https://www.ibm.com/products/netezza) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
trask, line 279: my initial inclination is not to prefix db2, informix and netezza with `ibm`, because they are standalone databases that exist outside of ibm cloud (I think still?)

and so feel more like cassandra, geode, etc which we don't prefix with `apache`

model/database/registry.yaml
```diff
@@ -127,238 +127,185 @@ groups:
...
+              stability: development
             - id: opensearch
-              value: "opensearch"
-              brief: "OpenSearch"
-              stability: experimental
+              value: "opensearch"  # trademark on 'opensearch'
+              brief: "[OpenSearch](https://opensearch.org/)"
+              stability: development
             - id: oracle
               value: "oracle"
```
trask, line 278: does this work now that we are sort of considering this to point to a top-level namespace?

or maybe a better question, if we add oracle (database) specific attributes what namespace would they be under? (now that we also have oracle cloud related namespaces I think)

model/database/registry.yaml
```diff
@@ -127,238 +127,185 @@ groups:
...
-              brief: "Cloud Spanner"
-              stability: experimental
+              value: "redis"  # trademark on 'redis'
+              brief: "[Redis](https://redis.io/)"
+              stability: development
+            - id: sap.hana
+              value: "sap.hana"  # trademark on 'sap hana'
+              brief: "[SAP HANA](https://www.sap.com/products/technology-platform/hana/what-is-sap-hana.html)"
+              stability: development
+            - id: sap.maxdb
```
trask, line 293: these get tricky...

I think my biases are showing...

if it's a brand name I'm familiar with like informix, I prefer not prefixing it with `ibm.`

but in this case I don't mind

docs/database/database-spans.md
lmolkova: Not everything has a trademark or a unique name - e.g. IBM MQ would have to become `ibm.mq` regardless of where it's hosted. At this moment we'd become slightly inconsistent.

Also, we use `cloud.provider.ibm_cloud` to talk about the cloud - not sure if we want to keep it though. Here `ibm` is a company

I feel there are two options:
1. prefix with a company name when a company provides multiple products (i.e. use `ibm` here) regardless of a trademark or cloud provider. Some company names are the same as product names (`mongoDB`, `easticsearch` is similar - it's already prefixed). Apache projects are similar - there is no ambiguity. 
2. use company prefix only when there is a potential ambiguity (`ibm.mq`) and then I'd probably keep `spanner`, `cosmosdb`, `dynamodb`, `redshift`. At the end of the day, you can host some of those on prem, run in emulators, and cloud provider prefix is not meaningful anyway

model/database/registry.yaml
lmolkova: I'm not a big fan of having `sap.hana` and `maxdb` both being SAP databases but following different naming patterns

model/database/registry.yaml
lmolkova: renamed to `oracle.db`, but it looks funky along with `couchdb`. Need to sleep on this

It's the case where we have one company and may offerings `oracle.db`, [`oracle.nosql_db`?](https://www.oracle.com/database/nosql/), `oracle.web_logic`, `oracle.jvm` 😅, oracle could is being addedas `oracle_cloud`

docs/database/database-spans.md
```diff
@@ -244,65 +244,57 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+| `cockroachdb` | [CockroachDB](https://www.cockroachlabs.com/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `couchbase` | [Couchbase](https://www.couchbase.com/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `couchdb` | [Apache CouchDB](https://couchdb.apache.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `derby` | [Apache Derby](https://db.apache.org/derby/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `elasticsearch` | [Elasticsearch](https://www.elastic.co/elasticsearch) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `firebirdsql` | [Firebird](https://www.firebirdsql.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `gcp.spanner` | [Google Cloud Spanner](https://cloud.google.com/spanner) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `geode` | [Apache Geode](https://geode.apache.org/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `h2` | [H2 Database](https://h2database.com/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
trask, line 272: maybe should be `h2database` when considered as a top-level namespace

model/database/registry.yaml
trask: I think I could live with `hana` and `maxdb` as both top-level namespaces (related to https://github.com/open-telemetry/semantic-conventions/issues/608#issuecomment-2585006195)

model/database/registry.yaml
```diff
@@ -127,238 +127,181 @@ groups:
...
         type:
           members:
             - id: other_sql
               value: "other_sql"
               brief: "Some other SQL database. Fallback only. See notes."
-              stability: experimental
+              stability: development
             - id: adabas
-              value: "adabas"
+              value: "adabas"  # trademark on 'adabas'
```
trask, line 146:
```suggestion
              value: "software_ag.adabas"  # trademark on 'adabas'
```

model/database/registry.yaml
```diff
@@ -127,238 +127,181 @@ groups:
...
             - id: instantdb
-              value: "instantdb"
-              brief: "InstantDB"
-              stability: experimental
-            - id: interbase
-              value: "interbase"
-              brief: "InterBase"
-              stability: experimental
+              value: "instantdb"  # no trademark
+              brief: "[InstantDB](https://www.instantdb.com/)"
```
trask, line 243: from their website, looks like they're calling it just "Instant" (but we could still use `instantdb` as the value based on their domain name

```suggestion
              brief: "[Instant](https://www.instantdb.com/)"
```

model/database/registry.yaml
trask: resolving, I'm coming around to your https://github.com/open-telemetry/semantic-conventions/issues/608#issuecomment-2584900801

trask (commented):
I think I'm coming around to your proposal in https://github.com/open-telemetry/semantic-conventions/issues/608#issuecomment-2584900801

docs/database/database-spans.md
trask: resolving, coming around to your proposal in https://github.com/open-telemetry/semantic-conventions/issues/608#issuecomment-2584900801

docs/database/database-spans.md
trask: `microsoft.sql_server`?

docs/database/elasticsearch.md
```diff
@@ -28,15 +28,15 @@ The **span name** follows the [general database span name guidelines](database-s
...
-| [`db.response.status_code`](/docs/attributes-registry/db.md) | string | The HTTP response code returned by the Elasticsearch cluster. [5] | `200`; `201`; `429` | `Conditionally Required` If response was received. | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
+| [`db.response.status_code`](/docs/attributes-registry/db.md) | string | The HTTP response code returned by the Elasticsearch cluster. [4] | `200`; `201`; `429` | `Conditionally Required` If response was received. | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
+| [`elasticsearch.path_parts.<key>`](/docs/attributes-registry/elasticsearch.md) | string | A dynamic value in the url path. [5] | `elasticsearch.path_parts.index=test-index`; `elasticsearch.path_parts.doc_id=123` | `Conditionally Required` when the url has dynamic values | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
maryliag, line 32: not sure if is within the scope of this PR, but just wondering if the `parts` is really necessary here, and why not use just `elasticsearch.path.*`. It doesn't seem consistent with others, for example, we could say that the `parts` is required for `elasticsearch.node_parts.*`. 
I would prefer remove it from here.

model/database/registry.yaml
```diff
@@ -127,238 +127,181 @@ groups:
...
             - id: couchdb
-              value: "couchdb"
-              brief: "CouchDB"
-              stability: experimental
-            - id: db2
-              value: "db2"
-              brief: "IBM Db2"
-              stability: experimental
+              value: "couchdb"  # trademark on 'couchdb'
+              brief: "[Apache CouchDB](https://couchdb.apache.org/)"
```
maryliag, line 187: I keep checking the official names to see if the name has `DB` together like this one, or separate like Cosmos... why can't it be consistent? 😅

model/database/spans.yaml
```diff
@@ -254,14 +254,14 @@ groups:
         examples: ["mykeyspace"]
         requirement_level:
           conditionally_required: If available.
-      - ref: db.cassandra.page_size
-      - ref: db.cassandra.consistency_level
+      - ref: cassandra.page_size
+      - ref: cassandra.consistency.level
```
maryliag, line 258: are there any other keys for `consistency` besides `level`? 
the impression I had was that we used `.` when there are other possible keys, so I was just curious on why this one changed to `consistency.level`, since I don't see other keys being used for it (unless I missed it)

For example, the one right above kept as `page_size`, so wondering why just this one changed

schema-next.yaml
```diff
@@ -16,6 +16,25 @@ versions:
...
+              db.cassandra.speculative_execution_count: cassandra.speculative_execution.count
+              db.cosmosdb.client_id: az.client.id
+              db.cosmosdb.connection_mode: az.cosmosdb.connection.mode
+              db.cosmosdb.consistency_level: az.cosmosdb.consistency.level
+              db.cosmosdb.request_charge: az.cosmosdb.request.charge
+              db.cosmosdb.request_content_length: az.cosmosdb.request.content_length
+              db.cosmosdb.regions_contacted: az.cosmosdb.request.regions_contacted
+              db.cosmosdb.sub_status_code: az.cosmosdb.response.sub_status_code
+              db.elasticsearch.node.name: elasticsearch.node.name
+              # db.elasticsearch.path_parts.<key> -> elasticsearch.path_parts.<key>
```
maryliag, line 37: is this commented out by mistake?

docs/database/elasticsearch.md
lmolkova: @gregkalapos any thoughts on this? `elasticsearch.path.doc_id` looks nicer than `elasticsearch.path_parts.doc_id`. Since we are renaming things anyway, we can clean up a few extra things

schema-next.yaml
lmolkova: commented out because it's a template attribute and rename transformation does not support them, added a comment about comment

model/database/spans.yaml
lmolkova: Great point and good test for our recent additions to the naming guidance:

https://github.com/open-telemetry/semantic-conventions/blob/8d9d4a1ce84b060e4d13b5f44b3495648767deb5/docs/general/naming.md?plain=1#L57-L68

1. `consistency.level` makes sense to me 
2.  I can come up with an artificial example of `consistency.quorum` - I had to think for a while though. 

So based on this guidance, it *could* be `.`, so it should be `.`.

I applied the same litmus test to page_size and it also makes sense plus I can come up with some examples like `page.number`, `page.next_link`. So I changed it all to `page.size`.

Really curious what do you think about this new guidance

docs/database/database-spans.md
lmolkova: BTW we have tons of `aws.dynamodb` attributes here https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/aws.md#amazon-dynamodb-attributes

 AWS was following this guidance from the start.

docs/database/elasticsearch.md
gregkalapos: Hmm, yeah, good point. I suspect this comes from the fact that the set of these things is referred to as  `path parts`, but here we actually have specific fields for each given specific part - so already the plural confuses me :) E.g. why `path_parts.doc_id` and not `path_part.doc_id`? 

This is btw. also somewhat similar to [`http.request.header.<key>`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/http.md) (except these values aren't in header, but in the url path).

Anyways, I think it's ok to drop `parts`. I was thinking that just `path` sounds a bit ambiguous - just throwing in maybe `url_path`? So `elasticsearch.url_path.<key>`. But then seeing that, I still think that `elasticsearch.path.<key>` looks nicer. So overall, I think `elasticsearch.path.<key>` is ok - dropping `parts` is an improvement.

model/database/spans.yaml
maryliag: I like the new guidance! I was more curious because I saw `page_size` and `consistency_level` one right after the other but they were treated differently. 
Now with both using `.`, I'm happy 😄

model/cassandra/registry.yaml
```diff
@@ -0,0 +1,75 @@
...
+        stability: development
+        brief: >
+          Whether or not the query is idempotent.
+      - id: cassandra.page.size
+        type: int
+        stability: development
+        brief: >
+          The fetch size used for paging, i.e. how many rows will be returned at once.
+        examples: [5000]
+      - id: cassandra.speculative_execution_count
```
trask, line 70:
```suggestion
      - id: cassandra.speculative_execution.count
```

model/database/registry.yaml
```diff
@@ -127,238 +127,181 @@ groups:
...
           members:
             - id: other_sql
               value: "other_sql"
               brief: "Some other SQL database. Fallback only. See notes."
-              stability: experimental
-            - id: adabas
-              value: "adabas"
+              stability: development
+            - id: software_ag.adabas
+              value: "software_ag.adabas"  # trademark on 'adabas'
```
trask, line 146: maybe can remove the trademark comments for values that are now embedded under company name?

model/database/registry.yaml
```diff
@@ -127,238 +127,181 @@ groups:
...
               brief: "Adabas (Adaptable Database System)"
```
trask, line 147:
```suggestion
              brief: "[Adabas (Adaptable Database System)](https://documentation.softwareag.com/?pf=adabas)"
```

model/database/registry.yaml
trask, line 146: we could consider `softwareag.adabas` (prioritizing domain name when deciding on top-level namespaces)

model/database/registry.yaml
```diff
@@ -127,238 +127,181 @@ groups:
...
-              stability: experimental
-            - id: hanadb
-              value: "hanadb"
-              brief: "SAP HANA"
-              stability: experimental
+              value: "geode"  # trademark on 'geode'
+              brief: "[Apache Geode](https://geode.apache.org/)"
+              stability: development
+            - id: h2database
+              value: "h2database "  # no trademark
```
trask, line 210:
```suggestion
              value: "h2database"  # no trademark
```

model/cassandra/registry.yaml
lmolkova: I agree with this, the reason I didn't do it is [http.request.resend_count](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/http.md#http-request-resend-count), but let's do it

docs/attributes-registry/azure.md
```diff
@@ -6,11 +6,51 @@
...
 # Azure
 
-## Azure SDK Attributes
+- [Azure Client Library Attributes](#azure-client-library-attributes)
+- [Azure Cosmos DB Attributes](#azure-cosmos-db-attributes)
 
-This document defines generic attributes for Azure SDK.
+## Azure Client Library Attributes
+
+This document defines generic attributes used by Azure Client Libraries.
```
trask, line 14:
```suggestion
This section defines generic attributes used by Azure Client Libraries.
```

docs/attributes-registry/azure.md
```diff
@@ -6,11 +6,51 @@
...
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
 | <a id="az-namespace" href="#az-namespace">`az.namespace`</a> | string | [Azure Resource Provider Namespace](https://learn.microsoft.com/azure/azure-resource-manager/management/azure-services-resource-providers) as recognized by the client. | `Microsoft.Storage`; `Microsoft.KeyVault`; `Microsoft.ServiceBus` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 | <a id="az-service-request-id" href="#az-service-request-id">`az.service_request_id`</a> | string | The unique identifier of the service request. It's generated by the Azure service and returned with the response. | `00000000-0000-0000-0000-000000000000` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| <a id="azure-client-id" href="#azure-client-id">`azure.client.id`</a> | string | The unique identifier of the client instance. | `3ba4827d-4422-483f-b59f-85b74211c11d`; `storage-client-1` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+## Azure Cosmos DB Attributes
+
+This group defines attributes for Azure Cosmos DB.
```
trask, line 24:
```suggestion
This section defines attributes for Azure Cosmos DB.
```

model/azure/registry.yaml
```diff
@@ -1,20 +1,91 @@
...
+      - id: azure.cosmosdb.request.charge
+        type: double
+        stability: development
+        brief: Request units consumed for the operation.
+        examples: [46.18, 1.0]
+      - id: azure.cosmosdb.request.body.size
+        type: int
+        stability: development
+        brief: Request payload size in bytes.
+      - id: azure.cosmosdb.request.regions_contacted
```
trask, line 55: two thoughts:
* `contacted_regions` is more clearly "one thing" (when considering `_` vs `.`) compared to `regions_contacted`
  * maybe a tip we can give for using `_` is `adjective_noun`. since `adjective` can't be its own namespace.
  * similar to `returned_rows`, `speculative_execution`, `active_instance`
* ends in "s", which is kind of nice for array types(?)
* 

```suggestion
      - id: azure.cosmosdb.request.contacted_regions
```

model/cassandra/registry.yaml
```diff
@@ -0,0 +1,75 @@
...
+              value: "any"
+              stability: development
+            - id: serial
+              value: "serial"
+              stability: development
+            - id: local_serial
+              value: "local_serial"
+              stability: development
+        stability: development
+      - id: cassandra.idempotence
```
trask, line 59: doesn't follow the [`{domain}.{thing}.{a property of the thing}`](https://github.com/open-telemetry/semantic-conventions/issues/1581#issuecomment-2486925892) pattern (I don't have any better idea here, just commenting in case you have thoughts)

docs/database/cosmosdb.md
```diff
@@ -26,11 +26,11 @@ extend and override the [Database Semantic Conventions](database-spans.md).
 
 ## Attributes
 
-`db.system` MUST be set to `"cosmosdb"` and SHOULD be provided **at span creation time**.
+`db.system.name` MUST be set to `"az.cosmosdb"` and SHOULD be provided **at span creation time**.
```
trask, line 29:
```suggestion
`db.system.name` MUST be set to `"azure.cosmosdb"` and SHOULD be provided **at span creation time**.
```

docs/database/cosmosdb.md
```diff
@@ -299,17 +299,17 @@ This metric is [required][MetricRequired].
 
 It captures the total time taken by an Azure Cosmos DB operation. This metric follows the common [db.client.operation.duration](/docs/database/database-metrics.md#metric-dbclientoperationduration) definition.
 
-Refer [db.client.cosmosdb.operation.request_charge](#metric-dbclientcosmosdboperationrequest_charge) metrics for dimensions.
+Refer [az.cosmosdb.client.operation.request_charge](#metric-azcosmosdbclientoperationrequest_charge) metrics for dimensions.
```
trask: (and in multiple places below)
```suggestion
Refer [azure.cosmosdb.client.operation.request_charge](#metric-azurecosmosdbclientoperationrequest_charge) metrics for dimensions.
```

docs/database/elasticsearch.md
trask: +1 for singular (no opinion otherwise), similar to `http.request.header.<key>`

trask (approved):
👍

model/cassandra/registry.yaml
lmolkova: thanks! changed to `cassandra.query.is_idempotent`

model/azure/registry.yaml
lmolkova: awesome suggestion, created https://github.com/open-telemetry/semantic-conventions/issues/1755 so we don't forget to document it

model/cassandra/registry.yaml
trask: i'm slightly anti on using the `is_` prefix, wdyt?

docs/database/elasticsearch.md
lmolkova: Looking into https://www.elastic.co/docs/api/doc/elasticsearch/v8/operation/operation-create-1, the official docs call it "Path parameters" which also aligns with OpenAPI language - https://swagger.io/docs/specification/v3_0/describing-parameters/#path-parameters

So I'm going to suggest 3 possible options:
- `elasticsearch.path.parameter.<key>` 
- elastic could reuse `db.operation.parameter.<key>` - is this important to record this info in `elasticsearch.path.` namespace?
- we could call it `url.path.parameter.key`  (no, we can't because `url.path` is taken, but we could figure something out)

model/cassandra/registry.yaml
trask: but can totally live with it if we apply consistently (or have reasons for when we apply it and when we don't)

lmolkova: would you prefer `cassandra.query.idempotent` ?

trask: yeah

maryliag (approved):

docs/database/sql.md
```diff
@@ -8,7 +8,7 @@ linkTitle: SQL
 
 The SQL databases Semantic Conventions describes how common [Database Semantic Conventions](database-spans.md) apply to SQL databases.
 
-The following database systems (defined in the [`db.system`](./database-spans.md#notes-and-well-known-identifiers-for-dbsystem) set) are known to use SQL as their primary query language:
+The following database systems (defined in the [`db.system.name`](./database-spans.md#notes-and-well-known-identifiers-for-dbsystemname) set) are known to use SQL as their primary query language:
 
 - `cockroachdb`
```
lmolkova, line 13: todo: need to update this list too

docs/database/database-spans.md
```diff
@@ -195,65 +195,57 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
-| `spanner` | Cloud Spanner | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `sqlite` | SQLite | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `sybase` | Sybase | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `teradata` | Teradata | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `trino` | Trino | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
-| `vertica` | Vertica | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `actian.ingres` | [Actian Ingres](https://www.actian.com/databases/ingres/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `aws.dynamodb` | [Amazon DynamoDB](https://aws.amazon.com/pm/dynamodb/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `aws.redshift` | [Amazon Redshift](https://aws.amazon.com/redshift/) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `azure.cosmosdb` | [Azure Cosmos DB](https://learn.microsoft.com/azure/cosmos-db) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
lmolkova, line 211: TODO: update this and other constants in README

docs/database/elasticsearch.md
```diff
@@ -165,17 +165,17 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
 | `url.full`                          | `"https://elasticsearch.mydomain.com:9200/my-index-000001/_search?from=40&size=20"`                                                 |
-| `db.elasticsearch.path_parts.index` | `"my-index-000001"`                                                                                                                 |
 | `db.namespace`                      | `"my-cluster"`                                                                                                                      |
-| `db.elasticsearch.node.name`        | `"instance-0000000001"`                                                                                                             |
+| `elasticsearch.node.name`           | `"instance-0000000001"`                                                                                                             |
+| `elasticsearch.path_parts.index`    | `"my-index-000001"`                                                                                                                 |
```
lmolkova, line 178:
```suggestion
| `db.operation.parameter.index`    | `"my-index-000001"`                                                                                                                 |
```

model/azure/registry.yaml
```diff
@@ -1,20 +1,92 @@
...
+        brief: Request payload size in bytes.
+      - id: azure.cosmosdb.operation.contacted_regions
+        type: string[]
+        stability: development
+        brief: >
+          List of regions contacted during operation in the order that they were contacted. If there is more than one region listed,
+          it indicates that the operation was performed on multiple regions i.e. cross-regional call.
+        note: >
+          Region name matches the format of `displayName`
+          in [Azure Location API](https://learn.microsoft.com/rest/api/subscription/subscriptions/list-locations?view=rest-subscription-2021-10-01&tabs=HTTP#location)
```
chalin, line 64: The following link in `model/azure/registry.yaml` for `id: azure.cosmosdb.operation.contacted_regions` is invalid (404):

> in [Azure Location API](https://learn.microsoft.com/rest/api/subscription/subscriptions/list-locations?view=rest-subscription-2021-10-01&tabs=HTTP#location)

Screenshot:

> Image

chalin: Let's track this via:

- #2844

PR #1737, opened 2025-01-10
adjust process.executable.name semantic expectation

### LINKED ISSUE #1736: `process.executable.name` shouldn't use `/proc/[pid]/status`
### Area(s)

area:process

Today, the `process.executable.name` attribute has the following description:
``` 
The name of the process executable. On Linux based systems, can be set
to the `Name` in `proc/[pid]/status`. On Windows, can be set to the
base name of `GetProcessImageFileNameW`.
```

The Windows guidance in the description will provide an accurate result, but the Linux guidance is problematic. `/proc/[pid]/status` may not have the executable name under two circumstances:

1. The executable name is longer than 15 characters (`TASK_COMM_LEN`-1) ([man page](https://man7.org/linux/man-pages/man5/proc_pid_status.5.html))
2. The process name is changed by the process using `prctl(PR_SET_NAME,...)` ([man page](https://man7.org/linux/man-pages/man2/prctl.2.html))

The current description for `process.executable.name` should instead be applied to a new attribute called `process.name`, and `process.executable.name` should have guidance adjusted so that it actually will represent the executable name.

_No response_

### Additional context

Issue was first brought forward by @christos68k in the Semantic Conventions WG meeting on Jan 6, 2025. Further discussion took place in the [`#otel-system-metrics` CNCF Slack channel](https://cloud-native.slack.com/archives/C05CTFE9U4A/p1736373545603729).

braydonk:
Fixes #1736

## Changes

This PR adds a new attribute `process.name` that uses the description that used to apply to `process.executable.name`. The `process.executable.name` attribute's description is adjusted such that the value of the attribute will reliably contain the executable name.


model/process/registry.yaml
```diff
@@ -11,6 +11,14 @@ groups:
         brief: >
           Process identifier (PID).
         examples: [1234]
+      - id: process.name
+        type: string
+        stability: experimental
+        brief: >
+          The name of the process. On Linux based systems, can be set
+          to the `Name` in `proc/[pid]/status`. On Windows, can be set to the
```
christos68k, line 19:
```suggestion
          to the value in `/proc/[pid]/comm`. On Windows, can be set to the
```

We can also use `/proc/[pid]/comm` which requires no parsing, unlike extracting `Name` out of `/proc/[pid]/status`. The values should be equivalent.

Also see:

- [proc_pid_comm(5)](https://man7.org/linux/man-pages/man5/proc_pid_comm.5.html)
- [prod_pid_status(5)](https://man7.org/linux/man-pages/man5/proc_pid_status.5.html)

.chloggen/process_name.yaml
```diff
@@ -0,0 +1,25 @@
...
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [1736]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The new `process.name` attribute uses the original guidance for `process.executable.name`,
+  suggesting use of the `Name` field from `/proc/[pid]/status` on Linux. `process.executable.name`
```
christos68k, line 24:
```suggestion
  suggesting use of `/proc/[pid]/comm` which is equivalent to the `Name` field
  from `/proc/[pid]/status` on Linux. `process.executable.name`
```

christos68k (approved):
Thanks!

model/process/registry.yaml
braydonk: What do you think about suggesting both since they're the same anyway? For example, in the hostmetricsreceiver we'll already have parsed `/proc/[pid]/status` for other information, so just getting the name from that is fine in our case, but in other cases people might prefer to just get `/proc/[pid]/comm`.

braydonk: Done in [5a679fd](https://github.com/open-telemetry/semantic-conventions/pull/1737/commits/5a679fd2d86cbc3f5b6584f59b37bd5051dd6f79)

model/process/registry.yaml
```diff
@@ -11,6 +11,14 @@ groups:
         brief: >
           Process identifier (PID).
         examples: [1234]
+      - id: process.name
+        type: string
+        stability: experimental
+        brief: >
+          The name of the process. On Linux based systems, can be set to
```
lmolkova, line 18: nit: can or SHOULD ?

model/process/registry.yaml
```diff
@@ -65,8 +73,8 @@ groups:
         stability: experimental
         brief: >
           The name of the process executable. On Linux based systems, can be set
-          to the `Name` in `proc/[pid]/status`. On Windows, can be set to the
-          base name of `GetProcessImageFileNameW`.
+          to the base name of the target of `/proc/[pid]/exe`. On Windows,
+          can be set to the base name of `GetProcessImageFileNameW`.
         examples: ['otelcol']
```
lmolkova, line 78: would it be possible to have an example that's different between `process.name` and `process.executable.name` ?

lmolkova (approved):

model/process/registry.yaml
braydonk: I didn't write this originally so I'm not sure what the original intent of the author was, but I imagine SHOULD is probably correct.

model/process/registry.yaml
braydonk: Where would be the best place for that? As part of this description?

model/process/registry.yaml
```diff
@@ -65,8 +73,8 @@ groups:
         stability: experimental
         brief: >
           The name of the process executable. On Linux based systems, can be set
-          to the `Name` in `proc/[pid]/status`. On Windows, can be set to the
-          base name of `GetProcessImageFileNameW`.
+          to the base name of the target of `/proc/[pid]/exe`. On Windows,
+          can be set to the base name of `GetProcessImageFileNameW`.
```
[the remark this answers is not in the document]
braydonk: Given this specification, it is the case that `process.name` and `process.executable.name` will be the same on Windows. I dug as far as I could into Win32 and .NET APIs to see if there was ever a way to change a process's name at runtime like in Linux, and I could not find any way to do so. So I think it is fundamentally the case even outside of this specification that process name and executable name will always be the same on Windows.

I **think** there is a good use-case for maintaining `process.executable.name` as a cross-platform name, and that's [how it's used in CLI](https://github.com/open-telemetry/semantic-conventions/blob/cb3a6db7ca115f6ef8400d10bb7f9b194ba8dcc9/model/cli/spans.yaml#L9). In this case, it seems the way it's used is for the attribute to be the executable name on both Linux and Windows. On Linux this distinction matters, whereas on Windows it doesn't. However, this ensures that on either platform the `cli.internal` span will always have a correct executable name, and doesn't need to worry about special attributes based on the platform.

Perhaps I could add a note in `process.executable.name`'s description that on Windows it will always be the same, and can be excluded if you're already using `process.name`?

model/process/registry.yaml
braydonk: Added it as a note on `process.name` in [cd4c335](https://github.com/open-telemetry/semantic-conventions/pull/1737/commits/cd4c3357d8325b865b3221e9236b14169226834b)

model/process/registry.yaml
braydonk: Made that change in [cd4c335](https://github.com/open-telemetry/semantic-conventions/pull/1737/commits/cd4c3357d8325b865b3221e9236b14169226834b)

model/process/registry.yaml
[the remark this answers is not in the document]
rockdaboot: Maybe we can generalize this a bit to be OS-agnostic?
- OTEL is not only about Linux and Windows
- missing features in Windows or other OSes may appear with the next version/upgrade

So, can we just say that the two fields *may* have the same value?
That is often true on Linux and always true for current Windows versions and below.

It still makes sense for Windows clients to send both fields, otherwise there should be a written hint/rule like "if `process.name` isn't set, fallback to `process.executable.name`" or the other way round.

docs/attributes-registry/process.md
```diff
@@ -43,17 +44,19 @@ An operating system process.
...
 | <a id="process-working-directory" href="#process-working-directory">`process.working_directory`</a> | string | The working directory of the process. | `/root` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
 **[1] `process.args_count`:** This field can be useful for querying or performing bucket analysis on how many arguments were provided to start a process. More arguments may be an indication of suspicious activity.
 
-**[2] `process.title`:** In many Unix-like systems, process title (proctitle), is the string that represents the name or command line of a running process, displayed by system monitoring tools like ps, top, and htop.
+**[2] `process.name`:** The value of this attribute will be equivalent to `process.executable.name` on Windows, but may not be on Linux. On Linux, the process name from `/proc/[pid]/comm` is truncated if its name is longer than `TASK_COMM_LEN`-1, and it can be manually changed by the process itself via [`prctl(2)`](https://man7.org/linux/man-pages/man2/prctl.2.html). On Windows, it won't be necessary to have both `process.name` and `process.executable.name`, but it may be on Linux depending on your use case.
```
rockdaboot, line 55: How should a backend operate if one of the fields is omitted? Take the value from the other field? What if one of the names isn't available for some reason?

What if on Linux both fields have the same value (a common case), is it OK to just send one of the two fields?

Also, is Windows the only OS where both fields are always the same? Can we have a completely list? If not, should we better not mention "Windows" here as the only affected OS?

Even if we say something like "In cases where `process.name` and `process.executable.name` are identical, only one of the fields is required.", there is room for interpretation. Why not drop the last sentence, which may lead to confusion?

model/process/registry.yaml
lmolkova: I meant if we could have realistic example that's different on linux in `        examples: ['otelcol']`, it's minor.

model/process/registry.yaml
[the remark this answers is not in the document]
lmolkova: We use `process.executable.name` in CLI conventions because it was the only one that existed. 

The question I feel we should address:

If I have some instrumentation that would benefit from having a process name, but does not need a lot of details ("General Class") - which one of those attributes I should use? I'm going to say `process.name` is the first candidate just because it's shorter and looks very general.

In this sense, I would even suggest to change `process.name` definition to be the best known name (yes, you can get it using OS APIs, but maybe you have a smart way to generate better process names, or maybe you want to self-report friendly process name when self-reporting it from within a process).

Let me try to phrase it (will leave a separate comment with suggestion)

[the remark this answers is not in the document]
lmolkova: added suggestion in https://github.com/open-telemetry/semantic-conventions/pull/1737/files#r1913586365

model/process/registry.yaml
rockdaboot: > would it be possible to have an example that's different between process.name and process.executable.name ?

For Linux, one example where the process name isn't changed programmatically is using a symbolic link, for example `ln -s foo bar`. Running `./bar` gives you `bar` from `/proc/<PID>/comm`, but the basename `foo` from `readlink /proc/<PID>/exe`.

A real example for this is `unlz4` which is a link to the `lz4` executable (on my Debian system). So as you say in your suggestion, `process.name` seems to be more relevant/precise here.

Unfortunate, there are other examples where `process.name` needs to be combined with `process.executable.path`:
```
UnicodeNameMappingGenerator-16 -> ../lib/llvm-16/bin/UnicodeNameMappingGenerator
UnicodeNameMappingGenerator-17 -> ../lib/llvm-17/bin/UnicodeNameMappingGenerator
```
`process.name`: UnicodeNameMap (in both cases)
`process.executable.name`: UnicodeNameMappingGenerator (in both cases)

docs/attributes-registry/process.md
mjwolf: In Linux, when using memfd, there can be processes created with no executable file.

So I don't think the backend/SDKs should be copying the fields when missing or omitting an existent field, since it could cause confusion as to whether `process.executable.name` really existed or not.

On the other hand, I agree the fields would commonly be duplicated, so it would be nice have a way to only send one and reduce data volumes. Especially since some workloads can create very large number of processes, so the data volume involving process telemetry can be very large.

model/process/registry.yaml
```diff
@@ -11,6 +11,22 @@ groups:
...
+          the value of `/proc/[pid]/comm` or to the `Name` field in `proc/[pid]/status`
+          (these values are equivalent). On Windows, this SHOULD be set to the
+          base name of `GetProcessImageFileNameW`.
+        note: >
+          The value of this attribute will be equivalent to `process.executable.name`
+          on Windows, but may not be on Linux. On Linux, the process name from `/proc/[pid]/comm`
+          is truncated if its name is longer than `TASK_COMM_LEN`-1, and it can be manually
+          changed by the process itself via [`prctl(2)`](https://man7.org/linux/man-pages/man2/prctl.2.html).
+          On Windows, it won't be necessary to have both `process.name` and `process.executable.name`, but
+          it may be on Linux depending on your use case.
```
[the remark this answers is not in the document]
mjwolf: I think we should avoid calling anything the "general process name", or that it belongs in `process.name`.  There's at least three three different things that might be considered the process name.

1. `process.name`, the name of the actual process/ the runnable process struct in the kernel.
2. `process.executable.name`, the name of the executable file that was used to create the process.
3. `process.title`, the value from `proctitle`, or the human readable name/title.

It's better to ensure the correct field is used for the correct data, and not suggest it's optional which information goes into the different fields. I think part of the confusion before this change is because the differences between these weren't clear.

Which one is the most important might also depend on the use case. Sometimes the executable file name might be considered more important than the process name.

braydonk:
Seeing lots of good points in the discussions but having trouble figuring out where to take this. Here's my attempt to synthesize the current state:

* It's unclear when each attribute should be used, and how consumers should handle the common case where `process.executable.name` and `process.name` will be the same
* The notes are specific only to Linux and Windows right now
  - This is partially because I was adapting the original notes that only mentioned the two operating systems, but I am myself biased because they are the only two operating systems I know anything about
* Descriptions of the attributes should provide examples of how `process.executable.name` and `process.name` could differ
  - There was a suggestion to make use of the `examples` field for this but I'm not sure how to codify it
* `process.title` is another attribute that can be considered in this arena of the process's name
* The idea of the user's interpretation of the best name for the process is a good candidate for [what we consider the "General Class" of instrumentation within our group](https://github.com/open-telemetry/semantic-conventions/blob/35695e1ac5151f1b07a9e17136c157978538b891/docs/non-normative/groups/system/design-philosophy.md#general-class-a-generalized-cross-platform-use-case-we-want-any-user-of-instrumentation-to-be-able-to-easily-access), however the names exist as they are today because the different ways to "name" the process (name, executable name, title) need to be instrumented in specific ways to avoid user confusion (as described in the notes, albeit not clear enough based on feedback)

For the most part I agree with every point of feedback I've seen, but unfortunately that means I agree with some that are that conflict with each other. `process.name` feels like "General Class" instrumentation and yet its definition is inextricably linked with the exact concept of the "process name" from the OS. The descriptions do only mention Linux and Windows and should find ways to be more general to operating systems, and yet if we aren't somewhat prescriptive on instrumentation in the descriptions then the relationship between these attributes could become muddy if different instrumentation makes different decisions about what values to use.

So I hate to say it, but I'm kinda stuck. I don't have a good idea where to take these attributes from here. I think the only thing I'm confident on is that `process.title` should be represented somehow (I am conflicted on exact naming because `proctitle` is Linux-exclusive iirc and the equivalent on Windows would I guess be `MainWindowTitle` but idk about other OS's).

I'm open to suggestions on what to do with this one.

christos68k:
> * It's unclear when each attribute should be used, and how consumers should handle the common case where `process.executable.name` and `process.name` will be the same

In terms of avoiding duplication, we could specify that if `process.name` is not present, the consumer may retrieve it from `process.executable.name`. This should work for all cases presented in this issue, including `memfd` on Linux. The trade-off is some extra complexity and stateful processing on the consumer side. Alternatively, we could opt for simplicity, allow for the same value to be present in both attributes and rely on protocol-level deduplication mechanisms such as a string table which we use in OTLP profiling (outside of which I don't have any context as to the extent of the issue).

lmolkova:
I don't have a strong opinion, a few things we could consider to get unstuck:
- not adding `process.name` for now. Sticking with `process.executable.name`, just changing its description to fix #1736
- going forward with `process.name`, but leaving its description open and soft so it can evolve after attribute is stabilized. 
- going forward with `process.name` in whatever form, but excluding it from initial stability scope.

> For the most part I agree with every point of feedback I've seen, but unfortunately that means I agree with some that are that conflict with each other. process.name feels like "General Class" instrumentation and yet its definition is inextricably linked with the exact concept of the "process name" from the OS. The descriptions do only mention Linux and Windows and should find ways to be more general to operating systems, and yet if we aren't somewhat prescriptive on instrumentation in the descriptions then the relationship between these attributes could become muddy if different instrumentation makes different decisions about what values to use.

If we go down this road ("process.name" is the best known process name), we'd be:
- prescriptive on how to populate it if instrumentation has no additional context
- would say that instrumentations that know better than that MAY populate it in different way. `process.executable.name` would be the precise one.
- would expect app developers to set whatever value they feel right. 

Either way, prototyping and final stabilization push are usually good time to clean up descriptions and meaning and also if we don't believe that some of it is essential for stability, let's just not add it or let's keep it experimental.

christos68k:
> I don't have a strong opinion, a few things we could consider to get unstuck:
> 
> * not adding `process.name` for now. Sticking with `process.executable.name`, just changing its description to fix [`process.executable.name` shouldn't use `/proc/[pid]/status` #1736](https://github.com/open-telemetry/semantic-conventions/issues/1736)
> * going forward with `process.name`, but leaving its description open and soft so it can evolve after attribute is stabilized.
> * going forward with `process.name` in whatever form, but excluding it from initial stability scope.

All of these options imply that we'd switch the semantics for `process.executable.name` to track the actual executable name. So if we keep being undecided on the semantics for `process.name`, I suggest we remove it from this PR and continue with `process.executable.name` . There is value in having improved semantics for that key and it would also unblock profiling work.

> Either way, prototyping and final stabilization push are usually good time to clean up descriptions and also if we don't believe that some of it is essential for stability, let's just not add it or let's keep it experimental.

My preference for `process.name` would be to introduce it with simplicity in mind and not  let the complexity arising out of deduplication derail us, at least initially.

braydonk:
> So if we keep being undecided on the semantics for process.name, I suggest we remove it from this PR and continue with process.executable.name.

That is now the purpose of this PR. `process.name` related stuff should all be removed, and only the change to the description of `process.executable.name` remains. I have opened #1928 to track adding `process.name`.

lmolkova:
@braydonk @christos68k @rockdaboot I think all the discussions can be resolved now and this PR should be ready to go.
Could you please check?

rockdaboot:
> @braydonk @christos68k @rockdaboot I think all the discussions can be resolved now and this PR should be ready to go. Could you please check?

From my side yes, but don't have the permission to resolve my own comment 😃

braydonk:
Should be good from my side since any open areas of concern have now been moved to #1928

Issue #1656, opened 2024-12-06
Add k8s resource attributes used by the Collector


ChrsMark:
### Area(s)

area:k8s

### Is your change request related to a problem? Please describe.

There are missing resource attributes that are already used in the Collector. This issue aims to list and keep track of them.

### Describe the solution you'd like

We would like to add the following as resource attributes semantic conventions:

- `k8s.hpa.{name,uid}`: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a49008324b95f428d518cb4372e2002adf9564b1/receiver/k8sclusterreceiver/metadata.yaml#L140-L148
- `k8s.replication_controller.{name,uid}`: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a49008324b95f428d518cb4372e2002adf9564b1/receiver/k8sclusterreceiver/metadata.yaml#L80-L88
- `k8s.resourcequota.{name,uid}`: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/metadata.yaml#L90-L98

### Additional context

This would be related to respective metrics' additions:

- https://github.com/open-telemetry/semantic-conventions/pull/1636
- https://github.com/open-telemetry/semantic-conventions/pull/1649#discussion_r1872258043

PR #1636, opened 2024-11-29
Add metrics for deployment, replicaset, replication_controller, statefulset and hpa

ChrsMark:
## Changes

Part of https://github.com/open-telemetry/semantic-conventions/issues/1032.

Fixes https://github.com/open-telemetry/semantic-conventions/issues/1644
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1637

Adds the following metrics:

- [`k8s.deployment.desired_pods`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.114.0/receiver/k8sclusterreceiver/documentation.md#k8sdeploymentdesired) (emitted by [collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9dfa2f7813b11500d001622f3d8c1dd8d9ac58fd/receiver/k8sclusterreceiver/internal/deployment/deployments.go#L32)) [⚠️ BREAKING change for the Collector]
- [`k8s.deployment.available_pods`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.114.0/receiver/k8sclusterreceiver/documentation.md#k8sdeploymentavailable) (emitted by [collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9dfa2f7813b11500d001622f3d8c1dd8d9ac58fd/receiver/k8sclusterreceiver/internal/deployment/deployments.go#L33)) [⚠️ BREAKING change for the Collector]
- [`k8s.replicaset.desired_pods`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.114.0/receiver/k8sclusterreceiver/documentation.md#k8sreplicasetdesired) (emitted by [collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9dfa2f7813b11500d001622f3d8c1dd8d9ac58fd/receiver/k8sclusterreceiver/internal/replicaset/replicasets.go#L31)) [⚠️ BREAKING change for the Collector]
[...]

PR #1742, opened 2025-01-14
Add hpa, resource_quota, replication_controller k8s attributes

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1656

## Changes

Adds the following k8s resources:

- `k8s.hpa.{name,uid}`: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a49008324b95f428d518cb4372e2002adf9564b1/receiver/k8sclusterreceiver/metadata.yaml#L140-L148
- `k8s.replication_controller.{name,uid}`: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a49008324b95f428d518cb4372e2002adf9564b1/receiver/k8sclusterreceiver/metadata.yaml#L80-L88
- `k8s.resource_quota.{name,uid}`: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.115.0/receiver/k8sclusterreceiver/metadata.yaml#L90-L98


TylerHelmuth (approved):

model/k8s/registry.yaml
```diff
@@ -117,6 +117,30 @@ groups:
         brief: >
           The name of the ReplicaSet.
         examples: ['opentelemetry']
+      - id: k8s.replication_controller.uid
```
lmolkova, line 120: it seems we used `replicaset`, `statefulset`, `daemonset`, etc in the past (no separator) and now are use  underscore in `replication_controller`, `resource_quota`, ...

I agree that `_` works best for long words. Please consider renaming old ones to follow the same pattern (e.g. `k8s.replica_set.uid`)

ChrsMark: That's true. The problem here is that `replica_set` looks a bit odd and `daemon_set` even more. So I wonder where we should draw the line of using `_` and where not.

Maybe we can just change everything and use one single non-separated word instead since it's common for kubernetes users anyways.

```console
> kubectl api-resources --sort-by=name
NAME                                SHORTNAMES   APIVERSION                        NAMESPACED   KIND
apiservices                                      apiregistration.k8s.io/v1         false        APIService
bindings                                         v1                                true         Binding
certificatesigningrequests          csr          certificates.k8s.io/v1            false        CertificateSigningRequest
clusterrolebindings                              rbac.authorization.k8s.io/v1      false        ClusterRoleBinding
clusterroles                                     rbac.authorization.k8s.io/v1      false        ClusterRole
componentstatuses                   cs           v1                                false        ComponentStatus
configmaps                          cm           v1                                true         ConfigMap
controllerrevisions                              apps/v1                           true         ControllerRevision
cronjobs                            cj           batch/v1                          true         CronJob
csidrivers                                       storage.k8s.io/v1                 false        CSIDriver
csinodes                                         storage.k8s.io/v1                 false        CSINode
csistoragecapacities                             storage.k8s.io/v1                 true         CSIStorageCapacity
customresourcedefinitions           crd,crds     apiextensions.k8s.io/v1           false        CustomResourceDefinition
daemonsets                          ds           apps/v1                           true         DaemonSet
deployments                         deploy       apps/v1                           true         Deployment
endpoints                           ep           v1                                true         Endpoints
endpointslices                                   discovery.k8s.io/v1               true         EndpointSlice
events                              ev           v1                                true         Event
events                              ev           events.k8s.io/v1                  true         Event
flowschemas                                      flowcontrol.apiserver.k8s.io/v1   false        FlowSchema
horizontalpodautoscalers            hpa          autoscaling/v2                    true         HorizontalPodAutoscaler
ingressclasses                                   networking.k8s.io/v1              false        IngressClass
ingresses                           ing          networking.k8s.io/v1              true         Ingress
jobs                                             batch/v1                          true         Job
leases                                           coordination.k8s.io/v1            true         Lease
limitranges                         limits       v1                                true         LimitRange
localsubjectaccessreviews                        authorization.k8s.io/v1           true         LocalSubjectAccessReview
mutatingwebhookconfigurations                    admissionregistration.k8s.io/v1   false        MutatingWebhookConfiguration
namespaces                          ns           v1                                false        Namespace
networkpolicies                     netpol       networking.k8s.io/v1              true         NetworkPolicy
nodes                               no           v1                                false        Node
persistentvolumeclaims              pvc          v1                                true         PersistentVolumeClaim
persistentvolumes                   pv           v1                                false        PersistentVolume
poddisruptionbudgets                pdb          policy/v1                         true         PodDisruptionBudget
pods                                po           v1                                true         Pod
podtemplates                                     v1                                true         PodTemplate
priorityclasses                     pc           scheduling.k8s.io/v1              false        PriorityClass
prioritylevelconfigurations                      flowcontrol.apiserver.k8s.io/v1   false        PriorityLevelConfiguration
replicasets                         rs           apps/v1                           true         ReplicaSet
replicationcontrollers              rc           v1                                true         ReplicationController
resourcequotas                      quota        v1                                true         ResourceQuota
rolebindings                                     rbac.authorization.k8s.io/v1      true         RoleBinding
roles                                            rbac.authorization.k8s.io/v1      true         Role
runtimeclasses                                   node.k8s.io/v1                    false        RuntimeClass
secrets                                          v1                                true         Secret
selfsubjectaccessreviews                         authorization.k8s.io/v1           false        SelfSubjectAccessReview
selfsubjectreviews                               authentication.k8s.io/v1          false        SelfSubjectReview
selfsubjectrulesreviews                          authorization.k8s.io/v1           false        SelfSubjectRulesReview
serviceaccounts                     sa           v1                                true         ServiceAccount
services                            svc          v1                                true         Service
statefulsets                        sts          apps/v1                           true         StatefulSet
storageclasses                      sc           storage.k8s.io/v1                 false        StorageClass
subjectaccessreviews                             authorization.k8s.io/v1           false        SubjectAccessReview
tokenreviews                                     authentication.k8s.io/v1          false        TokenReview
validatingadmissionpolicies                      admissionregistration.k8s.io/v1   false        ValidatingAdmissionPolicy
validatingadmissionpolicybindings                admissionregistration.k8s.io/v1   false        ValidatingAdmissionPolicyBinding
validatingwebhookconfigurations                  admissionregistration.k8s.io/v1   false        ValidatingWebhookConfiguration
volumeattachments                                storage.k8s.io/v1                 false        VolumeAttachment
```

WDYT?

/cc @dashpole @TylerHelmuth @open-telemetry/semconv-k8s-approvers

TylerHelmuth: I prefer not changing the name from what the collector already uses whenever possible, even if it means some have `_` and some don't.

lmolkova: I understand that you don't want to break users, but that's the last chance to clean things up. We're going to 🤞 live with these names for years. 

So I think we should make consistent choice on how/if to separate words.

Having said this, I agree that `replica_set`, `stateful_set`, `daemon_set` look weird and we can say we make exception for them and potentially a few more, but the default should be `_`.

ChrsMark: I would be personally in favor of the single word approach the k8s follows. Otherwise it will definitely consume us time in the future to determine what to split and what not.

lmolkova: I'd be happy to propose update to semconv naming guidance and add a principle that we prioritize domain-specific naming over otel naming conventions as long as it results in valid attribute/metric/etc name.

jinja2: I am also in favor of keeping the existing metrics names as-is even if there are some inconsistencies. I have some suggestions [here](https://github.com/open-telemetry/semantic-conventions/pull/1636#issuecomment-2536162847) for standardizing names going forward. But @ChrsMark's suggestion to drop the `_` also sounds ok. We should also decide when to use or not use the short form names.

ChrsMark: > We should also decide when to use or not use the short form names.

If we go with following the K8s API definition then I guess we will have to change the short names with the full ones. I'd be fine by that.

ChrsMark:
Based on the discussions from https://github.com/open-telemetry/semantic-conventions/pull/1742#discussion_r1926306420 I have followed the K8s single word approach for referring to resources. Let me know what you think.

If we agree on this we will need a follow-up issue to fix the related metrics' names as well.

model/k8s/registry.yaml
dashpole: I also prefer the single-word approach.  It is nice that it aligns with something I can "kubectl get"

.chloggen/add_k8s_resources.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: k8s
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add hpa, resource_quota and replication_controller resource attributes
```
lmolkova, line 13: nit
```suggestion
note: Add horizontalpodautoscaler , resourcequota and replicationcontroller resource attributes
```

But TBH I'd prefer HPA since (I think) it fits under 'widely recognized' under k8s namespace

https://github.com/open-telemetry/semantic-conventions/blob/867ed9c88f916d587c154e0c7367d50511e11d6e/docs/general/naming.md?plain=1#L94-L101

model/k8s/registry.yaml
```diff
@@ -153,6 +177,18 @@ groups:
         brief: >
           The name of the DaemonSet.
         examples: ['opentelemetry']
+      - id: k8s.horizontalpodautoscaler.uid
```
lmolkova, line 180: see my other comment, I think `k8s.hpa` is fine (assuming it's a common abbreviation in k8s world)

model/k8s/registry.yaml
jinja2: > If we go with following the K8s API definition then I guess we will have to change the short names with the full ones. I'd be fine by that.
I only noticed this with `hpa`. I think it would be good to standardize for new additions but leave hpa as-is to reduce changes in collector

model/k8s/registry.yaml
jinja2: yeah, it's a short name that can be used with kubectl

ChrsMark: Alright, moved it back to `hpa`, thank's!

AlexanderWert (approved):

Issue #236, opened 2023-08-07
Proposal: Define mapping from k8s well-known labels to semconv


smithclay:
Many OpenTelemetry users run their applications in Kubernetes, where it is useful to use a collector processor like [k8sattributes](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) to enrich metrics, spans, or logs with k8s metadata like pod name.

The k8sattributes processor also supports enriching metrics, logs, and spans with resource attributes based on Kubernetes labels, including [Well Known Labels](https://kubernetes.io/docs/reference/labels-annotations-taints/). It would be helpful to OpenTelemetry users what the recommended best practice is for mapping well-known labels to OpenTelemetry resource attributes.

Two of the most obvious candidates (but open for feedback), are:
* `app.kubernetes.io/name` -> `service.name`
* `app.kubernetes.io/version` -> `service.version`

To configure k8sattributes to apply these rules, the collector YAML looks like this: https://github.com/lightstep/otel-collector-charts/blob/main/charts/otel-cloud-stack/values.yaml#L166-L168

It seems like these recommendations could (should?) live as an section in this doc: https://opentelemetry.io/docs/specs/otel/resource/semantic_conventions/k8s/

smithclay:
cc @dmitryax

arminru:
> To configure k8sattributes to apply these rules, the collector YAML looks [like this](https://github.com/lightstep/otel-collector-charts/blob/main/charts/collector-k8s-noprom/values.yaml).

@smithclay The "like this" link does not work (any longer)

smithclay:
> > To configure k8sattributes to apply these rules, the collector YAML looks [like this](https://github.com/lightstep/otel-collector-charts/blob/main/charts/collector-k8s-noprom/values.yaml).
> 
> @smithclay The "like this" link does not work (any longer)

Updated here: https://github.com/lightstep/otel-collector-charts/blob/main/charts/otel-cloud-stack/values.yaml#L166-L168 -- thank you!

joaopgrassi:
This was closed by [mistake by the stale bot.](https://github.com/open-telemetry/semantic-conventions/pull/771) Re-opening

PR #349, opened 2023-09-26
Add recommendation for well-known k8s labels

smithclay:
## Changes

Adds recommendations for mapping Kubernetes well-known labels to existing OpenTelemetry resource attributes.

See https://github.com/open-telemetry/semantic-conventions/issues/236 for original issue.

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] [CHANGELOG.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md) updated for non-trivial changes.
* [ ] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

PR #1756, opened 2025-01-17
Kubernetes: How to populate resource attributes based on attributes, labels and transformation

zeitlinger:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/236

This is a new take on the issue. https://github.com/open-telemetry/semantic-conventions/pull/349 has been created before and not completed.

So what's the difference now?
1. The proposed rules are already implemented in the OTel operator and work well: https://github.com/open-telemetry/opentelemetry-operator#configure-resource-attributes - the PR (at least in it's initial form) makes the operator implementation fully compatible with this spec PR.
2. The specification work on `service.instance.id` has been completed - which was effectively a blocker before.


docs/resource/k8s.md
```diff
@@ -354,3 +354,69 @@ A CronJob creates Jobs on a repeating schedule.
...
+since users may not be aware that their labels are being used for this purpose.
+
+Each attribute has a priority order for how it should be calculated as described below.
+
+### How `service.namespace` is calculated
+
+Choose the first value found:
+
+1. `pod.annotation[resource.opentelemetry.io/service.namespace]`
+2. `pod.label[app.kubernetes.io/part-of]` (well-known label
```
dashpole, line 389: the part-of label feels odd for the namespace to me mainly because I don't expect it to often differentiate multiple instances of the same application (e.g. multiple collector deployments).

dashpole, line 389: Should we add k8s.namespace.name as a fallback here?

docs/resource/k8s.md
```diff
@@ -354,3 +354,69 @@ A CronJob creates Jobs on a repeating schedule.
...
+- `pod.label[app.kubernetes.io/version]` (well-known label
+  [app.kubernetes.io/version](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-version))
+- `if (contains(container docker image tag, '/') == false) container docker image tag`
+
+### How `service.instance.id` is calculated
+
+Choose the first value found:
+
+- `pod.annotation[resource.opentelemetry.io/service.instance.id]`
+- `concat([k8s.namespace.name, k8s.pod.name, k8s.container.name], '.')`
```
dashpole, line 422: nit: should we use a delimiter that is less common in names?  Maybe `/` or `:`?

dashpole (commented):
Love the overall direction.  This makes a lot of sense

docs/resource/k8s.md
zeitlinger: this more like:

shop/shop-frontend and shop/order-api

zeitlinger: good idea

@cyrille-leclerc wdyt?

docs/resource/k8s.md
zeitlinger: `.` is already typically used for namespacing in otel

docs/resource/k8s.md
cyrille-leclerc: +1 to fallback to `k8s.namespace.name`.

zeitlinger: added

cyrille-leclerc: > the `part-of` label feels odd for the namespace to me...

We saw a logic to have mapped the Kubernetes concept `part-of` to OTel `service.namespace` as the definitions share the same:

* `part-of`: The name of a higher level application this one is part of. Example `wordpress`. Definition [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels)
* `service.namespace`: A namespace for `service.name.`... A string value having a meaning that helps to distinguish a group of services, for example the team name that owns a group of services. `service.name` is expected to be unique within the same namespace... Example `Shop`. Definition [here](https://opentelemetry.io/docs/specs/semconv/attributes-registry/service/).

> feels odd for the `namespace` to me mainly because I don't expect it to often differentiate multiple instances of the same application (e.g. multiple collector deployments)

I'm not clear why you refer to distinct instances. Shouldn't it be `service.instance.id` rather than `service.namespace` that is intended to differentiates different instances of the same service?

docs/resource/k8s.md
```diff
@@ -354,3 +354,70 @@ A CronJob creates Jobs on a repeating schedule.
 <!-- endsemconv -->
 
 [DocumentStatus]: https://opentelemetry.io/docs/specs/otel/document-status
+
+## Specify resource attributes using Kubernetes annotations
```
ChrsMark, line 358: I wonder if this should be an OTEP (or non-normative guide) instead. 
In the current format it looks more like guidance not a spec.

Also I wonder if that could be part of the `service.*` definition. What we are trying to do here is to define how `service.*` attributes should be populated in a specific environment not how K8s attributes are defined/affected.

@open-telemetry/specs-semconv-approvers thoughts?

docs/resource/k8s.md
dashpole: This isn't really namespacing, though. Is it?  I guess at least most namespaces seem to use dashes... e.g. `kube-system` or `kube-addon-manager`.  So maybe `.` isn't too bad.

docs/resource/k8s.md
dashpole: Namespaces are often (but not always) used to differentiate copies of the same high-level application/service.  The classic examples are test vs staging vs production or redis (owned by team A) vs redis (owned by team B).  In both cases, the `part-of` concept won't often distinguish them, but the k8s namespace would.

> Shouldn't it be service.instance.id rather than service.namespace that is intended to differentiates different instances of the same service?

I meant it should differentiate "instances of a service", rather than "instances within a service".  E.g. namespace should differentiate between dev and prod deployments of my application, or deployment of the same application by different teams.

cyrille-leclerc: Thanks David. You are right, using the Kubernetes namespace to distinguish the environment like production and staging is a very popular convention.
It totally makes sense to me to exclude `part-of` from the naming strategy for `service.namespace` as the association is not intuitive at all. 

BTW I prefer to explicitly use the `resource.opentelemetry.io/*` annotations in K8s manifests even if they overlap with existing metadata like `metadata/name`, `spec/template/metadata/labels/app`, or `pod.name` to prevent misunderstandings.
Adding those `resource.opentelemetry.io/*` is intrusive but it removes obscure magic and it's rarely complicated to modify K8s deployment manifests.

dmitryax:
We briefly discussed this during the k8s SemConv WG meeting. There is no consensus on the document yet, but we decided to recommend moving it from the `k8s` resource section of the semantic conventions repo to `service` because it prescribes a way of populating values of `service` attributes, not attributes in `k8s` namespace.

docs/resource/k8s.md
```diff
@@ -354,3 +354,70 @@ A CronJob creates Jobs on a repeating schedule.
...
+2. `pod.label[app.kubernetes.io/part-of]` (well-known label
+   [app.kubernetes.io/part-of](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-part-of))
+3. `k8s.namespace.name`
+
+### How `service.name` is calculated
+
+Choose the first value found:
+
+- `pod.annotation[resource.opentelemetry.io/service.name]`
+- `pod.label[app.kubernetes.io/name]` (well-known label
```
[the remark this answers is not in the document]
dashpole: `service.name` isn't meant to differentiate multiple instances of the same application, right?  The example from https://opentelemetry.io/docs/specs/semconv/attributes-registry/service/#service-attributes is "shoppingcart"

docs/resource/k8s.md
```diff
@@ -354,3 +354,70 @@ A CronJob creates Jobs on a repeating schedule.
...
+- `pod.label[app.kubernetes.io/version]` (well-known label
+  [app.kubernetes.io/version](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-version))
+- `if (contains(container docker image tag, '/') == false) container docker image tag`
+
+### How `service.instance.id` is calculated
+
+Choose the first value found:
+
+- `pod.annotation[resource.opentelemetry.io/service.instance.id]`
+- `concat([k8s.namespace.name, k8s.pod.name, k8s.container.name], '.')`
```
jinja2, line 423: We discussed `service.instance.id` in the semconv meeting and had some doubts about the scenario where a container is restarted without a pod recreate (e.g. kubelet restarting a container for OOM). The confusion was whether this id should change to reflect that a new instance of the container was started which is what would happen i think if we left this to an sdk.

docs/resource/k8s.md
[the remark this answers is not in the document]
jinja2: To clarify the usage of `instance of application`, let me use an example. I have to run 2 different statfulsets of the `mysql` db with greater than 1 pod in each in the same namespace. Following the guidance of [k8s std labels](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-instance), i'll have below values

Statefulset 1 - `app.kubernetes.io/name=mysql`, `app.kubernetes.io/instance: mysql-1`
Statefulset 2 - `app.kubernetes.io/name=mysql`, `app.kubernetes.io/instance: mysql-2`

My reading of the attributes definition (`Logical name of the service.`) makes me think the unique `mysql-1` and `mysql-2` names are more appropriate since i can differentiate which installation of the db the metric is coming from.

docs/resource/k8s.md
jinja2: The convention for `service.instance.id` states that it `MUST be unique for each instance of the same service.namespace,service.name pair (in other words service.namespace,service.name,service.instance.id triplet MUST be globally unique).` For a pod which is owned by a statefulset this is not the case. I guess the question is whether this id should be used to differentiate between different instances of a service at a given time or across all instances ever run. In case of a statefulset, the id is not unique enough to differentiate between instances of a pod with same sts ordinal but k8s will only run one instance of a pod with a given ordinal (the id will remain same after an sts pod is recreated).

docs/resource/k8s.md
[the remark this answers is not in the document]
dashpole: @open-telemetry/specs-semconv-approvers can someone help clarify the intention behind `service.name` here?

docs/resource/k8s.md
jinja2: > IMO this is somewhat moot since SDKs are going to start generating unique IDs for service.instance.id anyways.
Are sdks going to ignore any `service.instance.id` values passed in with the env var `OTEL_RESOURCE_ATTRIBUTES`? My understanding is that the documented values are currently being calculated and set in this var by operator when auto-instrumenting.

zeitlinger: > > IMO this is somewhat moot since SDKs are going to start generating unique IDs for service.instance.id anyways.
> 
> Are sdks going to ignore any `service.instance.id` values passed in with the env var `OTEL_RESOURCE_ATTRIBUTES`? My understanding is that the documented values are currently being calculated and set in this var by operator when auto-instrumenting.

no, SDKs must honor whatever attributes are set using `OTEL_RESOURCE_ATTRIBUTES`

zeitlinger:
> We briefly discussed this during the k8s SemConv WG meeting. There is no consensus on the document yet, but we decided to recommend moving it from the `k8s` resource section of the semantic conventions repo to `service` because it prescribes a way of populating values of `service` attributes, not attributes in `k8s` namespace.

@open-telemetry/specs-semconv-approvers the service page is automatically generated - how can I move the content of this PR to the service page?

docs/resource/k8s.md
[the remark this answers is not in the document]
zeitlinger: looking at https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#applications-and-instances-of-applications - I would say that `app.kubernetes.io/instance` is the right translation for `service.name` - I didn't see that before.

Definition of service.name - see https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/service.md

> service.name: MUST be the same for all instances of horizontally scaled services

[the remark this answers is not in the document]
zeitlinger: I'm wondering is we should fall back to `app.kubernetes.io/name` as the next best thing if `app.kubernetes.io/instance` is not found.

docs/resource/k8s.md
```diff
@@ -354,3 +354,70 @@ A CronJob creates Jobs on a repeating schedule.
...
+- `k8s.container.name`
+
+### How `service.version` is calculated
+
+Choose the first value found:
+
+- `pod.annotation[resource.opentelemetry.io/service.version]`
+- `pod.label[app.kubernetes.io/version]` (well-known label
+  [app.kubernetes.io/version](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-version))
+- `if (contains(container docker image tag, '/') == false) container docker image tag`
```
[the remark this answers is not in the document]
zeitlinger: I'm unsure about this part of the spec.

I took this from the way the operator logic - collector does it a bit differently.

docs/resource/k8s.md
zeitlinger: > I've generally interpreted "globally unique" here to mean unique at a given time, rather than unique across all time.

that is also my understanding

docs/resource/k8s.md
zeitlinger: removed `app.kubernetes.io/part-of`

trask:
> @open-telemetry/specs-semconv-approvers the service page is automatically generated - how can I move the content of this PR to the service page?

maybe as subsection(s) under https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/README.md#service

docs/resource/k8s.md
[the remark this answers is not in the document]
jinja2: > I'm wondering is we should fall back to app.kubernetes.io/name as the next best thing if app.kubernetes.io/instance is not found.

yeah, that sounds good to me

docs/resource/README.md
```diff
@@ -22,6 +22,11 @@ This document defines standard attributes for resources. These attributes are ty
   - [Semantic Attributes with Dedicated Environment Variable](#semantic-attributes-with-dedicated-environment-variable)
   - [Semantic Attributes with SDK-provided Default Value](#semantic-attributes-with-sdk-provided-default-value)
 - [Service](#service)
+  - [Service in Kubernetes](#service-in-kubernetes)
+    - [How `service.namespace` is calculated](#how-servicenamespace-is-calculated)
```
jinja2, line 26: Since this is currently applicable to auto-instrumentation with operator, I think we should clarify that these are the guidelines suggested by conventions and adopted in operator. So maybe change the heading to

```suggestion
    - [How `service.namespace` can be calculated](#how-servicenamespace-can-be-calculated)
```

zeitlinger: good point

maybe "should" is better?

docs/resource/k8s.md
zeitlinger: thanks @jsuereth for the feedback on this question, which I have now included in the PR itself:

> Note that the container restart count is not included in the `service.instance.id` calculation, because it makes
troubleshooting harder when the ID changes on every restart, e.g. in a crash loop.

zeitlinger:
@jsuereth @tigrannajaryan thanks for the positive feedback in the entities SIG call. A added the notes about the object hierarchy for `service.name` and value stability for `service.instance.id`. Please let me know if this addresses everything.

zeitlinger:
@jinja2 thanks for the detailed review - everything should be addressed now

dashpole (approved):

jinja2 (approved):

zeitlinger:
@open-telemetry/specs-semconv-approvers please take a look 

Approval from the k8s semconv and entities SIG has been granted now

docs/resource/k8s.md
```diff
@@ -439,3 +439,17 @@ A ResourceQuota provides constraints that limit aggregate resource consumption p
 <!-- endsemconv -->
 
 [DocumentStatus]: https://opentelemetry.io/docs/specs/otel/document-status
+
+## Specify resource attributes using Kubernetes annotations
```
jsuereth, line 443: I'm not sure this is the best place for these docs.

I agree with what's written, just not location.  It's ok here for now

jsuereth (approved):

docs/resource/k8s.md
zeitlinger: @trask I think this was your suggestion - IIUC

docs/resource/README.md
```diff
@@ -138,6 +143,78 @@ namespace = Company
...
+- `k8s.replicaset.name`
+- `k8s.statefulset.name`
+- `k8s.daemonset.name`
+- `k8s.cronjob.name`
+- `k8s.job.name`
+- `k8s.pod.name`
+- `k8s.container.name`
+
+The rationale is to go from the ancestor to the descendant in the Kubernetes resource hierarchy, e.g. from `deployment`
+to `pod` - see [Kubernetes Object Hierarchy](https://gist.github.com/fardjad/ea3c38d566c845e0b353237d3959e365).
```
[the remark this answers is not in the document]
zeitlinger: at least I couldn't find one

docs/resource/README.md
```diff
@@ -138,6 +143,79 @@ namespace = Company
...
 
+### Service in Kubernetes
+
+The following [service resource attributes](../attributes-registry/service.md) are recommended for Kubernetes services.
+
+There are different ways to calculate the service attributes.
+
+1. [Well-Known Labels](https://kubernetes.io/docs/reference/labels-annotations-taints/)
+2. Annotations on the pod template that have the `resource.opentelemetry.io/` prefix as described in  
+    [resource attributes using Kubernetes annotations)](k8s.md#specify-resource-attributes-using-kubernetes-annotations)
```
joaopgrassi, line 154:
```suggestion
    [resource attributes using Kubernetes annotations](k8s.md#specify-resource-attributes-using-kubernetes-annotations)
```

joaopgrassi (approved):

docs/resource/k8s.md
zeitlinger: I tried adding it to the yaml - along with modifying the jinja template.

It looked strange to have markdown in yaml - so I'm wondering where we could put the text.
The goal was to have the k8s guidelines for service attributes part of the service spec.

Here's how it looks embedded in yaml:

```yaml
groups:
  - id: registry.service
    type: attribute_group
    display_name: Service Attributes
    brief: >
      A service instance.
    note: >
      Note: `service.namespace` and `service.name` are not intended to be concatenated for the purpose of forming a single globally unique name for the service. For example the following 2 sets of attributes actually describe 2 different services (despite the fact that the concatenation would result in the same string):

      ```
      # Resource attributes that describes a service.
      namespace = Company.Shop
      service.name = shoppingcart
      ```

      ```
      # Another set of resource attributes that describe a different service.
      namespace = Company
      service.name = Shop.shoppingcart
      ```

      ### Service in Kubernetes
```

@joaopgrassi or @lmolkova do you have a suggestion?

joaopgrassi: Since this all is just "suggestion", I think it would be nicer if it was its own document, inside https://github.com/open-telemetry/semantic-conventions/tree/main/docs/non-normative. So then everything in this PR goes to a new document there. WDYT?

joaopgrassi:
@open-telemetry/specs-semconv-maintainers this seems like a good PR to include in https://changelog.opentelemetry.io. Should we add the label to it? 🤔

trask:
> @open-telemetry/specs-semconv-maintainers this seems like a good PR to include in https://changelog.opentelemetry.io. Should we add the label to it? 🤔

sure! I don't think it strictly falls under "New semantic convention areas or stability changes, as well as releases", but I think it's always ok for @open-telemetry/specs-semconv-maintainers to call out particularly interesting or broadly relevant changes on https://changelog.opentelemetry.io

cc @austinlparker

docs/resource/k8s.md
zeitlinger: good idea - moved it there @joaopgrassi

zeitlinger:
> @open-telemetry/specs-semconv-maintainers this seems like a good PR to include in https://changelog.opentelemetry.io. Should we add the label to it? 🤔

now that the content is moved to non-normative, it should automatically get published under https://opentelemetry.io/docs/specs/semconv/non-normative/ - IIUC

trask:
@zeitlinger can you change this line in CODEOWNERS to cover this new file also (I think `/docs/non-normative/k8s-*`)?
https://github.com/open-telemetry/semantic-conventions/blob/d963d60dc076d80d630acadcdf265c2bfb17e8a1/.github/CODEOWNERS#L64

@open-telemetry/semconv-k8s-approvers does anyone else want to review before we merge? thanks!

zeitlinger:
> @zeitlinger can you change this line in CODEOWNERS to cover this new file also (I think `/docs/non-normative/k8s-*`)?

thanks - added

lmolkova:
@zeitlinger could you please resolve discussions (that are resolved)? Thanks!

zeitlinger:
> @zeitlinger could you please resolve discussions (that are resolved)? Thanks!

@lmolkova all are resolved now :smile:

.chloggen/k8s-label-recommendations.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: enhancement
+
+component: k8s
+
+note: How to populate resource attributes based on attributes, labels and transformation
+
+issues: [236]
+
+subtext:
```
trask, line 9: not sure why CI is complaining about untidy yaml, maybe it want a newline here?
```suggestion
subtext:

```

.chloggen/k8s-label-recommendations.yaml
```diff
@@ -0,0 +1,10 @@
...
+
```
trask, line 10: new try to fix CI untidy complaint
```suggestion
```

trask:
need #1939 to be merged first to fix CI

Issue #1744, opened 2025-01-14
Semconv schema: enum types - remove `value`, inherit it from the `id`


lmolkova:
We've made a couple of similar mistakes in the past when renaming enum members:

(still there now)

https://github.com/open-telemetry/semantic-conventions/blob/991d6caa40f5b25352ac3f772981662870e96faa/model/gen-ai/registry.yaml#L174-L176

and 

(fixed on main)

https://github.com/open-telemetry/semantic-conventions/blob/19f2ba7d0c0b0adc1b9efb8ab189c5300e038e4d/model/registry/messaging.yaml#L144-L146
      

Where we have different values between `id` and `value`. It creates confusion:
- codegen uses ID as a constant name and value as a value, but usually those are the same
- we enforce back-compat for id and not for the value, but both are important for back-compat
- value is not subject to `a-z0-9._` charset rule, while id is. In some cases enum value becomes part of the attribute name (`db.system=foo` and `db.foo.property=bar`)  and then needs to be `a-z0-9._` - it's not enforced in any way

----
Proposal:

- Option 1: stop using `value`.

  - don't allow values that's different from `id` - don't support non `a-z0-9._`  chars in values. 
  - weaver would need to populate value from id for codegen back-compat.
  - in the future we can combine it with #1462 to support extended chars in the value.
 
- Option 2: make `value` optional.

  - default to `id` when not specified
  - enforce back-compat (you can't remove or change value)
  - use `id` to create attribute name when doing (`db.system=foo` and `db.foo.property=bar`)

I'd prefer Option 1 combined with #1462.  Both options result in the same outcome, but with Option 1 we start from being strict and relax it once #1462 comes.

lmolkova:
I think we have to start from Option 2 because we have a bunch of legit cases for different id and value:

- HTTP methods (and other standard things like continent code) defined in ALLCAPS in standards
  - similar with `db.cosmosdb.consistency_level` which uses formally defined and documented constants in OpenAPI spec in PascalCase
  - `k8s.volume.type` uses the casing used in k8s docs
- back-compat (`otel.status.code` is `OK`)
- int value or digit as a first letter (grpc status codes and HTTP versions)

```log
Violation: Attribute 'http.request.method' member 'connect' has value 'CONNECT'
....
Violation: Attribute 'geo.continent.code' member 'af' has value 'AF'
...
[...]

PR #1757, opened 2025-01-17
Introduce generic `gen_ai.output.type` instead of `gen_ai.openai.request.response_format`

lmolkova:
Response format today could only work for structured outputs, but it should be useful in multi-modal cases.
In this PR, the proposal is to capture modality in the `gen_ai.output.type` and have dedicated attributes to describe format for specific types such as `gen_ai.output.json.schema.name`.

E.g. for structured outputs `gen_ai.output.type = json` and `gen_ai.output.json.schema.name = my_type_name`.
For images it could be `gen_ai.output.type = image`, `gen_ai.output.image.format = url`.

See [Cohere](https://docs.cohere.com/reference/chat#request.body.response_format), [Azure AI Inference](https://learn.microsoft.com/en-us/azure/ai-studio/reference/reference-model-inference-chat-completions#request-body),  [Vertex AI](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/control-generated-output), [Vertex AI](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#request) docs

Also fixing a tiny bug where we have different `id` and `value` for completion token types and other minor nits.


aabmass (commented):
Vertex AI/Gemini also supports response schema. I'll take a look and see if it works with this PR

karthikscale3 (approved):
LGTM

trentm (approved):

docs/gen-ai/openai.md
```diff
@@ -116,6 +108,16 @@ Instrumentations SHOULD document the list of errors they report.
...
 | `text_completion` | Text completions operation such as [OpenAI Completions API (Legacy)](https://platform.openai.com/docs/api-reference/completions) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
 
+---
+
+`gen_ai.request.response_format` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `json_object` | JSON object response format | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `json_schema` | JSON schema response format | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
lmolkova: that's what [OpenAI](https://platform.openai.com/docs/api-reference/audio/createTranslation), Azure AI and [Cohere](https://docs.cohere.com/reference/chat#request.body.response_format) do:

- OpenAI /Azure AI Inference
  - `{"type": "json_object"}` - json without schema
  - `{"type": "json_schema", "json_schema": {...}}` - json with schema
- Cohere
  - { "type": "json_object" }- json without schema
  - `{"type": "json_object", "json_schema": {...}}` - json with schema
- Vertex AI (correct me if I'm wrong, reading [this](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#request))
  - `responseMimeType`: `application/json` | `text/plain`  | `text/x.enum`
  - `responseSchema` 

I.e. if we apply existing things to vertex, it would 
- check if responseMimeType is json
  - if `responseSchema` is set, then `gen_ai.request.response_format=json_schema`
  - `gen_ai.request.response_format =json_object` otherwise
- if responseMimeType  is text, `gen_ai.request.response_format=text`
- in other cases, probably `gen_ai.request.response_format={responseMimeType}` 

I guess the question is whether that's we should continue doing?

docs/attributes-registry/gen-ai.md
```diff
@@ -90,31 +101,21 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
 | Value  | Description | Stability |
 |---|---|---|
 | `input` | Input tokens (prompt, input, etc.) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+| `output` | Output tokens (completion, response, etc.) | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `output`. |
 | `output` | Output tokens (completion, response, etc.) | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
```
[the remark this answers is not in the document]
lmolkova: there was a enum member with s bug

https://github.com/open-telemetry/semantic-conventions/blob/991d6caa40f5b25352ac3f772981662870e96faa/model/gen-ai/registry.yaml#L174-L176
as you can see `id` and `value` are different - more details in #1744

improving it here https://github.com/open-telemetry/semantic-conventions/pull/1764

docs/gen-ai/openai.md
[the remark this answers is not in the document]
aabmass: I think you're reading of Vertex API is correct. Looking ahead, there is a separate API for generating images which also supports more mime types https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/imagen-api#output-options. I'm guessing `gen_ai.request.response_format` attribute would be shared for multimodal as well–what do you think of changing this attribute to a mime type?

Regarding json schema, I opened https://github.com/open-telemetry/semantic-conventions/issues/1760 to support capturing the schema if available.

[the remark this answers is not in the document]
lmolkova: > I'm guessing `gen_ai.request.response_format` attribute would be shared for multimodal as well–what do you think of changing this attribute to a mime type?

great point!

Just did a basic research on types:
- OpenAI 
  - [Audio response format types](https://platform.openai.com/docs/api-reference/audio/createSpeech): mp3, opus, aac, flac, wav, and pcm.
  - [Realtime response format types](https://platform.openai.com/docs/api-reference/realtime-sessions/create): pcm16, g711_ulaw, or g711_alaw - this one is called `output_audio_format`
  - [Image response format types](https://platform.openai.com/docs/api-reference/images/create): url or b64_json
- Vertex AI uses [mime-types for images too](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/imagen-api#output-options)

I think mime-type won't work in some cases (e.g. image URL), but also it requires to map everything to a mime type. It's not straightforward - e.g. `b64_json` DALL-E produces seems to be `webp` but it's not documented (!?) and internet suggests to decode and look at the image header - not something instrumentation should do.

Seeing `output_audio_format` makes me think that there is a future where you could specify multiple formats for in out request (generate that video, this audio, etc) and we might need different attributes to capture `text_format`, `audio_format`, etc. 

----
I think I see two options:

1. Use response_format as a union of all possible formats across modalities: (`text | json | mp3 | url | .. `)
   - we can define a few well-known types (e.g. `json` would be used for all json things - `json_object`, `json_schema`, `application/json`, etc)
   - for everything else it would match constant used by the specific system, Vertex can use `audio/x-alaw-basic` and openai would do `g711_alaw`, we don't really try to unify all 
2. Break it down into format-per-modality 
   - `gen_ai.request.text.response.format = plain  | json | bson | xml | python`
   - `gen_ai.request.text.response.schema = {schema type name}` - this is only for text
   - `gen_ai.request.image.response_format = url | png | webp | b64_json` - we can do some normalization for a few things and record the rest as is
   - `gen_ai.request.image.response.aspect_ratio = 4:3` - there are other properties that would be worth recording
   - ...

Either way we could/should add something like
`gen_ai.request.response.type` = `text | image | video | audio | ...` - this is the modality 

(all attribute names need polishing, but I hope you get the idea)

I hope we can also make it symmetrical with request format and the actual response output type. 
Let me do a bit more research, but I'm leaning towards option 2 since it provides typed way to record modality-specific options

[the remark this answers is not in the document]
aabmass: > I hope we can also make it symmetrical with request format and the actual response output type. Let me do a bit more research, but I'm leaning towards option 2 since it provides typed way to record modality-specific options

Option 2 sounds good to me on cursory read through. It sounds like a bigger follow up task, so this PR LGTM if you want to address the rest later.

aabmass (approved):

docs/gen-ai/openai.md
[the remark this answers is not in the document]
lmolkova: I introduced `gen_ai.output.json.schema.name` and `gen_ai.output.type` following option 2, let's discuss on the GenAI call tomorrow

TaoChenOSU (approved):

docs/gen-ai/openai.md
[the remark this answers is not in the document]
aabmass: Did you push the updated code yet?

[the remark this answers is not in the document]
lmolkova: apparently no, did it now, thanks!

schema-next.yaml
```diff
@@ -16,6 +16,10 @@ versions:
               code.filepath: code.file.path
               code.lineno: code.line.number
               code.column: code.column.number
+        # https://github.com/open-telemetry/semantic-conventions/pull/1757
+        - rename_attributes:
+            attribute_map:
+              gen_ai.openai.request.response_format: gen_ai.request.response_format
```
trentm, line 22: This is inaccurate now.

.chloggen/1757.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: breaking
+component: gen-ai
+note: "Rename `gen_ai.openai.request.response_format` to `gen_ai.request.response_format`"
```
trentm, line 3: This is inaccurate now.
As well, the PR title should change.

model/gen-ai/registry.yaml
```diff
@@ -198,30 +203,61 @@ groups:
...
+          This attribute specifies the output modality and not the actual output format.
+          For example, if an image is requested, the actual output could be a
+          URL pointing to an image file.
+
+          Additional output format details may be recorded in `gen_ai.output.{type}.*` attributes, such
+          as `gen_ai.output.json.schema.name`.
+      - id: gen_ai.output.json.schema.name
+        stability: development
+        type: string
+        brief: Represents the name of the JSON schema requested by the client.
```
aabmass, line 243: Maybe this is a difference vs Vertex's API but I'm not sure what to put for `gen_ai.output.json.schema.name`.

Vertex accepts an OpenAPI schema ([proto source](https://github.com/googleapis/googleapis/blob/89a0f8e22df3d09eb1a283a32cb3772ad2553fce/google/cloud/aiplatform/v1/openapi.proto#L59)).

[example](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/control-generated-output#example_curl_command)
```json
{
  "contents": {
    "role": "user",
    "parts": {
      "text": "List a few popular cookie recipes."
    }
  },
  "generation_config": {
    "maxOutputTokens": 2048,
    "responseMimeType": "application/json",
    "responseSchema": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "recipe_name": {
            "type": "string",
          },
        },
        "required": ["recipe_name"],
      },
    }
  }
}
```

What would schema name before this request?

gyliu513 (approved):

model/gen-ai/deprecated/registry-deprecated.yaml
```diff
@@ -39,3 +39,22 @@ groups:
...
+            - id: json_object
+              value: "json_object"
+              brief: 'JSON object response format'
+              stability: experimental
+            - id: json_schema
+              value: "json_schema"
+              brief: 'JSON schema response format'
+              stability: experimental
+        brief: >
+          Deprecated, use `gen_ai.request.response_format`.
```
drewby, line 59:
```suggestion
          Deprecated, use `gen_ai.output.type`.
```

model/gen-ai/registry.yaml
```diff
@@ -198,30 +203,61 @@ groups:
...
+          This attribute SHOULD be used when the client requests output of a
+          specific type. The model may return zero or more outputs of this type.
+
+          This attribute specifies the output modality and not the actual output format.
+          For example, if an image is requested, the actual output could be a
+          URL pointing to an image file.
+
+          Additional output format details may be recorded in `gen_ai.output.{type}.*` attributes, such
+          as `gen_ai.output.json.schema.name`.
+      - id: gen_ai.output.json.schema.name
```
drewby, line 240: I'm still trying to understand where we use a dot vs an underscore. I'm surprised this one is not `schema_name`? Will there be a `schema.value` at some point, is that why?

model/gen-ai/spans.yaml
```diff
@@ -28,9 +28,13 @@ groups:
         requirement_level: recommended
       - ref: gen_ai.request.seed
         requirement_level:
-          conditionally_required: if appliable and if the request includes a seed
-      - ref: gen_ai.request.encoding_formats
-        requirement_level: recommended
+          conditionally_required: if applicable and if the request includes a seed
+      - ref: gen_ai.output.type
```
drewby, line 32: I missed the discussion on when we introduced `gen_ai.output` vs using request/response. Is this in one of the recent SIG meetings or slack threads? I don't have a strong opinion, but just want to listen to the context.

drewby (approved):
One fix, and a couple questions. Otherwise LGTM.

model/gen-ai/registry.yaml
lmolkova: there is a note below that basically says that if there is no name, don't set the name

if you think we need an indication that there was a schema provided, I could change it to

```diff
         This attribute SHOULD be set when `gen_ai.output.type` is `json` and
         the client has requested output in a specific schema.
         Semantic conventions for a specific GenAI system SHOULD document
         how to populate this attribute based on the available request properties.
         If the request does not include a schema this attribute SHOULD NOT be set.
+        If schema definition provided in the request does not contain schema name, 
+        this attribute SHOULD be set to `unknown`
```
(`other`, etc)

WDYT ?

model/gen-ai/spans.yaml
lmolkova: there was no discussion on this. I don't like how `gen_ai.request.response` looks like, plus I have a hope to reuse these attributes on  events where there is no request and there is just some content.

model/gen-ai/registry.yaml
lmolkova: we've documented that `.` is the first choice - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md

https://github.com/open-telemetry/semantic-conventions/blob/b865f63bc7ba3039ad87504d75ee40104149d1bd/docs/general/naming.md?plain=1#L61-L64

drewby: Got it, so we expect `schema` object to have other properties.

lmolkova: yes, but the main motivation is "Use namespaces (and dot separator) whenever it makes sense. For example 
 when introducing an attribute representing a property of some object, follow `*{object}.{property}` pattern". 

Name is a property of a schema and `schema.name` makes sense. Doesn't matter if we end up adding new properties at some point in the future.

model/gen-ai/registry.yaml
aabmass: That would work but it would be helpful to know which providers have name besides OpenAI. It looks like Vertex and [Cohere](https://docs.cohere.com/v2/docs/structured-outputs#json-schema-mode) do not have name unless I'm misreading

lmolkova: I updated the PR with this note. The type info (e.g. a class name) could be available on the SDK (or framework) level even if REST API does not have it as a parameter.

The alternative would be to have a boolean flag, but then we'd need it and a schema name when it's available. I think having `gen_ai.output.json.schema.name = unnamed` solves both cases with one attribute

model/gen-ai/deprecated/registry-deprecated.yaml
lmolkova: thanks!

model/gen-ai/registry.yaml
aabmass: OK my 2 cents–it seems like OpenAI schema name is the exception and not the rule and it might be annoying to see `unnamed` most of the time.

Class name would be nice but it might be opaque to an SDK in the case of an Agent framework generating the schema.

Not blocking this PR though

aabmass (approved):

model/gen-ai/registry.yaml
lmolkova: > Class name would be nice but it might be opaque to an SDK in the case of an Agent framework generating the schema. 

Instrumentation can be done on the framework level too and the attribute can be used on framework spans..

E.g. here's [llama-index example](https://docs.llamaindex.ai/en/stable/examples/structured_outputs/structured_outputs/#build-rag-pipeline-define-structured-output-schema) with class name, same in [langchain](https://python.langchain.com/v0.1/docs/modules/model_io/chat/structured_output/).

Since the attribute would be useful for frameworks, I'd prefer not to have OpenAI-specific attribute like `openai.json.output.schema.name` and then the choice is between 

0. `gen_ai.output.json.schema.name = unnamed` - PR in its current shape
1.  only set `gen_ai.output.json.schema.name` when schema name is known (no indication that schema was provided if it was not named).
2. `gen_ai.output.json.schema.name` when known and `gen_ai.output.json.schema.provided = true | false`
3. come back to having `json_object` and `json_schema` in `gen_ai.output.type`. I  don't like this option since the format and schema seem to be orthogonal 

In all 4 cases, we'd capture full schema definition in an event. 

is there a solution you'd prefer @aabmass ?

lmolkova: Discussed in the sig call on 1/30:

let's not introduce `gen_ai.output.json.schema.name` yet - it should be done in scope of  https://github.com/open-telemetry/semantic-conventions/issues/1835

model/gen-ai/spans.yaml
```diff
@@ -28,9 +28,10 @@ groups:
         requirement_level: recommended
       - ref: gen_ai.request.seed
         requirement_level:
-          conditionally_required: if appliable and if the request includes a seed
-      - ref: gen_ai.request.encoding_formats
-        requirement_level: recommended
```
trask, line 33: just checking that removing this ref was intentional?

trask (approved):

model/gen-ai/spans.yaml
lmolkova: great catch, thanks!

PR #1762, opened 2025-01-17
Move release instructions to RELEASING.md

trask:
I think it's nice to keep release instructions out of `CONTRIBUTING.md`, and using `RELEASING.md` is a reasonably common convention across otel repos: https://github.com/search?q=org%3Aopen-telemetry+path%3ARELEASING.md&type=code


lmolkova (approved):

RELEASING.md
```diff
@@ -0,0 +1,26 @@
+# Making a Release
+
+- Ensure the referenced specification version is up to date. Use
+  [tooling to update the spec](./CONTRIBUTING.md#updating-the-referenced-specification-version)
+  if needed.
+- Run [opentelemetry.io workflow](https://github.com/open-telemetry/opentelemetry.io/actions/workflows/build-dev.yml)
+  against `semantic-conventions` submodule as a smoke-test for docs. Fix broken links, if any.
```
lmolkova, line 7: we don't need it anymore, right?

```suggestion
```

trask: it seems we do, this is how I found #1766 (it was an [intra-spec link to a page title](https://github.com/open-telemetry/opentelemetry-specification/pull/4345#issuecomment-2573599798))

joaopgrassi (approved):

PR #1764, opened 2025-01-18
Don't render deprecated enum members

lmolkova:
I *think* this was the intention - I don't see a reason why we want to render them (they should remain in yaml for history and codegen though). Plus it results in weird things like this https://github.com/open-telemetry/semantic-conventions/pull/1757#discussion_r1920611213


templates/registry/markdown/enum_macros.j2
```diff
@@ -1,5 +1,5 @@
 {% import 'stability.j2' as stability %}
-{% macro filter(member) %}{% if snippet_type is not defined or (member.deprecated is none or member.deprecated == "") %}{{ "True" }}{% else %}{{ "False" }}{% endif %}{% endmacro %}
+{% macro filter(member) %}{% if (member.deprecated is none or member.deprecated == "") %}{{ "True" }}{% else %}{{ "False" }}{% endif %}{% endmacro %}
```
lmolkova, line 2: `snippet_type` is not used anywhere else

trask (approved):

arminru (approved):
I think it's fine to not render them and by keeping them in YAML it also prevents people from accidentally re-introducing them.
Alternatively we could also render them in a separate section or move them to the end of the table, but this might just bulk up markdown further.

PR #1765, opened 2025-01-18
Release automation

trask:
Here's the PR it creates (tested on my fork): https://github.com/trask/semantic-conventions/pull/4

Will add ensuring spec link versions are up-to-date as part of release automation later.

Will send immediate follow-up PR to update release instructions. Can't update it now b/c I want to include link to run the (not yet created) workflow. (and also b/c of conflicts with #1762)

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,60 @@
+name: Prepare release
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'The version to release, e.g. 1.30.0'
+        required: true
+
+jobs:
+  create-pull-request:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Validate version
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,60 @@
...
+          version=${{ inputs.version }}
+          sed -i "0,/^  next:$/s//  $version:/" schema-next.yaml
+          cp schema-next.yaml "schemas/$version"
+          git add "schemas/$version"
+
+          sed -i "0,/^  $version:$/s//  next:\n  $version:/" schema-next.yaml
+
+      - name: Update change log
+        run: |
+          make chlog-update VERSION=${{ inputs.version }}
```
lmolkova, line 42: nit: I think it should be 

```suggestion
          make chlog-update VERSION=v${{ inputs.version }}
```

at least that's what we have in the changelog

lmolkova (approved):
wow, that's awesome!

joaopgrassi (approved):
Awesome!

PR #1767, opened 2025-01-19
Label all new issues with `triage:needs-triage` label

lmolkova:
we currently label issues that were created using template and don't label blank ones.

### DIFF
--- a/.github/workflows/scripts/prepare-new-issue.sh
+++ b/.github/workflows/scripts/prepare-new-issue.sh
@@ -28,7 +28,7 @@ if [[ -z "${ISSUE:-}" || -z "${BODY:-}" || -z "${OPENER:-}" ]]; then
   exit 0
 fi
 
-LABELS=""
+LABELS="triage:needs-triage,"
 AREAS_SECTION_START=$( (echo "${BODY}" | grep -n '### Area(s)' | awk '{ print $1 }' | grep -oE '[0-9]+') || echo '-1' )
 BODY_AREAS=""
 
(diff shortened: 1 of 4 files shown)


.github/workflows/scripts/prepare-new-issue.sh
```diff
@@ -51,8 +51,8 @@ for AREA in ${BODY_AREAS}; do
     LABELS+="${AREA}"
 done
 
-if [[ -v PINGED_AREAS[@] ]]; then
-  echo "The issue was associated with areas:" "${!PINGED_AREAS[@]}"
+if [[ -v BODY_AREAS[@] ]]; then
+  echo "The issue was associated with areas:" "${!BODY_AREAS[@]}"
```
trask, line 55: probably can remove the `if [[ -n "${LABELS}" ]]` below now that never empty

trask (approved):

.github/workflows/scripts/prepare-new-issue.sh
```diff
@@ -45,14 +45,11 @@ for AREA in ${BODY_AREAS}; do
     continue
   fi
 
-  if [[ -n "${LABELS}" ]]; then
-      LABELS+=","
-  fi
-    LABELS+="${AREA}"
+  LABELS+=",${AREA}"
```
trask, line 48: oh, I didn't see this one! there's another `if [[ -n "${LABELS}" ]]` near the end of the file also

joaopgrassi (approved):

.github/workflows/scripts/prepare-new-issue.sh
lmolkova: good catch!

PR #1769, opened 2025-01-19
Label PRs that contain breaking changes based on the changelog

lmolkova:
This allows to label PRs as "breaking" simplifying filtering and querying breaking changes before they are released.

Tested here https://github.com/lmolkova/semantic-conventions/pull/6, https://github.com/lmolkova/semantic-conventions/pull/9


.github/workflows/prepare-new-pr.yml
```diff
@@ -0,0 +1,22 @@
+name: 'Prepare new PR'
+on:
+  pull_request:
+    types: [opened, ready_for_review]
+
+jobs:
+  prepare-new-pr:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
```
trask, line 10: I suspect this only worked in your test because the PR wasn't from a forked repo:

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflows-in-forked-repositories

> The `GITHUB_TOKEN` has read-only permissions in pull requests from forked repositories.

this kind of automation probably requires `pull_request_target`, which brings security considerations

lmolkova: It was necessary in my fork and failed without it. 

I assume it should not even be necessary (since https://github.com/open-telemetry/semantic-conventions/blob/e73d5005d64b897dc83487e8a9c4274106cbc79f/.github/workflows/scripts/prepare-new-issue.sh#L67 runs perfectly fine without any permissions

lmolkova: thanks! yeah, you're right.

So the `pull_request_target` executes workflows from the target branch (main) where the GH_TOKEN will be exposed to the version of this script checked into main. Is there any security concern with it?

I believe it's used today by [collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1b93d013649d4cb9cc5ea545ec81e48d6ad5a60e/.github/workflows/add-codeowners-to-pr.yml) and [.net](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/89f0af731441a364779693a1874c57f33b1e721c/.github/workflows/add-labels.yml#L50) and in the https://github.com/open-telemetry/assign-reviewers-action, so I assume it's considered to be safe and ok to use

trask: > I believe it's used today by [collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1b93d013649d4cb9cc5ea545ec81e48d6ad5a60e/.github/workflows/add-codeowners-to-pr.yml) and [.net](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/89f0af731441a364779693a1874c57f33b1e721c/.github/workflows/add-labels.yml#L50) and in the https://github.com/open-telemetry/assign-reviewers-action

these don't also check out the PR branch though, which is where things get tricky

e.g. you don't want to run a build on the PR branch, since PR could modify the build steps and print out secrets

your automation only needs to read a file, which sounds safe, but I'm not sure if the checkout alone could be compromised, e.g. by a post-checkout hook that prints out secrets (I'll test this... now I'm curious...)

trask: ah, git hooks have to be explicitly enabled after checkout, e.g. `git config --local core.hooksPath .githooks/`, so that at least seems safe

I'm good with pursuing this, just want to be careful about it

.github/workflows/prepare-new-pr.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  prepare-new-pr:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    if: ${{ github.repository_owner == 'open-telemetry' }}
+    steps:
+      - uses: actions/checkout@v4
```
trask, line 15: I think this defaults to checking out `main` after switching to `pull_request_target` (I suspect due to the security concerns), I think need:

```suggestion
      - uses: actions/checkout@v4
        with:
          ref: ${{ github.head_ref }}
```

.github/workflows/scripts/prepare-new-pr.sh
```diff
@@ -0,0 +1,36 @@
...
+if [ -z ${PR:-} ]; then
+    echo "PR number is required"
+    exit 1
+fi
+
+CHLOG="$(gh pr view $PR --json files --jq '.files.[].path | select (. | startswith(".chloggen/"))')"
+echo "Change log file(s): ${CHLOG}"
+
+if [ -z "$CHLOG" ]; then
+    echo "No changelog found in the PR. Ignoring this change."
```
joaopgrassi, line 23: Trick here, because people may forgot to add the changelog and this will only run on PR open. Once they add it later no label will be there. Not sure there's a way around it though

lmolkova: great catch! added `synchronize` type which covers push

.github/workflows/prepare-new-pr.yml
lmolkova: thanks for educating me about GH actions 😅 

I ended up breaking it down into two jobs, one with read permissions to read content on the branch, another one with write permissions to set label, but on main. I *think* is should be safe now

.github/workflows/prepare-new-pr.yml
```diff
@@ -0,0 +1,53 @@
...
+      change_type: ${{ steps.get-change-type.outputs.change_type }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+      - id: get-change-type
+        name: Run get-change-type.sh
+        run: |
+          set +e
+          change_type=$(./.github/workflows/scripts/get-change-type.sh)
```
trask, line 24: this is in the danger zone because it's running this script from the PR branch under elevated permissions

I'm not sure it's vulnerable b/c this part isn't passing any `env` vars, but it's probably an area better to avoid

I think a good way to mitigate this risk would be to checkout main, then checkout the PR branch in a subdir, e.g.

```
      - uses: actions/checkout@v4
      - uses: actions/checkout@v4
        with:
          ref: ${{ github.head_ref }}
          path: prbranch
```

then it's clearer we're only using the files from the prbranch, and avoids temptation to execute any scripts in it

I haven't used it before, but it looks like could even add a `sparse-checkout: .chloggen` to the checkout action to make it even clearer that we're using those files from the PR branch

trask: sorry, didn't see your most recent updates when I left this, looking...

trask: this is still giving `GITHUB_TOKEN` to the script in the PR branch

.github/workflows/prepare-new-pr.yml
trask: I think there may still be an issue, but let's continue in the other discussion below

.github/workflows/prepare-new-pr.yml
lmolkova: yeah, thanks again! I updated it to your proposal to sparse checkout chloggen dir and only use PR branch to read the corresponding file.

.github/workflows/prepare-new-pr.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  prepare-new-pr:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    if: ${{ github.repository_owner == 'open-telemetry' }}
+    outputs:
+      change_type: ${{ steps.get-change-type.outputs.change_type }}
```
trask, line 16:
```suggestion
```

.github/workflows/scripts/prepare-new-pr.sh
```diff
@@ -0,0 +1,41 @@
...
+#  - if label already exist, this is a no-op.
+#  - if any error happens, the script quietly exits with 0.
+
+if [ -z ${PR:-} ]; then
+    echo "PR number is required"
+    exit 1
+fi
+
+if [ -z ${PR_CHANGELOG_PATH:-} ]; then
+    echo "PR number is required"
```
trask, line 20:
```suggestion
    echo "PR_CHANGELOG_PATH is required"
```

.github/workflows/scripts/prepare-new-pr.sh
```diff
@@ -0,0 +1,41 @@
...
+echo "Change log file(s): ${CHLOG}"
+
+if [ -z "$CHLOG" ]; then
+    echo "No changelog found in the PR. Ignoring this change."
+    exit 0
+fi
+
+COUNT="$(echo "$CHLOG" | wc -l)"
+if [ $COUNT -eq 1 ]; then
+    CHANGE_TYPE=$(awk -F': ' '/^change_type:/ {print $2}' "$CHLOG" | xargs)
```
trask, line 34: suggestion, instead of prepending it above
```suggestion
    CHANGE_TYPE=$(awk -F': ' '/^change_type:/ {print $2}' "$PR_CHANGELOG_PATH/$CHLOG" | xargs)
```

.github/workflows/scripts/prepare-new-pr.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo $CHANGE_TYPE
+    gh pr edit "${PR}" --add-label "${CHANGE_TYPE}" || true
+else
+    echo "Found multiple changelog files. Ignoring this change."
+fi
+
+exit 0
```
trask, line 41:
```suggestion
```

trask (approved):

noahfalk (approved):
It wasn't clear how this impacted .NET sem-conv really, but I'm happy to trust you are making things work better procedurally so go for it 👍

.github/workflows/scripts/prepare-new-pr.sh
```diff
@@ -0,0 +1,39 @@
...
+if [ -z "$CHLOG" ]; then
+    echo "No changelog found in the PR. Ignoring this change."
+    exit 0
+fi
+
+COUNT="$(echo "$CHLOG" | wc -l)"
+if [ $COUNT -eq 1 ]; then
+    CHANGE_TYPE=$(awk -F': ' '/^change_type:/ {print $2}' "$PR_CHANGELOG_PATH/$CHLOG" | xargs)
+    echo $CHANGE_TYPE
+    gh pr edit "${PR}" --add-label "${CHANGE_TYPE}" || true
```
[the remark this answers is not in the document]
lmolkova: this would gracefully and quietly exit without setting any labels. It's our choice to decide which change types  to have labels for. We don't need to support all of them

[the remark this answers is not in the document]
lmolkova: See my other comment: https://github.com/open-telemetry/semantic-conventions/pull/1769#discussion_r1924344478

I don't think we need to label PRs with all change types, my goal is to give people a way to get notified about breaking changes before they are merged and participate in discussions. We can always add those labels if necessary

christophe-kamphaus-jemmic (approved):

PR #1772, opened 2025-01-20
Update `process.runtime` example for the Python language

Viicos:
## Changes

[The Python runtimes specification](https://opentelemetry.io/docs/specs/semconv/resource/process/#python-runtimes) says:

> `process.runtime.version`- Fill in the [`sys.implementation.version`](https://docs.python.org/3/library/sys.html#sys.implementation) values separated by dots.

The PyPy example currently uses version 7.3.2 (4 years old). In this version, it seems that PyPy did not implement `sys.implementation.version` correctly:

```python
Python 3.7.4 (87875bf2dfd8, Sep 24 2020, 07:26:36)
[PyPy 7.3.2-alpha0 with GCC 7.3.1 20180303 (Red Hat 7.3.1-5)] on linux
>>>> sys.implementation.version
(major=3, minor=7, micro=4, releaselevel='final', serial=0)
```

`sys.implementation.version` documentation:

> version is a named tuple, in the same format as [`sys.version_info`](https://docs.python.org/3/library/sys.html#sys.version_info). It represents the version of the Python implementation. This has a distinct meaning from the specific version of the Python language to which the currently running interpreter conforms, which `sys.version_info` represents. For example, for PyPy 1.8 `sys.implementation.version` might be `sys.version_info(1, 8, 0, 'final', 0)`, whereas `sys.version_info` would be `sys.version_info(2, 7, 2, 'final', 0)`. For CPython they are the same value, since it is the reference implementation.

In the latest PyPy release (7.3.17), this was fixed:

```python
Python 3.10.14 (39dc8d3c85a7, Aug 27 2024, 14:32:27)
[PyPy 7.3.17 with GCC 10.2.1 20210130 (Red Hat 10.2.1-11)] on linux
>>>> sys.implementation.version
sys.pypy_version_info(major=7, minor=3, micro=17, releaselevel='final', serial=0)
```
So the example was fixed accordingly. However, Is this expected? Should `sys.version_info` be used instead?


Viicos:
> What would be the benefit of using `sys.version_info`?

There's no real benefit in using `sys.version_info` vs `sys.implementation.version`, it's just that they don't provide the same information. `sys.version_info` provides the Python implementation version, while `sys.implementation.version` provides the version of the current implementation.

I believe it's still best to keep `sys.implementation.version`, as coupled with `process.runtime.name`, we know exactly which PyPy version was used and getting the Python implementation version used by this PyPy version is just a matter of looking up online at https://downloads.python.org/pypy/. If we were to use `sys.version_info`, we would be missing the exact PyPy version (e.g. PyPy 7.3.16 and 7.3.17 both use Python 3.10.14).

Just wanted to double check that we still want to stick to `sys.implementation.version`

mx-psi (approved):
> I believe it's still best to keep `sys.implementation.version`, as coupled with `process.runtime.name`, we know exactly which PyPy version was used and getting the Python implementation version used by this PyPy version is just a matter of looking up online at [downloads.python.org/pypy](https://downloads.python.org/pypy/). If we were to use `sys.version_info`, we would be missing the exact PyPy version (e.g. PyPy 7.3.16 and 7.3.17 both use Python 3.10.14).
> 
> Just wanted to double check that we still want to stick to `sys.implementation.version`

This makes sense to me. LGTM

trask:
cc @open-telemetry/python-approvers

xrmx (approved):

docs/resource/process.md
```diff
@@ -250,12 +250,9 @@ Examples for some Python runtimes:
...
 | CPython 3.7.3 on Windows | cpython | 3.7.3 | 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] |
 | CPython 3.8.6 on Linux | cpython | 3.8.6 | 3.8.6 (default, Sep 30 2020, 04:00:38) <br>[GCC 10.2.0] |
-| PyPy 3 7.3.2 on Linux | pypy | 3.7.4 | 3.7.4 (?, Sep 27 2020, 15:12:26)<br>[PyPy 7.3.2-alpha0 with GCC 10.2.0] |
+| PyPy 3 7.3.17 on Linux | pypy | 7.3.17 | 3.10.14 (39dc8d3c85a7, Aug 27 2024, 14:32:27)<br>[PyPy 7.3.17 with GCC 10.2.1 20210130 (Red Hat 10.2.1-11)] |
 
-Note that on Linux, there is an actual newline in the `sys.version` string,
-and the CPython string had a trailing space in the first line.
-
-Pypy provided a CPython-compatible version in `sys.implementation.version` instead of the actual implementation version which is available in `sys.version`.
+Note that on GCC/Clang Linux builds (up to and 3.9), there is an actual newline in the `sys.version` string.
```
Viicos, line 255: https://github.com/python/cpython/pull/24606, [introduced in 3.10.0a6](https://docs.python.org/3.10/whatsnew/changelog.html?highlight=borrowed%20reference#id175).

trask (approved):

Issue #1672, opened 2024-12-11
Add k8s.container.status.waiting metric to semantic conventions


povilasv:
### Area(s)

area:k8s

### Is your change request related to a problem? Please describe.

K8s Cluster receiver uses would like to monitor K8s CrashLoop Back off state. We had a multiple issues about it:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32457
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33797#issuecomment-2533555814

Previously I tried to model `status.waiting` as Resource attribute, but because it's mutable it cannot be Resource attribute. See https://github.com/open-telemetry/semantic-conventions/pull/997

I would like to propose adding this as metric to unblock this PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35668

### Describe the solution you'd like

```
  # k8s.container.* metrics
  - id: metric.k8s.container.status.waiting
    type: metric
    metric_name: k8s.container.status.waiting
    stability: experimental
    brief: "Whether container is in waiting state. (0 for no, 1 for yes)"
    instrument: gauge
    unit: ""
```

ChrsMark:
Thank's for filing this @povilasv, modeling this as a metric makes sense!

There is a proposal on how we should model such "state"/"phase"/"status" metrics in SemConv. It is already in use by `hw` metrics and jmx metrics as mentioned at https://github.com/open-telemetry/semantic-conventions/issues/1032#issuecomment-2420399295 (see https://github.com/open-telemetry/semantic-conventions/issues/1554).
Process' status will also be modeled like this: https://github.com/open-telemetry/semantic-conventions/pull/1212#issuecomment-2511960736
I have filed another one for `k8s.namespace.phase`: https://github.com/open-telemetry/semantic-conventions/pull/1668 which I think will help us verify the proposal.

My only question here would be how the modeling should actually look like. What you propose aligns with [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/pod-metrics.md) but I would like to see how this will be combined with other statuses like `running` and `terminated` as well as how we reflect the `reason` part. Can we deal with them all at once so as to ensure that the decision will scale and cover all? 

FWIWI, If I'm not mistaken we can't have `k8s.container.status.waiting` and `k8s.container.status.waiting.reason` (metric name cannot be namespace at the same time).

ChrsMark:
[...]

PR #1784, opened 2025-01-22
add k8s.container.status.state and k8s.container.status.reason metrics

povilasv:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1672

## Changes

Adds `k8s.container.status.state` metric, it would allow us to alert and monitor containers in not ready state.

I'm still not sure if this should be multiple different metrics or a single one :thinking: 

The current problems, with single metric:
- Running state doesn't have a reason, so we would set reason to empty.
- Waiting state and Terminated state have different sets of reasons:

Waiting state - "ContainerCreating", "CrashLoopBackOff", "CreateContainerConfigError", "ErrImagePull", "ImagePullBackOff"

Terminated state - "OOMKilled", "Completed", "Error", "ContainerCannotRun"

Alternative approach would be to do what KSM does:

- `k8s.container.status.state` metric without `reason` attribute.
- `k8s.container.status.waiting_reason` metric for waiting reason enum.
- `k8s.container.status.terminated_reason` metric for terminated reason enum.

This is not intended to merge, I would appreciate any feedback to see what we want to do here.


model/k8s/metrics.yaml
```diff
@@ -63,6 +63,21 @@ groups:
...
 
+  # k8s.container.* metrics
+  - id: metric.k8s.container.status.state
+    type: metric
+    metric_name: k8s.container.status.state
+    stability: experimental
+    brief: ""
+    instrument: updowncounter
+    unit: ""
+    attributes:
```
ChrsMark, line 74: I think we need to name these as `k8s.container.status.state` and `k8s.container.status.reason` so as to have meaningful names in case we want to re-use them in other signals like Entities (`k8s.container.reason` does not make a lot of sense).

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,21 @@ groups:
...
+      - ref: k8s.container.state
+        requirement_level: required
+      - ref: k8s.container.reason
+        requirement_level: required
```
ChrsMark, line 78: If we have it as `required`, the problem is that states which do not have a reason won't have it provided. Maybe that could be left "unknown" or just empty?

@open-telemetry/semconv-k8s-approvers @open-telemetry/specs-semconv-approvers thoughts?

.chloggen/container-status.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: k8s
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add k8s.container.status.state"
```
ChrsMark, line 13:
```suggestion
note: "Add k8s.container.status.state metric"
```

model/k8s/registry.yaml
```diff
@@ -232,3 +232,69 @@ groups:
               value: 'terminating'
               brief: "Terminating namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase)"
               stability: experimental
+      - id: k8s.container.status.state
```
trask, line 235: @open-telemetry/semconv-system-approvers sorry I forget, where did we end up with adding mutable resource attributes prior to entities?

ChrsMark: These are attributes in the registry which will be used as metrics attributes.

We have an issue to better formalize this at https://github.com/open-telemetry/semantic-conventions/issues/1554.

ref: https://github.com/open-telemetry/semantic-conventions/issues/1032#issuecomment-2420399295

trask, line 235: are there corresponding K8s APIs that you can link to, similar to in `k8s.namespace.phase` above?

model/k8s/registry.yaml
```diff
@@ -232,3 +232,69 @@ groups:
...
+              stability: experimental
+            - id: waiting
+              value: 'waiting'
+              brief: 'The container is waiting.'
+              stability: experimental
+        examples: ['terminated', 'running', 'waiting']
+      - id: k8s.container.status.reason
+        type:
+          members:
+            - id: container_creating
```
jinja2, line 257: I don't have any objection to snake case, just want to make sure we want to stick with this convention going forward for multi word attribute values?

zwindler:
This PR was auto closed but I feel that this change would be useful (at least, for me, and there are 2 issues open for this)

Can this be re-opened? Is there anything I can do to help?

zwindler:
I totally understand :)

Thank you

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,19 @@ groups:
       - ref: network.interface.name
       - ref: network.io.direction
 
+  # k8s.container.* metrics
+  - id: metric.k8s.container.status.state
+    type: metric
+    metric_name: k8s.container.status.state
+    stability: experimental
+    brief: ""
```
ChrsMark, line 71:
```suggestion
    brief: "Describes number of K8s containers that are currently in a given state for a specific reason."
```

I wonder though if we should split `status.state` and `status.reason` to 2 different metrics. I would like to get more opinions from @open-telemetry/semconv-k8s-approvers on this.

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,19 @@ groups:
...
+    instrument: updowncounter
+    unit: ""
```
ChrsMark, line 73: Based on the discussion from https://github.com/open-telemetry/semantic-conventions/pull/1668#discussion_r1907556971 I think that should be `{container}`?

model/k8s/registry.yaml
ChrsMark: Maybe @open-telemetry/specs-semconv-approvers could provide a direction on this?

model/k8s/metrics.yaml
jinja2: Yeah, I think these should be 2 different metrics. For a given state, a k8s container can go through different reasons. For example, the `waiting` state can have a reason of `ErrImagePull`, which might transition to `ImagePullBackOff` after retries fail. If we have reason as a separate metric, users can visualize reason transitions over time.

model/k8s/registry.yaml
trask: snake_case is the convention used for enum id 👍

btw, in case it's not clear, enum ids are only used for code generated constant names (they won't appear in the captured telemetry itself)

model/k8s/metrics.yaml
mveitas: @povilasv Anything I can do to help move this along. I would love to see this push forward as I have a whole engineering teams looking for this as we are moving from DataDog to Grafana and leveraging otel

ChrsMark: I guess the `k8s.container.status.reason` should not be emitted when there is no `reason`?

ChrsMark (approved):

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,28 @@ groups:
...
+    unit: "{container}"
+    attributes:
+      - ref: k8s.container.status.state
+        requirement_level: required
+  - id: metric.k8s.container.status.reason
+    type: metric
+    metric_name: k8s.container.status.reason
+    stability: experimental
+    brief: ""
+    instrument: updowncounter
```
jsuereth, line 82: What's the reason for this being an updowncounter?

Also, this seems like a string-valued metric here.

I see other comments to this effect, but the gist is that these types of metrics have very subtle edges when you use them in practice, leading to surprising issues.  THe prometheus "avoid missing metrics" is an example.

While, longer term, this type of data is something we absolutely want in the Entities signal, we should probably put some caveats on "state as metric" concerns that you'll run into with these.

jsuereth: Alright, the logic makes sense for default aggregation.

And if i understand your example, will EVERY possible reason be reported every time interval? If so this makes a lot more sense.  That wasn't fully clear to me reading the description.

ChrsMark: Maybe add this explanation as part of `note`?

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,28 @@ groups:
       - ref: network.interface.name
       - ref: network.io.direction
 
+  # k8s.container.* metrics
+  - id: metric.k8s.container.status.state
+    type: metric
+    metric_name: k8s.container.status.state
+    stability: experimental
+    brief: ""
```
ChrsMark, line 71: Can we also provide a brief for this metric? (missed that it's missing)

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,28 @@ groups:
...
+    instrument: updowncounter
+    unit: "{container}"
+    attributes:
+      - ref: k8s.container.status.state
+        requirement_level: required
+  - id: metric.k8s.container.status.reason
+    type: metric
+    metric_name: k8s.container.status.reason
+    stability: experimental
+    brief: ""
```
ChrsMark, line 81: Same here a `brief` should be provided unless there is a reason for not doing so.

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,28 @@ groups:
       - ref: network.interface.name
       - ref: network.io.direction
 
+  # k8s.container.* metrics
+  - id: metric.k8s.container.status.state
+    type: metric
+    metric_name: k8s.container.status.state
+    stability: experimental
+    brief: "The state of the container"
```
ChrsMark, line 71: I think we should be aligned with https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/metrics.yaml#L461:
```suggestion
    brief: "Describes the number of K8s containers that are currently in a given state"
```

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,28 @@ groups:
...
+    instrument: updowncounter
+    unit: "{container}"
+    attributes:
+      - ref: k8s.container.status.state
+        requirement_level: required
+  - id: metric.k8s.container.status.reason
+    type: metric
+    metric_name: k8s.container.status.reason
+    stability: experimental
+    brief: "The reason of the container state"
```
ChrsMark, line 81:
```suggestion
    brief: "Describes the number of K8s containers that are currently in a state for a given reason"
```

ChrsMark (approved):

model/k8s/metrics.yaml
ChrsMark: Looks good to me. @jsuereth please also review.

jsuereth: Looks good!

jsuereth (approved):

jinja2 (approved):

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,34 @@ groups:
...
+  - id: metric.k8s.container.status.reason
+    type: metric
+    metric_name: k8s.container.status.reason
+    stability: experimental
+    brief: "Describes the number of K8s containers that are currently in a state for a given reason"
+    instrument: updowncounter
+    unit: "{container}"
+    note: |
+      All possible container states will be reported at each time interval to avoid missing metrics.
+      Only the value corresponding to the current state will be non-zero.
```
jinja2, line 89:
```suggestion
    note: |
      All possible container state reasons will be reported at each time interval to avoid missing metrics.
      Only the value corresponding to the current state reason will be non-zero.
```

docs/attributes-registry/k8s.md
```diff
@@ -160,6 +162,33 @@ conflict.
 
 ---
 
+`k8s.container.status.reason` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `` | Reason unknown. | ![Development](https://img.shields.io/badge/-development-blue) |
```
jinja2, line 169: Does this mean that the metric `k8s.container.status.reason` will show with value 1 and the attribute value for same should be empty? I am bit confused here

dashpole (approved):

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,34 @@ groups:
...
+    metric_name: k8s.container.status.reason
+    stability: experimental
+    brief: "Describes the number of K8s containers that are currently in a state for a given reason"
+    instrument: updowncounter
+    unit: "{container}"
+    note: |
+      All possible container state reasons will be reported at each time interval to avoid missing metrics.
+      Only the value corresponding to the current state reason will be non-zero.
+    attributes:
+      - ref: k8s.container.status.reason
```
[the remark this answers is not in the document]
lmolkova: thanks, so it sounds like there is no additional structured information that would describe the underlying error code (when error happens) deeply than `ErrImagePull`?

It seem to be present in the message as `rpc error: code = NotFound`  but it's not structured.

Feel free to resolve the discussion if you believe we don't have additional error information to add to metrics.

lmolkova (approved):

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,34 @@ groups:
       - ref: network.interface.name
       - ref: network.io.direction
 
+  # k8s.container.* metrics
+  - id: metric.k8s.container.status.state
```
lmolkova, line 67: please add these metric definitions to k8s metrics markdown file - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/k8s-metrics.md

PR #1786, opened 2025-01-22
Temporarily ignore failing link

trask:
I suspect these links are only temporarily failing. This happened once before.


lmolkova:
> I suspect these links are only temporarily failing.

unless management APIs were retroactively removed from Java 17 :)

jsuereth (approved):

PR #1793, opened 2025-01-23
Less release steps

trask:
Automatically sends a PR to update spec repo links whenever there is a spec release.

Example PR: https://github.com/trask/semantic-conventions/pull/7


.github/scripts/update-spec-repo-links.sh
trask, line 1: this file needed to be outside of the `workflows` directory in order for the automation to be able to update it, so I went ahead and moved all of the scripts to the same place

.github/workflows/auto-update-spec-repo-links.yml
```diff
@@ -0,0 +1,80 @@
...
+      already-opened: ${{ steps.check-versions.outputs.already-opened }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - id: check-versions
+        name: Check versions
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          current_version=$(grep "PREVIOUS_SPECIFICATION_VERSION=v.*" \
```
joaopgrassi, line 23: Should this be getting the value of`LATEST_SPECIFICATION_VERSION` and replacing it on `PREVIOUS_SPECIFICATION_VERSION` as the guide said?

trask: I'm not sure I totally follow, I think I removed the need for that step by only maintaining a single version in the file (which aligns with the version currently used in the links)

lmolkova (approved):
🚀

.github/workflows/auto-update-spec-repo-links.yml
joaopgrassi: Ah I see it now. I somehow thought while initially reviewing that you only moved the `update_specification_version.sh` and didn't change it. All good, sorry for the confusion!

joaopgrassi (approved):

PR #1794, opened 2025-01-23
Added new AWS Attributes for various services

AsakerMohd:
## Changes

Added new AWS Attributes for various services including SQS, SNS, Step Functions, Bedrock as well as a new attribute for Lambda. 


model/aws/registry.yaml
```diff
@@ -393,6 +393,13 @@ groups:
           (`Lambda-Runtime-Invoked-Function-Arn` header on the `/runtime/invocation/next` applicable).
         note: This may be different from `cloud.resource_id` if an alias is involved.
         examples: ["arn:aws:lambda:us-east-1:123456:function:myfunction:myalias"]
+      - id: aws.lambda.resource_mapping.id
+        type: string
+        stability: experimental
+        brief: >
+          The UUID of the [AWS Lambda EvenSource Mapping](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html).
```
ppittle, line 400: typo - should be Lambda **EventSource** Mapping

ppittle (changes requested):

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
           and [upload-part-copy](https://docs.aws.amazon.com/cli/latest/reference/s3api/upload-part-copy.html) operations.
           The `part_number` attribute corresponds to the `--part-number` parameter of the
           [upload-part operation within the S3 API](https://docs.aws.amazon.com/cli/latest/reference/s3api/upload-part.html).
+  - id: registry.aws.sqs
```
lmolkova, line 491: if we need any aws.sqs attributes on top of generic messaging, please add a span definition for it and reference all applicable attributes, see [kafka yaml](https://github.com/open-telemetry/semantic-conventions/blob/5b24784e032ada99312d5a17d6970da434f5ce88/model/messaging/spans.yaml#L138) and [md](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/kafka.md) conventions as an example.

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
...
+        examples: ["arn:aws:states:us-east-1:123456789012:activity:get-greeting"]
+      - id: aws.stepfunctions.state_machine.arn
+        type: string
+        stability: experimental
+        brief: >
+          The ARN of the AWS Step Functions State Machine.
+        examples: ["arn:aws:states:us-east-1:123456789012:stateMachine:myStateMachine:1"]
+  - id: registry.aws.bedrock
+    type: attribute_group
+    display_name: Amazon Bedrock Attributes
```
lmolkova, line 556: please add a bedrock-specific section into gen_ai conventions. See [openai](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/openai.md) as an example

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
...
+    brief: >
+        This document defines attributes for AWS Bedrock.
+    attributes:
+      - id: aws.bedrock.agent.id
+        type: string
+        stability: experimental
+        brief: >
+          The unique identifier of the AWS Bedrock Agent. [Agents](https://docs.aws.amazon.com/bedrock/latest/userguide/agents.html) assist users in accomplishing complex tasks and are capable of interacting with various knowledge bases, data sources, and APIs.
+        examples: ["WM5GWNK1AI"]
+      - id: aws.bedrock.data_source.id
```
lmolkova, line 566: let's define a generic attribute for it, this is a shared concept across all genai systems

lmolkova (changes requested):
Please review existing semantic conventions:
- there are several existing attributes that should be used instead of AAWS-specific ones 
- some attributes should be defined as generic ones 

We avoid defining attributes alone without also defining spans, metrics, events, or resources that reference them, so please add corresponding signal definitions

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
...
+        brief: >
+          The ARN of the AWS Step Functions Activity.
+        examples: ["arn:aws:states:us-east-1:123456789012:activity:get-greeting"]
+      - id: aws.stepfunctions.state_machine.arn
+        type: string
+        stability: experimental
+        brief: >
+          The ARN of the AWS Step Functions State Machine.
+        examples: ["arn:aws:states:us-east-1:123456789012:stateMachine:myStateMachine:1"]
+  - id: registry.aws.bedrock
```
lmolkova, line 554: cc @open-telemetry/semconv-genai-approvers and specifically @xrmx who's adding bedrock instrumentation in python 

https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python-contrib+bedrock&type=pullrequests

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
           and [upload-part-copy](https://docs.aws.amazon.com/cli/latest/reference/s3api/upload-part-copy.html) operations.
           The `part_number` attribute corresponds to the `--part-number` parameter of the
           [upload-part operation within the S3 API](https://docs.aws.amazon.com/cli/latest/reference/s3api/upload-part.html).
+  - id: registry.aws.sqs
+    type: attribute_group
+    display_name: Amazon SQS Attributes
```
lmolkova, line 493: /cc @open-telemetry/semconv-messaging-approvers

model/aws/registry.yaml
xrmx: Thanks for pinging me, I think I won't need any of the proposed attributes in my work atm (and so have not much clue about them) but I'll keep an eye on this PR.

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
...
+        This document defines attributes for AWS SQS.
+    attributes:
+      - id: aws.sqs.queue.url
+        type: string
+        stability: experimental
+        brief: >
+          The URL of the AWS SQS Queue. It's a unique identifier for a queue in Amazon Simple Queue Service (SQS)
+          and is used to access the queue and perform actions on it. 
+        examples: ["https://sqs.us-east-1.amazonaws.com/123456789012/MyQueue"]
+      - id: aws.sqs.queue.name
```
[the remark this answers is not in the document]
AsakerMohd: Sure I can use that instead.

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
...
+          The name of the AWS SQS Queue the request refers to. Corresponds to the `--queue-name` parameter of the
+          SQS [get-queue-url](https://docs.aws.amazon.com/cli/latest/reference/sqs/get-queue-url.html) operation.
+        examples: ["some-sqs-queue-name"]
+  - id: registry.aws.sns
+    type: attribute_group
+    display_name: Amazon SNS Attributes
+    brief: >
+        This document defines attributes for AWS SNS.
+    attributes:
+      - id: aws.sns.topic.arn
```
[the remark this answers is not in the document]
AsakerMohd: In this case, idk if this generic attribute best describes the arn. In this case, I might add another generic AWS ARN attribute here: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/aws.md#general-aws-attributes.

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
...
+        brief: >
+          The ARN of the AWS Step Functions State Machine.
+        examples: ["arn:aws:states:us-east-1:123456789012:stateMachine:myStateMachine:1"]
+  - id: registry.aws.bedrock
+    type: attribute_group
+    display_name: Amazon Bedrock Attributes
+    brief: >
+        This document defines attributes for AWS Bedrock.
+    attributes:
+      - id: aws.bedrock.agent.id
```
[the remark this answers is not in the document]
AsakerMohd: Sure I can use that.

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
...
+        brief: >
+          The unique identifier of the AWS Bedrock Data Source. A [data source](https://docs.aws.amazon.com/bedrock/latest/userguide/knowledge-base-ds.html) contains files or content with information that can be retrieved when your knowledge base is queried.
+        examples: ["H7STPQYOND"]
+      - id: aws.bedrock.guardrail.id
+        type: string
+        stability: experimental
+        brief: >
+          The unique identifier of the AWS Bedrock Guardrail. A [guardrail](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails.html) helps safeguard and prevent unwanted behavior from model responses or user messages.
+        examples: ["sgi5gkybzqak"]
+      - id: aws.bedrock.knowledge_base.id
```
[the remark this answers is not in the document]
AsakerMohd: If the generic one is likely to be deprecated in the future, I'm in favor of keeping these as AWS specific ones. Does that work?

model/aws/registry.yaml
AsakerMohd: Let me add that.

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
           and [upload-part-copy](https://docs.aws.amazon.com/cli/latest/reference/s3api/upload-part-copy.html) operations.
           The `part_number` attribute corresponds to the `--part-number` parameter of the
           [upload-part operation within the S3 API](https://docs.aws.amazon.com/cli/latest/reference/s3api/upload-part.html).
+  - id: registry.aws.sqs
+    type: attribute_group
+    display_name: Amazon SQS Attributes
+    brief: >
+        This document defines attributes for AWS SQS.
+    attributes:
+      - id: aws.sqs.queue.url
```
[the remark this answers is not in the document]
AsakerMohd: I can probably still use `messaging.destination.name` in this case.

model/aws/registry.yaml
AsakerMohd: Probably won't need any more attributes so will remove this section.

model/aws/registry.yaml
AsakerMohd: Let me do that.

model/aws/registry.yaml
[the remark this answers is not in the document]
AsakerMohd: Taking a look at our use case, we can end up having both the queue URL and the queue name. Using the same attribute for both will not work in this case :(

model/gen-ai/registry.yaml
```diff
@@ -218,6 +218,11 @@ groups:
             Client-side operations are actions taken on the user's end or within the client application.
           Datastore: A tool used by the agent to access and query structured or unstructured external data for retrieval-augmented tasks or knowledge updates.
         examples: ['function', 'extension', 'datastore']
+      - id: gen_ai.data_source.id
```
lmolkova, line 221: Documenting quick research on the terminology used by others:

- Google calls them [data stores](https://cloud.google.com/dialogflow/cx/docs/concept/data-store)
- OpenAI uses [vector stores](https://platform.openai.com/docs/api-reference/vector-stores), 
- Azure OpenAI uses [data sources](https://learn.microsoft.com/en-us/azure/ai-services/openai/use-your-data-quickstart?tabs=keyless%2Ctypescript-keyless%2Cpython-new&pivots=ai-foundry-portal)
- Llamaindex calls them [data sources](https://docs.cloud.llamaindex.ai/llamacloud/data_sources)

So `gen_ai.data_source.id` sounds good to me.

model/gen-ai/registry.yaml
```diff
@@ -218,6 +218,11 @@ groups:
...
+        stability: development
+        type: string
+        brief: The data source identifier.
```
lmolkova, line 224: could we explain it a bit more? E.g. something along these lines:

```suggestion
        brief: The data source identifier.
        note: Data sources are used by AI agents and RAG applications to store grounding data. A data source may be an external database, object store, document collection, website, or any other storage system used by the GenAI agent or application. The `gen_ai.data_source.id` SHOULD match the identifier used by the GenAI system rather than a name specific to the external storage, such as a database or object store. Semantic conventions referencing `gen_ai.data_source.id` MAY also leverage additional attributes, such as `db.*`, to further identify and describe the data source.
```

model/gen-ai/registry.yaml
```diff
@@ -311,3 +316,21 @@ groups:
...
+    brief: >
+        This document defines attributes for AWS Bedrock.
+    attributes:
+      - id: gen_ai.aws.bedrock.guardrail.id
+        type: string
+        stability: experimental
+        brief: >
+          The unique identifier of the AWS Bedrock Guardrail. A [guardrail](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails.html) helps safeguard and prevent unwanted behavior from model responses or user messages.
+        examples: ["sgi5gkybzqak"]
+      - id: gen_ai.aws.bedrock.knowledge_base.id
```
lmolkova, line 331: is there a difference between knowledge_base and data_source? do we need both ids?

model/aws/sdk-spans.yml
```diff
@@ -297,3 +297,57 @@ groups:
...
+    span_kind: client
+    brief: "Attributes that exist for SQS request types."
+    attributes:
+      - ref: aws.sqs.queue.url
+        requirement_level: recommended
+      - ref: messaging.destination.name
+        requirement_level: recommended
+
+  - id: span.aws.sns.client
+    extends: span.aws.client
```
lmolkova, line 314: I don't think the result looks right in the markdown - the span does not follow messaging semantic conventions. Take a look at other messaging spans - https://github.com/open-telemetry/semantic-conventions/tree/main/docs/messaging they inherit from one of the messaging groups and include typical messaging attributes. Here's e.g. [kafka yaml](https://github.com/open-telemetry/semantic-conventions/blob/78c42c576a25743902192466cf7ff81889bf3630/model/messaging/spans.yaml#L141)

model/aws/sdk-spans.yml
```diff
@@ -297,3 +297,57 @@ groups:
...
+    stability: development
+    span_kind: client
+    brief: "Attributes that exist for Kinesis request types."
+    attributes:
+      - ref: aws.kinesis.stream_name
+        requirement_level: recommended
+
+  - id: span.aws.stepfunctions.client
+    extends: span.aws.client
+    type: span
```
lmolkova, line 335: do you need those defined as spans? I don't think any of them is referenced in markdown and given `"Attributes that exist for Step Functions request types.` explanation, it sounds like this group does not describe any span. 

If you want to document them as spans, could you please document them in markdown including span name format and link from [aws-sdk page](https://github.com/open-telemetry/semantic-conventions/blob/78c42c576a25743902192466cf7ff81889bf3630/docs/cloud-providers/aws-sdk.md#aws-service-specific-attributes)?

model/gen-ai/spans.yaml
```diff
@@ -210,3 +210,18 @@ groups:
...
+    span_kind: client
+    type: span
+    brief: >
+      Describes an AWS Bedrock operation span.
+    attributes:
+      - ref: gen_ai.aws.bedrock.guardrail.id
+        requirement_level: required
+      - ref: gen_ai.aws.bedrock.knowledge_base.id
+        requirement_level: recommended
+      - ref: gen_ai.data_source.id
```
lmolkova, line 226: I think it should be referenced in `span.gen_ai.client.create_agent` since it's applicable to other agents, not just bedrock

docs/faas/aws-lambda.md
```diff
@@ -52,6 +52,7 @@ and the [cloud resource conventions][cloud]. The following AWS Lambda-specific a
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
 | [`aws.lambda.invoked_arn`](/docs/attributes-registry/aws.md) | string | The full invoked ARN as provided on the `Context` passed to the function (`Lambda-Runtime-Invoked-Function-Arn` header on the `/runtime/invocation/next` applicable). [1] | `arn:aws:lambda:us-east-1:123456:function:myfunction:myalias` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`aws.lambda.resource_mapping.id`](/docs/attributes-registry/aws.md) | string | The UUID of the [AWS Lambda EvenSource Mapping](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html). An event source is mapped to a lambda function. It's contents are read by Lambda and used to trigger a function. | `587ad24b-03b9-4413-8202-bbd56b36e5b7` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 55: cc @open-telemetry/lambda-extension-approvers could you please take a look?

docs/gen-ai/aws-bedrock.md
```diff
@@ -0,0 +1,146 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: AWS Bedrock
+--->
+
+# Semantic conventions for AWS Bedrock operations
```
lmolkova, line 5: please add a link to this (and other new AWS docs) from https://github.com/open-telemetry/semantic-conventions/blob/78c42c576a25743902192466cf7ff81889bf3630/docs/cloud-providers/aws-sdk.md#aws-service-specific-attributes

model/aws/sdk-spans.yml
```diff
@@ -297,3 +297,57 @@ groups:
         requirement_level: recommended
       - ref: aws.s3.part_number
         requirement_level: recommended
+  
+  - id: span.aws.sqs.client
+    extends: span.aws.client
```
lmolkova, line 302: JFYI: there is an active issue against this span definition - https://github.com/open-telemetry/semantic-conventions/issues/1545, from semconv perspective we'd like to fix it at some point which may cause breaking changes if you use `rpc.*` attributes in AWS SDK instrumentations.

would appreciate your input on this.

model/gen-ai/registry.yaml
xrmx: From AWS doc "A data source contains files or content with information that can be retrieved when your knowledge base is queried.", so two different ids sounds legit to track the source of the data when you have more than one source in your knowledge base.

docs/faas/aws-lambda.md
tylerbenson: Is this available in the lambda context object?  If not, how should it be obtained?

AsakerMohd: This isn't available in the context or the lambda runtime env you're right. This is going to be populated by the AWS SDK for each language when that UUID is present. It's going to store UUID of the Event Source Mapping. Some of these operations are Create/Delete/Get/List/Update EventSourceMapping.

model/aws/sdk-spans.yml
AsakerMohd: So for this, why don't we just add `aws-api` to the list of known values here: https://github.com/open-telemetry/semantic-conventions/blob/520702f1557743fa05588999757106007fb21ba3/model/rpc/registry.yaml#L216-L240 and maybe update this md file https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/docs/cloud-providers/aws-sdk.md to make it less confusing? 

We should probably keep the rpc.system/rpc.operation for AWS because it's heavily used my AWS SDKs in multiple languages and instrumentation agents looks for these values. Changing the semcov will break all of these which wouldn't be ideal.

model/gen-ai/registry.yaml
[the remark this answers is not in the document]
AsakerMohd: Removed gen_ai prefix

model/gen-ai/registry.yaml
AsakerMohd: Added this explanation in the registry.

model/aws/sdk-spans.yml
AsakerMohd: So for this, it's meant to extend the aws attributes like rpc system etc and not messaging attributes like message.operation. AFAIK, there isn't a way to extend multiple ones from different files no?

model/aws/sdk-spans.yml
AsakerMohd: removed them from the span definition.

model/gen-ai/spans.yaml
AsakerMohd: moved the data source to be in the `span.gen_ai.client.create_agent` definition.

model/gen-ai/registry.yaml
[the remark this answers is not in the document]
lmolkova: I think it's still there, maybe you didn't push some of the changes?

model/aws/sdk-spans.yml
lmolkova: AWS SDK works on top of HTTP, correct? It's not an RPC protocol or middleware, but an SDK on top of REST API, so it's not an RPC system - PTAL at the https://github.com/open-telemetry/semantic-conventions/pull/1622#pullrequestreview-2503504000

docs/messaging/sns.md
```diff
@@ -0,0 +1,51 @@
...
+described on this page.
+
+<!-- semconv span.aws.sns.client -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
```
lmolkova, line 21: are there any general messaging conventions AWS SDK would report on SNS and SQS spans?

model/aws/registry.yaml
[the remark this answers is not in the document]
lmolkova: > Taking a look at our use case, we can end up having both the queue URL and the queue name. Using the same attribute for both will not work in this case :(

I'm not sure I understand the concern. From messaging  semconv perspective we recommend recording this information in the following way:

- `server.address = "sqs.us-east-1.amazonaws.com"`
- `server.port = 443` (or not set)
- `messaging.destination.name = MyQueue`

The question is how to record account id (123456789012), so perhaps

`[cloud.account.id](https://github.com/open-telemetry/semantic-conventions/blob/6814d83dac7ebd6b29dada91c669a99275f674c8/model/cloud/registry.yaml#L46C13-L46C29) = "123456789012"`

?

model/aws/sdk-spans.yml
lmolkova: there is no way to extend multiple, but you can extend one and reference attributes one-by-one (that are applicable to SNS/SQS), I would recommend extending messaging and this way avoiding referencing `rpc` attributes.

docs/faas/aws-lambda.md
tylerbenson: I suggest mentioning such, otherwise it would be confusing to not see this in instrumentation.

model/aws/registry.yaml
```diff
@@ -481,3 +488,96 @@ groups:
...
+        This document defines attributes for AWS Kinesis.
+    attributes:
+      - id: aws.kinesis.stream_name
+        type: string
+        stability: experimental
+        brief: >
+          The name of the AWS Kinesis [stream](https://docs.aws.amazon.com/streams/latest/dev/introduction.html) the request refers to. Corresponds to the `--stream-name` parameter of the
+          Kinesis [describe-stream](https://docs.aws.amazon.com/cli/latest/reference/kinesis/describe-stream.html) operation.
+        examples: ["some-stream-name"]
+  - id: registry.aws.stepfunctions
```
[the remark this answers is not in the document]
tylerbenson: I don't have sufficient experience with step functions to weigh in here.

lmolkova:
@AsakerMohd this PR got stale, would you like to keep working on it?

docs/faas/aws-lambda.md
AsakerMohd: Mentioned it in the description.

model/aws/sdk-spans.yml
AsakerMohd: I moved it to messaging to extend it there.

model/gen-ai/registry.yaml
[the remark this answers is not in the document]
AsakerMohd: Yeah sorry it was missed. It should be removed now.

model/aws/registry.yaml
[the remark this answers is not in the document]
lmolkova: it seems aws step functions are not related to lambda, so the only remaining question if we want to have `step_function` or `stepfunction` - the former would align with general naming conventions

model/messaging/spans.yaml
```diff
@@ -311,3 +311,23 @@ groups:
...
+  - id: messaging.aws.sqs
+    type: attribute_group
+    stability: development
+    extends: attributes.messaging.trace.minimal
+    brief: "Attributes that exist for SQS request types."
+    attributes:
+      - ref: aws.sqs.queue.url
+        requirement_level: recommended
+      - ref: messaging.destination.name
+        requirement_level: recommended
```
lmolkova, line 324: if you feel it's goo, please add
nit

```suggestion
        requirement_level: recommended
     - ref: aws.request_id
        requirement_level: recommended 
```

model/aws/registry.yaml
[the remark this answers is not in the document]
lmolkova: talked offline, can do `messaging.destination.name = MyQueue` and `aws.sqs.queue.url` or put url into `messaging.destination.name` - either way works

model/aws/sdk-spans.yml
lmolkova: fixed

lmolkova (approved):

model/messaging/spans.yaml
AsakerMohd: Added!

docs/attributes-registry/aws.md
```diff
@@ -165,3 +189,36 @@ This applies in particular to the following operations:
...
+|---|---|---|---|---|
+| <a id="aws-secretsmanager-secret-arn" href="#aws-secretsmanager-secret-arn">`aws.secretsmanager.secret.arn`</a> | string | The ARN of the Secret stored in the Secrets Mangger | `arn:aws:secretsmanager:us-east-1:123456789012:secret:SecretName-6RandomCharacters` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+## Amazon SNS Attributes
+
+This document defines attributes for AWS SNS.
+
+| Attribute | Type | Description | Examples | Stability |
+|---|---|---|---|---|
+| <a id="aws-sns-topic-arn" href="#aws-sns-topic-arn">`aws.sns.topic.arn`</a> | string | The ARN of the AWS SNS Topic. An Amazon SNS [topic](https://docs.aws.amazon.com/sns/latest/dg/sns-create-topic.html) is a logical access point that acts as a communication channel. | `arn:aws:sns:us-east-1:123456789012:mystack-mytopic-NZJ5JSMVGFIE` | ![Development](https://img.shields.io/badge/-development-blue) |
```
xrmx, line 207: In python botocore instrumentation we send this as `messaging.destination.name`

xrmx (approved):
The `gen_ai` part looks good to me

model/aws/registry.yaml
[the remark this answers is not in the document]
AsakerMohd: updated it to step_function

model/aws/registry.yaml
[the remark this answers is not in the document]
AsakerMohd: we decided it'll be better to have both `messaging.destination.name = MyQueue` and `aws.sqs.queue.url`

docs/attributes-registry/aws.md
AsakerMohd: We can continue using that and also add the `aws.sns.topic.arn` attribute to be more specific.

Issue #1662, opened 2024-12-09
`process.open_file_descriptor.count` and `process.context_switches` both have unexpected unit `{count}`


trask:
### Area(s)

area:process

### What happened?

`process.open_file_descriptor.count` and `process.context_switches` both have unexpected unit `{count}`.

Probably should be something like `{open_file_descriptor}` and `{context_switch}`?

### Semantic convention version

v1.29.0

trask:
cc @open-telemetry/semconv-system-approvers

mx-psi:
Discussed on the meeting on 2024-12-12, the name of these metrics was changed on open-telemetry/semantic-conventions/pull/330 and we agree the units should have been changed as well

PR #1797, opened 2025-01-23
Fix units of process.open_file_descriptor.count and process.context_switches

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1662

## Changes

This PR fixes the units of `process.open_file_descriptor.count` and `process.context_switches` metrics.
I was not sure if it should be `{open_file_descriptor}` or just `{file_descriptor}` but went with what was proposed in https://github.com/open-telemetry/semantic-conventions/issues/1662.


mx-psi (approved):

braydonk (approved):

model/process/metrics.yaml
```diff
@@ -81,15 +81,15 @@ groups:
     stability: experimental
     brief: "Number of file descriptors in use by the process."
     instrument: updowncounter
-    unit: "{count}"
+    unit: "{open_file_descriptor}"
```
lmolkova, line 84: maybe `{descriptor}` or `{file_descriptor}`?

I assume that if we had some other counter for file descriptors, it'd be right to use the same unit

lmolkova (approved):

model/process/metrics.yaml
arminru: ^ +1

arminru (approved):

model/process/metrics.yaml
ChrsMark: Answerwing with my original concern from the PR description :) : 

> I was not sure if it should be {open_file_descriptor} or just {file_descriptor} but went with what was proposed in https://github.com/open-telemetry/semantic-conventions/issues/1662.

It makes sense to change it to `file_descriptor` all together as @braydonk suggests at https://github.com/open-telemetry/semantic-conventions/pull/1797#issuecomment-2610104207. 

I'm fine if we want to merge this one and iterate on it, or if we want to fix in a different PR entirely.

braydonk: Let's merge this, I opened #1798 and if folks all agree we can make the change as part of that.

ChrsMark:
@open-telemetry/specs-semconv-maintainers shall we merge this?

model/process/metrics.yaml
lmolkova: just to clarify, should we merge `{open_file_descriptor}` and we'd revisit it in #1798? 
I'd prefer to change to `{file_descriptor}` now and update, if needed, in #1798, but it's not a strong opinion

braydonk: Yeah I was suggesting we merge it with the `{file_descriptor}` unit now, and revisit the metric name itself in the future.

ChrsMark: Ah, I missed we decided to change it to `{file_descriptor}` now, sorry folks. Changed that now.

braydonk (approved):

lmolkova (approved):

PR #1807, opened 2025-01-24
Update spec repo links to v1.41.0

opentelemetrybot:
Update spec repo links to `v1.41.0`.

PR #1808, opened 2025-01-24
[chore] opentelemetrybot doesn't have permission to update labels

trask:
the PR is created successfully (#1807), but the labeling step fails

another option is to give opentelemetrybot triage permissions on the repo, but i'd prefer to not give it any special permissions at this point since it's using the same PAT as all the other repos

### DIFF
--- a/.github/workflows/auto-update-spec-repo-links.yml
+++ b/.github/workflows/auto-update-spec-repo-links.yml
@@ -22,7 +22,7 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           current_version=$(grep "PREVIOUS_SPECIFICATION_VERSION=v.*" \
-                                 internal/tools/update_specification_version.sh \
+                                 .github/scripts/update-spec-repo-links.sh \
                             | sed "s/PREVIOUS_SPECIFICATION_VERSION=//")
           latest_version=$(gh release view \
                                --repo open-telemetry/opentelemetry-specification \


lmolkova (approved):

.github/workflows/auto-update-spec-repo-links.yml
```diff
@@ -77,5 +77,4 @@ jobs:
           git push --set-upstream origin $branch
           gh pr create --title "$message" \
```
arminru, line 78:
```suggestion
          gh pr create --title "[chore] $message" \
```
This would also make it skip changelog verification.

arminru (approved):

.github/workflows/auto-update-spec-repo-links.yml
arminru: https://github.com/open-telemetry/semantic-conventions/actions/runs/13010412553/job/36286799062
🤔

arminru: Ah, not a problem for the bot-opened PRs but for this one here:

Renaming does not trigger an event currently, so the check runs against the old name until one would synchronize (which trask will have to do anyway before merging):

https://github.com/open-telemetry/semantic-conventions/blob/5b24784e032ada99312d5a17d6970da434f5ce88/.github/workflows/changelog.yml#L9

We could also add the `edited` event to the workflow triggers:

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

arminru: https://github.com/open-telemetry/semantic-conventions/pull/1837

trask: nice!

PR #1844, opened 2025-01-28
[CI] Report non-normalized link from docs into docs

chalin:
- Adds `normalized-link-check` target that detects non-normalized links as those fixed in #1843
- Adds `normalized-link-check` as a pre-check to the already existing `markdown-link-check` (so it will be check via the existing workflow)

### Note - expect this PR to fail it contains a _seeded error_ to test the new check

/cc @trask @lmolkova

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,13 +5,14 @@ PWD := $(shell pwd)
 # Determine OS & Arch for specific OS only tools on Unix based systems
 OS := $(shell uname | tr '[:upper:]' '[:lower:]')
 ifeq ($(OS),darwin)
-	SED := gsed
+	SED ?= gsed
 else
-	SED := sed
+	SED ?= sed
 endif
 
 TOOLS_DIR := $(PWD)/internal/tools
 
+MARKDOWN_LINK_CHECK_ARG= # pass extra arguments such as --exclude '^http'
 MISSPELL_BINARY=bin/misspell
 MISSPELL = $(TOOLS_DIR)/$(MISSPELL_BINARY)
 
@@ -66,14 +67,24 @@ misspell:	$(MISSPELL)
 misspell-correction:	$(MISSPELL)
 	$(MISSPELL) -w $(ALL_DOCS)
 
+.PHONY: normalized-link-check
+normalized-link-check:
+	@if grep -R '\.\./docs/' docs model/database; then \
+		echo "\nERROR: Found occurrences of '../docs/'; see above."; \
+		echo "       Remove the '../docs/' from doc page links referencing doc pages."; \
+		exit 1; \
+	else \
+		echo "Normalized-link check passed."; \
+	fi
+
 .PHONY: markdown-link-check
-markdown-link-check:
+markdown-link-check: normalized-link-check
 	docker run --rm \
 		--mount 'type=bind,source=$(PWD),target=/home/repo' \
 		lycheeverse/lychee \
 		--config home/repo/.lychee.toml \
 		--root-dir /home/repo \
-		--verbose \
+		--verbose $(MARKDOWN_LINK_CHECK_ARG) \
 		home/repo
 
 .PHONY: markdown-link-check-changelog-preview
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,8 @@ markdown-link-check: normalized-link-check
 		lycheeverse/lychee \
 		--config home/repo/.lychee.toml \
 		--root-dir /home/repo \
-		--verbose $(MARKDOWN_LINK_CHECK_ARG) \
+		--verbose \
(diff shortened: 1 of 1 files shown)


chalin:
The following confirms that the check detects non-normalized paths, see https://github.com/open-telemetry/semantic-conventions/actions/runs/13020486464/job/36319834545?pr=1844:

```
Run make markdown-link-check
docs/attributes-registry/db.md:**[7] `db.query.text`:** For sanitization see [Sanitization of `db.query.text`](../../docs/database/database-spans.md#sanitization-of-dbquerytext).
model/database/registry.yaml:          For sanitization see [Sanitization of `db.query.text`](../../docs/database/database-spans.md#sanitization-of-dbquerytext).

ERROR: Found occurrences of '../docs/'; see above.
       Remove the '../docs/' from doc page links referencing doc pages.
make: *** [Makefile:[7](https://github.com/open-telemetry/semantic-conventions/actions/runs/13020486464/job/36319834545?pr=1844#step:4:8)2: normalized-link-check] Error 1
```

I'll remove the seeded error now.

Makefile
```diff
@@ -5,13 +5,14 @@ PWD := $(shell pwd)
 # Determine OS & Arch for specific OS only tools on Unix based systems
 OS := $(shell uname | tr '[:upper:]' '[:lower:]')
 ifeq ($(OS),darwin)
-	SED := gsed
+	SED ?= gsed
```
chalin, line 8: I don't have `gsed` in my environment. By doing this I can pass my sed command via the command line. Hopefully this is ok with you.

Makefile
```diff
@@ -66,14 +67,24 @@ misspell:	$(MISSPELL)
...
 .PHONY: markdown-link-check
-markdown-link-check:
+markdown-link-check: normalized-link-check
 	docker run --rm \
 		--mount 'type=bind,source=$(PWD),target=/home/repo' \
 		lycheeverse/lychee \
 		--config home/repo/.lychee.toml \
 		--root-dir /home/repo \
-		--verbose \
+		--verbose $(MARKDOWN_LINK_CHECK_ARG) \
```
chalin, line 87: Convenient to allow external links to be excluded or any other arguments to be passed to Lychee.

Makefile
```diff
@@ -66,14 +67,24 @@ misspell:	$(MISSPELL)
 misspell-correction:	$(MISSPELL)
 	$(MISSPELL) -w $(ALL_DOCS)
 
+.PHONY: normalized-link-check
+normalized-link-check:
+	@if grep -R '\.\./docs/' docs model/database; then \
```
chalin, line 72: Note: this checks both `docs` and `model/database`.

trask (approved):

Makefile
lmolkova, line 87: nit:

```suggestion
		--verbose \
		$(MARKDOWN_LINK_CHECK_ARG) \
```

just to make it more obvious it's not the value  of `--verbose`

Makefile
lmolkova: why `model/database`  only and not everything in `model`? Is it just an example?

Makefile
```diff
@@ -66,14 +67,24 @@ misspell:	$(MISSPELL)
...
+		echo "\nERROR: Found occurrences of '../docs/'; see above."; \
```
lmolkova, line 73: BTW, should we just enforce absolute paths everywhere? relative is always error prone

Makefile
```diff
@@ -5,13 +5,14 @@ PWD := $(shell pwd)
...
-	SED := gsed
+	SED ?= gsed
 else
-	SED := sed
+	SED ?= sed
 endif
 
 TOOLS_DIR := $(PWD)/internal/tools
 
+MARKDOWN_LINK_CHECK_ARG= # pass extra arguments such as --exclude '^http'
```
lmolkova, line 15: would we expect someone to modify the value in the file or could I just pass it in cmd line somehow? (Sorry, I'm a noob when it comes to makefiles)

chalin: The variable is empty, the part `# ...` is just a comment. So yes, a dev would have to pass a value on the command line.

Makefile
chalin: Oh! Good point. I've changed it to `model/[a-z]*` and added the following explanation:

```makefile
# NOTE: we check `model/[a-z]*` to avoid `model/README.md`, which contains
# valid occurrences of `../docs/`.
```

Makefile
chalin: I think that relative paths are useful. Besides, I might be able to eliminate the need for this normalization, but I'll experiment with that later, once I'm done with https://github.com/open-telemetry/opentelemetry.io/issues/6101.

lmolkova (approved):

Issue #1821, opened 2025-01-25
[editorial] Inline Hugo front matter


chalin:
This is the semconv equivalent of:

- https://github.com/open-telemetry/opentelemetry-specification/issues/4029

Inlining will allow us to simplify the importing and processing of doc pages into OTel.io

/cc @open-telemetry/docs-maintainers

joaopgrassi:
Hi @chalin! Is this something you will be working on?

chalin:
Yes, I've been actively working on this for the past couple of weeks. Why do you ask? I hope to wrap this up this month :).

PR #1846, opened 2025-01-28
[editorial] Drop path_base_for_github_subdir from front matter

chalin:
- Contributes to #1821
- Drops `path_base_for_github_subdir` from the Hugo front matter for all `docs/**/README.md` files. Note that we keep the cascaded value in the docs root readme.


trask (approved):

lmolkova (approved):
I have no idea what `path_base_for_github_subdir ` does, but I trust you @chalin  :)

chalin:
> I have no idea what `path_base_for_github_subdir` does, but I trust you @chalin :)

@lmolkova - every doc page on OTel.io has a "View source" link. [`path_base_for_github_subdir`](path_base_for_github_subdir) is used to compute the value of the "View source" URL for pages whose source isn't the OTel.io repo itself. `README.md` files were special in that they needed to be renamed, and so each such file required its own path_base config. We won't need that any more (to know why, see https://github.com/open-telemetry/opentelemetry.io/pull/6103). I hope that helps a bit to understand the purpose of that field. :) 

(Btw, the "View source" is temporarily broken. A fix should land soon: https://github.com/open-telemetry/opentelemetry.io/pull/6107.)

Issue #1755, opened 2025-01-17
Naming guidance evolving practices


lmolkova:
This issue collects evolving naming best(?) practices that are identified, but are not yet documented.

lmolkova:
pieces of wisdom from @trask https://github.com/open-telemetry/semantic-conventions/pull/1734#discussion_r1919466072

- Use `_` in `adjective_noun`

> - maybe a tip we can give for using _ is adjective_noun. since adjective can't be its own namespace.
> - similar to returned_rows, speculative_execution, active_instance

- Attribute name pluralization: use plural for arrays

lmolkova:
More things to consider:

- use `.count` vs `_count` 
- `is_*` for boolean flags

lmolkova:
also: `state` vs `status` vs `phase` vs `mode` (as in `cpu.mode` or `network.connection.state`) - seems to be domain-specific, no convention

lmolkova:
https://github.com/open-telemetry/semantic-conventions/pull/1742#discussion_r1927639614

Prioritizing domain-specific naming over otel naming conventions (e.g. k8s `daemonset` vs `daemon_set` since that's how it's used in k8s)

trask:
I think we are solid when it comes to choosing underscore vs dot.

But I think we have chaos when it comes to choosing underscore vs straight concat (e.g. `nodejs` vs `node_js`).

Looking at just the top-level namespaces:

- `aspnetcore` - ❓ why not `asp_net_core`
- `cicd` - ❓ why not `ci_cd`
- `cloudfoundry` - ✅ follows TLD rule for top-level namespace (https://cloudfoundry.org)
- `cloudevents` - ✅ follows TLD rule for top-level namespace (https://cloudevents.io)
- `dotnet` - why not `dot_net`?
- `graphql` - ✅ follows TLD rule for top-level namespace (https://graphql.org)
- `nodejs` - ✅ follows TLD rule for top-level namespace (https://nodejs.org)
- `opentracing` - ✅ follows TLD rule for top-level namespace (https://opentracing.io)
- `signalr` - ❓ why not `signal_r`
- `v8js` - ❓ why not `v8_js`
- `webengine` - ❓ why not `web_engine`
- `rabbitmq`* - ✅ follows TLD rule for top-level namespace (https://rabbitmq.com)
- `rocketmq`* - ❓ why not `rocket_mq`
[...]

PR #1848, opened 2025-01-29
Make replication controller metrics consistent with resource attributes

ChrsMark:
## Changes

Follow up from https://github.com/open-telemetry/semantic-conventions/pull/1742 to follow the naming of K8s api for resources. This PR applies the following changes:

```
k8s.replication_controller.desired_pods -> k8s.replicationcontroller.desired_pods
k8s.replication_controller.available_pods -> k8s.replicationcontroller.available_pods
``` 


.chloggen/fix_k8s_metrics.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: k8s
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Change k8s.replication_controller metrics to k8s.replicationcontroller
```
trask, line 13: sorry I missed the discussion in https://github.com/open-telemetry/semantic-conventions/pull/1742#discussion_r1926306420

I feel like this opens a pretty big door across semconv for debating when to apply underscores, because we could point to lots of domain-specific apis that concat words instead of underscoring them

I understand (and agree) with `replicaset`, `statefulset`, `daemonset`, even outside of the k8s domain. could we just make an exception for those (`*set` things)?

lmolkova: to me this boils down to https://github.com/open-telemetry/semantic-conventions/issues/1755#issuecomment-2610956092: k8s uses `replicationcontroller` for everything and we're being consistent with the domain.

Similarly to https://github.com/open-telemetry/semantic-conventions/blob/b865f63bc7ba3039ad87504d75ee40104149d1bd/model/azure/registry.yaml#L78-L80

there is some external source of this value and we pick this external spelling for attributes limited to this use-case.

trask: the benefit of doing this for attribute _values_ seems clear to me (e.g. avoids instrumentations having to map domain-specific values to something else)

the benefit to doing this for attribute _names_ is less clear to me

lmolkova: how would we approach something like `replicaset`, `statefulset`, `daemonset`  vs `replication_controller` in the future? There are some other examples in https://github.com/open-telemetry/semantic-conventions/pull/1742#discussion_r1926519917 like `cronjob` or `apiservice` where we'd have ambiguity

lmolkova: also `mariadb` and `cosmosdb` could arguably be `maria_db` and `cosmos_db`. I'm not suggesting it, I'm saying that I genuinely don't know when and why we apply `_`

model/k8s/deprecated/metrics-deprecated.yaml
```diff
@@ -0,0 +1,31 @@
+groups:
+  - id: metric.k8s.replication_controller.desired_pods
+    type: metric
+    metric_name: k8s.replication_controller.desired_pods
+    stability: experimental
+    deprecated: "Replaced by `k8s.replicationcontroller.desired_pods`."
+    brief: "Deprecated, use `k8s.replication_controller.desired_pods` instead."
```
trask, line 7:
```suggestion
    brief: "Deprecated, use `k8s.replicationcontroller.desired_pods` instead."
```

model/k8s/deprecated/metrics-deprecated.yaml
```diff
@@ -0,0 +1,31 @@
...
+      This metric SHOULD, at a minimum, be reported against a
+      [`k8s.replicationcontroller`](../resource/k8s.md#replicationcontroller) resource.
+    instrument: updowncounter
+    unit: "{pod}"
+  - id: metric.k8s.replication_controller.available_pods
+    type: metric
+    metric_name: k8s.replication_controller.available_pods
+    stability: experimental
+    deprecated: "Replaced by `k8s.replicationcontroller.available_pods`."
+    brief: "Deprecated, use `k8s.replication_controller.available_pods` instead."
```
trask, line 22:
```suggestion
    brief: "Deprecated, use `k8s.replicationcontroller.available_pods` instead."
```

trask (approved):
I think we [have some work to do](https://github.com/open-telemetry/semantic-conventions/issues/1755#issuecomment-2626226669) to establish guidance around when to concat words with `_` and when to concat with nothing, but I think that can take place outside of this PR

ChrsMark:
@open-telemetry/semconv-k8s-approvers could you give a +1 on this so as we can merge it?

AlexanderWert (approved):

PR #1863, opened 2025-02-03
Clarify notes on system-specific databases for operation and collection names

### LINKED ISSUE #1573: Override attribute notes for `db.collection.name` and `db.operation.name` to be more precise for specific database vendors
See
* https://github.com/open-telemetry/semantic-conventions/pull/1566#discussion_r1837003099
* https://github.com/open-telemetry/semantic-conventions/pull/1566#discussion_r1837010418
* https://github.com/open-telemetry/semantic-conventions/pull/1566#discussion_r1837018528
* https://github.com/open-telemetry/semantic-conventions/pull/1566#discussion_r1837029078

probably others...

lmolkova:
Fixes #1573

Update `db.collection.name` and `db.operation.name` notes when database does not support cross-table query or multiple operations in the same query.


model/database/spans.yaml
```diff
@@ -276,6 +272,11 @@ groups:
       - ref: cassandra.consistency.level
       - ref: db.collection.name
         brief: The name of the Cassandra table that the operation is acting upon.
+        note: |
+          It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.
+
+          For batch operations, if the individual operations are known to have the same collection name
```
maryliag, line 278: For the other example, are say it should be prepended by `BATCH `, `MULTI `, `PIPELINE `, etc. Isn't there anything here to indicate the same? Can't we use `BATCH ` too?

model/database/spans.yaml
```diff
@@ -2,7 +2,6 @@ groups:
   - id: trace.db.common.minimal
     extends: attributes.db.client.minimal
     type: attribute_group
-    stability: experimental
```
maryliag, line 5: just to confirm, all these traces attributes are not `experimental`? are they stable?

lmolkova: great question. attributes groups are weird. They are not a telemetry signal - they just group attributes. 

Attributes, metrics, spans, resources all require stability property, but it means nothing on the attribute groups.
It's not documented in the https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/semantic-convention-groups.md#semantic-convention-groups:

> Groups that have attribute_group type do not describe semantic convention and are used for auxiliary purposes.

But if you think it's confusing or a bad idea to skip stability for them in some other way, please let me know, I want to hear it :)

model/database/spans.yaml
lmolkova: I've tried to adjust generic `BATCH ` recommendation to redis - it does not use term batching, but it has transactions (`MULTI`/`EXEC` command) and pipelining when multiple commands are sent one after another without awaiting result and them results are returned once pipeline is executed.

e.g.

```java
pipe = jedis.pipelined();

Response<String> resp0 = pipe.get("seat:0");  // this is a promise, not the actual response
Response<String> resp3 = pipe.get("seat:3");  // commands could be the same or different
Response<String> resp4 = pipe.get("seat:4");

pipe.sync(); // this is where things are going to happen and what would be called `PIPELINE` operation. 
```

and for transactions

```java
jedis.multi();

trans.incrBy("counter:1", 1);   // commands could be the same or different
trans.incrBy("counter:2", 2);
trans.incrBy("counter:3", 3);

trans.exec(); // this would be `MULTI` operation 
```

model/database/spans.yaml
maryliag: My confusion was mostly because the traces got their tag removed and the span was updated from `experimental` to `development`, but I might have missed some info about how this distinction was made.
That was pretty much me trying to find out how to check what should be `stable` vs `experimental` vs `development`

model/database/spans.yaml
maryliag: any experts on Redis around here to let us know if creating a distinction between these cases (pipeline, transactions) is helpful? 😅 
in general I think having some indication that was a "_group_" (batch/multi/etc) operation would be helpful

trask (approved):

maryliag (approved):

Issue #1878, opened 2025-02-05
Adding Agent version and Agent role to Agent Semantic Convention


gyliu513:
### Area(s)

area:gen-ai

### What's missing?

- Agent Role: Agent roles are classified based on job responsibilities:
  – Worker. Agents can take on roles such as “Researcher” or “Writer,” functioning as part of a team with specific skills and responsibilities.
  – Developers can also create Coordinator/Supervisor agents, under which multiple worker agents operate. These subordinate agents follow the instructions of the supervisor.
- Agent Version
  - AI Agent Version refers to a specific iteration of an AI-powered agent, incorporating updates, improvements, and optimizations in functionality, performance, and learning capabilities for enhanced decision-making and task execution.

This will be leveraged by https://github.com/open-telemetry/semantic-conventions-genai/issues/11

@lmolkova ^^

lmolkova:
version added in https://github.com/open-telemetry/semantic-conventions/pull/3428, see https://github.com/open-telemetry/semantic-conventions/pull/1879#discussion_r1943477219 on the role

I think this can be closed now, feel free to reopen if role is still relevant

PR #1879, opened 2025-02-05
Added version and role to Agent Semantic Convention

gyliu513:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1878

## Changes

### DIFF
--- a/model/gen-ai/registry.yaml
+++ b/model/gen-ai/registry.yaml
@@ -192,6 +192,37 @@ groups:
         type: string
         brief: Human-readable name of the GenAI agent provided by the application.
         examples: ["Math Tutor", "Fiction Writer"]
+      - id: gen_ai.agent.version
+        stability: development
+        type: string
+        brief: Human-readable version of the GenAI agent provided by the application.
+        note: >
+          AI Agent Version refers to a specific iteration of an AI-powered agent, incorporating
+          updates, improvements, and optimizations in functionality, performance, and learning
+          capabilities for enhanced decision-making and task execution.
+        examples: ["1.0"]
+      - id: gen_ai.agent.role
+        stability: development
+        type:
+          members:
+            - id: worker
+              value: "worker"
+              brief: 'Agent with specific skills and responsibilities'
+              stability: development
+            - id: coordinator
+              value: "coordinator"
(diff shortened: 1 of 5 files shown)


model/gen-ai/registry.yaml
```diff
@@ -192,6 +192,37 @@ groups:
...
+        stability: development
+        type: string
+        brief: Human-readable version of the GenAI agent provided by the application.
+        note: >
+          AI Agent Version refers to a specific iteration of an AI-powered agent, incorporating
+          updates, improvements, and optimizations in functionality, performance, and learning
+          capabilities for enhanced decision-making and task execution.
+        examples: ["1.0"]
+      - id: gen_ai.agent.role
+        stability: development
```
lmolkova, line 205: are there some existing agent frameworks or services that support it? How would it be different from the name and/or description?

model/gen-ai/registry.yaml
```diff
@@ -192,6 +192,37 @@ groups:
         type: string
         brief: Human-readable name of the GenAI agent provided by the application.
         examples: ["Math Tutor", "Fiction Writer"]
+      - id: gen_ai.agent.version
```
lmolkova, line 195: are there some existing frameworks out there that have a concept of agent version?

lmolkova: let's not add version now  then - we can always add attributes when we need them. Also, it's a good question whether they should be supplied in instrumentation scope (when creating tracer/meter) vs in span attributes.

model/gen-ai/registry.yaml
lmolkova: Is it used by other frameworks? We can add have crewai-specific conventions, but I'm not sure if it's used by anyone else, so it might not be the right candidate for the general attribute.

Also, would you classify it as something essential for agent observability? I mean humans can figure out the role from name and description and not sure what visualization/tools would do with role information. So is there some specific scenario where it'd be important to have the role formalized?

gyliu513: MetaGPT also have roles concept https://github.com/geekan/MetaGPT/tree/main/metagpt/roles , like architect, assistant, engineer, researcher etc. But seems MetaGPT is using some other way to describe roles, let me do more research for this, thanks @lmolkova

PR #1880, opened 2025-02-06
Rework device events

breedx-splk:
Also rolls back deprecation for required registry attributes.

## Changes

Migrating `device.app.lifecycle` event from using `body` to using attributes for `ios.state` and `android.state`. This also requires reintroducing those attributes (which were previously deprecated).


model/device/events.yaml
```diff
@@ -16,77 +16,73 @@ groups:
...
+        stability: development
+        requirement_level:
+          conditionally_required: if and only if `os.name` is `ios`
+        note: >
+          The iOS lifecycle states are defined in the [UIApplicationDelegate documentation](https://developer.apple.com/documentation/uikit/uiapplicationdelegate),
+          and from which the `OS terminology` column values are derived.
+        brief: >
+          This attribute represents the state the application has transitioned into at the occurrence of the event.
+        examples: ["active"]
+        type: enumeration
```
trask, line 30:
```suggestion
```

model/ios/registry.yaml
```diff
@@ -0,0 +1,43 @@
...
+      The iOS platform on which the iOS application is running.
+    attributes:
+      - id: ios.state
+        stability: development
+        brief: >
+          This attribute represents the state of the application.
+        note: >
+          The iOS lifecycle states are defined in the [UIApplicationDelegate documentation](https://developer.apple.com/documentation/uikit/uiapplicationdelegate),
+          and from which the `OS terminology` column values are derived.
+        type: enumeration
```
trask, line 15:
```suggestion
```

model/device/events.yaml
```diff
@@ -16,77 +16,72 @@ groups:
...
+        stability: development
+        requirement_level:
+          conditionally_required: if and only if `os.name` is `android`
+        brief: >
+          This attribute represents the state the application has transitioned into at the occurrence of the event.
+        note: >
+          The Android lifecycle states are defined in [Activity lifecycle callbacks](https://developer.android.com/guide/components/activities/activity-lifecycle#lc),
+          and from which the `OS identifiers` are derived.
+        examples: ["created"]
+        type: enumeration
```
trask, line 68:
```suggestion
```

model/ios/registry.yaml
```diff
@@ -1,45 +1,43 @@
 groups:
-  - id: registry.ios.deprecated
+  - id: registry.ios
     type: attribute_group
-    display_name: Deprecated iOS Attributes
+    display_name: iOS Attributes
     brief: >
-        The iOS platform on which the iOS application is running.
+      The iOS platform on which the iOS application is running.
```
lmolkova, line 6:
```suggestion
      This group describes iOS-specific attributes.
```

model/ios/registry.yaml
```diff
@@ -1,45 +1,43 @@
...
     attributes:
       - id: ios.state
```
lmolkova, line 8: I know  it's been like this before, but `ios.state` naming does not seem quite right. It looks like it's a state of iOS and not the app.  If we have some other state on ios like `ios.wifi.state`, it would be even more confusing to have `ios.state` without any additional qualifier.

Maybe we can do `ios.app.state` and `andrioid.app.state` ?

model/ios/registry.yaml
```diff
@@ -1,45 +1,43 @@
...
-        stability: experimental
-        deprecated: "Moved to a payload field of `device.app.lifecycle`."
+        stability: development
+        brief: >
+          This attribute represents the state of the application.
         note: >
           The iOS lifecycle states are defined in the [UIApplicationDelegate documentation](https://developer.apple.com/documentation/uikit/uiapplicationdelegate),
           and from which the `OS terminology` column values are derived.
```
lmolkova, line 14: what are the `OS terminology` and `OS identifiers` things?

model/device/events.yaml
```diff
@@ -16,77 +16,20 @@ groups:
...
-              stability: development
+    attributes:
+      - ref: ios.state
+        requirement_level:
+          conditionally_required: if and only if `os.name` is `ios`
+        note: >
+          The iOS lifecycle states are defined in the [UIApplicationDelegate documentation](https://developer.apple.com/documentation/uikit/uiapplicationdelegate),
+          and from which the `OS terminology` column values are derived.
+        brief: >
+          This attribute represents the state of the application.
```
lmolkova, line 27:
```suggestion
```

note and brief seem to match the definition, no need to repeat here

model/device/events.yaml
```diff
@@ -16,77 +16,20 @@ groups:
...
+      - ref: android.state
+        requirement_level:
+          conditionally_required: if and only if `os.name` is `android`
+        brief: >
+          This attribute represents the state of the application.
+        note: >
+          The Android lifecycle states are defined in [Activity lifecycle callbacks](https://developer.android.com/guide/components/activities/activity-lifecycle#lc),
+          and from which the `OS identifiers` are derived.
```
lmolkova, line 35:
```suggestion
```

same, no need to repeat the definition

.chloggen/1880.yaml
```diff
@@ -0,0 +1,8 @@
+change_type: enhancement
```
lmolkova, line 1: We should mark it as breaking (breaking enhancement) since body fields will no longer be in the payload

model/ios/registry.yaml
breedx-splk: It's less confusing when it's simply scope to an event. 🙃

.chloggen/1880.yaml
breedx-splk: so what's the literal string that should go in here then?

model/ios/registry.yaml
breedx-splk: I'm literally just moving things around and not trying to rewrite/redefine/clarify too much.

.chloggen/1880.yaml
lmolkova:
```suggestion
change_type: breaking 
```

lmolkova (approved):
LGTM!

model/android/deprecated/registry-deprecated.yaml
```diff
@@ -7,7 +7,7 @@ groups:
     attributes:
       - id: android.state
         stability: development
-        deprecated: "Replaced by `device.app.lifecycle`."
+        deprecated: "Renamed to `android.app.state`"
         brief: >
           Deprecated use the `device.app.lifecycle` event definition including
           `android.state` as a payload field instead.
```
trask, line 13: looks like brief needs to be updated also?

trask (approved):

PR #1890, opened 2025-02-08
Set span kind on span definitions or change ambiguous spans to attribute groups

### LINKED ISSUE #1513: Make span_kind required in yaml and break down multi-kind span definitions
Following spans (after https://github.com/open-telemetry/semantic-conventions/pull/1512) would remain kindless as they describe multiple kinds of spans (or it's not clear what it should be)

- `span.rpc.connect_rpc`
- `span.rpc.grpc`
- `span.rpc.jsonrpc`
- `span.faas`
- `span.faas.datasource`
- `span.faas.timer`
- `span.messaging`
- `span.aws.lambda`

We should break them down and describe individual kinds (as they have different attribute sets and identities)

lmolkova:
Fixes #1513

In just released weaver 0.13.0 we'll start requiring span kinds. This PR adds span kind to span definitions that describe a single span kind. It changes ambiguous definitions to attribute groups.


model/faas/spans.yaml
```diff
@@ -35,9 +36,10 @@ groups:
       - ref: faas.document.time
       - ref: faas.document.name
 
-  - id: span.faas.timer
+  - id: span.faas.timer.server
```
lmolkova, line 39: FaaS conventions recommend using `SERVER` for all incoming invocations

https://github.com/open-telemetry/semantic-conventions/blob/4f77620fe731c10d40f7d50c543d4e5c73a46ebf/docs/faas/faas-spans.md?plain=1#L130

model/rpc/spans.yaml
```diff
@@ -61,44 +61,34 @@ groups:
       - ref: network.type
         requirement_level: recommended
 
-  - id: span.rpc.grpc
-    type: span
-    stability: experimental
-    extends: rpc
+  - id: rpc.grpc.attributes
```
lmolkova, line 64: `rpc.*flavor` spans capturing both client and server. If we want to preserve them as spans, we should break them down into two definitions. Otherwise they could stay as attribute groups.

trask (approved):

Issue #1873, opened 2025-02-05
Introduce an optional normalized total CPU utilization metric


rogercoll:
### Propose new conventions

Knowing what "cpu utilization" means isn’t trivial. There can be different interpretations of it and some users would want control over whether iowait or steal states count towards the usage or not. But there should be a simple, opinionated, and standardized metric that doesn’t need custom processors, or advanced knowledge of PromQL and cpu state philosophy to be useful for most use cases.

As part of the standardization, we need to align which CPU states count towards "cpu utilization". A lot [of](https://grafana.com/grafana/dashboards/9617-cpu/) [examples](https://stackoverflow.com/questions/66169803/cpu-usage-for-each-node-in-prometheus) recommend something like 1-idle. While there are [other](https://grafana.demo.do.prometheus.io/d/BEAYT9PWz/node-exporter-nodes?orgId=1&from=now-1h&to=now&timezone=utc&var-datasource=default&var-cluster=&var-instance=demo.do.prometheus.io:9100&refresh=30s&inspect=panel-2&inspectTab=query) https://github.com/prometheus/node_exporter/pull/2194 that recommend 1-(idle|iowait|steal). The system.cpu.total.norm.pct metric of the [Elastic System integration](https://www.elastic.co/guide/en/integrations/current/system.html#system-cpu) is defined as 1-(idle|iowait).

While it is possible to create a custom processor that calculates another derived metric, or using a complex PromQL query, standardizing the semantics for an overall cpu utilization would be a big win in terms of usability for such a core metric. We don’t want to open the floodgates on adding derived/convenience metrics for every possible use case but CPU utilization is such a fundamental metric that almost everyone needs, where the semantics aren’t obvious, and where complex queries are required at the moment.

Also, replicating these complex PromQL queries isn’t feasible for all backends. So from a perspective of being more vendor-agnostic, adding a standardized metric that is easily aggregatable makes sense, too.

Similar to [system.cpu.utilization](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemcpuutilization), this should be opt-in. In contrast to system.cpu.utilization, it should not have attributes for [cpu.mode](https://opentelemetry.io/docs/specs/semconv/attributes-registry/cpu/) or [system.cpu.logical_number](https://opentelemetry.io/docs/specs/semconv/attributes-registry/system/).

To introduce an optional normalized total CPU utilization metric, we propose the following solutions:

[...]

PR #1896, opened 2025-02-10
refactor: move system cpu-related metrics to cpu namespace

rogercoll:
Partially fixes https://github.com/open-telemetry/semantic-conventions/issues/1873 (renaming)

## Changes

Current `system.cpu.*` metrics do not reflect the CPU time/usage over the whole system but relative to a logical CPU. Each reported metric is specific to a logical CPU (system.cpu.logical_number). This PR proposes moving them to a new CPU namespace that will define metrics relative to a unique logical CPU. Additionally, the `system.cpu.logical_number` has been renamed to `cpu.logical_number` as it only applies to `cpu` namespace metrics.


braydonk (approved):
As it is this PR LGTM and matches the changes we discussed making as a group.

model/cpu/metrics.yaml
```diff
@@ -0,0 +1,35 @@
...
+        note: "Following modes SHOULD be used: `user`, `system`, `nice`, `idle`, `iowait`, `interrupt`, `steal`"
+      - ref: cpu.logical_number
+
+  - id: metric.cpu.frequency
+    type: metric
+    metric_name: cpu.frequency
+    stability: development
+    brief: "Operating frequency of the logical CPU in Hertz."
+    instrument: gauge
+    unit: "{Hz}"
```
lmolkova, line 33: nit: `Hz` is defined in [UCUM](https://unitsofmeasure.org/ucum#para-30)


```suggestion
    unit: "Hz"
```

lmolkova (approved):

model/cpu/metrics.yaml
rogercoll: fix and regenerated docs in https://github.com/open-telemetry/semantic-conventions/pull/1896/commits/ca0ab62e3e80bb9f33ec3f9ad367a054033c85c5

thank you!

PR #1897, opened 2025-02-10
Create `installation.id`, redefine `device.id`

### LINKED ISSUE #1874: `device.id` vs `installation.id` on Android and iOS
### Area(s)

area:device

### What's missing?

Take a look at the current definition of `device.id`:

https://github.com/open-telemetry/semantic-conventions/blob/b855a36cf4b8a3c5461ad16fa4ace0c615d11c2a/model/device/registry.yaml#L14-L21

The obligatory recommendation for Android ([_Firebase Installation ID_](https://firebase.google.com/docs/reference/android/com/google/firebase/installations/FirebaseInstallations#getId()) or _a globally unique UUID which is persisted across sessions in your application_) mostly mirrors the [recommendation for Google](https://developer.android.com/identity/user-data-ids#best-practices-android-identifiers)

> Use a Firebase installation ID (FID) or a privately stored GUID whenever possible for all other use cases, except for payment fraud prevention and telephony. For the vast majority of non-ads use cases, an FID or GUID should be sufficient.

The one for iOS suggests using [`identifierForVendor`](https://developer.apple.com/documentation/uikit/uidevice/identifierforvendor).

> The value of this property is the same for apps that come from the same vendor running on the same device. A different value is returned for apps on the same device that come from different vendors, and for apps on different devices regardless of vendor.

Now I have multiple remarks to these recommendations:

1. None of the two Android recommendations and the one iOS recommendation are really _device ids_, they are more like _installation ids_. For consumer apps the difference is probably insignificant, however, for enterprise apps on managed Android/iOS devices it is important. One could have many managed applications on the same device, so the expectation for `device.id`s would be that they are identical across all apps, which neither the Firebase Installation ID, nor the random generated UUID or the iOS `identifierForVendor` fulfill.

2. There is another ID, which fits my use case better on Android, called [`Settings.Secure.ANDROID_ID`](https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID), because I don't want to use Firebase, but I want an ID which survives uninstalls, and which is consistent across all my in-house apps.

   > **`ANDROID_ID`:**
   > On Android 8.0 (API level 26) and higher versions of the platform, a 64-bit number (expressed as a hexadecimal string), unique to each combination of app-signing key, user, and device. [...]

3. Although not recommended by Google for Android, one could still use actual device IDs, such as the IMEI or proprietary APIs provided by the vendor (e.g. [Zebra's API](https://techdocs.zebra.com/oeminfo/consume/#acquiredevicedata)), which are bound to the device for its whole lifetime. For enterprise-managed hardware there might be a legitimate use for such actual device ids, so these IDs would rather deserve to be the definition of `device.id`.

I would recommend these 3 changes:
1. Introduce a new attribute, called `installation.id`, and recommend it for most use-cases on Android and iOS.
2. Use a similar description for `installation.id`, that `device.id` currently has, but extend it with the possibility to use `Settings.Secure.ANDROID_ID`
3. Discourage the usage of `device.id` due to privacy reasons, but leave it open to custom implementations, such as IMEI or proprietary APIs.

bencehornak:
Fixes #1874

## Changes

Added `installation.id` (the first attribute in the `installation` namespace), which is _a unique identifier representing the installation of an application on a specific device_. Its definition is more-or-less identical to the old definition of `device.id`, which has always identified installations and not devices.

I have given a new definition to `device.id`, which now truly identifies the device. It has been marked as opt-in though, as its usage is generally discouraged due to privacy implications, and should only be enabled in special cases, when it does not violate the privacy of the user.

For more context see #1874.


joaopgrassi (commented):
I think this changes makes sense given the discussion in the issue. But I think it would be better if the changes related to `device` would be in a separate PR. Then you can include the changelog for it as a breaking change, since it is now Opt-In.

docs/attributes-registry/installation.md
```diff
@@ -0,0 +1,36 @@
...
+## Installation Attributes
+
+Describes installation attributes.
+
+| Attribute | Type | Description | Examples | Stability |
+|---|---|---|---|---|
+| <a id="installation-id" href="#installation-id">`installation.id`</a> | string | A unique identifier representing the installation of an application on a specific device [1] | `identifier_for_vendor:2ab2916d-a51f-4ac8-80ee-45ac31a28092`; `firebase:2ab2916d-a51f-4ac8-80ee-45ac31a28092`; `guid:2ab2916d-a51f-4ac8-80ee-45ac31a28092`; `app_set_id:2ab2916d-a51f-4ac8-80ee-45ac31a28092`; `android_id:2ab2916d-a51f-4ac8-80ee-45ac31a28092` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1] `installation.id`:** Its value MUST remain the same between launches of the same installation of an application.
+On a particular device it MIGHT be the same across different applications of the same vendor, however, it MUST be different across applications of different vendors.
```
LikeTheSalad, line 15: > On a particular device it MIGHT be the same across different applications of the same vendor

I'm guessing this comes from the "[app-set id](https://developer.android.com/identity/app-set-id#google-play-developer-scope)" approach mentioned in the issue, where an ID is the same across all apps within the same google developer account. I'm not sure if mentioning use cases where multiple apps are involved is needed or if it might cause confusion instead, I'm saying so because it sounds strange to me that the same "Installation ID" is reused across different apps installations 😅 though I guess that as long as the ID of a single app changes on a reinstall, then it's fine, regardless of whether the same ID is used elsewhere or not?

docs/attributes-registry/installation.md
```diff
@@ -0,0 +1,36 @@
...
+It SHOULD change if the application is uninstalled or if all applications of the vendor are uninstalled.
```
LikeTheSalad, line 16: If an application is uninstalled and then re-installed later, I'm guessing it should have different `installation.id` for the second install. So I'm curious why we're using "SHOULD" here?

LikeTheSalad (commented):
Thanks! I like the idea of using a name that better describes what we were doing with `device.id`. I've just added some questions on specific details, but I'm on board with the change.

docs/attributes-registry/installation.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Installation Attributes
+
+Describes installation attributes.
+
+| Attribute | Type | Description | Examples | Stability |
+|---|---|---|---|---|
+| <a id="installation-id" href="#installation-id">`installation.id`</a> | string | A unique identifier representing the installation of an application on a specific device [1] | `identifier_for_vendor:2ab2916d-a51f-4ac8-80ee-45ac31a28092`; `firebase:2ab2916d-a51f-4ac8-80ee-45ac31a28092`; `guid:2ab2916d-a51f-4ac8-80ee-45ac31a28092`; `app_set_id:2ab2916d-a51f-4ac8-80ee-45ac31a28092`; `android_id:2ab2916d-a51f-4ac8-80ee-45ac31a28092` | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
+
+**[1] `installation.id`:** Its value MUST remain the same between launches of the same installation of an application.
```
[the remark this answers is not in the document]
bencehornak: Would you be fine with 'SHOULD'?

docs/attributes-registry/installation.md
bencehornak: Right, the identical app id for a given vendor is a feature provided by [App set ID](https://developer.android.com/identity/app-set-id) and [`Settings.Secure.ANDROID_ID`](https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID) on Android, and the [`identifierForVendor`](https://developer.apple.com/documentation/uikit/uidevice/identifierforvendor) on iOS, see [this comparison](https://github.com/open-telemetry/semantic-conventions/issues/1874#issuecomment-2638963799).

bencehornak: I am not hundred percent satisfied with my own definitions I think because of the same reason you also found this line confusing. Let's try to approach it from this angle: what should _installation_ exactly mean?

* Option 1: The installation of **an application** on a specific device.
* Option 2: The installation of **an application OR an application group** on a specific device. An _application group_ is one or more applications from the same vendor.

I chose the 1st one when drafting the PR, however, thinking more carefully about it I'd go with option 2. The problem with option 1 is that it contradicts the possibility of identical IDs across apps from the same vendor.

So if you like option 2, then we can rephrase this definition as follows:

> **[1] `installation.id`:** Its value MUST remain the same between launches of the same installation of an application.
> If multiple applications belong to the same _application group_ (i.e. they share the same vendor), they MIGHT have the same installation IDs, however, applications of different _application groups_ MUST have different installation IDs.
> It SHOULD change if the application is uninstalled or if all applications of the application group are uninstalled.

If we go with option 1, then I am with you, I think it is confusing to allow identical `installation.id`s for different applications. But that also implies that we mustn't allow [App set ID](https://developer.android.com/identity/app-set-id), [Settings.Secure.ANDROID_ID](https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID) and [identifierForVendor](https://developer.apple.com/documentation/uikit/uidevice/identifierforvendor) as `installation.id`.

docs/attributes-registry/installation.md
bencehornak: My interpretation of [`Settings.Secure.ANDROID_ID`](https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID) is that it survives uninstalls and re-installs:

> On Android 8.0 (API level 26) and higher versions of the platform, a 64-bit number (expressed as a hexadecimal string), unique to each combination of app-signing key, user, and device. Values of ANDROID_ID are scoped by signing key and user. The value may change if a factory reset is performed on the device or if an APK signing key changes.

So if we want to allow this to be used as `installation.id`, I think we cannot use "MUST" here.

docs/attributes-registry/installation.md
```diff
@@ -0,0 +1,36 @@
...
+On a particular device it MIGHT be the same across different applications of the same vendor, however, it MUST be different across applications of different vendors.
+It SHOULD change if the application is uninstalled or if all applications of the vendor are uninstalled.
+Hardware IDs (e.g. serial number, IMEI, MAC address) MUST NOT be used as the installation identifier.
+The `installation.id` SHOULD be prefixed with the type of the identifier and a colon separator (e.g. `firebase:`).
```
bencehornak, line 18: I do want to get some opinions about this prefix. My reasoning was that if we allow heterogeneous use for `device.id` and `installation.id`, it's kind of important to now, what the installation ID was derived from, because the different IDs have different attributes. If I, as the consumer of telemetry data, only see `installation.id=a82cdd25-cce3-46cf-9a7d-cbdf53d477b1`, I don't know, if it is an app-set-id, a random GUID or something else.

However, it is arguably an implementation detail, which should not leak through the abstraction of OpenTelemetry.

Have you come across any similar situation with other semconvs? I'm fine removing the prefix clauses, but I'd really appreciate a second opinion 😄

bencehornak:
> I think this changes makes sense given the discussion in the issue. But I think it would be better if the changes related to `device` would be in a separate PR. Then you can include the changelog for it as a breaking change, since it is now Opt-In.

@joaopgrassi Thanks for your feedback! My reasoning for adding two changes to a single PR was that I see the two changes as a single transaction. If, say, the `device.id` changes are merged first, then the SDK users won't have any recommended fields to identify different installations until the 2nd PR with the `installation.id` changes are merged. If we do it other way around, for a short period of time there will be significant overlaps between the definitions of `device.id` and `installation.id`, so that would also not be ideal.

So right now I don't know, how to put the two changes in topological order given their circular dependency, but if you help me figure this out, I'm happy to split this PR into two.

model/device/registry.yaml
```diff
@@ -3,23 +3,32 @@ groups:
...
-          On iOS (Swift or Objective-C), this value MUST be equal to the [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
-          On Android (Java or Kotlin), this value MUST be equal to the Firebase Installation ID or a globally unique
-          UUID which is persisted across sessions in your application. More information can be found [here](https://developer.android.com/training/articles/user-data-ids)
-          on best practices and exact implementation details.
-          Caution should be taken when storing personal data or anything which can identify a user. GDPR and
-          data protection laws may apply, ensure you do your own due diligence.
-        examples: ['2ab2916d-a51f-4ac8-80ee-45ac31a28092']
+        note: |
+          **The `device.id` SHOULD NOT be used in most user-facing applications due to privacy regulations.
+          Consequently, instrumentations MUST provide it as an OPT-IN feature.**
```
lmolkova, line 15:
```suggestion
          Consequently, instrumentations that provide it MUST provide it as an opt-in feature.**
```

there is no formal definition of OPT-IN, so let's use lower case

model/device/registry.yaml
```diff
@@ -3,23 +3,32 @@ groups:
...
+
+          Its value SHOULD be identical for all apps on a device and it SHOULD NOT change if an app is uninstalled and re-installed.
+          However, it MIGHT be resettable by the user for all apps on a device.
```
lmolkova, line 18:
```suggestion
          However, it might be resettable by the user for all apps on a device.
```

again, MIGHT is not normative language. Feel fee to highlight it in **some** *other* way

model/installation/registry.yaml
```diff
@@ -0,0 +1,43 @@
...
+    display_name: Installation Attributes
+    brief: >
+      Describes installation attributes.
+    stability: development
+    attributes:
+      - id: installation.id
+        type: string
+        stability: development
+        brief: >
+          A unique identifier representing the installation of an application on a specific device
```
lmolkova, line 13: I think `installation` is too broad. Let's consider something like `device.app.installation.id`, `mobile.app.installation.id`, or broader `app.installation.id` or `software.installation.id` (so it can cover desktop OS apps too)

In the current form it could mean anything - like an installation artwork identifier, building/construction or military installation.

We should not have attributes that covers different unrelated fields.

model/installation/registry.yaml
```diff
@@ -0,0 +1,43 @@
...
+        note: |
+          Its value MUST remain the same between launches of the same installation of an application.
+          On a particular device it MIGHT be the same across different applications of the same vendor, however, it MUST be different across applications of different vendors.
```
lmolkova, line 16:
```suggestion
          On a particular device it could be the same across different applications of the same vendor, however, it MUST be different across applications of different vendors.
```

model/installation/registry.yaml
```diff
@@ -0,0 +1,43 @@
...
+      Describes installation attributes.
+    stability: development
+    attributes:
+      - id: installation.id
+        type: string
+        stability: development
+        brief: >
+          A unique identifier representing the installation of an application on a specific device
+        note: |
+          Its value MUST remain the same between launches of the same installation of an application.
```
lmolkova, line 15: > Its value MUST remain the same between launches of the same installation of an application.

Is installation id generation is something that instrumentation is responsible for? If not, we should avoid normative language or use SHOULD

model/installation/registry.yaml
lmolkova: > it MUST be different across applications of different vendors.

is this something in instrumentation control? I don't think it's possible to guarantee global uniqueness

model/installation/registry.yaml
```diff
@@ -0,0 +1,43 @@
...
+          On a particular device it MIGHT be the same across different applications of the same vendor, however, it MUST be different across applications of different vendors.
+          It SHOULD change if the application is uninstalled or if all applications of the vendor are uninstalled.
+          Hardware IDs (e.g. serial number, IMEI, MAC address) MUST NOT be used as the installation identifier.
+          The `installation.id` SHOULD be prefixed with the type of the identifier and a colon separator (e.g. `firebase:`).
```
lmolkova, line 19:
```suggestion
          The `installation.id` value SHOULD be prefixed with the type of the identifier and a colon separator (e.g. `firebase:`).
```

model/installation/registry.yaml
```diff
@@ -0,0 +1,43 @@
...
+          But it MAY be equal to other identifiers such as:
+
+          - [App set ID](https://developer.android.com/identity/app-set-id).
+          - [`Settings.getString(Settings.Secure.ANDROID_ID)`](https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID).
+
+          More information about Android identifier best practices can be found [here](https://developer.android.com/training/articles/user-data-ids).
+
+          > [!WARNING]
+          >
+          > Caution should be taken when storing personal data or anything which can identify a user. GDPR and laws may apply, ensure you do your own due diligence.
```
lmolkova, line 37: does this field contain PII or linkable PII? If not, let's remove the warning, otherwise, let's explicit say that it contains PII.

model/installation/registry.yaml
```diff
@@ -0,0 +1,43 @@
...
+        note: |
+          Its value MUST remain the same between launches of the same installation of an application.
+          On a particular device it MIGHT be the same across different applications of the same vendor, however, it MUST be different across applications of different vendors.
+          It SHOULD change if the application is uninstalled or if all applications of the vendor are uninstalled.
+          Hardware IDs (e.g. serial number, IMEI, MAC address) MUST NOT be used as the installation identifier.
+          The `installation.id` SHOULD be prefixed with the type of the identifier and a colon separator (e.g. `firebase:`).
+
+          On iOS this value SHOULD be equal to the [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/identifierforvendor).
+
+          On Android this value SHOULD be equal to either:
```
lmolkova, line 23: would it be possible to have one list of possible id sources ordered by preference ?
It's not clear how/why someone would pick one firebase id, guid or one of those mentioned as MAY below

schema-next.yaml
```diff
@@ -2,6 +2,11 @@ file_format: 1.1.0
 schema_url: https://opentelemetry.io/schemas/next
 versions:
   next:
+    all:
+      changes:
+        - rename_attributes:
+            attribute_map:
+              device.id: installation.id
```
lmolkova, line 9: Is this correct? `device.id` is still there, and I assume could still be used. 

I think we should not define any transformation and just explain this change in the changelog

model/device/registry.yaml
```diff
@@ -3,23 +3,32 @@ groups:
...
+          Hardware IDs (e.g. vendor-specific serial number, IMEI or MAC address) MIGHT be used as values.
+          The `device.id` SHOULD be prefixed with the type of the identifier and a colon separator (e.g. `imei:`).
+
+          More information about Android identifier best practices can be found [here](https://developer.android.com/training/articles/user-data-ids).
+
+          > [!WARNING]
+          >
+          > Caution should be taken when storing personal data or anything which can identify a user. GDPR and data protection laws may apply, ensure you do your own due diligence.
```
lmolkova, line 26: let's explicitly mention that this attribute contains PII like

```suggestion
          > This attribute contains sensitive (PII) information. Caution should be taken when storing personal data or anything which can identify a user. GDPR and data protection laws may apply, ensure you do your own due diligence.
```

model/device/registry.yaml
```diff
@@ -3,23 +3,32 @@ groups:
...
+          > See [`installation.id`](../attributes-registry/installation.md#installation-id) as a more privacy-preserving alternative.
```
lmolkova, line 27: let's use absolute path from the repo root - you never know where this document will be rendered and if relative urls will still work

joaopgrassi:
> > I think this changes makes sense given the discussion in the issue. But I think it would be better if the changes related to `device` would be in a separate PR. Then you can include the changelog for it as a breaking change, since it is now Opt-In.
> 
> @joaopgrassi Thanks for your feedback! My reasoning for adding two changes to a single PR was that I see the two changes as a single transaction. If, say, the `device.id` changes are merged first, then the SDK users won't have any recommended fields to identify different installations until the 2nd PR with the `installation.id` changes are merged. If we do it other way around, for a short period of time there will be significant overlaps between the definitions of `device.id` and `installation.id`, so that would also not be ideal.
> 
> So right now I don't know, how to put the two changes in topological order given their circular dependency, but if you help me figure this out, I'm happy to split this PR into two.

Yes, I understand. We have been asked by users though that such bundling breaking changes with other changes such as this makes it harder for them to track breaking changes, which is understandable. We have been trying recently to better identify/label breaking changes PR to help in such cases. 

I think we could still make it work to merge both of them separately. You will need approvals in both anyway and once both are good we just merge them together in the same release. We can for example, add the PRs into the same milestone.

docs/attributes-registry/installation.md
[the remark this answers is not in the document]
LikeTheSalad: Yes, at least in the case where a locally generated GUID is used, there's no guarantee that it will stay the same during the whole installation as users may clear the data that holds it.

docs/attributes-registry/installation.md
LikeTheSalad: > it contradicts the possibility of identical IDs across apps from the same vendor

I see what you mean, however, I don't see a bullet-proof way of preventing this from ever happening (more than one app using the same installation ID, even if it doesn't come from the same vendor) so I don't think it matters much if we mention it as a possibility or not, and considering that there will be other attributes involved to tell telemetry sources apart, such as `service.name` for example, I'm not sure it'll be a problem either. So I'd focus on what a single app needs, such as making a "best effort" on keeping the same `installation.id` value across a single installation (which to me also sounds like it must change in case the app is uninstalled and reinstalled later, though it's all semantics at this point this is not a blocker for me), regardless of whatever might happen on other apps.

So to summarise, I'd just remove the whole line 15 and also any mention of "vendor" or group-related IDs.

docs/attributes-registry/installation.md
LikeTheSalad: Got it, thanks. I think is fine to leave SHOULD here.

model/installation/registry.yaml
LikeTheSalad: Related [comment](https://github.com/open-telemetry/semantic-conventions/pull/1897#discussion_r1950578435).

docs/attributes-registry/installation.md
bencehornak: Alright, happy to simplify, removed the line in question.

docs/attributes-registry/installation.md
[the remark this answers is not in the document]
bencehornak: Changed to 'SHOULD'

model/installation/registry.yaml
bencehornak: In my opinion the implementation could be provided by the platform or instrumentation. If the platform can provide something fulfilling these criteria/recommendations, the instrumentation libraries might rely on those. E.g. on Android and iOS there are a few options to choose from, however, for desktop JVM/native apps the situation is far less obvious. Ultimately, I would argue that the application developer should have the control to choose one or another.

Since it is partially provided by the platform, I have relaxed it to SHOULD, as you requested.

model/installation/registry.yaml
bencehornak: I have removed this line as requested by @LikeTheSalad in the referenced thread.

model/installation/registry.yaml
bencehornak: Good point, this field should not qualify as PII if implemented correctly in contrast to `device.id`, which definitely does. I have removed the notice

schema-next.yaml
bencehornak: Alright, I can see that it is confusing for some, will revert the changes to `schema-next.yaml`.

model/installation/registry.yaml
bencehornak: Regarding the order I would try to stay reasonably close to [Google's recommendations](https://developer.android.com/identity/user-data-ids#best-practices-android-identifiers) here, as they have more authority to prescribe an implementation for the Android ecosystem:

> Use a Firebase installation ID (FID) or a privately stored GUID whenever possible for all other use cases, except for payment fraud prevention and telephony. For the vast majority of non-ads use cases, an FID or GUID should be sufficient.

Since Google didn't clearly lay out a precedence between the two, I tried to do the same - just listing the two options in the same order. It didn't mention the two other options, but since they are totally legit IDs, I thought, they could be added to the notes too.

Personally, I find the Firebase option a bad universal choice, since Firebase is not part of the Android platform, but it is an opinionated library closely tied to Google Services, which are not necessarily available on all devices (in contrast to `Settings.getString(Settings.Secure.ANDROID_ID)`). I think the OpenTelemetry Android project should refrain from adding Firebase as a dependency, as it might be undesirable for some users. The GUID choice is a safe one, since it's pretty universal, however, it might not satisfy all users (e.g. developers, who would like to correlate IDs of apps on the same device).

My recommendation is to leave the choice open on the semconv level and to leave the opinionated decision to the instrumentation maintainers. Additionally, instrumentations may give the instrumentation users the freedom to pick an ID other than the default one to fulfill various developer interests.

bencehornak: As to why somebody would pick a different ID you can refer to [this comparison](https://github.com/open-telemetry/semantic-conventions/issues/1874#issuecomment-2638963799). As you can see the various implementations have slightly different features.

My gut feeling is that 99% of applications developers don't care about what ID their instrumentations provide, yet 1% might have an opinion about it. That 1% can find more information on the internet about the pros and cons of various implementations (e.g. this Medium post: [Decoding Unique Identifiers in Android: Risks, Impacts, and Solutions](https://medium.com/digdroid/decoding-unique-identifiers-in-android-risks-impacts-and-solutions-gaid-fid-guid-app-set-id-1076b56d0dd4)), I think a detailed comparison is out-of-scope for the semconv docs.

bencehornak: @lmolkova So as you can see it's a lot of context, which is implied here, I can see that it might overwhelm instrumentation maintainers and application developers. To cater to the 99% who has no strong preference about one implementation or another, what do you think about the following suggestions:

1. I rephrase the statement like this to deter most people from choosing something from the 'MAY' section:

   ```diff
              - [Firebase Installation ID](https://firebase.google.com/docs/projects/manage-installations).
              - A globally unique UUID which is persisted across sessions in your application.
    
   -          But it MAY be equal to other identifiers such as:
   +          But it MAY be equal to other identifiers as well, if the above implementations are not preferred for some reason, such as:
    
              - [App set ID](https://developer.android.com/identity/app-set-id).
              - [`Settings.getString(Settings.Secure.ANDROID_ID)`](https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID).
   ```

2. We leave the choice between Firebase and GUID to the instrumentation maintainers (I strongly suggest GUID for the reference implementation)
3. Application developers may be given a way to override the instrumentation's choice if they want to

model/installation/registry.yaml
bencehornak: @lmolkova Excellent point, I certainly didnt't think about that from this perspective. Thanks for your suggestions too!

I'd argue that the scope of this field is mainly client applications (be it mobile, IoT or desktop) used by end users, so I find `app.installation.id` the most appropriate, because _app_ usually refers to end-user-apps (see [GeeksforGeeks](https://www.geeksforgeeks.org/difference-between-program-and-application/)). In my understanding _apps_ can be used to refer to _desktop programs_ too, so I'd encourage the use of the proposed `app.` namespace on desktop too.

`software.installation.id` is definitely another option, at this point it's quite subjective, what one considers more appropriate for describing client-side programs: _software_ or _app_. My personal preference is the namespace _app_.

I don't see the point in restricting this attribute to mobile apps only, so I'd advise against `mobile.app.installation.id`. Moreover, I'd vote against `device.app.installation.id` too, as I don't see a reason to nest `app` under `device`. It's true, that they are both related to client apps, but I think they should be on the same level (similarly to `service` vs `os`). If further grouping is desired, one option would be nesting both of them under a common group, like `client.app.*` and `client.device.*`, although I wasn't planning to go that far with this PR.

bencehornak: If there is no objection to it, I'd go forward with `app.installation.id`.

bencehornak:
@joaopgrassi alright, I'll restrict this PR to the new definition of `device.id`, and will add the new attribute to a separate PR.

docs/attributes-registry/installation.md
LikeTheSalad: Having a prefix, and now knowing the context around why is there, might make me think that there are other `installation.id`s with other prefixes belonging to the same installation. I guess what I'm trying to say is that generally speaking, I would only make a distinction if I were aware of multiple values that could be present at the same time.

There might be other use cases that I'm not aware of where making this distinction is helpful, so I'm not against having it, although if you have and could share an example of a use case where an end-user might need to know the origin of the installation.id value it'd help to better understand the reason why it should be stated in the conventions.

bencehornak: Sure, some context wouldn't hurt 😄

I'm working on a [Kotlin Multiplatform client app](https://kotlinlang.org/docs/multiplatform-intro.html), which I compile for Android and the desktop (and iOS in the future). For the various targets most of the resource attributes (e.g. `service.name` and `service.namespace`) will be identical, yet others will be platform-specific (e.g. `os.name` and of course `(app.)instalation.id`). For each target I'm planning to use a platform-specific implementation for `(app).installation.id`, which all have various characteristics (see the [comparison](https://github.com/open-telemetry/semantic-conventions/issues/1874#issuecomment-2638963799)). If we add prefixes to the IDs, it suggests to the reader of the telemetry data that the IDs behave differently (i.e. `firebase:` IDs are consistent across all apps from the vendor on the same device in the same account, however `guid:` IDs are not).

On the other hand, one could say that this information should not leak through the abstraction of OpenTelemetry, and the prefix makes the definition of this field unnecessarily complicated.

Having the prefix is not a hard requirement for me, I can leave without it too. So what do you think, should this extra information be visible through OpenTelemetry?

LikeTheSalad: I see, thank you for providing this context!

So if I understood correctly, the underlying use case seems to be related to telling users whether the same id is used across other apps or not. I'm not sure if that's something that the conventions should mention because I don't think it's something that developers can fully ensure to happen consistently. For example, the [app-set](https://developer.android.com/identity/app-set-id#app-scope) ID seems to be the same across all apps from a vendor only if the apps are installed via the Google Play Store, and the device where the app is installed has Google Play Services installed, and if some of those conditions aren't met, the ID will have the prefix from this google's tool although it wouldn't necessary always mean it's shared.

On top of that, having a prefix I think would also mean that we'd have to define what are the valid values (at least if we want to ensure that this info is useful in a vendor-agnostic way), otherwise maybe some people might decide to use `google` and others `appset` and so on, when they're all using the same [google play services appset tool](https://developer.android.com/identity/app-set-id#add-sdk) to get their ids.

So I'd say that it's better to keep it simple, at least for now, and not include prefixes in the convention.

bencehornak:
Sorry folks for the delay, now I'm back from holiday!

@joaopgrassi I have extracted the changes to `device.id` to #1951 as you requested, this PR only contains the addition of the `app.installation.id` attribute.

@lmolkova, @LikeTheSalad, @bidetofevil I believe that I have addressed all of your remarks, I would appreciate your feedback on the new additions.

LikeTheSalad (approved):
Thank you! 🙏

bencehornak:
@joaopgrassi @lmolkova, the two PRs (this and #1951) are ready for your review, could you take a look, if you get a chance?

trask:
it would be great if we can get approvals from the iOS and browser perspectives as well (if this attribute is applicable there)

MSNev (approved):

docs/attributes-registry/app.md
```diff
@@ -0,0 +1,32 @@
...
+It SHOULD change if the application is uninstalled or if all applications of the vendor are uninstalled.
+Additionally, users might be able to reset this value (e.g. by clearing application data).
+If an app is installed multiple times on the same device (e.g. in different accounts on Android), each `app.installation.id` SHOULD have a different value.
+Hardware IDs (e.g. serial number, IMEI, MAC address) MUST NOT be used as the `app.installation.id`.
+
+On iOS this value SHOULD be equal to the [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/identifierforvendor).
+
+On Android this value SHOULD be equal to either:
+
+- [Firebase Installation ID](https://firebase.google.com/docs/projects/manage-installations).
```
bidetofevil, line 24: Any reason why Firebase is called out specifically? Typically, each SDK or library who want to uniquely identify app instances will have their own unique ID. Having the same value for different SDKs for the same app might cause confusion in the recorded telemetry as to their provenance.

docs/attributes-registry/app.md
```diff
@@ -0,0 +1,32 @@
...
+
+## App Attributes
+
+Describes attributes related to client-side applications (e.g. web apps or mobile apps).
+
+| Attribute | Type | Description | Examples | Stability |
+|---|---|---|---|---|
+| <a id="app-installation-id" href="#app-installation-id">`app.installation.id`</a> | string | A unique identifier representing the installation of an application on a specific device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[1] `app.installation.id`:** Its value SHOULD remain the same between launches of the same installation of an application.
```
bidetofevil, line 14: This should also persist between normal app upgrades so that version differences on the same install can map back to the same ID

docs/attributes-registry/app.md
```diff
@@ -0,0 +1,32 @@
...
+
+On Android this value SHOULD be equal to either:
+
+- [Firebase Installation ID](https://firebase.google.com/docs/projects/manage-installations).
+- A globally unique UUID which is persisted across sessions in your application.
+
+But it might be equal to other identifiers as well, if the above implementations are not preferred for some reason, such as:
+
+- [App set ID](https://developer.android.com/identity/app-set-id).
+- [`Settings.getString(Settings.Secure.ANDROID_ID)`](https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID).
```
bidetofevil, line 30: I would actually avoid using this because the app or SDK using it will not be able to control this or make sure it's constant across upgrade versions - it depends on the app signing key as a component in its calculation so it will change if the signing key is rotated.

docs/attributes-registry/app.md
```diff
@@ -0,0 +1,32 @@
...
+
+# App
+
+## App Attributes
+
+Describes attributes related to client-side applications (e.g. web apps or mobile apps).
+
+| Attribute | Type | Description | Examples | Stability |
+|---|---|---|---|---|
+| <a id="app-installation-id" href="#app-installation-id">`app.installation.id`</a> | string | A unique identifier representing the installation of an application on a specific device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | ![Development](https://img.shields.io/badge/-development-blue) |
```
bidetofevil, line 12: Given that there's no easy way to share or manage this value across all the different modules (first and third parties) that comprise an app, the same app instance might emit different values for this depending on instrumentation. How do we want to account for this in the definition - just that a particular app instance can emit different values, or something more formal?

docs/attributes-registry/app.md
bencehornak: All of them have pros and cons (see my [comparison](https://github.com/open-telemetry/semantic-conventions/issues/1874#issuecomment-2638963799)), that's why I would leave this decision ultimately the SDK user. Is that fine for you?

docs/attributes-registry/app.md
bencehornak: I wanted to stay as close to the [official recommendation](https://developer.android.com/identity/user-data-ids#instance-ids-guids) as possible, that's why I called out Firebase and the GUID.

bencehornak: Regarding the ID's provenance, I had this idea to prefix the value with exactly that (e.g. `firebase:` or `guid:`), but we later discarded it, see [this thread](https://github.com/open-telemetry/semantic-conventions/pull/1897/files/606df0e978fbf29ad36dae83fd7a7c3801e22a38#r1949477663). Feel free to reopen, if you would like to add further points.

docs/attributes-registry/app.md
bencehornak: Not sure if I get this, you mean that if the app uses multiple tracking libraries (e.g. Otel Android SDK, Embrace SDK, Crashlytics, Instabug, Sentry, Posthog, ...) then each of these SDKs might have a different ID? Or are you referring to the possibility that an app has multiple instances of an Otel-compatible SDK in it, thus it is described by multiple Otel Resources (with possibly differing `app.installation.id` attributes)?

docs/attributes-registry/app.md
bencehornak: Will change the sentence to: `Its value SHOULD persist across launches of the same application installation, including through application upgrades.`

docs/attributes-registry/app.md
bidetofevil: Firebase is not universally used, and using it ties the instrumentation to it, which means tying itself to Firebase. I think it's not an unreasonable recommendation if you use Firebase, but the recommendation in the docs probably makes the most sense to me:

```
val uniqueID = UUID.randomUUID().toString()
```

bidetofevil: re: prefix - Ah, I see you were thinking along those lines too, and just using `app` signals the intention to have skip over that implementation detail. Yeah, makes ense.

docs/attributes-registry/app.md
bidetofevil: > are you referring to the possibility that an app has multiple instances of an Otel-compatible SDK in it

Yeah. This may be me being overly pedantic, but my thinking is that this will likely be set as a `resource` attribute before the OTel SDK instance starts, so it's up to whoever starts it to populate it with a value.

If your app has one OTel SDK instance running, no problem. There is only one. But if there are *multiple* OTel SDK instances running at the same time (which is theoretically possible depending on language/impl), we could have different, unrelated code (even third party) setting its own `app.installation.id` at startup.

A contrived example is if some ad SDK uses OTel under the hood, and it'll start its own specialized OTel SDK instance. This will live along side the instance the app uses for its custom telemetry and whatever other instrumentation you've configured it to use.

Perhaps a warning that this is tied to each OTel SDK instance that's running, and mentioning the multi-instance case as a example of there being more than one, is sufficient?

bencehornak: I see, I think I get what you mean, the possibility to have multiple Otel SDKs in an app is tricky. Ultimately, we need to decide, whether the proposed attribute should identify the 1. _app installation_ or 2. the _OTel SDK instance_.

Based on the suggested `app.installation.id` name I would go with the first choice, in which case it would be the user's responsibility to ensure consistency across all OTel SDKs within an app. Practically speaking, I think that should be possible with most (all?) SDKs, at least the OTel SDKs I've worked with allow setting/overriding any resource attributes, including the Android SDK, but that you surely know better than me 😄

If it is intended to have different IDs for each OTel SDK within the app, I suggest to draft a new PR with a new attribute name and optionally reject this one.

I think both IDs make sense in the end, 2. is easier to implement, however 1. is arguably more useful, as it allows correlating the telemetry stemming from each of those SDKs. Perhaps we can agree on the following phrasing:

> If multiple OpenTelemetry SDKs are used within the same application, they SHOULD use the same value for `app.installation.id`

This allows deviation from this, but states that the harmonized `app.installation.id` is the expected behavior. What do you think?

docs/attributes-registry/app.md
bencehornak: > Firebase is not universally used, and using it ties the instrumentation to it, which means tying itself to Firebase. I think it's not an unreasonable recommendation if you use Firebase, but the recommendation in the docs probably makes the most sense to me

Couldn't agree more, I'd also not choose to use Firebase in the OTel Android SDK as the primary implementation, because that would pull in a huge dependency, which is unwanted in many cases. I could imagine it being an alternative implementation like this:

```kotlin
OpenTelemetryRum.builder(this, config)
  .appInstallationIdProvider(UuidInstallationIdProvider()) // default behavior
  // or
  .appInstallationIdProvider(FirebaseInstallationIdProvider())
  // ...
```

where `FirebaseInstallationIdProvider` could be defined in a separate Firebase-specific Gradle module.

I would resolve this thread, unless you would like to change something in the wording.

bidetofevil: In terms of wording, I think saying it SHOULD be unique, and give some examples of how to get one, e.g. the Fireebase impl, random UUID, etc.. When I first read this, the fact it said "SHOULD be equal to either" seems like a recommendation of sorts. If you can remove that implication, that would be great. If not, it's not a big deal either and wouldn't be blocking to me.

docs/attributes-registry/app.md
bidetofevil: That sounds good. I agree that semantically, it's closer to 2, but making it 1 makes more sense in practice. That additional line you added is sufficient for me to close/acknowledge that loophole.

And agreed about it not being best practice to have multiple SDK instances for a single app process. If you're using an Agent-like package like `opentelemetry-android` or the Embrace Android SDK that offers a gateway to the OTel API, the OTel instance should come from that. Anything else is effectively instrumentation, and they shouldn't instantiate their own SDK instance. And if they do, they better let the app set things like resource attributes :-)

bryce-b (approved):

docs/attributes-registry/app.md
bencehornak: Added the new sentence.

docs/attributes-registry/app.md
bencehornak: Added the new sentence.

docs/attributes-registry/app.md
bencehornak: @bidetofevil I have changed the sentence to:

> For Android examples of `app.installation.id` implementations include:

Does it sounds better?

joaopgrassi (approved):
I don't immediately see a problem with introducing this new top-level area `app`. Let's get more eyes and approvals from other maintainers here.

docs/attributes-registry/app.md
bidetofevil: Works for me!

bidetofevil (approved):

model/app/resources.yaml
```diff
@@ -0,0 +1,9 @@
+groups:
+  - id: resource.app
+    type: resource
+    stability: development
+    name: app
+    brief: >
+      A client-side application.
```
lmolkova, line 7: nit:  maybe

```suggestion
      An app used directly by end users — like mobile, web, or desktop.
```

lmolkova (approved):

PR #1910, opened 2025-02-15
Add browser page view event

martinkuba:
## Changes

This adds a new event for capturing the page view event from browser/web applications.

Related to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2372 


docs/browser/events.md
```diff
@@ -0,0 +1,51 @@
...
+| `url` | string | Full HTTP request URL in the form scheme://host[:port]/path?query[#fragment]. Usually the fragment is not transmitted over HTTP, but if it is known, it should be included nevertheless. | `https://en.wikipedia.org/wiki/Main_Page`; `https://en.wikipedia.org/wiki/Main_Page#foo` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+`type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `0` | Initial page load within the browser which will generally also precede a PageNavigationTiming event. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `1` | This is for Single Page Applications (SPA) where the framework provides the ability to perform client side only page "navigation", the exact definition of what a virtual page change is determined by the SPA and the framework it is using. | ![Development](https://img.shields.io/badge/-development-blue) |
```
trask, line 44: do these values (`0` and `1`) come from the browser / browser spec, or did we create them?

docs/browser/events.md
```diff
@@ -0,0 +1,51 @@
...
+
+This event represents a browser page load/view.
+
+The primary use case is to capture metrics about the number of views of a page. Page view can be a hard page load in a browser as well as virtual navigation in a SPA (single-page application). The event should be collected as soon as possible without waiting for potentially delayed data such as the Navigation timing browser API.
+
+**Body fields:**
+
+| Body Field  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| `change_state` | string | Type of state change used for the virtual page navigation | `pushState`; `replaceState` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
trask, line 33: just from my reading of the description this could be better?

```suggestion
| `state_change` | string | Type of state change used for the virtual page navigation | `pushState`; `replaceState` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```

model/browser/events.yaml
```diff
@@ -0,0 +1,67 @@
...
+    body:
+      id: browser.page_view
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: referrer
+          type: string
+          stability: development
+          brief: Referring Page URI (document.referrer) whenever available.
```
david-luna: would be nice to have a link here. https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer

model/browser/events.yaml
```diff
@@ -0,0 +1,67 @@
...
+                change is determined by the SPA and the framework it is using.
+          stability: development
+          requirement_level: required
+          brief: Type of navigation
+          examples: [0, 1]
+        - id: title
+          type: string
+          stability: development
+          requirement_level: recommended
+          brief: Page title DOM property
```
david-luna: most applications use a generic text in `document.title` which may map the the routes defined in it. But in some cases they are customized for the user so we can be in situations of high cardinality and also there might be some PII issues.

I quite new here so I don't know if that needs to be covered here. At least instrumentations sending sch event should offer the option to not send this field or to redact it.

model/browser/events.yaml
```diff
@@ -0,0 +1,67 @@
...
+          examples: ["Home", "Checkout"]
+        - id: url
+          type: string
+          stability: development
+          requirement_level: required
+          brief: >
+            Full HTTP request URL in the form scheme://host[:port]/path?query[#fragment].
```
david-luna: query string is not mentioned here. If the intention is not to have it I think we should be explicit about it.

model/browser/events.yaml
```diff
@@ -0,0 +1,67 @@
...
+              stability: development
+              brief: >
+                This is for Single Page Applications (SPA) where the framework
+                provides the ability to perform client side only page
+                "navigation", the exact definition of what a virtual page
+                change is determined by the SPA and the framework it is using.
+          stability: development
+          requirement_level: required
+          brief: Type of navigation
+          examples: [0, 1]
```
thompson-tomo: Feel that these should be attributes

model/browser/events.yaml
```diff
@@ -0,0 +1,67 @@
...
+            Full HTTP request URL in the form scheme://host[:port]/path?query[#fragment].
+            Usually the fragment is not transmitted over HTTP, but if it is
+            known, it should be included nevertheless.
+          examples: ["https://en.wikipedia.org/wiki/Main_Page", "https://en.wikipedia.org/wiki/Main_Page#foo"]
+        - id: change_state
+          type: string
+          stability: development
+          requirement_level: recommended
+          brief: Type of state change used for the virtual page navigation
+          examples: ["pushState", "replaceState"]
```
thompson-tomo: Feel that these should be attributes

model/browser/events.yaml
thompson-tomo: I am in favour of removing as I fail to see what benefit it brings.

thompson-tomo (changes requested):

danielgblanco:
Thinking about the notion of combining hard and soft navigation on `browser.page_view`. As specified here, the intention is not to measure any timings, and instead use `browser.navigation_timing`, and I agree that's necessary if we want to represent purely counts of both types of navigations, otherwise there would be too many condition measurements depending on one or the other.

However, this brings up a follow up question. Do we expect to follow the same pattern for soft navigations? i.e. having a separate event to report timings for soft navigations (i.e. the time from interaction to render/stabilising). I think that'd make sense to me.

model/browser/events.yaml
```diff
@@ -0,0 +1,67 @@
...
+          examples: ["Home", "Checkout"]
+        - id: url
+          type: string
+          stability: development
+          requirement_level: required
+          brief: >
+            Full HTTP request URL in the form scheme://host[:port]/path?query[#fragment].
+            Usually the fragment is not transmitted over HTTP, but if it is
+            known, it should be included nevertheless.
+          examples: ["https://en.wikipedia.org/wiki/Main_Page", "https://en.wikipedia.org/wiki/Main_Page#foo"]
```
danielgblanco: Should this be a reference to https://github.com/open-telemetry/semantic-conventions/blob/main/model/url/registry.yaml#L39 ?

```suggestion
      - ref: url.full
        requirement_level: required
```

model/browser/events.yaml
```diff
@@ -52,3 +52,70 @@ groups:
...
+                Initial page load within the browser which will generally also
+                precede a PageNavigationTiming event.
+            - id: virtual_page
+              value: 1
+              stability: development
+              brief: >
+                This is for Single Page Applications (SPA) where the framework
+                provides the ability to perform client side only page
+                "navigation", the exact definition of what a virtual page
+                change is determined by the SPA and the framework it is using.
```
thompson-tomo, line 96: Value should be meaningful rather than numbers.

thompson-tomo (changes requested):

docs/browser/events.md
martinkuba: We defined these, but on a second thought I am inclined to use human-readable values instead. I think it could either be `hard` / `soft` or `page_load` /  `route_change`.

model/browser/events.yaml
martinkuba: added

model/browser/events.yaml
martinkuba: query is included in the example form, or I must be misunderstanding

model/browser/events.yaml
martinkuba: Is it now possible to reference an existing attribute as a body field? Also, this event has two URL fields (url and referrer). I think in this case the `url.full` name is fine, as in the context of the event it still makes sense. But if we needed to use a different name (like with referrer), is it still possible to reference url.full (for format, description etc.) but use a different name?

thompson-tomo: No but it should be an attribute rather than a body property with attributes useful when the value appears in more than 1 measurement which is the case here. In relation to referrer url, that would be a different property as mentioned in #1430

docs/browser/events.md
tedsuo: Human readable values are a best practice for enums in smeconv, so let's go with that 👍

model/browser/events.yaml
```diff
@@ -52,3 +52,70 @@ groups:
...
+    name: browser.page_view
+    brief: >
+      This event represents a browser page load/view.
+    note: >
+      The primary use case is to capture metrics about the number of views of
+      a page. Page view can be a hard page load in a browser as well as virtual
+      navigation in a SPA (single-page application). The event should be
+      collected as soon as possible without waiting for potentially delayed
+      data such as the Navigation timing browser API.
+    body:
```
tedsuo: Should all of this be moved out of the body and instead be set as attributes? Or one complex attribute?

lmolkova: +1

But this is where the fun starts - we don't have means to describe complex attribute types in semconv.

It should not be a blocker to this PR though. 
What we do in GenAI (#2179) is defining JSON schema for the types (it's a more natural way, it's been really hard to use yaml-based type definitions). 
In fact, we write models in Python and generate JSON schema from them. 

You folks might  find it more natural to write your models in JS and put them in this repo as non-normative guidance + have formal JSON schema mentioned in the brief/notes. 

There will definitely be a time when we'd all want complex attribute types  expressed formally in yaml, validate-able and code-generatable.

martinkuba: I have changed body fields to attributes.

model/browser/events.yaml
martinkuba: I have switched to using attributes, which now includes the `url.full` attribute.

martinkuba:
I have made a few changes based on feedback:

- switched from using body fields to attributes
- all new attributes are in the browser.* namespace
- for the URL field, I am now using the existing `url.full` attribute
- renamed page_view.title to page.title , since that seems to be really a property of the page
- updated the type enum to use human readable values (page_load and soft_navigation - I felt these make most sense, but open to other ideas)

There are a couple of things that still need to be resolved

- per @david-luna’s comment, the page title attribute could contain PII values, which complicates the instrumentation. It is also not clear if it provides much value. I am open to removing it (at least for now). lmkwyt
- I think the change_state attribute needs clarification. Is it intended to capture the mechanism of how the URL was changed or whether the new URL was added to the history stack? There are other ways of changing the URL than using the history API, so if it complicates things, I am also open to removing this for now.

model/browser/registry.yaml
```diff
@@ -50,3 +50,38 @@ groups:
           This value is intended to be taken from the Navigator API
           `navigator.language`.
         examples: ["en", "en-US", "fr", "fr-FR"]
+      - id: browser.referrer
+        type: string
+        stability: development
+        brief: Referring Page URI ([document.referrer](https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer)) whenever available.
+        examples: ["https://en.wikipedia.org/wiki/Main_Page"]
```
thompson-tomo: Feel this should be browser.page.referrer_url as it is describing the referrer of the page rather than details of the browser.

model/browser/registry.yaml
```diff
@@ -50,3 +50,38 @@ groups:
...
+              value: soft_navigation
+              stability: development
+              brief: >
+                This is for Single Page Applications (SPA) where the framework
+                provides the ability to perform client side only page
+                "navigation", the exact definition of what a virtual page
+                change is determined by the SPA and the framework it is using.
+        stability: development
+        brief: Type of navigation
+        examples: ["page_load", "soft_navigation"]
```
thompson-tomo, line 77: Similar to the other comment, I feel browser.page.view_type or browser.document.view_type might be a better fit as it is describing the page being shown in the browser.

model/browser/registry.yaml
```diff
@@ -50,3 +50,38 @@ groups:
...
+      - id: browser.page_view.change_state
+        type: string
+        stability: development
+        brief: Type of state change used for the virtual page navigation
+        examples: ["pushState", "replaceState"]
```
thompson-tomo: Similar to the other comment, I feel browser.page.change_state or browser.document.change_state might be a better fit as it is describing the page being shown in the browser.

model/browser/events.yaml
```diff
@@ -52,3 +52,27 @@ groups:
           brief: A unique ID representing this particular metric instance.
           requirement_level: required
           examples: ["v3-1677874579383-6381583661209"]
+
+  - id: event.browser.page_view
```
thompson-tomo: Maybe event.browser.page.rendered/viewed or event.browser.document.rendered/viewed so that other page events can be grouped with this one.

model/browser/registry.yaml
david-luna: From the MDN document linked in the brief
> The Document.referrer property returns the [URI](https://www.w3.org/Addressing/#background) of the page that linked to this page.

So I agree that referrer is not a browser detail.

But I'm more inclined to add it to the `page_view` namespace. In my mental model a page could have many referrer so maybe is not a good attribute to describe it. On the other side  when you visit a page you do from a previous one (none is a possible previous) and there can't be multiple values so the value describes better the event.

thompson-tomo: See my other comments, page view seems an odd attribute namespace for me. If we don't want to use browser.page maybe browser.document would be a better fit especially given document links.

model/browser/registry.yaml
david-luna: similar here

model/browser/registry.yaml
david-luna: similar here

model/browser/registry.yaml
joaquin-diaz: I wouldn't go with `browser.document.view_type`, I think we should reserve that namespace for attributes that are coming from `window.document` so it's easier to understand where is that coming from. I agree that it makes sense for referrer to be `browser.document.referrer` 

It looks like we're dealing with two namespaces here: 
* page
* page_view

Maybe having a proper definition of what they are can help assigning the attributes. If I have to give it a go I would say:
* page_view: it represents the action of navigating to a certain URL, whether is a full load (cold start) or soft navigation (like in SPAs)
* page: it contains information of the current page that the user is seeing in the browser, such us URL, title, etc 

Having said that I think here we're trying to describe the navigation type, so I agree that it should be in the `page_view` namespace

model/browser/registry.yaml
martinkuba: This makes sense to me, made the change

david-luna: all right. I'm fine with it :)

david-luna (approved):

model/browser/registry.yaml
martinkuba: I agree with @joaquin-diaz, the intent is to describe the type of navigation. Page itself does not have a type, especially when the values are either "page_load" or "soft_navigation". Maybe rename to `browser.navigation.type`?

model/browser/registry.yaml
martinkuba: The way this attribute is currently described is to capture detail about the "virtual page navigation". It specifically applies to soft navigation only. So, perhaps we should rename to `browser.navigation.state_change`. I am also open to removing this for now, as I am not 100% convinced this is even useful.

model/browser/events.yaml
martinkuba: I am open to making this change - I am thinking `browser.page.view` to more align with events like `session.start` or `app.screen.click` (noun rather than verb). With that said, I would at the same time rename attributes `browser.page_view.type` to `browser.navigation.type` and `browser.page_view.state_change` to `browser.navigation.state_change`.  This is because the term `page_view` was really only relevant to the event previously. WDYT?

model/browser/registry.yaml
thompson-tomo: I agree the page doesn't have a type hence suggestion was view_type but I do like your suggestion of browser.navigation.type as it is clear what it is describing.

model/browser/registry.yaml
thompson-tomo: The navigation sub-namespace feels a better fit. Maybe browser.navigation.change_type would be more appropriate if it is kept but I will let others decide if it is of value.

model/browser/events.yaml
thompson-tomo: Agree with the reason for changing the name. Happy with the current solution or perhaps browser.page.navigation but don't have a strong opinion

model/browser/registry.yaml
joaquin-diaz: Agree, I also like `browser.navigation.type`

overbalance: Same. There are other nav types: reload and back_forward. What about:
```
load
soft
reload
back_forward
```

model/browser/registry.yaml
thompson-tomo: Attribute has been removed. So discussion resolved.

model/browser/events.yaml
martinkuba: removed, we can add it later if we find there is value to it

docs/browser/events.md
```diff
@@ -51,4 +51,65 @@ semantic convention tooling supports complex attributes
...
+<!-- semconv event.browser.page.view -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `browser.page.view`.
```
overbalance, line 65:
```suggestion
The event name MUST be `browser.page_view`.
```

thompson-tomo: There was a discussion about removing usages of browser.page_view hence this was changed to this option especially as browser.page_view is not used anywhere else

overbalance: Ah, so the title is wrong then.

thompson-tomo: Correct, good  spot.

martinkuba: I have updated the title.

model/browser/registry.yaml
thompson-tomo: @martinkuba did we want to go with the change to browser.navigation.type especially as it would also potentially allow the additional types

martinkuba: We discussed this in the SIG meeting last week and decided to keep it like this. Navigation type already exists in the [navigation API](https://developer.mozilla.org/en-US/docs/Web/API/NavigateEvent/navigationType), so did not want to overload it.

We might want to add `navigation.type` attribute in the future, but it would have this other meaning (captured from `navigation.activation.navigationType`).

martinkuba:
This proposal has been superseded by [#2806 - Add browser navigation event](https://github.com/open-telemetry/semantic-conventions/pull/2806)

PR #1911, opened 2025-02-17
Add Semantic definitions for Oracle Database

sudarshan12s:
Fixes #

## Changes

Add semantic conventions for Oracle Database.


schema-next.yaml
```diff
@@ -7,6 +7,22 @@ versions:
         - rename_metrics:
             k8s.replication_controller.desired_pods: k8s.replicationcontroller.desired_pods
             k8s.replication_controller.available_pods: k8s.replicationcontroller.available_pods
+    all:
+      changes:
+        # https://github.com/open-telemetry/semantic-conventions/pull/TODO
+        # add semantic conventions for Oracle Database
+        - add_attributes:
```
jsuereth, line 14: I don't think this file supports an add_attributes field.

model/database/spans.yaml
```diff
@@ -838,3 +838,26 @@ groups:
       - ref: azure.cosmosdb.operation.contacted_regions
         requirement_level:
           conditionally_required: If available.
+  - id: span.db.oracledb.client
+    type: span
+    stability: experimental
```
lmolkova, line 843: nit:

```suggestion
    stability: development
```

docs/attributes-registry/README.md
```diff
@@ -82,6 +82,7 @@ Currently, the following namespaces exist:
 - [NodeJS](nodejs.md)
 - [OCI](oci.md)
 - [OpenTracing](opentracing.md)
+- [Oracledb](oracledb.md)
```
lmolkova, line 85: please add acronym `OracleDB` to https://github.com/open-telemetry/semantic-conventions/blob/d3bff0dcb5dae5d1969f3d1dcced070cb3da88fd/templates/registry/markdown/weaver.yaml#L8

model/oracledb/registry.yaml
```diff
@@ -0,0 +1,42 @@
+groups:
+  - id: registry.oracledb
+    type: attribute_group
+    display_name: Oracle Database Attributes
+    brief: >
+      This section defines attributes for Oracle Database.
+    attributes:
+      - id: oracledb.implicit_release
```
lmolkova, line 8: we avoid adding attributes without referencing them on the spans, metrics, events, resources, etc. 
Is there a span/metrics that would use this attribute? Could you define it?

Otherwise, let's postpone adding this and other attributes until we're ready to define signals they should be set on.

lmolkova, line 8: system-specific attribute namespace needs to match the `db.system.name` identifier (i.e. it has to be `oracle.db`

model/oracledb/registry.yaml
```diff
@@ -0,0 +1,42 @@
...
+        brief: >
+          The name of the pluggable database (PDB) the connection is using.
+        examples: "PDB1"
+      - id: oracledb.pool.increment
+        type: int
+        stability: development
+        brief: >
+          The number of connections added when the pool needs to expand.
+        examples: 5
+      - id: oracledb.pool.max
```
lmolkova, line 31: please check out DB connection pool metrics - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#connection-pools and check if anything oracle-specific is necessary.

lmolkova (changes requested):

model/oracledb/registry.yaml
sudarshan12s: Yes modified `oracledb.implicit_release `to `oracle.db.implicit_release`

similarly changed for other attributes too.

docs/attributes-registry/README.md
sudarshan12s: Added it.

model/oracledb/registry.yaml
sudarshan12s: Yes this attribute  `oracle.db.client.connection.implicit_release` will appear in span, when a query is executed on the connection and an implicit release of connection is done. It has only value `true` and if there is no implicit release of connection, it won't appear in the span. 

[node-oracledb](https://github.com/oracle/node-oracledb) @6.7.0 onwards has code to add this information passed to OT [module](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2612) to add it in span. 

I added the attribute in `model/database/spans.yaml `.Please suggest if this looks fine?

model/oracledb/registry.yaml
sudarshan12s: Thanks for the link. I removed `oracledb.pool.max` and `oracledb.pool.min` as they are specified in standard as
 `db.client.connection.max` and `db.client.connection.idle.min  `respectively.
 
I was adding these parameters in span when application creates a pool with the pool configuration (max, min ,..). I think the standard talks about this attribute to be included in metrics as `UpDownCounter`, Can I still keep the same name in spans .

docs/database/oracledb.md
```diff
@@ -0,0 +1,115 @@
...
+| [`db.query.summary`](/docs/attributes-registry/db.md) | string | Low cardinality representation of a database query text. [7] | `SELECT wuser_table`; `INSERT shipping_details SELECT orders`; `get user by id` | `Recommended` [8] | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
+| [`db.query.text`](/docs/attributes-registry/db.md) | string | The database query being executed. [9] | `SELECT * FROM wuser_table where username = :mykey` | `Recommended` [10] | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
+| [`db.response.returned_rows`](/docs/attributes-registry/db.md) | int | Number of rows returned by the operation. | `10`; `30`; `1000` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`oracle.db.client.connection.implicit_release`](/docs/attributes-registry/oracledb.md) | boolean | Boolean flag Indicating if the internal connection is released back to pool or not after executing a query. [11] | `true` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 43: I'm not sure if that's possible to add it to DB spans. E.g. if I take a JDBC example:

```java
  public static void viewTable() throws SQLException {

    String query = "select COF_NAME, SUP_ID, PRICE, SALES, TOTAL from COFFEES";
    try (Connection conn = odsconn.getConnection()) {
      try (Statement stmt = con.createStatement()) {
        ResultSet rs = stmt.executeQuery(query);
        // this is where SQL span ends
        while (rs.next()) {
          ...
        }
      } catch (SQLException e) {
      }
    }
  } // this where the connection is returned to the pool
```

it seems, DB span will be ended by the time connection is closed. 

Let me know if I'm missing something

docs/database/oracledb.md
```diff
@@ -0,0 +1,115 @@
...
+| [`oracle.db.client.connection.increment`](/docs/attributes-registry/oracledb.md) | int | The number of connections added when the pool needs to expand. | `5` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`oracle.db.instance`](/docs/attributes-registry/oracledb.md) | string | The name of the Oracle Database instance. | `ORCL1` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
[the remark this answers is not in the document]
sudarshan12s: Thanks. I understand, your suggestion is to combine  `instance name` and` pdb name` with `db.namespace`(it only has service name now).


The `service name` is the input given by user for getting a connection along with `hostname`, `port`. 
```
Ex: 
const configSV1 = "hostname:port/servicename" // configuration for getting a connection to database.
const conn1 = await oracledb.getConnection(configSV1) // get first connection to database, It can connect to instance1.
const conn2 = await oracledb.getConnection(configSV1) // get second connection to database, It can connect to instance2.
```
 The `instance name` and `pdb name `are filled after a successful connection. So it is possible that `conn1` can get a  different` instance name`  from `conn2`.  

Will this be an issue, If `db.namespace` includes additional details (runtime detected) like `instance name` and `pdb name` for the same user input, `configSV1` like co-relating spans,.. ?

docs/database/oracledb.md
sudarshan12s: > As part of `stmt.executeQuery(query);`, the `oracledb` driver adds a span attribute , `oracle.db.client.connection.implicit_release = true`, if the completion of the query execution can implicitly [release](https://node-oracledb.readthedocs.io/en/latest/user_guide/connection_handling.html#implicitpool) the connection back to pool. Hence this flag, `implicit_release` is detected at end of reading the query response from server.
> 
> Probably I should rename this `oracle.db.client.connection.implicit_release to` oracle.db.query.implicit_release` ?.

just to add on this, the server sends the information that an implicit release is done , only if there are no more rows (there are more other conditions to be satisfied) to be fetched from application, So server can send this flag as part of `rs.next()` call too.  I hope i could clarify.

model/oracledb/registry.yaml
sudarshan12s: I removed oracledb.pool.max,  oracledb.pool.incr and oracledb.pool.min .

docs/database/oracledb.md
```diff
@@ -0,0 +1,115 @@
...
+| [`db.query.text`](/docs/attributes-registry/db.md) | string | The database query being executed. [9] | `SELECT * FROM wuser_table where username = :mykey` | `Recommended` [10] | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
+| [`db.response.returned_rows`](/docs/attributes-registry/db.md) | int | Number of rows returned by the operation. | `10`; `30`; `1000` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`oracle.db.client.connection.implicit_release`](/docs/attributes-registry/oracledb.md) | boolean | Boolean flag Indicating if the internal connection is released back to pool or not after executing a query. [11] | `true` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`oracle.db.client.connection.increment`](/docs/attributes-registry/oracledb.md) | int | The number of connections added when the pool needs to expand. | `5` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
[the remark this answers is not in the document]
sudarshan12s: I removed `oracle.db.client.connection.increment` attribute.

docs/database/oracledb.md
```diff
@@ -0,0 +1,115 @@
...
+| [`oracle.db.instance`](/docs/attributes-registry/oracledb.md) | string | The name of the Oracle Database instance. | `ORCL1` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`oracle.db.pdb_name`](/docs/attributes-registry/oracledb.md) | string | The name of the pluggable database (PDB) the connection is using. | `PDB1` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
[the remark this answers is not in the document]
sudarshan12s: I documented db.namespace to include this component too . Removed  `oracle.db.pdb_name`

sudarshan12s:
@lmolkova, @jsuereth, could you please take another look at this PR? I've addressed all the feedback. Thanks!

docs/database/oracledb.md
lmolkova: the problem is that DB span is limited to the scope of `ResultSet rs = stmt.executeQuery(query);` - we don't know if user will ever call `rs.next()` (and if they would read to the end). 

I'm still not sure how we can populate this attribute on DB spans. It might make sense to look into separate telemetry for oracle connections (spans or events that would describe connection transitions and durations). Here's an attempt to define it for HTTP - https://github.com/open-telemetry/semantic-conventions/pull/1192

docs/database/oracledb.md
```diff
@@ -0,0 +1,110 @@
...
+<!-- semconv span.db.oracledb.client -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`db.namespace`](/docs/attributes-registry/db.md) | string | The database associated with the connection, qualified by the instance name, database name and service name. [1] | `ORCL1|PDB1|db_high.adb.oraclecloud.com`; `ORCL1|DB1|db_low.adb.oraclecloud.com` | `Conditionally Required` If available without an additional network call. | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
```
lmolkova, line 35: I would expect the domain name to be in `server.address` and `server.port`, is there a reason to put it in `db.namespace`?

model/database/spans.yaml
```diff
@@ -838,3 +838,53 @@ groups:
...
+      Attributes for Oracle Database
+    attributes:
+      - ref: db.namespace
+        sampling_relevant: true
+        brief: >
+          The database associated with the connection, qualified by the instance name,
+          database name and service name.
+        note: |
+          `db.namespace` SHOULD be set to the combination of instance name, database name and
+          service name following the `{instance_name}|{database_name}|{service_name}` pattern.
```
lmolkova, line 856: any reason not to follow the same pattern as MS SQL? (use `.` as a separator)?

https://github.com/open-telemetry/semantic-conventions/blob/e29f8dd779832b6011437e5e49288505ec8fe19c/model/database/spans.yaml#L128

I.e. `{instance_name},{database_name}`,

also check out my other comment on `service_name` - it should probably be in the `server.address` and don't need to be in the namespace

docs/database/oracledb.md
sudarshan12s: Yes, If the response to DB operation, `ResultSet rs = stmt.executeQuery(query);` indicates the flag is not true, this DB span will not have this attribute.

Each database operation synchronously updates this state in its response. So from the client side, It is only detected on explicit roundtrip to DB server for DB operation.  It is not an asynchronous event triggered from server.

If user has never called `rs.next()` or never read all rows, the attribute would never appear in the spans. Does it makes sense? 


I will try to explain the flow, Please see if it clarifies?

```
ResultSet rs = stmt.executeQuery(query); =>
createSpan , sp1
Run the API stmt.executeQuery(query);
sp1.setSpanAttributes(db.query.text = query, .. oracle.db.client.connection.implicit_release = true,..) // set span attributes parsing the result (we determine if flag, implicit_release is set) , inputs (set db.query.text based on config, opt-in) ..
sp1.end() // end span
```

just to add, this release of connection indicated by server will not disconnect client socket from server, it just informs the client that the DRCP server process is free to serve other connections. The physical connection from client to server(now to broker not to server process) still remains.

sudarshan12s: just to elaborate : 

When the application executes `ResultSet rs = stmt.executeQuery(query);`, a trace is generated with a root span containing key details such as db.query.text (holding the query string), db.namespace, db.serveraddr, and other relevant database attributes.

Additionally, a child span is created as part of the roundtrip to the database. This roundtrip span is responsible for capturing and setting the implicit_release attribute, which indicates if server process was implicitly released.

I am specifically explaining the code changes made for the JavaScript driver, as this functionality has not yet been implemented in the JDBC driver.

In the JavaScript node-oracledb driver, this state is set during the roundtrip process at [this location](https://github.com/oracle/node-oracledb/blob/2e8963a5a9c3037901d46c947ffff56deabcfc43/lib/thin/protocol/protocol.js#L182).

In the OpenTelemetry module, this same information in populated in span [here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2612/files#diff-e1900808d7c0d6f55d09d17a210a08591ca21f28f5c39e6322b96f9853eacd30R269).

sudarshan12s: @lmolkova , I can rename `oracle.db.client.connection.implicit_release` to `oracle.db.drcp.server_process_release` to be more precise, or I can remove this attribute for now and address it in a separate PR for further discussion and conclusion.

lmolkova: Looking into https://github.com/open-telemetry/opentelemetry-js-contrib/blob/5defed847c8c67e8cff321a10d6b2c6cfadb3f1f/plugins/node/opentelemetry-instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts#L335

```js
if (traceContext.additionalConfig.implicitRelease) {
  span.setAttribute(
    AttributeNames.ORACLE_IMPLICIT_RELEASE,
    traceContext.additionalConfig.implicitRelease
  );
}
```

it seems like a static configuration property reported on every span. It also does not represent an actual DB behavior. Do I miss something? 

This can as well be reported by the end users as an event describing their DB configuration or, if they want to report it on every span, they can write a span processor that would set it.

docs/database/oracledb.md
lmolkova: thanks! so the service name is usually not the same as domain name? Does it usually follow FQDN pattern? it makes sense.

 If it's not super-common to use FQDN for service-name, would it be possible to update an example?

model/database/spans.yaml
lmolkova: it sounds good, please keep the `|`.

I'm leaving this discussion unresolved for other reviewers to see it

lmolkova (commented):
I believe the only unresolved discussion is around `oracle.db.client.connection.implicit_release` - suggesting to remove it and tackle in a follow-up PR along with transport span definition. 

Thanks a lot for pushing through!

docs/database/oracledb.md
sudarshan12s: Yes the service name need not follow FQDN pattern. I have added an example without FQDN:

`"ORCL1|DB1|order-processing-service"`

docs/database/oracledb.md
sudarshan12s: Thanks @lmolkova . I understood., I will raise a different PR to describe this..                                                             
 This attribute does not appear in every round trip span issued for running db query —only in specific cases where the query execution on server caused server process to be freed. Since it only has a `true` value and is omitted otherwise, it represents a runtime behaviour .  If `traceContext.additionalConfig.implicitRelease` is defined span will see this attribute with value true.

sudarshan12s:
> I believe the only unresolved discussion is around `oracle.db.client.connection.implicit_release` - suggesting to remove it and tackle in a follow-up PR along with transport span definition.
> 
> Thanks a lot for pushing through!

I have removed the attribute  `oracle.db.client.connection.implicit_release` as suggested.

docs/database/oracledb.md
```diff
@@ -0,0 +1,116 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Oracle Database
+--->
+
+# Semantic conventions for Oracle Database
+
+**Status**: [Release Candidate][DocumentStatus]
```
lmolkova, line 7: I think we should start with Development. We plan to declare stability for ones marked with RC in ~ a month, but we need to let Oracle bake for a bit.

For example, we'd need prototypes/instrumentations in 3 languages. 

```suggestion
**Status**: [Development][DocumentStatus]
```

lmolkova (commented):
It looks great with a tiny caveat on the doc status.

lmolkova (approved):

alexkeh:
@Kielek I'm asking one of our Oracle team members to review this. Thanks for alerting us!

.chloggen/add_oracledb_semconv.yaml
```diff
@@ -0,0 +1,23 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Adding semnatic conventions for `oracledb` instrumentations.
```
trask, line 13:
```suggestion
note: Adding semantic conventions for `oracledb` instrumentations.
```

model/database/spans.yaml
trask: @alanwest @lmolkova what do you think of changing the sqlserver separator from `.` to `|`?

trask (approved):

.chloggen/add_oracledb_semconv.yaml
sudarshan12s: Done

model/database/spans.yaml
trask: resolving discussion, I opened an issue to discuss / decide this for sqlserver

PR #1915, opened 2025-02-17
Add service.instance.pool

### LINKED ISSUE #1404: Add service.instance.pool
### Area(s)

area:service

Instances of a service can be grouped into pools. For example, a service could be split into two pools: `canary` and `default`. Canary being a small pool that is deployed before default to check that everything is working as expected. Some pools can also be test pools of one or two instances to test a change in production.

The metrics can be aggregated by pool to easily compare the canary and default pools but currently there is no standard labels for that concept.

A new label `service.instance.pool` which designates the pool the instance is part of. The value would depend on the company but some recommended values would be:
- `default`
- `canary`
- `test-1`
- `test-2`

- `service.instance.pool.name` in case a pool can have other fields, such as an id + name
- `service.pool` but I think it makes more sense to add it at the instance level

### Additional context

_No response_

verdie-g:
Fixes #1404

## Changes

Add `service.instance.pool` in the service resource. More info in the issue.


jsuereth:
The service resource is treated specially in open telemetry, allowing interaction with environment variables.   What the intention for how to fil out an instance pool name?

verdie-g:
Answered in the issue https://github.com/open-telemetry/semantic-conventions/issues/1404.

PR #1919, opened 2025-02-18
Add Browser navigation timing event

martinkuba:
## Changes

This adds a new event for capturing the navigation timing date in browser/web applications.

Related to https://github.com/open-telemetry/opentelemetry-browser/issues/16


model/browser/events.yaml
```diff
@@ -0,0 +1,85 @@
...
+             navigation, and it is measured in milliseconds.
+          requirement_level: recommended
+          examples: [189.0]
+        - id: domComplete
+          type: double
+          stability: development
+          brief: >
+            The time immediately before the user agent sets the document's
+            readyState to "complete". The value is relative to the start of the
+            navigation, and it is measured in milliseconds.
```
trask, line 74: I'd suggest moving the second sentence to the `note` (to make the `brief` shorter)

and maybe add links to the properties in each case, to be super explicit that these are defined externally, e.g. https://www.w3.org/TR/navigation-timing-2/#dom-performancenavigationtiming-domcomplete

model/browser/events.yaml
```diff
@@ -0,0 +1,85 @@
...
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: name 
+          type: string
+          stability: development
+          brief: The resolved URL of the requested resource. 
+          requirement_level: recommended
+          examples: ["https://en.wikipedia.org/wiki/Main_Page"]
```
thompson-tomo, line 31: Feel that this should be an attribute

model/browser/events.yaml
```diff
@@ -0,0 +1,85 @@
+groups:
+  - id: event.browser.navigation_timing
+    stability: development
+    type: event
```
thompson-tomo, line 4: Based on description shouldn't this be metric?

model/browser/events.yaml
```diff
@@ -0,0 +1,85 @@
...
+          examples: [506.5]
+        - id: loadEventEnd
+          type: double
+          stability: development
+          brief: >
+            The time immediately after the current document's load event handler
+            completes. The value is relative to the start of the navigation, and
+            it is measured in milliseconds.
+          requirement_level: recommended
+          examples: [506.70000000018626]
```
thompson-tomo, line 85: If represented as a metric we could have just a state attribute.

thompson-tomo (changes requested):

PR #1631, opened 2024-11-29
Add SDK span telemetry metrics

JonasKunz:
## Changes

With this PR I'd like to start a discussion around adding SDK self-monitoring metrics to the semantic conventions.
The goal of these metrics is to give insights into how the SDK is performing, e.g. whether data is being dropped due to overload / misconfiguration or everything is healthy.
I'd like to add these to semconv to keep them language agnostic, so that for example a single dashboard can be used to visualize the health state of all SDKs used in a system.

We checked the SDK implementations, it seems like only the Java SDK currently has some health metrics implemented.
This PR took some inspiration from those and is intended to improve and therefore supersede them.

I'd like to start out with just span related metrics to keep the PR and discussions simpler here, but would follow up with similar PRs for logs and traces based on the discussion results on this PR.

### Prior work

This PR can be seen as a follow up to the closed [OTEP 259](https://github.com/open-telemetry/oteps/pull/259):
 * The OTEP originally superseded #184, which initially focused only on SDK exporter metrics.
 * #184 was closed in favor of the predecessor of [OTEP 259](https://github.com/open-telemetry/oteps/pull/259) to instead allow monitoring of entire "pipelines" using unified metrics across SDK exporters and collector components : A pipeline starts with an SDK exporter and goes through the processing of collector(s)
 * Finally, that OTEP was closed in favor of [this collector RFC](https://github.com/open-telemetry/opentelemetry-collector/pull/11406) which adds metrics just for the collector components.

 So we kind of have gone full circle: The discussion started with just SDK metrics (only for exporters), going to an approach to unify the metrics across SDK-exporters and collector, which then ended up with just collector metrics.
 So this PR can be seen as the required revival of #184 (see also [this comment](https://github.com/open-telemetry/oteps/pull/259#issuecomment-2422249166)).

 In my opinion, it is a good thing to separate the collector and SDK self-metrics:
  * There have been [concerns](https://github.com/open-telemetry/semantic-conventions/pull/184#issuecomment-1631394829) about both using the same metrics for both: How do you distinguish the metrics exposed by collector components from the self-monitoring metrics exposed by an Otel-SDK used in the collector for e.g. tracing the collector itself?
[...]

PR #1921, opened 2025-02-19
Add SDK Log health metrics

JonasKunz:
Follow up of #1631 adding SDK health metrics for logs.

## Changes

Adds:
 * A metric to count the number of logs generated submitted to the SDK (only enabled loggers count)
 * Count metrics for LogRecordProcessors and LogRecordExporters to track data loss

The processor and exporter metrics are very similar to the span-ones in #1631. In fact, they are pretty much a copy-paste.
So I think with the concrete examples at hand now, it would be a good time to revisit the decision and reasoning whether to keep the metrics separated per signal or combine them with and differentiate via an attribute (e.g. `otel.signal.kind`=`span`,`log`,`metric`).

The detailed reasoning for separation can be found in [this comment](https://github.com/open-telemetry/semantic-conventions/pull/1631#issuecomment-2589745739).

I still think that reasoning holds for processors: Log and Span Processors do have the same name, but work differently (e.g. span processor has multiple callbacks). Therefore it makes sense to have different namespaces for their metrics to not make things ugly in the future when new, span or log specific metrics need to be added or the processor specs diverge further. This already shows in the fact that metrics do not have processors, but `MetricReaders`.

~For the exporters I'm not so sure anymore: The exporters are separated in the spec, but it is in fact mostly copy-pasted and I think they are unlikely to diverge. So if we wanted to reduce the size of the spec here and ease some queries (e.g. "are there errors in any exporter?"), we could think of combining them.~
In the SIG meeting we decided to keep them separate for now, see #2002.

Looking forward to hearing your opinions here!

* [ ] ~[schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.~


model/otel/metrics.yaml
```diff
@@ -110,3 +110,98 @@ groups:
...
+          recommended: when applicable
+
+  - id: metric.otel.sdk.exporter.log.exported.count
+    type: metric
+    metric_name: otel.sdk.exporter.log.exported.count
+    stability: development
+    brief: "The number of LogRecords for which the export has finished, either successful or failed"
+    note: |
+      For successful exports, `error.type` MUST NOT be set. For failed exports, `error.type` must contain the failure cause.
+      For exporters with partial success semantics (e.g. OTLP with `rejected_spans`), rejected spans must count as failed and only non-rejected LogRecords count as success.
```
AlexanderWert, line 193: does this note about partial success also apply to log exporters? Currently it mentions `rejected spans`, which is a copy of the corresponding span metrics, I guess?

AlexanderWert (approved):

AlexanderWert:
@open-telemetry/specs-semconv-approvers @open-telemetry/specs-semconv-maintainers 
Can we get more reviews on this please?

This PR is a continuation of https://github.com/open-telemetry/semantic-conventions/pull/1631, now doing it for the Logs signal, but otherwise analogous to #1631.

model/otel/metrics.yaml
JonasKunz: Yeah, copy-paste mistake. I even already adjusted the end of the sentence: `and only non-rejected LogRecords count as success`

lmolkova:
related to https://github.com/open-telemetry/semantic-conventions/issues/1979

JonasKunz:
In accordance with #1979, I've adjusted the metric names in [bf5cd8d](https://github.com/open-telemetry/semantic-conventions/pull/1921/commits/bf5cd8d7638a1a85b9839c9bab8a4b27ecd9b12f).

lzchen (approved):

model/otel/metrics.yaml
```diff
@@ -110,3 +110,98 @@ groups:
...
+
+  - id: metric.otel.sdk.exporter.log.inflight_logs
+    type: metric
+    metric_name: otel.sdk.exporter.log.inflight_logs
+    stability: development
+    brief: "The number of LogRecords which were passed to the exporter, but that have not been exported yet (neither successful, nor failed)"
+    note: |
+      For successful exports, `error.type` MUST NOT be set. For failed exports, `error.type` must contain the failure cause.
+    instrument: updowncounter
+    unit: "{LogRecord}"
```
trask, line 175: I think snake case has been the existing convention for units:

```suggestion
    unit: "{log_record}"
```

model/otel/metrics.yaml
```diff
@@ -110,3 +110,98 @@ groups:
...
+      - ref: server.address
+        requirement_level:
+          recommended: when applicable
+      - ref: server.port
+        requirement_level:
+          recommended: when applicable
+
+  - id: metric.otel.sdk.exporter.log.exported_logs
+    type: metric
+    metric_name: otel.sdk.exporter.log.exported_logs
```
trask, line 188: I think it might be ok to remove the `_logs` suffix on all the metrics, similar to `jvm.class.loaded` / `jvm.class.unloaded`

```suggestion
    metric_name: otel.sdk.exporter.log.exported
```

model/otel/metrics.yaml
JonasKunz: Fixed in [1e1af04](https://github.com/open-telemetry/semantic-conventions/pull/1921/commits/1e1af04e42d5e39bafb5b5da586566e6ff038916)

model/otel/metrics.yaml
JonasKunz: SGTM. I also noticed I forgot about `otel.sdk.log.count` and renamed it to `otel.sdk.log.created` in [07f45ec](https://github.com/open-telemetry/semantic-conventions/pull/1921/commits/07f45ecf589cb520162178d3e52298efa5e08aa3).

docs/otel/sdk-metrics.md
```diff
@@ -20,6 +20,12 @@ This document describes metrics emitted by the OpenTelemetry SDK components them
   - [Metric: `otel.sdk.processor.span.processed.count`](#metric-otelsdkprocessorspanprocessedcount)
   - [Metric: `otel.sdk.exporter.span.inflight.count`](#metric-otelsdkexporterspaninflightcount)
   - [Metric: `otel.sdk.exporter.span.exported.count`](#metric-otelsdkexporterspanexportedcount)
```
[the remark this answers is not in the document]
JonasKunz: Definitely, I just wanted to do it in a follow-up PR after this one is merged, so that all naming discussions are settled prior. See https://github.com/open-telemetry/semantic-conventions/pull/1921#issuecomment-2732559616.

docs/otel/sdk-metrics.md
```diff
@@ -441,10 +467,414 @@ E.g. for Java the fully qualified classname SHOULD be used in this case.
...
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `otel.sdk.log.created` | Counter | `{log_record}` | The number of logs submitted to enabled SDK Loggers | ![Development](https://img.shields.io/badge/-development-blue) |
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `otel.otel.sdk.processor.log.queue.size`
```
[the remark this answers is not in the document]
JonasKunz: Fixed in https://github.com/open-telemetry/semantic-conventions/pull/1921/commits/0f2605d71d629b4feb254b19a7c7e5a38f1c4d0e

PR #1922, opened 2025-02-19
Define a way to indicate the number of requested response candidates.

michaelsafyan:
Proposes a mechanism to capture arguments related to the requested number of responses.

Both ChatGPT and Vertex AI offer the ability to request a specific number of candidate responses; in ChatGPT, this option is called `n`, while in Vertex AI, this option is called `candidateCount`.

This provides a means of capturing that option, while being general enough to allow for other APIs (by representing it as a range rather than a specific target number of responses).


model/gen-ai/registry.yaml
```diff
@@ -90,6 +90,18 @@ groups:
         type: int
         brief: The maximum number of tokens the model generates for a request.
         examples: [100]
+      - id: gen_ai.request.min_response_candidates
```
lmolkova, line 93: are there any APIs that have range? if not, I'd avoid adding this attribute:

https://github.com/open-telemetry/semantic-conventions/blob/c1f8e0a86410560da33f8cf3f18237ddf819fa1a/docs/general/how-to-define-semantic-conventions.md?plain=1#L50C8-L50C9

We can always add one later if/when the need comes

model/gen-ai/registry.yaml
```diff
@@ -90,6 +90,18 @@ groups:
...
+        stability: development
+        type: int
+        brief: The minimum number of candidate responses to return.
+        examples: [3]
+        note: Includes the specified value (a value of 3 means ">=3").
+      - id: gen_ai.request.max_response_candidates
```
lmolkova, line 99: I believe existing APIs don't consider it as a max, but generate as many responses as you request, so I'm not sure if `max` makes sense.

Suggesting `gen_ai.request.completion.count` (since we use `completion` terminology in docs) or `gen_ai.request.output.count` (since we use `output` in output tokens).

Given that OpenAI REST API is de-facto used by many providers, and OpenAI calls them choices, calling it `gen_ai.request.choice.count` would be familiar to the most of the users.

model/gen-ai/spans.yaml
```diff
@@ -16,6 +16,12 @@ groups:
         requirement_level: required
       - ref: gen_ai.request.max_tokens
         requirement_level: recommended
+      - ref: gen_ai.request.min_response_candidates
```
lmolkova, line 19: please generate MD files - https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#2-update-the-markdown-files

model/gen-ai/registry.yaml
michaelsafyan: Seems reasonable. Updated to `gen_ai.request.choice.count`.

model/gen-ai/registry.yaml
michaelsafyan: Fair point. Switched to a single target value.

model/gen-ai/spans.yaml
michaelsafyan: Thanks. I'm currently blocked on tooling:

https://github.com/open-telemetry/semantic-conventions/pull/1926

If you're able to help me get that in so that I can run the `make` workflow, I would appreciate that.

michaelsafyan: Thank for helping get the tooling dependency through. Updated the MD files using that process.

lmolkova (approved):
LGTM!

mattmccleary (approved):

jsuereth (approved):

Issue #984, opened 2024-04-30
Database: add examples


lmolkova:
It'd be great to have examples of DB spans for the following cases:

- SQL sanitization is enabled or disabled
- SQL parametrized query
- SQL stored procedure
- bulk/batch?

trask:
Targeting for post-frozen, pre-stability

trask:
cc @alanwest this could be addressed by your idea to add a "test suite" of examples

trask:
Resolved by #1923

PR #1923, opened 2025-02-19
Add test cases for DB query parsing and sanitization

alanwest:
Towards #984

Opening this PR to start a discussion for how we want to format test cases for database query parsing and sanitization.

Each test case includes the following fields

* A `name` identifying the test case
* A `query` statement representing the input
* `db.system.name` indicates which dialect that the test case targets. A value of `other_sql` indicates it is applicable to all dialects.
* An array of valid `db.query.text` values.
  * For example, the specification states that `IN` clauses MAY be collapsed. So, both `IN (?)` and `IN (?,?,?)` are valid.
* `db.query.summary` represents what `db.query.summary` should equal

For this first iteration I would like to settle on the format for presenting these test cases. In follow ups we can expand the test cases and also handle additional idiosyncrasies of other dialects.


alanwest:
@trask @lmolkova I plan to attend the DB semconv meeting tomorrow so we have a chance to discuss this synchronously first.

docs/non-normative/db-test-cases.json
```diff
@@ -0,0 +1,315 @@
...
+            "mssql"
+        ]
+    },
+    {
+        "name": "in_clause",
+        "sql": "SELECT * FROM table WHERE value IN (123, 456, 'abc')",
+        "sanitized": [
+            "SELECT * FROM table WHERE value IN (?)",
+            "SELECT * FROM table WHERE value IN (?, ?, ?)"
+        ],
```
trask, line 177:
```suggestion
        "sanitized": "SELECT * FROM table WHERE value IN (?, ?, ?)",
        "sanitizedWithCollapsedInClause": "SELECT * FROM table WHERE value IN (?)",
```

docs/non-normative/db-test-cases.json
```diff
@@ -0,0 +1,315 @@
...
+        "sql": "SELECT 12, -12, +12",
+        "sanitized": [
+            "SELECT ?, ?, ?"
+        ],
+        "summary": "SELECT",
+        "operation": "SELECT",
+        "collection": null,
+        "dialects": [
+            "mssql"
+        ]
```
trask, line 13:
```suggestion
        "db.system.name": "other_sql"
```

docs/non-normative/db-test-cases.json
trask, line 1: `db-sql-test-cases.json`

docs/non-normative/db-sql-test-cases.json
```diff
@@ -0,0 +1,219 @@
+[
```
lmolkova, line 1: let's put it into a sub-folder? e.g. `test-cases` or `database-test-cases`?

docs/non-normative/db-sql-test-cases.json
```diff
@@ -0,0 +1,219 @@
...
+    {
+        "name": "numeric_literal_integers",
```
lmolkova, line 3: would it be possible to add a small and informal description of this schema? Also, would it be possible to change it to differentiate inputs/outputs - e.g.

```json 
  {
    "test_name": "numeric_literal_integers",
    "input": { 
        "query": "SELECT 12, -12, +12",
        "db.namespace": "test"
     },
     "expected": {
       "span_name": "SELECT",
       "attributes": {
          "db.system.name": "other_sql",
          "db.query.text": "SELECT ?, ?, ?",
          "db.query.summary": "SELECT"
       }
  }
```

docs/non-normative/db-sql-test-cases.json
```diff
@@ -0,0 +1,219 @@
...
+        "db.system.name": "other_sql",
+        "query": "SELECT 12, -12, +12",
+        "db.query.text": [
```
lmolkova, line 6: any reason it's an array?

docs/non-normative/db-sql-test-cases.json
```diff
@@ -0,0 +1,219 @@
...
+        "db.query.summary": "SELECT Orders OrderDetails"
+    },
+    {
+        "name": "cross_join_comma_separated_syntax",
+        "db.system.name": "other_sql",
+        "query": "SELECT * FROM Orders o, OrderDetails od",
+        "db.query.text": [
+            "SELECT * FROM Orders o, OrderDetails od"
+        ],
+        "db.query.summary": "SELECT Orders"
```
lmolkova, line 190: why not `SELECT Orders, OrderDetails` ?

docs/non-normative/db-sql-test-cases.json
trask: > differentiate inputs/outputs

nice 👍

alanwest: Yes I like this idea. Few things to consider though

1. `db.system.name` in this context is treated as an input rather than an output. Originally, I called this field `dialects`. The idea is that many test cases are applicable to multiple dialects. So, it would have a structure like following:

```json
{
    "test_name": "numeric_literal_integers",
    "input": { 
        "query": "SELECT 12, -12, +12",
        "dialects": [
            "mssql",
            "postgres",
            "mysql"
        ]
     },
     "expected": {
       "span_name": "SELECT",
       "attributes": {
          "db.system.name": "whatever the input dialect was",
          "db.query.text": "SELECT ?, ?, ?",
          "db.query.summary": "SELECT"
       }
  }
```

When @trask and I discussed this, we toyed with this idea of using `other_sql` to indicate that a test case was applicable to all dialects simplifying things so the array was unnecessary.

2. What your idea regarding adding `db.namespace` as an input?

docs/non-normative/db-sql-test-cases.json
alanwest: This was a mistake. Fixed. It should be `SELECT Orders OrderDetails` (i.e., no comma between Orders and OrderDetails)

docs/non-normative/db-sql-test-cases.json
lmolkova: having an array of dialects sounds like a good future possibility, I like the idea of using other_sql to keep things simple now.

> What your idea regarding adding db.namespace as an input?

I was thinking about edge cases when you'd run a query with database a target (e.g. list tables), then the [span name would include the namespace](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#name). It can wait until we have a test-case for it.

docs/non-normative/database-test-cases/db-sql-test-cases.json
```diff
@@ -0,0 +1,315 @@
...
+        "name": "create_table",
+        "input": {
+            "db.system.name": "other_sql",
+            "query": "CREATE  TABLE MyTable (\n    ID NOT NULL IDENTITY(1,1) PRIMARY KEY\n)"
+        },
+        "expected": {
+            "db.query.text": [
+                "CREATE  TABLE MyTable (\n    ID NOT NULL IDENTITY(?,?) PRIMARY KEY\n)"
+            ],
+            "db.query.summary": "CREATE  TABLE MyTable"
```
trask, line 286: while preserving case has benefits (some people prefer lower vs upper, and some systems might be case sensitive), do you think there's any downside to normalizing spaces?

alanwest: Right, forgot we discussed this. I normalized the whitespace in `db.query.summary` but left `db.query.text` alone. WDYT?

lmolkova (approved):

trask (approved):

maryliag (approved):

PR #1941, opened 2025-02-26
Add Browser UserAction Event

Karlie-777:
Fixes #

## Changes


model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
...
+          requirement_level: recommended
+          examples: ["button"]
+        - id: element_xpath
+          type: string
+          stability: development
+          brief: Target element xpath
+          requirement_level: recommended
+          examples: ["//*[@id='testBtn']"]
+        - id: user_action_type
+          type: enum
```
Karlie-777, line 27: need to confirm the `user_action_type` enum. Currently, I am using the definitions from https://github.com/microsoft/ApplicationInsights-JS/blob/main/extensions/applicationinsights-clickanalytics-js/src/Enums.ts

model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
+groups:
+  - id: event.browser.user_action
+    stability: development
+    type: event
+    name: browser.user_action
+    brief: >
+      This event describes actions performed by the user such as click, scroll, zoom, resize, etc.
+    body:
```
breedx-splk, line 8: I would like to see a comment here to clarify that this is temporary or that this will actually be an attribute. That way readers do not expect to find content in the actual body.

model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
...
+      id: browser.user_action
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: element
+          type: string
+          stability: development
+          brief: Target element tag name and it is obtained via event.target.tagName.
```
breedx-splk, line 17:
```suggestion
          brief: Target element tag name and it is obtained via `event.target.tagName`.
```

model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
...
+            Type of interaction.
+            See enum [here](https://github.com/microsoft/ApplicationInsights-JS/blob/main/extensions/applicationinsights-clickanalytics-js/src/Enums.ts) for potential values we could add support for.
+          requirement_level: required
+          examples: ["cl"]
+        - id: click_coordinates
+          type: string
+          stability: development
+          brief: Click coordinates captured as a string in the format {x}X{y}.
+          requirement_level: recommended
+          examples: ["345X23"]
```
breedx-splk, line 76: Is this specific enough? Should this also specify if the coordinates are relative to the element or the page or the window or ???

model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
+groups:
+  - id: event.browser.user_action
+    stability: development
+    type: event
+    name: browser.user_action
```
breedx-splk, line 5: Do we think there's an opportunity to use another prefix/namespace other than `browser`? It seems like this is something that mobile (ios/android) could certainly also benefit from as well. If we can't make it broader to support multiple clients, then I fear we will encounter duplication later. I also undertand if the answer is "no, it doesn't make sense because browser already has all this stuff (eg. element xpath below, etc)"...but I think we should at least consider it.

Possible alternative namespaces: `client.user_action`, `app.user_action`, `

breedx-splk (approved):
I think this is good, especially as a first pass "development" event, thanks! I left some comments and think there is room for improvement and possibly sharing/broadening, but I don't want those to block. Thanks for making this bit-sized!

MSNev (approved):

model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
...
+          type: string
+          stability: development
+          brief: Target element xpath
+          requirement_level: recommended
+          examples: ["//*[@id='testBtn']"]
+        - id: user_action_type
+          type: enum
+          members:
+            - id: clickleft
+              value: "cl"
```
trask, line 30: is `cl` an existing abbrevation defined somewhere (e.g. a browser spec)? thanks!

model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
...
+              value: "o"
+              stability: development
+              brief: User actions that are not listed above.
+          stability: development
+          brief: >
+            Type of interaction.
+            See enum [here](https://github.com/microsoft/ApplicationInsights-JS/blob/main/extensions/applicationinsights-clickanalytics-js/src/Enums.ts) for potential values we could add support for.
+          requirement_level: required
+          examples: ["cl"]
+        - id: click_coordinates
```
trask, line 71: what do you think about
- `click.x`
- `click.y`

or is {x}Y{y} a browser spec thing? thanks

model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
+groups:
+  - id: event.browser.user_action
```
Karlie-777, line 2: May use client?

model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
...
+    name: browser.user_action
+    brief: >
+      This event describes actions performed by the user such as click, scroll, zoom, resize, etc.
+    body:
+      id: browser.user_action
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: element
```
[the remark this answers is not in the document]
MSNev: @trask Sorry No!
These are NOT top-level attributes and do not (and should not) conform to the full attribute naming guidelines (ONLY top-level attributes should and MUST conform to that specification) -- NOT the contents of the body / value of a complex attribute.

If / when the `body` becomes a standard complex attribute then there will need to be a convention in place about how these event specific attributes are named, which will most likely need to include the actual event name as the `prefix` (`browser.user_action.[payload/data]` -- for a single complete attribute).

Then there is just the overhead of creating individual on-off fields (only used for a single event) and going through the approvals / discussions about each individual field and how it can / should be made generic. Vs having a single discussion about the "naming" of fields within a specific event. If / When the community decides that they want to reuse the same type of "thing" identified by the individual event specific name, then at that point for those few fields (like the already defined `url.full`) it makes sense to have the broader naming / namespace discussions and then the event specific "changes" to pass that value as a top-level attribute rather than in the single value body / payload complex attribute.

While I know there has been discussions about "allowing" complex types to be defined as individual fields but sent in a complete manner, this is would not scalable and would also mean that for every event specific field for every event the names would become something like `browser.user_action.XXXX`, so not whenever an event has to be constructed the these long attribute names have to be constructed (and destructed -- for transport). This WILL result in
* larger code sizes (if these are included in the global semantic conventions and consumed) -- yes instrumentations could construct by thats another issue
* larger install sizes (for servers) if these are now included in the global semantic conventions
* More CPU used in constructing these events (to create the log record) and then even more to convert and send them
* More memory usage in generating these events
* More storage costs for these events (as browsers will generate billions/trillions of these events), a single user will generate hundreds of these events in a single session and large sites have 100's millions of daily users


Additional, part of the reason that the browser environment NEEDS to keep these in a single location is because
- (already mentioned but) These are specific and unique to each event and should only be compared against other events with the same name it does not need to be elevated to a top-level attribute
- These need to be kept as small as possible as this is specific for an environment that has extremely strict size, memory, network bandwidth constraints
- If / when this moves from `body` to a `single` complex attribute whose value contains these, then again it doesn't need to have a full namespace.

The client Sig spent months defining the "common" set of values for the event payloads based on different companies (well before some of the attributes became stable and renamed (like the `url.*` discussions), so while moving "some" of the fields to top-level attributes, most of the fields should not.

model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
...
+          type: string
+          stability: development
+          brief: Target element tag name and it is obtained via event.target.tagName.
+          requirement_level: recommended
+          examples: ["button"]
+        - id: element_xpath
```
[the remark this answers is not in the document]
MSNev: Some of the documentation (if you care to read) around how we got to these names in the client SIG are located in the Archived section of the [Client SIG meeting notes](https://docs.google.com/document/d/16Vsdh-DM72AfMg_FIt9yT9ExEWF4A_vRbQ3jRNBe09w/edit?tab=t.0#heading=h.vwohigf22rip) this inlcudes documents and spreadsheets used to get consensus on names to be used

If your objecting the the `_` then I would propose using just `xpath`

model/browser/events.yaml
```diff
@@ -0,0 +1,82 @@
...
+          type: string
+          stability: development
+          brief: Target element xpath
+          requirement_level: recommended
+          examples: ["//*[@id='testBtn']"]
+        - id: user_action_type
```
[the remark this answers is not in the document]
MSNev: No, because it's identifying the "type" of user actions, we have only defined a small common initial set of user actions that most RUM companies currently monitor, there are more but we (client SIG) needs to start with the initial common (most used) types of user interactions that get monitored.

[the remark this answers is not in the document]
MSNev: For a quick example here is a collection of **common** user actions that are collected by [some Microsoft teams](
https://[js.monitor.azure.com/scripts/c/behaviors.js](https://js.monitor.azure.com/scripts/c/behaviors.js), where this JS file is downloaded and used by our SDK's.
This does not include additional team specific user actions that some additional frameworks have.

model/browser/events.yaml
MSNev: We had this discussion in the Client Sig when we came up with this name, it went through many iterations and we settled on a single field formatted in this manner so that backends (and users) can add any processing they want around this vs the overhead of populating, sending, storing and processing of individual fields.

model/browser/events.yaml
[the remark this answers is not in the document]
trask: I'm not proposing a top-level `element.*` namespace, this would still be under the `browser.user_action` complex attribute in the future, e.g. `browser.user_action.element.tag.name`

model/browser/events.yaml
MSNev: Maybe, but that is a larger discussion for the Client SIG. We can always deprecate this event in favor of any stable standard. But generally it would be "nice" to have to define a commonly used event once -- but there are some very browser centric definitions here at the moment.

model/browser/events.yaml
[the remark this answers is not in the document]
trask: @MSNev I didn't follow why you disagree with the suggested change of renaming this? (especially since I proposed a shorter name)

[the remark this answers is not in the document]
MSNev: Ok, I merged the question from this point with the next issue on the "cl" and came up with this response, talking about what are the types.

I went to look at our [source document](https://docs.google.com/document/d/1_5MtDy_8uovK3cQEmwrVmFz0ApLQvQGsYX4ME9Kl3pg/edit?tab=t.0#heading=h.6qpmeyw546db) for the agreements that we had in the Client SIG and we (Client SIG) actually also identified that should be just "type".

So, yes `type` is what this should be.

model/browser/events.yaml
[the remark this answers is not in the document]
trask: I'm ok with keeping this as just `element`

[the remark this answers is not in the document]
trask: on further discussion, probably want `element.tag.name` since otherwise we can't have both `.element` and `element.id`

model/browser/events.yaml
breedx-splk: For reference, the Android `MotionEvent` provides x and y as separate fields. https://developer.android.com/reference/android/view/MotionEvent

trask: discussion in SIG meeting leaning towards 2-value number array

model/browser/events.yaml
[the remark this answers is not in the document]
MSNev: @breedx-splk , based on us changing this from a `browser` focus to a generic `user` focus is `element` the correct description of the "thing" being tracked here?

I know for .NET there is no "Element" (so the tag name is not relevant as it's more "type" (button, text box)) but it does have an `id` for the GUI object...

Originally (years ago), we started out calling this `target`. So really looking for input on a generic name that would work for most languages / frameworks.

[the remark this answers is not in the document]
MSNev: And to be a little clearer here, (if) `element` is fine for android as it's a common term for identifying this, then I'm fine to have this be called `element` (as previously agreed by the client Sig) -- especially as we don't have any .Net Client representation. We could just add additional documentation describing what we mean by the term `element` and call it done.

model/browser/events.yaml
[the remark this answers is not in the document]
trask: just `xpath` works for me

model/browser/events.yaml
Karlie-777: event.enduser/client.user_action

model/browser/events.yaml
Karlie-777: enduser.user_action

model/browser/events.yaml
Karlie-777: will change the enum to long name

model/browser/events.yaml
Karlie-777: example: [10,10]

model/browser/events.yaml
Karlie-777: changed to full names, thanks!

model/browser/events.yaml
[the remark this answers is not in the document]
Karlie-777: changed to xpath

model/browser/events.yaml
```diff
@@ -0,0 +1,83 @@
...
+          type: enum
+          members:
+            - id: clickleft
+              value: "clickleft"
```
trask, line 30: (see convention we're using in db.system.name now)

```suggestion
            - id: click.left
              value: "click.left"
```

Karlie-777: changed
`clickright` to `click.right`
`clickleft` to `click.left`
`clickmiddle` to `click.middle`

model/browser/events.yaml
```diff
@@ -0,0 +1,83 @@
...
+            - id: zoom
+              value: "zoom"
+              stability: development
+              brief: An element is zoomed by a user.
+            - id: resize
+              value: "resize"
+              stability: development
+              brief: An element is resized by a user.
+            - id: keyboardenter
+              value: "keyboardenter"
```
trask, line 54:
```suggestion
            - id: keyboard.enter
              value: "keyboard.enter"
```

model/browser/events.yaml
```diff
@@ -0,0 +1,83 @@
...
+              stability: development
+              brief: An element is entered via keyboard by a user.
+            - id: keyboardspace
+              value: "keyboardspace"
```
trask, line 58: just curious, why is space special compared to other characters?

does `keyboard.enter` not include space?

```suggestion
            - id: keyboard.space
              value: "keyboard.space"
```

model/browser/events.yaml
trask: it looks like this was resolved with this update:

> Click coordinates(in pixels) relative to the entire document with format

model/browser/events.yaml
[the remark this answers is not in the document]
trask: I think ok to resolve this

model/browser/events.yaml
Karlie-777: Updated, Thanks Trask!

model/browser/events.yaml
Karlie-777: Updated, Thanks Trask!

model/browser/events.yaml
trask: @Karlie-777 @breedx-splk it looks like there are still several browser centric definitions in this PR.

I think it will need a good amount of help from the Android folks to generalize this.

Otherwise I'd recommend landing it as a browser event and consider generalizing it as a follow-up.

As @MSNev says

> We can always deprecate this event in favor of any stable standard.

model/browser/events.yaml
trask: resolved now

model/browser/events.yaml
breedx-splk: > Otherwise I'd recommend landing it as a browser event and consider generalizing it as a follow-up.

I'm into this. And because I opened this thread I'm happy to resolve it...I don't think it should block.

.chloggen/browser-event.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: browser
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add browser and enduser events.
```
trask, line 13:
```suggestion
note: Add `browser.user_action` event.
```

trask (changes requested):
blocking temporarily to discuss whether we can unify screen coordinate representation with #2070

I've added this topic to next week's client instrumentation SIG meeting agenda

model/browser/events.yaml
```diff
@@ -0,0 +1,83 @@
...
+              value: "other"
+              stability: development
+              brief: User actions that are not listed above.
+          stability: development
+          brief: >
+            Type of interaction.
+            See enum [here](https://github.com/microsoft/ApplicationInsights-JS/blob/main/extensions/applicationinsights-clickanalytics-js/src/Enums.ts) for potential values we could add support for.
+          requirement_level: required
+          examples: ["click.right"]
+        - id: click_coordinates
```
trask, line 71:
```suggestion
        - id: click_xy
```

lmolkova: same as https://github.com/open-telemetry/semantic-conventions/pull/2070/files#r2054998523 - is there an explanation, pro/con analysis on this choice?

.chloggen/browser-event.yaml
Karlie-777: updated, and changelog re-uses the one from web-vital PR

model/browser/events.yaml
```diff
@@ -52,3 +52,92 @@ groups:
...
+      stability: development
+      type: map
+      fields:
+        - id: element
+          type: string
+          stability: development
+          brief: Target element tag name and it is obtained via `event.target.tagName`.
+          requirement_level: recommended
+          examples: ["button"]
+        - id: xpath
```
trask, line 73: `xpath`, `element` and `tags` feel browser specific, while the event name `enduser.user_action` makes it sound generic

Karlie-777: updated `enduser.user_action` to `browser.user_action`

model/browser/events.yaml
```diff
@@ -52,3 +52,92 @@ groups:
...
+            - 10
+        - id: click.y
+          type: int
+          stability: development
+          brief: Click y(vertical) coordinates(in pixels) relative to the entire document.
+          requirement_level: recommended
+          examples:
+            - 10
+        - id: tags
+          type: string
```
trask, line 139: this sounds like an array?

```suggestion
          type: string[]
```

model/browser/events.yaml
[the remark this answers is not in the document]
trask: @Karlie-777 it soulds like there's agreement on renaming this from `user_action_type` to `type`

model/browser/events.yaml
```diff
@@ -52,3 +52,92 @@ groups:
...
+    name: browser.user_action
+    brief: >
+      This event describes actions performed by the user such as click, scroll, zoom, resize, etc.
+    body:
+      id: browser.user_action
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: element
```
martinkuba, line 67: I wonder if we should include "name" here to be clear what this field represents (the name of the element). And if we do that, should we just be consistent with the API and call it "tagName" instead?

model/browser/events.yaml
```diff
@@ -52,3 +52,92 @@ groups:
...
+          type: string
+          stability: development
+          brief: Target element tag name and it is obtained via `event.target.tagName`.
+          requirement_level: recommended
+          examples: ["button"]
```
martinkuba, line 72: The browser API will provide this capitalized for HTML documents.

model/browser/events.yaml
```diff
@@ -52,3 +52,92 @@ groups:
...
+              value: "other"
+              stability: development
+              brief: User actions that are not listed above.
+          stability: development
+          brief: >
+            Type of interaction.
+            See enum [here](https://github.com/microsoft/ApplicationInsights-JS/blob/main/extensions/applicationinsights-clickanalytics-js/src/Enums.ts) for potential values we could add support for.
+          requirement_level: required
+          examples: ["click.right"]
+        - id: click.x
```
martinkuba, line 124: I think the name click.* might be ambiguous and too specific at the same time. The browser API provides clientX/clientY, pageX/pageY, and screenX/screenY (see [here](https://developer.mozilla.org/en-US/docs/Web/CSS/CSSOM_view/Coordinate_systems) for more details). The description here says "relative to the entire document", so I think click.x/click.y is intended to represent pageX/pageY. But given that there are multiple coordinate systems, shouldn't we be consistent and use the name that describes which coordinates are used?

Also, touch events have coordinates as well (also represented by pageX/pageY), so then these attributes could be used for both mouse and touch events.

model/browser/events.yaml
```diff
@@ -52,3 +52,93 @@ groups:
...
+        - id: tagName
+          type: string
+          stability: development
+          brief: Target element tag name and it is obtained via `event.target.tagName`.
+          requirement_level: recommended
+          examples: ["BUTTON"]
+        - id: xpath
+          type: string
+          stability: development
+          brief: Target element xpath, see [xpath](https://www.w3.org/TR/xpath/).
```
trask: reference browser API that can be used to capture this

model/browser/events.yaml
```diff
@@ -52,3 +52,93 @@ groups:
...
+    name: browser.user_action
+    brief: >
+      This event describes actions performed by the user such as click, scroll, zoom, resize, etc.
+    body:
+      id: browser.user_action
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: tagName
```
trask:
```suggestion
        - id: tag_name
```

model/browser/events.yaml
trask: discussed in SIG meeting, let's lean into browser terminology, e.g.

- `page.x`
- `page.y`

model/browser/events.yaml
```diff
@@ -52,3 +52,95 @@ groups:
...
+              stability: development
+              brief: An element is zoomed by a user.
+            - id: resize
+              value: "resize"
+              stability: development
+              brief: An element is resized by a user.
+            - id: keyboard.enter
+              value: "keyboard.enter"
+              stability: development
+              brief: An element is entered via keyboard by a user.
```
david-luna, line 111: This description is not clear to me. is that similar to `keyboard.space` below? so we track when user hits enter while an element (probably an input) has the focus?

If so I wonder why we want to differentiate between these two keys.

model/browser/events.yaml
```diff
@@ -52,3 +52,95 @@ groups:
...
+            - id: keyboard.space
+              value: "keyboard.space"
+              stability: development
+              brief: A space is entered via keyboard by a user.
+            - id: other
```
david-luna, line 116: another interesting event IMHO is [input](https://developer.mozilla.org/en-US/docs/Web/API/Element/input_event)

model/browser/events.yaml
```diff
@@ -52,3 +52,95 @@ groups:
...
+          requirement_level: recommended
+          examples:
+            - 10
+        - id: page.y
+          type: int
+          stability: development
+          brief: Click y(vertical) coordinates(in pixels) relative to the entire document.
+          requirement_level: recommended
+          examples:
+            - 10
```
thompson-tomo, line 139: Could we use the `app.screen.coordinate.*` namespace  ?

model/browser/events.yaml
```diff
@@ -52,3 +52,95 @@ groups:
...
+        - id: tags
+          type: string[]
+          stability: development
+          brief: Grab data from data-otel-* attributes in tree.
+          requirement_level: recommended
+          examples:
+            - ["id", "name"]
```
thompson-tomo, line 146: Attributes should be set as attributes on the event and don't think that these need to be defined here.

model/browser/events.yaml
```diff
@@ -52,3 +52,95 @@ groups:
...
+            - id: other
+              value: "other"
+              stability: development
+              brief: User actions that are not listed above.
+          stability: development
+          brief: >
+            Type of interaction.
+            See enum [here](https://github.com/microsoft/ApplicationInsights-JS/blob/main/extensions/applicationinsights-clickanalytics-js/src/Enums.ts) for potential values we could add support for.
+          requirement_level: required
+          examples: ["click.right"]
```
thompson-tomo, line 125: Feel that these should be attributes.

model/browser/events.yaml
```diff
@@ -52,3 +52,95 @@ groups:
...
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: tag_name
+          type: string
+          stability: development
+          brief: Target element tag name and it is obtained via `event.target.tagName`.
+          requirement_level: recommended
+          examples: ["BUTTON"]
```
thompson-tomo, line 72: Feel that this should be an attribute

model/browser/events.yaml
```diff
@@ -52,3 +52,95 @@ groups:
           brief: A unique ID representing this particular metric instance.
           requirement_level: required
           examples: ["v3-1677874579383-6381583661209"]
+  - id: event.browser.user_action
```
thompson-tomo, line 55: Would like to see the id/name be related to the group they come from with the list defined at: https://www.w3schools.com/jsref/obj_events.asp that way we can have a more tailored body/attributes.

thompson-tomo (changes requested):

lmolkova:
What's the status of this PR, is this ready for the general SemConv review?

model/browser/events.yaml
Karlie-777: https://github.com/open-telemetry/semantic-conventions/pull/1941#discussion_r2138377069 and https://github.com/open-telemetry/semantic-conventions/pull/2070 should already cover the app click events

joaquin-diaz:
Since @Karlie-777 is not actively working on this I am continuing her work here https://github.com/open-telemetry/semantic-conventions/pull/2992 (I couldn't push to the same branch unfortunately). I am already working on the instrumentation side and we want to keep things moving 

@thompson-tomo Could you take another look there? For starters I moved all the body fields to attributes, could you replicate any other comments you have here there so we can continue the conversation? 

@david-luna I see you also have a pending comment, I replaced the `keyboard.space` and `keyboard.enter` for `input` as it wasn't very clear to me what was the intention for those either. Input should cover them and the rest of the keys

@MSNev and @breedx-splk You both approved this PR, could you please take another look at the new PR? 

Thanks everyone!

tedsuo:
Closing this in favor of #2992

PR #1942, opened 2025-02-26
Add Browser Exception Event

Karlie-777:
Fixes #

## Changes


model/browser/events.yaml
```diff
@@ -0,0 +1,54 @@
...
+          requirement_level: recommended
+          examples: ["Can't convert 'int' object to str implicitly"]
+        - id: stacktrace
+          type: string
+          stability: development
+          brief: >
+            A stacktrace as a string in the natural representation for the language runtime.
+            The representation is to be determined and documented by each language SIG.
+          requirement_level: recommended
+          examples: ["Exception in thread main java.lang.RuntimeException:
```
Karlie-777, line 45: need to reformat the examples to make the generated table look nicer

model/browser/events.yaml
```diff
@@ -0,0 +1,54 @@
+groups:
+  - id: event.browser.exception
+    stability: development
+    type: event
+    name: browser.exception
+    brief: >
+      This event describes an error or exception that occurs in a browser application.
+    body:
```
breedx-splk, line 8: I would like to see a comment here to clarify that this is temporary or that this will actually be an attribute. That way readers do not expect to find content in the actual body.

model/browser/events.yaml
```diff
@@ -0,0 +1,54 @@
+groups:
+  - id: event.browser.exception
```
breedx-splk, line 2: Similar to comments I left on #1941, I think other client instrumentation would benefit from also having these relaxed to be non-browser. Alternatives: `client.exception`, `app.exception`, `runtime.exception`.....

MSNev: @Karlie-777 can you please review https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2751 which is an initial implementation of a browser "event" which is using the existing top-level 3 attributes.

model/browser/events.yaml
```diff
@@ -0,0 +1,54 @@
...
+    stability: development
+    type: event
+    name: browser.exception
+    brief: >
+      This event describes an error or exception that occurs in a browser application.
```
MSNev, line 7: Lets also "add" a top-level attribute `url.full` (redacted) or include a cutdown `url` within the data), as while having the "filename" is useful, we should also be including the "hosting" page as the source file (bundle) may throw at different locations and for different reasons when used from different pages / scenarios.

trask (commented):
can we reuse the general exception event: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-logs.md

we could add `code.*` attributes as opt-in on the general exception event

and if you want them to be "recommended" specifically for browser, we can probably figure out how to do that

do you think other browser-specific attributes would be added to this event? if that's the case, then may make sense to keep the "browser" event name, but would still be good to "extend" the existing general exception event, and just layer in the additional attributes

MSNev:
> can we reuse the general exception event? https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-logs.md
> 
> we could add `code.*` attributes as opt-in on the general exception event
> 
> and if you want them to be "recommended" specifically for browser, we can probably figure out how to do that
> 
> do you think other browser-specific attributes would be added to this event? if that's the case, then may make sense to keep the "browser" event name, but would still be good to "extend" the existing general exception event, and just layer in the additional attributes

Yeah, this is essentially what I'm saying also as we have had "exceptions" for spans and logs like for ages now, if we want to define a "standard" exception "event" (basically just give the Logs exception and event_name) then that would work, as long as we can add some required additional browser specific fields (I want to avoid having the discussions about "which" exception event should we create).

One issue (which I think already exists to some extent today for different languages), is that for "browsers" specifically, there is no common "standard" of the "stacktrace", so different browsers (including between versions of browser) "present" the same stack trace with different formatting which makes it extremely problematic for backends to process and group the same basic common failure (vs a browser / runtime specific failure (like IE not liking missing comma's for example)), as such it is common (at least for the Microsoft JS Sdk's) to pass back the "raw" as well as a "normalized" (parsed) stack to pull out the common fields of a stack into their components (filename, line, col, source bundle) as the JS runtime often has more context about what the current runtime is (Opera, Chromium, Firefox, iOS, specific versions, as well as some runtimes that "act" (provide) like browser electron, etc).
* So while this specific issue is not addressed in this PR the ability to "add" some common "additional" (known) attributes to a standard exception is going to be required whether we (preferably) have a common exception event or not

Karlie-777:
we already have logs exceptions, will reuse it, will close this one. (need to add the additional features to log exceptions)

lmolkova:
closing this PR based on the https://github.com/open-telemetry/semantic-conventions/pull/1942#issuecomment-2744186490

PR #1940, opened 2025-02-26
Add Browser Web Vitals Event 

Karlie-777:
Fixes #

## Changes

Please provide a brief description of the changes here.

Note: if the PR is touching an area that is not listed in the [existing areas](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/README.md), or the area does not have sufficient [domain experts coverage](https://github.com/open-telemetry/semantic-conventions/blob/main/.github/CODEOWNERS), the PR might be tagged as [experts needed](https://github.com/open-telemetry/semantic-conventions/labels/experts%20needed) and move slowly until experts are identified.

## Merge requirement checklist

* [ ] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [ ] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

PR #1943, opened 2025-02-26
Add Browser Resource Timing Event 

Karlie-777:
Fixes #

## Changes


model/browser/events.yaml
```diff
@@ -0,0 +1,104 @@
...
+          stability: development
+          brief: >
+            The time immediately after the browser finishes the domain name lookup for the resource,
+            and it is measured in milliseconds.
+          requirement_level: recommended
+          examples: [100.6000000014906]
+        - id: connectStart
+          type: double
+          stability: development
+          brief: >
```
Karlie-777, line 55: need to reformat/shorten the brief to make the generated table look nicer

model/browser/events.yaml
```diff
@@ -0,0 +1,104 @@
...
+      This event describes the timing metrics as provided by PerformanceResourceTiming Performance API.
+      These metrics are related to fetching a resource, such as
+      XMLHttpRequest, Fetch, sendBeacon APIs, SVG, image or script.
+    note: >
+      This event captures data from the [ResourceTiming](https://www.w3.org/TR/resource-timing/).
+      It is recommended to be captured by using the PerformanceResourceTiming API.
+      If the page starts unloading before the page loads,
+      then the partial data can be captured by calling the [performance.getEntriesByType](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType) method.
+
+    body:
```
breedx-splk, line 16: I would like to see a comment here to clarify that this is temporary or that this will actually be an attribute. That way readers do not expect to find content in the actual body.

model/browser/events.yaml
```diff
@@ -0,0 +1,104 @@
...
+      id: browser.resource_timing
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: name
```
breedx-splk, line 22: I don't think `name` is the right thing for a url here. Maybe just `url` would be more concise?

MSNev: Yeah, now that `url.full` is stable maybe we should consider lifting this to a top level attribute, where from an instrumentation perspective it would include the full redacted query string -- although for this event you "generally" don't need the full url, just the scheme, domain and path -- but splitting out to 3 different attributes doesn't really fit the need of this event....

Lets change this to `url` and add wording that it explicitly doesn't include the query string or any leading credentials (so it's formatted as `<scheme>:\\<full domain>\<path>` only
What I mean by this is that
`https://username:password@www.example.com/path?a=b&c=d` would be sent as `https://www.example.com/path` basically both removing any potential PII and reducing the cardinality of the name.

Karlie-777: and change it to required

Karlie-777: Make sure the Note is present:  it explicitly doesn't include the query string or any leading credentials (so it's formatted as <scheme>:\\<full domain>\<path> only https://username:password@www.example.com/path?a=b&c=d would be sent as https://www.example.com/path basically both removing any potential PII and reducing the cardinality of the name. (auto-reconstruct url instead of actual full url)

model/browser/events.yaml
```diff
@@ -0,0 +1,104 @@
...
+            The time immediately after the browser finishes the domain name lookup for the resource,
+            and it is measured in milliseconds.
+          requirement_level: recommended
+          examples: [100.6000000014906]
+        - id: connectStart
+          type: double
+          stability: development
+          brief: >
+             The time immediately immediately before the user agent starts establishing the connection to the server to retrieve the resource,
+             and it is measured in milliseconds.
```
trask, line 57:
```suggestion
             The time immediately before the user agent starts establishing the connection to the server to retrieve the resource, in milliseconds.
```

model/browser/events.yaml
```diff
@@ -0,0 +1,104 @@
...
+          examples: [100.6000000014901]
+        - id: domainLookupEnd
+          type: double
+          stability: development
+          brief: >
+            The time immediately after the browser finishes the domain name lookup for the resource,
+            and it is measured in milliseconds.
+          requirement_level: recommended
+          examples: [100.6000000014906]
+        - id: connectStart
```
trask, line 52: let's add links to external definition in notes for all attributes, e.g. https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/connectStart

model/browser/events.yaml
```diff
@@ -0,0 +1,104 @@
...
+             The time immediately immediately before the user agent starts establishing the connection to the server to retrieve the resource,
+             and it is measured in milliseconds.
+          requirement_level: recommended
+          examples: [476.6000000014901]
+        - id: secureConnectionStart
+          type: double
+          stability: development
+          brief: >
+            The time immediately before immediately before the browser starts the handshake process to secure the current connection,
+            and it is measured in milliseconds.
```
trask, line 65: (same for others)
```suggestion
            The time immediately before immediately before the browser starts the handshake process to secure the current connection, in milliseconds.
```

model/browser/events.yaml
trask: blocked by #1980

model/browser/events.yaml
```diff
@@ -0,0 +1,104 @@
...
+    note: >
+      This event captures data from the [ResourceTiming](https://www.w3.org/TR/resource-timing/).
+      It is recommended to be captured by using the PerformanceResourceTiming API.
+      If the page starts unloading before the page loads,
+      then the partial data can be captured by calling the [performance.getEntriesByType](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType) method.
+    attributes:
+      - ref: url.full
+        requirement_level: required
+        note: >
+          Generally the url doesn't include any query strings or any leading credentials,
```
trask, line 19: @open-telemetry/specs-semconv-maintainers do you see any issue with using `url.full` for something that doesn't include the query string?

ladd:
Would it make sense to use integer nanosecond timestamps to be consistent timestamp/duration measurement in the rest of [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L191) 

Looking beyond the browser, these are potentially applicable to a native app scenario via vendor APIs like [URLSessionTaskTransactionMetrics](https://developer.apple.com/documentation/foundation/urlsessiontasktransactionmetrics) on iOS.

model/browser/events.yaml
Karlie-777: Updated, changes are
1. modified and shortened the brief
2. added deep links
3. changed measurement units from milliseconds to seconds

model/browser/events.yaml
Karlie-777: added added deep links to each field

model/browser/events.yaml
Karlie-777: this is covered by previous comments, so resolved this one.

model/browser/events.yaml
Karlie-777: this is covered by previous comments, so resolved this one

model/browser/events.yaml
Karlie-777: to be discussed: (mentioned by ladd, see details below):
1. measurement units
2. should we consider other potential metrics?

model/browser/events.yaml
```diff
@@ -0,0 +1,100 @@
...
+      id: browser.resource_timing
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: fetchStart
+          type: double
+          stability: development
+          brief: >
+            The time immediately before the browser starts to fetch the resource, in seconds.
```
trask, line 32: I'd suggest just copying in the mozilla wording (and keeping it in milliseconds since that's how the underlying browser event is defined)
```suggestion
            A [timestamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) immediately before the browser starts to fetch the resource.
```

trask:
> Would it make sense to use integer nanosecond timestamps to be consistent timestamp/duration measurement in the rest of [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L191)

hey @ladd 👋

I think this event is a bit different because we're bridging an external event as-is, and so we are prioritizing preserving the external event structure/names/values over consistency with normal semantic conventions

> Looking beyond the browser, these are potentially applicable to a native app scenario via vendor APIs like [URLSessionTaskTransactionMetrics](https://developer.apple.com/documentation/foundation/urlsessiontasktransactionmetrics) on iOS.

@open-telemetry/android-approvers wdyt?

@MSNev @martinkuba how important is it for this event to exactly match the browser event? or would it be ok as long as you have the ability to capture the same set of data, but it was modeled as a generic event with some potentially browser specific attributes on top of it?

model/browser/events.yaml
trask: resolving, but we can continue some of these discussions below https://github.com/open-telemetry/semantic-conventions/pull/1943#issuecomment-2762382976

model/browser/events.yaml
MSNev: Part of the problem with calling these "timestamps" (even though the standard calls these "DOMHighResTimeStamp") is that in practice that are not actually timestamps for the resources. they are a recorded number of milliseconds.

As per the spec, so these values fall into "the difference in time between two discrete points in time", where the timeOrigin can (and often is) zero.

> This type can be used to describe a discrete point in time or a time interval (the difference in time between two discrete points in time). The starting time can be either a specific time determined by the script for a site or app, or the [time origin](https://developer.mozilla.org/en-US/docs/Web/API/Performance/timeOrigin).

So I would prefer if this was either kept as `time` as it more accurately represents what it is (a duration of elapsed time not a specific timestamp)

trask: @MSNev so this?

```
The [time](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) immediately before the browser starts to fetch the resource.
```

Karlie-777:
For the changelog, it will reuse and update the one from this PR https://github.com/open-telemetry/semantic-conventions/pull/1940

MSNev:
> @MSNev @martinkuba how important is it for this event to exactly match the browser event? or would it be ok as long as you have the ability to capture the same set of data, but it was modeled as a generic event with some potentially browser specific attributes on top of it?

## The case for keeping as is

As this is the `browser.resource_timing` event, it is very important that this be kept the way the industry and existing solutions (incl OpenTelemetry span events) record and measure these values, otherwise anyone "moving" to this new definition would have to perform backend calculations on every query or on receiving the event when processing this data.

Even in OpenTelemetry there is existing code (hacks) to capture these currently as "Span" events https://github.com/open-telemetry/opentelemetry-js/blob/b322a24b4a3296577fb908c0bc9b33efc57c501e/packages/opentelemetry-sdk-trace-web/src/utils.ts#L63-L127 which record the values as is (which for most browsers will be a zero based "duration" (which is also why the hack to send as a span event (long before logs existed))

## The case for Making Consistent

Now if this was (or becomes) a non-browser specific event definition and a general OpenTelemetry Semantic convention for all consumers of different types of "resource" types, then I would 100% agree with aligning and recording as nanoseconds. As a differently named event.

## General Background Details

As a general comment, in the browser world nothing is generally recorded in nanoseconds and when it is supported (as a double - as fractional microseconds) it is generally deliberately crippled (by adding random time "adjustments" with both positive and negative values) to avoid being used to "fingerprint" (identify / classify) a specific user / environment. These "adjustments" where added because this was occurring in the real world (with techniques like Canvas fingerprinting)

trask:
> For the changelog, it will reuse and update the one from this PR #1940

it's better to create a change log entry (file) for each PR

trask:
thanks

similar to https://github.com/open-telemetry/semantic-conventions/pull/1941#discussion_r2019321206, let's go ahead and land this browser-specific event initially, and explore generalizing it as follow-up work in the client instrumentation SIG

we can deprecate the browser-specific event later if/when a generalized version is landed

model/browser/events.yaml
```diff
@@ -0,0 +1,103 @@
...
+      id: browser.resource_timing
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: fetchStart
+          type: double
+          stability: development
+          brief: >
+            The [timestamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) immediately before
```
trask, line 32: from @MSNev's suggestion (and need to make the same change elsewhere)
```suggestion
            The [time](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) immediately before
```

model/browser/events.yaml
trask: unblocked/resolved now

model/browser/events.yaml
Karlie-777: updated! Thanks Trask!

MSNev:
> @MSNev @martinkuba how important is it for this event to exactly match the browser event? or would it be ok as long as you have the ability to capture the same set of data, but it was modeled as a generic event with some potentially browser specific attributes on top of it?

Very important, as anyone upgrading from something other than OTel would end up with a 1000x time difference, and the browser really isn't going to give you this level of reliable precession. ie. it should match the W3C definition

model/browser/events.yaml
```diff
@@ -52,3 +52,110 @@ groups:
...
+      This event describes the timing metrics as provided by PerformanceResourceTiming Performance API.
+      These metrics are related to fetching a resource, such as
+      XMLHttpRequest, Fetch, sendBeacon APIs, SVG, image or script.
+    note: >
+      This event captures data from the [ResourceTiming](https://www.w3.org/TR/resource-timing/).
+      It is recommended to be captured by using the PerformanceResourceTiming API.
+      If the page starts unloading before the page loads,
+      then the partial data can be captured by calling the [performance.getEntriesByType](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType) method.
+    attributes:
+      - ref: url.full
```
martinkuba, line 69: I think this URL is for the resource described by this event, not the page URL, correct? I think we should note this in the description of this attribute, so that it is clear.

Also, if this is the resource URL, why is it listed as an attribute rather than a body field like the measurements?

martinkuba:
All fields that are available from the browser API are described here
https://www.w3.org/TR/resource-timing/#sec-performanceresourcetiming

This event includes only a subset. I recall we discussed including only the most relevant fields a long time ago in the Client SIG. Now, after another review, I think there are additional fields we might want to consider including, for example:

- transferSize (to determine whether local cache was used)
- decodedBodySize, encodedBodySize (to determine whether the content was compressed)
- nextHopProtocol (to determine if a fast protocol was used)
- additional context: initiatorType, contentType, deliveryType, renderBlockingStatus

This PR has all fields marked as recommended. If we included all the fields and made them all recommended (not required), it could be left to the instrumentation to determine which to include. This could also be configurable, so that users can decide what fields they care about.

david-luna:
Could we re-open this PR?

model/browser/events.yaml
overbalance: Some platforms use querystring params to resize images. If you strip it down to just domain and path, you won't have a complete picture of which size assets loaded. I understand the high cardinality issue but wanted to call this case out.

Karlie-777:
This is PR is under open status, still making changes to this PR

model/browser/events.yaml
joaquin-diaz: Should we include query string separately?

Karlie-777:
@hectorhdzg will take over this PR and continue addressing comments after https://github.com/open-telemetry/semantic-conventions/pull/1943#issuecomment-2762784057 or he will open a new PR based on this PR.

Karlie-777:
close this PR since https://github.com/open-telemetry/semantic-conventions/pull/3069 is opened for the following changes. Thanks @hectorhdzg

PR #1944, opened 2025-02-27
clarify code function name

### LINKED ISSUE #1677: Clarify values for `code.function.name` and `code.namespace`
This is part of #1599 where we aim to make `code.*` attributes as release-candidate.

Some of the `code.*` attributes are being renamed with #1624, but this issue is about the values of those attributes and having a clear definition for them, in particular:

- `code.function.name` (previously `code.function`).
- `code.namespace`

In the discussion of #1624 we found that not having explicit per-language examples leaves interpretation open ([here](https://github.com/open-telemetry/semantic-conventions/pull/1624#discussion_r1870985463) and [here](https://github.com/open-telemetry/semantic-conventions/pull/1624#issuecomment-2534949287)), and is likely to cause ambiguity and potential inconsistencies, also there might be per-language constraints or overhead to provide those values.

Those attributes are currently in `experimental`, and we aim to promote them to `release candidate` with #1599, as per [this comment](https://github.com/open-telemetry/semantic-conventions/pull/1624#discussion_r1881292830) we consider that they are not used enough to justify a migration plan to prevent unexpected breaking changes (for example with `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable).

Here the goal would be to define for each language:
- the value of `code.function.name`
- the value of `code.namespace`
- what value to use when only a single value is provided by the language/platform, in other words when to split and how
- what value to use when only a partial value is available, for example anonymous lambdas/functions that might not have an explicit name in code (but likely have a "technical name").

For now, we aim to maximize consistency, however if the overhead to provide those values (for example due to extra allocation or string splitting overhead), it might be possible to provide "slightly inconsistent values" to avoid those.

## Checklist
- [ ] Java [comment](https://github.com/open-telemetry/semantic-conventions/issues/1677#issuecomment-2569081934) 
- [ ] Go [comment](https://github.com/open-telemetry/semantic-conventions/pull/1624#issuecomment-2534949287) for lambda, single value would be preferable but [could be split with minimal overhead](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6415).
- [ ] PHP [comment](https://github.com/open-telemetry/semantic-conventions/issues/1677#issuecomment-2576776269) and [comment](https://github.com/open-telemetry/semantic-conventions/issues/1677#issuecomment-2577545831). single value would be preferable/simpler but could be split with minimal overhead.
- [ ] Python [comment](https://github.com/open-telemetry/semantic-conventions/issues/1677#issuecomment-2575581985)
- [ ] Node/Javascript [comment](https://github.com/open-telemetry/semantic-conventions/issues/1677#issuecomment-2576797137)
- [ ] Erlang/Elixir [comment](https://github.com/open-telemetry/semantic-conventions/issues/1677#issuecomment-2576792921)
- [ ] Rust [comment](https://github.com/open-telemetry/semantic-conventions/issues/1677#issuecomment-2585650950)
- [ ] Profiling [comment](https://github.com/open-telemetry/semantic-conventions/issues/1677#issuecomment-2590986638)
- [ ] ...

SylvainJuge:
Fixes #1677

## Changes

- modify the definition of `code.function.name` to include fully qualified name
- deprecate `code.namespace` as the information is included in `code.function.name`.

This is a breaking change for which any producer/consumer of `code.function.name`, `code.function` (deprecated) or `code.namespace` is impacted. The changes are however expected to be performed once all attributes in `code.*` namespace have been made stable.

Because the change is not a simple rename, we can't use the schema transformation, even if we had a way to perform complex transformations made possible by https://github.com/open-telemetry/opentelemetry-specification/issues/4427, the transformation would have to be platform-specific and we might not have the ability to do that by relying only on the available data.


internal/tools/go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/open-telemetry/opentelemetry-specification/internal/tools
 
-go 1.22.0
+go 1.23.0
+
 toolchain go1.23.4
```
pellared, line 5: Nit: if you're touching go.mod I think it would be good to remove the toolchain directive

pellared (approved):
👍

internal/tools/go.mod
SylvainJuge: I had to change this just to make CI and checks work locally, this is definitely outside of my expertise area so maybe this change is not required nor relevant.

model/code/registry.yaml
```diff
@@ -9,15 +9,17 @@ groups:
...
+          runtime, which is also likely the same used within `code.stacktrace` attribute value.
+        examples: >
+          com.example.MyHttpService.serveRequest (Java)
+          com.mycompany.Main$1.myMethod (Java anynymous class)
+          com.mycompany.Main$$Lambda/0x0000748ae4149c00.myMethod (Java lambda)
+          GuzzleHttp\Client::transfer (PHP)
+          github.com/my/repo/pkg.foo.func5 (Go)
+          OpenTelemetry.Ctx.new (Elixir)
+          opentelemetry_ctx:new (Erlang)
+          playground::my_module::my_cool_func (Rust)
```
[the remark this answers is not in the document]
SylvainJuge: I've added one in 8837a93a, if you have other examples like C++ or C# it could also be welcome additions.

kamphaus (approved):

internal/tools/go.mod
adrielp: This usually is automatically added by go when one of the underlying dependencies pins their required version of go.

model/code/registry.yaml
```diff
@@ -9,15 +9,18 @@ groups:
...
-      - id: code.namespace
-        type: string
-        stability: development
-        brief: >
-          The "namespace" within which `code.function.name` is defined. Usually the qualified class or module name,
-          such that `code.namespace` + some separator + `code.function.name` form a unique identifier for the code unit.
-        examples: com.example.MyHttpService
+          The method or function fully-qualified name without arguments. The value should fit the natural representation of the language
+          runtime, which is also likely the same used within `code.stacktrace` attribute value.
+        examples: >
```
trask, line 14: I think examples is supposed to be an array, maybe include just a couple of short-ish examples, and move the complete set of examples to the notes section?

model/code/registry-deprecated.yaml
```diff
@@ -32,3 +33,11 @@ groups:
         brief: >
           Deprecated, use `code.column.number`
         examples: 16
+      - id: code.namespace
+        type: string
+        stability: development
+        deprecated: Value should be included in `code.function.name` which is expected to be a fully-qualified name.
+        brief: >
+          The "namespace" within which `code.function.name` is defined. Usually the qualified class or module name,
+          such that `code.namespace` + some separator + `code.function.name` form a unique identifier for the code unit.
```
lmolkova, line 42: The brief is no longer applicable, plus we usually change brief when deprecating attributes, so maybe

```suggestion
          Deprecated, namespace is now included into `code.function.name`
```

model/code/registry.yaml
SylvainJuge: Done in 467ce34

PR #1951, opened 2025-02-27
Change the definition of `device.id`

### LINKED ISSUE #1874: `device.id` vs `installation.id` on Android and iOS

bencehornak:
Fixes #1874

> [!WARNING]
> This PR must be merged together with #1897 as they are highly related.

## Changes

I have given a new definition to device.id, which now truly identifies the device. It has been marked as opt-in though, as its usage is generally discouraged due to privacy implications, and should only be enabled in special cases, when it does not violate the privacy of the user.

The old definition of `device.id` is more-or-less covered by the proposed `app.installation.id` in #1897, which is recommended for general use instead of this, as it provides more privacy to the users.

### DIFF
--- a/docs/attributes-registry/device.md
+++ b/docs/attributes-registry/device.md
@@ -9,12 +9,24 @@ Describes device attributes.
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
-| <a id="device-id" href="#device-id">`device.id`</a> | string | A unique identifier representing the device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="device-id" href="#device-id">`device.id`</a> | string | A unique identifier representing the device [1] | `123456789012345`; `01:23:45:67:89:AB` | ![Development](https://img.shields.io/badge/-development-blue) |
 | <a id="device-manufacturer" href="#device-manufacturer">`device.manufacturer`</a> | string | The name of the device manufacturer [2] | `Apple`; `Samsung` | ![Development](https://img.shields.io/badge/-development-blue) |
 | <a id="device-model-identifier" href="#device-model-identifier">`device.model.identifier`</a> | string | The model identifier for the device [3] | `iPhone3,4`; `SM-G920F` | ![Development](https://img.shields.io/badge/-development-blue) |
 | <a id="device-model-name" href="#device-model-name">`device.model.name`</a> | string | The marketing name for the device model [4] | `iPhone 6s Plus`; `Samsung Galaxy S6` | ![Development](https://img.shields.io/badge/-development-blue) |
 
-**[1] `device.id`:** The device identifier MUST only be defined using the values outlined below. This value is not an advertising identifier and MUST NOT be used as such. On iOS (Swift or Objective-C), this value MUST be equal to the [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor). On Android (Java or Kotlin), this value MUST be equal to the Firebase Installation ID or a globally unique UUID which is persisted across sessions in your application. More information can be found [here](https://developer.android.com/training/articles/user-data-ids) on best practices and exact implementation details. Caution should be taken when storing personal data or anything which can identify a user. GDPR and data protection laws may apply, ensure you do your own due diligence.
+**[1] `device.id`:** **The `device.id` SHOULD NOT be used in most user-facing applications due to privacy regulations.
+Consequently, instrumentations that provide it MUST provide it as an opt-in feature.**
+
+Its value SHOULD be identical for all apps on a device and it SHOULD NOT change if an app is uninstalled and re-installed.
+However, it might be resettable by the user for all apps on a device.
+Hardware IDs (e.g. vendor-specific serial number, IMEI or MAC address) MIGHT be used as values.
+
+More information about Android identifier best practices can be found [here](https://developer.android.com/training/articles/user-data-ids).
+
+> [!WARNING]
+>
+> This attribute contains sensitive (PII) information. Caution should be taken when storing personal data or anything which can identify a user. GDPR and data protection laws may apply, ensure you do your own due diligence.
+> See [`installation.id`](/docs/attributes-registry/installation.md#installation-id) as a more privacy-preserving alternative.
 
 **[2] `device.manufacturer`:** The Android OS provides this field via [Build](https://developer.android.com/reference/android/os/Build#MANUFACTURER). iOS apps SHOULD hardcode the value `Apple`.
 
--- a/docs/resource/device.md
+++ b/docs/resource/device.md
@@ -16,18 +16,30 @@
 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
-| [`device.id`](/docs/attributes-registry/device.md) | string | A unique identifier representing the device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`device.manufacturer`](/docs/attributes-registry/device.md) | string | The name of the device manufacturer [2] | `Apple`; `Samsung` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`device.model.identifier`](/docs/attributes-registry/device.md) | string | The model identifier for the device [3] | `iPhone3,4`; `SM-G920F` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`device.model.name`](/docs/attributes-registry/device.md) | string | The marketing name for the device model [4] | `iPhone 6s Plus`; `Samsung Galaxy S6` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`device.manufacturer`](/docs/attributes-registry/device.md) | string | The name of the device manufacturer [1] | `Apple`; `Samsung` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`device.model.identifier`](/docs/attributes-registry/device.md) | string | The model identifier for the device [2] | `iPhone3,4`; `SM-G920F` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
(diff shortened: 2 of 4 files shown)


docs/resource/device.md
```diff
@@ -16,18 +16,31 @@
...
+However, it might be resettable by the user for all apps on a device.
+Hardware IDs (e.g. vendor-specific serial number, IMEI or MAC address) MIGHT be used as values.
+
+More information about Android identifier best practices can be found [here](https://developer.android.com/training/articles/user-data-ids).
+
+> [!WARNING]
+>
+> This attribute contains sensitive (PII) information. Caution should be taken when storing personal data or anything which can identify a user. GDPR and data protection laws may apply,
+> ensure you do your own due diligence.
+> See [`app.installation.id`](/docs/attributes-registry/app.md#app-installation-id) as a more privacy-preserving alternative.
```
bencehornak, line 43: The CI fails at this step, because this file will only be available after #1897 is merged.

trask:
cc @open-telemetry/semconv-mobile-approvers

model/device/registry.yaml
```diff
@@ -3,23 +3,31 @@ groups:
...
-          The device identifier MUST only be defined using the values outlined below. This value is not an advertising
-          identifier and MUST NOT be used as such.
-          On iOS (Swift or Objective-C), this value MUST be equal to the [vendor identifier](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor).
-          On Android (Java or Kotlin), this value MUST be equal to the Firebase Installation ID or a globally unique
-          UUID which is persisted across sessions in your application. More information can be found [here](https://developer.android.com/training/articles/user-data-ids)
-          on best practices and exact implementation details.
-          Caution should be taken when storing personal data or anything which can identify a user. GDPR and
-          data protection laws may apply, ensure you do your own due diligence.
-        examples: ['2ab2916d-a51f-4ac8-80ee-45ac31a28092']
+        note: |
```
bidetofevil, line 13: I think we should list examples where this would be appropriate to use, where privacy concerns aren't relevant. 

In addition, we should make it explicit that for commercial mobile apps, they should use `app.installation.id` instead, as folks might gravitate towards this one for simplicity.

docs/resource/device.md
```diff
@@ -16,18 +16,31 @@
...
-**[3] `device.model.identifier`:** It's recommended this value represents a machine-readable version of the model identifier rather than the market or consumer-friendly name of the device.
+**[3] `device.model.name`:** It's recommended this value represents a human-readable version of the device model rather than a machine-readable alternative.
 
-**[4] `device.model.name`:** It's recommended this value represents a human-readable version of the device model rather than a machine-readable alternative.
+**[4] `device.id`:** **The `device.id` SHOULD NOT be used in most user-facing applications due to privacy regulations.
+Consequently, instrumentations that provide it MUST provide it as an opt-in feature.**
+
+Its value SHOULD be identical for all apps on a device and it SHOULD NOT change if an app is uninstalled and re-installed.
+However, it might be resettable by the user for all apps on a device.
+Hardware IDs (e.g. vendor-specific serial number, IMEI or MAC address) MIGHT be used as values.
```
bidetofevil, line 35: Lets come up with examples that are practical where this should be used. iOS and Android devices no longer expose these for the privacy reasons you talked about in the PR, so maybe something like kiosk serial numbers that map to the hardware?

model/device/registry.yaml
bidetofevil: We can put in a sterner warning, that if `device.id` is used and it complies to the spec here, it will be rejected by Apple App Store and Google Play Store review for privacy reasons. So folks who use it probably should be hardware OEMs or other folks who don't have to go through that process,

trask:
it would be great if we can get approvals from the iOS and browser perspectives as well (if this attribute is applicable there)

docs/attributes-registry/device.md
```diff
@@ -9,12 +9,25 @@ Describes device attributes.
...
+
+Its value SHOULD be identical for all apps on a device and it SHOULD NOT change if an app is uninstalled and re-installed.
+However, it might be resettable by the user for all apps on a device.
+Hardware IDs (e.g. vendor-specific serial number, IMEI or MAC address) MIGHT be used as values.
+
+More information about Android identifier best practices can be found [here](https://developer.android.com/training/articles/user-data-ids).
+
+> [!WARNING]
+>
+> This attribute contains sensitive (PII) information. Caution should be taken when storing personal data or anything which can identify a user. GDPR and data protection laws may apply,
```
MSNev, line 28: I would change this to "This attribute may contain sensitive ..." as not runtimes will populate this with potential PII content.

docs/resource/device.md
```diff
@@ -16,18 +16,31 @@
...
 |---|---|---|---|---|---|
-| [`device.id`](/docs/attributes-registry/device.md) | string | A unique identifier representing the device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`device.manufacturer`](/docs/attributes-registry/device.md) | string | The name of the device manufacturer [2] | `Apple`; `Samsung` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`device.model.identifier`](/docs/attributes-registry/device.md) | string | The model identifier for the device [3] | `iPhone3,4`; `SM-G920F` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`device.model.name`](/docs/attributes-registry/device.md) | string | The marketing name for the device model [4] | `iPhone 6s Plus`; `Samsung Galaxy S6` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`device.manufacturer`](/docs/attributes-registry/device.md) | string | The name of the device manufacturer [1] | `Apple`; `Samsung` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
MSNev, line 19: Suggestion: As a generic "namespace" for devices, its probably also worth considering an additional `device.class` which describes the type of device in a more generic form with possible values like
- Mobile
- Desktop
- Android
- Windows
- XBox
- iOS
- Browser
- etc

Although, there may be a collection of other attributes which could be used by backends to derive some of this information.

bencehornak: Sounds reasonable to me, I would keep this suggestion to a separate PR though, if you don't mind.

docs/resource/device.md
bencehornak: rephrased the disclaimer

bidetofevil (approved):
The new description sufficiently neuters `device.id` and gives the appropriate warnings for commercial apps and SDKs to NOT user it, so this works for me.

lmolkova (approved):

model/device/registry.yaml
```diff
@@ -3,23 +3,33 @@ groups:
...
+          Its value SHOULD be identical for all apps on a device and it SHOULD NOT change if an app is uninstalled and re-installed.
+          However, it might be resettable by the user for all apps on a device.
+          Hardware IDs (e.g. vendor-specific serial number, IMEI or MAC address) MIGHT be used as values.
```
trask, line 16: (just the more common normative word used in the spec)
```suggestion
          Hardware IDs (e.g. vendor-specific serial number, IMEI or MAC address) MAY be used as values.
```

trask:
> This PR must be merged together with https://github.com/open-telemetry/semantic-conventions/pull/1897 as they are highly related, dear maintainers, please add them to the same milestone.

will get this merged soon 👍

docs/attributes-registry/device.md
```diff
@@ -9,12 +9,25 @@ Describes device attributes.
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
-| <a id="device-id" href="#device-id">`device.id`</a> | string | A unique identifier representing the device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="device-id" href="#device-id">`device.id`</a> | string | A unique identifier representing the device [1] | `123456789012345`; `01:23:45:67:89:AB` | ![Development](https://img.shields.io/badge/-development-blue) |
```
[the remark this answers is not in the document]
trask: checked w/ @MSNev and confirmed ok to mark this discussion as resolved

PR #1954, opened 2025-02-28
messaging: add celery to messaging system values

xrmx:
## Changes

Celery site here:
https://docs.celeryq.dev/en/stable/index.html

And we have an instrumentation for celery in opentelemetry-python-contrib.

### DIFF
--- a/.chloggen/add-messaging-system-celery.yaml
+++ b/.chloggen/add-messaging-system-celery.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: messaging
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `celery` to messaging system values
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [1954]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/docs/attributes-registry/messaging.md
+++ b/docs/attributes-registry/messaging.md
@@ -76,6 +76,7 @@ size should be used.
 |---|---|---|
 | `activemq` | Apache ActiveMQ | ![Development](https://img.shields.io/badge/-development-blue) |
 | `aws_sqs` | Amazon Simple Queue Service (SQS) | ![Development](https://img.shields.io/badge/-development-blue) |
+| `celery` | Celery | ![Development](https://img.shields.io/badge/-development-blue) |
 | `eventgrid` | Azure Event Grid | ![Development](https://img.shields.io/badge/-development-blue) |
 | `eventhubs` | Azure Event Hubs | ![Development](https://img.shields.io/badge/-development-blue) |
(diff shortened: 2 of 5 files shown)


lmolkova (changes requested):
Celery seems to be a background task processing library rather than a messaging system. Messaging semantic conventions might not really apply to it since they are focused on common messaging scenarios involving network calls. I would imaging celery integrations for specific systems (like RabbitMQ) to either leverage underlying instrumentation for this library or, if celery needs to be instrumented, it should set the messaging system to the actual broker used by the library.

xrmx:
> Celery seems to be a background task processing library rather than a messaging system. Messaging semantic conventions might not really apply to it since they are focused on common messaging scenarios involving network calls. I would imaging celery integrations for specific systems (like RabbitMQ) to either leverage underlying instrumentation for this library or, if celery needs to be instrumented, it should set the messaging system to the actual broker configured by the user.

We already have a celery instrumentation but we were not setting the `messaging.system` https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3265 . The original PR had `queue` as value which does not seem much informative. 
We may be able to get the broker from the client, a popular broker for celery is redis, will it fit in that list?

lmolkova:
> We already have a celery instrumentation but we were not setting the `messaging.system` attribute, see [open-telemetry/opentelemetry-python-contrib#3265](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3265) . The original PR had `queue` as value which does not seem much informative. We may be able to get the broker from the client, a popular broker for celery is redis, will it fit in that list?

yep, redis would fit. 
If you have strong reasons to believe we need celery to be in the `messaging.system`, e.g. if there are some celery-specific behaviors, or the underlying broker is not known, please let me know. I'm happy to change my opinion

shivanshuraj1333:
some thoughts https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3265#discussion_r1981102250

lmolkova:
@shivanshuraj1333 @xrmx I replied in the https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3265/files#r1981937391 - looking into existing instrumentation I don't believe it has a significant intersection with [messaging semconv](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md#messaging-attributes) - it only populates message id, conversation id and destination name from messaging, but populates a lot of celery-specific attributes.

I don't see how setting messaging.system would be helpful - backends will assume that spans follow messaging semconv, but these spans would lack the details. 

I'm suggesting to think about generic task processing (which maybe in-process) provided by celery as a layer on top of messaging semconv.

shivanshuraj1333:
what could be the attribute name for it? any suggestions?

lmolkova:
created https://github.com/open-telemetry/semantic-conventions/issues/1963 - we should document the limitations of messaging semconv. 

//cc @open-telemetry/semconv-messaging-approvers

lmolkova:
> what could be the attribute name for it? any suggestions?

could you elaborate? attribute name for what?

shivanshuraj1333:
id you run a demo application like this, you'd see attributes coming like below:

```
- celery.action ⇒ `run`
- celery.delivery_info ⇒ `{'exchange': '', 'routing_key': 'distilled-guidance-dpo-25-10k-v1', 'priority': 0, 'redelivered': False}`
- celery.hostname ⇒ `gen1@bestapi-6dfffd8486-6h9x5`
- celery.reply_to ⇒ `86b14e3b-bbdb-3d12-8838-c6125a236bc6`
- celery.state ⇒ `SUCCESS`
- celery.task_name ⇒ `agent_telemetry_processor`
- messaging.conversation_id ⇒ `90054b2c-4e21-4edd-8999-605c1aa703b0`
- messaging.destination ⇒ `distilled-guidance-dpo-25-10k-v1`
- messaging.message_id ⇒ `90054b2c-4e21-4edd-8999-605c1aa703b0`
```

Now I see attributes with `messaging.*` in the backend to do correlations, but as per discussion we can not have `messaging.system=celery`.

Now one way to know that task processor `celery` is used is to check the presence of `celery.*` attributes or capture it under some attribute name. Is there a sem conv for that or we can have something like `messaging.task.processor=celery`

PR #1972, opened 2025-03-07
Add top-level heading into k8s non-normative doc

### LINKED ISSUE #1971: [docs] Broken link - hash (fragment) does not exist - non-normative/k8s-attributes/#service-attributes
- Context: https://github.com/open-telemetry/opentelemetry.io/pull/6122, specifically the most recent commit https://github.com/open-telemetry/opentelemetry.io/pull/6122/commits/8b2b4b800ff462bba55405c75fa8391112be966f
- Log https://github.com/open-telemetry/opentelemetry.io/actions/runs/13718790997/job/38369381823?pr=6122: reported link-check failure, "hash does not exist" for `non-normative/k8s-attributes/#service-attributes`:
  ```text
  resource/k8s/index.html
    hash does not exist --- docs/specs/semconv/resource/k8s/index.html --> /docs/specs/semconv/non-normative/k8s-attributes/#service-attributes
  ```

lmolkova:
Fixes #1971
(🤞 )

Attempt to fix https://github.com/open-telemetry/opentelemetry.io/actions/runs/13718790997/job/38369381823?pr=6122

For some reason K8s attributes doc does not appear in the preview - https://deploy-preview-6122--opentelemetry.netlify.app/docs/specs/semconv/non-normative/k8s-migration/

image

The assumption is that something is broken in that doc since because of it not having a top-level heading.

### DIFF
--- a/docs/non-normative/k8s-attributes.md
+++ b/docs/non-normative/k8s-attributes.md
@@ -1,6 +1,10 @@
 <!--- Hugo front matter used to generate the website version of this page:
-linkTitle: K8s attributes
+linkTitle: Resource attributes in K8s
 --->
+
+# Populating resources attributes in Kubernetes
+
+This document describes how to specify resource attributes for Kubernetes services.
 
 ## Specify resource attributes using Kubernetes annotations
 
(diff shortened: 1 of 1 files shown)


lmolkova:
@chalin could you please take a look? I'm not sure if this will fix the broken link - for some reason the whole referenced doc does not show up in the preview.

dmitryax (approved):

chalin (changes requested):
Right, the page doesn't appear because there's no page title. I think that promoting the first heading, and keeping the linkTitle as is would be fine. See #1974. WDYT?

PR #1982, opened 2025-03-11
Rename `feature_flag.provider_name` to `feature_flag.provider.name`

dyladan:
Refs https://github.com/open-telemetry/semantic-conventions/pull/1708

This follows guidance in #1708 for provider naming conventions.


trask:
sorry, I know I suggested this change, but I think it might be blocked on #1462

dyladan:
@trask @lmolkova what do you suggest we do then? The FF group is at a point where we're getting ready to wind down and declare stable (or at least RC). This is one of the last few nits we were going to tackle. I don't think it makes sense to keep the SIG around just to wait on an upstream blocker.

beeme1mr (approved):
I'm fine with this change. It should allow us to add additional provider-related metadata if necessary in the future.

If we're going to make this change, now is the time. We're actively adding support in the OpenFeature SDKs and would like to ramp down the feature flag OTel project.

lmolkova:
> @trask @lmolkova what do you suggest we do then? The FF group is at a point where we're getting ready to wind down and declare stable (or at least RC). This is one of the last few nits we were going to tackle. I don't think it makes sense to keep the SIG around just to wait on an upstream blocker.

Let's bring it up on the semconv call on Monday? One possibility is to implement relatively small tooling feature -  https://github.com/open-telemetry/semantic-conventions/issues/1462#issuecomment-2655314451 that would make it easy.

If we allow exception for this single attribute in the policy, we'll need to communicate it well to all language SIGs and help them take an action of excluding conflicting attribute from codegen. I don't want to go down this route.

The last of all options is to keep feature flag conventions unstable because of this, but that's the worst option in my opinion.

trask:
> Let's bring it up on the semconv call on Monday?

I've added it to the agenda

kinyoklion (approved):

lukas-reining (approved):

joaopgrassi (approved):

askpt (approved):

dyladan:
@trask is this unblocked now?

lmolkova:
> @trask is this unblocked now?

need a new weaver release - most weaver folks at at the KubeCon, let me check when we can release

lmolkova:
This PR should unblock this https://github.com/open-telemetry/semantic-conventions/pull/2086

model/feature-flags/registry.yaml
```diff
@@ -11,7 +11,7 @@ groups:
         stability: development
         brief: The lookup key of the feature flag.
         examples: ["logo-color"]
-      - id: feature_flag.provider_name
```
lmolkova, line 14: you'd have to deprecate it, but add the annotation to exclude from codegen following example in https://github.com/open-telemetry/semantic-conventions/pull/2086

dyladan: @lmolkova done in feada1a98ca86286b5164d79cbfc66af78870c65 PTAL

trask (approved):

PR #1990, opened 2025-03-12
Move the evaluated value from the event body to attributes

dyladan:
This came as a result of @trask reviewing the feature flag semconv to move away from body fields to attributes. `type` is also added because the evaluated value can be any type.

PR #1992, opened 2025-03-12
Define generic `error.message`

dyladan:
Context: https://github.com/open-telemetry/semantic-conventions/pull/1990#discussion_r1991582570

This defines a generic `error.message` attribute for human-readable error messages to provide additional context and details about an error which occurred. It is meant to be used in conjunction with `error.type` which is low-cardinality.


beeme1mr (approved):

trask:
I'm supportive of this. I've added it to both the general semconv meeting and log SIG meeting agendas just in case we don't get feedback here on the PR.

dyladan:
Created #1994 for the feature flag implementation of this just to keep things separate

docs/attributes-registry/error.md
```diff
@@ -9,9 +9,12 @@ This document defines the shared attributes used to report an error.
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
-| <a id="error-type" href="#error-type">`error.type`</a> | string | Describes a class of error the operation ended with. [1] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| <a id="error-message" href="#error-message">`error.message`</a> | string | A message providing more detail about an error in human-readable form. [1] | `TypeError: Cannot read properties of undefined (reading 'toString')` | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 12: I'm supportive of this attribute but slightly worried about the duplication between `exception.*` and `error.*`. This PR does not change status quo, but makes the problem more obvious.

Could we come up with some litmus test on when to use `error.message` vs `exception.message` and add it to the note? Or at least say that semantic conventions  should wither avoid referencing both or should document the distinction and avoid duplicating the same info in both.

Also to this point, the example looks like `exception.message` - is there a reason we need a more generic `error.message` here?

docs/attributes-registry/error.md
```diff
@@ -9,9 +9,12 @@ This document defines the shared attributes used to report an error.
...
+| <a id="error-type" href="#error-type">`error.type`</a> | string | Describes a class of error the operation ended with. [2] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 
-**[1] `error.type`:** The `error.type` SHOULD be predictable, and SHOULD have low cardinality.
+**[1] `error.message`:** `error.message` should provide additional context and detail about an error. It is NOT RECOMMENDED to duplicate the value of `error.type` in `error.message`.
```
lmolkova, line 15: Can we say that this attribute would usually be used on the logs/events and should not be captured on spans where the same information is in the status description? also not captured on metrics due to cardinality.

It could also be useful to add a link to https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md

WDYT?

dyladan: what do you think of the wording in 93ec7cf1cf7da521be0e5294b4180bec29075976

model/error/registry.yaml
```diff
@@ -37,3 +37,14 @@ groups:
...
+      - id: error.message
+        type: string
+        stability: development
+        examples: ["TypeError: Cannot read properties of undefined (reading 'toString')"]
+        brief: A message providing more detail about an error in human-readable form.
+        note: |
+          `error.message` should provide additional context and detail about an error.
+          It is NOT RECOMMENDED to duplicate the value of `error.type` in `error.message`.
+
+          `error.message` is recommended for logs, events, and span events,
```
trask, line 49: we aren't going to define semantic conventions for logs or span events

```suggestion
          `error.message` is recommended for events
```

docs/attributes-registry/error.md
dyladan: We need an error message for cases where an exception is not thrown. The case of feature flag evaluation failure for example. There may be e.g. a timeout but no exception thrown.

lmolkova: makes sense, could we still add a note saying that it's not recommended to duplicate `error.message` and `exception.message`?

dyladan: sure

dyladan: I also changed the example to not appear to be a duplicate of an exception

lmolkova (approved):

jsuereth (approved):

docs/attributes-registry/error.md
```diff
@@ -9,9 +9,16 @@ This document defines the shared attributes used to report an error.
...
+| <a id="error-message" href="#error-message">`error.message`</a> | string | A message providing more detail about an error in human-readable form. [1] | `Unexpected input type: string`; `The user has exceeded their storage quota` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="error-type" href="#error-type">`error.type`</a> | string | Describes a class of error the operation ended with. [2] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 
-**[1] `error.type`:** The `error.type` SHOULD be predictable, and SHOULD have low cardinality.
+**[1] `error.message`:** `error.message` should provide additional context and detail about an error.
+It is NOT RECOMMENDED to duplicate the value of `error.type` in `error.message`.
+It is also NOT RECOMMENDED to duplicate the value of `exception.message` in `error.message`.
+
+`error.message` is recommended for events but not metrics or spans due to its unbounded cardinality and overlap with span status.
```
trask, line 19:
```suggestion
`error.message` is NOT RECOMMENDED for metrics or spans due to its unbounded cardinality and overlap with span status.
```

.chloggen/generic-error.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: error
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `error.message` property for human-readable error message.
```
trask, line 13:
```suggestion
note: Add `error.message` property for human-readable error message on events.
```

trask (approved):

model/error/registry.yaml
```diff
@@ -37,3 +37,14 @@ groups:
 
           - Use a domain-specific attribute
           - Set `error.type` to capture all errors, regardless of whether they are defined within the domain-specific set or not.
+      - id: error.message
+        type: string
+        stability: development
+        examples: ["Unexpected input type: string", "The user has exceeded their storage quota"]
+        brief: A message providing more detail about an error in human-readable form.
+        note: |
```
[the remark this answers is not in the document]
trask: does the text below cover your concern?

> `error.message` is recommended for events but not metrics or spans due to its unbounded cardinality and overlap with span status.

lukas-reining (approved):

carlosalberto (approved):

joaopgrassi (approved):

model/error/registry.yaml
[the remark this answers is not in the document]
dyladan: @jsuereth does that cover your concern?

dyladan:
made a mistake with git history. marking as draft until i resolve it

dyladan:
Fixed

PR #1994, opened 2025-03-12
Use generic `error.message` in feature flag evaluation event

dyladan:
Depends on #1992 

Replaces `feature_flag.evaluation.error.message` with `error.message`

Draft until #1992 is settled


model/feature-flags/events.yaml
```diff
@@ -46,9 +46,9 @@ groups:
           | `targeting_key_missing` | The provider requires a targeting key and one was not provided in the evaluation context. | ![Development](https://img.shields.io/badge/-development-blue) |
           | `type_mismatch` | The type of the flag value does not match the expected type. | ![Development](https://img.shields.io/badge/-development-blue) |
           | `general` | The error was for a reason not enumerated above. | ![Development](https://img.shields.io/badge/-development-blue) |
-      - ref: feature_flag.evaluation.error.message
+      - ref: error.message
         requirement_level:
-          recommended: If and only if an error occurred. It's NOT RECOMMENDED to duplicate the value of `error.type` in `feature_flag.evaluation.error.message`.
+          recommended: If and only if an error occurred during flag evaluation.
```
dyladan, line 51: Would we want a further caveat on the error message? For example if type is `flag_not_found` we don't want to have message 'flag not found' right? We would want the message to provide additional context IFF there is additional context to provide.

```suggestion
          recommended: If and only if an error occurred during flag evaluation and `error.type` does not sufficiently describe the error.
        note: >
          Should not simply duplicate the value of `error.type`, but should provide more context.
          For example, if `error.type` is `invalid_context` the `error.message` may enumerate which context keys are missing or invalid.
```

schema-next.yaml
```diff
@@ -2,6 +2,12 @@ file_format: 1.1.0
 schema_url: https://opentelemetry.io/schemas/1.31.0
 versions:
   next:
+    all:
+      changes:
+        # todo link
```
lmolkova, line 7:
```suggestion
        # https://github.com/open-telemetry/semantic-conventions/pull/1994
```

joaopgrassi (approved):

trask (approved):

PR #1383, opened 2024-09-03
Add VCS metrics from Github receiver

christophe-kamphaus-jemmic:
Fixes #1372 

## Changes

This PR adds the VCS metrics from the GitHub Receiver to Semantic Conventions.
These metrics are based on https://github.com/adrielp/opentelemetry-collector-contrib/blob/903eec5382352e4d663fadd2dc7f1cc1e37f62fd/receiver/githubreceiver/documentation.md

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
  * [x] Sign CLA
* [x] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [x] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

## TODO

* [x] Update `docs/cicd/cicd-metrics.md` after receiving changes from https://github.com/open-telemetry/semantic-conventions/pull/1411

PR #1997, opened 2025-03-16
[vcs] add organization and provider name VCS attributes

### LINKED ISSUE #1452: Add organization name to attribute list when entity definition becomes clearer
# Overview

Update semantic conventions to contain the following resource attributes.

Ensure these attributes won't be misaligned with new entity definitions of attributes.

- [ ] `vcs.owner.name` - the group owner within the version control system. (ie. myorg, myteam)
- [ ] `vcs.vendor.name` - the vendor name of the version control system. (ie. github/gitlab)

afonsonf:
Fixes #1452

## Changes

Add organization and provider name to VCS attribute list


.chloggen/1452-vcs-metrics.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: vcs
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add organization and vendor name to VCS attribute list
```
adrielp, line 13:
```suggestion
note: Add organization and vendor name to VCS attribute registry
```

model/vcs/metrics.yaml
adrielp, line 1: Do we want to recommend these metrics have those attributes on them, or do we want to say they are opt in?

model/vcs/registry.yaml
```diff
@@ -239,3 +239,23 @@ groups:
...
         examples: ["open", "closed", "merged"]
+      - id: vcs.owner.name
+        type: string
+        stability: development
+        brief: >
+           The group owner within the version control system.
+        examples:
+          [
+            "my-org",
+            "myteam",
```
adrielp, line 250: I hadn't thought of owner as a team name. I think `service.namespace` can be considered a team name as well. I'd be fine w/ owner being a team name I think, but also I'd add maybe a group/subgroup level of ownership. Or business unit as an example?
```suggestion
            "myteam",
```

model/vcs/registry.yaml
```diff
@@ -239,3 +239,23 @@ groups:
...
+          ]
+      - id: vcs.vendor.name
+        type: string
+        stability: development
+        brief: >
+          The vendor name of the version control system.
+        examples:
+          [
+            "github",
+            "gitlab",
```
adrielp, line 260: Let's add some additional example so as to not act like we're showing favoritism towards those two. 

Maybe add, `gittea` and `bitbucket` as other options, thoughts?

afonsonf: Added the new examples

model/vcs/registry.yaml
afonsonf: the `myteam` was one of the examples in the issue, is `service.namespace` a new attribute to be added to vcs?
I added `business-unit` as an example

adrielp: Great! Thanks. 

re: `service.namespace` -- [service.namespace](https://opentelemetry.io/docs/specs/semconv/attributes-registry/service/#service-namespace) is one of the de-facto attributes that get attached to spans, etc. It is often used to denote the owning team of a service. However, it's subject to interpretation and can denote other things. I was only mentioning it here as a thought. The way you have presented `vcs.owner.name` is reasonable. Nothing needs to be down about `service.namespace` right now.

adrielp (approved):

model/vcs/registry.yaml
afonsonf: thanks for the explanation!

model/vcs/metrics.yaml
afonsonf: I've added the attributes to the metrics that had the `vcs.repository.name` as I considered it similar metadata

However I have read a bit more on the discussions originating the ticket and `vcs.owner.name` seems to have been intended for the metric `metric.vcs.repository.count`, which I missed, should I add it there as required or just recommended as the other ones? (https://github.com/open-telemetry/semantic-conventions/pull/1383#discussion_r1742695951)
I'll also add the vendor attribute to that metric

For the opt_in vs recommended in the other metrics:
-  for the owner attribute, I think it makes sense to be recommended in all of them (to group/analyse the metrics per organization)
- for the vendor attribute, I can change it to opt_in as from my initial view it does not bring much to the metrics (was it intended for some instrumentation?)

model/vcs/registry.yaml
```diff
@@ -239,3 +239,26 @@ groups:
...
+        stability: development
+        brief: >
+           The group owner within the version control system.
+        examples:
+          [
+            "my-org",
+            "myteam",
+            "business-unit",
+          ]
+      - id: vcs.vendor.name
```
lmolkova, line 253: provider would be a broader and more neutral term here. We also have some examples of `*.provider.name` (e.g. feature_flag.provider.name, https://github.com/open-telemetry/semantic-conventions/pull/1982) 

```suggestion
      - id: vcs.provider.name
```

adrielp: `provider` sounds good, especially if there's consistency there w/ other attributes.

model/vcs/registry.yaml
```diff
@@ -239,3 +239,26 @@ groups:
...
+        type: string
+        stability: development
+        brief: >
+          The vendor name of the version control system.
+        examples:
+          [
+            "github",
```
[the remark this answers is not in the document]
adrielp: good call

model/vcs/metrics.yaml
afonsonf: Added the attributes to the `metric.vcs.repository.coun` metric and changed the attribute `vcs.`provider.name` to `opt_in` in all metrics

model/vcs/registry.yaml
afonsonf: Thanks for review, updated

model/vcs/registry.yaml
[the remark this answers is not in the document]
afonsonf: Thanks, updated

model/vcs/registry.yaml
```diff
@@ -239,3 +239,26 @@ groups:
...
+          [
+            "my-org",
+            "myteam",
+            "business-unit",
+          ]
+      - id: vcs.vendor.name
+        type: string
+        stability: development
+        brief: >
+          The vendor name of the version control system.
```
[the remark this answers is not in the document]
afonsonf: Thanks, updated

lmolkova (approved):

docs/attributes-registry/vcs.md
```diff
@@ -96,6 +98,18 @@ the `.git` extension.
 
 ---
 
+`vcs.provider.name` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
kamphaus, line 101: Since we may use a custom value is there a reason to include `other_provider`?

afonsonf: good point, I'll create a PR to remove it

afonsonf: https://github.com/open-telemetry/semantic-conventions/pull/2020

PR #2006, opened 2025-03-18
Add GCP AppHub resource labels

jsuereth:
Adds AppHub label definitions which GCP will automatically provide for registered workloads or services. These labels are provided by GCP, and documented here for integrations.


braydonk (approved):
LGTM

docs/resource/cloud-provider/gcp/apphub.md
```diff
@@ -0,0 +1,102 @@
+# Google Cloud AppHub
+
+**Status**: [Development][DocumentStatus]
+
+These conventions are used by Google Cloud when using AppHub. Observability
```
aabmass, line 5:
```suggestion
These conventions are used by Google Cloud when using [AppHub](https://cloud.google.com/app-hub/docs/overview). Observability
```

aabmass (approved):

model/gcp/registry.yaml
```diff
@@ -39,6 +39,75 @@ groups:
...
+        brief: >
+          The name of the application as configured in AppHub.
+        examples: ['my-application']
+      - id: gcp.apphub.service.id
+        type: string
+        stability: development
+        brief: >
+          The name of the service as configured in AppHub.
+        examples: ['my-service']
+      - id: gcp.apphub.service.environment_type
```
lmolkova, line 73: applying our current naming guidance https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md, this one should be 

```suggestion
      - id: gcp.apphub.service.environment.type
```

(we default to `.` whenever it makes sense)

model/gcp/registry.yaml
```diff
@@ -39,6 +39,75 @@ groups:
...
+        type: string
+        stability: development
+        brief: >
+          Environment of an application, service, or workload is the stage of a software lifecycle, for example, production, staging, or development.
+        note: >
+          [See AppHub properties and attributes](https://cloud.google.com/app-hub/docs/overview#properties-attributes)
+        examples: ['DEVELOPMENT', 'TEST', 'STAGING', 'PRODUCTION']
+      - id: gcp.apphub.service.criticality_type
```
lmolkova, line 81: would 
```suggestion
      - id: gcp.apphub.service.criticality.type
```

work too?

model/gcp/registry.yaml
```diff
@@ -39,6 +39,75 @@ groups:
...
+        note: >
+          [See AppHub properties and attributes](https://cloud.google.com/app-hub/docs/overview#properties-attributes)
+        examples: ['CRITICAL', 'HIGH', 'MEDIUM', 'LOW']
+      - id: gcp.apphub.workload.id
+        type: string
+        stability: development
+        brief: >
+          The name of the workload as configured in AppHub.
+        examples: ['my-workload']
+      - id: gcp.apphub.workload.environment_type
```
lmolkova, line 95: I assume [`deployment.environment.name`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/deployment.md) does not work (since there are two types - one for `service`, another one for `workload` and they may be different)?

model/gcp/registry.yaml
```diff
@@ -39,6 +39,75 @@ groups:
...
+        type: string
+        stability: development
+        brief: >
+          Environment of an application, service, or workload is the stage of a software lifecycle, for example, production, staging, or development.
+        note: >
+          [See AppHub properties and attributes](https://cloud.google.com/app-hub/docs/overview#properties-attributes)
+        examples: ['DEVELOPMENT', 'TEST', 'STAGING', 'PRODUCTION']
```
lmolkova, line 102: would it make sense to define those in an enum? also, any reason for ALL CAPS?

model/gcp/registry.yaml
```diff
@@ -39,6 +39,75 @@ groups:
...
+        type: string
+        stability: development
+        brief: >
+          The name of the workload as configured in AppHub.
+        examples: ['my-workload']
+      - id: gcp.apphub.workload.environment_type
+        type: string
+        stability: development
+        brief: >
+          Environment of an application, service, or workload is the stage of a software lifecycle, for example, production, staging, or development.
```
lmolkova, line 99: nit:

```suggestion
          Environment of workload is the stage of a software lifecycle.
```

?

jsuereth: I copied this from apphub docs.  Happy to rephrase.

model/gcp/registry.yaml
jsuereth: Yes - effectively this is the same, however as GCP is providing these labels on top of metrics, we did not want to "take away" or override user-configured environment name.

More specifically, it's possible if a user is already configuring `deployment.environment.name` then the apphub environment type may not match.

While we can/should use deployment.environment.name when possible, we need a fallback to ensure the AppHub labels don't conflict.

Additionally, as these labels are only produced by GCP itself -> did not want to rely on unstable portions of semantic conventions for them.


Again, for context, AppHub will annnotate telemetry data with these labels.  The reason for this PR is to advertise what those are and how to consume them.

model/gcp/registry.yaml
jsuereth: see other comment

model/gcp/registry.yaml
jsuereth: Ok, updated and used the better link in the notes now.

model/gcp/registry.yaml
lmolkova: then I see no problem with keeping the original naming, the `.` is just the default

model/gcp/registry.yaml
```diff
@@ -39,6 +39,139 @@ groups:
...
+              value: "TEST"
+              brief: Test environment.
+              stability: development
+            - id: development
+              value: "DEVELOPMENT"
+              brief: Development environment.
+              stability: development
+        stability: development
+        brief: >
+          Environment of a workload is the stage of a software lifecycle, for example, production, staging, or development.
```
lmolkova, line 148: super-nit - it's just a dup of examples (and having shorter briefs makes our markdown tables more readable

```suggestion
          Environment of a workload is the stage of a software lifecycle.
```

model/gcp/registry.yaml
```diff
@@ -39,6 +39,139 @@ groups:
...
+              value: "TEST"
+              brief: Test environment.
+              stability: development
+            - id: development
+              value: "DEVELOPMENT"
+              brief: Development environment.
+              stability: development
+        stability: development
+        brief: >
+          Environment of a service is the stage of a software lifecycle, for example, production, staging, or development.
```
lmolkova, line 94: see the other comment 

```suggestion
          Environment of a service is the stage of a software lifecycle.
```

Issue #1906, opened 2025-02-13
SDK exporter health metrics: Possibly enhance with duration


JonasKunz:
#1631 added SDK health metrics for span-related processing. A discussion that came up towards the end of this PR is whether to add a duration metric for exporters in addition to or as replacement for `otel.sdk.exporter.span.exported.count`.
See [this comment](https://github.com/open-telemetry/semantic-conventions/pull/1631#issuecomment-2652534950) and [this comment](https://github.com/open-telemetry/semantic-conventions/pull/1631#issuecomment-2655276918).

The goal of this issue is to evaluate that addition and maybe also consider exporter-request metrics for duration, size and errors.

JonasKunz:
To continue the discussion here, the content of the last comment:

>> At the same time I can't really see the general importance / usefulness of having the exporter durations: It feels like more of a nice to have. What conclusions does this metric allow you to draw? Do you have concrete examples?
> 
>Debugging connectivity with my backend - network issues, throttling, slow backend response, retries, retry backoff interval optimizations.
> As @lzchen mentioned, absolute majority of OTel SDKs suppress all instrumentation around exporter calls, so you'd not have protocol-level spans and/or metrics. So a metric like this would be the only way to know how good your connectivity is.
>
> Counts are good, but they won't tell you that your P99 is 10 sec after all tries because your backoff interval is wrong. You'd just see less of them and will have no idea.
>
> I don't think it's nice to have.
>Given it's experimental, we can go ahead with just counts and potentially deprecating them when/if histograms are added.
>
> As a cost mitigation strategy, we can always use a small amount of buckets by default and users can always reconfigure them if they need less/more.

Here are my thoughts:

> Debugging connectivity with my backend - network issues, throttling, slow backend response, retries, retry backoff interval optimizations.

I think for those it is better to have dedicated `otel.sdk.exporter.request.*` metrics for request-based exporters. We can lend the overall design from the existing HTTP metrics here. I think that should also be applicable to other types of network requests, too:
 * `otel.sdk.exporter.request.duration`
 * `otel.sdk.exporter.request.body.size`

These metrics would be applicable to all kinds of exporters (span, metric, traces) and should cover the above use-cases except for retries.

IIRC retries was one of the motivations to go for a `otel.sdk.exporter.span.duration` metric: It would cover the time a span takes from being submitted to the exporter until it is either exported successfully or is dropped as failed, including potential retries.

[...]

PR #2007, opened 2025-03-18
Add http request duration to SDK metrics

lzchen:
Related to #1631 and partially addresses #1906

## Changes

Adds a metric to track the duration of http requests made within an exporter.

Following the discussion in this [issue](https://github.com/open-telemetry/semantic-conventions/issues/1906), I wasn't exactly sure how we want to provide more detail regarding partial success. The `error.type` can indicate that it was a partial success but we can possibly also include information about which items succeeded/failed somewhere (perhaps in another attribute).

I also made the distinction of making this metric specific to `http` but am open to defining a generic one regardless of transport protocol (we can probably have protocol + response code as additional attributes if so).

Looking forward to hearing your thoughts!


model/otel/metrics.yaml
```diff
@@ -110,3 +110,30 @@ groups:
           recommended: when applicable
       - ref: error.type
         examples: ["rejected", "timeout", "500", "java.net.UnknownHostException"]
+
+  - id: metric.otel.sdk.exporter.http.request.duration
```
JonasKunz, line 114: Any reason for limiting this metric to `http` instead of defining it for all network-request based exporters?
E.g. `otel.sdk.exporter.request.duration` instead of `otel.sdk.exporter.http.request.duration`? I think we could still keep the `http.response.status_code` around with `recommended: when applicable`.

docs/otel/sdk-metrics.md
```diff
@@ -452,5 +454,104 @@ E.g. for Java the fully qualified classname SHOULD be used in this case.
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Request Metrics
+
+### Metric: `otel.sdk.exporter.http.request.duration`
+
+This metric is [recommended][MetricRecommended].
```
JonasKunz, line 461: I think in general histogram health metrics should be opt-in, due to the higher storage costs of histograms.
Though for this specific metric, I think it might be debatable, as it is a very fundamental metric for network-based exporters.

docs/otel/sdk-metrics.md
```diff
@@ -452,5 +454,104 @@ E.g. for Java the fully qualified classname SHOULD be used in this case.
...
+<!-- semconv metric.otel.sdk.exporter.http.request.duration -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `otel.sdk.exporter.http.request.duration` | Histogram | `s` | The duration of any http request(s) made from the exporter during export. [1] | ![Development](https://img.shields.io/badge/-development-blue) |
```
jsuereth, line 476: At a high level, just want to understand why this wouldn't use the general HTTP semantic convention duration with a required label denoting it's the SDK exporter.

trask: separate monitoring pipeline metrics could make it easier for ingestion services to handle them differently (e.g. maybe not consider them "user data", surface pipeline issues, etc)

lmolkova: I hope we can have a export call duration metrics that covers all tries and has a final export status - the logical rather than physical. So I'd also prefer to remove `http` and would suggest to call it `otel.sdk.export.duration` (`otel.sdk.exporter.operation.duration`)

model/otel/metrics.yaml
lzchen: No strong reason in particular. I just wanted the status code to be represented as an attribute.

docs/otel/sdk-metrics.md
lzchen: @jsuereth 

Due to the change after [this comment](https://github.com/open-telemetry/semantic-conventions/pull/2007#discussion_r2003101279), the metric is now being used for all network based exporters instead of just http. 

@lmolkova 

I have removed the `http`. 

> covers all tries and has a final export status

This metric is meant to capture requests sent as a batch. I believe covering all retries constitutes knowing how each individual item in the batch has behaved until final export status. I don't know how that would translate to duration and also I don't think that's within the scope of a batch level metric. Perhaps that would be better suited for a separate metric, as our use case requires knowing the duration of requests on a batch level.

JonasKunz: > I hope we can have a export call duration metrics that covers all tries and has a final export status - the logical rather than physical

IMO we should stay with `otel.sdk.exporter.request.duration` and keep this about the physical request.
The problem with tracking the total duration including retries is that I don't think most implementations would treat retried batches and new batches separetly and might combine them onto a single, physical request. This makes it impossible to track the duration correctly with partial successes. I've given a detailed explanation and example in [this comment](https://github.com/open-telemetry/semantic-conventions/issues/1906#issuecomment-2665550188).

Another benefit of having a duration per physical request that you also get the `error.type` which caused the retry: e.g. was it a network error or did the receiver return `503 Temporarily Unavailable`?

I would suggest to add a separate metric if we want to track the "logical" request (= a batch of data to be exported). Something like `otel.sdk.exporter.batch.latency`. If an exporter does not have partial success semantics OR retries batches in isolation, it can produce this metric. Otherwise we will still have `otel.sdk.exporter.request.duration`, which also remains useful in combination with `otel.sdk.exporter.batch.latency` due to giving insights on the physical requests.

lmolkova: patrial success is a terminal status code, it won't  be retried - the exporter would need to create a new batch if they want to re-send it

> The client MUST NOT retry the request when it receives a partial success response where the partial_success is populated.

from https://opentelemetry.io/docs/specs/otlp/#partial-success

We do have protocol-level metrics for HTTP and gRPC, we just need to let them do the work (not suppress metrics from exporters), the logical duration is a separate metrics, it can't be replaced with a physical ones.

JonasKunz: > patrial success is a terminal status code, it won't be retried - the exporter would need to create a new batch if they want to re-send it

The problem I'm thinking of is not partial-successes being retried. It is about requests with partial success containing a mix of retried and new data:

> 1. A batch of spans is submitted to the exporter
> 2. The request fails for a network reason
> 3. In the meantime, a second batch of spans is submitted for export
> 4. The second request is issued, containing the spans from both batches
> 5. The request finishes with partial success, returning that 5 spans were rejected

In this example, it is not possible to identify if the rejected spans belong to the first or second batch, therefore it is not possible to correctly populate the `error.type` for the batches.

I don't know if any exporter does this or whether this is forbidden, I just wanted to highlight that it might not be possible to correctly collect this metric in certain scenarios.

> We do have protocol-level metrics for HTTP and gRPC, we just need to let them do the work (not suppress metrics from exporters), the logical duration is a separate metrics, it can't be replaced with a physical ones.

For that I'm with @trask :

> separate monitoring pipeline metrics could make it easier for ingestion services to handle them differently (e.g. maybe not consider them "user data", surface pipeline issues, etc)

I think that telemetry about opentelemetry itself should be clearly separated from "user data" too. Some examples where not doing this would lead to problems:

* A user has a HTTP-statistics dashboard for a mission-critical service. He now needs to somehow filter out the Opentelemetry-Exporter requests to not distort the data he is actually interested in. To do so, he needs deep deployment knowledge about the deployed OTLP-endpoints to filter them out correctly. Even worse, he might have to change his dashboard if the monitoring setup changes and e.g. collectors are deployed under different URLs.

* It makes sense for mission critical environments to have a separate "user-data" telemetry backend and a health-telemetry backend. This way if your "user-data" backend is overloaded, you'll still have your health-telemetry to investigate the problem. If you have both in a single monitoring backend, you won't be able to access your health telemetry in case the backend is overloaded, which ironically is when you need those metrics most. Having the health metrics in well-defined, separated namespaces (e.g. `otel.sdk.*`) makes it easy to set this up: Just setup two metric exporters in your SDKs, one exporting just `otel.sdk.*` to your health-telemetry backend and one exporting everything except `otel.sdk.*`. If you have to seaprate metrics based on attribute values things get a lot more complicated.

I'm not asking to necessarily add physical request metrics in this PR, I just would like to reserve the `otel.sdk.exporter.request.*` for those and use `otel.sdk.exporter.batch.duration` or `otel.sdk.export.duration` for the logical duration instead.

lmolkova: @JonasKunz 

the export duration would measure this call:

https://github.com/open-telemetry/opentelemetry-go/blob/796596abbad67e221a89e225174715a74888101c/exporters/otlp/otlptrace/exporter.go#L36

> 	err := e.client.UploadTraces(ctx, protoSpans)

there is no logic that re-batches things on retry - the retry attempts to re-send the same batch. 

Do I miss something? Is there an SDK exporter that has some dynamic batching mechanism?

JonasKunz: I wasn't talking about a concrete implementation and also not only thinking about OTLP. But it is possible to provide an implementation where my example scenario can occur.

Anyway, this shouldn't be problematic as long as the metric is not a MUST. So I'd propose to just move ahead with the metric for the logic duration, but reserve `otel.sdk.exporter.request.*` for physical request metrics?

docs/otel/sdk-metrics.md
```diff
@@ -452,5 +454,131 @@ E.g. for Java the fully qualified classname SHOULD be used in this case.
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Request Metrics
+
+### Metric: `otel.sdk.exporter.request.duration`
```
lmolkova, line 459: suggesting 
```suggestion
### Metric: `otel.sdk.exporter.operation.duration`
```

to match other logical operations

model/otel/metrics.yaml
```diff
@@ -110,3 +110,34 @@ groups:
           recommended: when applicable
       - ref: error.type
         examples: ["rejected", "timeout", "500", "java.net.UnknownHostException"]
+
+  - id: metric.otel.sdk.exporter.request.duration
+    type: metric
+    metric_name: otel.sdk.exporter.request.duration
+    stability: development
+    brief: "The duration of any request(s) made from the exporter during export."
```
lmolkova, line 118:
```suggestion
    brief: "The duration of exporting a back of telemetry items. It cover the duration including all retries"
```

model/otel/metrics.yaml
```diff
@@ -110,3 +110,34 @@ groups:
...
+    note: |
+      This metric defines successful requests using the full success definitions for [http](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/docs/specification.md#full-success-1)
+      and [grpc](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/docs/specification.md#full-success). Anything else is defined as an unsuccessful request. For successful requests,
+      `error.type` MUST NOT be set. For unsuccessful requests, `error.type` must contain a relevant failure cause.
```
lmolkova, line 122:
```suggestion
      `error.type` MUST NOT be set. For unsuccessful export operations, `error.type` must contain a relevant failure cause.
      It should be reported for export calls limited to a single batch. If exporter updates batch between retries, this metric SHOULD NOT be reported.
```

model/otel/metrics.yaml
```diff
@@ -110,3 +110,34 @@ groups:
...
         examples: ["rejected", "timeout", "500", "java.net.UnknownHostException"]
+
+  - id: metric.otel.sdk.exporter.request.duration
+    type: metric
+    metric_name: otel.sdk.exporter.request.duration
+    stability: development
+    brief: "The duration of any request(s) made from the exporter during export."
+    note: |
+      This metric defines successful requests using the full success definitions for [http](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/docs/specification.md#full-success-1)
+      and [grpc](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/docs/specification.md#full-success). Anything else is defined as an unsuccessful request. For successful requests,
```
lmolkova, line 121:
```suggestion
      and [grpc](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/docs/specification.md#full-success). Anything else is defined as an unsuccessful operations. For successful operations,
```

model/otel/metrics.yaml
```diff
@@ -110,3 +110,34 @@ groups:
...
       - ref: error.type
         examples: ["rejected", "timeout", "500", "java.net.UnknownHostException"]
+
+  - id: metric.otel.sdk.exporter.request.duration
+    type: metric
+    metric_name: otel.sdk.exporter.request.duration
+    stability: development
+    brief: "The duration of any request(s) made from the exporter during export."
+    note: |
+      This metric defines successful requests using the full success definitions for [http](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/docs/specification.md#full-success-1)
```
lmolkova, line 120:
```suggestion
      This metric defines successful operations using the full success definitions for [http](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/docs/specification.md#full-success-1)
```

model/otel/metrics.yaml
```diff
@@ -110,3 +110,34 @@ groups:
...
+      - ref: otel.component.name
+      - ref: server.address
+        requirement_level:
+          recommended: when applicable
+      - ref: server.port
+        requirement_level:
+          recommended: when applicable
+      - ref: error.type
+        requirement_level:
+          conditionally_required: If request has ended with an error
```
lmolkova, line 136:
```suggestion
          conditionally_required: If the operation has ended with an error
```

model/otel/metrics.yaml
```diff
@@ -110,3 +110,34 @@ groups:
...
+        examples: ["rejected", "timeout", "500", "java.net.UnknownHostException"]
+      - ref: http.response.status_code
```
lmolkova, line 138:
```suggestion
      - ref: http.response.status_code
        brief: The HTTP status code of the last HTTP request performed in scope of this export call.
```

model/otel/metrics.yaml
```diff
@@ -110,3 +110,34 @@ groups:
...
+        requirement_level:
+          recommended: when applicable
+      - ref: rpc.grpc.status_code
```
lmolkova, line 141:
```suggestion
      - ref: rpc.grpc.status_code
        brief: The gRPC status code of the last gRPC requests performed in scope of this export call.
```

.chloggen/requests-duration.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: 'otel'
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Adds SDK self-monitoring metrics for exporter request duration
```
lmolkova, line 13:
```suggestion
note: Adds SDK self-monitoring metrics for exporter call duration
```

docs/otel/sdk-metrics.md
lmolkova: let's go ahead with it, but I would expect that physical requests are covered with standard protocol-level metrics. We can add a flag indicating it came from exporter or just rely on the `server.address:port` to uniquely identify the exporter.

From semconv perspective we avoid defining multiple metrics for the same operation

docs/otel/sdk-metrics.md
```diff
@@ -882,5 +887,136 @@ E.g. for Java the fully qualified classname SHOULD be used in this case.
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Operation Metrics
+
+### Metric: `otel.sdk.exporter.operation.duration`
+
+This metric is [opt-in][MetricOptIn].
```
lmolkova, line 894: Discussed at the SemConv SIG meeting 4/7/25:

- this is an important metrics that allows to count number of export calls (not a number of spans) and final (after all tries) success rate. It should be recommended
- to offset the costs associated with histogram, we should consider reducing the number of buckets. E.g. we've done 4 buckets in [JVM GC duration](https://github.com/open-telemetry/semantic-conventions/blob/c1d8d371e8ad7cf47b50b6e0dd976746acdaeddf/docs/runtime/jvm-metrics.md#metric-jvmgcduration). We can even start with one bucket here. 

/cc @JonasKunz to keep me honest here

PS: if we go ahead with one bucket, I'd suggest `1` sec as something I would use for P95 export duration target.

trask: > PS: if we go ahead with one bucket, I'd suggest `1` sec as something I would use for P95 export duration target.

just a terminology clarification, on the SIG call I was thinking one bucket histogram would mean no boundaries

lzchen: @trask 

Just to clarify, does this me a `ExplicitBucketBoundaries` defined with `[double.max]` as a single entry?

lmolkova: I believe it's an empty array which results in a single bucket covering -Inf, Inf range. You still get count and sum, i.e. average from it.

lmolkova: @lzchen could you please also make the metric recommended? Thanks!

model/otel/metrics.yaml
```diff
@@ -205,3 +205,36 @@ groups:
...
+    stability: development
+    brief: "The duration of exporting a batch of telemetry items."
+    note: |
+      This metric defines successful operations using the full success definitions for [http](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/docs/specification.md#full-success-1)
+      and [grpc](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/docs/specification.md#full-success). Anything else is defined as an unsuccessful operation. For successful
+      operations, `error.type` MUST NOT be set. For unsuccessful export operations, `error.type` must contain a relevant failure cause.
+    instrument: histogram
+    unit: "s"
+    attributes:
+      - ref: otel.component.type
```
lmolkova, line 221: nit:

```suggestion
      - ref: otel.component.type
        examples: ["otlp_grpc_span_exporter", "com.example.MySpanExporter"]
```

(just to avoid having `batching_span_processor` in the examples)

lmolkova (commented):
LGTM

.chloggen/requests-duration.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: 'otel'
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Adds SDK self-monitoring metrics for exporter call duration
```
JonasKunz, line 13: Nit: we add just one metric here
```suggestion
note: Adds SDK self-monitoring metric for exporter call duration
```

JonasKunz (approved):
LGTM (after applying @lmolkova's suggestions)

lzchen:
Is there something failing with the builds? Doesn't look like it's related to this pr. :O

model/otel/metrics.yaml
```diff
@@ -205,3 +205,37 @@ groups:
           recommended: when applicable
       - ref: error.type
         examples: ["rejected", "timeout", "500", "java.net.UnknownHostException"]
+
+  - id: metric.otel.sdk.exporter.operation.duration
+    type: metric
+    metric_name: otel.sdk.exporter.operation.duration
+    stability: development
+    brief: "The duration of exporting a batch of telemetry items."
```
trask, line 213: I think this is more commonly used
```suggestion
    brief: "The duration of exporting a batch of telemetry records."
```

lmolkova (approved):

docs/otel/sdk-metrics.md
```diff
@@ -882,5 +887,135 @@ E.g. for Java the fully qualified classname SHOULD be used in this case.
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Operation Metrics
+
+### Metric: `otel.sdk.exporter.operation.duration`
```
trask, line 892: is the idea for this to include metric exporter durations as well?

lzchen: I believe so.

lmolkova:
@lzchen could you please resolve the conflict? It's be ready to go otherwise

lzchen:
@lmolkova 
Should be good to go!

Issue #1111, opened 2024-06-02
CI/CD conventions for metrics


christophe-kamphaus-jemmic:
### Area(s)

area:new

### Is your change request related to a problem? Please describe.

This issue is to discuss attributes specific to metrics and as part of the CI/CD Working Group and Semantic Conventions WG.
Also a challenge specific to metrics can the time series cardinality when CICD observes metrics for individual builds.

### Describe the solution you'd like

Following https://github.com/open-telemetry/semantic-conventions/pull/1075 (by adjusting the vocabulary here below to align with #1075) we should define metric attributes for
* duration of pipelineRuns (by status, pipeline)
* count of pipelineRuns (by status, pipeline)
* count of agents
* queue length of pending pipelineRuns
* duration for how long a pipelineRun is in the queue before starting execution

Additionally it should be possible to opt-in to metrics specific to a particular pipelineRun.
These could be metrics about the agent which executes a pipelineRun, the OS, network, jvm, the number of failed/total tests …
We need to specify the attribute which should link these metrics to the pipelineRun, eg. `pipeline.run.id`

Metrics specific to a pipelineRun are of high cardinality. We should document this as a warning and give guidance how these metrics can be efficiently encoded in the OTel protocol, ie by using resource attributes instead of metric attributes wherever possible.

### Describe alternatives you've considered

Span metrics could be used for duration and count of pipelineRuns, however this relies on the pipelineRuns having completed.
This is due to limitations inherent in using traces to represent pipelineRuns, a span can only be sent when complete.
Due to this limitation it could be preferable for the CICD system to expose metrics directly about the duration, count and status of pipelineRuns. These pipelineRuns could account also for in progress builds.

### Additional context

CICD metrics were discussed at KubeCon March 2024 SemConv users meeting.
Notes on how to deal with cardinality were:
* Could we use Exemplars? We could link to the build trace from some metrics.
This added information might make it easier to identify pipelineRuns that need investigation.
* Using the resource attribute for the build ID is fine for the OTel protocol,
but backends (eg. Prometheus) would still have the cardinality issue when storing the time series
(metric / resource attributes would be flattened into time series).

christophe-kamphaus-jemmic:
We can use label [area:cicd](https://github.com/open-telemetry/semantic-conventions/labels/area%3Acicd) instead of `area:new`.

adrielp:
[...]

PR #2013, opened 2025-03-25
Add CICD spans and resources

### LINKED ISSUE #1713: Add CICD conventions for spans
Related to https://github.com/open-telemetry/semantic-conventions/issues/1645

Define SemConv conventions for CICD spans related to pipelines, pipeline runs, stages, steps, …

kamphaus:
Fixes #1713

## Changes

This PR defines spans for CICD pipeline runs as well as the associated resources (both CICD and VCS).
Some attributes that are needed in these spans and resources were added.


docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,106 @@
...
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`cicd.pipeline.result`](/docs/attributes-registry/cicd.md) | string | The result of a pipeline run. | `success`; `failure`; `timeout`; `skipped` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+The following attributes can be important for making sampling decisions
+and SHOULD be provided **at span creation time** (if provided at all):
+
+* [`cicd.pipeline.result`](/docs/attributes-registry/cicd.md)
```
christophe-kamphaus-jemmic, line 44: So that's what the `sampling_relevant` flag is for.
This is not right since the result can only be set at the end of the pipeline run.
→ review use of `sampling_relevant` flag.

docs/attributes-registry/cicd.md
```diff
@@ -16,10 +16,14 @@ This group describes attributes specific to pipelines within a Continuous Integr
...
 | <a id="cicd-pipeline-task-name" href="#cicd-pipeline-task-name">`cicd.pipeline.task.name`</a> | string | The human readable name of a task within a pipeline. Task here most closely aligns with a [computing process](https://wikipedia.org/wiki/Pipeline_(computing)) in a pipeline. Other terms for tasks include commands, steps, and procedures. | `Run GoLang Linter`; `Go Build`; `go-test`; `deploy_binary` | ![Development](https://img.shields.io/badge/-development-blue) |
 | <a id="cicd-pipeline-task-run-id" href="#cicd-pipeline-task-run-id">`cicd.pipeline.task.run.id`</a> | string | The unique identifier of a task run within a pipeline. | `12097` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="cicd-pipeline-task-run-result" href="#cicd-pipeline-task-run-result">`cicd.pipeline.task.run.result`</a> | string | The result of a task run. | `success`; `failure`; `timeout`; `skipped` | ![Development](https://img.shields.io/badge/-development-blue) |
```
kuisathaverat, line 19: Do we need also `canceled`?

docs/attributes-registry/cicd.md
```diff
@@ -46,6 +50,19 @@ This group describes attributes specific to pipelines within a Continuous Integr
...
 ---
 
+`cicd.pipeline.task.run.result` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `cancellation` | The task run was cancelled, eg. by a user manually cancelling the task run. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `error` | The task run failed due to an error in the CICD system, eg. due to the worker being killed. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `failure` | The task run did not finish successfully, eg. due to a compile error or a failing test. Such failures are usually detected by non-zero exit codes of the tools executed in the task run. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `skip` | The task run was skipped, eg. due to a precondition not being met. | ![Development](https://img.shields.io/badge/-development-blue) |
```
kuisathaverat, line 60: `skip` or `skipped`

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,106 @@
...
+---
+
+`cicd.pipeline.result` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `cancellation` | The pipeline run was cancelled, eg. by a user manually cancelling the pipeline run. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `error` | The pipeline run failed due to an error in the CICD system, eg. due to the worker being killed. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `failure` | The pipeline run did not finish successfully, eg. due to a compile error or a failing test. Such failures are usually detected by non-zero exit codes of the tools executed in the pipeline run. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `skip` | The pipeline run was skipped, eg. due to a precondition not being met. | ![Development](https://img.shields.io/badge/-development-blue) |
```
kuisathaverat, line 55: `skip` or `skipped`

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,106 @@
...
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`cicd.pipeline.task.name`](/docs/attributes-registry/cicd.md) | string | The human readable name of a task within a pipeline. Task here most closely aligns with a [computing process](https://wikipedia.org/wiki/Pipeline_(computing)) in a pipeline. Other terms for tasks include commands, steps, and procedures. | `Run GoLang Linter`; `Go Build`; `go-test`; `deploy_binary` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`cicd.pipeline.task.run.id`](/docs/attributes-registry/cicd.md) | string | The unique identifier of a task run within a pipeline. | `12097` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`cicd.pipeline.task.run.result`](/docs/attributes-registry/cicd.md) | string | The result of a task run. | `success`; `failure`; `timeout`; `skipped` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
```
kuisathaverat, line 77: `cancelation` and `error` are missing, and `required` is not described below

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,106 @@
...
+---
+
+`cicd.pipeline.task.run.result` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `cancellation` | The task run was cancelled, eg. by a user manually cancelling the task run. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `error` | The task run failed due to an error in the CICD system, eg. due to the worker being killed. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `failure` | The task run did not finish successfully, eg. due to a compile error or a failing test. Such failures are usually detected by non-zero exit codes of the tools executed in the task run. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `skip` | The task run was skipped, eg. due to a precondition not being met. | ![Development](https://img.shields.io/badge/-development-blue) |
```
kuisathaverat, line 97: `skip` or `skipped`

docs/resource/cicd.md
```diff
@@ -0,0 +1,191 @@
...
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+**type:** `cicd.pipelinerun`
```
kuisathaverat, line 64:
```suggestion
**type:** `cicd.pipeline.run`
```

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,106 @@
...
+<!-- semconv span.cicd.pipeline.run.server -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`cicd.pipeline.result`](/docs/attributes-registry/cicd.md) | string | The result of a pipeline run. | `success`; `failure`; `timeout`; `skipped` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
```
kuisathaverat, line 39: I do not see `cancelation` and `error` described below

docs/resource/cicd.md
```diff
@@ -0,0 +1,191 @@
...
+|---|---|---|---|---|---|
+| [`vcs.ref.head.name`](/docs/attributes-registry/vcs.md) | string | The name of the [reference](https://git-scm.com/docs/gitglossary#def_ref) such as **branch** or **tag** in the repository. [1] | `my-feature-branch`; `tag-1-test` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`vcs.ref.head.revision`](/docs/attributes-registry/vcs.md) | string | The revision, literally [revised version](https://www.merriam-webster.com/dictionary/revision), The revision most often refers to a commit object in Git, or a revision number in SVN. [2] | `9d59409acf479dfa0df1aa568182e43e43df8bbe28d60fcf2bc52e30068802cc`; `main`; `123`; `HEAD` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`vcs.ref.type`](/docs/attributes-registry/vcs.md) | string | The type of the [reference](https://git-scm.com/docs/gitglossary#def_ref) in the repository. | `branch`; `tag` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
kuisathaverat, line 162: A commit sha, Could be a type?

model/cicd/resources.yaml
```diff
@@ -0,0 +1,31 @@
...
+    name: cicd.pipeline
+    brief: >
+        A pipeline is a series of automated steps that helps software teams deliver code.
+    attributes:
+      - ref: cicd.pipeline.name
+
+  - id: resource.cicd.pipelinerun
+    type: resource
+    stability: development
+    name: cicd.pipelinerun
```
kuisathaverat, line 14:
```suggestion
    name: cicd.pipeline.run
```

docs/attributes-registry/cicd.md
joaopgrassi: Also `cancelled` instead of `cancellation`?

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,106 @@
...
+  - [Pipeline run](#pipeline-run)
+  - [Pipeline task run](#pipeline-task-run)
+
+<!-- tocstop -->
+
+## CICD Spans
+
+The conventions described in this section are specific to Continuous Integration / Continuous Deployment (CICD) systems.
+
+**Disclaimer:** These are initial CICD spans and attributes
```
joaopgrassi, line 21: Maybe we don't need this disclaimer. I believe this is somewhat implicit by the document status.

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,106 @@
...
+but more may be added in the future.
+
+Also consider setting other attributes of the [CICD and VCS resource conventions][cicdres].
```
joaopgrassi, line 24: IMO If we think those are valuable on the spans here, we should add them here with proper requirement levels instead of being generic like this.

docs/cicd/cicd-spans.md
kamphaus: Fixed in bbce19df8fecc23c220c58ca8f73f9c80a73361a

docs/attributes-registry/cicd.md
kamphaus: This new attribute is based on `cicd.pipeline.result` with the descriptions adapted to refer to task run instead of pipeline run.
For completeness and to remain aligned with `cicd.pipeline.result` I would keep `cancellation` and `error`.

docs/attributes-registry/cicd.md
kamphaus: We had this discussion in a previous PR: https://github.com/open-telemetry/semantic-conventions/pull/1681#discussion_r1907454943

We decided to standardize on all result values being nouns.

docs/cicd/cicd-spans.md
kamphaus: It's described a few lines below: https://github.com/open-telemetry/semantic-conventions/blob/bbce19df8fecc23c220c58ca8f73f9c80a73361a/docs/cicd/cicd-spans.md?plain=1#L43-L52

Do you mean that the description should be improved?

docs/cicd/cicd-spans.md
kamphaus: These are just examples and not an exhaustive list.
A list of well-known values is shown a few lines below:
https://github.com/open-telemetry/semantic-conventions/blob/bbce19df8fecc23c220c58ca8f73f9c80a73361a/docs/cicd/cicd-spans.md?plain=1#L84-L93

Is there a better way to point this list out?

docs/resource/cicd.md
kamphaus: The `vcs.ref.type` refers to the type of `vcs.ref.head.name` whether it refers to a branch or a tag.
The commit sha is stored in `vcs.ref.head.revision`.

Do you mean that `sha` could be an additional value for `vcs.ref.type`?

docs/resource/cicd.md
kamphaus: Renamed in 77cca2d9657f7c4b02bfafdc6c444be5aafd7a30

model/cicd/resources.yaml
kamphaus: Renamed in 77cca2d9657f7c4b02bfafdc6c444be5aafd7a30

docs/cicd/cicd-spans.md
kamphaus: Removed in 41762ab10dc5eb134c55bd84bedab0dafd8ae9a3

docs/cicd/cicd-spans.md
kamphaus: Let's reword this to make the intent clearer.

We want to reference the relevant resource conventions to be used:
> Any resources of the [CICD and VCS resource conventions] that apply SHOULD be used.

Do we want to allow setting the attributes from the resources as span attributes as well (eg. if for technical reasons they cannot be set as resources)?
We could add the following:
> If resources cannot be used for technical reasons, the resource attributes MAY be set as span attributes.

docs/resource/cicd.md
kuisathaverat: Yes, it the case `vcs.ref.head.revision` is a `sha` How we inform `vcs.ref.type`?  maybe `sha` or `id` or something along these lines

docs/cicd/cicd-spans.md
kuisathaverat: For some reason, I thought it was a complete list of values instead of a sample

christophe-kamphaus-jemmic: Is it good as it is or should I improve the description?

docs/resource/cicd.md
christophe-kamphaus-jemmic: The `vcs.ref.head.revision` should always be set to the most precise revision which in most cases is the sha.

Since `vcs.ref.type` refers to the type of `vcs.ref.head.name` I would not add a value for sha unless `vcs.ref.head.name` itself contains a sha.
If `vcs.ref.head.name` is not set then I would leave `vcs.ref.type` unset as well even if `vcs.ref.head.revision` contains a sha.

kuisathaverat (approved):

model/cicd/resources.yaml
```diff
@@ -0,0 +1,31 @@
+groups:
+  - id: resource.cicd.pipeline
+    type: resource
+    stability: development
+    name: cicd.pipeline
+    brief: >
+        A pipeline is a series of automated steps that helps software teams deliver code.
+    attributes:
+      - ref: cicd.pipeline.name
```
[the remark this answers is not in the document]
kamphaus: Since the `cicd.pipeline.run.*` attributes would be shared across all spans of a pipeline run I thought it would be more efficient to have them as a `cicd.pipeline.run` resource. This indeed means we need a separate resource per pipeline run. Is there a downside to having a resource per pipeline run?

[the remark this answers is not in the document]
kamphaus: How does where the spans are emitted relate to whether we are using different resources or span attributes?
Do child spans inherit the resources of their parent span or do they have their own resources?

kamphaus:
I have now also added guidance on span kind and name for pipeline runs.

model/cicd/resources.yaml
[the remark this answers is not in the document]
kamphaus: > Indeed I can improve the description to make this clearer.

I took a stab at improving the description for pipeline run resource:

b901fafb46dcdec1e48cc227c10207a6c61d1e33 and 6b406541979442a1b3914822cbe5516bd57f671e

docs/resource/cicd.md
```diff
@@ -0,0 +1,195 @@
...
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### CICD pipeline run
+
+To efficiently set the following attributes for all spans of a pipeline run they should be defined as the following resource.
+
+Since this resource is unique for each pipeline run, this means instantiating separate [TracerProviders](https://opentelemetry.io/docs/specs/otel/trace/api/#tracerprovider) for each pipeline run in order to be able to configure the resources of any spans emitted through that tracer.
```
kamphaus, line 56: Related to https://github.com/open-telemetry/semantic-conventions/issues/1111 https://github.com/open-telemetry/semantic-conventions/issues/1184#issuecomment-2773594630, we should add a warning here that if the `CICD pipeline run` resource is used for metrics, then it might cause issues with high-cardinality.

kamphaus: I added a warning in 7f6b5b38b0620ea6ef75cfe24bffa4aa2b95cab7.

docs/resource/cicd.md
```diff
@@ -0,0 +1,197 @@
...
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## Version Control System (VCS)
+
+A Version Control System tracks changes to source code.
+
+### VCS repository
```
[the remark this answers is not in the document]
kamphaus: The discussion https://github.com/open-telemetry/semantic-conventions/pull/2013#discussion_r2016086021 about needing to instantiate several TracerProviders also applies to the VCS resource especially for vcs.ref

I would prefer to resolve this discussion in this PR for both cicd and vcs resources.

What benefit is there in splitting off the vcs part? Which parts of this PR relating to VCS do you propose to split off in its own PR?

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,108 @@
...
+
+[cicdres]: /docs/resource/cicd.md (CICD and VCS resource conventions)
+
+### Pipeline run
+
+For all pipeline runs, a span with kind `SERVER` SHOULD be created corresponding to the execution of the pipeline run.
+
+The span name MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.43.0/specification/trace/api.md#span).
+
+The span name SHOULD be `{action} {pipeline}` if there is a (low-cardinality) `pipeline` available.
```
adrielp, line 31: I'm not entirely following this section when reading it in isolation. What is `{action}` and what is `{pipeline}` -- aka, where do those come from? 

Additionally, I don't immediately understand the point around low-cardinality in the names and why that matters. 

In my mind, if we made this the convention, then the `{action}` described below would require everyone to update their pipelines to support conventions? In the current pipeline, the span names map directly to the `cicd.pipeline.{run, task}.name`. This makes it easy to read in the visualizing backend, but also reference by attribute.

Thoughts on the path here?

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,108 @@
...
+|---|---|---|---|---|---|
+| [`cicd.pipeline.result`](/docs/attributes-registry/cicd.md) | string | The result of a pipeline run. | `success`; `failure`; `timeout`; `skipped` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+---
+
+`cicd.pipeline.result` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `cancellation` | The pipeline run was cancelled, eg. by a user manually cancelling the pipeline run. | ![Development](https://img.shields.io/badge/-development-blue) |
```
adrielp, line 55: Something that we experienced in the GitHub receiver for skip & cancelled pipelines is crazy durations on the spans set improperly. This is something messed up on the event side, but it makes me wonder if we should provide guidance around cancelled, skipped, and timedout pipeline span durations, eg.

cancelled pipelines should have a duration of cancelled at - started at when available, otherwise a default value of 0 should be used for the duration.

skipped pipelines should have a duration of 0 when emitted as a span.

timed out should be the same as cancelled.

docs/resource/cicd.md
```diff
@@ -0,0 +1,197 @@
...
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### CICD pipeline run
+
+To efficiently set the following attributes for all spans of a pipeline run they should be defined as the following resource.
+
+Since this resource is unique for each pipeline run, this means instantiating separate [TracerProviders](https://opentelemetry.io/docs/specs/otel/trace/api/#tracerprovider) for each pipeline run in order to be able to configure the resources of any spans emitted through that tracer.
+
+Using the CICD pipeline run resource with metrics inherently causes high cardinality and might cause issues with some metric storage backends. Thus using CICD pipeline run resource for metrics MUST be opt-in.
```
adrielp, line 58:
```suggestion
Using the CICD pipeline run resource with metrics inherently causes high cardinality and may increase costs with some metric storage backends. Thus, using CICD pipeline run resource for metrics MUST be opt-in.
```

Two things:

1) We probably shouldn't have the cicd.pipeline.name attribute on every metric. 
2) We should provide guidance around attaching the metrics directly to the spans?

model/cicd/resources.yaml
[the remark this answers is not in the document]
adrielp: They inherit a copy and update. 

One thing to think about is the definition of cicd.pipeline.name. There are layers here, where in the example of GitHub, a workflow is a pipeline, a job is a pipeline, and the job contains the tasks. 

Basically the workflow is a pipeline w/ many child pipelines. Maybe updating the attribute description and guidance will make it a little more clear in certain contexts.

docs/cicd/cicd-spans.md
adrielp: This has been addressed. It was actually a ME problem (i wasn't setting an end span time in steps 🤦🏼  and that didn't get caught). 

However, I have added dedicated queue spans to jobs: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39511

This has added an attribute `cicd.pipeline.run.queue.duration` (which is basically a metric). 

We also did some queue span modeling, where the `run, queue, etc` prefix's come into a little more. I'm glad they're not mandatory though, as it's not possible to tell what is really a build (it's CI, but not necessarily a build, could be a lint, a test, a format, etc).

docs/resource/cicd.md
kamphaus: I will update the PR with this suggested change...

> We probably shouldn't have the cicd.pipeline.name attribute on every metric

This PR does not add guidance which resource to attach to which signal.
It only says for spans "Any resources of the CICD and VCS resource conventions that apply SHOULD be used."

The cicd pipeline names may also cause cardinality issues if less of a problem than for pipeline runs.
Should I also add a similar warning for the pipeline resource "Using the CICD pipeline resource with metrics may cause issues with high cardinality..."?

> We should provide guidance around attaching the metrics directly to the spans?

What do you mean with attaching the metrics to the span?

docs/resource/cicd.md
```diff
@@ -0,0 +1,197 @@
...
+
+<!-- tocstop -->
+
+See also:
+
+- The [Trace semantic conventions for CICD](/docs/cicd/cicd-spans.md)
+
+## Continuous Integration and Continuous Deployment (CICD)
+
+CICD aims to increase early defect discovery, increase productivity, and provide faster release cycles through automating builds, tests and deployment of software.
```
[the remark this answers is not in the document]
kamphaus: Not really, I will remove it...

docs/cicd/cicd-spans.md
kamphaus: I didn't get the point on what should be changed in this PR.

> This has been addressed.

I will mark this conversation as resolved.
Please clarify if changes related this discussion should be made.

model/cicd/resources.yaml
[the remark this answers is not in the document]
kamphaus: As I understand your point is that the current definition of pipeline does not map cleanly to the GitHub workflow / job concepts.

How could this be improved?

Would adding an attribute `cicd.pipeline.group` to group multiple pipelines together help (this would be in addition to grouping pipelines via `vcs.repository.name`)?

docs/cicd/cicd-spans.md
kamphaus: I was basing myself on `http-spans.md` to define the span name, hence splitting the name into action and object on which the action is performed.

As mentioned a few lines below `{pipeline}` is the same as the `cicd.pipeline.name` attribute.

The low-cardinality phrase also comes from `http-spans.md` where it refers to the span name usually being the http controller which is usually the URL path. The low cardinality there is to exclude path parameters from being included in the span name. The low-cardinality is required so meaningful statistics per span name can be calculated.

For cicd I can't think of any similar issue as with http path parameters, but I think it makes sense to keep the low-cardinality phrasing just in case.

In my opinion it's good to have `{action}` since different cicd systems might want to have different names for the action (eg. `SYNC` vs `BUILD`). Also the same CICD might have different actions for the same pipeline (eg. `BUILD` vs `UPDATE`).

In my opinion it makes sense to have the `{action} {pipeline}` definition for span names since this aligns with several conventions for span names (http, gen-ai, database, faas, messaging).

docs/resource/cicd.md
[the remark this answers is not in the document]
kamphaus: Removed in 29f122accda138bed61d214e2333efbf2e82e6bf.

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,124 @@
...
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+Semantic Convention for tasks in a pipeline run
+
+**Span kind** SHOULD be `SERVER`.
```
cyrille-leclerc, line 88: In Jenkins, build steps, which seem to map to OTel tasks, use `SpanKind=Internal`. Setting them as `SpanKind=Server` could look weird in the trace view and would require to have a `SpanKind=client to trigger those tasks.`

docs/cicd/cicd-spans.md
cyrille-leclerc: In Jenkins we use `BUILD {pipeline}`

kamphaus: CICD SIG meeting notes:

Use same language for action as for other enums -> open-ended.
Eg. tool chains can use their own custom actions (`scan`, `test`, ...).

docs/cicd/cicd-spans.md
kamphaus: Note: if a pipeline span is of kind server and is triggered by another pipeline, then a server span could be a child span of an internal span.

kamphaus: I changed span kind to INTERNAL for tasks in 9a0eaabb6f1b5a4ee24c1a5fd750fb124daf184f.

Considering https://opentelemetry.io/docs/concepts/signals/traces/#span-kind for the span kind of pipeline runs I'm not sure any fit perfectly. I'm tending towards `CONSUMER` due to the potential asynchronous processing of jobs however `SERVER` could also apply for jobs that are triggered and executed synchronously by other pipelines.
To be discussed in SemConv SIG...

docs/cicd/cicd-spans.md
kamphaus: I updated the language to be open-ended for `{action}` in 97b023fa9d79fd79e1b61d6ed1c3f381b3f7c388.

kamphaus: ### Discussed in SemConv SIG 2025-03-28

When something is unbonded then it is high-cardinality. Ie if it has the potential to be infinite / limitless.
An example of unboundedness and thus high cardinality would be user-provided content like the URL path.

Having a lot of combination alone is not high cardinality. It's the fact that they are unbounded that that would be high cardinality. But even if you have, eg. 300000, but if it's constantly that number, they don't go over, then it might be okay.

Another rule of thumb: if it's low-cardinality enough to use as a metric attribute, then it's good enough to use as a span name.

---

All discussion points in this conversation look resolved to me.
Feel free to reopen if I missed something.

docs/cicd/cicd-spans.md
kamphaus: ### Discussed in SemConv SIG 2025-04-28

The span kind is used for

* displaying the span differently in observability backends
* [service graph connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/servicegraphconnector/README.md) for emitting metrics representing the service graph based on client/server and producer/consumer span pairs
* critical path analysis
  * to determine whether synchronous client/server calls are on the critical path, while asynchronous producer/consumer calls might be off the critical path
  * for cicd it might be more appropriate to consider every job even if triggered asynchronously because the job might only be done when every triggered pipeline is done (ie by using client/server span kind)

The client/server for synchronous and producer/consumer for asynchronous spans terminology does not work well for CICD because some pipeline runs might be triggered synchronously while others might be triggered by an event and execution might be asynchronous.

We should consistently use either Client/Server or Producer/Consumer span pairs.
Or stay with internal if we don't want to make any assumptions.

After some discussion the conclusion was to stay with server spans for pipeline runs for now.

kamphaus: Regarding the inconsistency: the step triggering the child pipeline should be of span kind client, then it's fine.
It's possible for the client span to be much shorter in the case of asynchronous triggering of the child pipeline or for the client span to be larger than the child pipeline run span in case of synchronous execution of the child pipeline.

adrielp (approved):

model/cicd/spans.yaml
```diff
@@ -0,0 +1,26 @@
...
+    span_kind: server
+    brief: 'Semantic Convention for pipeline runs'
+    stability: development
+    attributes:
+      - ref: cicd.pipeline.result
+        requirement_level: required
+  - id: span.cicd.pipeline.task.internal
+    type: span
+    span_kind: internal
+    brief: 'Semantic Convention for tasks in a pipeline run'
```
lmolkova, line 13:
```suggestion
    brief: This span describes CICD pipeline task execution
```

model/cicd/spans.yaml
```diff
@@ -0,0 +1,26 @@
+groups:
+  - id: span.cicd.pipeline.run.server
+    type: span
+    span_kind: server
+    brief: 'Semantic Convention for pipeline runs'
```
lmolkova, line 5:
```suggestion
    brief: This span describes CICD pipeline run
```

model/cicd/spans.yaml
```diff
@@ -0,0 +1,26 @@
+groups:
+  - id: span.cicd.pipeline.run.server
+    type: span
```
lmolkova, line 3: please add entity/resource associations similar to https://github.com/open-telemetry/semantic-conventions/pull/2123 for the resources you expect to be populated

model/cicd/spans.yaml
```diff
@@ -0,0 +1,26 @@
...
+    span_kind: server
+    brief: 'Semantic Convention for pipeline runs'
+    stability: development
+    attributes:
+      - ref: cicd.pipeline.result
```
lmolkova, line 8: please also add `error.type` and describe which `cicd.pipeline.result` should be reported as errors

model/cicd/spans.yaml
```diff
@@ -0,0 +1,26 @@
+groups:
+  - id: span.cicd.pipeline.run.server
+    type: span
+    span_kind: server
+    brief: 'Semantic Convention for pipeline runs'
+    stability: development
```
lmolkova, line 6: could you please detailed description you have in markdown for this span into the yaml? 

See https://github.com/open-telemetry/semantic-conventions/blob/08728eb08de176cb5c0616022feabb5018741e00/model/database/spans.yaml#L110 as an example.

It helps us keep all these notes at least available for the tooling.

Also, same concern on the span.cicd.pipeline.task.internal

model/cicd/spans.yaml
```diff
@@ -0,0 +1,26 @@
...
+    attributes:
+      - ref: cicd.pipeline.result
+        requirement_level: required
+  - id: span.cicd.pipeline.task.internal
+    type: span
+    span_kind: internal
+    brief: 'Semantic Convention for tasks in a pipeline run'
+    stability: development
+    attributes:
+      - ref: cicd.pipeline.task.name
```
lmolkova, line 16: please add error.type

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,124 @@
...
+### Pipeline run
+
+For all pipeline runs, a span with kind `SERVER` SHOULD be created corresponding to the execution of the pipeline run.
+
+The span name MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.43.0/specification/trace/api.md#span).
+
+The span name SHOULD be `{action} {pipeline}` if there is a (low-cardinality) `pipeline` available.
+If there is no (low-cardinality) `{pipeline}` available, the span name SHOULD be `{action}`.
+
+The `{action}` SHOULD be one of the following values: `BUILD`, `RUN`, `SYNC`. If none of these values applies a custom value MAY be used.
```
lmolkova, line 34: any reason for ALLCAPS? Also, would we have action appearing on metrics in some shape?  We probably want a dedicated attribute for it.

docs/cicd/cicd-spans.md
```diff
@@ -0,0 +1,124 @@
...
+
+The `{pipeline}` SHOULD be the [`cicd.pipeline.name`](/docs/attributes-registry/cicd.md).
```
lmolkova, line 36: nit: do you need a placeholder for `{pipeline}` given it's always `cicd.pipeline.name` ?

```suggestion
The span name SHOULD be `{action} {cicd.pipeline.name}` if there is a (low-cardinality) pipeline name available.
If pipeline name is not available or is likely to have high cardinality, the span name SHOULD be `{action}`.

The `{action}` SHOULD be one of the following values: `BUILD`, `RUN`, `SYNC`. If none of these values applies a custom value MAY be used.
```

docs/cicd/cicd-spans.md
kamphaus: I used ALLCAPS because I based this on the the http semconv which uses the http method (ie `GET`, `POST`, ...) and because Jenkins already uses `BUILD` in its pipeline run span name.

I think it makes sense to use ALLCAPS because it clearly distinguishes the action from the pipeline name.

I will make it an enum attribute, since it also makes it easier to have consistent language for the listed values and add additional ones in the future.
I don't think we will have this attribute on metrics.

model/cicd/resources.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+  - id: resource.cicd.pipeline.run
+    type: resource
```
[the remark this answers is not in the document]
kamphaus: I will resolve this conversation since there is nothing to be changed.

docs/cicd/cicd-spans.md
kamphaus: I made it similar to the http semconv where such placeholders are used.

Since we refer to several times (`if ... then ...`) it's more readable to use this placeholder instead of writing out the full attribute name.

model/cicd/registry.yaml
```diff
@@ -86,6 +86,40 @@ groups:
           [
             "https://github.com/open-telemetry/semantic-conventions/actions/runs/9753949763/job/26920038674?pr=1075",
           ]
+      - id: cicd.pipeline.task.run.result
```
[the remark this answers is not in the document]
kamphaus: Discussed with Adriel:
We do not expect both attributes to be set on the same span.
the cicd.pipeline.result would only be set on a pipeline run span, while the cicd.pipeline.task.run.result is only set on a task run span.
There a pros and cons to leaving it separate or combining them: leaving them separate makes analysis easier, while having them combined requires the necessity to distinguish between task and pipeline run spans during analysis.
I have created a follow-up issue to continue the discussion for combining both attributes:
https://github.com/open-telemetry/semantic-conventions/issues/2208

model/cicd/resources.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - ref: cicd.pipeline.run.url.full
+
+  - id: resource.cicd.worker
+    type: resource
+    stability: development
+    name: cicd.worker
+    brief: >
+      A CICD worker is a component of the CICD system that performs work (eg. running pipeline tasks or performing sync).
+    attributes:
+      - ref: cicd.worker.id
```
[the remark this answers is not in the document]
kamphaus: Sometime the worker might not have a url (eg. github hosted runners), that's why I made the url attribute opt-in.
The worker id could be the host.id, a pod name or some other value like the agent public IP allowing to identify the agent.
The worker name is more descriptive and could be shared among several workers (eg. `Ubuntu`, `Mac OS`).
I will improve the examples to make this clearer and make the worker.id required.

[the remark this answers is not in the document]
kamphaus: Done in 2e552848c97d9e547843e0c605ef56aa1c098113.

model/cicd/spans.yaml
kamphaus: Done in 637ee78fcca2927a7969eb5abacc8de0de1d690f.

model/cicd/spans.yaml
kamphaus: Done in 9298f48736e4b56fa3678c241e304c4412fa428a.

model/cicd/spans.yaml
kamphaus: Done in 4a803cee2e1de393676d7d2b9e5b9cd650f07d40.

model/cicd/spans.yaml
kamphaus: Done in aa5e4771d592943d8e3caa8750b898e408ff66bd.

model/cicd/spans.yaml
kamphaus: Done in e41f4f520fb814e4545591cfa7c27ac6c47476e0.

There was no markdown for cicd.pipeline.task to be moved to the yaml.
I will keep it in mind if we decide to add some later.

docs/cicd/cicd-spans.md
kamphaus: I improved the note like this in a8d7dea51d776dab86d45b042cbc82c5bb693dc3.

docs/cicd/cicd-spans.md
kamphaus: Made `{action}` an attribute: 3cefced99c742652925188a3acbf1bfea858297f

model/cicd/spans.yaml
kamphaus: I added entity associations in 7f81a46a421b6a245446f68bab7b7c874dde7e43.

model/cicd/registry.yaml
```diff
@@ -139,6 +173,54 @@ groups:
         brief: >
           The result of a pipeline run.
         examples: ["success", "failure", "timeout", "skipped"]
+      - id: cicd.pipeline.action
```
lmolkova, line 176:
```suggestion
      - id: cicd.pipeline.action.name
```

we usually end up needing to define more properties, so we recommend 

https://github.com/open-telemetry/semantic-conventions/blob/1000be4d897fafbc8554b373267992e47c1b86ac/docs/general/naming.md?plain=1#L76-L78

model/cicd/resources.yaml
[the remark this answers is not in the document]
lmolkova: thanks! 

> Sometime the worker might not have a url (eg. github hosted runners), that's why I made the url attribute opt-in.

just checking if you want url to be opt-in, you can absolutely make it `recommended: when available` - either way works for me

model/cicd/spans.yaml
```diff
@@ -0,0 +1,45 @@
...
+      For all pipeline runs, a span with kind `SERVER` SHOULD be created corresponding to the execution of the pipeline run.
+
+      **Span name** MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.43.0/specification/trace/api.md#span).
+
+      The span name SHOULD be `{action} {pipeline}` if there is a (low-cardinality) pipeline name available.
+      If the pipeline name is not available or is likely to have high cardinality, then the span name SHOULD be `{action}`.
+
+      The `{action}` SHOULD be the [`cicd.pipeline.action`](/docs/attributes-registry/cicd.md#cicd-pipeline-action).
+
+      The `{pipeline}` SHOULD be the [`cicd.pipeline.name`](/docs/attributes-registry/cicd.md#cicd-pipeline-name).
```
lmolkova, line 16: moving https://github.com/open-telemetry/semantic-conventions/pull/2013#discussion_r2067121403 discussion here, why not

```suggestion
      The span name SHOULD be `{cicd.pipeline.action} {cicd.pipeline.name}` if there is a (low-cardinality) pipeline name available.
      If the pipeline name is not available or is likely to have high cardinality, then the span name SHOULD be `{cicd.pipeline.action}`.
```

in HTTP there are multiple options for each component, which is not the case here.

Check out DB : https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#name

lmolkova (commented):
Looks great!

I left some comments, I'd love to resolve the `cicd.pipeline.action.name` one, the rest is non-blocking

model/cicd/registry.yaml
kamphaus: I did the rename in a9f2be76b22553df3e633e404d676bcbc4c288a5.

model/cicd/resources.yaml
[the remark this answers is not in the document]
kamphaus: Thanks, I improved this in 39133d85b6ac900664e2364dce58da1a26de5b0d.

model/cicd/spans.yaml
kamphaus: Having tried this in ac67d6bf457a90fa80228b3aeefee9c8efbf40d0 I noticed that the markdown did not display the link to the attribute registry well.
I thus reverted it in 61f48f92df41f2c94e96bb9269d1b5f5a32ae99e since I preferred to have the link to the attribute registry.

kamphaus:
I missed to update the changelog after my recent changes, I thus made the updates in 787e908321c0d130e1acbd01990ef8d0f1e8f45a, e13e5f096b3abee3b1c5cbfded7aa70a01e1ce48 and 14b1f5a254c4e9734767325657c56d875e784b43.

lmolkova (approved):

Issue #1961, opened 2025-03-05
Add MultiAgent Semantic Conventions


PRATIBHA-Moogi:
### Area(s)

area:gen-ai

### **Is your change request related to a problem? Please describe.**

Based on the work on "A Taxonomy of AgentOps for Enabling Observability of
Foundation Model based Agents"  we need to have coverage on all the attributes of MultiAgentic systems which govern Agenitc Behaviour & its performance in a holistic manner.
Ref Publication [https://arxiv.org/html/2411.05285v1]

Based on the taxonomy published by the above work, MultiAgentic systems can follow different topologies or workflows defined over multiple agents (a set of experts). 

Task can be composed over a specific Agent call or a set of Agents, with a Gen AI model, given Task description, given expected output specification etc. It can also follow a specific topology that of defining a DAG or a workflow over multiple agents.

Here, Task key attributes such as Task description, Agents involved, expected_output can be added to the semantic conventions to well define Task resources.

Each Agent further can be composed with a bundle of tools required to carry out Agent's goal given its role, given a prompt_template, in-context info.  So Agent Goal, Role, Task, Tool, Expectated output can characterise individual Agents attributes and can be added to the semantic conventions for covering Agent resources.

Each tool also can be characterised with a set of attributes - tool_type, tool_description, tool_argument etc. So those attributes can also be added into semantic conventions.

Each prompt can also be described by means of prompt_template_type, prompt_template_info. And further to that other constructs of prompt_template_info - User Goal, Instruction, Query, Few-Shot examples, Output format ask, Tools List,  ChatHistory etc can also be added as a set of attributes to characterise Prompts.

The above some of the key MultiAgentic System attributes, if added to the semantic conventions, can further enable rich set of Observability views as the below ;

1. Discovery of MultiAgentic system topology view
2. Discovery of Tasks, Task-Agent relationship
3. Discovery of Agent-Tool relationship
4. Discovery of Agent-Prompt attributes relationship
5. Discovery of Tools usage, Tools Types & associated performance metrics
6. Correlations of Agent's performance with Prompts attributes
7. System level metrics associated at each components of MultiAgentic System

**### ##Describe the solution you'd like**

A new semantic conventions with the key attributes to describe multiagentic systems / applications.

/cc @gyliu513

gyliu513:
Related with https://github.com/open-telemetry/semantic-conventions/issues/1530

TaoChenOSU:
[...]

Issue #1688, opened 2024-12-16
Unified semantic conventions for tasks, workflows, pipelines, jobs


svrnm:
### Area(s)

area:new

### Is your change request related to a problem? Please describe.

While reviewing the [AI Agent Span Semantic Convention](https://github.com/open-telemetry/semantic-conventions/pull/1657) I made the comment, that the definition of `ai_agent.workflow.*` and `ai_agent.task.*` should not be unique, since there tasks & workflows that can be modeled similarly outside of the `ai_agent` scope. Same is true for the existing experimental [CICD pipeline](https://opentelemetry.io/docs/specs/semconv/attributes-registry/cicd/#cicd-pipeline-attributes) attributes. Other examples might be cronjobs, business processes, build tools, like make or goyek, which defines their own `goyek.flow.*` tasks in https://pkg.go.dev/github.com/goyek/x/otelgoyek#example-package (thanks to @pellared for pointing me to goyek)

### Describe the solution you'd like

The solution I would like to see is a unified set of attributes that describe such "workflow" and then only to have the unique attributes in the related specifications, similar as we have it today for the HTTP SemConv, where `server.*` or `url.*` attributes are used where applicable.

So instead of the current proposals, the future _**may**_ look like the following:

* cicd.pipeline.name => workflow.name
* cicd.pipeline.run.id => workflow.run.id
* cicd.pipeline.task.name => workflow.task.name
* cicd.pipeline.task.run.id => workflow.task.run.id
* cicd.pipeline.task.run.url.full => workflow.task.url.full (or just url.full if it is a "task span"?) 
* cicd.pipeline.task.type => workflow.task.type
* ...

and

* ai_agent.workflow.name => workflow.name
* ai_agent.task.name => workflow.task.name
* ai_agent.task.output => workflow.task.output
* ...

and

* goyek.flow.output => workflow.output
* ...

Both @open-telemetry/semconv-genai-approvers @open-telemetry/semconv-cicd-approvers are very active working groups, pushing semantic conventions forward, and it would be great to see some broader thinking about "workflows" (or however this should be named in a unified way)

### Additional context

Note, that this may also require to make progress on the long standing question of "long running 'spans'"[^1]:

https://github.com/open-telemetry/opentelemetry-specification/issues/373,
https://github.com/open-telemetry/opentelemetry-specification/issues/2692 and other previous issues touched upon this topic, but so far there is no solution for long-running (++minutes, hours, days) or even "infinite" spans

[...]

PR #2014, opened 2025-03-25
added multiagent semantic conventions along with single agent additio…

PRATIBHA-Moogi:
To address - https://github.com/open-telemetry/semantic-conventions-genai/issues/24#issue-2897759988. 
This PR addresses the need for defining multiagent semantic conventions

Fixes # NIL

## Changes

Added multiagent application key attributes in addition to that added extra attributes for single agent application.

### DIFF
--- a/model/gen-ai/registry.yaml
+++ b/model/gen-ai/registry.yaml
@@ -202,6 +202,41 @@ groups:
         type: string
         brief: Free-form description of the GenAI agent provided by the application.
         examples: ["Helps with math problems", "Generates fiction stories"]
+      - id: gen_ai.agent.tools
+        stability: development
+        type: string[]
+        brief: Lists the name of the tools configured with agent.
+        example: ["search-web","summarise-searched-results","publish-results"]
+      - id: gen_ai.agent.llm
+        stability: development
+        type: string
+        brief: Name of the llm configured with Agent application.
+        example: ["OpenAI", "Ollama", "Anthropic"]
+      - id: gen_ai.multiagent.id
+        stability: development
+        type: string
+        brief: The unique identifier of the GenAI multiagent.
+        example: ['multiasst_5j66UpCpwteGg4YSxUnt7lPY']
+      - id: gen_ai.multiagent.name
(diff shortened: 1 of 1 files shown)


lmolkova (changes requested):
I don't believe we're ready to define multi-agent semantic conventions. We don't even have a clear definition of a single agent ones, especially when it comes to client-side agent frameworks.

Multi-agent support is tight to general-purpose workflow #1688 and as suggested in the https://github.com/open-telemetry/semantic-conventions/issues/1961#issuecomment-2741220343 may need a few attributes and events or spans defined (e.g. for transition between agents).

From general semconv perspective, we [don't encourage](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/how-to-define-semantic-conventions.md) defining attributes without a corresponding span/event/metric or at least an instrumentation that would use them. 

I don't believe these new attributes would apply to existing GenAI spans the conventions cover. 

TL;DR: The suggestion is:
- please help us finalize  conventions for a single agent 
- if you're interested in multi-agent conventions:
  - consider how to express it in a more general way - any workflow involving agents, humans, or non-AI bots would have similar characteristics as multi-agent one
  - it's best to start by prototyping the actual instrumentation and thinking about information that would be unique and essential for multi-agent flows.

PRATIBHA-Moogi:
Thanks for your comments @lmolkova . I am working on it along with @gyliu513 .  Regarding current gen_ai.agent.xx conventions, it doesn't represent a multiagent set of attributes. We can surely expand that one first.  I also find bit of overlap on different concepts such as Workflow, flows, tasks, agents, tools, as well as runnable etc. as expressed by different agentic-application-frameworks e.g. langchain/langgraph, crew.ai, autogen, IBM Bee framework, Llamaindex etc.  Is there we are maintaining a document to freeze on a common set of defnitions across these middleware framework ?

gyliu513:
@lmolkova totally agree with you, I will work with @PRATIBHA-Moogi to improve the single agent first, and then extend to multi-agent

Issue #2002, opened 2025-03-18
Reality-check if exporter health metrics should be merged across signals


JonasKunz:
This is issue is about whether the exporter health metrics should be merged across signals.
E.g. currently the exporter metrics are separated per signal:
* `otel.otel.sdk.exporter.log.exported.count`
* `otel.otel.sdk.exporter.span.exported.count`

The metric definitions are pretty much copy-paste, with only adjustments to the unit. The question is whether it makes more sense to combine them into a single metric (e.g. `otel.otel.sdk.exporter.exported.count`), differentiating the signal types using an attribute (`otel.signal`=`span/log/metric`).

Per discussion in the SIG Meeting on 17.03.2025 we decided to not merge them for now, even though their definitions are very similar. Some concerns where:
 * Confusion when building dashboards: Other health metrics would not have the `otel.signal` attribute, making it a bit more difficult/ error prone to build dashboards.
 * Non-intuitive metric units: One would need to check the exact metric defintion to understand that the unit is actually dependent on the `otel.singal` type
 * Risk of diverging: Exporters have dedicated specifications per signal. If those diverge, we would run into the risk of not being able to represent the different behaviours in the exporter metrics.

The decision therfore was to keep them separated for now. After the metrics have been implemented and used in the wild, we will revisit this decision if there are too many drawbacks discovered. We can then adjust, as the metrics are experimental.

trask:
I'm glad that `otel.sdk.exporter.operation.duration` is unified across signals (#2007).

Keeping the others separate makes sense to me, especially due to the difference in units and the differences between span/log pipelines (which are similar) and metric pipelines (which are fairly different and so may be hard to unify across all three).

joaopgrassi:
Should we close this as it's nothing is changing or we prefer to leave it open? Was there any new insights since this was opened?

JonasKunz:
I don't think the implementation of these metrics has come far enough to really make a decision, however so far I personally am still happy with the design. Imo we could close this and reopen if needed.

joaopgrassi:
Alright I will keep it open in the backlog then.

PR #2016, opened 2025-03-26
Metric health metrics

JonasKunz:
Adds health metrics for the SDK metric parts. Based on the span health metrics from #1631 and log health metrics from #1921.

## Changes

* Adds a new metric for the number of metric collections per metric-reader
* Adds exporter metrics for metric-exporters:  these are pretty much a copy-paste, which is in accordance with #2002.


model/otel/metrics.yaml
```diff
@@ -206,6 +206,63 @@ groups:
...
+      - ref: server.address
+        requirement_level:
+          recommended: when applicable
+      - ref: server.port
+        requirement_level:
+          recommended: when applicable
+      - ref: error.type
+        examples: ["rejected", "timeout", "500", "java.net.UnknownHostException"]
+
+  - id: metric.otel.sdk.metricreader.collection.duration
```
lmolkova, line 251: nit: based on the naming guidance, we should probably call it `metric_reader` since these are two inseparable words.

/cc @trask

JonasKunz: Fixed in https://github.com/open-telemetry/semantic-conventions/pull/2016/commits/39378f6fd0bd8f6abe9fc7b5e6122d26d4722e56

AlexanderWert (approved):

lmolkova (approved):

model/otel/metrics.yaml
```diff
@@ -206,6 +206,63 @@ groups:
       - ref: error.type
         examples: ["rejected", "timeout", "500", "java.net.UnknownHostException"]
 
+  - id: metric.otel.sdk.exporter.metric.inflight
+    type: metric
+    metric_name: otel.sdk.exporter.metric.inflight
+    stability: development
+    brief: "The number of metric data points which were passed to the exporter, but that have not been exported yet (neither successful, nor failed)"
+    note: |
+      For successful exports, `error.type` MUST NOT be set. For failed exports, `error.type` must contain the failure cause.
```
trask, line 215:
```suggestion
      For successful exports, `error.type` MUST NOT be set. For failed exports, `error.type` MUST contain the failure cause.
```

model/otel/metrics.yaml
```diff
@@ -206,6 +206,63 @@ groups:
...
+        requirement_level:
+          recommended: when applicable
+
+  - id: metric.otel.sdk.exporter.metric.exported
+    type: metric
+    metric_name: otel.sdk.exporter.metric.exported
+    stability: development
+    brief: "The number of metric data points for which the export has finished, either successful or failed"
+    note: |
+      For successful exports, `error.type` MUST NOT be set. For failed exports, `error.type` must contain the failure cause.
```
trask, line 234:
```suggestion
      For successful exports, `error.type` MUST NOT be set. For failed exports, `error.type` MUST contain the failure cause.
```

model/otel/registry.yaml
```diff
@@ -75,12 +75,12 @@ groups:
             - id: batching_log_processor
               value: "batching_log_processor"
               brief: >
-                The builtin SDK Batching LogRecord Processor
+                The builtin SDK Batching log_record Processor
```
trask, line 78: (and similar below if you agree with this)
```suggestion
                The builtin SDK batching log record processor.
```

model/otel/metrics.yaml
```diff
@@ -206,6 +206,63 @@ groups:
       - ref: error.type
         examples: ["rejected", "timeout", "500", "java.net.UnknownHostException"]
 
+  - id: metric.otel.sdk.exporter.metric.inflight
+    type: metric
+    metric_name: otel.sdk.exporter.metric.inflight
```
trask, line 211: I think current name could be confusing since we generally refer to a "metric" as a thing that has multiple timeseries (e.g. [in proto](https://github.com/open-telemetry/opentelemetry-proto/blob/42319f8b5bf330f7c3dd4a097384f9f6d5467450/opentelemetry/proto/metrics/v1/metrics.proto#L187))

what do you think of

- `otel.sdk.exporter.metric_data_point.inflight`, or
- `otel.sdk.exporter.timeseries.inflight`

JonasKunz: Good point, I like `otel.sdk.exporter.metric_data_point.inflight`. This way the metric will still show up when searching for `metric` and therefore makes it more easily discoverable than `timeseries`.

trask (approved):

PR #2037, opened 2025-03-29
Add `db.stored_procedure.name`

### LINKED ISSUE #1491: [database] reporting stored procedures/prepared statements

trask:
Fixes #1491


model/database/registry.yaml
```diff
@@ -106,6 +106,28 @@ groups:
...
+        stability: release_candidate
+        brief: The name of a stored procedure within the database.
+        note: |
+          It is RECOMMENDED to capture the value as provided by the application
+          without attempting to do any case normalization.
+
+          For batch operations, if the individual operations are known to have the same
+          stored procedure name then that stored procedure name SHOULD be used.
+        examples: ["GetCustomer"]
+      - id: db.prepared_statement.name
```
lmolkova, line 120: it seems cassandra supports prepared statements: https://docs.datastax.com/en/developer/java-driver/3.0/manual/statements/prepared/index.html

model/database/registry.yaml
```diff
@@ -106,6 +106,28 @@ groups:
             'INSERT shipping_details SELECT orders',
             'get user by id',
           ]
+      - id: db.stored_procedure.name
```
lmolkova, line 109: I think redis functions/scripts would fit into this category?
 
https://redis.io/docs/latest/commands/fcall/
https://redis.io/docs/latest/commands/evalsha/

maryliag (approved):

.chloggen/2037.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `db.stored_procedure.name` and `db.prepared_statement.name`
```
lmolkova, line 13: should we also mention them in the migration guide ?

model/database/registry.yaml
trask: I don't think it should apply to these kinds of "unnamed" prepared statements.

Thinking about it more... I decided to remove db.prepared_statement.name altogether for now, because even in databases that support a "name" (https://github.com/open-telemetry/semantic-conventions/issues/1491#issuecomment-2470528067), it's still a name that's only local to a session, as opposed to something that is database-wide.

For normal (unnamed) prepared statements, I would expect the instrumentation to track these and capture the statement itself (as opposed to some reference to the statement). I guess ideally instrumentation would do the same for "named" prepared statements.

Also, I don't think named prepared statements are used much and so I'm comfortable leaving them out and dealing with making them opt-in somehow on metrics if we ever do add them.

docs/non-normative/db-migration.md
```diff
@@ -65,6 +65,8 @@ to
 | New: `db.operation.batch.size`                      |                                                                                                             |
 | New: `db.response.status_code`                      |                                                                                                             |
 | New: `db.operation.parameter.<key>`                 | Opt-In                                                                                                      |
+| New: `db.stored_procedure.name`                     |                                                                                                             |
+| New: `db.prepared_statement.name`                   |                                                                                                             |
```
trask, line 69:
```suggestion
```

lmolkova (approved):

Issue #1979, opened 2025-03-10
Document convention that Counter metrics shouldn't end with `.count`?


trask:
This isn't mentioned in https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#naming-rules-for-counters-and-updowncounters, but the convention has been followed pretty closely to only use `.count` suffix on UpDownCounter metrics (and not to use it on Counter metrics).

I'm not thinking of reasons right now, maybe related to the ambiguity of Counters being commonly exported as both delta and cumulative, and maybe related to the Prometheus exporter converting these to `_count_total`(?)

There are just a few metrics that violate this convention:

* `otel.sdk.span.ended.count`
* `otel.sdk.processor.span.processed.count`
* `otel.sdk.exporter.span.exported.count`
* `dotnet.thread_pool.work_item.count` - this one is stable, can't rename it

One option for the `otel.sdk.*` metric names could be:
* `otel.sdk.ended_spans`
* `otel.sdk.processor.processed_spans`
* `otel.sdk.exporter.exported_spans`

cc @JonasKunz

JonasKunz:
> `otel.sdk.ended_spans`
Sounds good to me, but I think we should also rename `otel.sdk.span.live.count` -> `otel.sdk.live_spans` for consistency?

> `otel.sdk.exporter.exported_spans`

Initially it was intended to have different namespaces per signal: E.g. `otel.sdk.exporter.span.*` `otel.sdk.exporter.log.*` etc. However I think it is okay to share metrics across signal types for exporters, see my at reasoning https://github.com/open-telemetry/semantic-conventions/pull/1921.

So this change is fine with me, we might even remove the `_spans` suffix and introduce an attribute instead to distinguish the telemetry type.

> `otel.sdk.processor.processed_spans``

Here I'm not that that happy about dropping the `span` in the namespace. There is no general concept of `processor` in the spec, there is just span-processors and log-processors, which are different things and evolve independently. This also shows with metrics: There is no `MetricProcessor`, just `MetricReader`s. Tehre's also some more reasoning given in https://github.com/open-telemetry/semantic-conventions/pull/1921.

So I'm not opposing to removing the `.count` suffix, I'm just concerned about dropping `span` from the metric namespace turning it from "span processor" to just "processor"

trask:
> I'm just concerned about dropping span from the metric namespace turning it from "span processor" to just "processor"

[...]

Issue #1999, opened 2025-03-17
More ideas about UpDownCounter naming


trask:
Based on @braydonk's https://github.com/open-telemetry/semantic-conventions/issues/1798#issuecomment-2730486556

> these are going to be UpDownCounters anyway which implies that any counted instances of handles/descriptors are nominally "open"

I thought it was worth a quick look at existing similar-ish updowncounters:

### those that include some kind of "open/active" term:

* `azure.cosmosdb.client.active_instance.count` -- seems like a possibility to rename to `azure.cosmosdb.client.instance.count` without confusion
* `cicd.pipeline.run.active` -- maybe align with `http.server.active_requests` and rename to `cicd.pipeline.active_runs`
* `http.server.active_requests` -- renaming this to `http.server.request.count` seems like it could lead to confusion since counting total requests is common (even though it's an UpDownCounter, I think the name alone could give some people confusion at first glance)
* `http.client.open_connections` -- I think this should probably be `http.client.connection.count`
* `http.client.active_requests` -- renaming this to `http.client.request.count` seems like it could lead to confusion
* `kestrel.active_connections` ( / `kestrel.queued_connections`) - interesting case where we want to differentiate between active and queued (pre- active)
* `kestrel.active_tls_handshakes` -- seems to make sense to follow `http.server.active_requests`
* `process.open_file_descriptor.count` -- I agree with removing the `open_` prefix, I don't think it's common to capture the total file descriptors created(?) and so I don't think would lead to confusion
* `signalr.server.active_connections` -- I think this should probably be `signalr.server.connection.count` (but it's already stable)

### those that don't:

* `cicd.worker.count`
* `db.client.connection.count`
* `dotnet.thread_pool.thread.count`
* `dotnet.timer.count`
* `dotnet.assembly.count`
* `go.goroutine.count`
* `jvm.buffer.count`
* `jvm.thread.count`
* `jvm.class.count` ( / `jvm.class.loaded` / `jvm.class.unloaded`)
* `process.thread.count`
* `system.network.connections`
* `system.process.count` ( / `system.process.created`)
* `vcs.change.count`
* `metric.vcs.repository.count`
* `metric.vcs.ref.count`

braydonk:
Thanks for thinking of this; I didn't think about how this concept could apply to other UpDownCounters but I think it can make sense in more cases than just the one I was thinking of.

[...]

PR #2042, opened 2025-03-31
Remove .count suffix from span health metrics

JonasKunz:
Renames the SDK span health metrics to remove the `.count` suffixes.
Partially adresses #1979. This is a follow-up of #1921, adjusting the exporter metric names based on the decisions in that PR.


AlexanderWert (approved):

model/otel/metrics.yaml
```diff
@@ -1,27 +1,27 @@
 groups:
-  - id: metric.otel.sdk.span.live.count
+  - id: metric.otel.sdk.live_spans
     type: metric
-    metric_name: otel.sdk.span.live.count
+    metric_name: otel.sdk.live_spans
```
lmolkova, line 4: same here, wouldn't `otel.sdk.span.live` be more consistent?

model/otel/metrics.yaml
```diff
@@ -1,27 +1,27 @@
...
+      For spans with `recording=true`: Implementations MUST record both `otel.sdk.live_spans` and `otel.sdk.ended_spans`.
+      For spans with `recording=false`: If implementations decide to record this metric, they MUST also record `otel.sdk.ended_spans`.
     attributes:
       - ref: otel.span.sampling_result
 
-  - id: metric.otel.sdk.span.ended.count
+  - id: metric.otel.sdk.ended_spans
     type: metric
-    metric_name: otel.sdk.span.ended.count
+    metric_name: otel.sdk.ended_spans
```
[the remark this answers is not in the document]
trask: I think we're a bit inconsistent, I found these examples of similar Counter metrics that were counting a "dimension" of something:

* jvm.class.loaded / jvm.class.unloaded (updowncounter: jvm.class.count)
* cpython.gc.collected_objects / cpython.gc.uncollectable_objects
* k8s.job.failed_pods (#2060)
* kestrel.rejected_connections
* system.process.created (updowncounter: system.process.count)

I do like the `jvm.class.*` and `system.process.*` examples best though, and they seem to best match this example which also has an associated updowncounter

model/otel/metrics.yaml
trask: related to #1999

I think I like your suggestion

* `otel.sdk.span.ended`
* `otel.sdk.span.live` (or `otel.sdk.span.active`)

especially if we also like this pattern for `http.server.active_requests` and rename that to `http.server.request.active`

JonasKunz: Fixed in https://github.com/open-telemetry/semantic-conventions/pull/2042/commits/6af9056f429677cda826f9162e0bf793675f95ff . I stayed with `live` because I personally find `active` confusing in the context of spans because to me that is ambigous with a span being part of the current context of a thread, which is independent of a span being ended or not.

docs/otel/sdk-metrics.md
```diff
@@ -483,6 +484,8 @@ E.g. for Java the fully qualified classname SHOULD be used in this case.
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Log metrics
```
JonasKunz, line 487: I just noticed that I forgot to add this heading in the previous PR, therefore sneaking it into this one.

trask (approved):

lmolkova (approved):

model/otel/metrics.yaml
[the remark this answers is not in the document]
lmolkova: created https://github.com/open-telemetry/semantic-conventions/issues/2073 so we don't forget

PR #2045, opened 2025-04-01
Initial proposal for MCP semantic conventions

### LINKED ISSUE #2043: Semantic conventions for Model Context Protocol
### Propose new conventions

MCP is a protocol for being able to add tools, resources and prompts to LLMs, and is gaining a lot of interest and traction in the LLM community for being able to connect LLMs with additional data sources.

MCP servers and clients can be a heterogeneous mix of languages and implementations, and as `stdio` is used as one of the communication mechanism, there is not a good way to emit logs, and so OTel/OTLP would be a good solution.

samsp-msft:
Fixes #2043

## Changes

Creates a doc version of semantic conventions for Model Context Protocol (MCP).

The [Model Context Protocol (MCP)](https://modelcontextprotocol.io/) is an open protocol for interactions between applications and AI models, supporting structured input and output formats. As MCP adoption grows, there's a need to establish semantic conventions for telemetry to ensure consistent observability across implementations.

### DIFF
--- a/docs/gen-ai/modelcontextprotocol.md
+++ b/docs/gen-ai/modelcontextprotocol.md
@@ -1,0 +1,97 @@
+# Semantic Conventions for Model Context Protocol (MCP)
+
+## Background
+
+The [Model Context Protocol (MCP)](https://modelcontextprotocol.io/) is an open protocol for interactions between applications and AI models, supporting structured input and output formats. As MCP adoption grows, there's a need to establish semantic conventions for telemetry to ensure consistent observability across implementations.
+
+## Proposal
+
+This issue proposes semantic conventions for [Model Context Protocol operations](https://github.com/modelcontextprotocol/specification) to standardize how traces, metrics, and logs are reported across MCP client and server implementations.
+
+## Spans
+
+### MCP Session
+
+An MCP session represents the lifecycle of a connection between a client and server.
+
+| Attribute | Type | Description | Examples | Required |
+|---|---|---|---|---|
+| `mcp.session.id` | string | Identifier for the MCP session | `c52f5ef3-c7a1-4f7a-a2e3-e38a7a4966d2` | Yes |
+| `mcp.transport` | string | Transport mechanism used | `streamable-http`, `stdio` | No |
+
+#### Span creation
+
+- Create a span when an MCP session is established and end it when the session is terminated
+- Span name: `mcp.session`
+
+### MCP Operation
+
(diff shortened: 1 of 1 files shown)


lmolkova (changes requested):
We define semantic conventions formally in YAML and generate MD from it, please refer to the [contributing.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) for the details.

It also seems that MCP is heavily based on JSON RPC and we have conventions defined for it [here](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/rpc).

I assume there is some MCP specifics that's worth recording, so we'll need to also figure out whether it should be expressed on JSON RPC spans/metrics with extra attributes or  as a separate set of conventions.

eranra:
Hi @samsp-msft, very interesting.

@samsp-msft can you generalize the semantic to something like "genai.tools.repo" with broader focus beyond just the focus on MCP. This will allow future extensions to other types of tool repositories and protocols for GenAI. I am aware that MCP is beyond just tools, but this seems to be the most needed area to observe. 

Also, you can add a specific semantic for "genai.tools.repo.type" that will enum value of "mcp"

samsp-msft:
> It also seems that MCP is heavily based on JSON RPC and we have conventions defined for it [here](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/rpc).

There are multiple transports, including stdio, JSON RPC and a new Http Streaming one. The MCP standard is evolving rapidly at this point. I would expect the spans for MCP to be wrapped by those for http / RPC based on the way that the communication is used. The attributes here are more about the MCP aspect of calling tools.

> @samsp-msft can you generalize the semantic to something like "genai.tools.repo" with broader focus beyond just the focus on MCP. This will allow future extensions to other types of tool repositories and protocols for GenAI. I am aware that MCP is beyond just tools, but this seems to be the most needed area to observe.

This makes sense. MCP is codifying its own transport semantics, but stepping back the same concepts exist when exposing other mechanisms for integrating data sources, tools and prompts into gen_ai. Most of this is not specific to MCP - probably just the concept of a session with an id, and some of the semantics of the type of data being requested, but the latter could be a namespaced value.

For example, if the generalized namespace is `gen_ai.integration`  then we could use a namespaced attribute value that indicates the type eg `gen_ai.integration.request = mcp.tools.call`.

gen_ai already has a section on tool calling - I think MCP is extending that concept further. What should the base namespace be for the attributes and I can rework this based on that?

samsp-msft:
Duplicate #2083

PR #2046, opened 2025-04-02
Update GenAI conventions based on system-specific naming guidance

lmolkova:
Implements #1708 for GenAI conventions

  - Rename `gen_ai.system` to `gen_ai.provider.name`
  - Remove `gen_ai` prefix from `gen_ai.openai.*` attributes.
  - Rename `az.ai.*` system names to `azure.ai.*`.

Related: #608


model/gen-ai/registry.yaml
```diff
@@ -5,7 +5,7 @@ groups:
     brief: >
       This document defines the attributes used to describe telemetry in the context of Generative Artificial Intelligence (GenAI) Models requests and responses.
     attributes:
-      - id: gen_ai.system
+      - id: gen_ai.system.name
```
aabmass, line 8: This is creates some confusion with `gen_ai.system.message` https://github.com/open-telemetry/semantic-conventions/blob/beb66215d37557d8bfc9303abd4ce3e2ef47da01/model/gen-ai/events.yaml#L10-L17

aabmass, line 8: From SIG: some confusion when the provider (Grok, OpenAI, etc.) is different from the protocol.
- We could add `gen_ai.protocol.name` separate attribute
- We could rename this to `gen_ai.provider.name`

Prior art:
- "system" has prior art in DB, messaging
- feature flags has `provider` and RPC has `protocol`

Each convention picks something better than system if there is something better in that space. Provider is more convention in GenAI "inference providers".

lmolkova: renamed to `gen_ai.provider.name`

docs/gen-ai/aws-bedrock.md
```diff
@@ -139,14 +139,14 @@ Additional output format details may be recorded in the future in the `gen_ai.ou
 
 ---
 
-`gen_ai.system` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+`gen_ai.provider.name` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
sergey-splunk, line 142: Let's keep the mentioning of the `gen_ai.system` as the previous attribute naming, so people can still discover this spec by searching `gen_ai.system` and have a connection to the new naming

Cirilla-zmh: +1. We also have `messaging.system` in the messaging semantic conventions, and obviously, it helps users understand where the data originates. The attribute name `gen_ai.provider.name` appears to refer specifically to the backend service provider. These two attributes can coexist with distinct semantics.

One important point to note is that we intend to use this attribute not only on the client side but also on the server side (in model services built with frameworks like vLLM or SGLang). In such scenarios, using the name "system" seems more reasonable.

model/gen-ai/registry.yaml
```diff
@@ -5,7 +5,7 @@ groups:
     brief: >
       This document defines the attributes used to describe telemetry in the context of Generative Artificial Intelligence (GenAI) Models requests and responses.
     attributes:
-      - id: gen_ai.system
+      - id: gen_ai.provider.name
```
lmolkova, line 8: Based on the SIG discussion, I've been considering other possible attribute names that:
- describe something that's always available
- reflect that client does not always know which provider it talks to
- can be consistent across client and server in the perfect world

Here're the candidates:

1. `gen_ai.protocol.name` 
2. `gen_ai.api.namespace | style | flavor | family | provider.name` 
3. `gen_ai.telemetry.format`

I think provider is still better for the following reasons:
- protocol/api are too low-level, they are not available or are impl detail on the higher-level framework level (langchain, etc)
- protocol/api imply that model is remote and while it's the main use-case, it won't works nicely for local models. Provider name vagueness is useful here and works with `llama.cpp` or `hugging_face.transformers`. 
- telemetry format feels really artificial, `gen_ai.telemetry.format = openai | gcp.vertex_ai` feels weird
- in many cases (client-side openai being the most problematic) provider name and protocol | api family are the same. Perhaps we should treat `openai` as `unknown` 

---------
Based on this, I believe `gen_ai.provider.name` is still the best choice among these candidates.

model/gen-ai/registry.yaml
```diff
@@ -90,20 +78,30 @@ groups:
...
-          instrumentation's best knowledge, instead of the actual system. The `server.address`
-          attribute may help identify the actual system in use for `openai`.
+          Multiple systems, including Azure OpenAI, Gemini, and AI hosting platforms
+          are accessible using the OpenAI REST API and corresponding client libraries,
+          but may proxy or host models from different providers.
 
-          For custom model, a custom friendly name SHOULD be used.
-          If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
-        examples: 'openai'
+          The `gen_ai.request.model`, `gen_ai.response.model, and `server.address`
```
Cirilla-zmh: A backtick \` is missing after `gen_ai.response.model`.

model/gen-ai/registry.yaml
```diff
@@ -90,20 +78,30 @@ groups:
...
+          attributes may help identify the actual system in use.
+
+          For custom models, a friendly custom provider name SHOULD be used.
+          If none of these options apply, the `gen_ai.provider.name` SHOULD be
+          set to `_OTHER`.
```
Cirilla-zmh: Should we add _OTHER into the members?

model/gen-ai/registry.yaml
Cirilla-zmh: I'll go with `gen_ai.provider.name`.

docs/gen-ai/openai.md
```diff
@@ -34,7 +56,7 @@ The Semantic Conventions for [OpenAI](https://openai.com/) extend and override t
 
 Semantic Conventions for [OpenAI](https://openai.com/) client spans extend and override the semantic conventions for [Gen AI Spans](gen-ai-spans.md).
 
-`gen_ai.system` MUST be set to `"openai"` and SHOULD be provided **at span creation time**.
+`gen_ai.provider.name` MUST be set to `"openai"` and SHOULD be provided **at span creation time**.
```
aabmass, line 60: Any idea why `gen_ai.provider.name` does not appear in the table of attributes?

docs/gen-ai/gen-ai-spans.md
```diff
@@ -69,17 +89,27 @@ client or when the GenAI call happens over instrumented protocol such as HTTP.
...
+
+The `gen_ai.request.model`, `gen_ai.response.model, and `server.address`
+attributes may help identify the actual system in use.
 
-The actual GenAI product may differ from the one identified by the client.
-Multiple systems, including Azure OpenAI and Gemini, are accessible by OpenAI client
-libraries. In such cases, the `gen_ai.system` is set to `openai` based on the
-instrumentation's best knowledge, instead of the actual system. The `server.address`
-attribute may help identify the actual system in use for `openai`.
+For custom models, a friendly custom provider name SHOULD be used.
```
aabmass: Most custom models are served as OpenAI API (for example vLLM). But the OpenAI guidance below says 

> `gen_ai.provider.name` MUST be set to `"openai"` and SHOULD be provided **at span creation time**.

Is this line about custom models only intended to apply on the inference server side?

model/gen-ai/registry.yaml
```diff
@@ -90,20 +78,30 @@ groups:
...
+
+          The `gen_ai.provider.name` attribute acts as a discriminator that
+          identifies the GenAI telemetry format flavor. It SHOULD be set
```
aabmass: What do you mean by telemetry format flavor here? The JSON schemas we are working on in https://github.com/open-telemetry/semantic-conventions/pull/2179 defines a standard  telemetry format, so it will always be the same flavor. Is this intended as an intermediate state before we have that format pinned down?

model/gen-ai/registry.yaml
aabmass: Do you mean `gen_ai.api.namespace | style | flavor | family | provider.name` as alternative names for `gen_ai.provider.name`, but with the same semantics? If that is the intention of `gen_ai.provider.name`, those are more clear IMO.

> * protocol/api are too low-level, they are not available or are impl detail on the higher-level framework level (langchain, etc)

I feel like framework level instrumentation should maybe not be emitting inference level spans. Should the framework level even have access to the provider name anyway, or just use it's own intermediate representation?

Cirilla-zmh: Based on the current semantics of `gen_ai.provider.name`, I believe we should not relate its value to a specific API. This is because any provider can offer an API in any style, just like openai-style API. Once the provider name cannot be clearly determined in inference clients' instrumentations, it seems that we should assign it as "unknown".

If that's the case, then it appears that the inference client will never be able to obtain an accurate provider.name, unless we determine it based on the base URL being accessed — and I don't think that's a maintainable approach, but it seems like the only option we have.

docs/gen-ai/gen-ai-spans.md
xrmx: Eh it's hard to draw the line when you are talking about `openai` the provider or the client library. Related to that some span attributes have openai in the name because they come from the openai client but the provider may be different.

aabmass: I'm not arguing against one or the other, I just want to fix this contradiction.

gyliu513:
@lmolkova do we have any blockers for this to be merged? This is the foundation for some future AI Semantic Convention, it will be great this can be merged soon, thanks!

model/gen-ai/registry.yaml
lmolkova: I mean the broader flavor: today we have OpenAI, Azure AI Inference, AWS bedrock conventions, and GCP is on the way - https://github.com/open-telemetry/semantic-conventions/pull/2125

So it's additional flavor applied by specific provider. There is an example in the next paragraph and I'm adjusting it slightly to

>       The `gen_ai.provider.name` attribute acts as a discriminator that
>       identifies the GenAI telemetry format flavor specific to that provider.
>       It SHOULD be set consistently with provider-specific attributes and signals.
>       For example, GenAI spans, metrics, and events related to AWS Bedrock
>       should have the `gen_ai.provider.name` set to `aws.bedrock` and include
>       applicable `aws.bedrock.*` attributes and are not expected to include
>      `openai.*` attributes.

Do you think it addresses your question?

docs/gen-ai/gen-ai-spans.md
lmolkova: Make sense, I'm  going to remove 

> For custom models, a friendly custom provider name SHOULD be used.

I think it's redundant and covered below with 

> `gen_ai.provider.name` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.

lmolkova: updated in https://github.com/open-telemetry/semantic-conventions/pull/2046/commits/86abbadfef871a3c1270de6ec662e7970d4dc62f

model/gen-ai/registry.yaml
lmolkova: I removed this sentence in https://github.com/open-telemetry/semantic-conventions/pull/2046/commits/86abbadfef871a3c1270de6ec662e7970d4dc62f, I'm not aware of any cases where we don't know the provider name. We can always bring it back if i turns out to be necessary.

docs/gen-ai/openai.md
lmolkova: because we don't reference it in the yaml  -  the reason for it is the tolling limitations https://github.com/open-telemetry/weaver/issues/479.

When we reference this enum attribute, we reference all the values and have no means to tell that it's only valid to use `openai`. So the full enum would appear. 

That's the imperfect practice we've been following in other conventions until we figure the tooling out

model/gen-ai/registry.yaml
lmolkova: thanks for your points!

We need to be able to accommodate multiple layers of GenAI instrumentation:
- framework/agent
- model client lib 
- model server
- in-process model

Some of them will not be emitting inference spans, but would still emit other GenAI ones (e.g. invoke_agent or something else we'll come up with). Plus the line between basic inference vs reasoning vs inference + query rewrite + RAG is very blurry, you can easily hide the RAG complexity on the server side  - I assume this API does it  https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/rag-api#prediction-api, but you can also do the same pattern with langchain or another client-side framework.

The `gen_ai.provider.name` could work for all of these layers (being imperfect whenever openai api is used). 

The `api` or `protocol` in the name would be weird for these local vs service-side RAG scenarios and won't really work for in-proc models in the future.

docs/gen-ai/README.md
```diff
@@ -7,9 +7,24 @@ linkTitle: Generative AI
...
+> (or prior):
+>
+> * SHOULD NOT change the version of the GenAI conventions that they emit by default.
+>   Conventions include, but are not limited to, attributes, metric, span and event names,
+>   span kind and unit of measure.
+> * SHOULD introduce an environment variable `OTEL_SEMCONV_EXPERIMENTAL_OPT_IN`
+>   as a comma-separated list of category-specific values. The list of values
+>   includes:
+>   * `gen-ai-latest` - emit the latest (supported by the instrumentation) GenAI
+>     conventions and dot not emit the old one (v1.34.0 or prior).
```
trask:
```suggestion
>     conventions and do not emit the old one (v1.34.0 or prior).
```

docs/gen-ai/README.md
```diff
@@ -7,9 +7,24 @@ linkTitle: Generative AI
...
+> [v1.34.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.34.0/docs/gen-ai/README.md)
+> (or prior):
+>
+> * SHOULD NOT change the version of the GenAI conventions that they emit by default.
+>   Conventions include, but are not limited to, attributes, metric, span and event names,
+>   span kind and unit of measure.
+> * SHOULD introduce an environment variable `OTEL_SEMCONV_EXPERIMENTAL_OPT_IN`
+>   as a comma-separated list of category-specific values. The list of values
+>   includes:
+>   * `gen-ai-latest` - emit the latest (supported by the instrumentation) GenAI
```
trask: wdyt of?
```suggestion
>   * `gen-ai-latest-experimental` - emit the latest (supported by the instrumentation) GenAI
```

docs/gen-ai/README.md
```diff
@@ -7,9 +7,24 @@ linkTitle: Generative AI
...
+>     conventions and dot not emit the old one (v1.34.0 or prior).
+>   * The default behavior (in the absence of one of `gen-ai-latest`) is to continue
+>     emitting whatever version of the GenAI conventions the instrumentation
+>     was emitting (1.34.0 or prior).
+> * SHOULD update the baseline version to a stable one in the next major version
+>   after GenAI conventions are declared stable.
```
trask: I found this a bit confusing without listing the `genai` (stable) value above

docs/gen-ai/README.md
lmolkova: I don't mind, but note the env var name `OTEL_SEMCONV_EXPERIMENTAL_OPT_IN`, do you think `-experimental` is necessary in the value as well?

trask: oh I missed that, thx!

trask:
```suggestion
>   * `gen-ai-latest` - emit the latest (supported by the instrumentation) experimental GenAI
```

docs/gen-ai/README.md
trask: resolving, this was also due to me missing the word EXPERIMENTAL in the env var

trask: will `OTEL_SEMCONV_OPT_IN` be brought into the picture before stabilizing?

docs/gen-ai/README.md
lmolkova: WDYT about https://github.com/open-telemetry/semantic-conventions/pull/2046/commits/1e977ce277e860af596fa8f3f503c414282d2a53 ?

docs/gen-ai/README.md
lmolkova: I'd prefer not to add anything about stable versions until we have stability in sight. So I removed this point and added 

> This transition plan will be updated to include stable version before the
> GenAI conventions are marked as stable.

trask: resolved by https://github.com/open-telemetry/semantic-conventions/pull/2046#discussion_r2162629205

docs/gen-ai/aws-bedrock.md
```diff
@@ -6,12 +6,33 @@ linkTitle: AWS Bedrock
...
+>   as a comma-separated list of category-specific values. The list of values
+>   includes:
+>   * `gen-ai-latest-experimental` - emit the latest experimental version of
+>     GenAI conventions (supported by the instrumentation) and do not emit the
+>     old one (v1.34.0 or prior).
+>   * The default behavior is to continue emitting whatever version of the GenAI
+>     conventions the instrumentation was emitting (1.34.0 or prior).
+>
+> This transition plan will be updated to include stable version before the
+> GenAI conventions are marked as stable.
```
trask: 👍

model/gen-ai/registry.yaml
```diff
@@ -90,20 +78,27 @@ groups:
               value: "mistral_ai"
```
trask, line 77: any preference on how these namespaces are derived from their domain name?
* `mistral_ai` -- https://mistral.ai
* `xai` -- https://x.ai

I think `mistral_ai` was correct when we were deriving it from the product name, but now that we're basing this on the domain name(?) we could probably go either way...

lmolkova: Let's keep mistral_ai - it's a trademark?

I renamed `xai` to `x_ai` for consistency and to follow the `_` separator policy.

model/gen-ai/registry.yaml
aabmass: That clarifies thanks. But are we planning to add custom fields within the JSON schemas in https://github.com/open-telemetry/semantic-conventions/pull/2179/ or is this discriminator just for the flat attributes on telemetry like we already documented.

model/gen-ai/registry.yaml
```diff
@@ -5,105 +5,99 @@ groups:
...
-              brief: 'Groq'
+              brief: '[Groq](https://groq.com/)'
             - id: mistral_ai
               stability: development
               value: "mistral_ai"
-              brief: 'Mistral AI'
+              brief: '[Mistral AI](https://mistral.ai/)'
 
-        brief: The Generative AI product as identified by the client or server instrumentation.
+        brief: The Generative AI model provider as identified by the client
```
aabmass: Based on your above comment 
> The gen_ai.provider.name could work for all of these layers (being imperfect whenever openai api is used).

It sounds like provider is not specific to model inference, but across GenAI telemetry like RAG, agents, etc. So, should we remove "model" throughout here?

model/gen-ai/registry.yaml
aabmass: If `gen_ai.provider.name` is just a 1:1 rename of `gen_ai.system` then I have no objection. I think we can address "api flavor" or whatever we call it in a future PR.

model/gen-ai/spans.yaml
```diff
@@ -168,7 +168,7 @@ groups:
       Semantic Conventions for [Azure AI Inference](https://learn.microsoft.com/azure/ai-studio/reference/reference-model-inference-api)
       client spans extend and override the semantic conventions for [Gen AI Spans](gen-ai-spans.md).
     note: |
-      `gen_ai.system` MUST be set to `"az.ai.inference"` and SHOULD be provided **at span creation time**.
+      `gen_ai.provide.name` MUST be set to `"azure.ai.inference"` and SHOULD be provided **at span creation time**.
```
aabmass:
```suggestion
      `gen_ai.provider.name` MUST be set to `"azure.ai.inference"` and SHOULD be provided **at span creation time**.
```

aabmass (approved):

model/gen-ai/spans.yaml
lmolkova: great catch!

model/gen-ai/registry.yaml
lmolkova: The JSON schemas could define system-specific properties - e.g. OpenAI can extend that schema and add `openai.foo` property on each output message, and Gemini could do the same and add `gemini.bar` property. This would be compatible with base definition.

Then `gen_ai.provider.name` openai would mean that backend could (but does not have to) expect OpenAI specific override of `gen_ai.output.messages`. 

I.e. `gen_ai.provider.name` identifies the variations in the format on the top-level attributes and inside complex attributes.

Let me know if you think we should reflect it in the semconv text.

docs/gen-ai/README.md
```diff
@@ -7,9 +7,25 @@ linkTitle: Generative AI
...
+> [v1.36.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/gen-ai/README.md)
+> (or prior):
+>
+> * SHOULD NOT change the version of the GenAI conventions that they emit by default.
+>   Conventions include, but are not limited to, attributes, metric, span and event names,
+>   span kind and unit of measure.
+> * SHOULD introduce an environment variable `OTEL_SEMCONV_STABILITY_OPT_IN`
+>   as a comma-separated list of category-specific values. The list of values
+>   includes:
+>   * `gen-ai-latest-experimental` - emit the latest experimental version of
```
trask: is there any preference between `gen-ai-latest-experimental` and `gen_ai_latest_experimental`?

trask (approved):

Issue #1426, opened 2024-09-23
Automate schema-next generation


lmolkova:
[schema-next](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) contains a list of transformations to apply to attribute/metrics/event/resource to convert them from previous version to current one.

It relies on contributors and reviewers to update it manually and there is no check that verifies that file was updated and that it's a valid transformation (both attributes/metrics/etc exist, the type is the same, etc).

We can automate generation of this file if we formalize it in yaml.

1. Attribute/metric/event rename is applied along with deprecation (#1419). We can formalize deprecation and update schema-next automatically.

    ```yaml
    - id: registry.foo.deprecated 
      ...
      attributes:
        - id:  foo.id
          deprecated:
            renamed_to: foo.unique_id
            note: >
              Optional note to provide any other info
        - id:  foo.another_attribute
          deprecated: removed
    ```

   This would cover all of the existing cases we support in `schema-next.yaml`.

2. There are changes that we don't yet have transformation defined for that don't involve deprecation - for those we probably need to analyze the diff between versions:
    - attribute type/metric unit or instrument was changed
    - default value has changed
    - ...

    For this category we can also look into combining it with the changelog - changelog could be auto-generated based on the actual diff. Or changelog format can be made more specific so that we can generate transformations from it.

Arguably, we need a combination of changelog and semantic conventions to be available to the codegen as well (e.g. if attribute type has changed, I might want to generate some extra documentation, do something custom like type conversion, or maybe get alerted in some custom way.

lquerel:
Thanks, Liudmila, for specifying this. With that in place in semconv, Weaver will be able to automatically update the OTEL schema for any future version of the registry.

I would suggest using a representation similar to renaming for the removal of attributes, as it's possible to include a note for both removal and renaming.

```yaml
- id: registry.foo.deprecated 
  ...
  attributes:
    - id:  foo.id
      deprecated:
        renamed_to: foo.unique_id
        note: >
          Optional note to provide any other info
    - id:  foo.another_attribute
      deprecated: 
        removed: true
        note: >
          Removed, report shared memory usage with `metric.system.memory.shared` metric
```

Another option, which could be grammatically more regular, might be…

```yaml
- id: registry.foo.deprecated 
  ...
  attributes:
    - id: foo.id
      deprecated:
        action: renamed
        renamed_to: foo.unique_id
        note: >
          Optional note to provide any other info
[...]

Issue #1419, opened 2024-09-19
Don't require brief for deprecated attributes, auto-generate it in markdown and code


lmolkova:
We currently do this: 

```yaml
brief: 'Deprecated, use `server.address`, `server.port` attributes instead.'
deprecated: "Replaced by `server.address` and `server.port`."
```

It'd be nice to avoid duplication - https://github.com/open-telemetry/semantic-conventions/pull/1418#discussion_r1766981411

lmolkova:
Syntax suggestion:

```yaml
- id: registry.foo.deprecated 
  ...
  attributes:
    - id:  foo.id
      deprecated:
        renamed_to: foo.unique_id
        note: >
          Optional note to provide any other info
    - id:  foo.another_attribute
      deprecated: removed
```

PR #2047, opened 2025-04-02
Use deprecated struct instead of text, add policies 

lmolkova:
> [!WARNING]
>
> This is a breaking change for code-generation using (old) [build-tools](https://github.com/open-telemetry/build-tools/issues/322).
> If merged, the (new) [weaver](https://github.com/open-telemetry/weaver) will become the only supported tool to generate code and markdown starting with the next release.

Leverage new `deprecated` structure introduced in https://github.com/open-telemetry/weaver/pull/400

- convert all `deprecated: "text"` into new struct and require it via policy
- check that `renamed_to` value exists (https://github.com/open-telemetry/weaver/issues/673) and is not deprecated 

This change is the  next step on the way to more semconv automations including validate-able and auto-generated changelogs and schema transformations: #1426, #1419, #1596


braydonk (approved):

model/database/deprecated/metrics-deprecated.yaml
```diff
@@ -89,7 +102,8 @@ groups:
     type: metric
     metric_name: db.client.connections.wait_time
     stability: development
-    deprecated: "Replaced by `db.client.connection.wait_time`. Note: the unit also changed from `ms` to `s`."
+    deprecated:
+      reason: uncategorized
```
maryliag, line 106: do you want to add a note instead? Similar to what you did on the registry-deprecated?

model/database/deprecated/metrics-deprecated.yaml
```diff
@@ -101,7 +115,8 @@ groups:
     type: metric
     metric_name: db.client.connections.use_time
     stability: development
-    deprecated: "Replaced by `db.client.connection.use_time`. Note: the unit also changed from `ms` to `s`."
+    deprecated:
+      reason: uncategorized
```
maryliag, line 119: same comment as above

model/database/deprecated/metrics-deprecated.yaml
lmolkova: great catch, thanks!

.chloggen/2047.yaml
```diff
@@ -0,0 +1,12 @@
...
+note: Convert deprecated text to structured format.
+issues: [ 2047 ]
+
+subtext: >
+  This is a breaking change from the schema perspective, but does not change anything for instrumentations
+  or the end users.
+  It breaks compatibility with the (old) [code generation toolings](https://github.com/open-telemetry/build-tools/issues/322).
+  Please use [weaver](https://github.com/open-telemetry/weaver) to generate Semantic Conventions markdown or code.
+
+
```
trask, line 12:
```suggestion
```

model/code/registry-deprecated.yaml
```diff
@@ -8,35 +8,46 @@ groups:
       - id: code.function
         type: string
         stability: development
-        deprecated: Replaced by `code.function.name`
+        deprecated:
+          reason: renamed
+          renamed_to: code.function.name
```
trask, line 13: maybe better not to model as a rename since it also included the merging of code.namespace?

model/database/deprecated/metrics-deprecated.yaml
```diff
@@ -77,7 +89,9 @@ groups:
     type: metric
     metric_name: db.client.connections.create_time
     stability: development
-    deprecated: "Replaced by `db.client.connection.create_time`. Note: the unit also changed from `ms` to `s`."
+    deprecated:
+      reason: uncategorized
+      note: Replaced by `db.client.connection.create_time` with a `s` unit.
```
trask, line 94: (and same below if you take this suggestion)
```suggestion
      note: Replaced by `db.client.connection.create_time` with unit `s`.
```

model/http/deprecated/registry-deprecated.yaml
```diff
@@ -8,72 +8,97 @@ groups:
...
+          renamed_to: url.full
         examples: ['https://www.foo.bar/search?q=OpenTelemetry#SemConv']
       - id: http.target
         type: string
         brief: 'Deprecated, use `url.path` and `url.query` instead.'
         stability: development
-        deprecated: "Split to `url.path` and `url.query."
+        deprecated:
+          reason: obsoleted
+          note: Split to `url.path` and `url.query.
```
trask, line 45:
```suggestion
          note: Split to `url.path` and `url.query`.
```

model/http/deprecated/registry-deprecated.yaml
```diff
@@ -8,72 +8,97 @@ groups:
...
         examples: ['/search?q=OpenTelemetry#SemConv']
       - id: http.request_content_length
         type: int
         brief: 'Deprecated, use `http.request.header.<key>` instead.'
         stability: development
-        deprecated: "Replaced by `http.request.header.<key>`."
+        deprecated:
+          reason: uncategorized
+          note: Replaced by `http.request.header.<key>`.
```
trask, line 53: wdyt? (and if so would want to change the brief above as well)
```suggestion
          note: Replaced by `http.request.header.content_length`.
```

model/http/deprecated/registry-deprecated.yaml
```diff
@@ -8,72 +8,97 @@ groups:
...
         examples: 3495
       - id: http.response_content_length
         type: int
         brief: 'Deprecated, use `http.response.header.<key>` instead.'
         stability: development
-        deprecated: "Replaced by `http.response.header.<key>`."
+        deprecated:
+          reason: uncategorized
+          note: Replaced by `http.response.header.<key>`.
```
trask, line 61: same question as above

trask (approved):
👍

model/http/deprecated/registry-deprecated.yaml
lmolkova: I like it!

trask:
can update with new deprecated element after #1982 is merged

ChrsMark (approved):
Changes lgtm

model/code/registry-deprecated.yaml
```diff
@@ -8,35 +8,48 @@ groups:
       - id: code.function
         type: string
         stability: development
-        deprecated: Replaced by `code.function.name`
+        deprecated:
+          reason: uncategorized
+          note: >
```
joaopgrassi, line 13: Why don't we use the `renamed_to` here? 🤔

model/database/deprecated/metrics-deprecated.yaml
joaopgrassi: Can't we still cover it with the `renamed_to` and add an additional note? Isn't this way with only the note having the same limitations as before, with it being hard to do any code generation? Sorry if I missed something obvious

joaopgrassi (commented):
Looks good! I just have a general question as to why some attributes/metrics are not using the `renamed_to` field and just have a note. Maybe I'm missing something obvious

xrmx (approved):

model/database/deprecated/metrics-deprecated.yaml
trask: I believe `renamed_to` is an also indication to automated schema conversion tools (not only for markdown generation)

model/code/registry-deprecated.yaml
lmolkova: because `code.function` is not qualified e.g. `foo` and  `code.function.name` is fully qualified, e.g. `my_app.namespace.class.foo`, so in order to construct `code.function.name` you'd need to take deprecated  `code.namespace` and `code.function`.

`renamed_to` is essentially limited to renames that don't change anything important about the attribute (type, meaning, format).

we could in the future define more complex transformations, it's just not supported now.

Hope it helps!

model/database/deprecated/metrics-deprecated.yaml
lmolkova: yes, the main goal is automated schema transformation, note is still there for humans

lmolkova:
Related to https://github.com/open-telemetry/weaver/pull/737 which enforces structured deprecated property under --future

PR #2050, opened 2025-04-02
Propose an update to the values of 'gen_ai.system' for Google.

### LINKED ISSUE #1950: Rename `gen_ai.system` values of `vertex_ai` and `gemini`
### Area(s)

area:gen-ai

### What's missing?

Common nomenclature (or prefixing) for Google's GenAI systems.

This divergent naming can also make it difficult to know which value to populate in `gen_ai.system` at higher-level abstraction layers that build on top of the Google GenAI SDK (which provides the same library interface, regardless of whether Vertex AI or the Gemini backend is used).

## Common proposal

When Google is a backend, use the Google-specific `gcp.client.service` convention to identify the specific backend where known:

- `gcp.client.service`:
   - value == `generativelanguage` => `generativelanguage.googleapis.com` (Gemini)
   - value == `aiplatform` => `aiplatform.googleapis.com` (Vertex AI)

## Ideal solution

In conjunction with the common proposal above, reduce the two values of `gemini` and `vertex_ai` for `gen_ai.system` to a single value: `google`

## Alternative solution

In conjunction with the common proposal above:

Rename the existing `gen_ai.system` values for Google:

 - `gemini` => `google.gemini`
 - `vertex_ai` => `google.gcp.vertex`

And define an additional value:

 - `google` (for when the specific subsystem -- `vertex` vs `gemini` -- is not known)

This way there is at least a common prefix (`google`) for which shared properties can be attached.

michaelsafyan:
Fixes #1950 

## Changes

Refines the values of `gen_ai.system` for Google's AI products.

The Google Gen AI SDK defined here:

 - https://github.com/googleapis/python-genai

... provides a common client for both the Vertex AI and Gemini APIs and reflects some degree of intended sharing / convergence between these two systems. Having totally separate names (like `vertex_ai` and `gemini`) without a common prefix means that any vendor-specific attributes would need to be defined duplicatively for both `vertex_ai` and for `gemini`, even if supported by both systems.

Additionally, code built on top of the Google Gen AI SDK may not necessarily know which backend is being used.

This change proposes a common prefix (`gcp.gen_ai`) which may be used for any Google-specific Gen AI attributes that are common between both backend systems. More specific sub-systems (`gcp.gen_ai.vertex_ai` and `gcp.gen_ai.gemini`) still exist to allow for divergent properties where desireable. However, the common prefix should make it easier to provide unified Semantic Conventions where it makes sense to do so, simplifying conventions and instrumentation.

* [?] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.
   * _Is it possible to rename attribute values this way?_ I see that attribute keys are renamable, but it's not clear whether it is possible to change the values this way.


docs/attributes-registry/gen-ai.md
```diff
@@ -102,15 +102,22 @@ Datastore: A tool used by the agent to access and query structured or unstructur
 | `az.ai.openai` | Azure OpenAI | ![Development](https://img.shields.io/badge/-development-blue) |
 | `cohere` | Cohere | ![Development](https://img.shields.io/badge/-development-blue) |
 | `deepseek` | DeepSeek | ![Development](https://img.shields.io/badge/-development-blue) |
-| `gemini` | Gemini | ![Development](https://img.shields.io/badge/-development-blue) |
+| `gcp.gen_ai` | Any Google AI endpoint [6] | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 105: we have a difficult time with namespace and attribute collisions in semconv.
E.g. we don't allow `foo.bar` attribute if `foo.bar.baz` is defined. 

Assuming we can rename other two based on my suggestions, we'd resolve the collision as well

```
gcp.gemini
gcp.vertex_ai
gcp.gen_ai 
```

.chloggen/gcp_genai_system.yaml
```diff
@@ -0,0 +1,6 @@
+change_type: breaking
+component: gen-ai
+note: Refine the values for `gen_ai.system` related to Google's AI endpoints.
+issues: [1950]
+subtext: Enable sharing of attributes between Vertex AI and Gemini through a common prefix.
+
```
lmolkova, line 6:
```suggestion

```

docs/attributes-registry/gen-ai.md
michaelsafyan: If the `gen_ai.system` were `gcp.vertex_ai`, would it make sense for it to include attributes prefixed with `gcp.gena_ai`? Part of the thinking of having this prefix structure is that it conveys the intent that `vertex_ai` and `gemini` refine this common prefix, where the more specific systems can still reference/use attributes defined in the shorter prefix.

docs/attributes-registry/gen-ai.md
```diff
@@ -102,15 +102,22 @@ Datastore: A tool used by the agent to access and query structured or unstructur
...
+| `gcp.gen_ai.gemini` | Gemini [7] | ![Development](https://img.shields.io/badge/-development-blue) |
+| `gcp.gen_ai.vertex_ai` | Vertex AI [8] | ![Development](https://img.shields.io/badge/-development-blue) |
```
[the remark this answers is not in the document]
michaelsafyan: The intention here wasn't "SemConv domain", but rather to have a common prefix so that it is clear that the more specific variant refines/inherits from the shorter prefix.

Additionally, there is the matter that Google teams typically can lay claim to attributes in `gcp.<product>.`. For example, a team in Vertex AI might reasonably think that attributes prefixed with `gcp.vertex_ai.` are fair game (even if not defined in the OTel SemConv repo). This prefixing creates a namespace that deviates from `gcp.<product>.` that may help to avoid collisions (i.e. a namespace that is clearly to be defined/specified in the OTel SemConv repo, not outside of it).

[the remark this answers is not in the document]
lmolkova: > For example, a team in Vertex AI might reasonably think that attributes prefixed with gcp.vertex_ai. are fair game (even if not defined in the OTel SemConv repo). This prefixing creates a namespace that deviates from gcp.<product>. that may help to avoid collisions (i.e. a namespace that is clearly to be defined/specified in the OTel SemConv repo, not outside of it).

That's the eventual point. GCP Vertex AI is no different than GCP PubSub or GKE - they all are in semconv and someday I hope will be governed by Google Cloud. 

So I don't think we should deviate from this principle for Vertex or Gemini, why?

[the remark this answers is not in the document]
michaelsafyan: OK. I'm persuaded.

I updated these to:

 - `gcp` (no need to have `.gen_ai` if it is possible to have attributes that do not match `gen_ai.system`).
 - `gcp.vertex_ai`
 - `gcp.gemini`

My intention is to still use the `gcp.gen_ai.` prefix, though, for attributes that are common to `gcp.vertex_ai` and `gcp.gemini`.

docs/attributes-registry/gen-ai.md
michaelsafyan: Yes, I noted the `azure.ai.` prefix as part of the inspiration for the original naming. In any event, if common attributes for GCP can live under `gcp.gen_ai.`, then I am OK with the suggested alternative names. Updated.

docs/attributes-registry/gen-ai.md
```diff
@@ -102,15 +102,22 @@ Datastore: A tool used by the agent to access and query structured or unstructur
 | `az.ai.openai` | Azure OpenAI | ![Development](https://img.shields.io/badge/-development-blue) |
 | `cohere` | Cohere | ![Development](https://img.shields.io/badge/-development-blue) |
 | `deepseek` | DeepSeek | ![Development](https://img.shields.io/badge/-development-blue) |
-| `gemini` | Gemini | ![Development](https://img.shields.io/badge/-development-blue) |
+| `gcp` | Any Google generative AI endpoint [6] | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 105: are we going to ever report abstract Google AI thing? Don't we know which model family are we talking to? 

If it's the case, let's still pick a name for it, something generic like `gcp.ai` or `gcp.gen_ai` like the one you had before

docs/attributes-registry/gen-ai.md
```diff
@@ -102,15 +102,22 @@ Datastore: A tool used by the agent to access and query structured or unstructur
...
+| `gcp.gemini` | Gemini [7] | ![Development](https://img.shields.io/badge/-development-blue) |
+| `gcp.vertex_ai` | Vertex AI [8] | ![Development](https://img.shields.io/badge/-development-blue) |
 | `groq` | Groq | ![Development](https://img.shields.io/badge/-development-blue) |
 | `ibm.watsonx.ai` | IBM Watsonx AI | ![Development](https://img.shields.io/badge/-development-blue) |
 | `mistral_ai` | Mistral AI | ![Development](https://img.shields.io/badge/-development-blue) |
 | `openai` | OpenAI | ![Development](https://img.shields.io/badge/-development-blue) |
 | `perplexity` | Perplexity | ![Development](https://img.shields.io/badge/-development-blue) |
-| `vertex_ai` | Vertex AI | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 111: removing enum members is a breaking change, so we don't allow it, please keep the old ones as deprecated, check out this as an example:

https://github.com/open-telemetry/semantic-conventions/blob/1830ef2d0bd77a4a623bb49d686de4863c1ab6c2/model/gen-ai/registry.yaml#L179-L183

model/gen-ai/registry.yaml
```diff
@@ -13,14 +13,31 @@ groups:
...
+            - id: gcp
+              stability: development
+              value: "gcp"
+              brief: "Any Google generative AI endpoint"
+              note: "May be used when specific backend is unknown. May be accompanied by 'gcp.client.service' attribute differentiating the endpoint."
+            - id: gcp.vertex_ai
+              stability: development
+              value: "gcp.vertex_ai"
+              brief: 'Vertex AI'
+              note: "This refers to the 'aiplatform.googleapis.com' endpoint. May be accompanied by 'gcp.client.service' with value 'aiplatform'. May make use of common attributes from 'gcp.gen_ai.' prefix."
```
lmolkova, line 25:
```suggestion
              note: > 
                This refers to the 'aiplatform.googleapis.com' endpoint. 
```

Instead of mentioning extra attributes here, we should formally list all appliable attributes in the Google-specific document in the https://github.com/open-telemetry/semantic-conventions/tree/main/docs/gen-ai folder.

It can be done in a separate PR.

michaelsafyan: Thanks. I removed the mention of `gcp.client.service`. I'll look to add some of this in a separate PR.

docs/attributes-registry/gen-ai.md
michaelsafyan: OK to resolve?

model/gen-ai/registry.yaml
```diff
@@ -13,9 +13,9 @@ groups:
               stability: development
               value: "openai"
               brief: 'OpenAI'
-            - id: gcp
+            - id: gcp.gen_ai
               stability: development
-              value: "gcp"
+              value: "gcp.gen_ai"
               brief: "Any Google generative AI endpoint"
               note:  >
```
lmolkova, line 20:
```suggestion
              note: >
```

lmolkova (approved):

gyliu513 (approved):

jsuereth (approved):

PR #2066, opened 2025-04-08
events: remove uncaught exception example

pellared:
The uncaught exception in an controversial example of an event.

Why: https://github.com/open-telemetry/opentelemetry-specification/pull/4333


pellared:
Does it need a changelog entry?

trask (approved):

lmolkova (approved):

PR #2070, opened 2025-04-08
Introduce app click events

breedx-splk:
## Changes

This introduces two new click event definitions:
* `app.screen.click` - for a user clicking somewhere on a screen
* `app.widget.click` - for a user clicking on a GUI widget

The `docs/app/app.md` file is net new (didn't exist previously).


cleverchuk (approved):
LTGM except that `app.widget.id` and `app.widget.name` have same description throughout. It looks like it was intentional?

LikeTheSalad (approved):
Cheers!

docs/app/app.md
```diff
@@ -0,0 +1,117 @@
...
+| [`app.screen.coordinate.x`](/docs/attributes-registry/.md) | int | The x (horizontal) coordinate of a screen coordinate, in screen pixels. | `0`; `131` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.screen.coordinate.y`](/docs/attributes-registry/.md) | int | The y (vertical) component of a screen coordinate, in screen pixels. | `12`; `99` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.widget.id`](/docs/attributes-registry/.md) | string | An identifier that uniquely differentiates this widget from other widgets in the same application. [2] | `f9bc787d-ff05-48ad-90e1-fca1d46130b3`; `submit_order_1829` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.widget.name`](/docs/attributes-registry/.md) | string | The name of an application widget. [3] | `submit`; `attack`; `Clear Cart` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
bidetofevil, line 26: This is a human-readable description or the label that is seen on screen? Do we need to specify what language it is in another attribute or do we simply add something in the description to offer guidence?

docs/app/app.md
```diff
@@ -0,0 +1,117 @@
...
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`app.installation.id`](/docs/attributes-registry/.md) | string | A unique identifier representing the installation of an application on a specific device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.screen.coordinate.x`](/docs/attributes-registry/.md) | int | The x (horizontal) coordinate of a screen coordinate, in screen pixels. | `0`; `131` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
bidetofevil, line 23: I think very reasonable requirements will need to use different systems, e.g. screen/window percentage, device pixels vs physical pixels, 0,0 being the top left of a screen vs top left of the not-full-size window, given the different device screen size and density that exists, not to menu windows that aren't full screen.

As such, do we want to add an attribute that is effectively an enum that describes that system the x-y map to?

We can make it optional and default to something

docs/app/app.md
```diff
@@ -0,0 +1,117 @@
...
+
+The event name MUST be `app.screen.click`.
+
+This event represents an instantaneous click on the screen of an application.
+
+The `app.screen.click` event can be used to indicate that a user has clicked or tapped on the screen portion of an application. Clicks outside of an application's active area SHOULD NOT generate this event. This event does not differentiate between touch/mouse down and touch/mouse up. Implementations SHOULD give preference to generating this event at the time the click is complete, typically on touch release or mouse up. The location of the click event MUST be provided in absolute screen pixels.
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`app.screen.coordinate.x`](/docs/attributes-registry/app.md) | int | The x (horizontal) coordinate of a screen coordinate, in screen pixels. | `0`; `131` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
```
bidetofevil, line 76: Similar comments as the other one

docs/app/app.md
```diff
@@ -0,0 +1,117 @@
...
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`app.installation.id`](/docs/attributes-registry/.md) | string | A unique identifier representing the installation of an application on a specific device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.screen.coordinate.x`](/docs/attributes-registry/.md) | int | The x (horizontal) coordinate of a screen coordinate, in screen pixels. | `0`; `131` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.screen.coordinate.y`](/docs/attributes-registry/.md) | int | The y (vertical) component of a screen coordinate, in screen pixels. | `12`; `99` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
bidetofevil, line 24: Do we want to make the coordinates optional so folks can record a click without necessarily putting in coordinates that they aren't going to use?

model/app/registry.yaml
```diff
@@ -31,3 +31,34 @@ groups:
           More information about Android identifier best practices can be found [here](https://developer.android.com/training/articles/user-data-ids).
         examples:
           - 2ab2916d-a51f-4ac8-80ee-45ac31a28092
+      - id: app.screen.coordinate.x
+        type: int
+        brief: The x (horizontal) coordinate of a screen coordinate, in screen pixels.
+        stability: development
+        examples: [ 0, 131 ]
+      - id: app.screen.coordinate.y
```
trask, line 39: @breedx-splk @MSNev do you think we can be consistent (one way or the other) with how we represent coordinates across client instrumentation?

related to https://github.com/open-telemetry/semantic-conventions/pull/1941#discussion_r2001519799

docs/app/app.md
```diff
@@ -0,0 +1,117 @@
...
+<!-- semconv registry.app -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`app.installation.id`](/docs/attributes-registry/.md) | string | A unique identifier representing the installation of an application on a specific device [1] | `2ab2916d-a51f-4ac8-80ee-45ac31a28092` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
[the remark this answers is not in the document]
breedx-splk: This is a preexisting attribute, but the markdown doc is new. The entry for this attribute is auto-generated off of the existing yaml. If you want to take this up, please open a separate issue.

model/app/registry.yaml
breedx-splk: > @breedx-splk @MSNev do you think we can be consistent (one way or the other) with how we represent coordinates across client instrumentation?
> 
> related to [#1941 (comment)](https://github.com/open-telemetry/semantic-conventions/pull/1941#discussion_r2001519799)

Seems unlikely, especially if browser is pushing for a 2-element array. That's way too magical for me.

Additionally, having separate fields allows filtering/processing/whatever without parsing/splitting/extracting.

breedx-splk:
> LTGM except that `app.widget.id` and `app.widget.name` have same description throughout. It looks like it was intentional?

Yeah, so there are two separate fields per attribute -- the `brief` and the `note`. Correct me if I'm wrong, but I believe that the `brief` is intended to be a short description, typically one or two short sentences. The `note` is there to provide additional clarity/explanation. I thought it important to note what a widget is, because there isn't a lot of prior art in semconv so far.

If you feel like it's too redundant or if we can consolidate that somewhere else, I'm open to suggestions.

model/app/registry.yaml
MSNev: We also don't have this as a top level attribute as it will be part of the event content and not directly comparable across different events.

nit: The `type` says `int`, but you example is an array?

docs/app/app.md
breedx-splk: Same reply. Closing the thread here so we can continue in one place above.

docs/app/app.md
breedx-splk: Yes, the name is expected to be human-readable. I am not aware of other attributes providing for language, but that's something that could be expanded upon later. This isn't perfection.

model/app/registry.yaml
breedx-splk: > nit: The type says int, but you example is an array?

That's a yaml array, containing more than one integer example. Would it help if I expanded those to a multi-line list format?

trask (changes requested):
blocking temporarily to discuss whether we can unify screen coordinate representation with #1941

I've added this topic to next week's client instrumentation SIG meeting agenda

docs/app/app.md
bidetofevil: Sounds good.

docs/app/app.md
bidetofevil: Sure. This can be evolved. Stub values could also be used.

bidetofevil (approved):
LGTM. Additional attributes suggested can be added later on.

docs/app/app.md
```diff
@@ -0,0 +1,117 @@
...
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `app.screen.click`.
+
+This event represents an instantaneous click on the screen of an application.
+
+The `app.screen.click` event can be used to indicate that a user has clicked or tapped on the screen portion of an application. Clicks outside of an application's active area SHOULD NOT generate this event. This event does not differentiate between touch/mouse down and touch/mouse up. Implementations SHOULD give preference to generating this event at the time the click is complete, typically on touch release or mouse up. The location of the click event MUST be provided in absolute screen pixels.
```
jzwc, line 72: How is "application's active area" defined?

breedx-splk: It's not defined...but as written here it's intended to mean the area of the screen that is displaying the application.

trask:
> blocking temporarily to discuss whether we can unify screen coordinate representation with #1941

See #2145

docs/app/app.md
```diff
@@ -10,47 +10,7 @@ This document defines events and attributes related to client-side applications
...
-
-**[2] `app.widget.id`:** A widget is an application component, typically an on-screen visual GUI element.
-
-**[3] `app.widget.name`:** A widget is an application component, typically an on-screen visual GUI element.
-
-<!-- markdownlint-restore -->
-<!-- prettier-ignore-end -->
-<!-- END AUTOGENERATED TEXT -->
-<!-- endsemconv -->
+See the [app attributes](/docs/attributes-registry/app.md) registry for details.
```
lmolkova: not a strong opinion, but I'd rather not even reference all possible app attributes here. We consider attribute registry as a giant pool of might-be-related things and try to de-prioritize it a bit comparing to the telemetry signals.

If you feel super=strong about keeping it, I'd suggest something like

 ```suggestion
See the [app attributes](/docs/attributes-registry/app.md) registry for all application-related attributes that may appear on telemetry items.
```

I'd also move this section down, to be after all event definitions

File: docs/app/app.md [hunk 9]
```diff
@@ -10,47 +10,7 @@ This document defines events and attributes related to client-side applications
 
 ## Attributes
```
lmolkova, line 11: could you please add TOC?

docs/app/app.md
```diff
@@ -0,0 +1,77 @@
...
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`app.screen.coordinate.x`](/docs/attributes-registry/app.md) | int | The x (horizontal) coordinate of a screen coordinate, in screen pixels. | `0`; `131` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.screen.coordinate.y`](/docs/attributes-registry/app.md) | int | The y (vertical) component of a screen coordinate, in screen pixels. | `12`; `99` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+<!-- markdownlint-restore -->
```
lmolkova, line 39: are these two attributes sufficient to conclude anything useful? Or would we expect this event to be reported in a scope of a span? 

I'd imagine you'd want to record `user`, `enduser` and/or application name, device/brower info, os type, etc, would these be recorded as resources? If so, would be great to mention the list of resources that would make it complete in a free-form markdown text.

breedx-splk: Yes, they tell you where the user clicked on a screen. That's useful. 

I think `user` or `enduser` or whatever could be added later as needed. The app, device, os, etc. are on the resource. Is there prior art that lists out a resource like you're asking? Does it have to be in this PR?

trask (approved):

docs/app/app.md
lmolkova: let's postpone it until https://github.com/open-telemetry/semantic-conventions/pull/2123 lands - it adds a formal way to document the association

lmolkova (approved):

docs/app/app.md
```diff
@@ -0,0 +1,83 @@
...
+# App
+
+**Status**: [Development][DocumentStatus]
+
+This document defines events related to client-side applications
+(e.g. web apps or mobile apps).
+
+* Click events
+  * [`app.screen.click`](#event-appscreenclick)
+  * [`app.widget.click`](#event-appwidgetclick)
```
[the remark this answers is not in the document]
breedx-splk: Yes, but not before I say that all of this is so over-engineered.  🤣

docs/app/app.md
jzwc: I see. But is it even possible for an agent in an instrumented application to detect clicks outside of the application screen real estate? Is there something similar specified in, e.g., network instrumentation (that network requests made outside of the instrumented application should not be recorded)?

docs/app/app.md
```diff
@@ -0,0 +1,88 @@
...
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `app.screen.click`.
+
+This event represents an instantaneous click on the screen of an application.
+
+The `app.screen.click` event can be used to indicate that a user has clicked or tapped on the screen portion of an application. Clicks outside of an application's active area SHOULD NOT generate this event. This event does not differentiate between touch/mouse down and touch/mouse up. Implementations SHOULD give preference to generating this event at the time the click is complete, typically on touch release or mouse up. The location of the click event MUST be provided in absolute screen pixels.
```
jzwc, line 38: The definition of "location of the click" seems a bit ambiguous. 

Is it meant to refer to the device’s absolute hardware display coordinates, or to coordinates relative to the application's visible display area?

I suspect it is intended to be the latter, as absolute display coordinates are often inaccessible on platforms where applications are sandboxed and cannot determine their exact position on the physical screen (for example, in multitasking scenarios on an iPad).

Providing an explicit definition would greatly improve the clarity of this text.

breedx-splk: Oh I think I disagree. I think you'd have to be pretty creative indeed to try and specify coordinates for a "screen" click event whose location is specified in "screen pixels" with "application visible area".  If you have specific phrasing you would like to add or edit to clarify, I'm open to suggestions though.

It's definitely not intended to be application visible area. It's intended here, for now, as a first pass, to be absolute screen coordinates.

I understand that this may not be easy (or possible?) on all platforms. I would expect this semconv to evolve over time to include other coordinate systems. This first pass isn't trying to be comprehensive or perfect....and one could imagine another field/enum on the event that later indicates which coordinate system is used. Alternatively, there could be another event called `app.window.click` or `app.visible_area.click` whose coordinates could be relative to app visible area or window or whatever.

model/app/registry.yaml
```diff
@@ -31,3 +31,34 @@ groups:
           More information about Android identifier best practices can be found [here](https://developer.android.com/training/articles/user-data-ids).
         examples:
           - 2ab2916d-a51f-4ac8-80ee-45ac31a28092
+      - id: app.screen.coordinate.x
```
trask, line 34:
```suggestion
      - id: app.screen.coordinate.xy
```

model/app/registry.yaml
```diff
@@ -31,3 +31,28 @@ groups:
           More information about Android identifier best practices can be found [here](https://developer.android.com/training/articles/user-data-ids).
         examples:
           - 2ab2916d-a51f-4ac8-80ee-45ac31a28092
+      - id: app.screen.xy
```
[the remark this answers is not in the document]
breedx-splk: You can watch the last 2 recordings of the client sig if you want to completely understand how we ended up here. It's a compromise between being understandable and being optimized for size/performance.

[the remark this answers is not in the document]
breedx-splk: I do agree that it does go against the new guidance, and I still prefer to have separate fields.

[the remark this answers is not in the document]
lmolkova: Could someone with a strong desire to have int array to come forward and document the reasons behind this? Or come present it at the semconv SIG call?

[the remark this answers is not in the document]
breedx-splk: @MSNev ^^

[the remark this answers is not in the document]
MSNev: Adding @trask as the owner of the above issue and @tedsuo, as they were both also part of the discussions.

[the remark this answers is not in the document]
trask: > Or come present it at the semconv SIG call?

this seems pretty reasonable, and what has been asked of other SIGs when they have decided not to follow general semantic convention recommendations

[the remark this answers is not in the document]
lmolkova: So what I believe this analysis shows is that if all the application did was encoding coordinates as OTLP anyvalue, it would result in noticeable perf impact. I assume that client instrumentation will do a lot of other things beyond encoding coordinates meaning that perf implications of this decision in real life might be negligeable.

What this analysis does not include is:
- instrumentation ergonomics
- query/dashboard/end-user ergonomics

In OTel semconv and API we usually value ergonomics more than micro-benchmarking results.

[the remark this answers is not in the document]
jsuereth: To add one to what @lmolkova states, I think we can't myopically look at just OTLP-as-it-is today for this ONE event and optimise `xy` over `.x`+`.y` here.

A few higher level points:

- We're investigating a columnar approach to both events and metric storage that would NOT benefit from this micro-optimization in OTLP-Arrow.
- Profiling is introducing a dictionary concept that I'd like to more consistently use across OTLP that would actually dramatically improve the performance of events, e.g. for client-side instrumetnation.
- The overall shape of OTLP is *rather* expensive for client side events.  If we have these arguments, nickle and dime-ing performance out of standardized conventions, I think it's a losing battle.  I'd rather have a dedicate protocol (or protocol extensions) that are optimal for the client-side sig.  I think this kind of deviation for performance sake is a bit too acute and hurts the long term success of the SIG, where we *really* need benchmarks (both of sending the data, and querying it) with OTLP or OTLP alternatives.

I'm blocking this PR as I'd want to see us resolve two things first:

1. What is Client SIG's plans for optimal event communication such that we DON'T have repeated attempts to violate conventions across OTel.
2. Should we update our guidance on arrays to allow "sized vectors" as a conceptual thing, e.g. a vector-2 (xy), a vector-3 (xyz), etc.?   This is the only argument I could see for this type of convention that I'd be on board with, but we should make sure the general guidance allows and recommends this.

jsuereth (changes requested):

jsuereth:
I'm blocking this PR until https://github.com/open-telemetry/semantic-conventions/pull/2070#discussion_r2064061733 is resolved.

breedx-splk:
Thanks @jsuereth. To clarify for the record, and because this is my PR, I definitely prefer having separate fields for x and y. I made a concession and agreed to do the array thing purely as an attempt to move this new experimental spec forward. 

I believe that it's probably a mistake to try and micro-optimize at this level, especially before any usage has been established.

model/app/registry.yaml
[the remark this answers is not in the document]
MSNev: > 2. Should we update our guidance on arrays to allow "sized vectors" as a conceptual thing, e.g. a vector-2 (xy), a vector-3 (xyz), etc.? This is the only argument I could see for this type of convention that I'd be on board with, but we should make sure the general guidance allows and recommends this.

Which is effectively what has been proposed.

Points are typically used, processed, and compared as combinations (e.g., X/Y pairs). Handling them as single entities across the entire pipeline—covering creation, processing, encoding, decoding, storing, retrieving, and querying—avoids the overhead associated with managing multiple objects (attributes/key value-pairs). This approach also mitigates issues related to recombining attributes and handling edge cases where one attribute might be dropped due to limits, not just the concerns around OTLP handling.

breedx-splk:
Based on the above and today's client SIG, I've reverted the last commit to better adhere to the existing guidance.

lmolkova (approved):

jsuereth (approved):
My concerns are now resolved, thanks!

Issue #1571, opened 2024-11-12
Docker images not qualified, leading to errors when running `make check-policies`


michaelsafyan:
### Area(s)

area:other

## Description

Docker containers used in `Makefile` are not qualified leading to errors:

```
[~/Code/github/michaelsafyan/open-telemetry.semantic-conventions (refattr)]$ make check-policies
docker run --rm \
        --mount 'type=bind,source=/home/michaelsafyan/Code/github/michaelsafyan/open-telemetry.semantic-conventions/policies,target=/home/weaver/policies,readonly' \
        --mount 'type=bind,source=//home/michaelsafyan/Code/github/michaelsafyan/open-telemetry.semantic-conventions/model,target=/home/weaver/source,readonly' \
        otel/weaver:v0.10.0 registry check \
        --registry=/home/weaver/source \
        --baseline-registry=https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.28.0.zip[model] \
        --policy=/home/weaver/policies
Error: short-name "otel/weaver:v0.10.0" did not resolve to an alias and no unqualified-search registries are defined in "/etc/containers/registries.conf"
make: *** [Makefile:249: check-policies] Error 125
```

## Steps to Reproduce (if any)

1. Install `podman-is-docker` so that the `docker` command becomes an alias over `podman`
2. Run `make check-policies` inside the repo

## Expected Result

Should work

## Actual Result

Fails, because top-level `dependencies.Dockerfile` does not qualify the dependencies

### Semantic convention version

3c959946a0dcee57e63843790f0c0c7810f897b0

PR #2080, opened 2025-04-09
[chore] Make build fully support podman-is-docker.

### LINKED ISSUE #2052: Build operation "make check" fails with "podman-is-docker"
There appear to be a few issues:

 1. `markdown-lint-check` references local files but doesn't use `$(DOCKER_USER_IS_HOST_USER_ARG)`
 2. Container `lycheeverse/lychee` is not prefixed (should add "docker.io/" prefix)
 3. `table-check` references local files but doesn't use `$(DOCKER_USER_IS_HOST_USER_ARG)`
 4. `check-policies`  references local files but doesn't use `$(DOCKER_USER_IS_HOST_USER_ARG)`
 5. `compatibility-check` references local files but doesn't use `$(DOCKER_USER_IS_HOST_USER_ARG)`
 6. `schema_check.sh` uses unqualified container (`otel/build-tool-schemas`); needs `docker.io/` prefix.
 7. `schema_check.sh` omits necessary user parameters given reference to local files

It would be ideal to not only fix this issue, but to add a GitHub workflow that attempts to build the "Makefile" in an environment that uses "podman-is-docker" to ensure ongoing build compatibility with podman.

michaelsafyan:
Fixes #2052 #1571

## Changes

Updates the build machinery to:

 - Always specify the container repository
 - Always specify user/group parameters when using mount points which require local file access
 - Always specify `PODMAN_USERNS=keep-id` when using the user/group parameters

Also adds a GitHub workflow aimed at regression-proofing the podman-is-docker support.


michaelsafyan:
Thanks, @trask ! Looking into the cause...

.github/workflows/checks.yml
```diff
@@ -86,13 +95,33 @@ jobs:
...
+  semantic-conventions-registry-podman:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: install podman and podman-docker
+      run: sudo apt-get install podman podman-docker
+    - name: verify registry tables
+      run: |
+        make attribute-registry-generation
+        git diff --exit-code './docs/attributes-registry/*.md' || (echo 'Attribute registry markdown is out of date, please run "make attribute-registry-generation" and commit the changes in this PR.' && exit 1)
```
trask, line 107: what do you think of using a matrix build (e.g. podman `false`/`true`) to avoid the worry of keeping these jobs synchronized?

Makefile
```diff
@@ -122,19 +123,22 @@ normalized-link-check:
...
 markdown-link-check: normalized-link-check
 	$(DOCKER_RUN) --rm \
+	    $(DOCKER_USER_IS_HOST_USER_ARG) \
 		--mount 'type=bind,source=$(PWD),target=/home/repo' \
-		lycheeverse/lychee \
+		$(LYCHEE_CONTAINER) \
 		--config home/repo/.lychee.toml \
 		--root-dir /home/repo \
 		--verbose \
+		--timeout=60 \
```
trask, line 132: thanks, we just started seeing timeout errors recently, which maybe aligns with this release: https://github.com/lycheeverse/lychee-action/releases/tag/v2.4.0, though I can't see anything in the change log for https://github.com/lycheeverse/lychee/releases/tag/lychee-v0.18.1 to explain it

I'm wondering if it the retry isn't occurring after timeout, but this is a good idea to increase from the default 20s to see if that helps 👍

trask: I'm going to re-run the CI check a few times on your PR to see if we get consistently better results now...

michaelsafyan: Thanks. Good to know that this is likely unrelated to this change. I was wondering if it was related to GNU throttling traffic from GitHub (rather than anything to do with the lychee dependency).

.github/workflows/checks.yml
```diff
@@ -4,10 +4,18 @@ on:
...
+jobs:  
+
   markdownlint:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        container-runner: ["docker", "podman"]
     steps:
+    - name: install podman and podman-docker
+      if: ${matrix.container-runner}=="podman"
```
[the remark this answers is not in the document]
michaelsafyan: Thanks! Updated.

.github/workflows/checks.yml
michaelsafyan: Done. Switched to using the matrix mechanism. Thanks for the tip.

.github/workflows/checks.yml
```diff
@@ -4,10 +4,18 @@ on:
   push:
   pull_request:
 
-jobs:
+jobs:  
+
```
trask, line 8:
```suggestion
jobs:
```

.github/workflows/checks.yml
```diff
@@ -36,15 +44,23 @@ jobs:
...
     - name: install dependencies
       run: npm install
 
+    - name: install podman and podman-docker
+      if: ${{ matrix.container-runner == 'podman' }}
+      run: sudo apt-get install podman podman-docker
+
     - name: run markdown-link-check
       run: make markdown-link-check
+      timeout-minutes: 5
```
trask, line 63: I think it's ok to leave this off
```suggestion
```

Makefile
trask, line 132: unfortunately I'm still seeing sporadic timeouts, but let's keep this change anyways 👍

trask (approved):

.github/workflows/checks.yml
michaelsafyan: Done. Removed trailing whitespace.

.github/workflows/checks.yml
michaelsafyan: Done. Removed this line.

lmolkova (approved):

lmolkova:
Do we have to run two copies of each check in the CI? It's especially problematic for the link check which is long and we know people are blocking link checker (presumably for being spammy).

The one that's really useful locally is  `make table-generation attribute-registry-generation`, i.e. `semantic-conventions`  in the CI - if we can test it works with both, it'd be very unlikely for others to be broken.

So I'd propose to run `semantic-conventions` with podman and docker and keep the rest on docker.

michaelsafyan:
> Do we have to run two copies of each check in the CI? It's especially problematic for the link check which is long and we know people are blocking link checker (presumably for being spammy).
> 
> The one that's really useful locally is `make table-generation attribute-registry-generation`, i.e. `semantic-conventions` in the CI - if we can test it works with both, it'd be very unlikely for others to be broken.
> 
> So I'd propose to run `semantic-conventions` with podman and docker and keep the rest on docker.

That's fair feedback, @lmolkova .

I've created a separate workflow file that is executed only when the components of the build system change for exercising podman, which I think should address the concern of this adding too much overhead in normal processes.

This new workflow does, however, execute all of the steps that are present in the "CONTRIBUTING.md" guide. My concern with omitting some of the steps is that it then becomes easy to break/block some of the processes/workflows that might be used during local development for those using podman.

trask:
not sure what's going on with CI, closing and re-opening to kick it

lmolkova:
the names of the checks have changed but we still have old names in the branch protection rules as required. Wonder if we can preserve the old checks? Otherwise I can go and update them

lmolkova:
ah, the latest commit refactors  how checks are done, but I don't think it does all the checks we used to run. 
Is there a need for it?

Can we have one check that validates basic features with podman, but leave the rest intact?

lmolkova:
thanks a lot @michaelsafyan !

PR #2086, opened 2025-04-11
Update weaver to 0.14.0 and allow `_` -> `.` renames

lmolkova:
Related to #1118 and #1982

The new weaver supports excluding attributes from code generation allowing us to rename `foo_bar` to `foo.bar`!
It can be done by providing the following annotation on attribute to be excluded:

```yaml
- id: messaging.client_id
  ....
  annotations:
    code_generation:
      exclude: true
```

### DIFF
--- a/model/kestrel/metrics.yaml
+++ b/model/kestrel/metrics.yaml
@@ -40,9 +40,10 @@ groups:
       - ref: tls.protocol.version
       # yamllint disable rule:line-length
       - ref: error.type
-        brief: Captures the error type when a connection fails.
+        brief: The full name of exception type.
         requirement_level:
           conditionally_required: if and only if an error has occurred.
+        examples: ['System.OperationCanceledException', 'Contoso.MyException']
         # TODO: move note to yaml once https://github.com/open-telemetry/build-tools/issues/192 is supported
         note: |
           Starting from .NET 9, Kestrel `kestrel.connection.duration` metric reports
--- a/model/kestrel/metrics.yaml
+++ b/model/kestrel/metrics.yaml
@@ -38,14 +38,11 @@ groups:
       - ref: network.protocol.version
         examples: ["1.1", "2"]
       - ref: tls.protocol.version
-      - ref: error.type
-        brief: The full name of exception type.
-        requirement_level:
-          conditionally_required: if and only if an error has occurred.
-        note: "Captures the exception type when a connection fails."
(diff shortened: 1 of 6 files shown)


model/kestrel/metrics.yaml
```diff
@@ -38,14 +38,12 @@ groups:
...
       - ref: tls.protocol.version
+      # yamllint disable rule:line-length
       - ref: error.type
         brief: The full name of exception type.
         requirement_level:
           conditionally_required: if and only if an error has occurred.
-        note: "Captures the exception type when a connection fails."
         examples: ['System.OperationCanceledException', 'Contoso.MyException']
-      # yamllint disable rule:line-length
-      - ref: error.type
```
lmolkova, line 48: was referenced twice which is now flagged by weaver

model/kestrel/metrics.yaml
```diff
@@ -38,14 +38,12 @@ groups:
       - ref: network.protocol.version
         examples: ["1.1", "2"]
       - ref: tls.protocol.version
+      # yamllint disable rule:line-length
       - ref: error.type
         brief: The full name of exception type.
         requirement_level:
           conditionally_required: if and only if an error has occurred.
-        note: "Captures the exception type when a connection fails."
         examples: ['System.OperationCanceledException', 'Contoso.MyException']
```
trask, line 46: (using the examples from below)
```suggestion
        examples: ['connection_reset', 'invalid_handshake']
```

model/kestrel/metrics.yaml
trask: cool, good catch weaver 👍

policies/attribute_name_collisions.rego
```diff
@@ -82,11 +87,6 @@ is_property_set(obj, property) = true if {
...
 } else = false
 
-# This list contains exceptions for existing collisions that were introduced unintentionally.
-# We'll have a way to specify how collision resolution happens in the schema -
-# see phase 2 in https://github.com/open-telemetry/semantic-conventions/issues/1118#issuecomment-2173803006
-# For now we'll exclude existing collisions from the checks.
-# ADDING NEW EXCEPTIONS IS NOT ALLOWED.
-
-# DO NOT ADD ATTRIBUTES TO THIS LIST
-excluded_const_collisions := {"messaging.client_id"}
```
trask, line 92: 🎉

trask (approved):

lquerel (approved):
Nice!

PR #2088, opened 2025-04-11
Add standardized recommended system.cpu.utilization metric

rogercoll:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1873

## Changes


docs/system/system-metrics.md
```diff
@@ -143,6 +144,28 @@ This metric is [recommended][MetricRecommended].
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+### Metric: `system.cpu.utilization`
+
+This metric is [recommended][MetricOptIn].
```
trask, line 149: this seems pretty useful from a U/X perspective (even if it's just calculated from other metrics)

and it shouldn't be too costly given it has no buckets or attributes

what do you think about making it recommended?

rogercoll: I agree to the statement, it simplifies derived queries (and some backends might not be able to derive it).

@open-telemetry/semconv-system-approvers wdyt?

AlexanderWert: +1 on making it recommended

model/system/metrics.yaml
```diff
@@ -32,6 +32,20 @@ groups:
     unit: "{cpu}"
     attributes: []
 
+  - id: metric.system.cpu.utilization
+    type: metric
+    metric_name: system.cpu.utilization
+    stability: development
+    brief: Fraction of total CPU time spent in active (non-idle) modes.
```
ChrsMark, line 39: Should we mention that it's the summary for all of the cores of the system/host?

model/system/metrics.yaml
```diff
@@ -32,6 +32,20 @@ groups:
...
+    note: >
+      This metric provides a standardized and opinionated definition of CPU utilization.
+      It is defined as the proportion of CPU time spent in non-idle and non-iowait modes, following a formula similar to: 1 - (idle + iowait).
```
ChrsMark, line 42: Since this is a derived metric, could we become more descriptive here providing the exact formula that should be used for this calculation from the existing metrics?

I think it'd be extremely useful to show how we can calculate this from the existing `cpu.time` metrics for example.

model/system/metrics.yaml
rogercoll: Definitely, added in https://github.com/open-telemetry/semantic-conventions/pull/2088/commits/fc50a970aef5e44cef24f2fb82f26a69942ff84f

docs/system/system-metrics.md
ChrsMark: I remember we had been mentioning that only `cpu.time` should be considered as recommended or required and all the others that can be derived from it should be opt-in. However, if we all agree on making an exception for this single metric I would be fine.

model/system/metrics.yaml
rogercoll: Good point, let me know what you think of  https://github.com/open-telemetry/semantic-conventions/pull/2088/commits/fc50a970aef5e44cef24f2fb82f26a69942ff84f

model/system/metrics.yaml
```diff
@@ -36,10 +36,11 @@ groups:
     type: metric
     metric_name: system.cpu.utilization
     stability: development
-    brief: Fraction of total CPU time spent in active (non-idle) modes.
+    brief: Fraction of total CPU time spent in active modes, aggregated across all system CPU cores.
     note: >
       This metric provides a standardized and opinionated definition of CPU utilization.
-      It is defined as the proportion of CPU time spent in non-idle and non-iowait modes, following a formula similar to: 1 - (idle + iowait).
+      It is defined as the proportion of total system CPU time spent in non-idle and non-iowait modes — that is, the time spent in user, system, nice, interrupt, and steal modes. It is computed by aggregating `cpu.utilization` across all logical CPUs for these modes, and dividing the result by `system.cpu.logical.count` to produce a normalized, system-wide utilization value between 0.0 and 1.0:
+        system.cpu.utilization = sum(cpu.utilization{mode != "idle" and mode != "iowait"}) / system.cpu.logical.count
```
ChrsMark: Shall we also add the `cpu.time` corresponding one?

If I don't miss anything that should be the following: `system.cpu.utilization = sum(rate(cpu.time{mode != "idle" and mode != "iowait"})) / system.cpu.logical.count`

rogercoll: Good point, is the reasoning behind this proposal `cpu.utilization` being opt-in and `cpu.time` recommended?

docs/system/system-metrics.md
trask: > seeking alignment with other similar metrics

❤️

dmitryax: > Since for system.* we don't have a system.cpu.time or something similar I assume we can have it as recommended as it's the only one available for the whole system/host.

We do have `system.cpu.time` and it's being emitted [by default](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/cpuscraper/metadata.yaml#L25) in the collector's host metrics receiver.

> I remember we had been mentioning that only cpu.time should be considered as recommended or required and all the others that can be derived from it should be opt-in. However, if we all agree on making an exception for this single metric I would be fine.

This is what we've been following in the collector. I believe this is something that needs clarification in the OTel Spec.

dmitryax: > So it seems that cpu.utilization(per core)==cpu.usage. I wonder if that would make sense to rename it to cpu.usage to achieve alignment with other areas where we have this triplet of metrics: (https://github.com/search?q=repo%3Aopen-telemetry%2Fsemantic-conventions%20cpu.usage&type=code).

If aggregated over cpu cores, they are not the same. `cpu.utilization` is always within [0,1] range, while `cpu.usage` is within [0,<number of cores>]. I think both of them should be defined along with `cpu.time`, and only one of them should be recommended.

dmitryax: I've submitted https://github.com/open-telemetry/semantic-conventions/issues/2177 with an intention to define generic guidelines such metrics

model/system/metrics.yaml
rogercoll: changed in https://github.com/open-telemetry/semantic-conventions/pull/2088/commits/db4b15eb177ec716d95832f6bf7fdc5b4e1cab15

model/system/metrics.yaml
```diff
@@ -32,6 +32,22 @@ groups:
...
+    metric_name: system.cpu.utilization
+    stability: development
+    brief: Fraction of total CPU time spent in active modes, aggregated across all system CPU cores.
+    note: >
+      This metric provides a standardized and opinionated definition of CPU utilization.
+      It is defined as the proportion of total system CPU time spent in non-idle and non-iowait modes — that is, the time spent in user, system, nice, interrupt, and steal modes.
+      It is computed by aggregating `cpu.time` across all logical CPUs for these modes over a measurement interval, and dividing the result by the elapsed time and the `system.cpu.logical.count` to produce a normalized,
+      system-wide utilization value between 0.0 and 1.0: sum(rate(cpu.time{mode != "idle" and mode != "iowait"})) / system.cpu.logical.count
+      This represents time spent actively executing work, including time delayed by the hypervisor (steal), which reflects CPU demand in virtualized environments.
+      As a result, in cases of high CPU contention — such as 100% steal time — this metric will report 100% utilization, making such issues easy to detect.
```
ChrsMark, line 46: https://github.com/open-telemetry/semantic-conventions/pull/2123 is not merged yet but maybe we can put a note here about resource/entity association already (similar to https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/metrics.yaml#L153-L154)

rogercoll: Added https://github.com/open-telemetry/semantic-conventions/pull/2088/commits/b7ef1550900d53b5ae619b7cb6ac44f6f4cb2817

model/system/metrics.yaml
```diff
@@ -32,6 +32,26 @@ groups:
...
+    note: >
+      This metric provides a standardized and opinionated definition of CPU utilization.
+      It is defined as the proportion of total system CPU time spent in non-idle and non-iowait modes — that is, the time spent in user, system, nice, interrupt, and steal modes.
+      It is computed by aggregating `cpu.time` across all logical CPUs over a measurement interval,
+      and dividing the result by the elapsed time and the `system.cpu.logical.count` to produce a normalized,
+      system-wide utilization value between 0.0 and 1.0: sum(rate(cpu.time{mode != "idle" and mode != "iowait"})) / system.cpu.logical.count
+      This represents time spent actively executing work, including time delayed by the hypervisor (steal), which reflects CPU demand in virtualized environments.
+      As a result, in cases of high CPU contention — such as 100% steal time — this metric will report 100% utilization, making such issues easy to detect.
+
+      This metric SHOULD, at a minimum, be reported against a
```
lmolkova, line 49: it should be possible to record it in formal way now https://github.com/open-telemetry/semantic-conventions/pull/2123

model/system/metrics.yaml
```diff
@@ -32,6 +32,26 @@ groups:
...
+    stability: development
+    brief: Fraction of total CPU time spent in active modes, aggregated across all system CPU cores.
+    note: >
+      This metric provides a standardized and opinionated definition of CPU utilization.
+      It is defined as the proportion of total system CPU time spent in non-idle and non-iowait modes — that is, the time spent in user, system, nice, interrupt, and steal modes.
+      It is computed by aggregating `cpu.time` across all logical CPUs over a measurement interval,
+      and dividing the result by the elapsed time and the `system.cpu.logical.count` to produce a normalized,
+      system-wide utilization value between 0.0 and 1.0: sum(rate(cpu.time{mode != "idle" and mode != "iowait"})) / system.cpu.logical.count
+      This represents time spent actively executing work, including time delayed by the hypervisor (steal), which reflects CPU demand in virtualized environments.
+      As a result, in cases of high CPU contention — such as 100% steal time — this metric will report 100% utilization, making such issues easy to detect.
```
lmolkova, line 47: nit, formatting

```suggestion
    note: |
      This metric provides a standardized and opinionated definition of CPU utilization.
      It is defined as the proportion of total system CPU time spent in non-idle and non-iowait modes — that is, the time spent in user, system, nice, interrupt, and steal modes.
      It is computed by aggregating `cpu.time` across all logical CPUs over a measurement interval,
      and dividing the result by the elapsed time and the `system.cpu.logical.count` to produce a normalized,
      system-wide utilization value between 0.0 and 1.0: `sum(rate(cpu.time{mode != "idle" and mode != "iowait"})) / system.cpu.logical.count`.
      This represents time spent actively executing work, including time delayed by the hypervisor (steal), which reflects CPU demand in virtualized environments.
      As a result, in cases of high CPU contention — such as 100% steal time — this metric will report 100% utilization, making such issues easy to detect.
```

model/system/metrics.yaml
rogercoll: great! added the `host` entity_association https://github.com/open-telemetry/semantic-conventions/pull/2088/commits/ab7cf55420f1c40a8ff94054a5940ef2fcbd4c4b

docs/system/system-metrics.md
```diff
@@ -143,6 +144,34 @@ This metric is [recommended][MetricRecommended].
...
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability | Entity Associations |
+| -------- | --------------- | ----------- | -------------- | --------- | ------ |
+| `system.cpu.utilization` | Gauge | `1` | Fraction of total CPU time spent in active modes, aggregated across all system CPU cores. [1] | ![Development](https://img.shields.io/badge/-development-blue) | `host` |
+
+**[1]:** This metric provides a standardized and opinionated definition of CPU utilization.
+It is defined as the proportion of total system CPU time spent in non-idle and non-iowait modes — that is, the time spent in user, system, nice, interrupt, and steal modes.
+It is computed by aggregating `cpu.time` across all logical CPUs over a measurement interval,
+and dividing the result by the elapsed time and the `system.cpu.logical.count` to produce a normalized,
+system-wide utilization value between 0.0 and 1.0: `sum(rate(cpu.time{mode != "idle" and mode != "iowait"})) / system.cpu.logical.count`.
+This represents time spent actively executing work, including time delayed by the hypervisor (steal), which reflects CPU demand in virtualized environments.
```
bertysentry, line 167: This calculation will be challenged by end users: "why don't you take into account iowait?", "why don't you remove steal time?", etc. People will never come to an agreement, because different they have different use cases and needs.

docs/system/system-metrics.md
```diff
@@ -143,6 +144,34 @@ This metric is [recommended][MetricRecommended].
...
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability | Entity Associations |
+| -------- | --------------- | ----------- | -------------- | --------- | ------ |
+| `system.cpu.utilization` | Gauge | `1` | Fraction of total CPU time spent in active modes, aggregated across all system CPU cores. [1] | ![Development](https://img.shields.io/badge/-development-blue) | `host` |
+
+**[1]:** This metric provides a standardized and opinionated definition of CPU utilization.
+It is defined as the proportion of total system CPU time spent in non-idle and non-iowait modes — that is, the time spent in user, system, nice, interrupt, and steal modes.
+It is computed by aggregating `cpu.time` across all logical CPUs over a measurement interval,
```
bertysentry, line 164: This calculation doesn't take into account the fact that all CPUs don't run at the same frequency: one core may be at 1GHz while another one is clocked at 2.5GHz at the same time.

A more precise calculation of a weighted average would be: `sum(delta(cpu.time{mode != "idle" and mode != "iowait"})) / sum(delta(cpu.time))`.

bertysentry:
I really think this needs more discussion (see [my comment](https://github.com/open-telemetry/semantic-conventions/issues/1873#issuecomment-2853933803)).

docs/system/system-metrics.md
bertysentry: But even the above calculation could be challenged, where one would want to see the utilization percentage based on the potential time when processor is running at full speed (or nominal speed, or whatever is meaningful).

If a core is running at half speed, it may show a 50% time utilization. But if it was running at full speed, it would have shown 25% time utilization for the same task. Knowing that clock speed varies based on the workload, maybe the 25% figure is more accurate than 50%? Who knows? 🤷‍♂️ 

My point being: I don't see the community agreeing on any definition of CPU utilization, as different operating systems already have very different definitions.

bertysentry: @ChrsMark The definition doesn't specify it, and the implementation is in fact affected by processor speed... but doesn't know it. Again, what I'm saying is: let's just collect `system.cpu.time` and let the backend calculate processor utilization as a rate of this metric. It's more efficient and more accurate, and it lets end users decide what they want to see.

braydonk:
We are still working on this. I'm working on a document to describe how we intend to approach our guidance for this, which will largely be "here is the best attempt at a utilization metric because it was in the conventions before, but for xyz reasons you should just use CPU time".

braydonk:
I misunderstood; I thought this was the issue not the PR. I'm not sure what we're planning to do with this one.

rogercoll:
Closing this issue as "cpu utilization" being a gauge over an implicit time window which might provide wrong values or that can be interpreted incorrectly.

We expect creating a document providing the mathematical details and guidance on the above statement. We might consider reopening this PR after the document/guidance, that is if we find a way to include all the necessary information (e.g include `collection_interval`) for backends to properly downsample "cpu utilization".

PR #2090, opened 2025-04-11
Pin lychee link checker version

trask:
It doesn't seem to have tag-based versions in dockerhub, only github commit sha based versions.

Also, it seems that `latest` runs the lastest nightly release, instead of the latest GitHub (tag) release.

Want to see if any recent lychee changes are related the recent link check flakiness we've seen.


lmolkova (approved):

trask:
> Issues found in 1 input. Find details below.
>
> [home/repo/docs/resource/process.md]:
[TIMEOUT] https://gcc.gnu.org/frontends.html | Timeout

😢

I think this one url has been particularly problematic, I think I'll exclude it from the link check...

lmolkova:
wdyt about making timeout not an error for link checker?  It's inconclusive and if the link disappears eventually it'll turn into 404 and then we'll react. Not sure if it's easily possible, my 5 min search didn't give a simple answer.

trask:
> wdyt about making timeout not an error for link checker? It's inconclusive and if the link disappears eventually it'll turn into 404 and then we'll react. Not sure if it's easily possible, my 5 min search didn't give a simple answer.

there's an option to configure which response codes are considered successful, but I don't see a way to treat timeouts as successful

joaopgrassi (approved):

PR #2093, opened 2025-04-12
Add `db.query.parameter.<key>`

trask:
Also keeping `db.operation.parameter` but that will not apply for parameterized queries.

For now moved `db.operation.parameter` from rc back to development. Not that I don't have confidence in the attribute, but it could maybe use a bit better explanation of when to use it vs `db.query.parameter`. I could easily be convinced to move it back to rc.

Partial reasoning is that `db.operation.parameter` looks a bit odd when there's no `db.operation.name` which is often the case when there's a `db.query.parameter`.

Another reason is that operations could have parameters + the underlying query text could be parameterized.

In general, this seems to make the general case of `db.query.text` + `db.query.parameter.*` clearer.


docs/database/redis.md
```diff
@@ -74,20 +73,15 @@ Instrumentations SHOULD document how `error.type` is populated.
 
 **[10] `db.query.text`:** For **Redis**, the value provided for `db.query.text` SHOULD correspond to the syntax of the Redis CLI. If, for example, the [`HMSET` command](https://redis.io/docs/latest/commands/hmset) is invoked, `"HMSET myhash field1 'Hello' field2 'World'"` would be a suitable value for `db.query.text`.
 
-**[11] `db.query.text`:** Non-parameterized query text SHOULD NOT be collected by default unless there is sanitization that excludes sensitive data, e.g. by redacting all literal values present in the query text.
+**[11] `db.query.text`:** Query text SHOULD NOT be collected by default unless there is sanitization that excludes sensitive data, e.g. by redacting all literal values present in the query text.
 See [Sanitization of `db.query.text`](/docs/database/database-spans.md#sanitization-of-dbquerytext).
-Parameterized query text SHOULD be collected by default (the query parameter values themselves are opt-in, see [`db.operation.parameter.<key>`](/docs/attributes-registry/db.md)).
```
trask, line 79: I didn't find a reference to redis parameterized queries so removed this

model/database/registry.yaml
```diff
@@ -122,6 +122,19 @@ groups:
           For batch operations, if the individual operations are known to have the same
           stored procedure name then that stored procedure name SHOULD be used.
         examples: ["GetCustomer"]
+      - id: db.operation.parameter
+        type: template[string]
+        stability: development
```
trask, line 127: The only place I kept `db.operation.parameter` for now is in elasticsearch semconv, so marked it as `development` for now, though could easily be convinced to keep it rc.

model/database/registry.yaml
```diff
@@ -122,6 +122,19 @@ groups:
...
+        brief: >
+          A database operation parameter, with `<key>` being the parameter name,
+          and the attribute value being a string representation of the parameter value.
+        note: >
+          If an operation parameter has no name and instead is referenced only by index,
+          then `<key>` SHOULD be the 0-based index.
+
```
lmolkova, line 134: I think it's common for operation (not query) parameters to have name, so this is no longer needed?

```suggestion
```

lmolkova (approved):

maryliag (approved):

Issue #2076, opened 2025-04-09
[k8s] Define semantic conventions for k8s resource quota metrics


ChrsMark:
### Area(s)
area:k8s

### Is your change request related to a problem? Please describe.
As part of the K8s SemConv stability work we need to define k8s resource quota related metrics that are already in use by the [Opentelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.111.0/receiver/k8sclusterreceiver/documentation.md).

Related to https://github.com/open-telemetry/semantic-conventions/issues/1032

### Describe the solution you'd like
The existing metrics in use that we need to define as semantic conventions are the following:

 `k8s.resource_quota.hard_limit`
 `k8s.resource_quota.used`

jinja2:
Resource quotas aren't limited to cpu and memory. It can be set for storage, gpu, object counts (k8s native and custom resources), etc.  Here's an example - 
```
apiVersion: v1
kind: ResourceQuota
metadata:
  creationTimestamp: "2024-04-03T19:16:22Z"
  name: default-resource-quotas
  namespace: istio-ingress
  resourceVersion: "242026"
  uid: <uid>
spec:
  hard:
    count/gateways.networking.istio.io: "20"
    count/targetgroupbindings.elbv2.k8s.aws: "50"
    services.loadbalancers: "10"
    services.nodeports: "30"
status:
  hard:
    count/gateways.networking.istio.io: "20"
    count/targetgroupbindings.elbv2.k8s.aws: "50"
    services.loadbalancers: "10"
    services.nodeports: "30"
  used:
    count/gateways.networking.istio.io: "10"
    count/targetgroupbindings.elbv2.k8s.aws: "17"
    services.loadbalancers: "6"
    services.nodeports: "15"
```

Given the wide range of possible resource types, the existing pattern of using an attribute to indicate the resource type seems more scalable, and flexible to me.

ChrsMark:
Thank's @jinja2 for the note, I guess that might was the idea for it being implemented like this in the Collector.

However, I'm not sure this design can be covered in SemConv. Can the same metric be of different type/unit based on its attributes? For instance, memory quotas will be counted in Bytes while cpu ones in cores, right?

I wonder if scoping each resource under it's own namespace makes more sense here, potentially starting from the most common ones?

rogercoll:
[...]

Issue #2074, opened 2025-04-09
[k8s] Define semantic conventions for k8s container resource (cpu_limits, restarts, ready etc)


ChrsMark:
### Area(s)
area:k8s

### Is your change request related to a problem? Please describe.
As part of the K8s SemConv stability work we need to define k8s container resource related metrics that are already in use by the [Opentelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.111.0/receiver/k8sclusterreceiver/documentation.md).

Related to https://github.com/open-telemetry/semantic-conventions/issues/1032

### Describe the solution you'd like
The existing metrics in use that we need to define as semantic conventions are the following:

 `k8s.container.cpu_request`
 `k8s.container.cpu_limit`
 `k8s.container.memory_request`
 `k8s.container.memory_limit`
 `k8s.container.storage_request`
 `k8s.container.storage_limit`
 `k8s.container.ephemeralstorage_request`
 `k8s.container.ephemeralstorage_limit`
 `k8s.container.restarts`
 `k8s.container.ready`

ChrsMark:
All metrcis are now added. We can close this as completed.

Issue #1489, opened 2024-10-17
[k8s] Define semantic conventions for k8s cpu metrics


ChrsMark:
### Area(s)

area:k8s

### Is your change request related to a problem? Please describe.

As part of the K8s SemConv stability work we need to define k8s cpu metrics that are already in use by the [Opentelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.111.0/receiver/kubeletstatsreceiver/documentation.md).

Related to https://github.com/open-telemetry/semantic-conventions/issues/1032

### Describe the solution you'd like

The existing metrics in use that we need to define as semantic conventions are the following:

 `k8s.node.cpu.usage`:✅ https://github.com/open-telemetry/semantic-conventions/pull/1320
 `k8s.node.cpu.utilization`: Blocked until https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27885#issuecomment-2242780898 is covered
 `k8s.node.cpu.time`: ✅ https://github.com/open-telemetry/semantic-conventions/pull/1320
 `k8s.pod.cpu.usage`: ✅ https://github.com/open-telemetry/semantic-conventions/pull/1320
 `k8s.pod.cpu.utilization`: `Deprecated`
 `k8s.pod.cpu.time`: ✅ https://github.com/open-telemetry/semantic-conventions/pull/1320
 `container.cpu.usage`:  ✅ https://github.com/open-telemetry/semantic-conventions/pull/1128  
 `container.cpu.utilization`: `Deprecated`
 `container.cpu.time` ✅ https://github.com/open-telemetry/semantic-conventions/pull/282
 `k8s.container.cpu_limit_utilization`: ✅  https://github.com/open-telemetry/semantic-conventions/pull/2514
 `k8s.container.cpu_request_utilization`: ✅  https://github.com/open-telemetry/semantic-conventions/pull/2514
 `k8s.container.cpu.node.utilization`
`k8s.pod.cpu.node.utilization`
 `k8s.pod.cpu_limit_utilization`: waiting for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42485
 `k8s.pod.cpu_request_utilization`: waiting for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42485

- [ ] verify if `container.cpu.usage` is the same when collected from k8s, compared to when collected from docker: https://github.com/open-telemetry/semantic-conventions/issues/1489#issuecomment-3244406267

[...]

PR #2113, opened 2025-04-14
Add resourcequota cpu and memory metrics

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/2076

## Changes

This PR introduces the following metrics as part of https://github.com/open-telemetry/semantic-conventions/issues/1032:

```
k8s.resourcequota.cpu.hard_limit
k8s.resourcequota.cpu.hard_request
k8s.resourcequota.cpu.used_limit
k8s.resourcequota.cpu.used_request
k8s.resourcequota.memory.hard_limit
k8s.resourcequota.memory.hard_request
k8s.resourcequota.memory.used_limit
k8s.resourcequota.memory.used_request
```

This addition follows the K8s spec and compute docs:
- https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#resourcequotastatus-v1-core
- https://kubernetes.io/docs/concepts/policy/resource-quotas/#compute-resource-quota

### Extra context

A `ResourceQuota` object will come with a `status` like the following:
```yaml
status:
  hard:
    limits.cpu: "2"
    limits.memory: 2Gi
    requests.cpu: "1"
    requests.memory: 1Gi
  used:
    limits.cpu: 800m
    limits.memory: 800Mi
    requests.cpu: 400m
    requests.memory: 600Mi
```

([ref](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/quota-memory-cpu-namespace/))


rogercoll (commented):
Are those metrics specific per k8s namespace? Should we add the `k8s.namespace.name` attribute?

docs/system/k8s-metrics.md
```diff
@@ -65,6 +65,15 @@ and therefore inherit its attributes, like `k8s.pod.name` and `k8s.pod.uid`.
...
   - [Metric: `k8s.namespace.phase`](#metric-k8snamespacephase)
+- [Resource Quota metrics](#resource-quota-metrics)
+  - [Metric: `k8s.resourcequota.cpu.hard_limit`](#metric-k8sresourcequotacpuhard_limit)
+  - [Metric: `k8s.resourcequota.cpu.hard_request`](#metric-k8sresourcequotacpuhard_request)
+  - [Metric: `k8s.resourcequota.cpu.used_limit`](#metric-k8sresourcequotacpuused_limit)
+  - [Metric: `k8s.resourcequota.cpu.used_request`](#metric-k8sresourcequotacpuused_request)
+  - [Metric: `k8s.resourcequota.memory.hard_limit`](#metric-k8sresourcequotamemoryhard_limit)
+  - [Metric: `k8s.resourcequota.memory.hard_request`](#metric-k8sresourcequotamemoryhard_request)
+  - [Metric: `k8s.resourcequota.memory.used_limit`](#metric-k8sresourcequotamemoryused_limit)
+  - [Metric: `k8s.resourcequota.memory.used_request`](#metric-k8sresourcequotamemoryused_request)
```
rogercoll, line 76: Would it make sense to flip the `used/hard` state part into an attribute?

```
k8s.resourcequota.cpu.limit
k8s.resourcequota.cpu.request
k8s.resourcequota.memory.limit
k8s.resourcequota.memory.request

attributes:
resourcequota.state: "used" | "hard"
```

That would allow us reusing this attribute for future related metrics like:

```
k8s.resourcequota.pods.limit
k8s.resourcequota.pods.request


attributes:
resourcequota.state: "used" | "hard"
```

ChrsMark: Hmm, I think `used` should be renamed to sth like `usage` since it actually represents this. And `hard` should just be skipped since `hard_limit` is essentially just `limit` and same for `request.

In any case we would need to align this with other metrics like `k8s.container.cpu_limit_utilization` (see https://github.com/open-telemetry/semantic-conventions/issues/1489) and  `k8s.container.cpu_request` (see https://github.com/open-telemetry/semantic-conventions/issues/2074)

So probably the metrics will be:

```
# the `hard` section
k8s.resourcequota.cpu.limit
k8s.resourcequota.cpu.request
k8s.resourcequota.memory.limit
k8s.resourcequota.memory.request

# the `used` section
k8s.resourcequota.cpu.limit_usage
k8s.resourcequota.cpu.request_usage
k8s.resourcequota.memory.limit_usage
k8s.resourcequota.memory.request_usage
```

WDYT?

ChrsMark:
Still valid, waiting for https://github.com/open-telemetry/semantic-conventions/issues/2076..

docs/system/k8s-metrics.md
joaopgrassi: I guess this now can be solved: https://github.com/open-telemetry/semantic-conventions/issues/2076#issuecomment-2880652742

ChrsMark:
@jinja2 @dmitryax I have updated this PR based on https://github.com/open-telemetry/semantic-conventions/issues/2076#issuecomment-2880652742. Please take a look when you find the time. Thank's!

model/k8s/metrics.yaml
```diff
@@ -469,3 +469,200 @@ groups:
...
+
+  # k8s.resourcequota.cpu.* metrics
+  - id: metric.k8s.resourcequota.cpu.limit
+    type: metric
+    metric_name: k8s.resourcequota.cpu.limit
+    stability: development
+    brief: |
+      The CPU limits in a specific namespace.
+      Across all pods in a non-terminal state,
+      the sum of CPU limits cannot exceed this value.
```
jinja2, line 481:
```suggestion
    brief: |
      The CPU limits in a specific namespace.
      - If `k8s.resourcequota.state` is `hard`: this value is the maximum allowed CPU limit enforced by the quota.
      - If `k8s.resourcequota.state` is `used`: this value is the current usage of CPU limits requested by all non-terminal pods.
```

jinja2: We'll have to add such clarification in each metric :(

What do you think about adding a note in the attribute's definition and referencing that in the metrics?
Example note in attribute
```
  note: |
    The meaning of the metric value depends on the state:
    - If state is `hard`, the value represents the configured quota limit of the resource in the namespace.
    - If state is `used`, the value represents the current observed total usage of the resource in the namespace.
```
And in the metrics, put a line `See `k8s.resourcequota.state` attribute notes (<link>) for further details.`

model/k8s/metrics.yaml
```diff
@@ -469,3 +469,200 @@ groups:
...
+    attributes:
+      - ref: k8s.resourcequota.state
+        requirement_level: required
+      - ref: k8s.storage.class_name
+        requirement_level: opt_in
+
+  # k8s.resourcequota.persistentvolumeclaims.* metrics
+  - id: metric.k8s.resourcequota.persistentvolumeclaims.request
+    type: metric
+    metric_name: k8s.resourcequota.persistentvolumeclaims.request
```
jinja2, line 597: There is no request/limits for persistentvolumeclaims (the storage.* quotas are for quantity). The field `persistentvolumeclaims` in resourcequota is just the count of pvcs permitted or used. We can either make this part of `object_count` with `k8s.storage.class_name` a conditional_required attr when `k8s.resourcequota.resource_name = pvc` OR we just drop the `.request` and change the unit on this to `1`

ChrsMark: Good catch! Thank you!

model/k8s/metrics.yaml
```diff
@@ -469,3 +469,193 @@ groups:
...
+    note: |
+      More info: https://kubernetes.io/docs/concepts/policy/resource-quotas/#storage-resource-quota
+
+      This metric SHOULD, at a minimum, be reported against a
+      [`k8s.resourcequota`](../resource/k8s.md#resourcequota) resource.
+    attributes:
+      - ref: k8s.resourcequota.state
+        requirement_level: required
+      - ref: k8s.storage.class_name
+        requirement_level: opt_in
```
jinja2, line 585:
```suggestion
        requirement_level: conditionally_required
```

When quotas are set both for generic storage (e.g., requests.storage) and storage class specific resources (e.g., example.storageclass.storage.k8s.io/requests.storage) in the same namespace, omitting the storage class name results in two indistinguishable metrics. If this metric is enabled, then the class_name should be required when a resourcequota is defined for storageclass.

model/k8s/metrics.yaml
```diff
@@ -469,3 +469,193 @@ groups:
...
```
jinja2, line 606:
```suggestion
        requirement_level: conditionally_required
```

model/k8s/metrics.yaml
```diff
@@ -469,3 +469,193 @@ groups:
...
+        requirement_level: required
+
+  # k8s.resourcequota.object_count metric
+  - id: metric.k8s.resourcequota.object_count
+    type: metric
+    metric_name: k8s.resourcequota.object_count
+    stability: development
+    brief: |
+      The object count limits in a specific namespace.
+      See [`k8s.resourcequoate.state`](../registry/attributes/k8s.md) attribute for further details.
```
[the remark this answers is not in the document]
ChrsMark: Thank's!

docs/registry/attributes/k8s.md
```diff
@@ -169,6 +177,21 @@ conflict.
 
 ---
 
+`k8s.resourcequota.state` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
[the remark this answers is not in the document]
ChrsMark: Yeap, that's true. We did this for convenience but if it doesn't work eventually we need to change it.

The alternative would be to just split each metric adding with a `.used` suffixed metrics and a `.hard_limit` suffixed metric. @jinja2 any thoughts on this?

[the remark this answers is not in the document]
jinja2: I agree - the concern about incorrect aggregation and user confusion outweighs the simplicity of our label-based schema.

@ChrsMark To be in line with the [k8s spec](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/resource-quota-v1/#ResourceQuotaStatus), I would suggest `.hard` instead of `.hard_limit`, otherwise your suggestion sgtm. Our metric names keep getting longer 😅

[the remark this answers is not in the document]
ChrsMark: Cool, thank's for the feedback @jinja2 @dashpole! I have now applied the changes, please take another look when you get the chance.

> Our metric names keep getting longer 😅

Yeap 😓 .

dashpole:
I see a mixture of updowncounters and gauges.  Seems like they should all use the same instrument, given they are all usage and limit, but with different units.  I see our container request/limit metrics [are gauges](https://github.com/open-telemetry/semantic-conventions/blob/2bc97890c1ad82232745b4dd74fd3144476b6a5c/model/k8s/metrics.yaml#L511), but IMO updown counters seem more appropriate.

ChrsMark:
> I see a mixture of updowncounters and gauges. Seems like they should all use the same instrument, given they are all usage and limit, but with different units. 

Intuitively, I'd pick `gauge` type for resource (cpu, memory, storage) limits and specially usages, but I'd be fine going with `updowncounter` if we think it applies better.

>  I see our container request/limit metrics [are gauges](https://github.com/open-telemetry/semantic-conventions/blob/2bc97890c1ad82232745b4dd74fd3144476b6a5c/model/k8s/metrics.yaml#L511), but IMO updown counters seem more appropriate.

Same here, we can change it in a separate PR if we have agreement.

model/k8s/metrics.yaml
```diff
@@ -616,3 +616,341 @@ groups:
...
+  # k8s.resourcequota.persistentvolumeclaims.* metrics
+  - id: metric.k8s.resourcequota.persistentvolumeclaims.hard
+    type: metric
+    metric_name: k8s.resourcequota.persistentvolumeclaims.hard
+    stability: development
+    brief: |
+      The total number of PersistentVolumeClaims that can exist in the namespace.
+      The value represents the configured quota limit of the resource in the namespace.
+    instrument: updowncounter
+    unit: "1"
```
dashpole: Should this be {persistentvolumeclaim}?

model/k8s/metrics.yaml
```diff
@@ -616,3 +616,341 @@ groups:
...
+  # k8s.resourcequota.object_count metric
+  - id: metric.k8s.resourcequota.object_count.hard
+    type: metric
+    metric_name: k8s.resourcequota.object_count.hard
+    stability: development
+    brief: |
+      The object count limits in a specific namespace.
+      The value represents the configured quota limit of the resource in the namespace.
+    instrument: updowncounter
+    unit: "1"
```
dashpole: should this be {object}?

model/k8s/metrics.yaml
ChrsMark: Good catch. Fixed.

model/k8s/metrics.yaml
ChrsMark: Good catch. Fixed.

ChrsMark:
@dashpole eventually I changed the instrument types to `updowncounter`s ([ref](https://github.com/open-telemetry/semantic-conventions/pull/2113#issuecomment-2935782781))

@jinja2 please take a look as well when you get the chance :)

model/k8s/registry.yaml
```diff
@@ -524,3 +524,21 @@ groups:
           or
           [K8s ContainerStateTerminated](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#containerstateterminated-v1-core)
         examples: ['ContainerCreating', 'CrashLoopBackOff', 'CreateContainerConfigError', 'ErrImagePull', 'ImagePullBackOff', 'OOMKilled', 'Completed', 'Error', 'ContainerCannotRun']
+      - id: k8s.hugepage.size
+        type: string
+        stability: development
+        brief: >
+          The size (identifier) of the K8s huge page.
+        examples: [ '2Mi' ]
+      - id: k8s.storage.class_name
```
dashpole: The object here is "storageclass".  I would have expected `k8s.storageclass.name`?

ChrsMark: Thank's for catching this.

ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#storageclass-v1-storage-k8s-io

ChrsMark: Fixed.

model/k8s/registry.yaml
```diff
@@ -524,3 +524,23 @@ groups:
...
+        examples: [ '2Mi' ]
+      - id: k8s.storageclass.name
+        type: string
+        stability: development
+        brief: >
+          The name of K8s
+          [StorageClass](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#storageclass-v1-storage-k8s-io)
+          object.
+        examples: [ 'gold.storageclass.storage.k8s.io' ]
+      - id: k8s.resourcequota.resource_name
```
jinja2, line 541: The resource quota spec allows for two syntax for some core object types - [ref](https://kubernetes.io/docs/concepts/policy/resource-quotas/#object-count-quota). I think a note clarifying this difference would be helpful.

Suggestion -
```
The value for this attribute can be either the full `count/<resource>[.<group>]` string (e.g., count/deployments.apps, count/pods), or, for certain core Kubernetes resources, just the resource name (e.g., pods, services, configmaps).

Both forms are supported by Kubernetes for object count quotas. See [Kubernetes Resource Quotas documentation](https://kubernetes.io/docs/concepts/policy/resource-quotas/#object-count-quota) for more details.`
```

Or was your intention here to always add the count prefix?

ChrsMark: The intention for this attribute is to capture both syntax types. Adding your suggestion will better clarify this. Thank's!

model/k8s/metrics.yaml
```diff
@@ -616,3 +616,341 @@ groups:
...
+      The value represents the configured quota limit of the resource in the namespace.
+    instrument: updowncounter
+    unit: "By"
+    entity_associations:
+      - k8s.resourcequota
+    note: |
+      This metric is retrieved from the `hard` field of the
+      [K8s ResourceQuotaStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#resourcequotastatus-v1-core).
+    attributes:
+      - ref: k8s.hugepage.size
```
jinja2, line 752:
```suggestion
      - ref: k8s.hugepages.size
```
Wdyt think about using the plural form consistently? K8s spec uses `hugepages` and our metrics also have the plural form.

ChrsMark: I think we should avoid pluralisation here. `size` is an attribute of a `hugepage` so `hugepages.size` looks a bit weird.
For consistency, I think that the respective metrics should not be pluralised but use `count` instead so as to be aligned with `k8s.resourcequota.object_count.*`.
I suggest we have them as `k8s.resourcequota.hugepage_count.*` and `k8s.resourcequota.persistentvolumeclaim_count.*` accordingly. WDYT?

Tried that at [0eacaa7](https://github.com/open-telemetry/semantic-conventions/pull/2113/commits/0eacaa7f084b92e362d223d6397c042b953fe19d).

model/k8s/metrics.yaml
```diff
@@ -616,3 +616,341 @@ groups:
...
+  # k8s.resourcequota.hugepage_count.* metrics
+  - id: metric.k8s.resourcequota.hugepage_count.request.hard
+    type: metric
+    metric_name: k8s.resourcequota.hugepage_count.request.hard
+    stability: development
+    brief: |
+      The huge page requests in a specific namespace.
+      The value represents the configured quota limit of the resource in the namespace.
+    instrument: updowncounter
+    unit: "By"
```
dashpole: It is confusing to have "count" in the name of a metric that is measured in bytes.  If it is measured in bytes I would prefer `hugepage`

ChrsMark: Right, based on the [docs](https://kubernetes.io/docs/concepts/policy/resource-quotas/#compute-resource-quota) this represents `the number of huge page requests of the specified size cannot exceed this value.`
So the unit should be `{hugepage}`, right?

dashpole: Yes.  If it is a count of hugepages, then i'm more OK with the current naming

ChrsMark:
This one has already 2 approvals from K8s SIG. @open-telemetry/specs-semconv-approvers @open-telemetry/specs-semconv-maintainers feel free to take a look :)

PR #2125, opened 2025-04-16
Define dynamic request attributes for Google GenAI LLM inference spans

### LINKED ISSUE #2053: Dynamic "gcp.gen_ai.request.*" attributes in support of Gen AI SDK instrumentation
### Area(s)

area:gen-ai, area:gcp

### What's missing?

In [PR 3374: Include more request configuration options...](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3374), there is an attempt to flatten and dump the contents of `GenerateContentConfigOrDict` into span attributes. I'd like there to be formally conventions in place to support this.

The need for this is to be able to reflect/support any config property that might reasonably impact the behavior/output.

While it would be good to establish conventions for common request options, it is not likely that we will be able to keep up with all request options or establish conventions for all of them. It would be good to have a way to supply any information not otherwise defined in common conventions.

I'd like to define `gcp.gen_ai.request.*` as a set of dynamic span attributes on `gen_ai` spans. These will contain any request options that are common across Gemini and Vertex AI (e.g. options supplied to the Gen AI SDK that enables consistent configuration).

The naming is related to the naming choices made in:

 - https://github.com/open-telemetry/semantic-conventions/pull/2050 ([Propose an update to the values for `gen_ai.system` for Google](https://github.com/open-telemetry/semantic-conventions/pull/2050))

In other words, the name is subject to change dependent on the outcome of https://github.com/open-telemetry/semantic-conventions/pull/2050.

michaelsafyan:
Fixes #2053

## Changes

Defines a `gcp.gen_ai.operation.config.<key>` dynamic attribute in support of Google Gen AI SDK instrumentation (see: [PR 3374: include more request configuration options...](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3374)).  In support of this change, this also adds a Google-specific variant of the Gen AI docs. We anticipate that this will serve to enable additional documentation and Semantic Conventions for Google-specific elements beyond the dynamic request attributes being documented here. 


model/gcp/registry.yaml
```diff
@@ -195,3 +195,28 @@ groups:
...
+  - id: registry.gcp.gen_ai
+    type: attribute_group
+    display_name: Generative AI Attributes
+    brief: >
+      This documents defines attributes for Google generative AI systems.
+    stability: development
+    attributes:
+      - id: gcp.gen_ai.request
+        stability: development
+        type: template[string[]]
```
lmolkova, line 208: do you want value to be an array of strings or a single string?

I.e. `"foo.bar.baz": "5"` or `"foo.bar.baz": ["5"]`  ?

model/gcp/registry.yaml
```diff
@@ -195,3 +195,28 @@ groups:
...
+        brief: >
+          Request configuration options named in a Google-specific manner.
+        note: >
+          Key names are derived from the data structures in the Google
+          Gen AI SDK ("google.genai.types.*" in https://github.com/googleapis/python-genai).
```
lmolkova: is this specific to python? is there a language-agnostic REST API we can use instead ?

model/gcp/registry.yaml
```diff
@@ -195,3 +195,28 @@ groups:
...
         examples: ['my-host1234.example.com', 'sample-vm.us-west1-b.c.my-project.internal']
+
+  - id: registry.gcp.gen_ai
+    type: attribute_group
+    display_name: Generative AI Attributes
+    brief: >
+      This documents defines attributes for Google generative AI systems.
+    stability: development
+    attributes:
+      - id: gcp.gen_ai.request
```
lmolkova, line 206: We need to add `parameters` or similar to make it descriptive. Also do you care about separating request/response options? 

Throwing in some suggestions: `gcp.gen_ai.operation.parameters | options | metadata | config | properties`

model/gcp/registry.yaml
```diff
@@ -195,3 +195,28 @@ groups:
...
+        stability: development
+        type: template[string[]]
+        brief: >
+          Request configuration options named in a Google-specific manner.
```
lmolkova, line 210: maybe 

```suggestion
          Gen AI operation options, `<key>` being the flattened name of the configuration option passed to the client library.
```
?

model/gcp/registry.yaml
```diff
@@ -195,3 +195,28 @@ groups:
...
+          Gen AI SDK ("google.genai.types.*" in https://github.com/googleapis/python-genai).
+
+          Key names are dependent on the particular operation in question.
+          Each operation has its corresponding "*Config" data structure; for
+          example, the "generate_content" operation has "GenerateContentConfig",
+          while "generate_images" has a corresponding "GenerateImagesConfig".
+
+          Key names correspond to a flattened version of the config data
+          structure. For example {"foo": {"bar": {"baz": 5}}} results in
+          a key called "gcp.gen_ai.request.foo.bar.baz" with value 5.
```
lmolkova, line 222:
```suggestion
          a key called "gcp.gen_ai.request.foo.bar.baz" with value "5".
```

also see my other comment on whether you want an array or a single string

model/gcp/registry.yaml
```diff
@@ -195,3 +195,28 @@ groups:
...
+          Key names are derived from the data structures in the Google
+          Gen AI SDK ("google.genai.types.*" in https://github.com/googleapis/python-genai).
+
+          Key names are dependent on the particular operation in question.
+          Each operation has its corresponding "*Config" data structure; for
+          example, the "generate_content" operation has "GenerateContentConfig",
+          while "generate_images" has a corresponding "GenerateImagesConfig".
+
+          Key names correspond to a flattened version of the config data
+          structure. For example {"foo": {"bar": {"baz": 5}}} results in
```
lmolkova, line 221: would you need to support flattening for arrays?

not a blocker this PR:

Do you feel you want to go down the route of defining flattening for everything inside the instrumentation? 

Assuming complex attributes on spans were allowed, would you prefer to flatten this one or to be a (set of) complex attributes like `gcp.gen_ai.content_config = {"foo": {"bar": {"baz": 5}}}`, `gcp.gen_ai.images_config = {...}`?

In other words, do you envision someone aggregating/filtering/indexing over a specific flattened value like `foo.bar.baz`?

---
Also me being overly eager to prove my point - this is where we try very hard to express complex data with flat attributes and come up with creative (but still limited) ways. 

//cc @open-telemetry/specs-logs-approvers @open-telemetry/specs-semconv-maintainers  @austinlparker

model/gcp/registry.yaml
michaelsafyan: I actually want the value types to be `any`, but unfortunately this doesn't look to be supported.

Filed: https://github.com/open-telemetry/weaver/issues/705

model/gcp/registry.yaml
michaelsafyan: Yes and no.

There is a language-agnostic specification that defines a 1:1 mapping to/from the Vertex AI API and to/from the Gemini API. This language-agnostic schema is used to generate `types.py` in the googleapis/python-genai .

This is alluded to in https://github.com/googleapis/python-genai/blob/main/google/genai/types.py

```
# Code generated by the Google Gen AI SDK generator DO NOT EDIT.
```

Unfortunately, I do not believe that this mapping is published.

Additionally, there is a language-specific element in that the generator is likely to use different formatting for outputting in different languages (e.g. `camelCase` in Java, `lower_with_underscore` in Python).

You can see, for example, how similar data structures are auto-generated in Java here:

https://github.com/googleapis/java-genai/tree/main/src/main/java/com/google/genai/types

Because the source of truth from which it is generated is not public, I think it is easiest to reference the Python version. That also establishes that the key names use `lower_case_with_underscores`, which is more consistent with the key naming that appears in the OTel Semantic Conventions.

model/gcp/registry.yaml
michaelsafyan: Done. Renamed to `gcp.gen_ai.operation.config`. Used "config", because this is the name that the Google Gen AI SDK has chosen to use in all languages for this concept.

model/gcp/registry.yaml
michaelsafyan: Done. Applied suggested change to the `brief` attribute.

model/gcp/registry.yaml
michaelsafyan: Good question.

From an instrumentation standpoint, it would be convenient to be able to put anything into an attribute. Or, at the very least, to be able to put anything into an attribute that is supported by `AnyValue` in OTLP.

From a backend implementation standpoint, I am uncertain which of these two options is better:

   1. Send the data over the wire as a single `AnyValue` with the complex data type.
   2. Perform flattening under-the-hood in the OTel library as an implementation detail.

Here are some advantages to option 1:

  - OTLP representation matches what was written by the instrumentation
  - Is probably more compact over the wire (would need to verify)

Here are some advantages to option 2:

 - Enables process-level filtering of sub-elements of the data structure (like the allowlist used in the Gen AI SDK instrumentation for choosing which pieces of the config to include).

 - Enables downstream filtering, processing, or other transformation of specific fields in the data structure (e.g. it would be easier to configure the OTel Collector to target a piece of the data structure to redact or derive a time series from it).

 - Maintains the contract implied by the attribute specification.

 - Provides for consistent naming for querying across different backend systems.

Even if going with option 1, it may still be advantageous to have some standardized flattening algorithm to unambiguously define how to refer to pieces of the object; if such a standard flattening algorithm were defined, then backends could optionally choose to index the contents of the `AnyValue` using the corresponding flattened nomenclature. Additionally, this nomenclature could provide a way to reference the sub-fields for filtering/processing in the Collector (whether through processing complex objects in their `AnyValue` form with these names serving as references into the datastructure, or with the help of processors that apply the flattening algorithm [or reverse it] to facilitate pipelines that operate on the sub-fields of the object).

model/gcp/registry.yaml
lmolkova: Thanks for the explanation! If I understand correctly, you'd like all languages to use python casing and this way be consistent with each other? 

Mind adding a note on this? E.g. 

>           Key names are derived from the data structures in the Google
          Gen AI SDK ("google.genai.types.*" in https://github.com/googleapis/python-genai) and are language-agnostic.

or anything more descriptive

michaelsafyan:
@lmolkova reaching back to see what the next steps are to make forward progress here... for example, do we need to wait for Weaver to resolve https://github.com/open-telemetry/weaver/issues/705 before we can move forward? And do we need this PR to go in before we can submit/merge https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3374 ? 

It looks like Traceloop has some interest in iterating on the Google Gen AI SDK instrumentation (https://github.com/traceloop/openllmetry/issues/2675) ... so I'm wondering if it might make sense to accelerate the merge of https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3374 (even with this pending) simply to reduce the chance of future merge conflicts ... I know that code tends to rot the longer it is out-of-sync.

model/gen-ai/spans.yaml
```diff
@@ -175,6 +175,22 @@ groups:
...
+    extends: trace.gen_ai.client.common_attributes
+    stability: development
+    brief: Semantic Conventions for model inference spans when calling Google systems.
+    span_kind: client
+    note: |
+      `gen_ai.system` MUST be set to `"gcp.gen_ai"`, `"gcp.vertex_ai"`, or `"gcp.gemini"`.
+    attributes:
+     - ref: gcp.gen_ai.operation.config
+       requirement_level: opt_in
+     - ref: gcp.client.service
```
gyliu513, line 189: How about use `members` as `type` for this? Like https://github.com/open-telemetry/semantic-conventions/blob/main/model/gen-ai/registry.yaml#L255-L282

michaelsafyan: Is this a comment about `gcp.gen_ai.operation.config` or about `gcp.client.service`?

It looks like `members` defines an `enum` type. However `gcp.gen_ai.operation.config` is not an enum. The `gcp.cllient.service` is not defined by this PR (though this PR references it).

Is it valid to change the type where it is referenced?

michaelsafyan: I don't think that should be turned into an enum, because that would create a lot of toil whenever a new GCP API is launched and would fall quickly out of date. Additionally, I expect that most GCP client libraries will continue to be auto generated and therefore no one will think to update the value in OTel SemConv. Manually written GCP client libraries like Gen AI SDK are the exception rather than the rule.

model/gcp/registry.yaml
michaelsafyan: Updated the description to clarify, so marking resolved.

model/gcp/registry.yaml
michaelsafyan: Per conversation in LLM Conventions SIG today, will need to update this to just be a JSON string.

model/gcp/registry.yaml
michaelsafyan: No longer applicable. Marking resolved.

model/gcp/registry.yaml
michaelsafyan: No longer applicable. Marking resolved.

model/gcp/registry.yaml
michaelsafyan: I've updated to a JSON string per the LLM SIG feedback. That being said, I do have some concerns regarding ease of implementation in the instrumentation with this approach; it is slightly hostile to reusing the flattening logic to generate the "gen_ai.request.*" attributes via trivial renaming and will make selectively opting in to a subset of the data slightly more complicated. (At least in this particular case).

michaelsafyan:
I see some yamllint errors:

```
./docs/attributes-registry/gcp.md:114:204 MD055/table-pipe-style Table pipe style [Expected: leading_and_trailing; Actual: leading_only; Missing trailing pipe]
./docs/attributes-registry/gcp.md:114:204 MD056/table-column-count Table column count [Expected: 5; Actual: 3; Too few cells, row will be missing data]
./docs/attributes-registry/gcp.md:115:1 MD055/table-pipe-style Table pipe style [Expected: leading_and_trailing; Actual: trailing_only; Missing leading pipe]
./docs/attributes-registry/gcp.md:115:520 MD056/table-column-count Table column count [Expected: 5; Actual: 3; Too few cells, row will be missing data]
./docs/attributes-registry/gcp.md:118:523 MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]
```

... which I'm not quite sure how to fix. Before spending too much time on that, though, I want to confirm that the updated definition is acceptable.

model/gcp/registry.yaml
```diff
@@ -195,3 +195,42 @@ groups:
...
+    display_name: Generative AI Attributes
+    brief: >
+      This documents defines attributes for Google generative AI systems.
+    stability: development
+    attributes:
+      - id: gcp.gen_ai.operation.config
+        stability: development
+        type: string
+        brief: >
+          Operation-specific request configuration options for Google generative AI operations.
```
lmolkova, line 210:
```suggestion
          Operation-specific request configuration options for Google generative AI operations recorded as JSON string.
```

model/gcp/registry.yaml
```diff
@@ -195,3 +195,42 @@ groups:
...
+
+          The string value should be a valid JSON dictionary. The JSON dictionary should include
+          only an allowlisted subset of the request configuration that does not overlap with
+          configuration that has already been captured elsewhere through standard Semantic Conventions.
```
lmolkova, line 214: would you mind moving it to the note (to keep the brief short)?

model/gcp/registry.yaml
```diff
@@ -195,3 +195,42 @@ groups:
...
+          `com.google.genai.types.*` in <https://github.com/googleapis/java-genai> for examples.
+
+          Because the data structure is encoded in JSON format, the key names should be specified
+          using "camelCase". More generally, the names should align with the expectations of proto3
+          JSON encoding <https://protobuf.dev/programming-guides/proto3/#json>; for example, fields
+          that represent a timestamp in the local language should get mapped to the proto3 JSON
+          encoding of "google.protobuf.Timestamp" (i.e. RFC 3339 date format) per the specification
+          given in <https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp>. 
+
+          Instrumentations should not populate this attribute by default. Rather, instrumentations
```
lmolkova, line 233:
```suggestion
          Instrumentations SHOULD NOT populate this attribute by default. Rather, instrumentations
```

model/gcp/registry.yaml
```diff
@@ -195,3 +195,42 @@ groups:
...
+          should provide an opt-in mechanism through which the config information can be dumped
```
lmolkova, line 234:
```suggestion
          SHOULD provide an opt-in mechanism through which the config information can be dumped
```

we usually use MAY so that instrumentations don't have to make anything configurable, but given it's a recommendation to Google-specific instrumentation, I think it's ok to use SHOULD here.

docs/gen-ai/gcp.md
```diff
@@ -0,0 +1,126 @@
...
+| [`gen_ai.usage.output_tokens`](/docs/attributes-registry/gen-ai.md) | int | The number of tokens used in the GenAI response (completion). | `180` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`server.address`](/docs/attributes-registry/server.md) | string | GenAI server address. [9] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gcp.gen_ai.operation.config`](/docs/attributes-registry/gcp.md) | string | Operation-specific request configuration options for Google generative AI operations.
+The string value should be a valid JSON dictionary. The JSON dictionary should include only an allowlisted subset of the request configuration that does not overlap with configuration that has already been captured elsewhere through standard Semantic Conventions. [10] | `{"audioTimestamp":true}`; `{"safetySettings":[{"threshold":"OFF"}]}`; `{"thinkingConfig":{"includeThoughts":true},"routingConfig":{"autoMode":{"modelRoutingPreference":"BALANCED"}}}` | `Opt-In` | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 53: apparently having a line break in brief breaks table rendering. 

image


In any case, please check out my other comment on making brief brief.

lmolkova (commented):
Left some cosmetic suggestions, LGTM otherwise

.chloggen/gcp_genai_dynamic_req_attr.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: enhancement
+component: [gcp, gen-ai]
```
gyliu513, line 2: Can you update as below? I created a PR want to handle multiple components case at https://github.com/open-telemetry/semantic-conventions/pull/2209
```suggestion
component: gcp, gen-ai
```

lmolkova: `[]` would be more idiomatic way to represent multiple values in yaml, why limit us to concatenated strings?

gyliu513: Yes, but seems we are not following best practise here, like with one value, we are using `component: val1`, for multiple values, we are using `component: [val1, val2]`, we should probably go with `component: [val1]` even if there is only one value.

I have updated https://github.com/open-telemetry/semantic-conventions/pull/2209, it can handle following cases now, we should be good. 
- `component: val1`
- `component: [val1, val2]`
- `component: [val1]`

model/gcp/registry.yaml
michaelsafyan: Done.

model/gcp/registry.yaml
michaelsafyan: Done.

model/gcp/registry.yaml
michaelsafyan: Done.

model/gcp/registry.yaml
michaelsafyan: Done.

model/gcp/registry.yaml
```diff
@@ -195,3 +195,41 @@ groups:
...
+          See "google.genai.types.*" in <https://github.com/googleapis/python-genai> and 
+          `com.google.genai.types.*` in <https://github.com/googleapis/java-genai> for examples.
+
+          The JSON dictionary SHOULD include only an allowlisted subset of the request configuration. The
+          contents also SHOULD NOT overlap with configuration that has already been captured elsewhere through
+          standard Semantic Conventions.
+
+          Because the data structure is encoded in JSON format, the key names should be specified
+          using "camelCase". More generally, the names should align with the expectations of proto3
+          JSON encoding <https://protobuf.dev/programming-guides/proto3/#json>; for example, fields
```
lmolkova, line 227: nit, maybe 

```suggestion
          The names SHOULD align with the expectations of proto3
          JSON encoding <https://protobuf.dev/programming-guides/proto3/#json>; for example, fields
```

for brevity?

docs/gen-ai/gcp.md
```diff
@@ -0,0 +1,142 @@
...
+
+**[9] `gen_ai.response.model`:** If available. The name of the GenAI model that provided the response. If the model is supplied by a vendor, then the value must be the exact name of the model actually used. If the model is a fine-tuned custom model, the value should have a more specific name than the base model that's been fine-tuned.
+
+**[10] `server.address`:** When observed from the client side, and when communicating through an intermediary, `server.address` SHOULD represent the server address behind any intermediaries, for example proxies, if it's available.
+
+**[11] `gcp.gen_ai.operation.config`:** The data structure is dependent on the value of "gen_ai.operation.name" and should correspond with the "*Config" data structure in the Google Gen AI SDK. The Google Gen AI SDK provides these "*Config" data structures in multiple languages, generated from a common source of truth. See "google.genai.types.*" in <https://github.com/googleapis/python-genai> and  `com.google.genai.types.*` in <https://github.com/googleapis/java-genai> for examples.
```
lmolkova, line 91: Probably need to escape * here, the formatting is off:

image

lmolkova (approved):

jsuereth:
This LGTM once comments are addressed.

lmolkova:
reopening in case @aabmass would like to take it over

Rima-ag:
I'm happy to take this over, address the comments and update it to include the changes to the semantic convention since then
@lmolkova, could you please reopen the PR?

lmolkova:
@Rima-ag sure, done!

docs/gen-ai/gcp.md
Rima-ag: Addressed in #3486

model/gcp/registry.yaml
Rima-ag: Fixed in #3486

trask:
Closing in favor of @Rima-ag's open-telemetry/semantic-conventions#3486

PR #2127, opened 2025-04-16
Fix merge queue

trask:


.github/workflows/changelog.yml
```diff
@@ -46,7 +44,7 @@ jobs:
 
       - name: Ensure no changes to the CHANGELOG.md
         run: |
-          if [[ $(git diff --name-only $(git merge-base origin/main $PR_HEAD) $PR_HEAD ./CHANGELOG*.md) ]]
+          if [[ $(git diff --name-only origin/main HEAD ./CHANGELOG*.md) ]]
```
lmolkova, line 47: nice clean up!

lmolkova (approved):

.github/workflows/changelog.yml
trask: I'm not sure why the existing implementation was that way, but tested and this seems to work 🤞

arminru (approved):

PR #2128, opened 2025-04-17
Add some cache metrics for gen_ai server

Cirilla-zmh:
## Changes

Add `gen_ai.server.cache_hit_rate` and  `gen_ai.server.cached_tokens_total` to evaluate and analyze the performance of the model service, thereby guiding the optimization of the model service.


docs/gen-ai/gen-ai-metrics.md
```diff
@@ -546,6 +546,208 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
...
+**[7]:** May be used when specific backend is unknown. May use common attributes prefixed with 'gcp.gen_ai.'.
+
+**[8]:** This refers to the 'aiplatform.googleapis.com' endpoint. May use common attributes prefixed with 'gcp.gen_ai.'.
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `gen_ai.server.cached_tokens_total`
```
lmolkova, line 650: We don't use `_total` in metric names - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#metrics

Please consider the following name  `gen_ai.server.model.token.cached` - this would be more descriptive and  consistent with other metrics

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -546,6 +546,208 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
...
+<!-- semconv metric.gen_ai.server.cached_tokens_total -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `gen_ai.server.cached_tokens_total` | Counter | `{token}` | Count of cached prompt tokens | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 663: it's the number of tokens currently cached, so it has to be `UpDownCounter`

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -546,6 +546,208 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
...
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`gen_ai.operation.name`](/docs/attributes-registry/gen-ai.md) | string | The name of the operation being performed. [1] | `chat`; `generate_content`; `text_completion` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 667: this attribute can't be on this metric, right? number of cached tokens is a model server property, it's not per-operation thing

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -546,6 +546,208 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
...
+| [`gen_ai.system`](/docs/attributes-registry/gen-ai.md) | string | The Generative AI product as identified by the client or server instrumentation. [2] | `openai` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`error.type`](/docs/attributes-registry/error.md) | string | Describes a class of error the operation ended with. [3] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` if the operation ended in an error | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
lmolkova, line 669: it also is not appliable I believe

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -546,6 +546,208 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
...
+| [`gen_ai.request.model`](/docs/attributes-registry/gen-ai.md) | string | The name of the GenAI model a request is being made to. | `gpt-4` | `Conditionally Required` If available. | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`server.port`](/docs/attributes-registry/server.md) | int | GenAI server port. [4] | `80`; `8080`; `443` | `Conditionally Required` If `server.address` is set. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.response.model`](/docs/attributes-registry/gen-ai.md) | string | The name of the model that generated the response. | `gpt-4-0613` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 672: I think having both request and response model does not make sense on a server metric that's also not request/response specific. Let's use `gen_ai.request.model` for now and I'll create an issue to rename it to something applicable to a broader set of scenarios

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -546,6 +546,208 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+### Metric: `gen_ai.server.cache_hit_rate`
```
lmolkova, line 549: Please take a look at the https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#metrics

In OTel we don't report rate metrics, we report total counts and let users derive rate if they want to. 

A more OTel-idiomatic way would be to have a `gen_ai.server.token.usage` counter similar to `gen_ai.client.token.usage` that contains total number of input and output tokens regardless .

We also need a way to record cached/not cached type, we can add another attribute like `gen_ai.input.token.type = cached | other`. 

this way  cache hit can be calculated as `sum(gen_ai.server.token.usage{gen_ai.input.token.type="cached"}) / sum(gen_ai.server.token.usage{gen_ai.token.type="input"})`

docs/gen-ai/gen-ai-metrics.md
Cirilla-zmh: Thank you! But I'm not quite sure whether it should be named "server" or "model". In the gen ai world, do we have other types of servers besides model?

If the MCP Server counts, it doesn't seem like it should have the metric `gen_ai.server.time_to_first_token`.

docs/gen-ai/gen-ai-metrics.md
Cirilla-zmh: > rename it to something applicable to a broader set of scenarios
I fully support this!

In fact, when creating this PR, I went through a very long internal debate — all of our existing attribute sets are designed for request-response patterns, but most of the metrics in this PR are inherent metrics of the model itself and they are nothing to do with requests and responses.

I think we should create a new attribute set specifically for this type of metric, rather than reusing the existing `gen_ai.server` attributes set.

docs/gen-ai/gen-ai-metrics.md
```diff
@@ -546,6 +546,208 @@ If none of these options apply, the `gen_ai.system` SHOULD be set to `_OTHER`.
...
+
+This metric is [recommended][MetricRecommended] to report the cache hit rate of the model service for any model servers which support serving LLMs. It is measured by dividing the number of hited prompt tokens from the cache by the total number of prompt tokens. This is important in evaluating the efficiency of the caching mechanism, as a higher cache hit rate indicates better utilization of cached results and reduced computational overhead during inference.
```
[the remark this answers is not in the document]
Cirilla-zmh: Thanks for your comments! Let's have a discussion in above comments. ;)

Cirilla-zmh:
@lmolkova @achandrasekar 
Thanks for your review!

It seems that we really shouldn’t directly use the existing server-related attributes to record the inherent properties of the model, right? I support this idea. If you all agree, please allow me to make the necessary adjustments and resend this PR.

I believe we should distinguish metrics of `gen_ai.server` and `gen_ai.model`.

docs/gen-ai/gen-ai-metrics.md
Cirilla-zmh: This might be more challenging than expected. For some open-source inference engines, we may only be able to directly obtain pre-calculated rate metrics. For instance, in the case of sglang: [sglang metrics collector](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/metrics/collector.py).

Should we adopt the same approach for metrics exported by inference engines and those exported by model-serving platforms? They may have different observability priorities:

- For inference engines we care more about performance (e.g., cache hit rates for tokens) and resource load & utilization (e.g., some GPU-level metrics).
- For model-serving platforms we care more about service smoothness (e.g., TTFT/TPOT) and cost-efficiency metrics (e.g., token usage rates).

PR #2129, opened 2025-04-17
Add streaming metrics and attributes for sse and rpc

### LINKED ISSUE #6: GenAI (LLM): how to capture streaming
Some questions (and proposals) on capturing streaming LLM completions:

1. Should the GenAI span cover the duration till the last token in case of streaming? 
    - Yes, otherwise how do we capture completion, errors, usage, etc?
2. Do we need an event when the first token comes? Or another span to capture duration-to-first token from the beginning?
    - This might be too verbose/not quite useful
3. Do we need some indication on the span that it represents a streaming call?
4. Do we need new metrics?
    - see https://github.com/open-telemetry/semantic-conventions/pull/1103 for server streaming metrics:
      - Time-to-first-token 
      - Time-to-next-token
      - Number of active streams would also be useful - streaming seems to be quite hard and error prone and users would appreciate knowing they don't close streams, don't read them to the end, etc.
5. What should `gen_ai.client.operation.duration` capture?
    - same as span: time-to-last-token

Cirilla-zmh:
Partly fixes open-telemetry/semantic-conventions-genai#6 

## Changes

See this [comment](https://github.com/open-telemetry/semantic-conventions-genai/issues/6#issuecomment-2771791800) for details.

cc @lyred193


trask:
hi @Cirilla-zmh, I'd suggest that we wait on reviewing this until we kick off the upcoming RPC semconv stability effort: https://github.com/open-telemetry/community/pull/2684

Cirilla-zmh:
> hi @Cirilla-zmh, I'd suggest that we wait on reviewing this until we kick off the upcoming RPC semconv stability effort: [open-telemetry/community#2684](https://github.com/open-telemetry/community/pull/2684)

Thank you for the reminder. Let's proceed with this PR according to your timeline.

thompson-tomo:
Should we consider sse under the messaging namespace?

Cirilla-zmh:
> Should we consider sse under the messaging namespace?

Sorry, I don't think so. SSE is defined and implemented over HTTP and messaging systems that support SSE are just follow this specification.

thompson-tomo:
For messaging it is irrelevant of the transport being used and as such some of the messaging would be using http under the surface.

Cirilla-zmh:
> For messaging it is irrelevant of the transport being used and as such some of the messaging would be using http under the surface.

I'm not sure if I really understood what you meant. 

For me, SSE is a concept within the scope of HTTP and is not closely related to messaging systems. In other words, messaging systems can choose to publish messages using various protocols or methods, and SSE over HTTP is one of those options. Additionally, people often use SSE independently of any messaging system for other purposes as well.

Issue #1275, opened 2024-07-24
Need a new metric process.open_file_descriptor.max or process.open_file_descriptor.limit


xiangtianyu:
### Area(s)

area:jvm, area:os, area:system

### Is your change request related to a problem? Please describe.

Now we have a metric `process.open_file_descriptor.count` https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/process-metrics.md#metric-processopen_file_descriptorcount

since jdk provided both `getMaxFileDescriptorCount` and `getOpenFileDescriptorCount` https://github.com/eclipse-openj9/openj9/blob/9d6010511102076fb18bfcfe2b3609c4fb9c94f6/jcl/src/jdk.management/share/classes/com/ibm/lang/management/UnixOperatingSystemMXBean.java#L31 We may need a new metric.

### Describe the solution you'd like

We may need a new metrics like `process.open_file_descriptor.max` or `process.open_file_descriptor.limit`

xiangtianyu:
Is there any update about it?

lmolkova:
related to https://github.com/open-telemetry/semantic-conventions/issues/1838

trask:
let's solve #1838 first, and then come back to this issue and look at adding the `.limit` metric

PR #2139, opened 2025-04-17
feat: Add `jvm.file_descriptor.count` metric to jvm experimental metrics

### LINKED ISSUE #1838: Consider adding JVM metric for open file descriptors Semantic Conventions
### Propose new conventions

I'd like to add this metric to the [java-contrib](https://github.com/open-telemetry/opentelemetry-java-contrib/) project. 

Image

```yaml
  - bean: java.lang:type=OperatingSystem
    mapping:
      OpenFileDescriptorCount:
        metric: jvm.fd.open
        unit: 1
        desc: open file descriptors
```

Happy to discuss naming conventions on these as well but would like to collect this information!

I believe that this metric would be useful for monitoring resource usage and detecting potential file descriptor leaks and should be recommended metric. 

Apologies if there was already discussion on adding this metric I just did not see it in the current list found here: https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/

schmikei:
Resolves #1838 

## Changes

This adds the UNIX specific JVM metric to the specification and docs for JVM runtime metrics.

This is the first time I'm contributing to this portion of Otel so please let me know if there's any other file I need to update and any guidance maintainers would like to share!


model/jvm/metrics-experimental.yaml
```diff
@@ -67,3 +67,11 @@ groups:
     brief: "Number of buffers in the pool."
     instrument: updowncounter
     unit: "{buffer}"
+
+  - id: metric.jvm.file_descriptor.count
+    type: metric
+    metric_name: jvm.file_descriptor.count
+    stability: development
+    brief: "Number of open file descriptors by the JVM."
```
trask, line 75: can you include a notes section that describes what API(s) we would typically get this data from? I think you can see examples of this in the other JVM metrics

docs/runtime/jvm-metrics.md
```diff
@@ -600,6 +600,22 @@ This metric is obtained from [`BufferPoolMXBean#getCount()`](https://docs.oracle
...
+<!-- semconv metric.jvm.file_descriptor.count -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `jvm.file_descriptor.count` | UpDownCounter | `{buffer}` | Number of open file descriptors by the JVM. | ![Development](https://img.shields.io/badge/-development-blue) |
```
SylvainJuge, line 612:
```suggestion
| `jvm.file_descriptor.count` | UpDownCounter | `{file_descriptor}` | Number of open file descriptors by the JVM. | ![Development](https://img.shields.io/badge/-development-blue) |
```

schmikei: Ah thanks for the catch, apologies for the copy paste miss!

docs/runtime/jvm-metrics.md
```diff
@@ -600,6 +600,24 @@ This metric is obtained from [`BufferPoolMXBean#getCount()`](https://docs.oracle
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+<!-- semconv metric.jvm.file_descriptor.count -->
```
trask, line 603: can you add a heading for this metric and update the table of contents?

model/jvm/metrics-experimental.yaml
```diff
@@ -67,3 +67,14 @@ groups:
...
+    instrument: updowncounter
+    unit: "{file_descriptor}"
+    note: >
+      This metric is sourced from the `UnixOperatingSystemMXBean` which is only available on Unix-like systems.
+      [Reference](https://docs.oracle.com/en/java/javase/17/docs/api/jdk.management/com/sun/management/UnixOperatingSystemMXBean.html#getOpenFileDescriptorCount()).
```
trask, line 80: I think can remove this now that you've added it in the markdown (which seems good since more consistent with the other JVM metrics)
```suggestion
```

model/jvm/metrics-experimental.yaml
```diff
@@ -67,3 +67,14 @@ groups:
     brief: "Number of buffers in the pool."
     instrument: updowncounter
     unit: "{buffer}"
+
+  - id: metric.jvm.file_descriptor.count
+    type: metric
+    metric_name: jvm.file_descriptor.count
+    stability: development
+    brief: "Number of open file descriptors by the JVM."
```
trask, line 75:
```suggestion
    brief: "Number of open file descriptors as reported by the JVM."
```

.chloggen/add-file-descriptor-metric-to-jvm.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: jvm
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `jvm.file_descriptor.count` as a development metric to track the number of open file descriptors by the JVM.
```
trask, line 13:
```suggestion
note: Add `jvm.file_descriptor.count` as an in-development metric to track the number of open file descriptors as reported by the JVM.
```

trask (approved):
thanks!

SylvainJuge (approved):
LGTM, do you think we should also include `jvm.file_descriptor.limit` ?

From experience the open count is enough to monitor any file descriptor leak in applications, but knowing how far we are from the limit (for example when it has been set too low) could be relevant as well without being too much effort as it's already provided by the JVM.

I would suggest to do so in a separate PR though.

trask:
> LGTM, do you think we should also include `jvm.file_descriptor.limit` ?

I think we have tracking issue for this already: #1275

PR #2146, opened 2025-04-18
Create more registries

jsuereth:
Reworks our registry generation so we have a registry we can provide links to for both attributes and entities.


docs/registry/entities/aws-log.md
```diff
@@ -0,0 +1,29 @@
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/entity_entry.md.j2 -->
+
+
+
+# Aws Log
```
lmolkova, line 6: nit: acronym filter

docs/registry/entities/aws-ecs.md
```diff
@@ -0,0 +1,35 @@
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/entity_entry.md.j2 -->
+
+
+
+# Aws Ecs
```
lmolkova, line 6: leaving comment here, but it's a general question:

what would we do for metrics/events/spans in the future - would we group them by the root namespace or render each in a  separate file? 

I think it makes sense to group them (e.g. all HTTP metrics in one doc) -  having one per file feels too granular. But that's what we generate here and it matches current resource docs layout

docs/registry/entities/README.md
```diff
@@ -0,0 +1,256 @@
...
+> The following registry overview is a work in progress.
+>
+> Relationships and signal associations are a work in progress.
+
+Currently, the following namespaces exist:
+<table>
+<thead>
+<tr>
+  <th>Namespace</th>
+  <th>Entity</th>
```
lmolkova, line 25: maybe 

```suggestion
  <th>Entities</th>
```

?

The rendering is a bit weird

image


so it took me  a bit to understand there could be more than one, so maybe plural would be a bit more clear?

templates/registry/markdown/entity_entry.md.j2
```diff
@@ -0,0 +1,16 @@
...
+# {{ ctx.name | title_case }}
+
+**Status:** {{ stability.badge(ctx.stability, ctx.deprecated) }}
+
+**type:** `{{ ctx.name }}`
+
+**Description:** {{ ctx.brief }}
+
+{# TODO: Add identifying vs. descriptive attributes #}
+{{ at.generate(ctx.attributes, "", "../attributes", ctx.lineage.attributes) }}
```
lmolkova, line 15: wdyt about using absolute path from the repo root? I.e. `/docs/registry/attributes"

templates/registry/markdown/weaver.yaml
```diff
@@ -1,10 +1,20 @@
 templates:
-  - pattern: readme.md.j2
+  - pattern: attribute_readme.md.j2
     filter: '.groups | map(select(.type == "attribute_group")) | map(select(.id | startswith("registry"))) | group_by(.id | split(".") | .[1]) | map({id: .[0].id | split(".") | .[1], groups: .})'
     application_mode: single
+    file_name: 'registry/attributes/README.md'
```
lmolkova, line 5: nit: there is some inconsistency around `registry/attribute` vs `registry/attributes` in the current commit

trask (commented):
what do you think about renaming

`/docs/attributes-registry` --> `/docs/registry/attribute/`

in a separate PR first, to make it easier to review the mass mechanical changes separately?

docs/registry/entities/aws-ecs.md
jsuereth: Yeah - I think having one per file DOES make a lot of sense. I can update the entity registry to do that too.

docs/registry/entities/README.md
jsuereth: Yeah - I'm still toying with making this look good. This is a prototype that SHOULD be just enough wrong it give speople a good idea of what it should be :).

Agree it could be called Entities, but if you have other ideas for how to organize the information, I'm all ears.

jsuereth:
> what do you think about renaming
> 
> `/docs/attributes-registry` --> `/docs/registry/attribute/`
> 
> in a separate PR first, to make it easier to review the mass mechanical changes separately?

Of course.  This is a "demo" of what we could do, hence it's still draft.  I'd like to align on what we think "looks right" and then do all the updates.

The Website changes in this PR are also somewhat dubious, but hard to showcase without importing this PR into the opentelemetry.io project to show it off...

templates/registry/markdown/weaver.yaml
jsuereth: Fixed

templates/registry/markdown/entity_entry.md.j2
jsuereth: Fixed, good suggestion!

jsuereth:
> what do you think about renaming
> 
> `/docs/attributes-registry` --> `/docs/registry/attribute/`
> 
> in a separate PR first, to make it easier to review the mass mechanical changes separately?

Opened a PR for this now.

PR #2149, opened 2025-04-19
Remove RC status from deprecated `db.system`

trask:
This was just missed when we deprecated `db.system`.

### DIFF
--- a/model/database/deprecated/registry-deprecated.yaml
+++ b/model/database/deprecated/registry-deprecated.yaml
@@ -436,8 +436,8 @@ groups:
               stability: development
             - id: mariadb
               value: "mariadb"
-              brief: "MariaDB (This value has stability level RELEASE CANDIDATE)"
-              stability: release_candidate
+              brief: "MariaDB"
+              stability: development
             - id: maxdb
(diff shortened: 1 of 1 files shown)


model/database/deprecated/registry-deprecated.yaml
```diff
@@ -489,8 +489,8 @@ groups:
               stability: development
             - id: postgresql
               value: "postgresql"
-              brief: "PostgreSQL (This value has stability level RELEASE CANDIDATE)"
-              stability: release_candidate
+              brief: "PostgreSQL"
+              stability: development
```
lmolkova, line 493: Do you think we should downgrade to development when we deprecate things?
We wouldn't do this for stable ones (to keep code-gen producing the same code), so I wonder if we need to do it for other stability levels.

lmolkova: I don't mind us doing it now for this change, created https://github.com/open-telemetry/semantic-conventions/issues/2156 to track the general question

lmolkova (approved):

maryliag (approved):

Issue #2147, opened 2025-04-18
Review Oracle `db.namespace` attribute value


lmolkova:
It's currently modeled as `ORCL1|PDB1|db_high.adb.oraclecloud.com`

Service name (being the most broad?) should probably go first.

The `db.namespace`  is also associated with connection and is usually retrieved from connection string, and it's not clear if all three things can be specified there. It seems connection string would include service_name, instance id, or both of them, while the database name is an internal name not visible to the client.

trask:
Moved to post-stability since oracle semantic conventions is not in the initial stability wave

joaopgrassi:
@trask do we have a board for "post stability" db work items? If not, should we have one?

lmolkova:
@joaopgrassi it's already there (in the closed DB project)

<img width="362" height="254" alt="Image" src="https://github.com/user-attachments/assets/19ebb152-6f61-484e-8b2a-e01371b799da" />

sudarshan12s:
Sorry, I couldn’t review this issue earlier. I’m sharing additional details about the `db.namespace` attribute below. Let me know your inputs.. 

The `db.namespace` follows the syntax: instanceName | pdbName | serviceName.

One reason for placing the instanceName first—consistent with the review suggestion—is to align with the `db.namespace` format used by Microsoft SQL Server.

The database name and instance name become dynamically available only after successful authentication and hence not included in the connection string.


I also raised an [issue](https://github.com/open-telemetry/semantic-conventions/issues/2999) on changing the spec status for Oracle Database.

lmolkova:
@sudarshan12s sorry if I forgot some context.  Let me go back and (together with my friend chatgpt) try to clarify Oracle terminology:
- Instance is essentially a machine on which PDB is running. Usually one PDB runs on one instance, but not necessarily. Based on your description, you only know which instance you're connected after it happens and don't control it from the client side
- PDB is the database. You can have more than one per instance. you connect to a specific PDB using service name. You don't know until you connect
- service name is the service name that is used to connect to PDB, there could be multiple service names for the same DB

If I'm mostly correct, it sounds like:
- It seems like MS SQL server named instances and Oracle instance name here are not the same. In MS SQL server, instance is part of routing and connection string, it's not something associated with connection, it's necessary to create connection.
- the db.namespace is dynamic - if connection fails, it'd be different than if it succeeds. If I reconnect, I could end up with a different instance. 
- there is no hierarchy between PDB and service

[...]

PR #2151, opened 2025-04-19
Use pipe as the `db.namespace` separator

### LINKED ISSUE #2067: Change the sqlserver `db.namespace` separator from `.` to `|`?
@alanwest @lmolkova what do you think of changing the sqlserver separator from `.` to `|`?

_Originally posted by @trask in https://github.com/open-telemetry/semantic-conventions/pull/1911#discussion_r2033157885_

trask:
Resolves #2067


model/database/registry.yaml
```diff
@@ -26,9 +26,9 @@ groups:
           The name of the database, fully qualified within the server address and port.
         note: >
           If a database system has multiple namespace components, they SHOULD be concatenated
-          (potentially using database system specific conventions) from most general to most
-          specific namespace component, and more specific namespaces SHOULD NOT be captured without
-          the more general namespaces, to ensure that "startswith" queries for the more general namespaces will be valid.
+          from the most general to the most specific namespace component,
+          using `|` as a separator between the components.
+          Any missing components (and their associated separators) SHOULD be omitted.
```
trask, line 31: I intentionally removed this:

> more specific namespaces SHOULD NOT be captured without the more general namespaces, to ensure that "startswith" queries for the more general namespaces will be valid

since it seems better to capture something than nothing.

model/database/spans.yaml
```diff
@@ -951,6 +952,8 @@ groups:
         note: |
           `db.namespace` SHOULD be set to the combination of instance name, database name and
           service name following the `{instance_name}|{database_name}|{service_name}` pattern.
```
lmolkova, line 954: since you're already changing this note, mind changing the order here to 

```suggestion
          service name following the `{service_name}|{database_name}|{instance_name}` pattern.
```

We'll figure the `database_name` separately in https://github.com/open-telemetry/semantic-conventions/issues/2147

lmolkova (approved):

maryliag (approved):

alanwest (approved):

PR #2157, opened 2025-04-21
Add jank events for slow app rendering

breedx-splk:
This adds a new experimental (development) event called `app.jank` which can be used to report jank (slow renders). The event contains attributes that can help to indicate the threshold above which renders happened, a classification type of slow or frozen, and the number of renders that were slow. In the case of Android, the Activity name can also be provided.


docs/app/app.md
```diff
@@ -80,9 +80,53 @@ Use this event to indicate that visual application component has been clicked, t
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
-## Attributes
+## Jank Events
 
-See the [app attributes](/docs/attributes-registry/app.md) registry for all
-application-related attributes that may appear on telemetry items.
+### Event: `app.jank`
```
bidetofevil, line 85: Some general thoughts:

- I feels like it would be better if we can represent discrete events with concrete definitions that can be counted rather than an almost metrics-y summation of the number of events.
- To mitigate folks firing an event every time a frame is slow, causing dozens of events every second, perhaps we can just make a frozen frame one to start instead of letting folks shoot themselves in the foot?
- How are you proposing to handle long periods of frozenness? Cut an event every 700ms?

bidetofevil: I've added specific comments below assuming you still want to go with events that sum up multiple slow frames, but I feel like cutting discrete events with a minimum thresholds would be a better way of quantifying these.

docs/app/app.md
```diff
@@ -80,9 +80,53 @@ Use this event to indicate that visual application component has been clicked, t
...
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`app.jank.count`](/docs/attributes-registry/app.md) | int | A count of the number of frame renders that experienced jank. [1] | `9`; `42` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.jank.threshold_ms`](/docs/attributes-registry/app.md) | int | The minimum rendering threshold for this type of jank, in milliseconds. | `16`; `700`; `1024` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.jank.type`](/docs/attributes-registry/app.md) | string | A categorization of the type of jank. | `slow`; `frozen` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.jank.activity.name`](/docs/attributes-registry/app.md) | string | The name of the Activity that was active during the jank. [2] | `myActivity`; `checkout` | `Opt-In` | ![Development](https://img.shields.io/badge/-development-blue) |
```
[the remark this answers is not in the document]
LikeTheSalad: I'm not sure what's the reason for adding `app.jank.activity.name`. If there's not a strong one, I think we should remove it until someone asks for it. The reason is that in OTel Android we already set the activity/fragment name on every log as `screen.name`, so we'd be duplicating that data, plus this would be the only Android-specific attribute, so maybe reviewers might want to add other platform-specific attributes to it to represent a "screen" on their platforms which might clutter the spec and maybe delay the approvals. So I'd prefer to start as clean as possible, unless we have a good reason to add it.

docs/app/app.md
```diff
@@ -0,0 +1,132 @@
...
+---
+
+`app.jank.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `frozen` | Jank that causes the user to suspect that the application is frozen. [3] | ![Development](https://img.shields.io/badge/-development-blue) |
+| `slow` | Jank is classified as observably slow. [4] | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[3]:** Recommended threshold for slow is 700ms.
```
LikeTheSalad, line 126:
```suggestion
**[3]:** Recommended threshold for frozen is 700ms.
```

docs/app/app.md
LikeTheSalad: I agree, I don't like the idea of firing hundreds of these in a very short time. The way it's portrayed seems to fit better for a metric counter for each type of jank.

docs/app/app.md
```diff
@@ -80,9 +80,53 @@ Use this event to indicate that visual application component has been clicked, t
...
+
+The event name MUST be `app.jank`.
+
+This event indicates that the application has detected substandard UI rendering performance.
+
+Jank happens when the UI is rendered slowly enough for the user to experience some disruption or sluggishness. Jank may be classified with varying degrees, such as `slow` or `frozen`, when rendering times exceed a given threshold.
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`app.jank.count`](/docs/attributes-registry/app.md) | int | A count of the number of frame renders that experienced jank. [1] | `9`; `42` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
```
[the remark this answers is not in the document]
LikeTheSalad: Maybe this could be a way to avoid it getting too noisy as well, by firing the event only after a fixed amount of time has passed. Now, the fact that what the event would eventually send is a count value does make me think that we're talking about a metric.

docs/app/app.md
```diff
@@ -80,9 +80,53 @@ Use this event to indicate that visual application component has been clicked, t
...
+| [`app.jank.type`](/docs/attributes-registry/app.md) | string | A categorization of the type of jank. | `slow`; `frozen` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.jank.activity.name`](/docs/attributes-registry/app.md) | string | The name of the Activity that was active during the jank. [2] | `myActivity`; `checkout` | `Opt-In` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[1] `app.jank.count`:** Depending on platform limitations, the value provided MAY be approximation.
+
+**[2] `app.jank.activity.name`:** This is an Android specific attribute.
+
+---
+
+`app.jank.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
```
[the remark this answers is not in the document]
breedx-splk: I appreciate what you're saying, but the majority of the content in the markdown (everything between matching html/xml start/end comment blocks) is autogenerated by the semconv tooling when running `make table-generation attribute-registry-generation`.

.chloggen/click_events.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: app
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Defines two new click events for the app domain
```
xrmx, line 13:
```suggestion
note: "Defines two new click events for the app domain"
```

docs/app/app.md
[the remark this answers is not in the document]
breedx-splk: Yeah, having the duration seems like a good idea. I think we don't have it right now. 

I'm not convinced that we need client-side aggregation over time...that feels like more of a hassle.

docs/app/app.md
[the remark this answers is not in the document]
breedx-splk: Ok, sounds good, let's remove it for now.

docs/app/app.md
breedx-splk: > How are you proposing to handle long periods of frozenness? Cut an event every 700ms?

You all are thinking about implementation. I'm just trying to spec the event!

I do agree that firing these at a high rate is counterproductive. If we're concerned about implementation details, I think you could open an issue in the Android repo and we could start comparing what we have today with where we want to end up.

As it's written right now, the idea is simply to have a mechanism to report that some slowness happened and how many times, discretized into "slow" and "frozen", driven mainly by the Android descriptions of jank. If we wanted to take a step back and perhaps think about a histogram instead of an event, I'm ok with that....but as a follow-up effort.

This was primarily driven by a desire to convert our existing zero-duration span into an event, but I feel it's getting somewhat off course.

marandaneto: I like the idea of having a % app hang rate similar to % crash-free sessions/users
It's pretty challenging to make those app hangs actionable though (here is the problem and how to fix it).
Usually, those events tell you it's hanging, but not where or how to fix it.
Only profiling makes this actionable, so I don't think this metric is super helpful (well, it's better to have than not to have). Still, it's more distracting than helpful, in my opinion.

Sentry just released [something new](https://blog.sentry.io/get-better-visibility-into-app-hangs-on-apple-devices) around this topic, but I haven't checked the new implementation yet, it used to be just a watchdog with a synthetic stack trace back then, which was most of the time quite off.

My 2 cents would be to make this metric as actionable as possible instead of just having a dashboard with a number, but I'm not sure if this is possible without profiling.

breedx-splk:
I still wanna come back to this.

breedx-splk:
keepalive

model/app/registry.yaml
```diff
@@ -62,3 +62,39 @@ groups:
...
+              brief: 'Jank that causes the user to suspect that the application is frozen.'
+              note: 'Recommended threshold for frozen is 700ms.'
+              stability: development
+        examples: [slow, frozen]
+      - id: app.jank.threshold_ms
+        stability: development
+        type: int
+        brief: The minimum rendering threshold for this type of jank, in milliseconds.
+        examples: [16, 700, 1024]
+      - id: app.jank.activity.name
```
thompson-tomo: I would propose to go with app.activity.name from the start

model/app/events.yaml
```diff
@@ -37,3 +37,22 @@ groups:
         requirement_level: opt_in
       - ref: app.screen.coordinate.y
         requirement_level: opt_in
+  - id: event.app.jank
+    stability: development
+    type: event
+    name: app.jank
```
thompson-tomo, line 43: Could we perhaps go with a non android specific name such as app.rendering.state.changed

model/app/events.yaml
```diff
@@ -37,3 +37,22 @@ groups:
...
+      experience some disruption or sluggishness. Jank may be classified with
+      varying degrees, such as `slow` or `frozen`, when rendering times exceed
+      a given threshold.
+    attributes:
+      - ref: app.jank.count
+        requirement_level: required
+      - ref: app.jank.type
+        requirement_level: recommended
+      - ref: app.jank.threshold_ms
+        requirement_level: recommended
```
thompson-tomo, line 58: I would do this as a body attribute as it doesn't need to be in the registry.

model/app/events.yaml
```diff
@@ -37,3 +37,22 @@ groups:
...
+    name: app.jank
+    brief: >
+      This event indicates that the application has detected substandard UI
+      rendering performance.
+    note: >
+      Jank happens when the UI is rendered slowly enough for the user to
+      experience some disruption or sluggishness. Jank may be classified with
+      varying degrees, such as `slow` or `frozen`, when rendering times exceed
+      a given threshold.
+    attributes:
```
thompson-tomo, line 52: I suspect we should have at least one attribute to indicate what activity caused the jank. I have some thoughts in #2475

model/app/registry.yaml
```diff
@@ -62,3 +62,39 @@ groups:
...
         examples: [ "submit", "attack", "Clear Cart" ]
+      - id: app.jank.count
+        type: int
+        stability: development
+        brief: A count of the number of frame renders that experienced jank.
+        note: >
+          Depending on platform limitations, the value provided
+          MAY be approximation.
+        examples: [ 9, 42 ]
+      - id: app.jank.type
```
thompson-tomo, line 73: Could this be app.rendering.state so that it is generic?

model/app/registry.yaml
```diff
@@ -62,3 +62,39 @@ groups:
...
+              brief: 'Jank is classified as observably slow.'
+              note: 'Recommended threshold for slow is 16ms.'
+              stability: development
+            - id: frozen
+              value: 'frozen'
+              brief: 'Jank that causes the user to suspect that the application is frozen.'
+              note: 'Recommended threshold for frozen is 700ms.'
+              stability: development
+        examples: [slow, frozen]
+      - id: app.jank.threshold_ms
```
thompson-tomo, line 89: Not needed if moved to body attribute.

model/app/registry.yaml
```diff
@@ -62,3 +62,39 @@ groups:
           A widget is an application component, typically an on-screen visual
           GUI element.
         examples: [ "submit", "attack", "Clear Cart" ]
+      - id: app.jank.count
```
thompson-tomo, line 65: Not needed if moved to body attribute or metric

model/app/events.yaml
```diff
@@ -37,3 +37,22 @@ groups:
...
+      - ref: app.jank.count
+        requirement_level: required
+      - ref: app.jank.type
+        requirement_level: recommended
```
thompson-tomo, line 56: Shouldn't it be required?

model/app/events.yaml
```diff
@@ -37,3 +37,22 @@ groups:
...
+      This event indicates that the application has detected substandard UI
+      rendering performance.
+    note: >
+      Jank happens when the UI is rendered slowly enough for the user to
+      experience some disruption or sluggishness. Jank may be classified with
+      varying degrees, such as `slow` or `frozen`, when rendering times exceed
+      a given threshold.
+    attributes:
+      - ref: app.jank.count
+        requirement_level: required
```
thompson-tomo, line 54: Might also be better suited as a body attribute. also am unsure how this could be >1 as events should be be emitted in a similar fashion to logs ie as aggregation not possible.

model/app/registry.yaml
breedx-splk: True, but metrics aren't on the table right now for mobile...and using flat/basic attributes is currently much simpler than defining structured event bodies. There is prior art for defining attributes for events, and these events are pretty simple (eg. not a lot of deep structure or complicated types involved).

model/app/registry.yaml
breedx-splk: Not my preference, because this is attempting to specifically define a type of jank (slow screen rendering performance). It's not really a "rendering state", so I don't think `app.rendering.state` is a great fit.

model/app/registry.yaml
breedx-splk: there's a lot of history here about using simple attributes for simple events. There's a been a LOT of discussion. We want to favor using flat even attributes instead of bodies, where possible.

model/app/events.yaml
breedx-splk: That's fine. Can we agree to do this in a follow-up enhancement PR so that it doesn't stall this one? Thanks!

docs/app/app.md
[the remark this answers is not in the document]
breedx-splk: > Do you want to specify the total time duration being evaluated in which all the slow frames have been observed?  ow wide spread they are greatly impact how the user perceives them (e.g. 10 slow frames over 1 minute is much  different than 1 second).

@bidetofevil I see what you mean about 10 slow renders over 1 second vs 1 minute, this is currently a gap. I'll add something. Thanks!

model/app/events.yaml
breedx-splk: by allowing it to be `recommended` (not required), it allows instrumentation to report jank with a `app.jank.threshold_ms` that doesn't fit cleanly into the "slow" or "frozen" buckets. Feels more flexible that way to me.

docs/app/app.md
breedx-splk: I've added a period attribute now, so that it should hopefully be clearer that these are intended to be counted and reported periodically (like every 10 seconds or 1 second or whatever). Hopefully people don't sent zeros. :)

model/app/registry.yaml
breedx-splk: Good call, I agree. I've renamed this to `app.activity.name`. 👍🏻

model/app/events.yaml
breedx-splk: see my comment above about "state" not really being a good fit.

docs/app/app.md
```diff
@@ -80,9 +82,62 @@ Use this event to indicate that visual application component has been clicked, t
...
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
 ## Attributes
 
-See the [app attributes](/docs/registry/attributes/app.md) registry for all
+See the [app attributes](/docs/attributes-registry/app.md) registry for all
```
trask, line 140: I think if you revert this change it will fix the link check failure

model/app/registry.yaml
LikeTheSalad: I like the overall idea, @breedx-splk. What's very confusing to me is that it looks like a metric, swims like a metric, and quacks like a metric... And yet it's defined as a log event? I noticed that you recently mentioned that you've [added a period attribute](https://github.com/open-telemetry/semantic-conventions/pull/2157/files#r2196171769) to keep the count somewhere in memory and only sending the event after that period of time, and that's something that already happens with metrics too! 😅 - I just don't understand the idea of having to reinvent metrics as log events, it makes me wonder why do we have different signals then? In theory, I guess we can also replace spans by logs, but it doesn't seem like that's the way to go.

I was thinking what would make sense to me for this to be a log event, and as such, I think it would be simpler, like "There was a UI jank at this point in time" with maybe an attribute that shows for how long the UI was stuck and where in the app that happened. It would be helpful for me as an app developer to know that there are UI janks and where, and with the time attr I would be able to correlate that data with the host device info so that I can check how much priority I should give this issue since it's not uncommon to have UI jank in older devices. The problem with that approach is that we could potentially send hundreds of events in a very short period of time (unless we limit the amount of events per second or something like that, where we would only capture the first jank within a second and then ignore the rest during that time, for example), which is why I guess we came up with the "count" attribute. But by introducing a count value and a period of time, we essentially turn this into a metric (unless I missed something important), so it seems like it's just the nature of this event. We've recently added metrics support in OTel Android, so I'm not sure why it would be a problem to go with it as a metric.

It's a tricky event to define for sure, it's probably a good topic for the SIG meeting.

model/app/registry.yaml
LikeTheSalad: I wouldn't add this attribute, I think it might be more confusing than helpful as its items don't depend on constants but instead users can choose what values to set as thresholds, so if they decide to change them at some point (or just switch from an instrumentation impl to another with different defaults), then they would get misleading values across time. Plus, I think this categorization makes more sense to do on the query side, where I can define my threshold at query time, without having to rely on how it was categorized in the past. On top of that, it adds more size to the event itself, making it not too disk space friendly.

If someone needs it in the future, and they explain their use case in a way that we can get a better idea of how to make this attribute useful for them, I think we should add it then. But right now, since we don't know exactly what's needed for this event, I wouldn't add it because I think it's better to add stuff as the need arises, rather than having to make breaking changes and/or removing stuff later.

model/app/events.yaml
thompson-tomo: The problem is currently there is no context to what caused the event so that should be considered. This could be the activity or app.

model/app/registry.yaml
thompson-tomo: If we were to add it then the naming should be generic rather than jank so that it can be used on other os.

model/app/registry.yaml
thompson-tomo: My understanding on the guidance I have seen is that body properties are for attributes which are used as across multiple sihnals

model/app/registry.yaml
thompson-tomo: Agree it certainly appears to me to be a metric to me so I would rather see it expressed in that manner. The event should be recorded as soon as it happens to provide an indication that something happened.

model/app/events.yaml
thompson-tomo: Then we need another name, key thing would be to not be android specific.

model/app/events.yaml
thompson-tomo: Not sure what benefit reporting just the threshold would be

breedx-splk:
Well, this seems to be in a bad state now after rebasing. The builds keep failing and many unrelated files seem touched now...and so I'm inclined to recreate this as a new PR with a smaller history, but since nobody really seems to even want this and there's some strong desire to have high-cardinality metrics now, I think I'll just close it.

model/app/registry.yaml
bidetofevil: Jank is a specific type of UI artifact whereby some amount of frames were not rendered, leading to choppy or frozen UI. It's not state, per se, but more like the occurrence of something bad that is perceivable to users. 

I can see the desire to classify this into different types, but in a way that is what the period and count attributes do. Having a high-level box to put it into would require defining the parameters of said box, and I don't think we know enough how the usage to clearly say what they would be.

I think we should probably skip this for now. And if we want to go ahead and do it, I wouldn't make it more generic, as the whole idea of semantic conventions is to make them specific and meaningful. If there are other UI screen events we want to track, we can define a new event, as the underlying attributes would likely not be the same as what we have for jank.

model/app/registry.yaml
bidetofevil: I mentioned this in @breedx-splk's [PR](https://github.com/open-telemetry/opentelemetry-android/pull/1064) over in the Android repo, but I don't think making this a metric, thereby erasing the "when did it happen?" dimension from the signal is a good idea. That is the most important thing when this data is being used, and no amount of aggregation will give you more information than knowing when it happened within a session and what preceded and followed it.

Count in this case is metadata for the jank event and gives some qualification as to the magnitude of the jank event being seen by the user. Totalling up all the dropped frames but not stating when they were dropped renders this data pretty useless. We are not trying to count all the dropped frames in the entire fleet (or even on a single device) - we are trying to model the fact that users are seeing janky scrolling or unresponsive UI.

model/app/events.yaml
bidetofevil: So jank events are modelled as some number of dropped frames over a certain period in such a way that it is perceivable by a user. It is not that every single dropped frame will create an event - this is not the dropped frame event, after all.

This is why the count and period attributes are important: to describe what kind of jank event was captured.

model/app/registry.yaml
thompson-tomo: Actually metrics support exemplars which preserve that link hence not rendering them useless in fact for me make them more useful then events as you can see how widespread it is and then drill down.

PR #2164, opened 2025-04-22
[chore] Enable PR can add area label

gyliu513:
Fixes #

## Changes

### DIFF
--- a/.github/scripts/prepare-new-pr.sh
+++ b/.github/scripts/prepare-new-pr.sh
@@ -34,6 +34,9 @@ if [ $COUNT -eq 1 ]; then
     CHANGE_TYPE=$(awk -F': ' '/^change_type:/ {print $2}' "$PR_CHANGELOG_PATH/$CHLOG" | xargs)
     echo $CHANGE_TYPE
     gh pr edit "${PR}" --add-label "${CHANGE_TYPE}" || true
+    AREA=$(awk -F': ' '/^component:/ {print $2}' "$PR_CHANGELOG_PATH/$CHLOG" | xargs)
+    echo $AREA
+    gh pr edit "${PR}" --add-label "area:${AREA}" || true
 else
     echo "Found multiple changelog files. Ignoring this change."
 fi


.github/scripts/prepare-new-pr.sh
```diff
@@ -34,6 +34,9 @@ if [ $COUNT -eq 1 ]; then
     CHANGE_TYPE=$(awk -F': ' '/^change_type:/ {print $2}' "$PR_CHANGELOG_PATH/$CHLOG" | xargs)
     echo $CHANGE_TYPE
     gh pr edit "${PR}" --add-label "${CHANGE_TYPE}" || true
```
lmolkova, line 36: I believe you can add [multiple labels at once](https://cli.github.com/manual/gh_pr_edit) at once with something like `gh pr edit "${PR}" --add-label "${CHANGE_TYPE},{$AREA}"`

gyliu513: Done

.github/scripts/prepare-new-pr.sh
```diff
@@ -33,7 +33,9 @@ COUNT="$(echo "$CHLOG" | wc -l)"
 if [ $COUNT -eq 1 ]; then
     CHANGE_TYPE=$(awk -F': ' '/^change_type:/ {print $2}' "$PR_CHANGELOG_PATH/$CHLOG" | xargs)
     echo $CHANGE_TYPE
-    gh pr edit "${PR}" --add-label "${CHANGE_TYPE}" || true
+    AREA=$(awk -F': ' '/^component:/ {print $2}' "$PR_CHANGELOG_PATH/$CHLOG" | xargs)
+    echo $AREA
+    gh pr edit "${PR}" --add-label "area:${AREA} ${CHANGE_TYPE}" || true
```
lmolkova, line 38: I'm sorry I think I sent you on a wrong path. Apparently there is no way to ignore labels that don't exist. I *think* we validate the change type and it's pretty much guaranteed to be correct, but area is a free form text, so it's likely to be wrong.

If one of the labels not found, the command would not populate the other one. 
So have them run in two separate commands would be a more robust way.

Since area is a free form text (and we sometimes list more than one), populating area would not work every once in a while

gyliu513: Updated, thanks @lmolkova . BTW: I think for each PR, the reviewer needs to check and make sure the `component ` field is correct. :-)

lmolkova (approved):

trask (approved):

PR #2168, opened 2025-04-23
[k8s] Introduce semantic conventions for k8s Deployment labels and annotations

### LINKED ISSUE #2134: [k8s] Define semantic conventions for k8s Deployment labels and annotations
### Area(s)

area:k8s

### What's missing?

As part of the K8s SemConv stability work we should think forwards and define k8s Deployment labels and annotations that will be used by the Opentelemetry Collector in the future to avoid multiple versions of label and annotations attributes present in the components. Currently there is no usage in any of the collector components yet.

`k8s.deployment.label.<key>`
`k8s.deployment.annotation.<key>`

odubajDT:
Fixes #2134 

## Changes

Introduction of semantic conventions for k8s Deployment labels and annotations


model/k8s/resources.yaml
```diff
@@ -74,6 +74,9 @@ groups:
     attributes:
       - ref: k8s.deployment.uid
       - ref: k8s.deployment.name
+      - ref: k8s.deployment.label
```
ChrsMark, line 77: I know that's not the case so far but I wonder if we should consider having labels as `opt-in`.

@open-telemetry/semconv-k8s-approvers @open-telemetry/specs-semconv-approvers thoughts?

jinja2: I'm in favor of making `labels` opt-in (assuming we are discussing this change for all k8s resource types). They can be numerous and sometimes mutable, which may introduce high cardinality. Marking them as opt-in like annotations seems appropriate and consistent to me.

jinja2 (approved):

trask (approved):

PR #2171, opened 2025-04-23
[k8s] Introduce semantic conventions for k8s Job labels and annotations

### LINKED ISSUE #2137: [k8s] Define semantic conventions for k8s Job labels and annotations
### Area(s)

area:k8s

### What's missing?

As part of the K8s SemConv stability work we should think forwards and define k8s Job labels and annotations that will be used by the Opentelemetry Collector in the future to avoid multiple versions of label and annotations attributes present in the components. Currently there is no usage in any of the collector components yet.

`k8s.job.label.<key>`
`k8s.job.annotation.<key>`

odubajDT:
Fixes #2137 

## Changes

Introduction of semantic conventions for k8s Job labels and annotations

### DIFF
--- a/model/k8s/registry.yaml
+++ b/model/k8s/registry.yaml
@@ -245,6 +245,22 @@ groups:
         brief: >
           The name of the Job.
         examples: ['opentelemetry']
+      - id: k8s.job.label
+        type: template[string]
+        stability: development
+        brief: >
+          The label key-value pairs placed on the Job.
+        note: |
+          The `<key>` being the label name, the value being the label value, even if the value is empty.
+        examples: ['k8s.job.label.jobtype=ci', 'k8s.job.label.automated=']
+      - id: k8s.job.annotation
+        type: template[string]
+        stability: development
+        brief: >
+          The annotation key-value pairs placed on the Job.
+        note: |
+          The `<key>` being the annotation name, the value being the annotation value, even if the value is empty.
+        examples: [ 'k8s.job.annotation.number=1', 'k8s.job.annotation.data=' ]
       - id: k8s.cronjob.uid
         type: string
         stability: development
--- a/.chloggen/job-labels-annotations.yaml
+++ b/.chloggen/job-labels-annotations.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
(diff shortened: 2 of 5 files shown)


ChrsMark (approved):

model/k8s/registry.yaml
```diff
@@ -217,6 +217,22 @@ groups:
...
           The name of the Job.
         examples: ['opentelemetry']
+      - id: k8s.job.label
+        type: template[string]
+        stability: development
+        brief: >
+          The label key-value pairs placed on the Job.
+        note: |
+          The `<key>` being the label name, the value being the label value, even if the value is empty.
+        examples: ['k8s.job.label.jobtype=ci', 'k8s.job.label.automated=']
```
trask, line 227: @odubajDT can you update the examples in your PRs to remove the key names? we are in the process of doing this everywhere, see #2175

cc @lmolkova

trask (approved):

model/k8s/registry.yaml
odubajDT: Sure, but just to make things clear, in the PR for node annotation the following examples are present:

```
examples: [ '0', '' ]
```

Annotations are key-value pairs and the examples would be confusing to me. Please correct me if I understand it wrong. Thank you!

trask: > Annotations are key-value pairs and the examples would be confusing to me

yeah, I'll discuss with other semconv folks how we could potentially mitigate this confusion before proceeding with this change, we can always make this change here later and doesn't need to hold up your PR

joaopgrassi (approved):

model/k8s/registry.yaml
lmolkova: does this look confusing? 

image




I think the confusion comes from the attribute brief (and the way we define template attributes in yaml)

image

The attribute `k8s.job.label.<key>` is not a key-value pair. `k8s.job.label.*` is a set of attributes that are recorded as a set of flattened key value pairs.
If I try to implement this, I'd be confused on what I put into the value of such attribute - `k8s.job.label.jobtype=ci` ?

odubajDT: should be adapted

trask: @odubajDT sorry for the churn, it looks like we may try to preserve the key-value pairing, see https://github.com/open-telemetry/semantic-conventions/pull/2175/files#r2062701530

it will probably be a bit before we have a final decision and tooling, so I'd suggest go back to your original version that includes the key=value pairs in the examples and we can merge your PRs, and we will adapt your examples in the future based on the outcome

odubajDT: all good, thanks! adapted

PR #2172, opened 2025-04-23
[k8s] Introduce semantic conventions for k8s CronJob labels and annotations

### LINKED ISSUE #2138: [k8s] Define semantic conventions for k8s CronJob labels and annotations
### Area(s)

area:k8s

### What's missing?

As part of the K8s SemConv stability work we should think forwards and define k8s CronJob labels and annotations that will be used by the Opentelemetry Collector in the future to avoid multiple versions of label and annotations attributes present in the components. Currently there is no usage in any of the collector components yet.

`k8s.cronjob.label.<key>`
`k8s.cronjob.annotation.<key>`

odubajDT:
Fixes #2138 

## Changes

Introduction of semantic conventions for k8s CronJob labels and annotations


ChrsMark (approved):
Based on the fact that labels and annotations are essential information of a k8s resource similar to name and uid I think it makes sense to have them defined for the available resources as part of the resource specific namespace.

lmolkova (approved):

PR #2181, opened 2025-04-25
[k8s] make k8s Node and Pod labels optional

### LINKED ISSUE #2079: [k8s] Define semantic conventions for k8s node labels and annotations
### Propose new conventions

Area(s)

area:k8s
Is your change request related to a problem? Please describe.

As part of the K8s SemConv stability work we need to define k8s node resource condition metric that is already in use by the [Opentelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.111.0/receiver/k8sclusterreceiver/documentation.md).

Describe the solution you'd like

The existing metrics in use that we need to define as semantic conventions are the following:

k8s.node.condition

odubajDT:
Fixes #2079 

## Changes

Make k8s Node and Pod labels optional (context [here](https://github.com/open-telemetry/semantic-conventions/pull/2168#discussion_r2055650318))

### DIFF
--- a/.chloggen/node-label-opt.yaml
+++ b/.chloggen/node-label-opt.yaml
@@ -4,7 +4,7 @@
 # your pull request title with [chore] or use the "Skip Changelog" label.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: enhancement
+change_type: breaking
 
 # The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
 component: k8s
--- a/.chloggen/node-label-opt.yaml
+++ b/.chloggen/node-label-opt.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: k8s
+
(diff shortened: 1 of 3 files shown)


ChrsMark (commented):
Shall we just make all `k8s.*.label.*` opt-in with one PR? This will make the decision/review easier I think.

odubajDT:
> Shall we just make all the existing `k8s.*.label.*` attributes opt-in with one PR? This will make the decision/review easier I think.

Good point, will add Pod labels optional here as well. For every other k8s workload it's already adapted in the PR that is introducing the labels/annotations semConv, I guess it's good then?

.chloggen/node-label-opt.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
ChrsMark, line 7: I wonder if this should be `breaking`? @open-telemetry/specs-semconv-approvers thoughts?

ChrsMark (approved):

.chloggen/node-label-opt.yaml
trask: I think it would be considered breaking, since instrumentations will no longer capture something by default when they take in this change. at least I think that's how we would treat it if it were a span attribute, not sure if things are different for resource attributes

dmitryax (approved):

.chloggen/node-label-opt.yaml
odubajDT, line 7:
```suggestion
change_type: breaking
```

lmolkova (approved):

PR #2200, opened 2025-04-30
Make db.operation.name required where it's available, clarify what to do with multiple layers

### LINKED ISSUE #2098: Clarify whether to capture `db.query.text` in Redis (and other similar APIs?)
In Java, Redis client has a structured API, but then emits a query syntax. The Java instrumentation is written at the lower-level query layer and so captures `db.query.text`. However, if the instrumentation was written at the structured API layer, then I think it wouldn't capture `db.query.text`.

This could mean a lack of consistency between Redis instrumentations with respect to `db.query.text`, but there's already inconsistency with respect to `db.query.text` since it should only be captured if the instrumentation implements sanitization (which not all instrumentations will), so maybe it's ok.

lmolkova:
Partially addresses #2098

(all the changes are in the yaml, md only has auto-generated stuff)

There are a few more changes identified in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2693#discussion_r2040353050, will send a separate PR


model/database/spans.yaml
```diff
@@ -361,8 +368,7 @@ groups:
           For batch operations, if the individual operations are known to have the same operation name
           then that operation name SHOULD be used prepended by `BATCH `,
           otherwise `db.operation.name` SHOULD be `BATCH`.
-        requirement_level:
-          conditionally_required: If readily available.
+        requirement_level: required
```
lmolkova, line 371: HBASE always has operation name and does not have query language, no reason for conditionally_required

model/database/spans.yaml
```diff
@@ -511,8 +518,7 @@ groups:
           The name of the [MongoDB command](https://www.mongodb.com/docs/manual/reference/command/) being executed.
         note: ""
         examples: ['findAndModify', 'getMore', 'insertMany', 'bulkWrite']
-        requirement_level:
-          conditionally_required: If readily available.
+        requirement_level: required
```
lmolkova, line 521: mongo also always has an operation name

model/database/spans.yaml
```diff
@@ -796,6 +802,7 @@ groups:
           operations performed by Cosmos DB client.
         examples: ["Microsoft.DocumentDB"]
       - ref: db.operation.name
+        requirement_level: required
```
lmolkova, line 805: and cosmos too

model/database/spans.yaml
```diff
@@ -449,6 +455,7 @@ groups:
           if the individual operations are known to have the same command then that
           command SHOULD be used prepended by `MULTI ` or `PIPELINE `.
           Otherwise `db.operation.name` SHOULD be `MULTI` or `PIPELINE`.
+        requirement_level: required
```
lmolkova, line 458: given that redis does not even concat command and params into a single string, and that even low-level APIs like [`sendCommand`](https://www.javadoc.io/doc/redis.clients/jedis/3.3.0/redis/clients/jedis/Jedis.html#sendCommand-redis.clients.jedis.commands.ProtocolCommand-java.lang.String...-) take individual commands, I think we should always be able to populate it.

@trask you mentioned one of the Redis java client that's instrumented on the  lower level, can  you point me to it and I'll check?

trask: oh, I think I was mistaken.

It looks like the Java instrumentation reverse engineers something that looks like a "query text" but isn't really a query text:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/949b73c8f97471fc33c1a94c5a52e1f129b40ee8/instrumentation/jedis/jedis-3.0/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/Jedis30ClientTest.java#L121-L122

I think in this case probably Java should not be capturing any query text under the stable semconv.

model/database/spans.yaml
```diff
@@ -108,9 +108,16 @@ groups:
...
+      Instrumentations SHOULD, when possible, record database spans that cover the duration of
+      the corresponding API call as if it was observed by the caller (such as client application).
+      For example, if a transient issue happened and was retried within this database call, the corresponding
+      span should cover the duration of the logical operation with all retries.
+
+      When a database client provides both a low-level API to execute arbitrary queries
+      and higher-level convenience APIs for specific operations (e.g., calling stored
+      procedures), instrumentation SHOULD prefer instrumenting the convenience APIs.
+      These often allow setting `db.operation.*`` attributes, which usually are not
+      readily available at the generic query level.
```
trask, line 120:
```suggestion
      When a database client provides higher-level convenience APIs for specific operations
      (e.g., calling stored procedures), which internally generate and execute a query,
      it is RECOMMENDED to instrument the higher-level convenience APIs.
      These often allow setting `db.operation.*` attributes, which usually are not
      readily available at the generic query level.
```

model/database/spans.yaml
lmolkova: Thanks! The command name is still readily available there

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/949b73c8f97471fc33c1a94c5a52e1f129b40ee8/instrumentation/jedis/jedis-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/JedisConnectionInstrumentation.java#L40

[`ProtocolCommand`](https://www.javadoc.io/doc/redis.clients/jedis/3.10.0/index.html) is an interface with a small set of 
 implementations, primarily `Protocol.Command` which is effectively an enum of commands. 

we pass the command over to JedisRequest which just gets command name or converts `byte[]` representing command name to string in 

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/949b73c8f97471fc33c1a94c5a52e1f129b40ee8/instrumentation/jedis/jedis-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/JedisRequest.java#L34-L43

no parsing needed.

trask (approved):

docs/database/database-spans.md
```diff
@@ -83,9 +83,16 @@ For example, for an operation describing SQL query on an anonymous table like `S
...
+For example, if a transient issue happened and was retried within this database call, the corresponding
+span should cover the duration of the logical operation with all retries.
+
+When a database client provides higher-level convenience APIs for specific operations
+(e.g., calling a stored procedure), which internally generate and execute a generic query,
+it is RECOMMENDED to instrument the higher-level convenience APIs.
+These often allow setting `db.operation.*` attributes, which usually are not
+readily available at the generic query level.
 
 **Span name** is covered in the [Name](/docs/database/database-spans.md#name) section.
```
lmolkova, line 97: we might need to update the span name (to account for missing `db.query.summary` on stored procedures), discussion in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2693#discussion_r2067819756

alanwest (approved):

docs/database/database-spans.md
lmolkova: will follow up on this in a separate PR

PR #2206, opened 2025-04-30
Add an option to generate db.query.summary from operation/target, restrict its usage for non-query databases

lmolkova:
Another part of #2098 and related to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2693#discussion_r2040353050

For a code like this

1. Query API

   ```csharp
   var command = new SqlCommand();
   command.CommandType = CommandType.Text;
   command.CommandText = "some_stored_procedure @some_parameter"
   command.Parameters.AddWithValue("@some_parameter", "some value");
   ```

2. Non-query API

   ```csharp
   var command = new SqlCommand();
   command.CommandType = CommandType.StoredProcedure;
   command.CommandText = "some_stored_procedure"
   command.Parameters.AddWithValue("@some_parameter", "some value");
   ```

this PR suggests this (and generalizes for similar cases like `createTable` convenience):

| | Query API (`CommandType.Text`) | Non-query API (`CommandType.StoredProcedure`) |
| --- | --- | --- |
| `db.query.summary` | ✔️ (if parsing is supported) | ✔️ (generate from operation and stored procedure name) |
| `db.query.text` |  ✔️ (if sanitization/parsing  is supported) | ✖️ |
| `db.query.parameter.<key>` | ✔️ (opt-in) | ✔️ (opt-in) |
| `db.operation.name` | ✖️ | ✔️ |
| `db.stored_procedure.name` | ✖️ | ✔️ |

Substantial changes are in yaml files and in "sanitizing..." section in spans.md


docs/database/database-spans.md
```diff
@@ -336,8 +347,16 @@ which SHOULD have low-cardinality and SHOULD NOT contain any dynamic or sensitiv
...
+Instrumentation SHOULD set the query summary if it is readily available through instrumentation hooks or other sources.
+
+Otherwise:
+
+- When instrumenting higher-level APIs that build queries internally - for example,
+those that create a table or execute a stored procedure - instrumentations SHOULD
+generate a `db.query.summary` from available operation(s) and target(s) using the format described in this section.
+
+- When instrumenting APIs that operate at the query level, instrumentations that
+support query parsing SHOULD generate a query summary based on the `db.query.text`.
```
lmolkova, line 359: substantial MD-only change

docs/database/database-spans.md
```diff
@@ -420,6 +439,17 @@ name or target).
...
+   [Microsoft.Data.SqlClient](https://learn.microsoft.com/dotnet/api/microsoft.data.sqlclient.sqlcommand.commandtype):
+
+    ```csharp
+    var command = new SqlCommand();
+    command.CommandType = CommandType.StoredProcedure;
+    command.CommandText = "some_stored_procedure";
+    ```
+
+    the corresponding `db.query.summary` is `EXEC some_stored_procedure`, `db.query.text` is not populated.
+
```
lmolkova, line 452: also here

model/database/common.yaml
```diff
@@ -61,7 +61,7 @@ groups:
       - ref: db.query.text
       - ref: db.query.summary
         requirement_level:
-          recommended: if readily available or if instrumentation supports query summarization.
+          recommended: if available or if the instrumentation supports generating a query summary.
```
trask, line 64: if available seems maybe too generic

maybe?

```suggestion
          recommended: if available through instrumentation hooks or if the instrumentation supports generating a query summary.
```

model/database/registry.yaml
```diff
@@ -106,12 +106,16 @@ groups:
...
-          a class of database queries and is useful as a grouping key,
+          Low cardinality summary of a database query.
+        note: |
+          `db.query.summary` provides static summary of the query.
+          It describes a class of database queries and is useful as a grouping key,
           especially when analyzing telemetry for database calls involving complex queries.
 
+          The summary SHOULD NOT be reported for operations that do not involve
+          a text-based query - for example, when the database does not expose a
+          query language interface.
```
trask, line 117: what if executing a stored procedure is implemented via a native (non text-based) mechanism? or do you think this is a non-issue?

matt-hensley: @trask do you have an example? Likely a non-issue IMO but worth a quick check

trask: I don't, sounds like there may not be any examples

I think [this](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2693#discussion_r2067819756) is what made me think maybe SQLServer driver implemented stored procs in some other way besides regular (text-based) SQL:

> According to [chatgpt](https://chatgpt.com/share/68119770-94e8-800b-a788-d3b3f0d454d9), these two are not equivalent and may have different perf

but I just read the linked result and see it's about how the text-based SQL is structured as opposed to using something other than text-based SQL

lmolkova: I can remove text-based. I think redis is the one that doesn't really format a text query, yet you still have individual elements in it.

> The summary SHOULD NOT be reported for operations that do not involve query language. 

?

lmolkova: the more I think, the more I want to remove this sentence. Elastic/Open/AzureSearch support query text for full search, lucene query syntax, etc, but those run in scope of specific operation/collection and wouldn't report query summary for  them

docs/database/cosmosdb.md
```diff
@@ -63,12 +63,11 @@ Semantic conventions described in this document apply to the call-level spans on
 | [`azure.client.id`](/docs/attributes-registry/azure.md) | string | The unique identifier of the client instance. | `3ba4827d-4422-483f-b59f-85b74211c11d`; `storage-client-1` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`azure.cosmosdb.request.body.size`](/docs/attributes-registry/azure.md) | int | Request payload size in bytes. |  | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`db.operation.batch.size`](/docs/attributes-registry/db.md) | int | The number of queries included in a batch operation. [9] | `2`; `3`; `4` | `Recommended` | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
-| [`db.query.summary`](/docs/attributes-registry/db.md) | string | Low cardinality representation of a database query text. [10] | `SELECT wuser_table`; `INSERT shipping_details SELECT orders`; `get user by id` | `Recommended` [11] | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |
```
lmolkova, line 66: query syntax in cosmos is only used to refine items when querying from specific container, i.e. it only contains stuff we would not put into the summary. Operation and collection name come from the structured part of it and not from the query text.

So removing summary from cosmos, and also making query text not sampling relevant.

trask:
> | | Query API (`CommandType.Text`) | Non-query API (`CommandType.StoredProcedure`) |
> | --- | --- | --- |
> | `db.stored_procedure.name` | ✖️ | ✔️ |

what do you think about

| | Query API (`CommandType.Text`) | Non-query API (`CommandType.StoredProcedure`) |
| --- | --- | --- |
| `db.stored_procedure.name` | ✔️ (if parsing is supported) | ✔️ |

model/database/registry.yaml
trask: makes sense. I think we can get what we wanted (restricting the usage of db.query.summary for nosql databases) just by deciding whether to include it in the vendor-specific convention on a case-by-case basis like you did below for cosmos

docs/database/database-spans.md
```diff
@@ -336,8 +343,16 @@ which SHOULD have low-cardinality and SHOULD NOT contain any dynamic or sensitiv
 > only contain information that has a significant impact on the query, database,
 > or application performance.
 
-Instrumentations that support query parsing SHOULD generate a query summary when
-one is not readily available from other sources.
+Instrumentation SHOULD set the query summary if it is readily available through instrumentation hooks or other sources.
```
trask, line 346: above:

> The `db.query.summary` attribute captures a shortened representation of a query text
which SHOULD have low-cardinality and SHOULD NOT contain any dynamic or sensitive data.

--> something like

> The `db.query.summary` attribute can be used to capture a shortened representation of the query text.
It SHOULD have low-cardinality and SHOULD NOT contain any dynamic or sensitive data.

trask: oh, also the section header

> Generating a summary of the query text

-->

> Generating a summary of the query

model/database/registry.yaml
```diff
@@ -106,10 +106,10 @@ groups:
         type: string
         stability: release_candidate
         brief: >
-          Low cardinality representation of a database query text.
-        note: >
-          `db.query.summary` provides static summary of the query text. It describes
-          a class of database queries and is useful as a grouping key,
+          Low cardinality summary of a database query.
+        note: |
+          `db.query.summary` provides static summary of the query.
```
trask, line 111: (it was there previously, but I think the word possibly raises more questions than it answers)
```suggestion
          `db.query.summary` provides summary of the query.
```

trask: or could probably just remove this line since redundant with brief?

docs/database/database-spans.md
lmolkova: nice! addressed in the https://github.com/open-telemetry/semantic-conventions/pull/2206/commits/4c4d1ccacfb29542ff7c8b100e2145f5ff3344ff

lmolkova:
Discussed this 
| | Query API (`CommandType.Text`) | Non-query API (`CommandType.StoredProcedure`) |
| --- | --- | --- |
| `db.operation.name` | ✖️ | ✔️ |
| `db.stored_procedure.name` | ✔️ (if parsing is supported) | ✔️ |

with @trask offline, decided not to parse stored procedure to keep instrumentation code easier and stay consistent with operation name/collection name (which are not parsed from query).

We can always revisit it later

docs/database/database-spans.md
```diff
@@ -431,6 +449,17 @@ name or target).
...
+- Stored procedure is executed using convenience API such as one available in
+   [Microsoft.Data.SqlClient](https://learn.microsoft.com/dotnet/api/microsoft.data.sqlclient.sqlcommand.commandtype):
+
+    ```csharp
+    var command = new SqlCommand();
+    command.CommandType = CommandType.StoredProcedure;
+    command.CommandText = "some_stored_procedure";
+    ```
+
+    the corresponding `db.query.summary` is `EXEC some_stored_procedure`, `db.query.text` is not populated.
```
trask, line 461:
```suggestion
    the corresponding `db.query.summary` is `EXECUTE some_stored_procedure`, `db.query.text` is not populated.
```

docs/database/cosmosdb.md
trask: > query syntax in cosmos is only used to refine items when querying from specific container

maybe this should be stored somewhere else besides `db.query.text`?

my concern is that when people see `db.query.text` they will think that represents the whole query as opposed to part of the query

(this discussion can definitely be split out into an issue since cosmos isn't in the initial stability wave)

docs/database/database-spans.md
alanwest: I do not have any insights regarding the SQL standard for you, but I do slightly prefer `EXECUTE` over `EXEC`. Other operations are fully spelled out (e.g., `SELECT`, `INSERT`, `UPDATE`), I prefer spelling out `EXECUTE`.

alanwest: Actually, this line would be the following based on [this conversation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2693#discussion_r2069700445)

```suggestion
    the corresponding `db.query.summary` is `some_stored_procedure`, `db.query.text` is not populated.
```

trask: > based on https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2693#discussion_r2069700445

I think that conversation is about `CommandType.Text`, and this example is about `CommandType.StoredProcedure`

I was previously against capturing `db.query.summary` when we don't capture `db.query.text`, but this PR relaxes that and allows `db.query.summary` to be captured for higher-level apis as well (even when there's no `db.query.text`)

trask (approved):

docs/database/cosmosdb.md
lmolkova: filed https://github.com/open-telemetry/semantic-conventions/issues/2213 and added a comment on cosmos yaml so we don't forget to follow up when cosmos goes stable

docs/database/database-spans.md
lmolkova: updated to `EXECUTE` , @alanwest lmk if we're on the same page now

alanwest: > I think that conversation is about `CommandType.Text`, and this example is about `CommandType.StoredProcedure`

Yup, you're right, I overlooked that!

>  lmk if we're on the same page now

We're totally on the same page!

alanwest (approved):

PR #2211, opened 2025-05-01
Make `db.response.returned_rows` opt-in on `release_candidate` spans

trask:


model/database/spans.yaml
```diff
@@ -136,6 +136,9 @@ groups:
       classify as errors.
     span_kind: client
     extends: trace.db.common.full
+    attributes:
+      - ref: db.response.returned_rows
+        requirement_level: opt_in  # experimental attributes must be opt-in stable spans
```
lmolkova, line 141: nit, should we mention in the comment that we want it to be recommended once the attribute is stabilized? We can still dig up the level in the history once we ready, so no strong opinion here

lmolkova (approved):

model/database/spans.yaml
trask: ya, good idea, done

maryliag (approved):

PR #2215, opened 2025-05-01
Update DB migration guide

lmolkova:


trask (approved):

docs/non-normative/db-migration.md
```diff
@@ -58,38 +58,41 @@ to
...
 | `db.cosmosdb.container` &rarr; `db.collection.name` |                                                                                                             |
 | New: `db.query.summary`                             |                                                                                                             |
 | New: `db.operation.batch.size`                      |                                                                                                             |
 | New: `db.response.status_code`                      |                                                                                                             |
 | New: `db.operation.parameter.<key>`                 | Opt-In                                                                                                      |
+| New: `db.query.parameter.<key>`                     | Opt-In                                                                                                      |
+| New: `db.response.returned_rows`                    | Opt-In                                                                                                      |
```
trask, line 70: I think it would be good to mark these as still experimental here somehow

maryliag (approved):

docs/non-normative/db-migration.md
trask: (pushed a commit to resolve this)

alanwest (approved):

docs/non-normative/db-migration.md
```diff
@@ -43,7 +43,7 @@ This section summarizes the changes made to the HTTP semantic conventions
 from
 [v1.24.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/database/README.md).
 to
-[v1.31.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.31.0/docs/database/README.md).
+[v1.33.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.33.0/docs/database/README.md).
```
lmolkova, line 46: Adding a comment so we don't merge it accidentally.

the tag does not exist yet, so the link check fails. The plan is to merge it right before release and avoid breaking other PRs

PR #2218, opened 2025-05-02
Clarify `db.query.summary` for stored procedures

trask:
@matt-hensley dug out https://web.cecs.pdx.edu/~len/sql1999.pdf and found that the ANSI SQL standard is CALL

### DIFF
--- a/.chloggen/2218.yaml
+++ b/.chloggen/2218.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update stored procedure summary example
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [ 2218 ]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/.chloggen/2218.yaml
+++ b/.chloggen/2218.yaml
@@ -10,7 +10,7 @@ change_type: bug_fix
 component: db
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Update stored procedure summary example
+note: Clarify `db.query.summary` for stored procedures
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 # The values here must be integers.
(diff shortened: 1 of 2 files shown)


docs/database/database-spans.md
```diff
@@ -458,7 +458,7 @@ name or target).
     command.CommandText = "some_stored_procedure";
     ```
 
-    the corresponding `db.query.summary` is `EXECUTE some_stored_procedure`, `db.query.text` is not populated.
+    the corresponding `db.query.summary` is `CALL some_stored_procedure`, `db.query.text` is not populated.
```
trask, line 461: @alanwest raised a good question about this being under a SQLServer example where maybe it should be EXECUTE

matt-hensley:
CALL is also used by a number of databases:
https://mariadb.com/kb/en/call/
https://dev.mysql.com/doc/refman/8.4/en/call.html 
https://www.postgresql.org/docs/current/sql-call.html 
https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/CALL.html

MSSQL might be the odd one out as it uses EXEC instead.

docs/database/database-spans.md
alanwest: Here's some links for the purpose of leaving some breadcrumbs

* T-SQL's (SQL Server's SQL dialect) documentation for its [EXECUTE statement](https://learn.microsoft.com/sql/t-sql/language-elements/execute-transact-sql)
* Helpful [StackOverflow post](https://stackoverflow.com/questions/7846418/what-is-the-difference-between-call-and-exec-in-t-sql) on this topic.

alanwest: Something like

```suggestion
    the corresponding `db.query.summary` is `EXECUTE some_stored_procedure`, `db.query.text` is not populated. The `EXECUTE` statement is SQL Server's means to invoke a stored procedure. If the database system uses a different keyword for executing a stored procedure then that keyword SHOULD be used (e.g., `CALL some_stored_procedure`). 
```

trask:
JDBC uses the term "call" in its API: https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#prepareCall-java.lang.String-

docs/database/database-spans.md
```diff
@@ -449,16 +449,30 @@ name or target).
...
-- Stored procedure is executed using convenience API such as one available in
-   [Microsoft.Data.SqlClient](https://learn.microsoft.com/dotnet/api/microsoft.data.sqlclient.sqlcommand.commandtype):
+- Stored procedure is executed using a convenience API such as one available in
+  [JDBC](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#prepareCall-java.lang.String-):
+
+  ```java
+  connection.prepareCall("{call some_stored_procedure}");
+  ```
+
+  the corresponding `db.query.summary` is `call some_stored_procedure`,
```
trask, line 459: I went with lowercase `call` here to match the user input above

.chloggen/2218.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update stored procedure summary example
```
trask, line 13:
```suggestion
note: Clarify `db.query.summary` for stored procedures
```

PR #2231, opened 2025-05-05
(chore) Move off deprecated group type resource to entity group type.

jsuereth:
This just moves `type: resource` to `type: entity` for all groups in semantic conventions.


model/android/resources.yaml
```diff
@@ -1,6 +1,6 @@
 groups:
   - id: resource.android
```
kamphaus, line 2: Should IDs and file names also be renamed for consistency?
To be done in a later PR?
Eg. `model/android/resources.yaml` → `model/android/entities.yaml`
```suggestion
  - id: entity.android
```

jsuereth: Yeah I was thinking of doing this later.  I can do this now if we think we'd like to bump that consistency now too.

Id's are almost invisible to me.

kamphaus: For me it's fine to do it later.
I was just wondering whether it was considered and if anyone else had an opinion on it.

smith (approved):

lmolkova (approved):

trask (approved):

kamphaus (approved):

model/android/resources.yaml
jsuereth: if someone feels strongly, I'll update it now.

smith: I think we should change it here, or at least follow soon with another PR to change it. Unless there's certain tooling that depends on these specific ids. At the very least it's one less thing in the yaml that could confuse people.

dmitryax (approved):

model/android/resources.yaml
jsuereth: I think it needs to also rename resources.yaml to entities.yaml

jsuereth: Going to merge this and rename in follow-up.

Issue #2162, opened 2025-04-22
[Database] Enable trace context propagation between application and database via instrumentation libraries


sincejune:
### Area(s)

area:db

### What's missing?

At work, I've noticed a lack of service correlation between our application and SQL Server, such as passing the current `service name`, `trace ID`, and `span ID` to the database. Implementing this feature would greatly enhance our ability to trace application behaviors.

### Describe the solution you'd like

Introducing an option to enable instrumentation libraries to propagate context information to databases.

Using SQL Server as an example, we can implement a `ContextPropagationLevel` flag with three possible values:
- `disabled`: Disable context propagation, default value.
- `service`: Passing the current `service.name` to the database when executing queries.
- `trace`: Passing the current `service.name`, `trace ID` and `span ID` to the database when executing queries.

For an example of the `trace` level implementation in SQL Server, see the following PR: [open-telemetry/opentelemetry-dotnet-contrib#2709](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2709)

If you agree with this proposal, I'm willing to create an initial specification for SQL Server. Thank you!

Kielek:
Linking with https://github.com/open-telemetry/opentelemetry-specification/issues/2279

Kielek:
@trask, do you know what is the status of the donation from https://github.com/google/sqlcommenter?tab=readme-ov-file#sqlcommenter?

I mean mostly semantic convention/specification, it includes
 - Context propagation
    - W3C Traceparent
    - W3C Tracestate
 - Other fields (mandatory, or can be skipped for the context propagation?)
   - controller
   - route
   - db_driver
   - framework 
 - Should be consider also other context text based propagators if configured on the OTel (B3/B3 multi, etc).?

Should it be treated as a stable convention/experimental/not accepted?

trask:
> [@trask](https://github.com/trask), do you know what is the status of the donation from https://github.com/google/sqlcommenter?tab=readme-ov-file#sqlcommenter?

it hasn't landed in the specification or semantic convention repo, so I wouldn't consider it a part of OpenTelemetry yet, even though the IP was [donated to OpenTelemetry in 2021](https://github.com/open-telemetry/community/issues/783).

I'm also personally hesitant about recommending propagating trace context via SQL comments due to https://github.com/google/sqlcommenter/issues/284

sincejune:
[...]

PR #2236, opened 2025-05-06
[db] Add docs for context info propagation

sincejune:
Relevant to #2162 

## Changes
Added documentations for instrumentations to propagate context information to databases


docs/database/database-spans.md
```diff
@@ -478,6 +479,39 @@ Semantic conventions for individual database systems or specialized instrumentat
 MAY specify a different `db.query.summary` format as long as produced summary remains
 relatively short and its cardinality remains low comparing to the `db.query.text`.
 
+## Propagating context to databases
+
+Instrumentations SHOULD propagate the context information to databases.
+
+| Attribute      | Type   | Description                           | Require level     | Stability                                                      |
+|----------------|--------|---------------------------------------|-------------------|----------------------------------------------------------------|
+| `service.name` | string | Logical name of the service [1]       | `Required`        | ![Development](https://img.shields.io/badge/-development-blue) |
```
pellared, line 488: Why is this needed?

docs/database/database-spans.md
```diff
@@ -478,6 +479,39 @@ Semantic conventions for individual database systems or specialized instrumentat
 MAY specify a different `db.query.summary` format as long as produced summary remains
 relatively short and its cardinality remains low comparing to the `db.query.text`.
 
+## Propagating context to databases
+
+Instrumentations SHOULD propagate the context information to databases.
```
pellared, line 484: The propagators are defined here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#propagators-distribution (and https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration).

I think this PR could be simplified to

> Instrumentations SHOULD propagate the context information to the SQL queries following [sqlcommenter](https://google.github.io/sqlcommenter/spec/).

together with an example

docs/database/database-spans.md
```diff
@@ -478,6 +479,39 @@ Semantic conventions for individual database systems or specialized instrumentat
 MAY specify a different `db.query.summary` format as long as produced summary remains
 relatively short and its cardinality remains low comparing to the `db.query.text`.
 
+## Propagating context to databases
```
pellared, line 482: I think this needs to be marked as an experimental feature (which we call Development)

```suggestion
## Propagating context to databases

**Status**: [Development][DocumentStatus]
```

docs/database/sql-server.md
```diff
@@ -152,6 +153,21 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+Instrumentations SHOULD propagate the context information to databases.
+
+| Attribute      | Type   | Description                           | Require level | Stability                                                      |
+|----------------|--------|---------------------------------------|---------------|----------------------------------------------------------------|
+| `service.name` | string | Logical name of the service [1]       | `Required`    | ![Development](https://img.shields.io/badge/-development-blue) |
+| `traceparent`  | string | The trace context of current span [2] | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[1] `service.name`:** MUST be the same for all instances of horizontally scaled services. If the value was not specified, SDKs MUST fall back to `unknown_service:` concatenated with [process.executable.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/), e.g. `unknown_service:bash`. If `process.executable.name` is not available, the value MUST be set to `unknown_service`.
+
+**[2] `traceparent`:** MUST be in the [text format](https://www.w3.org/TR/trace-context/#traceparent-header).
```
pellared, line 167: As described before. I think this can be removed.

docs/database/sql-server.md
```diff
@@ -152,6 +153,21 @@ and SHOULD be provided **at span creation time** (if provided at all):
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+### Propagating context to SQL Server
```
pellared, line 156:
```suggestion
### Propagating context to SQL Server

**Status**: [Development][DocumentStatus]
```

docs/database/database-spans.md
XSAM: I think this is because of the limitation of actual implementation.

When propagating `traceparent` via sqlcommenter is not available for certain databases due to the performance impact (https://github.com/google/sqlcommenter/issues/284) and propagating `traceparent` via `SET CONTEXT_INFO` is also not available due to the additional network round trip delay, only pushing a `service.name` or (`peer.service`) via sqlcommenter become a solution to achieve limited correlation for knowing which service is making the query request.

Some questions here:
- What is the best name to use here? `service.name` or `peer.service`.
- Should we mark this field as `Optional` or `Recommended`?

docs/database/database-spans.md
sincejune: updated

docs/database/sql-server.md
sincejune: updated

docs/database/database-spans.md
```diff
@@ -478,6 +479,37 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+**Status**: [Development][DocumentStatus]
+
+Instrumentations SHOULD propagate the context information to the SQL queries following [sqlcommenter](https://google.github.io/sqlcommenter/spec/).
+
+| Attribute      | Type   | Description                           | Require level     | Stability                                                      |
```
pellared, line 488: Maybe we can change this table and name it something like "recommended propagators"?

pellared:
@open-telemetry/semconv-db-approvers, PTAL

docs/database/database-spans.md
```diff
@@ -478,6 +479,41 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+- Query with `service.name`:
+
+  ```sql
+  SELECT * FROM songs /* service.name=music-player:play */
+  ```
+
+- Query with `service.name` and `traceparent`
+
+  ```sql
+  SELECT * FROM songs /* service.name=music-player:play, traceparent=00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01 */
```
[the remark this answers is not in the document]
sincejune: URL encoding is required for sqlcommenter value serialization, so this doesn't appear to be an issue: https://google.github.io/sqlcommenter/spec/#key-serialization

docs/database/database-spans.md
sincejune: > Gotcha. In such case `service.name` feels more like a baggage item that you want to propagate. Then it could encoded something like
> 
> ```
> SELECT * FROM songs /* baggage: service.name=music-player:play */
> ```

Updated to use `baggage.service.name` as `baggage: ...` is not allowed in sqlcommenter

docs/database/database-spans.md
sincejune: Added header `Recommended attributes` to make it more intuitive.

docs/database/database-spans.md
```diff
@@ -478,6 +480,39 @@ Semantic conventions for individual database systems or specialized instrumentat
 MAY specify a different `db.query.summary` format as long as produced summary remains
 relatively short and its cardinality remains low comparing to the `db.query.text`.
 
+## Propagating context to databases
+
+**Status**: [Development][DocumentStatus]
+
+Instrumentations SHOULD propagate the context information to the SQL queries following [sqlcommenter](https://google.github.io/sqlcommenter/spec/).
+
+### Recommended attributes
```
pellared, line 489:
```suggestion
### Recommended sqlcommenter attributes
```

Just so that people are not confusing it with OTel attributes.

docs/database/database-spans.md
```diff
@@ -478,6 +480,39 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+| Attribute              | Type   | Description                           | Require level     | Stability                                                      |
+|------------------------|--------|---------------------------------------|-------------------|----------------------------------------------------------------|
+| `baggage.service.name` | string | Logical name of the service [1]       | `Required`        | ![Development](https://img.shields.io/badge/-development-blue) |
```
pellared, line 493: If `traceparent` is present I do not think that service.name is needed. Isn't it the best way to have one or another?

Do we even need to define "require level"? Maybe we should just describe the meaning of each sqlcommenter attribute?

docs/database/database-spans.md
```diff
@@ -478,6 +480,39 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+| `traceparent`          | string | The trace context of current span [2] | `Recommended` [3] | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[1] `baggage.service.name`:** MUST be the same for all instances of horizontally scaled services. If the value was not specified, SDKs MUST fall back to `unknown_service:` concatenated with [process.executable.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/), e.g. `unknown_service:bash`. If `process.executable.name` is not available, the value MUST be set to `unknown_service`.
+
+**[2] `traceparent`:** MUST be in the [text format](https://www.w3.org/TR/trace-context/#traceparent-header).
+
+**[3] `traceparent`:** `tracparent` have extremely high-cardinality. It's RECOMMENDED to propagate this info if the high-cardinality is safe for the behind databases
```
pellared, line 500: trying to figure out some better wording...

```suggestion
**[3] `traceparent`:** `traceparent` has extremely high-cardinality. It's RECOMMENDED to propagate this info if the high-cardinality is safe for the database observabilty engine.
```

docs/database/database-spans.md
```diff
@@ -478,6 +480,39 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+### Recommended attributes
+
+| Attribute              | Type   | Description                           | Require level     | Stability                                                      |
+|------------------------|--------|---------------------------------------|-------------------|----------------------------------------------------------------|
+| `baggage.service.name` | string | Logical name of the service [1]       | `Required`        | ![Development](https://img.shields.io/badge/-development-blue) |
+| `traceparent`          | string | The trace context of current span [2] | `Recommended` [3] | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[1] `baggage.service.name`:** MUST be the same for all instances of horizontally scaled services. If the value was not specified, SDKs MUST fall back to `unknown_service:` concatenated with [process.executable.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/), e.g. `unknown_service:bash`. If `process.executable.name` is not available, the value MUST be set to `unknown_service`.
```
pellared, line 496: Maybe we should somehow refer to https://www.w3.org/TR/baggage?

docs/database/database-spans.md
sincejune: > If `traceparent` is present I do not think that service.name is needed. Isn't it the best way to have one or another?

Updated, and I keep the `Require level` to make it as an annotation as well. How do you think about it?

sincejune: > If `traceparent` is present I do not think that service.name is needed. Isn't it the best way to have one or another?

Updated, and I keep the `Require level` to make it as an annotation as well. What do you think about it?

docs/database/sql-server.md
```diff
@@ -152,6 +154,29 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+
+**[1] `baggage.service.name`:** MUST be the same for all instances of horizontally scaled services. If the value was not specified, SDKs MUST fall back to `unknown_service:` concatenated with [process.executable.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/), e.g. `unknown_service:bash`. If `process.executable.name` is not available, the value MUST be set to `unknown_service`.
+
+[Baggage](https://www.w3.org/TR/baggage/) means you can pass data across services and processes, making it available to add to traces, metrics, or logs in those services.
+
+**[2] `baggage.service.name`:** SHOULD be set if `traceparent` cannot be injected due to performance issues.
+
+**[3] `traceparent`:** MUST be in the [text format](https://www.w3.org/TR/trace-context/#traceparent-header).
+
+Instrumentations SHOULD make use of [SET CONTEXT_INFO](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-context-info-transact-sql?view=sql-server-ver16) to add high-cardinality information(e.g. `traceparent`) as adding sql comments to queries changes their unique identifier in SQL Server.
```
zacharycmontoya, line 178: We should add to the guidance here that instrumentations will need to copy over the transaction (if present) to the new `set context_info` command. When we didn't do this, we ran into customer issues when there was an active transaction

docs/database/database-spans.md
```diff
@@ -478,6 +480,43 @@ Semantic conventions for individual database systems or specialized instrumentat
 MAY specify a different `db.query.summary` format as long as produced summary remains
 relatively short and its cardinality remains low comparing to the `db.query.text`.
 
+## Propagating context to databases
+
+**Status**: [Development][DocumentStatus]
+
+Instrumentations SHOULD propagate the context information to the SQL queries following [sqlcommenter](https://google.github.io/sqlcommenter/spec/).
```
zacharycmontoya, line 487: I wanted to share a learning from my org. We've implemented a sqlcommenter-like approach for our instrumentations and we found it more beneficial to _prepend_ the sql comment rather than _append_ to the query text. This runs counter to the sqlcommenter spec, but we found that most databases truncate the query in views like pg_stat_activity, so prepending the attributes offered a better guarantee that they wouldn't be truncated

docs/database/database-spans.md
```diff
@@ -478,6 +480,43 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+- Query with `baggage.service.name`:
+
+  ```sql
+  SELECT * FROM songs /* baggage.service.name=music-player%3Aplay */
+  ```
+
+- Query with `baggage.service.name` and `traceparent`
+
+  ```sql
+  SELECT * FROM songs /* baggage.service.name=music-player%3Aplay, traceparent=00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01 */
```
zacharycmontoya, line 517: Per the sqlcomment spec, the values should be wrapped with a `'` character, so I think this suggestion is a more accurate output (this applies to the previous example as well):

```suggestion
  SELECT * FROM songs /* baggage.service.name='music-player%3Aplay', traceparent='00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01' */
```

docs/database/database-spans.md
zacharycmontoya: I agree with the decision to add the service name as a low-cardinality alternative to trace context, but I'm not sure if we should be referencing the baggage concept or name for this attribute name. Typically the service name for a trace/log/metric isn't typically stored in baggage but instead as a resource.

@pellared what do you think of going back to `service.name` or something slightly different like `resource.service.name`?

XSAM: > What do you think about making it conditionally required, where there is not information about span?

@Kielek, it makes sense to me.

---

About baggage, I know there are no other built-in propagators to use, but I am not sure it is the best place to hold additional attributes.

> Updated to use baggage.service.name as baggage: ... is not allowed in sqlcommenter

 @sincejune If we use baggage without additional processing, then it would look like this for `peer.service=bar`

```
/*baggage='peer.service%3Dbar'*/
```

Result from otelsql, https://github.com/XSAM/otelsql/blob/312753965cc288af499887517f050e17460d420a/commenter_test.go#L72

docs/database/sql-server.md
XSAM: Could you elaborate more about what `copy over the transaction` is actually doing? Could we just run `set context_info` before every query?

docs/database/database-spans.md
sincejune: @zacharycmontoya Thank you for pointing this out!

I think it makes sense to prepend the sql comment instead. I'll update this PR accordingly if there are no objections. cc @open-telemetry/semconv-db-approvers @XSAM

docs/database/database-spans.md
pellared: > but I'm not sure if we should be referencing the baggage concept or name for this attribute name. 

We somehow need to propagate this information. I was thinking of reusing some existing propagator, instead of creating a new one. The alternative would be probably creating a new "service name propagator".

However, it feels that a baggage propagator can be used for this. From https://www.w3.org/TR/baggage/#example-use-case:

> For example, if you need to annotate logs with some request-specific information, you could propagate a property using the baggage header.

I just noticed that here is a nice doc for context propagation for AWS: https://github.com/open-telemetry/semantic-conventions/blob/45e91aedff7d347955ba09269b871f8fd9049ce3/docs/non-normative/compatibility/aws.md#context-propagation

> @pellared what do you think of going back to `service.name` or something slightly different like `resource.service.name`?

:+1: `service.name`  as I originally proposed. Especially that it is already a "reserved" key: https://github.com/open-telemetry/semantic-conventions/blob/45e91aedff7d347955ba09269b871f8fd9049ce3/docs/resource/README.md?plain=1#L58

> @pellared, do you know how to express an attribute in baggage in the semantic convention doc?

I think that we just need to write it down as I do not think we have any semantic conventions for baggage key-values so far.

CC @reyang, @dyladan, @yurishkuro, @arminru

docs/database/database-spans.md
XSAM: Using prepend conflicts the spec of sql commenter.

@zacharycmontoya Could you elaborate more about this? How serious and often the truncate will happen to certain databases if we use append.

docs/database/database-spans.md
pellared: I am not sure if we need even the require level. I think these are just different propagation options :wink: But we can keep it for now as I think there are other more important comments to be resolved/addressed.

docs/database/sql-server.md
zacharycmontoya: Here's the LOC I'm referencing for our C# client : https://github.com/DataDog/dd-trace-dotnet/pull/6233/files#diff-29ead36b0177265c725705ce75f5a441417f4c22e99f7080bc8ef0f486b723a7R145-R146

When creating a new `set context_info` command that we run before the original query, we simply copy the `Transaction` from the previous `DbCommand` object to the new one. Otherwise, we encountered a `InvalidOperationException`

docs/database/database-spans.md
trask, line 487: I'm concerned about pushing this as a standard in light of https://github.com/google/sqlcommenter/issues/284

nenadnoveljic: Postgres [limits query text to 1024 characters by default](https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW), but you can increase it.

sincejune: The plan is to use `sqlcommenter` only with attributes that do not change frequently. In the SQL Server documentation, I have added the section below. @trask Does this plan seem good to you?

> Instrumentations SHOULD make use of [SET CONTEXT_INFO](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-context-info-transact-sql?view=sql-server-ver16) to add high-cardinality information(e.g. `traceparent`) as adding sql comments to queries changes their unique identifier in SQL Server.

docs/database/database-spans.md
nenadnoveljic: > If traceparent is present I do not think that service.name is needed. 

Even if `traceparent` is present, `service.name` is still useful. Exact correlation using `trace_id` requires both a database sample and the full trace for the same request, but because of sampling, many traces and DB executions are missed. Since the chance of both being captured is low (`p_trace * p_db_sample`), including `service.name` helps support approximate correlation when exact matching isn’t possible.

docs/database/database-spans.md
nenadnoveljic: > What is the best name to use here? service.name or peer.service.
I slightly prefer peer.service.

`service.name`. `peer.service` applies to the remote (downstream service): [link](https://opentelemetry.io/docs/specs/semconv/attributes-registry/peer/)

docs/database/sql-server.md
nenadnoveljic: `SET context_info` must run on the same physical connection as the SQL statement it’s meant to instrument.

trask (changes requested):
I've added this to the Semantic Convention and Specification meeting agendas for next week.

I'd like to get some clarity from the OpenTelemetry community whether this is something we should pursue given the performance issues and apparent abandonment of sqlcommenter.

Related: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13714#pullrequestreview-2838103467

docs/database/database-spans.md
sincejune: Updated to use `peer.service`.

sincejune: @pellared Do you recommend using the format `/*baggage='service.name%3Dbar'*/` instead of `/*baggage.service.name='bar'*/`? I'm not sure if there is an existing propagator we can use. Just wanted to double-check.

docs/database/database-spans.md
sincejune: Another idea is to make it flexible by allowing the implementation of propagation for a specific database to choose between `append` or `prepend`.

sincejune: Modify this PR to incorporate the idea.

sincejune: Modified this PR to incorporate the idea.

sincejune: @lmolkova The `traceparent` format is more widely used, so I prefer adopting that. What are your thoughts?

sincejune:
> > Added documentations for instrumentations to propagate context information to databases
> 
> Should we clarify that this article is specific to MS SQL Server, since `context_info` is only available there? Other databases like Oracle and Postgres can use fields like [v$session.action](https://github.com/DataDog/dd-trace-java/pull/8090) or [application_name](https://github.com/DataDog/dd-trace-java/pull/7940) to propagate `tracecontext`.

@nenadnoveljic I only added the `context_info` section to the SqlServer's documentation, so I believe we should be fine with that. What do you think?

docs/database/sql-server.md
sincejune: Updated

docs/database/database-spans.md
pellared: I suggest exploring `/*baggage='service.name%3Dbar'*` and use of the baggage propagator.

sincejune: @pellared Updated the examples accordingly.

XSAM:
I have refactored the PR change to the following structure. @trask @nenadnoveljic @zacharycmontoya PTAL

- For every database
  - Put `service.name` on sql comments.
- For sqlserver
  - Put `traceparent` value via `SET CONTEXT_INFO`

---

Failed CI is irrelevant to this PR. https://github.com/open-telemetry/semantic-conventions/actions/runs/15172537474/job/42665853199?pr=2236

docs/database/database-spans.md
```diff
@@ -486,7 +486,7 @@ relatively short and its cardinality remains low comparing to the `db.query.text
 
 ### SQL commenter
 
-Instrumentations SHOULD propagate the context information to the SQL queries following [SQL commenter](https://google.github.io/sqlcommenter/spec/). The instrumentation implementation MAY choose to either `append` the comment to the end of the query or `prepend` the comment at the beginning of the query, depending on the specific database system's requirements or preferences.
+Instrumentations SHOULD propagate the context information to the SQL queries following [SQL commenter](https://google.github.io/sqlcommenter/spec/). The instrumentation implementation MAY choose to either **append** the comment to the end of the query or **prepend** the comment at the beginning of the query, depending on the specific database system's requirements or preferences.
```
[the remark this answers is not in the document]
XSAM: I think we will need a following discussion of https://github.com/open-telemetry/semantic-conventions/pull/2236#discussion_r2078566667.

I did a small experiment on postgres `17.5` by running two long queries with different comments style more than 1024 bytes.

For instance:
- For **prepend** comment, I ran `/*foo='content longer than 1024 bytes...'*/ SELECT state,query FROM pg_stat_activity;`
- For **append** comment, I ran `SELECT state,query FROM pg_stat_activity /*foo='content longer than 1024 bytes...'*/;`

The result I got from `pg_stat_activity` shows both queries get truncated.

- For **prepend** comment, I got `/*foo='content longer than 1024 bytes...`
  - Only the truncated comment is left.
- For **append** comment, I got `SELECT state,query FROM pg_stat_activity /*foo='content longer than 1024 bytes...`
  - Query remains untouched, as it is very short, but comment is also truncated.

So, postgres seems truncate anything more than 1024 bytes on `pg_stat_activity` whatever it is the query statement or comments. That means no matter the comments, a query that is more 1024 bytes is going to be truncated.

## Comparing these two approaches

### prepend

Pros:
- Always keep the comments. Propagated info won't be lost.

Cons:
- When truncating happens, the query statement get truncated.
  - Users won't see a complete query, and the correlation based on query text may not work.
- It breaks [pg_hint_plan](http://pg-hint-plan.readthedocs.io/en/latest/description.html#basic-usage) functionality https://github.com/open-telemetry/semantic-conventions/pull/2236#discussion_r2084548720. So the instrumentation needs to actively detect a hint. (though I was not able to reproduce the issue, the hint plans I got are always the same)

### append

Pros:
- Original query is kept (assuming the query itself won't be truncated)

Cons:
- Comments might be truncated. Propagated info is lost.

---

In short, assuming **prepend** can break `pg_hint_plan` while **append** won't. **append** seems better for easy implementation. About the truncate part, we could use `ALTER SYSTEM SET track_activity_query_size = 16384;` to increase the limit (though it needs a restart to be effective)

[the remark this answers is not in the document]
XSAM: @lmolkova PTAL, is this PR on a good shape?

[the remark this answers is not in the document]
lmolkova: I do not agree with this statement:

> Instrumentations SHOULD propagate the context information to the SQL queries following [SQL commenter](https://google.github.io/sqlcommenter/spec/).

1. recommending instrumentations to modify SQL queries by default is dangerous and questionable
2. it ruins prepared statements caches for most database systems

We can document other context propagation mechanisms (like there is something for MS SQL and postgres) for these specific databases.
It'd be ok-ish with having it opt-in for other instrumentations.

docs/database/database-spans.md
```diff
@@ -478,6 +480,28 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+**Status**: [Development][DocumentStatus]
+
+### SQL commenter
+
+Instrumentations SHOULD propagate the context information to the SQL queries following [SQL commenter](https://google.github.io/sqlcommenter/spec/). The instrumentation implementation MAY choose to either **append** the comment to the end of the query or **prepend** the comment at the beginning of the query, depending on the specific database system's requirements or preferences.
+
+| Attribute              | Type   | Description                           | Examples | Require level     | Stability                                                      |
+|------------------------|--------|---------------------------------------|----------|-------------------|----------------------------------------------------------------|
+| [`service.name`](/docs/registry/attributes/service.md) | string | Logical name of the service [22]   | `shoppingcart` | `Recommended`     | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova, line 493: this table talks about attributes, but it's under context propagation - I don't understand the suggestion here

docs/database/database-spans.md
```diff
@@ -478,6 +480,28 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+**[22] `service.name`:** Instrumentations MAY use [SDK-provided resource detectors](https://opentelemetry.io/docs/specs/semconv/resource/#semantic-attributes-with-sdk-provided-default-value) to set the default value for this attribute.
```
lmolkova, line 495: instrumentations don't have access to resources - resources are SDK concept and instrumentations only use API

docs/database/sql-server.md
```diff
@@ -152,6 +155,41 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+
+**Status**: [Development][DocumentStatus]
+
+### SQL commenter
+
+Instrumentations SHOULD **append** the comment to the end of the query.
+
+### SET CONTEXT_INFO
+
+Instrumentations MAY make use of [SET CONTEXT_INFO](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-context-info-transact-sql?view=sql-server-ver16) to add text format of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) before executing a query. This is an opt-in behavior that should be explicitly enabled by the user.
```
lmolkova, line 168: what's about tracestate, baggage and other possible headers?
context injection should be propagator concern, not instrumentation one

docs/database/sql-server.md
```diff
@@ -152,6 +155,41 @@ and SHOULD be provided **at span creation time** (if provided at all):
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Context Propagation
+
+**Status**: [Development][DocumentStatus]
+
+### SQL commenter
+
+Instrumentations SHOULD **append** the comment to the end of the query.
```
lmolkova, line 164: why position is important? what MS SQL instrumentation should put into the comment? in which format?

nenadnoveljic: > why position is important? 

In Postgres, prepended comments can break query hints, as shown [here](https://github.com/open-telemetry/semantic-conventions/pull/2236#discussion_r2084548720). Appended comments may be truncated in long queries. Proposal: use prepended comments by default, except for Postgres.

> what MS SQL instrumentation should put into the comment?

To transport static tags, such as `service` and `peer.service`, because they don't fit in `context_info`.

docs/database/sql-server.md
XSAM: > context injection should be propagator concern, not instrumentation one

I agree if SQL has a header concept (A list of key-value pairs), so it can have a dynamic space to put `tracestate` and `baggage`. We can also have a variety of propagators.

But, SQL does not have such a concept. And `CONTEXT_INFO` is the binary format in SQL Server, with a maximum size of 128 bytes. It is not feasible to assign more values due to this length limitation. Even if there is no limitation, I don't know what existing format we can use to put a list into a binary value.

So, I propose to use the value `traceparent`, a static value, for SQL Server.

docs/database/database-spans.md
XSAM: Good catch! I forgot this part.

Sounds like `service.name` should be considered a parameter of the instrumentation that users should provide. Until we have an EntityProvider in the future.

docs/database/database-spans.md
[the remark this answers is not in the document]
XSAM: > It'd be ok-ish with having it opt-in for other instrumentation.

Is it ok to doc `Instrumentations MAY propagate the context information to the SQL queries following [SQL commenter](https://google.github.io/sqlcommenter/spec/).` Use `MAY` and an annotation to express that it is an opt-in behavior?

I can also move it to MS SQL if you prefer, though I want to have a unified way to propagate context information.

docs/database/database-spans.md
XSAM: That means when using sqlcommenter for context propagation, we should propagate `service.name`.

The reason is here. https://github.com/open-telemetry/semantic-conventions/pull/2236#discussion_r2084585691. This is a tradeoff because the database itself won't provide trace info. We can use database receiver sending SQL to fetch query samples with trace info, but the chance of getting a complete trace (from SQL client to SQL server) is low due to sampling.

Do you think suggesting a `service.name` propagator would make this sentence easy to understand?

docs/database/sql-server.md
XSAM: > in which format?

Examples:

**append**: `SELECT * FROM songs /*service.name='shoppingcart'*/`
**prepend**: `/*service.name='shoppingcart'*/ SELECT * FROM songs`

Another discussion: https://github.com/open-telemetry/semantic-conventions/pull/2236#discussion_r2103582514

cheempz:
Just saw this and very interested in any effort to clarify the role of sqlcommenter in OTel. @tammy-baylis-swi and i will be sharing our learnings at the Open Observability Summit June 26th, would love to talk about it with any maintainers or TC members there.

XSAM:
Superseded by #2495

PR #2237, opened 2025-05-06
Add guidance and info metric for per cicd pipeline run metrics

### LINKED ISSUE #1184: [cicd] embed system attributes under cicd runners

kamphaus:
Fixes #1184

## Changes

This PR adds guidance on per pipeline run metrics and also adds the `cicd.pipeline.run.info` metric for linking a pipeline run entity to other entities.

## Prototype

A precursor metric `ci.podspan.info` (now named `cicd.pipeline.run.info` in this PR) was used in the [opentelemetry-agent-metrics-plugin](https://github.com/jenkinsci/opentelemetry-agent-metrics-plugin).

An otel-collector uses the span attributes and resource attributes (using their `cicd.pipeline` and `cicd.pipeline.run` entity links) to establish the link to k8s attributes (with `k8sattributes` processor) and emitting the info metric to allow querying.
https://github.com/jenkinsci/opentelemetry-agent-metrics-plugin/blob/main/collector.yaml

A Grafana dashboard allows querying for pod metrics (cpu, memory, network, disk, ...) when given the pipeline run id.
https://github.com/jenkinsci/opentelemetry-agent-metrics-plugin/blob/main/ci-pod-metrics-dashboard.json


adrielp (approved):

model/cicd/metrics.yaml
```diff
@@ -39,6 +39,21 @@ groups:
         requirement_level: required
     entity_associations:
       - cicd.pipeline
+  - id: metric.cicd.pipeline.run.info
+    type: metric
+    metric_name: cicd.pipeline.run.info
+    brief: 'This is an info metric linking pipeline runs to any other associated entities or information.'
```
adrielp, line 45: I don't have a helpful suggestion, but just a thought around how to make this brief more clear (if that's something we think needs to be more clear). Can we link to some reference documentation around what an info metric is?

lmolkova: +1 

It'd be helpful to include when it's reported - when the run starts? 

Since it results in high cardinality, does it need to be a metric? is it a time to consider using event instead? It would provide all the correlation too

docs/cicd/cicd-metrics.md
```diff
@@ -35,6 +37,36 @@ The conventions described in this section are specific to Continuous Integration
 **Disclaimer:** These are initial CICD metrics and attributes
 but more may be added in the future.
 
+### Guidance on per pipeline run metrics
+
+You MAY use metrics that include an attribute identifying the pipeline run by using the [`cicd.pipeline.run`](/docs/resource/cicd.md#cicd-pipeline-run) entity association.
```
lmolkova, line 42: does it apply to all metrics in this file?

I'm not sure it's possible to pick and choose which entity associations specific metric gets. When you create a meter it's associated with all entities that are defined on that meter provider. I don't see if it changes with https://github.com/open-telemetry/opentelemetry-specification/pull/456, checking with @jsuereth to confirm.


I.e. someone that adds `cicd.pipeline.run` entity should expect that all metrics emitted during the run would have association with it, including `cicd.pipeline.run.duration`.

model/cicd/metrics.yaml
```diff
@@ -39,6 +39,21 @@ groups:
...
+    unit: "1"
+    stability: development
+    entity_associations:
+      - cicd.pipeline
+      - cicd.pipeline.run
+      - cicd.worker
+      - host
+      - container
+      - k8s.namespace
+      - k8s.pod
```
lmolkova, line 56: is any of these entities happened to be provided on the worker, `cicd.pipeline.run.duration` would also be associated with them. Which makes me wonder what's the point of `cicd.pipeline.run.info` metric?

lmolkova:
Is there a prototype? Could you please add a link in the PR description? Thanks!

docs/cicd/cicd-metrics.md
kamphaus: The idea behind this guidance paragraph and the info metric is to be able to query for other metrics related to the runner: host or pod metrics for cpu, memory, disk.
To be able to easily show them in a dashboard we need to be able to select the run and then the agent for that run (in case multiple runners are involved).
There are two ways that allow to do this:
1. adding the `cicd.pipeline.run` entity to all the metrics (ie. to the host metrics and pod metrics).
2. or adding an additional metric which has both the `cicd.pipeline.run` entity and a `host` or `k8s.pod` entity. This would be the `cicd.pipeline.run.info` metric.

In this paragraph I meant to cover case 1. That the `cicd.pipeline.run` entity could be added to the host metrics and pod metrics.

model/cicd/metrics.yaml
kamphaus: > does it need to be a metric?

The `cicd.pipeline.run.info` metric is meant to cover [case 2](https://github.com/open-telemetry/semantic-conventions/pull/2237#discussion_r2250055298). We query this info metric in order to pivot from one attribute ([cicd.pipeline.run.id](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/cicd.md)) to the host or pod id as such it has to be a metric.

model/cicd/metrics.yaml
kamphaus: What do you mean?
None of the [cicd entities](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/entities/cicd.md) include a duration attribute.

model/cicd/metrics.yaml
kamphaus: > It'd be helpful to include when it's reported - when the run starts?

Although it would be nice to have a way to provide more information about a pipelines, pipeline runs and workers I think those would be more appropriately added as descriptive attributes on the entities.
It was my understanding from the entities SIG that at some point the descriptive attributes are meant to be conveyed in an efficient way (like an info metric maybe?), but I might be mistaken.

Since cicd.pipeline.run.info metric is primarily meant for linking the different entities I'd like to start with that and add other information later.

**Update:** Discussed in SemConv meeting 2025-08-04: Indeed entities to metrics signal transformations are on the Entities SIG roadmap, but could take a long time to arrive.

kamphaus: > Can we link to some reference documentation around what an info metric is?

The closest definition to what an "info" metric type is that I could find is this paragraph in the Otel specs: https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#info

Should I link to it?

kamphaus:
> Is there a prototype? Could you please add a link in the PR description? Thanks!

Done, I will update the metric name in the prototype once this PR is merged.

In the dashboard you can find examples of querying by the info metric:
https://github.com/jenkinsci/opentelemetry-agent-metrics-plugin/blob/5440c66be18d6563097963bcce37f1b37f21e9be/ci-pod-metrics-dashboard.json#L2558

docs/cicd/cicd-metrics.md
kamphaus: As discussed in SemConv meeting 2025-08-04:
In CICD we do not use a single SDK to attach entities to emitted signals.
Instead we use multiple SDKs where appropriate, for example:

* in the controller to emit controller-specific signals
* on the workers to emit signals related to the pipeline run executing on that worker

Alternatively we construct the otlp without using an SDK: for example in the otel-collector [github receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/githubreceiver).

> should expect that all metrics emitted during the run would have association with it

Since we can choose how a specific signal is emitted (eg. using which SDK) we may associate only the relevant entities to a given signal. In the example of the `cicd.pipeline.run.duration` it would most likely be emitted by the CICD controller without specific pipeline / pipeline run entity association. This allows using `cicd.pipeline.name` as an attribute for the `cicd.pipeline.run.duration` metric without having a specific pipeline run as a resource of that metric.

kamphaus:
### Discussed in SemConv meeting 2025-08-4

Fine to define info metrics to define entity relationships.

Best to have consistency with k8s info / state metrics

If we block this PR to have a wider discussion in SemConv maintainers, we can still move forward with defining such info metrics as opt-in in other system (in otel-collector as an example).

Improve the PR: give examples of use cases. Expand on the description.

kamphaus:
> Improve the PR: give examples of use cases. Expand on the description.

Since this PR is blocked on the discussion of info metrics (https://github.com/open-telemetry/semantic-conventions/issues/2595) I will split off the "Guidance on per pipeline run metrics" into its own PR to make progress on that.

Done: https://github.com/open-telemetry/semantic-conventions/pull/2618

kamphaus:
I'm closing this PR since we won't move forward on the blocking issue in the foreseeable future.
Once that issue is addressed we will open a new PR.

PR #2238, opened 2025-05-06
Expand the intended use of the cloud.region attribute

### LINKED ISSUE #2142: The correct attribution of cloud region for multi-regional client calls
Hi team, it is my first issue here, so let me know in case there are any missing details/practices. 

### Use case 

As part of https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3210 (Amazon Bedrock instrumentation), I'm hoping to introduce a proper semconv attribute for the cloud region. Looking at this made me wonder about the currently defined attributes for the cloud region and how they play with the scenarios of multi-regional calls from the client side.

When instrumenting client-side API calls to the cloud service running in different regions, it's valuable to capture the target region of the request. This may happen in the case of client's intended of balancing the load, regional failovers or sometimes the same service may offer additional features in different regions (for example Amazon Bedrock model availability differs per AWS region).

### Problem  

Currently, the standard semantic conventions include the incubating attribute [CLOUD_REGION](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating/attributes/cloud_attributes.py#L40)), defined as "The geographical region the resource is running in." This attribute correctly describes the location of the instrumented service (the resource emitting the telemetry).

However, there doesn't appear to be a standard semantic convention specifically designed to capture the cloud region of the remote peer or dependency service being called, particularly for inclusion on the client span representing that call.

### Proposal / Request for Discussion

I was hoping to receive some guidance on this if possible and see whether I'm misinterpreting the existing attribute or this is a viable use case for introducing a new attribute.

#### Alternatives considered:
- Using `CLOUD_REGION` from incubating attributes [here](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating/attributes/cloud_attributes.py#L40), but the description implied it being tied to a resource, not so much to the destination of an interaction represented by a span.  

#### Next steps:
I would appreciate feedback on:
1. Is this a recognised need? Do others require capturing the target service's region on client spans? So far I've seen such use case in botocore instrumentation [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py#L279).
2. Are the existing conventions already addressing this and I have missed their meaning?
   * In case this is a viable reason for a new attribute - ideally, this attribute would be associated with client spans and represent the cloud region of the target service being called. The exact mechanism for discovering this target region would be implementation-specific.

SrdjanLL:
Fixes #2142

## Changes

- Update the definition (brief) of the existing `cloud.region` attribute to clarify its application. The change ensures it covers use cases not only tied to describing resource location but also captures the region of the destination (dependency) being called.
- Update AWS docs to include this `cloud.region` as a span attribute recommendation (as mentioned [here](https://github.com/open-telemetry/semantic-conventions/issues/2142#issuecomment-2839931200)).


xrmx (approved):
A small nit but LGTM

lmolkova (approved):

joaopgrassi (approved):

Issue #1651, opened 2024-12-05
Logs/Events guidance on what to put to attributes vs body


lmolkova:
It's not clear what should go to event attribute vs body.

Some considerations:
- body, in practice, has much larger size limits
- body supports complex nested structure
- attributes are easier/more efficient to query and are easier to index

As a result, each event author needs to decide where each property goes and take into account backend limitations. 

We need some clear guidance in semantic conventions, but also this affects event API design.

jack-berg:
I know there's been a lot of back and forth about this, but I think there's a correct answer here: 

Use the body to record what happened. Use attributes as a means of capturing supplementary contextual information.

Semantic conventions should define fields for the body. Instrumentation authors should primarily record data to the body. The emit event API should prioritize the body, and provide great UX for adding to it. The emit event API should retain the ability to record attributes, but de-prioritize them and indicate that they are meant for contextualizing the event, rather than recording what happened. 

Reasoning:

* We want to be consistent. We've got to arrive at some guidance so that instrumentation and semconv authors are not each making decisions and reaching different conclusions. An ecosystem where some events are put their primary data in the body and others put their primary data in attributes is a bad result.
* We want to be able to differentiate between what happened and additional contextual information about it. I.e. users should be able to write a processor which enriches the log record with data from `Context`, without polluting the event itself. This is important because backends might want to do strict validation against the event schema and throw away information which doesn't conform. By putting the event in the body and contextual information in attributes, we have a clear distinction between the two. Backends can treat attributes as a sort of "user space" and apply different validation / retention rules than to the body. 
* On paper, log record body and attributes both support complex types, and no there's seemingly limited advantage to choosing one vs the other. But attributes only support complex types for the purpose of bridging. If we're designing a user facing emit event API, we could have this API accept the stricter [standard attribute definition][https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#standard-attribute]. This is symmetric to attributes in our other user facing APIs (spans, metrics, resources, etc). By putting the event's primary data in the body, we retain the ability to use complex types, and also retain a degree of symmetry with other APIs. We contain the user facing API surface area of complex types to a single place. Nice! 
* If you accept the "we want to be consistent" argument of the first bullet, then we're either putting the primary event data in the body or the attributes. We can prove that body is the right place by examining the result if we put the data in attributes:
[...]

Issue #828, opened 2024-03-21
Event semantic convention should document SeverityNumber 


lmolkova:
### Area(s)

area:other

## Description

Based on the spec-ed out Event parameters, events have optional [SeverityNumber](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md#eventlogger)

But none of the existing events document it and it's not covered in the https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/events.md

## Expected Result

We have a way to define and document severity for the events/

### Semantic convention version

1.24.0

MSNev:
Events SDK default for SeverityNumber 
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-sdk.md

lmolkova:
Discussed at semconv 6/27 call
- this is a tooling issue (schema)
- we'll need to discuss specific events that need severity

lmolkova:
Discussed at semconv WG 6/17:
- some events would need severity, we can discuss it when we introduce them. They are emitted by instrumentation libraries (e.g. GenAI response content should probably be Verbose) and severity can be used as the enablement mechanism for such events
- I've moved it to tooling project since the request here is to be able to specify the severity (i.e. support it in the schema definition and anything derived from it such as codegen)

PR #2240, opened 2025-05-07
Update general event guidance and policies to allow complex attributes on events only

### LINKED ISSUE #1669: Support complex attribute values on logs/events in semantic conventions

lmolkova:
Fixes #1669

Related: #1651, https://github.com/open-telemetry/opentelemetry-specification/pull/4485, https://github.com/open-telemetry/opentelemetry-specification/issues/4199

- Updates event guidance to use attributes instead of body
- removes mentions of Event API
- allows extended attributes on events
- adds policy that prevents usage of extended attributes beyond events for the time being.


docs/general/events.md
```diff
@@ -17,74 +17,56 @@ network connection severed, etc.
...
+OpenTelemetry Semantic Convention that define events SHOULD document the event name along
+with attributes and the type of the body.
 
 ## General event semantics
 
 * An event MUST have an [Event name property](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/logs/data-model.md#field-eventname)
-  that uniquely identifies the event. Event names are subject to the [Naming guidelines](/docs/general/naming.md).
-* Event MAY have [standard](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/common#attribute)
+  that uniquely identifies the event. Event names are subject to the [Naming guidelines](/docs/general/naming.md). Event name SHOULD be of a low-cardinality.
+* Events MAY have [extended](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/logs/data-model.md#field-attributes) attributes
```
lmolkova, line 31:
```suggestion
* Events MAY have [extended](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/logs/data-model.md#field-attributes) attributes
```

docs/general/events.md
```diff
@@ -17,74 +17,56 @@ network connection severed, etc.
...
-the event context.
+Events SHOULD be generated / produced / recorded using the
+[Logs API](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/logs/api.md#emit-a-logrecord)
+to ensure that the event is created using the configured SDK instance.
 
-Over time, some Events will be specified by OpenTelemetry and will have documented payload structure,
-field semantics, and stability and requirement levels. Other events may be user-defined and carry
-bespoke user semantics. When an Event name exists in the semantic conventions, its _payload_
-structure and semantics will also be defined.
+OpenTelemetry Semantic Convention that define events SHOULD document the event name along
```
trask, line 24:
```suggestion
OpenTelemetry Semantic Conventions that define events SHOULD document the event name along
```

docs/general/events.md
```diff
@@ -17,74 +17,56 @@ network connection severed, etc.
...
+with attributes and the type of the body.
```
trask, line 25:
```suggestion
with attributes and the type of the body if any.
```

docs/general/events.md
```diff
@@ -17,74 +17,56 @@ network connection severed, etc.
...
-    [Logs API](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/logs/api.md#emit-a-logrecord)
-    to ensure that the event is created using the configured SDK instance.
-  * The Emit Event API is not yet available in all OpenTelemetry SDKs. Check [spec-compliance matrix](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/spec-compliance-matrix.md#logs) to see the implementation status in corresponding language.
-* It's NOT RECOMMENDED to prefix the _payload_ (body) _fields_ with the `EventName` to
-  avoid redundancy and to keep the event definition clean.
-* The events SHOULD document their semantic conventions including event name,
-  attributes, and the payload.
+* Events MAY contain a body.
+* The event name uniquely identifies event structure: the type of the body and the
+  set of attributes.
```
trask, line 35: (just reordered to give priority to attributes)
```suggestion
* The event name uniquely identifies the event structure: the set of attributes and the type of the body.
```

docs/general/events.md
```diff
@@ -17,74 +17,56 @@ network connection severed, etc.
...
 Recommendations on using attributes vs. body fields:
 
-* If the field should be comparable across events with different `EventName` (or between an event and other telemetry items),
-  it should be an attribute.
-* If the field is specific to the event itself, then it should be a body field.
-* Body fields that belong to events with different event names are not comparable.
+It's RECOMMENDED to use the collection of attributes to represent the details of
+the event.
+
+Body SHOULD be used when bridging logs and MAY be used when mapping externally defined
```
trask, line 44: can we say something about bridging log attributes (e.g. https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/13787) to event attributes, and that this guidance only applies to something like log "message"?

docs/general/events.md
```diff
@@ -17,74 +17,56 @@ network connection severed, etc.
...
-When recording events from an existing system as OpenTelemetry Events, it is
-possible that the existing system does not have the equivalent of a name or
-requires multiple fields to identify the structure of the events. In such cases,
-OpenTelemetry recommends using a combination of one or more fields as the name
-such that the name identifies the event structurally. It is also recommended that
-the event names have low-cardinality, so care must be taken to use fields
-that identify the class of Events but not the instance of the Event.
+When recording events from an existing system as OpenTelemetry Events, the system
+may lack a single name field or require multiple fields to identify the event.
+In such cases, use a combination of fields to create a low-cardinality name.
```
trask, line 70:
```suggestion
In such cases, use a combination of fields to create a low-cardinality event name.
```

trask (approved):

docs/general/events.md
```diff
@@ -17,74 +17,55 @@ network connection severed, etc.
...
 In OpenTelemetry, Events are implemented as a specific type of [`LogRecord`](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/logs/api.md) that conforms to the conventions included here.
 
-In addition to a required name, an Event may contain a _payload_ (body) of any type permitted
-by the [LogRecord body](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-body).
-In its implementation, the Event _payload_ (body) will constitute the `Body` of the `LogRecord`.
-Like all other OpenTelemetry signals, an Event has optional attribute metadata that helps describe
-the event context.
+Events SHOULD be generated / produced / recorded using the
+[Logs API](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/logs/api.md#emit-a-logrecord)
+to ensure that the event is created using the configured SDK instance.
```
pellared, line 22: nit: The user does not need to use OTel SDK (it is fine to use a custom Logs API implementation). I suggest just removing this sentence to avoid confusion.

```suggestion
[Logs API](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/logs/api.md#emit-a-logrecord).
```

docs/general/events.md
```diff
@@ -17,74 +17,55 @@ network connection severed, etc.
...
 
-Over time, some Events will be specified by OpenTelemetry and will have documented payload structure,
-field semantics, and stability and requirement levels. Other events may be user-defined and carry
-bespoke user semantics. When an Event name exists in the semantic conventions, its _payload_
-structure and semantics will also be defined.
+OpenTelemetry Semantic Conventions that define events SHOULD document the event name along
+with attributes and the type of the body if any.
```
pellared, line 25: I think that it should also define severity level (even if this would be usually INFO).

docs/general/events.md
pellared: I suggest removing this paragraph.
I think we should have a separate document on "log bridge" guidelines and semantic conventions.

policies/attribute_types.rego
```diff
@@ -0,0 +1,24 @@
+package after_resolution
+import rego.v1
+
+# checks that complex attributes are not used in any groups except events
+deny contains group_stability_violation(description, group.id, attr.name) if {
```
jsuereth, line 5: I'm really happy with how readable these are for review.

docs/general/events.md
lmolkova: discussed at the Logs SIG meeting, let's remove it for now and revisit if we have use-cases for body in semconv

lmolkova: removed body and guidelines around it.

docs/general/events.md
lmolkova: I removed the sentence, we never prescribe how to create telemetry. you can create valid OTLP LogRecord in the collector as far as semconvs is concerned

docs/general/events.md
```diff
@@ -17,74 +17,31 @@ network connection severed, etc.
...
 
 ## General event semantics
 
 * An event MUST have an [Event name property](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/logs/data-model.md#field-eventname)
-  that uniquely identifies the event. Event names are subject to the [Naming guidelines](/docs/general/naming.md).
-* Event MAY have [standard](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/common#attribute)
+  that uniquely identifies the event. Event names are subject to the [Naming guidelines](/docs/general/naming.md). Event name SHOULD be of a low-cardinality.
+* Events MAY have [extended](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/logs/data-model.md#field-attributes)
   attributes that provide additional context about the event.
```
pellared, line 28: I suggest not using the "extended" term as it may confuse the users.
We do not have such term in the specification.
Moreover, the hyperlink references the log attributes which is correct.

```suggestion
* Events MAY have [attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/logs/data-model.md#field-attributes)
  that provide additional context about the event.
```

docs/general/events.md
```diff
@@ -17,74 +17,31 @@ network connection severed, etc.
...
-    [Logs API](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/logs/api.md#emit-a-logrecord)
-    to ensure that the event is created using the configured SDK instance.
-  * The Emit Event API is not yet available in all OpenTelemetry SDKs. Check [spec-compliance matrix](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/spec-compliance-matrix.md#logs) to see the implementation status in corresponding language.
-* It's NOT RECOMMENDED to prefix the _payload_ (body) _fields_ with the `EventName` to
-  avoid redundancy and to keep the event definition clean.
-* The events SHOULD document their semantic conventions including event name,
-  attributes, and the payload.
+* Events MAY contain a body of type [any](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/logs/data-model.md#field-body).
+* The event name uniquely identifies event structure: the set of attributes and
+* the type of the body.
```
pellared, line 31: Also would it not better to add it the first bullet point (or a second)?

```suggestion
* The event name uniquely identifies event structure: the set of attributes and
  the type of the body.
```

docs/general/events.md
```diff
@@ -17,74 +17,31 @@ network connection severed, etc.
...
-  attributes.
-* Events SHOULD be generated / produced / recorded using the
-    [Logs API](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/specification/logs/api.md#emit-a-logrecord)
-    to ensure that the event is created using the configured SDK instance.
-  * The Emit Event API is not yet available in all OpenTelemetry SDKs. Check [spec-compliance matrix](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.44.0/spec-compliance-matrix.md#logs) to see the implementation status in corresponding language.
-* It's NOT RECOMMENDED to prefix the _payload_ (body) _fields_ with the `EventName` to
-  avoid redundancy and to keep the event definition clean.
-* The events SHOULD document their semantic conventions including event name,
-  attributes, and the payload.
+* Events MAY contain a body of type [any](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/logs/data-model.md#field-body).
```
pellared, line 29:
```suggestion
* Events MAY contain a [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/logs/data-model.md#field-body) of type [`any`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/logs/data-model.md#type-any).
```

docs/general/events.md
```diff
@@ -17,74 +17,29 @@ network connection severed, etc.
...
-* Body fields that belong to events with different event names are not comparable.
-  For example, body field `id` on event `my_company.order_submitted` is semantically different from
-  field `id` on an event with name `session.start`.
+  that uniquely identifies the event. Event names are subject to the [Naming guidelines](/docs/general/naming.md). Event name SHOULD be of a low-cardinality.
+* Events MAY have [attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.45.0/specification/logs/data-model.md#field-attributes)
+  that provide additional context about the event.
+* Events MAY contain a [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.45.0/specification/logs/data-model.md#field-body) of type [`any`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.45.0/specification/logs/data-model.md#type-any).
+* The event name uniquely identifies event structure: the set of attributes and
+  the type of the body.
+* Events MAY have [severity number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.45.0/specification/logs/data-model.md#field-severitynumber).
```
pellared, line 32:
```suggestion
* Events SHOULD have [severity number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.45.0/specification/logs/data-model.md#field-severitynumber).
```
and move below 

> * An event MUST have an ...

Related to: https://github.com/open-telemetry/opentelemetry-specification/pull/4509

pellared (approved):

docs/general/events.md
lmolkova: I'd prefer to keep MAY here at least for the time being - none of the existing events have severity as of today and the tooling does not support setting it - https://github.com/open-telemetry/semantic-conventions/issues/828. 
I also have some doubts that all events should have fixed severity (e.g. resource provisioning event may be Info when successful, warning if retriable, error if completely failed)

PR #2267, opened 2025-05-14
[k8s] Add k8s node allocatable metrics

### LINKED ISSUE #2243: [k8s] Define semantic conventions for k8s node allocatable metric
## Area(s)

area:k8s

## Is your change request related to a problem? Please describe.

As part of the K8s SemConv stability work we need to define k8s node resource allocatable metric that is already in use by the [Opentelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.1.0/receiver/k8sclusterreceiver/internal/node/testdata/expected.yaml).

## Describe the solution you'd like

The existing metrics in use that we need to define as semantic conventions are the following:

`k8s.node.allocatable`

odubajDT:
Fixes #2243

## Changes

added 

```
k8s.node.allocatable.cpu
k8s.node.allocatable.ephemeral_storage
k8s.node.allocatable.memory
k8s.node.allocatable.pods
```

follow-up issue to handle the changes in k8scluster receiver [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40708)


model/k8s/metrics.yaml
```diff
@@ -74,6 +74,34 @@ groups:
       The actual accuracy would depend on the instrumentation and operating system.
     instrument: gauge
     unit: "s"
+  - id: metric.k8s.node.allocatable_cpu
```
joaopgrassi, line 77:
```suggestion
  - id: metric.k8s.node.allocatable.cpu
```
I wonder what others think, but I'd consider making `allocatable` also a namespace. https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#general-naming-considerations

Sorry if I missed a past discussion about this.

model/k8s/metrics.yaml
```diff
@@ -74,6 +74,34 @@ groups:
...
+    brief: "Amount of memory allocatable on the node"
+    instrument: gauge
+    unit: "By"
+  - id: metric.k8s.node.allocatable_pods
+    type: metric
+    metric_name: k8s.node.allocatable_pods
+    stability: development
+    brief: "Amount of pods allocatable on the node"
+    instrument: gauge
+    unit: ""
```
joaopgrassi, line 104:
```suggestion
    unit: "{pod}"
```

Any reason to have this one empty?

joaopgrassi (commented):
Is it correct that there are no attributes in any of these metrics?

model/k8s/metrics.yaml
odubajDT: I won't be against it, but this will require changes in the k8scluster receiver to match the new semconv

model/k8s/metrics.yaml
odubajDT: adapted

docs/system/k8s-metrics.md
```diff
@@ -256,6 +260,86 @@ The actual accuracy would depend on the instrumentation and operating system.
...
+<!-- semconv metric.k8s.node.allocatable_cpu -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability | Entity Associations |
+| -------- | --------------- | ----------- | -------------- | --------- | ------ |
+| `k8s.node.allocatable_cpu` | Gauge | `{cpu}` | Amount of cpu allocatable on the node | ![Development](https://img.shields.io/badge/-development-blue) |  |
```
dashpole, line 276: I think all of these should be `UpDownCounter`s

odubajDT: It's documenting the existing functionality, therefore it should be Gauge, since Gauge is used

model/k8s/metrics.yaml
ChrsMark: Probably we can just define SemConv "correctly" even if that'd be different from the current implementation. We do this for all the metrics of https://github.com/open-telemetry/semantic-conventions/issues/1032. We document these diffs at https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-migration.md for now and after we have stable SemConv we will adopt back the Collector components. I assume we can apply the same here?

docs/system/k8s-metrics.md
ChrsMark: As part of the wider K8s SemConv stability effort, we can focus on how it should be in SemConv and fix the Collector later. See https://github.com/open-telemetry/semantic-conventions/pull/2267#discussion_r2120289448.

model/k8s/metrics.yaml
```diff
@@ -74,6 +74,34 @@ groups:
...
+    stability: development
+    brief: "Amount of memory allocatable on the node"
+    instrument: gauge
+    unit: "By"
+  - id: metric.k8s.node.allocatable_pods
+    type: metric
+    metric_name: k8s.node.allocatable_pods
+    stability: development
+    brief: "Amount of pods allocatable on the node"
+    instrument: gauge
```
ChrsMark: I think `updowncounter` is more appropriate here:
```suggestion
    instrument: updowncounter
```

model/k8s/metrics.yaml
odubajDT: adapted accordingly

docs/system/k8s-metrics.md
odubajDT: adapted

model/k8s/metrics.yaml
```diff
@@ -102,6 +102,34 @@ groups:
       The actual accuracy would depend on the instrumentation and operating system.
     instrument: gauge
     unit: "s"
+  - id: metric.k8s.node.allocatable.cpu
```
ChrsMark, line 105: Please define the `entity_association` for these metrics. It should be `k8s.node`.

See for example https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/metrics.yaml#L507-L508

ChrsMark (commented):
Can we also list how these metrics are different to what Collect has today? We track those diffs at https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-migration.md#summary-of-changes

model/k8s/metrics.yaml
odubajDT: should be added

ChrsMark (approved):
LGTM! Thank's for contributing this!

dashpole (approved):

ChrsMark:
@open-telemetry/specs-semconv-maintainers @open-telemetry/specs-semconv-approvers this one has enough approvals from the K8s SIG. Should be on your side now :)

joaopgrassi (approved):
I think this PR is not affected by the discussion on naming UpDownCounters https://github.com/open-telemetry/semantic-conventions/pull/2317/files, so giving my approval.

jsuereth (approved):

PR #2278, opened 2025-05-16
Create Entities Registry

jsuereth:
Adds support for Entity registry and Entity stabilization policies.

- Creates an Entity Registry, which documents all entities configured across the `model/` directory.
- Updates YAML file so markdown generated is valid in the registry or in the existing file.

Replaces #2146


policies/entity_stability.rego
```diff
@@ -0,0 +1,48 @@
...
+    exceptions = {}
+    not exceptions[group.id]
+
+    ids := [attr |
+        some attr in group.attributes
+        attr.role == "identifying"
+    ]
+    count(ids) < 1
+
+    description := sprintf("Stable entity '%s' has no identifying attributes", [group.name])
```
lmolkova: do you think it should be a policy or a hard schema requirement? I.e. should become part of weaver in the long run?

model/telemetry/entities.yaml
```diff
@@ -8,10 +8,13 @@ groups:
     attributes:
       - ref: telemetry.sdk.name
         requirement_level: required
+        role: identifying
```
lmolkova: package name doesn't seem unique across languages, should language be identifying as well?

docs/registry/entities/android.md
```diff
@@ -0,0 +1,20 @@
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/entity_namespace.md.j2 -->
+
+# Android
```
lmolkova, line 6: should we remove convent of https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/android.md ? and redirect it to the new page? (and other resource definitions)

policies/entity_stability.rego
jsuereth: Hard requirement - and yes it should be in weaver.

I see three phases:

- Don't allow stable entities without roles in semconv
- Don't allow any entities without roles in semconv
- Don't allow any entities without roles in weaver

model/telemetry/entities.yaml
jsuereth: Good question.  I just took a rough cut first, I think language SHOULD be identifying here.

docs/registry/entities/android.md
jsuereth: I would like to, actually.

However - there is some content in markdown that can't be moved to YAML just yet, so I didn't pull the trigger.

If you want, I can open a tracking ticket for that and we can discuss how to get from A->B, but I don't think I can make that happen in this PR.

lmolkova: makes sense, yes, please create an issue

jsuereth: Done! https://github.com/open-telemetry/semantic-conventions/issues/2327

docs/registry/entities/zos.md
```diff
@@ -0,0 +1,27 @@
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/entity_namespace.md.j2 -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+# Zos
+
+## Zos Software
```
lmolkova: I think we're missing acronym filter in jinja

templates/registry/markdown/entity_namespace.md.j2
```diff
@@ -0,0 +1,46 @@
...
+
+**Descriptive Attributes:**
+
+{{ at.generate(desc_attrs, "", "/docs/registry/attributes", e.lineage.attributes) | trim }}
+{%- endif %}
+{%- if misc_attrs | length > 0 %}
+
+**Other Attributes:**
+
+_Note: Stable Entities MUST have attributes with a defined role._
```
lmolkova:
```suggestion
> :warning: This entity definition contains attributes without a role. Stable Entities MUST NOT have attributes without a defined role.
```

so it gives a warning to conventions authors that something needs to be changed in yaml and also tells how to fix it?

docs/registry/entities/zos.md
jsuereth: Good catch! Fixed

templates/registry/markdown/entity_namespace.md.j2
jsuereth: Fixed!, good suggestion.

docs/registry/entities/telemetry.md
```diff
@@ -0,0 +1,77 @@
...
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+**type:** `telemetry.distro`
+
+**Description:** The distribution of telemetry SDK used to capture data recorded by the instrumentation libraries.
+
+**Other Attributes:**
+
+> :warning: This entity definition contains attributes without a role.
+> Stable Entities MUST NOT have attributes without a defined role.
```
kamphaus, line 19: Is it documented that this means adding either of the following role property to the entity attribute (eg. here in `model/telemetry/entities.yaml`)?
```diff
+         role: identifying
```
or
```diff
+         role: descriptive
```

jsuereth: That should be in the modelling guide: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/how-to-write-conventions/resource-and-entities.md

kamphaus (approved):

.chloggen/wip-entity-registry.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: service
```
joaopgrassi:
```suggestion
component: entities
```

Since there's multiple "components" here, maybe we can use this [entities label](https://github.com/open-telemetry/semantic-conventions/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20label%3Aentities) I created sometime ago to group entities stuff together.

joaopgrassi (approved):

docs/registry/entities/otel.md
```diff
@@ -0,0 +1,27 @@
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/entity_namespace.md.j2 -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+# OTel
+
+## OTel Scope
+
+**Status:** ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Obsoleted.
```
kamphaus, line 10: Is it documented anywhere why this was obsoleted and what to use instead?

jsuereth: The `otel` attributes are for compatibility in non-otel environments.  There isn't any OpenTelemetry instrumentation that should be producing these attributes today, they represent concepts already modelled directly in OTLP, so just use those.

Basically, these should NOT have been an "entity" per-se, but just a set of attributes described for compatibility in non-OTLP environments.

PR #2283, opened 2025-05-19
[k8s] Adds k8s.hpa.metric.target.cpu.* metrics

### LINKED ISSUE #2182: [k8s.hpa] Add metrics/attrs for HPA metrics configuration
### Area(s)

area:k8s

### What's missing?

Introduce new metrics and attributes to provide further insight into k8s HPA's configuration. These metrics will be used to monitor the target values of the metrics specified in the HPA's `spec.metrics[]` field. HPA spec can be seen [here](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/horizontal-pod-autoscaler-v2/).

The metrics and attributes should convey the following information:
1. Track which metrics the HPA is configured to monitor.
2. Report the configured target values for these metrics, including relevant attributes to provide insight into the type.

Example of an HPA spec
```yaml
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  name: my-app-hpa
  namespace: default
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: my-app
  minReplicas: 1
  maxReplicas: 10
  metrics:
  - type: Object
    object:
      metric:
        name: requests-per-second
      describedObject:
        apiVersion: networking.k8s.io/v1
        kind: Ingress
        name: main-route
      target:
        type: Value
        value: 10k
  - type: ContainerResource
    resource:
      name: memory
      container: my-sidecar
      target:
        type: Value
        averageUtilization: 2Gi
````

#### **1. Metric: `k8s.hpa.metrics.target.value`**
- **Description**: Represents the target value of the metric being monitored by the HPA.
- **Instrument**: Gauge
- **Unit**: Varies by the metric type
- **Attributes**:
  - `k8s.hpa.metrics.target.type`: The type of metric source - possible values are `Resource`, `External`, `Object`, `Pods`, `ContainerResource`.
  - `k8s.hpa.metrics.target.name`: The name of the metric being tracked (e.g., `cpu`, `http_requests_per_second`, `queue_length`).
  - `k8s.hpa.metrics.target.value.type`: Specifies the type of value - possible values are `average_value`, `average_utilization`, `value`.
  - `k8s.hpa.metrics.target.container` (optional): The name of the container (if `type=ContainerResource`).
  - `k8s.hpa.metrics.target.object.kind` (optional): The kind of the object (if `type=Object`).
  - `k8s.hpa.metrics.target.object.name` (optional): The name of the object (if `type=Object`).
  - `k8s.hpa.metrics.target.object.apiversion` (optional): The name of the object (if `type=Object`).

*OR*

Break above metric into to 3 different metrics with the value of `k8s.hpa.metrics.target.value.type` attribute now being reflected in the metric name which I think is more inline with the semantic-conventions guidance -

#### **1. Metric: `k8s.hpa.metrics.target.value`**
- **Description**: Tracks the configured target value for the HPA metric.
- **Attributes**:
  - `k8s.hpa.metrics.target.type`: The type of metric source - possible values are `Resource`, `External`, `Object`, `Pods`, `ContainerResource`.
  - `k8s.hpa.metrics.target.name`: The name of the metric being tracked (e.g., `cpu`, `http_requests_per_second`, `queue_length`).
  - `k8s.hpa.metrics.target.container` (optional): The name of the container (if `type=ContainerResource`).
  - `k8s.hpa.metrics.target.object.kind` (optional): The kind of the object (if `type=Object`).
  - `k8s.hpa.metrics.target.object.name` (optional): The name of the object (if `type=Object`).
  - `k8s.hpa.metrics.target.object.apiversion` (optional): The name of the object (if `type=Object`).

#### **2. Metric: `k8s.hpa.metrics.target.average_value`**
- **Description**: Tracks the configured average target value for the HPA metric.
- **Attributes**:
  - Same as `k8s.hpa.metrics.target.value`.

### **3. Metric: `k8s.hpa.metrics.target.average_utilization`**
- **Description**: Tracks the configured average utilization target for the HPA metric (as a percentage).
- **Attributes**:
  - Same as `k8s.hpa.metrics.target.value`.

Given the variety of resources that HPAs can scale on (CPU, memory, external metrics, object metrics, etc.), the units for `k8s.hpa.metrics.target.value` and  `k8s.hpa.metrics.target.average_value` could vary significantly. For example, 500m for CPU, 2Gi for memory, or 1000 req/s for HTTP requests, etc. I think defining each possible combination (if even feasible) would introduce significant overhead. I am looking for suggestions on how best to represent the metrics' units in this situation.

### PR DESCRIPTION
**Pull request by jinja2**
Fixes #2182

## Changes

Adds metrics and attributes to track k8s HPA's metric target values for CPU resources.

### CODE REVIEW

File: model/k8s/metrics.yaml [hunk 1]
```diff
@@ -318,6 +318,69 @@ groups:
       [`k8s.hpa`](../resource/k8s.md#horizontalpodautoscaler) resource.
     instrument: updowncounter
     unit: "{pod}"
+  - id: metric.k8s.hpa.metric.target.cpu.value
```
jinja2, line 321:
```suggestion
  - id: metric.k8s.hpa.spec.metric.target.cpu.value
```
Should we make a distinction between spec vs status in k8s metrics if the field reported shows up in both?

dmitryax: I assume it's not feasible to get any metrics from the hpa status. If so, I don't think we need to add `spec`. It'd be a precedent that can cause significant complications to all other metric names :)

jinja2: The hpa reports the current metric in status, but if we want to report that in the future we can use currentmetric to differentiate with target. You are right, best to not complicate this further.

model/k8s/metrics.yaml
```diff
@@ -318,6 +318,72 @@ groups:
...
+  - id: metric.k8s.hpa.metric.target.cpu.average_utilization
+    type: metric
+    metric_name: k8s.hpa.metric.target.cpu.average_utilization
+    stability: development
+    brief: "Target average utilization, in percentage, for CPU resource in HPA config."
+    note: |
+      This metric aligns with the `averageUtilization` field of the
+      [K8s HPA MetricTarget](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#metrictarget-v2-autoscaling).
+
+      This metric SHOULD be reported against a [`k8s.hpa`](../resource/k8s.md#horizontalpodautoscaler)
```
ChrsMark: Now we can just use the `entity_associations` key to link metrics to entities.

model/k8s/metrics.yaml
```diff
@@ -318,6 +318,72 @@ groups:
...
+      resource. If the type of the metric is [`ContainerResource`](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-metrics-apis),
+      the `k8s.container.name` attribute MUST be set to identify the specific container within the pod to which the metric applies.
+    instrument: gauge
+    unit: "1"
+    attributes:
+      - ref: k8s.hpa.name
```
ChrsMark: We can use `entity_associations` and then there is no need to link the metric to the `k8s.hpa.name/uid` like this.

model/k8s/metrics.yaml
```diff
@@ -318,6 +318,72 @@ groups:
...
+      - ref: k8s.hpa.uid
+      - ref: k8s.namespace.name
```
ChrsMark: I wonder if this should be part of the `entity_associations`. An HPA metric is associated with an HPA, a namespace and conditionally to a k8s.container.

So I wonder if those should come from the entity associations instead of being metric attributes.

dashpole (approved):

model/k8s/metrics.yaml
jinja2: Yeah, these should be from the entity's attributes. <s>We don't define `k8s.namespace.name` as [entity attribute](https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/entities.yaml) on namespace-scoped k8s objects at present. Wdyt about adding it?</s>

jinja2: Oh, I see we can associate multiple entity. I am just going to add the `k8s.namespace` and `k8s.hpa` entities and don't think the attribute needs to be added anywhere. Thanks for the feedback!

jinja2: I have updated the PR with entity_associations.

jinja2: I think we should settle on our preferred approach for the namespace name attribute as mentioned [here](https://github.com/open-telemetry/semantic-conventions/pull/2322#issuecomment-2945683941).

ChrsMark (approved):
LGTM!

Not sure if the open question of https://github.com/open-telemetry/semantic-conventions/pull/2283#discussion_r2129864711 should block this or not. Probably we can proceed and revisit entity associations anyways.

jinja2:
@open-telemetry/specs-semconv-approvers this PR has SIG approvals. PTAL when you get a chance. Thanks!

jsuereth (approved):

joaopgrassi (approved):

ChrsMark:
@jinja2 that will need a rebase :)

jinja2:
Resolved conflicts

trask:
sorry, we didn't get to it in time, can you resolve the conflicts again? thanks

PR #2287, opened 2025-05-19
Add ONC/Sun RPC model and NFS system metrics

johnleslie:
## Changes

Add ONC/Sun RPC model and NFS system metrics

  * Added to rpc/registry.yaml:rpc.*  but did not change any existing elements


lmolkova:
@johnleslie thanks for the contribution!  
JFYI: we're starting RPC stabilization project https://github.com/open-telemetry/community/pull/2684, which is likely to result in significant refactoring of existing RPC conventions.

Please check out the blog post on the [scope and details](https://opentelemetry.io/blog/2025/stabilizing-rpc-conventions/)

model/system/metrics.yaml
```diff
@@ -470,3 +470,333 @@ groups:
       - ref: linux.memory.slab.state
     entity_associations:
       - host
+
+  # metric.system.nfs.* metrics
```
braydonk: I think there are a substantial enough number of `system.nfs.*` and `nfsd.*` metrics that it deserves its own namespace. It would make a lot of these metric names shorter, and we could move them into their own folder and documentation file.

You can make a `nfs` folder in `model`, move these into a `metrics.yaml` file that contains the `nfs.*` and `nfsd.*` and remove the `system.` prefix from them. That folder can be added to `.github/CODEOWNERS` with the general semconv approvers group and the system semconv group as owners.

When these metrics are moved to an `nfs` folder we'll also want to generate `.md` files with them. I have never created one of these files from scratch so I don't know if there's an easier way than what's written [in CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#2-update-the-markdown-files). @open-telemetry/specs-semconv-approvers does anyone know if there's an automated process to start this?

model/system/metrics.yaml
```diff
@@ -470,3 +470,333 @@ groups:
...
+  - id: metric.system.nfs.net.count
+    type: metric
+    metric_name: system.nfs.net.count
+    stability: development
+    brief: "Reports the count of kernel NFS client network requests handled"
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.netcnt, which includes both UDP datagrams and TCP segments
+    instrument: gauge
```
braydonk: OTLP makes a particular distinction between the `Gauge` and `Counter`/`UpDownCounter` instruments:
https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments
https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/#instrument-selection

The main difference is that `UpDownCounter` is to be used for metrics that track an additive measurement. If I'm not mistaken, every `gauge` in this set of metrics can be either a `Counter`/`UpDownCounter` and none fall into the OTLP definition for `Gauge`. I think they can all be changed to one or the other, where `Counter` is used for values that only ever increase and `UpDownCounter` would be for values that can decrease.

Unless there is something unique about this stat from `nfs`, I believe most of these metrics that would be `Counter`s since they appear to be counting ever-increasing values over time. The only metric I see here that strikes me as an `UpDownCounter` would be `nfsd.thread.count`, assuming that this number could be changed up or down during `nfsd` runtime. 

(Worth noting that some backends will treat `UpDownCounter` and `Gauge` as the same thing, though I think most do make the distinction between `Counter` and `Gauge`)

braydonk (commented):
I think the file permission changed on `model/system/metrics.yaml` and I think it was potentially unintentional. Should that be reverted?

johnleslie:
> I think the file permission changed on `model/system/metrics.yaml` and I think it was potentially unintentional. Should that be reverted?

fixed

model/system/metrics.yaml
johnleslie: I've renamed the system.nfs metrics to nfs and updated .github/CODEOWNERS.

I tried generating the docs by following the CONTRIBUTING.md instructions -- creating a template in docs/nfs/nfs-metrics.md and running $ make table-generation attribute-registry-generation

I've tried several values for the nfs-metrics.md template's `new-group-id` but consistently get the following error. ie, with:

```
docs/nfs/nfs-metrics.md:<!-- semconv nfs -->
docs/nfs/nfs-metrics.md:<!-- endsemconv -->
```

I get:

ℹ Updating: $/home/weaver/target/nfs/nfs-metrics.md
✖ Could not find: nfs

I've tried the latest version of weaver and get the same error. Any idea?

model/system/metrics.yaml
johnleslie: Most of these metrics do only increase (except `metric.nfs.server.thread.count` which is more of an `updowncounter`), however they can reset to 0 upon server restart or host reboot.. would that make them gauge or counter?

braydonk: It will still classify as a `Counter` under that scenario. All metrics in the OTLP data model will come with a `StartTime`, which is the timestamp of the first point observed in the series. For a `Counter` metric, if a value is detected that is not equal/increasing, this is considered a "counter reset" and the `StartTime` of the series is updated to the new value. Collection technologies (including the Collector where these metrics are currently targeting most closely) are built to handle this for `Counter` metrics automatically.

So I think for this PR, each of these metrics can become a `sum` instrument, and most can be `monotonic: true` other than `nfs.server.thread.count`.

model/system/metrics.yaml
braydonk: I think there needs to be a `semconv` group for every metric individually. Definitely a little bit tedious, but I don't think an automated process exists for it yet.

You'll also need to include a header section similar to other `md` files that include the preamble and table-of-contents autogen section. I think those can be mostly copied from any other `*-metrics.md` file.

I created a gist that you should be able to work from as a starting point: https://gist.github.com/braydonk/93845983e1ca6429aa0b5528fb83ff65 (you'll want to change it to the `Raw` view to see all the comments for autogen sections).

braydonk: Additionally, while the metrics in this file were moved correctly to their own file, it looks like the `system` name is still present in all of the metrics (at least according to the latest commit as of this comment's writing).

model/system/metrics.yaml
johnleslie: done

braydonk: Thanks! I made a mistake in the last comment, getting mixed up between OTel Collector yaml metric defs and semconv; `instrument: counter` is right and all the `monotonic: true` lines can be removed.

model/system/metrics.yaml
johnleslie: - created nfs-metrics.md using your template and `make table-generation attribute-registry-generation` -- looks good, thanks!
- removed obsolete system.nfs metrics

.github/CODEOWNERS
```diff
@@ -151,4 +151,8 @@
 /docs/mainframe                   @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
 /model/mainframe                  @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
 /docs/zos                         @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
-/model/zos                        @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
\ No newline at end of file
+/model/zos                        @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
+
+# NFS
+/docs/nfs                         @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
```
braydonk: `sig-mainframe-approvers` can change to `semconv-system-approvers`

docs/nfs/nfs-metrics.md
```diff
@@ -0,0 +1,246 @@
...
+linkTitle: NFS
+--->
+
+# Semantic conventions for NFS metrics
+
+**Status**: [Development][DocumentStatus]
+
+This document describes instruments and attributes for NFS and NFSD metrics. 
+
+<!-- toc -->
```
braydonk, line 11: I think a `make markdown-toc` will fill this section out.

braydonk (commented):
Nice, just about there!

.github/CODEOWNERS
johnleslie: done

docs/nfs/nfs-metrics.md
johnleslie: done

.chloggen/model-nfs.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: nfs
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add rpc.onc.* / rpc.nfs.* model to support new metric.nfs.*/metric.nfsd.*
```
braydonk: Suggested change to the changelog note to mention the metrics first and some formatting changes since we know it'll be rendered into Markdown.

```suggestion
note: Add `nfs.*`/`nfsd.*` metrics, along with supporting `rpc.onc.*`/`rpc.nfs.*` attributes.
```

docs/nfs/nfs-metrics.md
```diff
@@ -0,0 +1,654 @@
...
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`rpc.onc.procedure.name`](/docs/registry/attributes/rpc.md) | string | ONC/Sun RPC procedure name | `OPEN`; `READ`; `GETATTR` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`rpc.onc.version`](/docs/registry/attributes/rpc.md) | int | ONC/Sun RPC program version |  | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
```
braydonk, line 471: I think you can add these to the bottom of the document to resolve markdown lint:
```
[MetricRecommended]: /docs/general/metric-requirement-level.md#recommended
[MetricOptIn]: /docs/general/metric-requirement-level.md#opt-in
```

johnleslie: Done

model/rpc/registry.yaml
```diff
@@ -274,3 +274,30 @@ groups:
...
+      - id: rpc.onc.program.number
+        type: int
+        stability: development
+        brief: "ONC/Sun RPC program number"
+      - id: rpc.onc.program.name
+        type: string
+        stability: development
+        brief: "ONC/Sun RPC program name"
+        examples: ['portmapper', 'nfs']
+      - id: rpc.nfs.operation.name
```
thompson-tomo: Would like to see this in a storage namespace and ideally not tied to nfs

model/rpc/registry.yaml
```diff
@@ -274,3 +274,30 @@ groups:
...
+      - id: rpc.onc.procedure.name
+        type: string
+        stability: development
+        brief: "ONC/Sun RPC procedure name"
+        examples: ['OPEN', 'READ', 'GETATTR']
+      - id: rpc.onc.program.number
+        type: int
+        stability: development
+        brief: "ONC/Sun RPC program number"
+      - id: rpc.onc.program.name
```
thompson-tomo: Could we not use rpc.system?

model/rpc/registry.yaml
```diff
@@ -274,3 +274,30 @@ groups:
...
         brief: "Uncompressed size of the message in bytes."
+      - id: rpc.onc.version
+        type: int
+        brief: "ONC/Sun RPC program version"
+        stability: development
+      - id: rpc.onc.procedure.number
+        type: int
+        stability: development
+        brief: "ONC/Sun RPC procedure number"
+      - id: rpc.onc.procedure.name
```
thompson-tomo: Could we not use rpc.method?

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
...
+    metric_name: nfs.client.rpc.authrefresh.count
+    stability: development
+    brief: "Reports the count of kernel NFS client RPC authentication refreshes"
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.rpcauthrefresh
+    instrument: counter
+    unit: "{authrefresh}"
+    entity_associations:
+      - host
+
```
thompson-tomo: Should this be in the rpc namespace instead?

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
...
+    metric_name: nfs.client.net.tcp.connection.count
+    stability: development
+    brief: "Reports the count of kernel NFS client TCP connections accepted"
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.nettcpconn
+    instrument: counter
+    unit: "{connection}"
+    entity_associations:
+      - host
+
```
thompson-tomo: Could this not be in the network namespace?

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
+groups:
```
thompson-tomo, line 1: Needs further refinement after other feedback.

braydonk:
All the checks are passing and all my comments are resolved. I'll leave it now to folks for feedback about the metrics themselves, which I don't know enough to have an opinion on.

model/rpc/registry.yaml
johnleslie: That would mean enumerating several ONC/Sun RPC-specific protocols within rpc.system. It feels to me like a rpc.system member would be ONC/Sun RPC itself (I had it in an earlier revision, but it was unused so I removed it).

model/rpc/registry.yaml
johnleslie: NFSv4 has both operations and methods, so we can't overload rpc.method with both.

NFSv3 has i.e. OPEN as an ONC RPC method
NFSv4 would have COMPOUND as an ONC RPC method, and OPEN and READ as a NFSv4 operations.

model/rpc/registry.yaml
johnleslie: How would we differentiate operations and procedures if it's in a generic storage namespace?

model/nfs/metrics.yaml
johnleslie: This metric is the TCP connection count for all NFS clients under a `host`. How could we isolate a network.* metric (none exists currently) to NFS clients?

model/nfs/metrics.yaml
johnleslie: How would we isolate an RPC-namespaced metrics to values specific to NFS clients?

model/rpc/registry.yaml
thompson-tomo: I would added the network.protocol.name attribute to the metrics hence enabling your filtering.

model/rpc/registry.yaml
thompson-tomo: You are correct about rpc.system. I think that i meant to write rpc.service but yes I would also set rpc.system as you mentioned.

model/nfs/metrics.yaml
thompson-tomo: I would use the network.protocol.name attribute to enable the necessary drill down.

model/rpc/registry.yaml
thompson-tomo: I think we can describe the meaning of procedure given that it is a totally different name.

model/nfs/metrics.yaml
thompson-tomo: I would have the metrics include the network.protocol.name attribute.

model/rpc/registry.yaml
thompson-tomo: I would for v4 compound the operation and method but v3 just have method.

model/rpc/registry.yaml
johnleslie: hmm, the metrics themselves are already in the NFS namespace i.e. metric.nfs.client.net.udp.count. I think adding a  network.protocol.name="nfs" would be redundant. Am I misunderstanding?

model/rpc/registry.yaml
johnleslie: re-added onc_rpc to rpc.system

model/rpc/registry.yaml
johnleslie: Unfortunately in Linux we don't have combined statistics for operations... we only have counters for operations and counters for methods. Any issue leaving as-is?

model/nfs/metrics.yaml
johnleslie: authrefresh is RPC-specific though.. Can you provide an example?

model/nfs/metrics.yaml
johnleslie: Hmm, the NFS tcp connection counts would be such a tiny fraction of the overall count of available protocols I'm not sure this makes sense... we currently don't have any metrics for tcp connection counts by protocol, and the vast majority aren't available.

model/rpc/registry.yaml
thompson-tomo: So what I have done is add an id attribute as a part of #2405 and that way you can use `peer.service.*` namespace rather than introducing `rpc.onc.program.*` namespace. Note My initial suggestion was to use `rpc.service` but the peer is a nicer solution.

model/rpc/registry.yaml
thompson-tomo: In your scenario it would be redundant however the driver for using the storage namespace would be to have a metric such as metric.storage.volume.operation.count which you can filter based on protocol ie nfs, volume, operation.

This would provide a highly reusable solution and if someone wants to see samba/smb metrics it would be the same base as nfs. I have started sketching the attributes via: https://github.com/thompson-tomo/semantic-conventions/commit/d7bb890abeddd3ff954605e509f442b1d6ba1e28

model/rpc/registry.yaml
thompson-tomo: How about this: 
- `rpc.method` is used instead of `rpc.onc.procedure.name` regardless of version as originally proposed

That way we can reuse the existing convention for spans including the naming which requires a method & avoid duplication of attributes.

model/nfs/metrics.yaml
thompson-tomo: Agree nfs would likely be a small fraction but overall I think it would be a useful addition to the existing metrics. Scenario would be analysing metrics generated by a firewall (which is a system) hence the additional attribute would be useful.

model/nfs/metrics.yaml
thompson-tomo: So for auth refresh that scenario could be addressed by #2434 but others are certainly generic rpc.

model/nfs/metrics.yaml
braydonk: @open-telemetry/specs-semconv-maintainers @open-telemetry/specs-semconv-approvers - request for comment as well

Unification of the udp datagram and tcp segment count metrics could be cool, as there is information about these counts in the system that would be good to have metrics for. However, the method of retrieving the overall udp/tcp counts on the system would purely be an entire count for the system and not broken down by any particular protocol/identifying dimension. So if we were to share this hypothetical metric with `nfs`, both of which being reported under a host entity, the best way to tell which metric is intended for `nfs` would be to have some label that is always a specific value. I'm not sure how to nicely convey this in semconv yaml definitions, I guess just stated in the brief with normative language?   
Additionally, you couldn't aggregate these metrics together; the `nfs` counts would be a strict subset of `system.network` counts, meaning that summing them together would double-count all the `nfs` counts.

Given this, I am leaning towards recommending that these remain under the `nfs` namespace and not attempt to unify with `system.network`, as the benefits of sharing are outweighed by the difficulties in usage that are very difficult to convey to people consuming the conventions. Am I missing something in terms of the value of trying to share these metrics?

thompson-tomo: Alternatively if system.network doesn't seem feasible what is the thought about unifying tcp/UDP but putting it the rpc namespace?

braydonk: If it seems feasible that the semantics used here would work ported over to the `rpc` space with some manner of `protocol` attribute to define the `nfs` ones I could see that working. I don't know enough about RPC to say for sure but I could see it working. The only thing I wonder here is how the NFS metrics will be differentiated for some things like RPC count which are shared, but some are for client and some are for server. They both are sets of stats that come from the kernel so it still makes sense to report them against the same `host` resource, and so need another way to differentiate.

thompson-tomo: We would have agree we would have atleast the system name &/or network protocol to differentiate data.

model/rpc/registry.yaml
johnleslie: cc @braydonk how should we proceed here? I'm ok with i.e. metric.storage.volume.operation.count as in https://github.com/thompson-tomo/semantic-conventions/commit/d7bb890abeddd3ff954605e509f442b1d6ba1e28

We do need a way to accommodate both RPC procedures, which are the main operation types for NFSv1-3, as well as NFSv4 operations. NFSv1-3 has an RPCs operation of i.e. READDIR, whereas NFSv4 has an RPC operation of COMPOUND that includes multiple operations, such as OPEN and READ. We can't conflate the two, as the Linux kernel provides a count of NFSv4 procedures such as COMPOUND.

Something like:

- id: metric.storage.volume.operation.count
    - attributes:
      - ref: ???.name (operation/procedure name as in OPEN, READ)
      - ref: network.protocol.name (RPC, NFSv4)

model/rpc/registry.yaml
johnleslie: Hmm, I don't see how peer.service.name works for protocols such RPC and NFS, given the brief and examples:

      - ref: peer.service.name
        brief: "The name of the service to which a request is made, as returned by the AWS SDK."
        examples:
          - DynamoDB
          - S3

model/nfs/metrics.yaml
johnleslie: I don't think access_control is quite relevant here. This metric comes from a Linux-specific counter of the number of times a Sun/ONC RPC client has refresh authentication credentials. Other RPC clients (OSes) may have this counter as well.

What do you recommend we do here? Agree it can be more generic than metric.nfs.client.rpc.authrefresh.count, but I think it should be a Sun/ONC RPC specific metric.

model/rpc/registry.yaml
thompson-tomo: I would have an attribute ie storage.volume.volume.action/operation which in v1-3 is populated with the procedure value & v4 is populated with the operation value.

In all cases the procedure is set as the `rpc.method` as per https://github.com/open-telemetry/semantic-conventions/pull/2287#discussion_r2173633610

model/rpc/registry.yaml
thompson-tomo: Note this comment is in relation to `rpc.onc.program.number` & `rpc.onc.program.name` which has a name defined as 

> The service.name of the remote service. 

With an example of

> AuthTokenCache

There was no field for number hence adding of the if.

model/rpc/registry.yaml
johnleslie: For NFSv4 we need to report both operations _and_ procedures. My schema above works, but it is network specific, so I'm not sure if it makes sense to have network.protocol.name on metric.storage.volume.operation.count unless it can be left empty without issue.

 . I.e. for a single NFSv4 request of: procedure=COMPOUND, operation=OPEN,operation=READ we'd result in:

id: metric.storage.volume.operation.count=123+1
  - attributes:
    - ref: rpc.method=COMPOUND
    - ref: network.protocol.name=RPC

id: metric.storage.volume.operation.count=456+1
  - attributes:
    - ref: rpc.method=OPEN
    - ref: network.protocol.name=NFSv4

id: metric.storage.volume.operation.count=789+1
  - attributes:
    - ref: rpc.method=READ
    - ref: network.protocol.name=NFSv4

does this work?

model/nfs/metrics.yaml
johnleslie: As braydonk mentioned they are reported per-nfs-client and per-nfs-server. What's the proposed schema here?

model/rpc/registry.yaml
thompson-tomo: I was more thinking along the line of for the metric.
```
id: metric.storage.volume.operation.count
  attributes:
    ref: network.protocol.name=NFS
    ref: storage.volume.action/operation= OPEN
```

That way we have a consistent count of operation/actions.

model/nfs/metrics.yaml
thompson-tomo: Howabout at a minimum

```
metric.rpc.client.connection.count
  attributes:
    rpc.system
    network.transport
    network.type
```

with a similar solution for the server

johnleslie: For these Linux /proc NFS metrics, I can see:

rpc.system=onc_rpc
network.transport=tcp
network.type=??? it's not broken down by IPv4 or IPv6. do we leave this blank?

Also, we need to specify NFS here. how about if we add "rpc.onc.program.name" as an attribute?

model/rpc/registry.yaml
johnleslie: Can you sketch out what you're proposing here?

model/rpc/registry.yaml
johnleslie: can network.protocol.name be "onc_rpc" and "nfsv4"? If so this works.

model/nfs/metrics.yaml
thompson-tomo: Yes leave blank if not known, perhaps should be opt-in.

In relation to `rpc.onc.program.name` that seems too specific however `peer.service.name` does seem to fit for the client metric & `service.name` for the server metric.

johnleslie: Done (https://github.com/open-telemetry/semantic-conventions/pull/2287/commits/0240b4ddaee3e72f60a9632b2ca62278955cd8e9). Note I'm not 100% convinced about having network.type in there -- it's not obvious that you'd find the total in the metric with this field blank

model/nfs/metrics.yaml
johnleslie: @thompson-tomo friendly ping

model/rpc/registry.yaml
johnleslie: Can we get https://github.com/thompson-tomo/semantic-conventions/commit/d7bb890abeddd3ff954605e509f442b1d6ba1e28 merged so I can implement this?

johnleslie: Friendly ping

model/rpc/registry.yaml
johnleslie: SGTM -- just need to merge https://github.com/thompson-tomo/semantic-conventions/commit/d7bb890abeddd3ff954605e509f442b1d6ba1e28 so we can use storage.volume.volume.action/operation

model/rpc/registry.yaml
braydonk: Adding `network.protocol.name` to `storage.volume.operation.count` feels a bit weird. Will all other usages of `storage.volume.operation.count` also use that attribute? The network protocol here just feels like a way to add a dimension for NFS's sake rather than something universally applicable.

model/nfs/metrics.yaml
braydonk: Do other RPC clients count retransmit or authrefresh? Whenever I look into supporting materials on either one, everything that comes up is NFS related. These metrics are also designed to match up directly with Sun RPC counters from the kernel, the only consumer of which also appears to be NFS. If it isn't really clearly universally applicable, I'd rather not abstract it to the general RPC namespace just for the sake of it.

braydonk (commented):
Given that these metrics as they exist in the PR line up so closely with what `/proc/net/rpc/nfs` and `/proc/net/rpc/nfsd` provide, and as a result the common stat gathering practices like the `nfsstat` command, I'm skeptical about the usefulness of trying to abstract this stuff to generic `rpc` and `storage` namespaces. It looks like we will be cluttering those namespaces with things that would only make sense for `nfs`, and simultaneously making the usage of the conventions for NFS more difficult. I like what is here currently, and it has an [existing implementation PR in the Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40134) proving out the schema. I think there would need to be a really convincing reason to uproot this PR as it is.

model/rpc/registry.yaml
jsuereth: @thompson-tomo can you outline, more specifically, what use cases you're seeing here for treating RPC + NFS as the same?

Is this something most o11y users would do?  View NFS saturation the same as other RPC services?

I'm not sure I buy this abstraction as being an improvement to general observability and I'm struggling to understand why we would join these two patterns. 

We haven't written up our guidance (AFAIK) on when to provide abstract namespaces and when to divide, but I feel pushing NFS into RPC is one step too far.

cc @open-telemetry/specs-semconv-maintainers for thoughts.

lmolkova: I believe we advise against overly-generalizing and suggest to stick to external standards

> When defining a broad attribute applicable across multiple domains or systems, check for existing standards or widely accepted best practices in the industry. Avoid creating generic attributes that are not based on established standards.


https://github.com/open-telemetry/semantic-conventions/blob/69f40e1ccdb32e28c6ca63a44d7348a07760aeef/docs/how-to-write-conventions/README.md?plain=1#L127-L151

I don't believe treating NFS as RPC makes sense, we've blocked previous attempts to make RPC cover anything but RPC frameworks - https://github.com/open-telemetry/semantic-conventions/pull/1622

model/nfs/metrics.yaml
johnleslie: Currently NFS is the only Linux RPC client that exposes retransmit and authrefresh via /proc. It does this via rpc_proc_register/rpc_proc_show: https://github.com/search?q=repo%3Atorvalds%2Flinux%20rpc_proc_register&type=code

All Linux RPC clients count these, but they don't make the statistic available as NFS does.

model/rpc/metrics.yaml
```diff
@@ -160,3 +160,33 @@ groups:
       Should be 1 for all non-streaming RPCs.
 
       **Streaming**: This metric is required for server and client streaming RPCs
+
+  - id: metric.rpc.client.connection.count
```
braydonk: I'm not sold on this being in `rpc`. The convention was pretty clean when it was under `nfs`, since the meaning of the metric is clear from its definition rather than needing numerous dimensions to properly mould this to generic `rpc` usability. If this metric would be useful for `rpc` in general I think it could be done in a separate PR defining it in the way that makes the most sense, rather than trying to force it as part of the `nfs` conventions.

I suggest reverting these back to the original definitions in the `nfs` namespace.

johnleslie: Agree with this -- the 4 attributes under `rpc` make this hard to work with.

model/rpc/registry.yaml
thompson-tomo: Alternatively we could go with storage.volume.type that way we are contained with storage

model/rpc/registry.yaml
thompson-tomo: My preference is for nfs telemetry to be in a storage namespace as commented on this PR. That way we can compare nfs storage volumes to alternatives.

What I did suggest is based on PR initial formnis why don't we reuse existing attributes for things like procedure & program rather than introducing specific properties when alternatives are pre-existing

model/rpc/registry.yaml
jsuereth: agreed with @braydonk here - @thompson-tomo I think where you push for abstraction, we need to understand the value and how that abstraction will be used.

The goal of having a general "storage" set of metrics is laudable, but don't we already have that?  I see this PR adding the *deep* specific NFS metrics we need to monitor NFS directly.  Does the thing we're pushing to abstract actually provide value?   Should we force a "storage" paradigm beyond the notion of disk operations we already have for system metrics?

I think this decision lies with the @open-telemetry/semconv-system-approvers, but I'll add that I'm also unconvinced this abstraction makes sense here.

model/nfs/metrics.yaml
jsuereth: I'm not sure I understand that value of these labels on the metric.

model/rpc/metrics.yaml
johnleslie: reverted this

model/nfs/metrics.yaml
johnleslie: https://github.com/open-telemetry/semantic-conventions/commit/0240b4ddaee3e72f60a9632b2ca62278955cd8e9 is reverted

model/rpc/registry.yaml
braydonk: > My preference is for nfs telemetry to be in a storage namespace as commented on this PR. That way we can compare nfs storage volumes to alternatives.

I think it's possible for both of our goals to coexist.

We don't currently have a more generic `storage` namespace. I think it could and should exist. Within the @open-telemetry/semconv-system-approvers group though, we purposely strive to match the instrumentation source with the namespace.

In this case, I can envision instrumentation (focusing on Linux for the sake of example) that would get all the mounted filesystems, and use something like [`statvfs(2)`](https://man7.org/linux/man-pages/man2/statvfs.2.html) on each to get information out of each of them. The filesystem type would be presumably an attribute on an Entity or attribute on the Datapoint (presumably the former but I have not deeply thought about design implications of either). The metrics produced by this information would be under `storage.*`.

With the stated goal of comparing storage volumes of different types, in my opinion this accomplishes the goal more simply and directly. Different filesystems will have all kinds of different stats, and there are likely overlaps between them, but rather than trying to bolt different filesystem stats together in some shared abstract `storage.*` namespace metrics, the `storage.*` namespace only contains the types of stats that can be 1:1 directly compared regardless of filesystem type. I think the `storage` namespace should be separately designed with this in mind, rather than trying to force `nfs` conventions within this PR to fit into a mold that we then need to find ways for all filesystem types to fit their stats into.

The conventions within this PR are directly targeted to instrumentation for `nfs` in particular (i.e. doing the same sort of thing the `nfsstat` command does reading the stats from `procfs`). There could still be separate `storage` instrumentation that would also get its own information about those storage volumes, but I think that's fine and these conventions existing as they do does not lock us out of that future. These metrics exist to reflect the real life patterns of monitoring systems; there are generic APIs and mechanisms to get information about any storage volumes, and there are specific APIs for getting information about `nfs` storage volumes, and there are separate conventions for the important bits of both.

Given this, I feel pretty strongly that we should move ahead with this PR, and design the storage namespace to be generic and reflect the APIs and mechanism used to get general storage volume information.

---

> What I did suggest is based on PR initial formnis why don't we reuse existing attributes for things like procedure & program rather than introducing specific properties when alternatives are pre-existing

Similar to matching metrics to instrumentation sources, attributes and metadata should similarly match existing protocol definitions and the typical ways the information is used. ONC RPC programs have a particularly designed protocol and ways to identify programs and procedures (referring to [`RFC-5531`](https://datatracker.ietf.org/doc/html/rfc5531#page-4) here) that differs from RPC programs that use different protocols. Having these attributes reflect directly what is in the ONC RPC protocol definition lines up with our goals of metrics/attributes reflecting instrumentation.

.github/CODEOWNERS
```diff
@@ -151,4 +151,8 @@
 /docs/mainframe                   @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
 /model/mainframe                  @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
 /docs/zos                         @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
-/model/zos                        @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
\ No newline at end of file
+/model/zos                        @open-telemetry/specs-semconv-approvers @open-telemetry/sig-mainframe-approvers
+
+# NFS
+/docs/nfs                         @open-telemetry/semconv-system-approvers @open-telemetry/sig-mainframe-approvers
```
braydonk: For both of these, `sig-mainframe-approvers` can be removed, and `specs-semconv-approvers` can be added. `semconv-system-approvers` can stay.

braydonk (commented):
To resolve the CI issues, you'll need to run `make registry-generation` and add a `.` to the end of all briefs (this is a new validation rule that didn't exist when the PR was first written but is being caught now).

.chloggen/model-nfs.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: nfs
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `nfs.*`/`nfsd.*` metrics, along with supporting `rpc.onc.*`/`rpc.nfs.*` attributes.
```
braydonk: It seems that `nfsd.*` got changed to `nfs.server`, so we can just say `Add nfs.*` here.

braydonk (approved):

.chloggen/model-nfs.yaml
johnleslie: done

johnleslie:
@open-telemetry/semconv-db-approvers can I get a second review/approval on this?

johnleslie:
> @open-telemetry/semconv-db-approvers can I get a second review/approval on this?

@arminru @trisch-me can I get a second review/approval on this?

johnleslie:
> > @open-telemetry/semconv-db-approvers can I get a second review/approval on this?
> 
> @arminru @trisch-me can I get a second review/approval on this?

@trisch-me @dyladan @tedsuo can I get a second review/approval on this please?

braydonk:
@open-telemetry/semconv-db-approvers The changes in this PR related to DB are very minor; there's just a couple documents that have the new RPC enum added. Please take a look when you get a chance.

model/rpc/registry.yaml
```diff
@@ -274,3 +278,30 @@ groups:
         type: int
         stability: development
         brief: "Uncompressed size of the message in bytes."
+      - id: rpc.onc.version
```
lmolkova: following [naming recommendations](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#system-specific-attributes-1), this and other rpc.onc attributes should start with the `rpc.system` name, i.e.

```suggestion
      - id: onc_rpc.version
```

Other RPC conventions have not been updated yet to follow this guidance, but we're going to start RPC stabilization effort that  would take care of it later this month.

model/rpc/registry.yaml
```diff
@@ -249,6 +249,10 @@ groups:
               value: 'connect_rpc'
               brief: 'Connect RPC'
               stability: development
+            - id: onc_rpc
+              value: 'onc_rpc'
+              brief: 'ONC RPC (Sun RPC)'
```
lmolkova: It would be useful to add a link to the RFC here

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
...
+    type: metric
+    metric_name: nfs.client.net.tcp.count
+    stability: development
+    brief: "Reports the count of kernel NFS client TCP segments handled."
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.nettcpcnt
+    instrument: counter
+    unit: "{segment}"
+    entity_associations:
+      - host
```
lmolkova: is there a reason to define 3 metrics here? 

It could all be represented with `nfs.client.net.count` and `network.transport` attribute set t `tcp` or `udp`.

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
...
+    type: metric
+    metric_name: nfs.server.net.tcp.count
+    stability: development
+    brief: "Reports the count of kernel NFS server TCP segments handled."
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.nettcpcnt
+    instrument: counter
+    unit: "{segment}"
+    entity_associations:
+      - host
```
lmolkova: could these 3 metrics be merged into `nfs.server.net.count` with `network.transport` attribute to capture the transport?

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
...
+    metric_name: nfs.server.operation.count
+    stability: development
+    brief: "Reports the count of kernel NFSv4+ server operations."
+    instrument: counter
+    unit: "{operation}"
+    attributes:
+      - ref: rpc.onc.version
+      - ref: rpc.nfs.operation.name
+    entity_associations:
+      - host
```
lmolkova: the otel way to record these metrics would be to have one `nfs.server.operation.count` metric with `error.type` attribute. Successful operations would be reported without `error.type`, for unsuccessful ones, `error.type` would match `badclient`, `badfmt`, etc

model/rpc/registry.yaml
```diff
@@ -274,3 +278,30 @@ groups:
...
+      - id: rpc.onc.procedure.number
+        type: int
+        stability: development
+        brief: "ONC/Sun RPC procedure number."
+      - id: rpc.onc.procedure.name
+        type: string
+        stability: development
+        brief: "ONC/Sun RPC procedure name."
+        examples: ['OPEN', 'READ', 'GETATTR']
+      - id: rpc.onc.program.number
```
lmolkova: this attribute is not referenced on any metric - is there need to introduce it at all? We try to avoid defining attribute that are not used

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
...
+  - id: metric.nfs.client.net.count
+    type: metric
+    metric_name: nfs.client.net.count
+    stability: development
+    brief: "Reports the count of kernel NFS client network requests handled."
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.netcnt, which includes both UDP datagrams and TCP segments
+    instrument: counter
+    unit: "{request}"
+    entity_associations:
```
lmolkova: is there a way to add `server.address | port` attribute on this and other client metrics?

lmolkova: there should be a way to record if request was failed - we use `error.type` attribute for it.

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
...
+  - id: metric.nfs.client.net.tcp.connection.count
+    type: metric
+    metric_name: nfs.client.net.tcp.connection.count
+    stability: development
+    brief: "Reports the count of kernel NFS client TCP connections accepted."
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.nettcpconn
+    instrument: counter
+    unit: "{connection}"
+    entity_associations:
```
lmolkova, line 26: is there a way to report `server.*` attributes here?

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
...
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.nettcpcnt
+    instrument: counter
+    unit: "{segment}"
+    entity_associations:
+      - host
+
+  - id: metric.nfs.client.net.tcp.connection.count
+    type: metric
+    metric_name: nfs.client.net.tcp.connection.count
```
lmolkova: the connection count could be ambiguous (current vs accepted), so suggesting

```suggestion
    metric_name: nfs.client.net.tcp.connection.accepted
```

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
...
+    type: metric
+    metric_name: nfs.client.rpc.count
+    stability: development
+    brief: "Reports the count of kernel NFS client RPCs handled."
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.rpccnt
+    instrument: counter
+    unit: "{request}"
+    entity_associations:
+      - host
```
lmolkova, line 39: we should include `error.type` attribute and have means to capture the request status.

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,330 @@
...
+    type: metric
+    metric_name: nfs.server.io.write.count
+    stability: development
+    brief: "Reports the count of kernel NFS server bytes passed to write requests."
+    note: |
+      Linux: this metric is taken from the Linux kernel NFSD_STATS_IO_WRITE counter in the nfsd_net struct
+    instrument: counter
+    unit: "By"
+    entity_associations:
+      - host
```
lmolkova: could these metrics be merged and record io direction attribute similar to [`container.network.io`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/container-metrics.md#metric-containernetworkio)

lmolkova (changes requested):
It seems that metrics being introduced mirror what's already defined in `nfsd_*`, we'd define them quite differently in OTel.  E.g. we don't report each time series as its own metric and tend to merge multiple counts into one while using attributes to break them down (e,g, tcp/udp, error conditions, network direction are attributes, not parts of metric name). Is there a way we can define and report these metrics in 'otel way'?

model/nfs/metrics.yaml
braydonk: The unit would be the main reason. The unit name is distinct for all 3, and ~~there isn't really a great word to generalize it; a `{segment}` in TCP isn't really the same thing as a `{datagram}` in UDP.~~

braydonk: Reading more closely in the `/proc/net/rpc/nfsd` documentation, it seems like perhaps the unit `{packet}` could be used actually.

In that case we probably would have `network.transport` set to `tcp` or `udp` and not track the `svc_stat.netcnt` stat directly, under the assumption that aggregating the two `network.transport` attribute values would give you the full value of all packets received.

model/nfs/metrics.yaml
braydonk: This would be attached to a `host` entity. NFS is a kernel level service, so there wouldn't usually be multiple servers with different addresses on the same host. I would think the server address would be better reported as part of the entity, and the port wouldn't be particularly useful information for the timeseries.

model/nfs/metrics.yaml
KalleOlaviNiemitalo: I wonder if these count RPC "records" as in <https://www.rfc-editor.org/rfc/rfc5531#section-11>, rather than network packets.  I tried reading the Linux code that increments the counters but didn't figure it out.

`netcnt++`: <https://github.com/torvalds/linux/blob/c17b750b3ad9f45f2b6f7e6f7f4679844244f0b9/net/sunrpc/svc_xprt.c#L814>

`netudpcnt++`: <https://github.com/torvalds/linux/blob/c17b750b3ad9f45f2b6f7e6f7f4679844244f0b9/net/sunrpc/svcsock.c#L678>

`nettcpcnt++`: <https://github.com/torvalds/linux/blob/c17b750b3ad9f45f2b6f7e6f7f4679844244f0b9/net/sunrpc/svcsock.c#L1189>

johnleslie:
> It seems that metrics being introduced mirror what's already defined in `nfsd_*`, we'd define them quite differently in OTel. E.g. we don't report each time series as its own metric and tend to merge multiple counts into one while using attributes to break them down (e,g, tcp/udp, error conditions, network direction are attributes, not parts of metric name). Is there a way we can define and report these metrics in 'otel way'?

Done.

- metric.nfs.client.net.tcp.count / metric.nfs.client.net.udp.count -> metric.nfs.client.net.count + ref: network.transport
- removed old metric.nfs.client.net.count (kernel counter that simply combined tcp+udp)
- metric.nfs.server.io.read.count / metric.nfs.server.io.write.count -> network.io.direction + ref: network.io.direction
- metric.nfs.server.net.tcp.count / metric.nfs.server.net.udp.count -> metric.nfs.client.net.count + ref: network.transport

model/nfs/metrics.yaml
johnleslie: Done: metric.nfs.server.io.read.count / metric.nfs.server.io.write.count -> network.io.direction + ref: network.io.direction

model/rpc/registry.yaml
johnleslie: So I'll leave this as-is in this PR? "RPC and messaging semantic conventions don't follow the system-specific naming guidance yet, and will be updated one-by-one."

Once it's onc_rpc, hopefully in a future PR, would it go in a new model/onc_rpc/ dir?

model/rpc/registry.yaml
johnleslie: Done

model/nfs/metrics.yaml
johnleslie: Done: metric.nfs.server.net.tcp.count / metric.nfs.server.net.udp.count -> metric.nfs.client.net.count + ref: network.transport

model/nfs/metrics.yaml
johnleslie: Done

model/nfs/metrics.yaml
johnleslie: This is recorded in metric.nfs.server.rpc.bad.count which now has a error.type

model/nfs/metrics.yaml
johnleslie: we have metric.nfs.server.net.tcp.connection.count

model/nfs/metrics.yaml
johnleslie: done

model/nfs/metrics.yaml
johnleslie: This is recorded in metric.nfs.server.rpc.bad.count which now has a error.type

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,240 @@
...
+
+  - id: metric.nfs.client.operation.count
+    type: metric
+    metric_name: nfs.client.operation.count
+    stability: development
+    brief: "Reports the count of kernel NFSv4+ client operations."
+    instrument: counter
+    unit: "{procedure}"
```
braydonk: looks like a typo here
```suggestion
    unit: "{operation}"
```

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,240 @@
...
+    type: metric
+    metric_name: nfs.client.net.count
+    stability: development
+    brief: "Reports the count of kernel NFS client TCP segments and UDP datagrams handled."
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.netudpcnt and svc_stat.nettcpcnt
+    instrument: counter
+    attributes:
+      - ref: network.transport
+    unit: "{net}"
```
braydonk: I think this might be a more accurate unit
```suggestion
    unit: "{packet}"
```

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,240 @@
...
+    stability: development
+    brief: "Reports the count of kernel NFS server stale file handles."
+    note: |
+      Linux: this metric is taken from the Linux kernel NFSD_STATS_FH_STALE counter in the nfsd_net struct
+    instrument: counter
+    unit: "{fh}"
+    entity_associations:
+      - host
+
+  - id: metric.nfs.server.io.count
```
braydonk: This can just be `.io`, don't need `.count` for `By` metrics.

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,240 @@
...
+
+  - id: metric.nfs.server.net.count
+    type: metric
+    metric_name: nfs.server.net.count
+    stability: development
+    brief: "Reports the count of kernel NFS server TCP segments and UDP datagrams handled."
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.nettcpcnt and svc_stat.netudpcnt
+    instrument: counter
+    unit: "{net}"
```
braydonk: Same as above
```suggestion
    unit: "{packet}"
```

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,240 @@
...
+    stability: development
+    brief: "Reports the count of kernel NFS server TCP connections accepted."
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.nettcpconn
+    instrument: counter
+    unit: "{connection}"
+    entity_associations:
+      - host
+
+  - id: metric.nfs.server.rpc.count
```
braydonk, line 166: Does `svc_stat.rpccnt` include the bad RPCs?

If yes, leave this and `rpc.bad.count` as is.

If no, then I think this metric can receive the `error.type` dimension, where an empty `error.type` is the number of good RPCs.

model/nfs/metrics.yaml
johnleslie: Agree on record

model/nfs/metrics.yaml
johnleslie: Fixed

model/nfs/metrics.yaml
johnleslie: A single RPC record can be multiple TCP fragments/packets or a single UDP datagram/packet. Just updated this to {record} per the suggestion here: https://github.com/open-telemetry/semantic-conventions/pull/2287#discussion_r2282527909

See: https://datatracker.ietf.org/doc/html/rfc5531#section-11

model/nfs/metrics.yaml
johnleslie: A single RPC record can be multiple TCP fragments/packets or a single UDP datagram/packet. Just updated this to {record} per the suggestion here: https://github.com/open-telemetry/semantic-conventions/pull/2287#discussion_r2282527909

See: https://datatracker.ietf.org/doc/html/rfc5531#section-11

model/nfs/metrics.yaml
johnleslie: Done

model/nfs/metrics.yaml
braydonk: @KalleOlaviNiemitalo Thanks, looks like `{record}` is a sensible unit for these and that's what it's been changed to.

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,229 @@
...
+    type: metric
+    metric_name: nfs.server.rpc.count
+    stability: development
+    brief: "Reports the count of kernel NFS server RPCs handled."
+    note: |
+      Linux: this metric is taken from the Linux kernel's svc_stat.rpccnt, the count of good RPCs. this metric can have
```
braydonk:
```suggestion
      Linux: this metric is taken from the Linux kernel's svc_stat.rpccnt, the count of good RPCs. This metric can have
```

model/rpc/registry.yaml
johnleslie: Done

johnleslie:
> It seems that metrics being introduced mirror what's already defined in `nfsd_*`, we'd define them quite differently in OTel. E.g. we don't report each time series as its own metric and tend to merge multiple counts into one while using attributes to break them down (e,g, tcp/udp, error conditions, network direction are attributes, not parts of metric name). Is there a way we can define and report these metrics in 'otel way'?

Anything else before you can approve? I've addressed all of your feedback. Thanks!

johnleslie:
> > It seems that metrics being introduced mirror what's already defined in `nfsd_*`, we'd define them quite differently in OTel. E.g. we don't report each time series as its own metric and tend to merge multiple counts into one while using attributes to break them down (e,g, tcp/udp, error conditions, network direction are attributes, not parts of metric name). Is there a way we can define and report these metrics in 'otel way'?
> 
> Anything else before you can approve? I've addressed all of your feedback. Thanks!

Closing

johnleslie:
> > > It seems that metrics being introduced mirror what's already defined in `nfsd_*`, we'd define them quite differently in OTel. E.g. we don't report each time series as its own metric and tend to merge multiple counts into one while using attributes to break them down (e,g, tcp/udp, error conditions, network direction are attributes, not parts of metric name). Is there a way we can define and report these metrics in 'otel way'?
> > 
> > 
> > Anything else before you can approve? I've addressed all of your feedback. Thanks!
> 
> Closing

Friendly ping on this. Can you close this request? Trying to get this merged.

model/rpc/registry.yaml
lmolkova: > Once it's onc_rpc, hopefully in a future PR, would it go in a new model/onc_rpc/ dir?

Yes, but  is there a reason not to do it right away and this way minimize churn?

model/nfs/metrics.yaml
lmolkova: not sure I understand, this is a client metric and errors are only reported on the server one. 

If it's not possible to report an error on the client metric, let's include it into the description that this metric only reports successful RPCs

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,229 @@
...
+    type: metric
+    metric_name: nfs.server.repcache.misses
+    stability: development
+    brief: "Reports the kernel NFS server request reply cache miss count."
+    note: |
+      Linux: this metric is taken from the Linux kernel NFSD_STATS_RC_MISSES counter in the nfsd_net struct
+    instrument: counter
+    unit: "{miss}"
+    entity_associations:
+      - host
```
lmolkova: if we models a cache hit/miss metric in general, we'd probably have something like `some.cache.operation.count` with attribute `some.cache.response.status = hit | miss`.

I guess the reason to have two metrics is different units, but it makes metrics less discoverable in general and does not make it clear what's the full set of related measurements. In both cases the unit can be changed to `{request}` or  `{response}` and it'd also make it clear that a sum of them is the total number of calls to cache.

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,229 @@
...
+
+  - id: metric.nfs.server.repcache.nocache
```
lmolkova: `nfs.server.repcache.nocache` naming suggests it's one of possible `nfs.server.repcache.*`, but it does nto seem to be, it seems to be related to a response rather than a cache property. IF it's a correct assumption, the suggestion is to call it `nfs.server.reply.nocache` or similar.

model/nfs/metrics.yaml
```diff
@@ -0,0 +1,229 @@
...
+    type: metric
+    metric_name: nfs.server.repcache.nocache
+    stability: development
+    brief: "Reports the count of kernel NFS server request replies that are uncacheable."
+    note: |
+      Linux: this metric is taken from the Linux kernel NFSD_STATS_RC_NOCACHE counter in the nfsd_net struct
+    instrument: counter
+    unit: "{miss}"
```
lmolkova:
```suggestion
    unit: "{reply}"
```

?

lmolkova (commented):
Thanks for addressing most of my feedback, I still have a couple of comments.

model/nfs/metrics.yaml
johnleslie: Updated to: Reports the count of kernel NFS client RPCs sent, regardless of whether they're accepted/rejected by the server.

johnleslie: Updated the units to "{request}", and updated the note:

     note: |
       Linux: this metric is taken from the Linux kernel NFSD_STATS_RC_HITS counter in the nfsd_net struct
+      Linux: To get the total number of repcache requests, add hits, misses, and nocache.

model/nfs/metrics.yaml
johnleslie: Updated the units to "{request}", and updated the note:

     note: |
       Linux: this metric is taken from the Linux kernel NFSD_STATS_RC_HITS counter in the nfsd_net struct
+      Linux: To get the total number of repcache requests, add hits, misses, and nocache.

model/nfs/metrics.yaml
johnleslie: Done

model/rpc/registry.yaml
johnleslie: I'd like to get this merged, as it's been several months of review here and it's blocking a migration to OpenTelemetry. Would like to get to this in a future PR.

model/nfs/metrics.yaml
johnleslie: This still has to do with the reply cache (repcache). Reworded as:

    brief: "Reports the count of kernel NFS server replies that are uncacheable by the the reply cache."

johnleslie:
@jsuereth @braydonk @lmolkova looks like all comments are resolved -- can you merge?

model/rpc/registry.yaml
```diff
@@ -249,6 +249,10 @@ groups:
               value: 'connect_rpc'
               brief: 'Connect RPC'
               stability: development
+            - id: onc_rpc
+              value: 'onc_rpc'
+              brief: 'ONC RPC (Sun RPC): https://datatracker.ietf.org/doc/html/rfc5531'
```
braydonk:
```suggestion
              brief: '[ONC RPC (Sun RPC)](https://datatracker.ietf.org/doc/html/rfc5531)'
```

model/nfs/metrics.yaml
lmolkova: what I mean is that OTel-idiomatic way would be to define one metric with dimensions. E.g. `nfs.server.cache.requests` with dimension like `nfs.server.cache.state = hit | miss | nocache`. It'd match other metrics and make it easy to consume these metrics. Then explanation on the sum is not even necessary (still helpful)

model/rpc/registry.yaml
lmolkova: I'm sorry it's taking this long.  If it's merged like this, it will be changed and broken very soon. What's the benefit of merging something and breaking it immediately?

Any project that onboards on these naming would be incompatible with semconv soon and would need to be updated again. 

Change like this would probably take minutes.

model/nfs/metrics.yaml
johnleslie: I think that makes sense. Refactored as metric.nfs.server.repcache.requests with a attribute of nfs.server.repcache.status

model/rpc/registry.yaml
johnleslie: @lmolkova Any chance you could suggest the change? We now have onc_rpc as a member of rpc.system, but I don't know what you want as far as migrating rpc.onc.{version,procedure.number,procedure.name} under rpc.system.onc_rpc.

johnleslie: @lmolkova see above. "Any chance you could suggest the change? We now have onc_rpc as a member of rpc.system, but I don't know what you want as far as migrating rpc.onc.{version,procedure.number,procedure.name} under rpc.system.onc_rpc."

thompson-tomo: @johnleslie see the code suggestion. Effectively any onc_rpc specific attribute becomes onc_rpc.* ie onc_rpc.version rather than rpc.onc.*. This is what has already been done for Cassandra db for example.

lmolkova: I sent https://github.com/johnleslie/semantic-conventions/pull/1 with proposed changes. It also fixes markdown - generation failed due to metric renames - please double-check that part, I focused on making it just pass

johnleslie: Very much appreciated! Just approved https://github.com/johnleslie/semantic-conventions/pull/1

lmolkova: nice, only you can merge it @johnleslie or feel free to incorporate the changes  in some other way :)

johnleslie: merged

lmolkova (approved):

PR #2294, opened 2025-05-22
New convention: `raid`

jherbage:
Fixes #

## Changes

Attributes describing the linux only multiple device RAID arrays and their states.
Metrics to be generated by a new scraper in hostreceiver to run only on linux systems to collect metrics about these raid devices based on an existing exporter available in Prometheus' node_exporter

* [ x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
* [ x] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.
I think this is n/a - all new stuff.


model/linux/registry.yaml
```diff
@@ -17,3 +17,50 @@ groups:
         stability: development
         brief: "The Linux Slab memory state"
         examples: ["reclaimable", "unreclaimable"]
+  # mdraid device attributed
```
braydonk: If I understand correctly, `mdraid` already essentially indicates Linux Software RAID, meaning that these being in the `linux` namespace would be redundant. I think it would make more sense for `mdraid` attributes and metrics to make it into their own folder, similar to what's being done with NFS metrics in #2287.

You can make a `model/mdraid` folder, a `registry.yaml` to put these attributes in and a `metrics.yaml` file to put the below metrics in. All prefixes can be stripped from both, making `mdraid` the root namespace of both the metrics and attributes. You can then create a `docs/mdraid` and put a `mdraid-metrics.md` file there. I created this gist that you can use as a springboard by copying the `Raw` view, and changing the necessary sections to be about `mdraid` instead of `nfs`: https://gist.github.com/braydonk/93845983e1ca6429aa0b5528fb83ff65

If I'm not mistaken, a `registry/attributes/mdraid.md` file will be automatically generated from `model/mdraid/registry.yaml`.

model/system/metrics.yaml
```diff
@@ -102,7 +102,7 @@ groups:
     unit: "By"
     attributes:
       - ref: system.paging.state
-      - ref: system.device
+      - ref: linux.mdraid.device.name
```
braydonk: This seems unintentional; `system.device` is probably still the correct attribute for these.

model/system/metrics.yaml
```diff
@@ -430,3 +430,95 @@ groups:
       - ref: linux.memory.slab.state
     entity_associations:
       - host
+
+  - id: metric.system.linux.mdraid.disks
```
braydonk: Related to my above comment, these `mdraid` metrics can probably be moved to an `mdraid` model folder. This metric would be changed to just `mdraid.disks` and the same pattern would be applied to the others.

model/system/metrics.yaml
```diff
@@ -430,3 +430,95 @@ groups:
...
+    type: metric
+    metric_name: system.linux.mdraid.disks
+    stability: development
+    brief: "Number of active/failed/spare disks of device."
+    note: |
+      Reports the total number of disks in a linux RAID device.
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat).
+    instrument: gauge
```
braydonk: The following is copied from one of my comments on a similar PR.

---

OTLP makes a particular distinction between the Gauge and Counter/UpDownCounter instruments:
https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments
https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/#instrument-selection

The main difference is that UpDownCounter is to be used for metrics that track an additive measurement. If I'm not mistaken, every gauge in this set of metrics can be either a Counter/UpDownCounter and none fall into the OTLP definition for Gauge. I think they can all be changed to one or the other, where Counter is used for values that only ever increase and UpDownCounter would be for values that can decrease.

---

I see that within this PR, all metrics introduced use a `gauge` instrument, whereas I believe none of these metrics should be instrumented as a `gauge`.

As an example, this metric `mdraid.disks` is probably an `updowncounter`. It's counting something additive, and disks might be added/removed over the lifetime of the timeseries.

All of these metrics are probably `counter` or `updowncounter`. The key question will be whether it is something that only increases or is something that can always go up and down.

model/system/metrics.yaml
```diff
@@ -430,3 +430,95 @@ groups:
...
+    unit: "By"
```
braydonk: The `By` unit is "bytes", which is not the correct unit for this metric. For this metric it would likely be `{disk}`. I see that each of the metrics here use the `By` unit. I'm not sure any of these are tracking bytes, so the same adjustment will need to be made to each metrics's unit.

model/system/metrics.yaml
```diff
@@ -430,3 +430,95 @@ groups:
...
+    attributes:
+      - ref: linux.mdraid.device.name
+      - ref: linux.mdraid.device.state
+    entity_associations:
+      - host
+
+  - id: metric.system.linux.mdraid.activity_state
```
braydonk: This style of metric is a special case. I have had an open todo-list item for myself to write out the guidance for how to design these metrics correctly, I'm sorry I don't have that available. I will summarize the guidance here.

For a metric like this that is tracking the current state over an enum of possible states, it is to be a metric that reports a value of `1` if it is currently in a state, and `0` if not for all possible states. This means the unit is `1`. It is instrumented as an `updowncounter` because across other dimensions, you may want to sum up the number of things that are in a given state, making it a countable instance of something.

The template metric to refer to would be `hw.status`. You can design this metric essentially the same way: https://github.com/open-telemetry/semantic-conventions/blob/d97527f25fc8ab7dc9905451fcbf9825c9ff8f7e/model/hardware/common-metrics.yaml#L49

model/system/metrics.yaml
thompson-tomo: Why are we restricting all of these to just Linux using mdraid? Much prefer to make it generic ie metric.system.raid.disks. We could even have an attribute for type.

model/system/metrics.yaml
```diff
@@ -430,3 +430,95 @@ groups:
...
+      Reports the number of raid disks in a fully functional array for a Linux RAID device.
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat) and [md](https://docs.kernel.org/admin-guide/md.html).
+    instrument: gauge
+    unit: "By"
+    attributes:
+      - ref: linux.mdraid.device.name
+    entity_associations:
+      - host
+
+  - id: metric.system.linux.mdraid.degraded_raid_disks
```
thompson-tomo: Should this not be metric.system.raid.disks.degraded or metric.system.raid.degraded.disks

model/system/metrics.yaml
```diff
@@ -430,3 +430,95 @@ groups:
...
+      Reports the total number of blocks on a device. which are synced.
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat).
+    instrument: gauge
+    unit: "By"
+    attributes:
+      - ref: linux.mdraid.device.name
+    entity_associations:
+      - host
+
+  - id: metric.system.linux.mdraid.raid_disks
```
thompson-tomo: Should this not be metric.system.raid.disks.assigned or metric.system.raid.assigned.disks

model/linux/registry.yaml
```diff
@@ -17,3 +17,50 @@ groups:
...
+            - id: resync
+              value: 'resync'
+              stability: development
+            - id: check
+              value: 'check'
+              stability: development
+        stability: development
+        brief: "The device activity state"
+        examples: ["active"]
+      - id: linux.mdraid.device.state
```
thompson-tomo: Should this not just be system.disk.state, disk.state or storage.disk.state

model/linux/registry.yaml
```diff
@@ -17,3 +17,50 @@ groups:
...
+    type: attribute_group
+    display_name: General Linux RAID Device Attributes
+    brief: "Describes Linux RAID Device attributes"
+    attributes:
+      - id: linux.mdraid.device.name
+        type: string
+        stability: development
+        brief: "The device identifier"
+        examples: ["(identifier)"]
+      - id: linux.mdraid.device.activity_state
```
thompson-tomo: Should this not be system.disk.status, disk.status or storage.disk.status

model/linux/registry.yaml
```diff
@@ -17,3 +17,50 @@ groups:
         stability: development
         brief: "The Linux Slab memory state"
         examples: ["reclaimable", "unreclaimable"]
+  # mdraid device attributed
+  - id: registry.linux.mdraid
+    type: attribute_group
+    display_name: General Linux RAID Device Attributes
+    brief: "Describes Linux RAID Device attributes"
+    attributes:
+      - id: linux.mdraid.device.name
```
thompson-tomo: Suggestion would be to go with a name such as storage.volume.name or even just volume.name with the storage option being my preference.

thompson-tomo (commented):
Would be great if we did not restrict it to Linux.

model/linux/registry.yaml
jherbage: Made recommended changes @braydonk

model/system/metrics.yaml
jherbage: changed now

model/system/metrics.yaml
jherbage: Moved to a raid folder

model/system/metrics.yaml
jherbage: added

model/system/metrics.yaml
jherbage: changed to {disk}

model/system/metrics.yaml
jherbage: added

model/system/metrics.yaml
jherbage: changed to make it OS agnostic

model/system/metrics.yaml
jherbage: changed

model/system/metrics.yaml
jherbage: changed

model/linux/registry.yaml
jherbage: changed

model/linux/registry.yaml
jherbage: changed

model/linux/registry.yaml
jherbage: changed to recommended

model/raid/metrics.yaml
```diff
@@ -0,0 +1,99 @@
...
+    metric_name: raid.activity_state
+    stability: development
+    brief: "Reports the activity state of a RAID device"
+    note: |
+      Reports the activity state of a RAID device.
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat).
+
+      `raid.activity_state` is currently specified as an *UpDownCounter* but would ideally be represented using a
+      [*StateSet* as defined in OpenMetrics](https://github.com/prometheus/OpenMetrics/blob/v1.0.0/specification/OpenMetrics.md#stateset).
+      This semantic convention will be updated once *StateSet* is specified in OpenTelemetry. This planned change
```
braydonk: This comment implies that this is something in the works, but I haven't heard anything like this. Is there an effort to define `StateSet` in OTLP? 

It also implies that an UpDownCounter is somehow wrong for this metric, which I don't believe it is.

model/raid/metrics.yaml
```diff
@@ -0,0 +1,99 @@
+groups:
+  # raid.* metrics
+  - id: metric.raid.disks
+    type: metric
+    metric_name: raid.disks
```
braydonk: Guidance was recently merged that disallows pluralizing metric names. This should be changed to `raid.disk.count` or something of the like.

The guidance is written here: https://github.com/open-telemetry/semantic-conventions/blob/7a2ed6cb7937e9cd37a4be66ba5b4177bb203a6d/docs/general/naming.md?plain=1#L291

For the below `blocks` metrics, it's debatable whether namespaces should also not be pluralized. The reasoning for not pluralizing is largely about namespace preservation, which won't be a problem for those metrics. I've opened a thread in CNCF Slack asking about this.

model/raid/metrics.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+  - id: metric.raid.disks.assigned
+    type: metric
+    metric_name: raid.disks.assigned
+    stability: development
+    brief: "Number of raid disks in a fully functional array for a RAID device."
+    note: |
+      Reports the number of raid disks in a fully functional array for a RAID device.
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat) and [md](https://docs.kernel.org/admin-guide/md.html).
+    instrument: gauge
```
braydonk: This and `raid.disks.degraded` are sum-able, and should thus be UpDownCounters rather than Gauges.

model/raid/registry.yaml
```diff
@@ -0,0 +1,47 @@
...
+    brief: "Describes RAID Device attributes"
+    attributes:
+      - id: storage.volume.name
+        type: string
+        stability: development
+        brief: "The device identifier"
+        examples: ["(identifier)"]
+      - id: storage.disk.status
+        type:
+          members:
```
braydonk: Are the values in these enums generic or specific to raid?

model/raid/metrics.yaml
braydonk: We actually have guidance not to pluralize namespaces either: https://github.com/open-telemetry/semantic-conventions/blob/7a2ed6cb7937e9cd37a4be66ba5b4177bb203a6d/docs/general/naming.md?plain=1#L276

model/raid/metrics.yaml
jherbage: I took it from here https://github.com/open-telemetry/semantic-conventions/blob/d97527f25fc8ab7dc9905451fcbf9825c9ff8f7e/model/hardware/common-metrics.yaml#L60 @braydonk which I perhaps wrongly thought you wanted me to do based on this comment https://github.com/open-telemetry/semantic-conventions/pull/2294#discussion_r2138553296

For now I will just remove the comment as I certainly didnt intend to express an opinion either way

model/raid/registry.yaml
jherbage: The underlying values come out of linux's md raid code but they are interpreted here in the prometheus procfs code
https://github.com/prometheus/procfs/blob/master/mdstat.go#L125-L152

I took the expected values from the node exporter code the PR is based on so here
https://github.com/prometheus/node_exporter/blob/master/collector/mdadm_linux.go#L45-L74

But it looks like their exporter is out of date for the underlying code with the set of possible activity values being different. I need to look more closely at this to understand what the actual expected set of values are.

jherbage: Is it an option if its this dynamic to just not specify the enumeration in the semantics?

model/raid/metrics.yaml
braydonk: Ah right, I apologize as I'd forgotten they also state that in their definition. I don't know if that's actually based on anything, might ask them to remove that too. But yes I think you can remove it from these conventions

model/raid/registry.yaml
braydonk: This isn't the first time we've run into something like this, and I don't know if I have a good answer. My immediate opinion, based on nothing but my own taste, is that this attribute should be moved to the `raid` namespace, and the description should say that it is using activity values directly from `mdraid`, which covers the bases and should be descriptive enough to a consumer of these conventions familiar with raid.

@open-telemetry/specs-semconv-approvers @open-telemetry/specs-semconv-maintainers for opinions on this.

jherbage: Reading the code I think its OK for now at least to leave the unexposed handling of devices that are reshaping. I cant even find an example anywhere of mdstat output with reshape in it. But I am happy to be advised to enumerate with it or not enumerate at all on this state

jherbage: Hopefully I have addressed this comment in the way you intended @braydonk 
Other comments should also be addressed

model/raid/registry.yaml
```diff
@@ -0,0 +1,34 @@
...
+    brief: "Describes RAID Device attributes"
+    attributes:
+      - id: raid.disk.status
+        type: string
+        note: |
+          The set of values are pulled directly from mdraid via 
+          https://github.com/prometheus/procfs/blob/master/mdstat.go
+        stability: development
+        brief: "The raid array activity status"
+        examples: ["active"]
```
thompson-tomo: Could this be storage.volume.state? Or storage.raid.state?

model/raid/registry.yaml
```diff
@@ -0,0 +1,34 @@
+groups:
+  - id: registry.raid
```
thompson-tomo: Registry should only contain a single namespace.

model/raid/metrics.yaml
jherbage: done

model/raid/registry.yaml
```diff
@@ -0,0 +1,34 @@
...
+              stability: development
+            - id: spare
+              value: 'spare'
+              stability: development
+            - id: failed
+              value: 'failed'
+              stability: development
+        stability: development
+        brief: "The storage disk state"
+        examples: ["active"]
```
thompson-tomo: Could this leverage hw.state as it is describing the state of the hw?

model/raid/metrics.yaml
```diff
@@ -0,0 +1,93 @@
...
+    brief: "Number of degraded raid disks in an array for a RAID device."
+    note: |
+      Reports the number of degraded raid disks in an array for a RAID device..
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat) and [md](https://docs.kernel.org/admin-guide/md.html).
+    instrument: updowncounter
+    unit: "{disk}"
+    attributes:
+      - ref: storage.volume.name
+    entity_associations:
+      - host
```
thompson-tomo: Should these metrics be renamed to storage.volume.disk.{{adjective}} and introduce an attribute storage.volume.mode? Ie Raid5

model/raid/metrics.yaml
```diff
@@ -0,0 +1,93 @@
...
+    note: |
+      Reports the total number of disks in a RAID device.
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat).
+    instrument: updowncounter
+    unit: "{disk}"
+    attributes:
+      - ref: storage.volume.name
+      - ref: storage.disk.state
+    entity_associations:
+      - host
```
thompson-tomo: Should these metrics be renamed to storage.volume.disk.status and introduce an attribute storage.volume.mode? Ie raid5

model/raid/metrics.yaml
```diff
@@ -0,0 +1,93 @@
...
+      Reports the activity state of a RAID device.
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat).
+    instrument: updowncounter
+    unit: "{disk}"
+    attributes:
+      - ref: storage.volume.name
+      - ref: raid.disk.status
+    entity_associations:
+      - host
+
```
thompson-tomo: It would feel more correct if name ended in status and other part of name was storage.volume or storage.raid depending on attribute name

model/raid/metrics.yaml
```diff
@@ -0,0 +1,93 @@
...
+    brief: "Number of blocks on device."
+    note: |
+      Reports the total number of blocks on a device. which are synced.
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat).
+    instrument: updowncounter
+    unit: "{disk}"
+    attributes:
+      - ref: storage.volume.name
+    entity_associations:
+      - host
```
thompson-tomo: Should these metrics be renamed to storage.volume.block.{{adjective}} and introduce an attribute storage.volume.mode? Ie Raid5

model/raid/registry.yaml
```diff
@@ -0,0 +1,34 @@
...
+        note: |
+          The set of values are pulled directly from mdraid via 
+          https://github.com/prometheus/procfs/blob/master/mdstat.go
+        stability: development
+        brief: "The raid array activity status"
+        examples: ["active"]
+      - id: storage.volume.name
+        type: string
+        stability: development
+        brief: "The device identifier"
```
thompson-tomo: Update to "The name of the storage volume."

thompson-tomo (changes requested):

model/raid/registry.yaml
jherbage: Sorry @thompson-tomo I don't understand here. 
Am I right in thinking registry.raid is one namespace? What is the other one?
Or are you saying I can't use system and raid together in this file. 
Do I need to create a new registry for the storage attributes?

thompson-tomo: So registry.raid is a registry and raid is the namespace. This namespace should be the base namespace of the attributes and should also be the containing folder. If the state field is renamed, this folder can be renamed to storage.

model/raid/registry.yaml
jherbage: Just so I understand you right @thompson-tomo are you saying I should add a new hw.state of spare (failed and active exist already) and a new hw.type of raid or just use the existing hw.type value of "disk_controller"?

I will make those changes as that seems to make sense.

jherbage: Note I need to add a state of active as well as the current hw.states are ok, failed and degraded and only failed is used. OK is not the same as active.
And that then makes me think maybe this isnt the right attribute to use. I will make the changes and see if its what you want

model/raid/registry.yaml
jherbage: ok I can do that
Does that mean all the metrics in the namespace have to be prefixed storage as well (or move to their own raid namespace)?

model/raid/metrics.yaml
jherbage: have made an attempt to fix this

model/raid/metrics.yaml
jherbage: done

model/raid/metrics.yaml
jherbage: done

model/raid/metrics.yaml
jherbage: done

model/raid/registry.yaml
jherbage: I have done this and renamed the metrics to storage as well so raid is gone as a namespace

model/raid/registry.yaml
jherbage: done

model/raid/registry.yaml
jherbage: added this now but will leave comment open as not sure I have this right based on above

model/raid/registry.yaml
jherbage: changed to storage.raid.state but leaving comment open in case not right

model/raid/registry.yaml
thompson-tomo: This pattern of not all attributes applying is common to what we did in the hw metrics. Can close

model/storage/metrics.yaml
```diff
@@ -0,0 +1,100 @@
...
+      - ref: storage.volume.name
+      - ref: hw.state
+      - ref: hw.type
+      - ref: storage.volume.mode
+    entity_associations:
+      - host
+
+  - id: metric.storage.volume.state
+    type: metric
+    metric_name: storage.volume.state
```
thompson-tomo: Change this to status.

model/storage/metrics.yaml
```diff
@@ -0,0 +1,100 @@
+groups:
+  # storage.* metrics
+  - id: metric.storage.disk.count
+    type: metric
+    metric_name: storage.disk.count
```
thompson-tomo: Should this also be storage.volume?

docs/registry/attributes/raid.md
```diff
@@ -0,0 +1,27 @@
...
+
+---
+
+`storage.disk.state` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `active` | active | ![Development](https://img.shields.io/badge/-development-blue) |
+| `failed` | failed | ![Development](https://img.shields.io/badge/-development-blue) |
+| `spare` | spare | ![Development](https://img.shields.io/badge/-development-blue) |
```
thompson-tomo: Delete this file

thompson-tomo (changes requested):

model/raid/registry.yaml
thompson-tomo: Can close.

model/raid/metrics.yaml
thompson-tomo: Can close have raised a fresh comment on latest file. https://github.com/open-telemetry/semantic-conventions/pull/2294#discussion_r2205046358

docs/registry/attributes/raid.md
jherbage: done

model/storage/metrics.yaml
jherbage: changed

model/storage/metrics.yaml
jherbage: done

thompson-tomo (approved):

model/storage/metrics.yaml
```diff
@@ -0,0 +1,100 @@
+groups:
+  # storage.* metrics
+  - id: metric.storage.volume.disk.count
+    type: metric
+    metric_name: storage.volume.disk.count
+    stability: development
+    brief: "Number of active/failed/spare disks of device."
+    note: |
+      Reports the total number of disks in a RAID device.
```
braydonk: Should these descriptions still say `RAID` if the namespace is just `storage` now?

braydonk (approved):

model/storage/metrics.yaml
jherbage: changed to be more generic

docs/storage/metrics.md
```diff
@@ -0,0 +1,191 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Storage
+path_base_for_github_subdir:
+  from: content/en/docs/specs/semconv/http/_index.md
+  to: http/README.md
```
joaopgrassi, line 5:
```suggestion
```

The doc is also missing the main header

.chloggen/raid_metrics_in_hostreceiver.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: system
```
joaopgrassi, line 10:
```suggestion
component: storage
```

.chloggen/raid_metrics_in_hostreceiver.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Proposed metric semantics for a scraper in hostreceiver scraping raid metrics related to multiple device RAID disks
```
joaopgrassi, line 13: Maybe make this note more generic, something like: Introduce storage attributes and metrics

joaopgrassi (changes requested):

joaopgrassi:
Please also update the PR title as it's not up-to-date with the changes.

model/storage/metrics.yaml
```diff
@@ -0,0 +1,100 @@
...
+    brief: "Number of active/failed/spare disks of device."
+    note: |
+      Reports the total number of disks in a storage device.
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat).
+    instrument: updowncounter
+    unit: "{disk}"
+    attributes:
+      - ref: storage.volume.name
+      - ref: hw.state
+      - ref: hw.type
```
thompson-tomo, line 16: I don't think that we need this.

model/storage/metrics.yaml
```diff
@@ -0,0 +1,100 @@
...
+    stability: development
+    brief: "Number of active/failed/spare disks of device."
+    note: |
+      Reports the total number of disks in a storage device.
+      See also [mdadm](https://raid.wiki.kernel.org/index.php/Mdstat).
+    instrument: updowncounter
+    unit: "{disk}"
+    attributes:
+      - ref: storage.volume.name
+      - ref: hw.state
```
thompson-tomo, line 15: Am thinking we should go with disk.mode to avoid overloading hw.state.

lmolkova:
@jherbage - thanks for the contribution!

This PR adds new area - `storage` and we don't have an active SIG for it. Given that there is just one component in otel ecosystem that would currently follow this convention, I'd recommend documenting it in the corresponding collector receiver and bringing it to semconv once there is more than one instrumentation. 

I'm going to close this PR because of the above reasons. Feel free to reopen if @open-telemetry/semconv-system-approvers  would be ready to own the `storage` area and if multiple instrumentations would start following these conventions.

PR #2295, opened 2025-05-22
Revert `cpu.*` metrics to `system.cpu.*`

### LINKED ISSUE #1873: Introduce an optional normalized total CPU utilization metric

braydonk:
Related to #1873 

## Changes

In #1873 we decided that these metrics should move back to the `system` namespace to stay consistent with similar per-device-type metrics in the `system` namespace such as `disk` and `network`.


ChrsMark (approved):

bertysentry (approved):

docs/system/cpu-metrics.md
```diff
@@ -33,29 +33,7 @@ This metric is [recommended][MetricRecommended].
 
 | Name     | Instrument Type | Unit (UCUM) | Description    | Stability | Entity Associations |
 | -------- | --------------- | ----------- | -------------- | --------- | ------ |
-| `cpu.time` | Counter | `s` | Seconds each logical CPU spent on each mode | ![Development](https://img.shields.io/badge/-development-blue) |  |
-
```
lmolkova, line 37: could you please update md to generate correct metric here?

model/cpu/deprecated.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - id: metric.cpu.time
+    type: metric
+    metric_name: cpu.time
+    brief: "Deprecated. Use `system.cpu.time` instead."
+    deprecated:
+      reason: renamed
+      renamed_to: system.cpu.time
+    stability: development
+    instrument: counter
+    unit: "s"
```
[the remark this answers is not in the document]
braydonk: Fixed in [f5597bc](https://github.com/open-telemetry/semantic-conventions/pull/2295/commits/f5597bcf958db79ee40ef55c84e47c950c2bf0b4)

docs/system/cpu-metrics.md
braydonk: Since there are no more `cpu.*` metrics, I've removed this document in [f5597bc](https://github.com/open-telemetry/semantic-conventions/pull/2295/commits/f5597bcf958db79ee40ef55c84e47c950c2bf0b4)

lmolkova: do we have `system.cpu.time` metrics somewhere in markdown? It doesn't seem so

model/cpu/deprecated.yaml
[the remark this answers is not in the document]
braydonk: Done in [f5597bc](https://github.com/open-telemetry/semantic-conventions/pull/2295/commits/f5597bcf958db79ee40ef55c84e47c950c2bf0b4)

docs/system/system-metrics.md
```diff
@@ -143,6 +143,90 @@ This metric is [recommended][MetricRecommended].
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+### Metric: `system.cpu.time`
```
lmolkova, line 146: please update table of contents and then we should be good to go

lmolkova (approved):

PR #2296, opened 2025-05-22
Update recording errors doc - document when to use `error`, `exception` (or both) attributes

### LINKED ISSUE #2001: When to use `error` vs `exception` attributes
We have `error` and `exception` namespaces that describe similar things with some caveats:
- not every error has a corresponding exception (e.g. when downstream service returns 503 and span would have  `error.type = "503"` and no exception thrown
- not every exception is an error (see https://github.com/open-telemetry/opentelemetry-specification/pull/4333 for some examples)
- exception is a language/specific concept. E.g. where Java client library would throw, golang or rust would return error, but semantic conventions should be as language-agnostic  as possible.

Combining it with evolving guidance on reporting (terminal) exception details on the spans in https://github.com/open-telemetry/opentelemetry-specification/pull/4430 and https://github.com/open-telemetry/semantic-conventions/pull/1992, we'll have an overlap between exceptions and error attributes on spans and logs. 

We need to update [error-recording guidance](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md) and explain when to populate `error`, `exception` or both sets of attributes.

lmolkova:
Fixes #2001

Updates errors guidance to cover:
- errors vs exceptions and when to use which set of attributes
- reporting errors on logs/events
- reporting `error.message` attribute

### DIFF
--- a/docs/general/recording-errors.md
+++ b/docs/general/recording-errors.md
@@ -5,16 +5,24 @@
 <!-- toc -->
 
 - [What constitutes an error](#what-constitutes-an-error)
+- [Error vs exception](#error-vs-exception)
 - [Recording errors on spans](#recording-errors-on-spans)
 - [Recording errors on metrics](#recording-errors-on-metrics)
+- [Recording errors on logs and events](#recording-errors-on-logs-and-events)
 - [Recording exceptions](#recording-exceptions)
 
 <!-- tocstop -->
 
 This document provides recommendations to semantic convention and instrumentation authors
-on how to record errors on spans and metrics.
+on how to record errors on spans, metrics, and logs.
 
 Individual semantic conventions are encouraged to provide additional guidance.
+
+In the scope of this document, the terms error and exception are defined as follows:
+
+- *Error* refers to a general concept describing any non-success condition. This may include an, non-success status code, or an invalid response.
+
+- *Exception* specifically refers to runtime exceptions and their associated stack traces.
 
 ## What constitutes an error
 
@@ -39,6 +47,50 @@ An operation SHOULD be considered as failed if any of the following is true:
 Errors that were retried or handled (allowing an operation to complete gracefully) SHOULD NOT
 be recorded on spans or metrics that describe this operation.
 
+## Error vs exception
+
+When recording errors that also associated with an exception, instrumentation
+MAY record applicable [error](/docs/registry/attributes/error.md) *and*
+[exception](/docs/registry/attributes/exception.md) attributes.
+
+Semantic conventions SHOULD document the applicable set of attributes, which may
+include only `error`, only `exception`, or both, depending on the signal and
+convention.
+
+General recommendations:
+
+- Semantic conventions intended for cross-language applicability SHOULD use
+  `error.*` attributes.
+
+- `exception` attributes SHOULD be documented for conventions that target
+  exceptions directly such as [`exception` events](/docs/exceptions/exceptions-logs.md).
+
+- when the error and exception details are identical, and both sets
+  of attributes are documented by the convention, it's RECOMMENDED to
+  populate `error.type` and `error.message` (when applicable) instead
+  of the corresponding `exception.*` attributes.
+
+  **Example:**
+
+  The instrumentation catches network-related exception that does
+  not include any structured information about the failure. The instrumentation
+  should set `error.type` to the fully qualified exception type and omit
+  `exception.type`.
+
+- when the error and exception details differ, and both sets of
+  attributes are documented by the convention, it's RECOMMENDED to
+  populate the `error` attributes along with any `exception`
+  attributes that differ from the corresponding `error` ones.
+
+  **Example:**
+
+  The instrumentation catches a network-related exception that exposes a status
+  code.
+  The instrumentation should set `error.type` to the status code provided in the
+  exception. If the corresponding convention documents `exception` attributes,
+  the instrumentation also sets `exception.type` to the fully qualified
+  exception type.
+
 ## Recording errors on spans
 
 [Span Status Code][SpanStatus] MUST be left unset if the instrumented operation has
@@ -52,10 +104,12 @@ When the operation ends with an error, instrumentation:
   about the error which is not expected to contain sensitive details and aligns
   with [Span Status Description][SpanStatus] definition.
 
-  It's NOT RECOMMENDED to duplicate status code or `error.type` in span status description.
+  It's NOT RECOMMENDED to duplicate status code, `error.type`, `error.message`,
+  or `exception.message` in the span status description.
 
-  When the operation fails with an exception, the span status description SHOULD be set to
-  the exception message.
+  When an operation fails due to an exception, the span status description
+  SHOULD be set to the exception message, unless a more precise description is
+  available through other means.
 
 Refer to the [recording exceptions](#recording-exceptions) on capturing exception
 details.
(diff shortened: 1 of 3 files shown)


docs/general/recording-errors.md
```diff
@@ -5,17 +5,25 @@
...
-on how to record errors on spans and metrics.
+on how to record errors on spans, metrics, and logs.
 
 Individual semantic conventions are encouraged to provide additional guidance.
 
+In the scope of this document, the terms error and exception are defined as follows:
+
+- *Error* refers to a general concept describing any non-success condition. This may include an, non-success status code, or an invalid response.
+
+- *Exception* specifically refers to runtime exceptions and their associated stack traces.
```
braydonk, line 25: How does this mesh with languages that don't have "exceptions" and actually only have "errors" like Go or Rust?

docs/general/recording-errors.md
```diff
@@ -5,17 +5,25 @@
...
 
 This document provides recommendations to semantic convention and instrumentation authors
-on how to record errors on spans and metrics.
+on how to record errors on spans, metrics, and logs.
 
 Individual semantic conventions are encouraged to provide additional guidance.
 
+In the scope of this document, the terms error and exception are defined as follows:
+
+- *Error* refers to a general concept describing any non-success condition. This may include an, non-success status code, or an invalid response.
```
braydonk, line 23:
```suggestion
- *Error* refers to a general concept describing any non-success condition. This may include a non-success status code, or an invalid response.
```

docs/general/recording-errors.md
lmolkova: I refactored it slightly, please take a look, the key change:

added 

> Errors and exceptions are related but not the same: an error can occur without
an exception being thrown, and an exception does not necessarily constitute an
error.
>
> Exceptions and how they are recorded in telemetry are inherently
language-specific. Some languages, such as Rust or Go, do not use exceptions
at all.
>
> This document focuses on recording errors.


Plus there are a couple of points below that should address it

> - Semantic conventions intended for cross-language applicability SHOULD use
>   `error.*` attributes.
> 
> - `exception` attributes SHOULD be documented for conventions that target
>   exceptions directly such as [`exception` events](/docs/exceptions/exceptions-logs.md).

smaddock:
My 2¢: the “error vs exception” documentation is still too language-specific in its assumptions of “exceptions.” I’d suggest reframing it altogether:

Errors:
- “Error” indicates the **state** of what is being signaled. E.g., Span Status, Log Record Severity.
  - Note that the existance of an error code in a system does not necessarily mean the system being instrumented is in an error state. Include 404 example from PR.
- Additional error-domain attributes should be assigned per SemConv to provide additional information on the nature of that error state.
- Include much of the existing “recording errors on…” doc from this PR.

Exceptions:
- “Exception” as a general concept refers to an **event** where an exceptional condition occurred, i.e., some predefined condition was not met.
  - (Exceptions can occur at any layer of processing; in production environments, they are most often encountered while code is being executed, where they are referred to as “runtime exceptions.”)
- Programming languages may have specific mechanisms for defining the conditions, and for throwing and handling the events. Languages may also have different names for these conditions and events, and different definitions of what is a language-specific “exception”.
- Exceptions may be indicative of an error state, or may simply be a logic function employed in the code — determining how and whether to generate signals from exceptions, or add exception information to existing signals via attributes, requires an understanding of the system being instrumented and the nature of the exceptions encoutered.

SemConv Changes:
- Deprecate the existing exception domain and its attributes.
  - If an exception is indicative of an error state, the appropriate error attributes should be used instead.
  - Otherwise the instrumentation user should evalutate what is semantically being indicated by the exception and what needs to be observed. Suggest log events as a generic alternative.
  - The code-domain stacktrace attribute should be used any time a stacktrace needs to be added to a signal (except profiles).
- Language SIGs should provide examples of signaling errors and exceptional events, and clarify how their language’s “error” and/or “exception” constructs apply.

lmolkova:
@smaddock could you please leave specific suggestions on the content modified by this PR?

WRT 

> Deprecate the existing exception domain and its attributes

There is no active discussion on this - I'd suggest to start with an issue in the https://github.com/open-telemetry/opentelemetry-specification. You also probably want to read https://github.com/open-telemetry/opentelemetry-specification/pull/4333

docs/general/recording-errors.md
```diff
@@ -4,18 +4,82 @@
...
+Semantic conventions SHOULD document the applicable set of attributes, which may
+include only `error`, only `exception`, or both, depending on the signal and
+convention.
+
+General recommendations:
+
+- Semantic conventions intended for cross-language applicability SHOULD use
+  `error.*` attributes.
+
+- `exception` attributes SHOULD be documented for conventions that target
```
lmolkova, line 54: this maybe a breaking(?) change for go/rust/no-exception languages

jsuereth:
@lmolkova Is this PR still active?

docs/general/recording-errors.md
```diff
@@ -52,10 +116,12 @@ When the operation ends with an error, instrumentation:
   about the error which is not expected to contain sensitive details and aligns
   with [Span Status Description][SpanStatus] definition.
 
-  It's NOT RECOMMENDED to duplicate status code or `error.type` in span status description.
+  It's NOT RECOMMENDED to duplicate status code, `error.type`, `error.message`,
+  or `exception.message` in the span status description.
```
KalleOlaviNiemitalo, line 120: The note on the `error.message` attribute says it is NOT RECOMMENDED for spans.  This text about duplicating `error.message` in the span status description may give the false impression that the attribute is preferred.

docs/general/recording-errors.md
jsuereth: So for RUST at least - I'd tie the notion of "exception" to the presence of a [backtrace](https://doc.rust-lang.org/std/backtrace/index.html) on the error type.

Specifically, we want:

- In Rust, if you are signalling any type `T` which implements `std::error::Error` with a backtrace and are logging it, you fill out `exception.stacktrace` with the `Backtrace` from Rust.  
- Following [this guidance for accessing members by type on Error](https://doc.rust-lang.org/std/error/trait.Error.html#example-1) with a convention of attaching `Backtrace` as a field in Error.
- Ensuring this guidance works with `anyhow` crate.

If I read this guidance correctly, this means Rust would do the following:

- Always write `error.type` (the rust error type)
- Always write `error.message` (using `description` of the rust error)
- Optionally write `exception.stacktrace` (by accessing `Backtrace` field on `Error` when available.
- Optional write `exception.type` to match `error.type` if a Backtrace exists?

Can you confirm?

Also would be good for @open-telemetry/rust-maintainers to confirm my assumptions on error handling here.

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs#L83-L93
This is what we do today (only error.message)

docs/general/recording-errors.md
```diff
@@ -4,18 +4,82 @@
...
+- *Exception* specifically refers to runtime exceptions and their
+  associated stack traces.
+
+Errors and exceptions are related but not the same: an error can occur without
+an exception being thrown, and an exception does not necessarily constitute an
+error.
+
+Exceptions and how they are recorded in telemetry are inherently
+language-specific. Some languages, such as Rust or Go, do not use exceptions
+at all.
```
pellared, line 37: I think that a panic in Go be considered as an exception.
Then a deffered `span.End` can handle "the exception" (recover the panic and record it).
See: https://github.com/open-telemetry/opentelemetry-go/blob/b5b6989c7a08b5e4773236fc8d9fa0e5fe9788e1/sdk/trace/span.go#L476-L493

I think it would be good to clarify it.

docs/general/recording-errors.md
```diff
@@ -4,18 +4,82 @@
...
 
+## Error vs exception
+
+In the scope of this document, the terms error and exception are defined as follows:
+
+- *Error* refers to a general concept describing any non-success condition.
+  This may include a non-success status code, or an invalid response.
+
+- *Exception* specifically refers to runtime exceptions and their
+  associated stack traces.
```
pellared, line 29: @open-telemetry/go-maintainers,

While a personally like this distinction (and I find it correct), I am not sure if/how we are going to make it in Go in a non-breaking way given https://pkg.go.dev/go.opentelemetry.io/otel/trace#Span:

```go
	// RecordError will record err as an exception span event for this span. An
	// additional call to SetStatus is required if the Status of the Span should
	// be set to Error, as this method does not change the Span status. If this
	// span is not being recorded or err is nil then this method does nothing.
	RecordError(err error), options ...EventOption)
```

Maybe deprecate this function and create a new one that would record errors according to the new semantics?

Related issues:
- https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7254
- https://github.com/open-telemetry/opentelemetry-go/issues/6903
- https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7470

lmolkova:
Drafting this one and blocking on https://github.com/open-telemetry/opentelemetry-specification/pull/4333

lmolkova:
related: https://github.com/open-telemetry/opentelemetry-specification/issues/3496#issuecomment-3366843990

PR #2307, opened 2025-05-28
Automate schema-next generation in release workflow

### LINKED ISSUE #1426: Automate schema-next generation

lmolkova:
Fixes #1426

This PR automates schema-next generation. Proposed process changes:

- schema-next.yaml is deleted, PR authors should not update it
-  reviewers should review `deprecated` property - it has to be structured and should use proper reason. Only `renamed` on attributes and metrics is included into generated `schema-next`.
- schemas/$version is generated only at the release time using `weaver diff`


internal/tools/scripts/generate-schema-next.sh
```diff
@@ -0,0 +1,40 @@
...
+fi
+
+if ! grep -q "^versions:$" ${prev_version_schema_file}; then
+  echo "String 'versions:' not found in the file ${prev_version_schema_file}."
+  exit 1
+fi
+
+cp ${prev_version_schema_file} ${cur_version_schema_file}
+
+sed -i "s/^schema_url: .*/schema_url: https:\/\/opentelemetry.io\/schemas\/$cur_version/" ${cur_version_schema_file}
```
jsuereth: Is this something weaver could generate?

internal/tools/scripts/schema-diff/yaml/schema-diff.j2
```diff
@@ -0,0 +1,18 @@
+{{- template.set_file_name("schema-diff.yaml") }}
```
jsuereth, line 1: For clarity -

We're no longer doing renames *within* a metric or splits at this point right?

Previously defined diffs are preserved and we just don't allow those changes going forward?

jsuereth (approved):
Want to clarify some limitations, but this looks like a huge improvement for us.

internal/tools/scripts/generate-schema-next.sh
lmolkova: in theory yes, in practice - https://github.com/open-telemetry/weaver/issues/775

made a few changes to get closer to more things generated by weaver

internal/tools/scripts/schema-diff/yaml/schema-diff.j2
lmolkova: weaver.yaml filters changes to `renames` only. During code review we have to ensure things like scoped renames, splits or other not supported transformation are added as `uncategorized` instead of `renamed`. This is not new - we have to check this today as well.
 
So the short answer is yes,  this is only for proper renames.

internal/tools/scripts/schema-diff/yaml/schema-diff.j2
```diff
@@ -0,0 +1,18 @@
...
+{% for attr in ctx.registry_attributes %}
+              {{attr.old_name}}: {{attr.new_name}}
+{% endfor %}
+{% endif %}
+{% if ctx.metrics | length > 0 %}
+    metrics:
+      changes:
+        - rename_metrics:
+{% for metric in ctx.metrics %}
+              {{metric.old_name}}: {{metric.new_name}}
```
lmolkova: apparently this change is blocked on https://github.com/open-telemetry/weaver/issues/776 - old_name is not a name, but id

lmolkova:
blocked on https://github.com/open-telemetry/weaver/issues/776

internal/tools/scripts/schema-diff/yaml/schema-diff.j2
lmolkova: fixed in weaver 0.15.3

lmolkova:
Now blocked on https://github.com/open-telemetry/semantic-conventions/pull/2444 (weaver version update)

PR #2310, opened 2025-05-30
Detect yaml groups that are not used in md files, deprecate `otel.scope` entity

lmolkova:
There's been a few occurrences where we didn't notice yaml groups that are not in MD during code review or were close to not noticing. This PR automates detection of yaml groups that are not used in md files.

We might take a different approach once we have signal registries - https://github.com/open-telemetry/semantic-conventions/pull/2146, this is a quick patch and I won't regret us removing it in the future.

In the meantime - signals that only exist in yaml are not going to be shown on otel.io and won't exist for anyone browsing md docs here. We should try to avoid it.


internal/tools/scripts/find-dead-yaml.sh
```diff
@@ -0,0 +1,48 @@
...
+  exit 1
+fi
+
+declare -A semconv_snippets
+# Extract only lines of the form <!-- semconv ... --> from .md files
+echo "Indexing semconv lines in .md files under: $docs_folder ..."
+while IFS= read -r LINE; do
+  semconv_snippets["$LINE"]=1
+done < <(
+  grep -rhoP '<!--\s*semconv\s+\K[a-z0-9._]+(?=\s*-->)' "$docs_folder" --include="*.md"
```
lmolkova, line 27: all the credit goes to AI — my bash skills are nowhere near that level

docs/resource/otel.md
```diff
@@ -0,0 +1,24 @@
...
+
+<!-- semconv entity.otel.scope -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
```
trisch-me: so these 2 files are examples of missing docs for entities app and otel?

docs/resource/otel.md
```diff
@@ -0,0 +1,24 @@
...
+
+**type:** `otel.scope`
+
+**Description:** Attributes used by non-OTLP exporters to represent OpenTelemetry Scope's concepts.
```
jsuereth: So this isn't a *real* resource.  This is a set of attributes reserved for how you report OpenTelemetry's "InstrumentationScope" in non OpenTelemetry systems.

I think this one isn't modelled well and we should have an exemption for it rather than forcing it into the markdown here.  We actually *don't* want resources having these attributes.

docs/resource/otel.md
lmolkova: these two and metric `system.cpu.frequency`

docs/resource/otel.md
lmolkova: I deprecated corresponding entity, attributes are still there

docs/resource/app.md
```diff
@@ -0,0 +1,42 @@
+# App
+
+<!-- semconv entity.app -->
```
lmolkova: this won't be necessary once https://github.com/open-telemetry/semantic-conventions/pull/2278 is merged, so I'm going to remove entities from the check and delete this file

Issue #2306, opened 2025-05-28
Clarify `.count` guidance for UpDownCounters


ChrsMark:
The [Use count instead of pluralization for UpDownCounters](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#use-count-instead-of-pluralization-for-updowncounters) guidance is not clear right now.

We need to clarify/decide if the guidance applies only when there is an "overlap" between the namespace and the metric name. For example since `system.process.*` is an already existing namespace then we should use `system.process.count` and not `system.processes`.
 
The guidance does not enforce to use `.count` for all `UpDownCounter`s  and metric like [`k8s.deployment.desired_pods`](https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/metrics.yaml#L131-L142) is fine if there is no need for `k8s.deployment.desired_pod.*` namespace.

If that's the case we can either remove the guidance or just try to clarify its content. 

ref: https://github.com/open-telemetry/semantic-conventions/issues/1800#issuecomment-2915721562

Related to:
-  https://github.com/open-telemetry/semantic-conventions/pull/2191#discussion_r2107549865
- https://github.com/open-telemetry/semantic-conventions/issues/1999
- https://github.com/open-telemetry/semantic-conventions/issues/2301

Issue #1800, opened 2025-01-23
`system.network.connections` wrong pluralization


braydonk:
### What's missing?

`system.network.connections` is an `updowncounter`, and as such should be named `system.network.connection.count`.

ChrsMark:
I find the guidance to [Use count instead of pluralization for UpDownCounters](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#use-count-instead-of-pluralization-for-updowncounters) a bit unclear 🤔 .

Should we use count for all metrics that are of `updowncounter` type? The specific paragraph implies that there should a namespace for the entity but here we don't have a `system.network.connection.*` namespace.

braydonk:
Yeah you're right, the guidance could be clearer here.

Quoting from the linked guidance:
```
If the value being recorded represents the count of concepts signified by the namespace then the metric should be named count
```
This is where the ambiguity comes in for sure; `system.network.connection` isn't necessarily a namespace insofar as we have no other metrics stemming from `system.network.connection`. We could interpret it as such and say it's a namespace with only one metric for now being `.count` if we think there might be others added. I can't currently think of any more metrics we might add, but perhaps it's still worth taking the `.count` route so that we're not locked out of any other `system.network.connection.*` metrics ever being added?

Something the guidance should make a bit more clear is the reasoning for a SHOULD on `.count` with `UpDownCounters` in relation to using pluralization. If the goal is to surface clearly within the name when something is `UpDownCounter` vs `Counter` then that's fine but should probably be stated explicitly. This is how I've been treating it, but litigating the guidance more closely does reveal some ambiguity.

Tagging in @open-telemetry/specs-semconv-maintainers to help resolve this.

ChrsMark:
So based on the discussion from [Monday's May 26 2025 meeting](https://docs.google.com/document/d/10xG7DNKWRhxNmFGt3yYd3980a9uwS8lMl2LvQL3VNK8/edit?tab=t.0#heading=h.ylazl6464n0c) it seems that all metrics of type `UpDownCounter` should be suffixed with `.count`, right? 

Latest comments on https://github.com/open-telemetry/semantic-conventions/issues/1999 indicate that but we would potentially need to make it even more explicit in the guidance?

@braydonk @lmolkova please check me on this. Pinging @trask as well because of https://github.com/open-telemetry/semantic-conventions/issues/1999 :).

braydonk:
[...]

PR #2317, opened 2025-06-02
[chore] Clarify .count for updowncounters guidance

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/2306

## Changes

This PR tries to clarfiy the guidance regarding the usage of `.count` for `UpDownCounter`s as it was described at https://github.com/open-telemetry/semantic-conventions/issues/2306.

/cc @trask @lmolkova @braydonk 

### DIFF
--- a/docs/general/naming.md
+++ b/docs/general/naming.md
@@ -290,18 +290,17 @@ Examples:
 
 #### Do not pluralize UpDownCounter names
 
-If the value being recorded represents the count of concepts signified
-by the namespace then the metric SHOULD be named `count` within that namespace.
+UpDownCounter names MUST NOT be pluralized.
+
+If the value being recorded represents a count of the concepts defined by the namespace,
+a suffix such as `count`, `active`, etc. SHOULD be used within that namespace.
 
 For example if we have a namespace `system.process` which contains all metrics related
 to the processes then to represent the count of the processes we can have a metric named
 `system.process.count` instead of `system.processes`.
 
-It's not always necessary to use `count`.
-For example, a suffix like `active` should be enough when the metric represents the number of concepts
-in `active` state, i.e. `cicd.pipeline.run.active`.
-
-What is required, is to not pluralize UpDownCounter metric names.
+Similarly, a suffix like `active` SHOULD be used when the metric represents the number of
+concepts in an active state — for example, `cicd.pipeline.run.active`.
 
 #### Do not use `total`
 
--- a/docs/general/naming.md
+++ b/docs/general/naming.md
@@ -24,7 +24,7 @@ aliases: [attribute-naming]
 - [Metrics](#metrics)
   - [Naming rules for counters and UpDownCounters](#naming-rules-for-counters-and-updowncounters)
     - [Pluralization](#pluralization)
-    - [Use `count` instead of pluralization for UpDownCounters](#use-count-instead-of-pluralization-for-updowncounters)
+    - [Use `count` instead of pluralization for UpDownCounters (when a namespace exists)](#use-count-instead-of-pluralization-for-updowncounters-when-a-namespace-exists)
     - [Do not use `total`](#do-not-use-total)
   - [Instrument naming](#instrument-naming)
   - [Client and server metrics](#client-and-server-metrics)
@@ -288,14 +288,18 @@ Examples:
 - `system.paging.faults`, `system.disk.operations`, and `system.network.packets`
   should be pluralized, even if only a single data point is recorded.
 
-#### Use `count` instead of pluralization for UpDownCounters
+#### Use `count` instead of pluralization for UpDownCounters (when a namespace exists)
 
 If the value being recorded represents the count of concepts signified
-by the namespace then the metric should be named `count` (within its namespace).
+by the namespace then the metric SHOULD be named `count` within that namespace.
 
 For example if we have a namespace `system.process` which contains all metrics related
 to the processes then to represent the count of the processes we can have a metric named
 `system.process.count`.
+
+Using `count` is not required for UpDownCounters when there is no meaningful namespace to preserve.
+For instance, `k8s.deployment.desired_pods` is acceptable, as there is no need to use singular
+`k8s.deployment.desired_pod.*` as a namespace.
 
 #### Do not use `total`
 
--- a/docs/general/naming.md
+++ b/docs/general/naming.md
@@ -24,7 +24,7 @@ aliases: [attribute-naming]
 - [Metrics](#metrics)
   - [Naming rules for counters and UpDownCounters](#naming-rules-for-counters-and-updowncounters)
     - [Pluralization](#pluralization)
-    - [Use `count` instead of pluralization for UpDownCounters (when a namespace exists)](#use-count-instead-of-pluralization-for-updowncounters-when-a-namespace-exists)
+    - [Do not pluralize UpDownCounter names](#do-not-pluralize-updowncounter-names)
     - [Do not use `total`](#do-not-use-total)
   - [Instrument naming](#instrument-naming)
   - [Client and server metrics](#client-and-server-metrics)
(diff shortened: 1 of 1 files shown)


joaopgrassi (dismissed):

braydonk (approved):

mx-psi (approved):

rogercoll (approved):

trask (approved):
I've added this to Mon SIG meeting agenda just to confirm with more folks, thanks

docs/general/naming.md
```diff
@@ -285,15 +285,19 @@ should not be pluralized, even if many data points are recorded.
...
-by the namespace then the metric should be named `count` (within its namespace).
+by the namespace then the metric SHOULD be named `count` within that namespace.
 
 For example if we have a namespace `system.process` which contains all metrics related
 to the processes then to represent the count of the processes we can have a metric named
 `system.process.count`.
 
+Using `count` is not required for UpDownCounters when there is no meaningful namespace to preserve.
+For instance, `k8s.deployment.desired_pods` is acceptable, as there is no need to use singular
+`k8s.deployment.desired_pod.*` as a namespace.
```
lmolkova: wait, we've been actively aligning updowncounters to follow `k8s.deployment.desired_pod.count` pattern, why would we change direction now? it'd be even more confusing than today

ChrsMark: This comes from the discussion at https://github.com/open-telemetry/semantic-conventions/issues/1800#issuecomment-2913515338.

If we still want to enforce `.count` to all `updowncounter`s then we should change the guidance accordingly to make this clear. However, I doubt that's sth we want to do now if what @trask shared at https://github.com/open-telemetry/semantic-conventions/issues/1800#issuecomment-2913515338 was the original motivation.

In any case, whatever we decide on we need to fix the guidance accordingly to avoid confusion.

ps: I won't be able to join the Mon SIG meeting but please share any decision here so I can update the PR accordingly :)

lmolkova: I would be happy to completely drop this part of guidance 2 years ago, but we've been actively pushing everyone to follow it and implemented it in some stable conventions. 

Changing or removing it now would mean some chaos and more inconsistencies to come. I'll try to prepare a list of metrics so we can evaluate how many irreversible decisions we've made so far. 

To some extent it's a matter of taste and not a technical/rational decision to call something `foo.bar.active_requests` or `foo.bar.active_request.count` - it's like tab vs space - make decision once, add lint, and everyone should follow the repo policy. Having mixture is the worst situation to be in.

lmolkova: stable plural metrics violating the current no-pluralization guidance (8, self-contained within .NET):

- aspnetcore.rate_limiting.active_request_leases
- aspnetcore.rate_limiting.queued_requests
- kestrel.active_connections
- kestrel.active_tls_handshakes
- kestrel.queued_connections
- kestrel.queued_requests
- kestrel.upgraded_connections
- signalr.server.active_connections

development metrics violating the current guidance (29, 24 of them are k8s):

- db.client.connection.pending_requests
- http.client.active_requests
- http.client.open_connections
- http.server.active_requests
- k8s.cronjob.active_jobs
- k8s.daemonset.current_scheduled_nodes
- k8s.daemonset.desired_scheduled_nodes
- k8s.daemonset.misscheduled_nodes
- k8s.daemonset.ready_nodes
- k8s.deployment.available_pods
- k8s.deployment.desired_pods
- k8s.hpa.current_pods
- k8s.hpa.desired_pods
- k8s.hpa.max_pods
- k8s.hpa.min_pods
- k8s.job.active_pods
- k8s.job.desired_successful_pods
- k8s.job.failed_pods
- k8s.job.max_parallel_pods
- k8s.job.successful_pods
- k8s.replicaset.available_pods
- k8s.replicaset.desired_pods
- k8s.replicationcontroller.available_pods
- k8s.replicationcontroller.desired_pods
- k8s.statefulset.current_pods
- k8s.statefulset.desired_pods
- k8s.statefulset.ready_pods
- k8s.statefulset.updated_pods
- system.network.connections

metrics following current guidance:
- 55 development
- 16 stable

--------------------
I'd prefer the outcome that is 

1. minimally disruptive 
2. does not make things more confusing

I have two ideas:

1.  Just say to avoid pluralization (just because it's a common practice).  
     ```diff
    - Use count instead of pluralization for UpDownCounters
    + Do not pluralize UpDownCounter names
    ```

2. Just remove this part from the guidance. If pluralization is sometimes ok and sometimes not, just pick whatever name works, we don't need guidance.

ChrsMark:
@lmolkova @joaopgrassi @trask, should we then keep the guidance and make it more explicit with sth like `all updowncounters SHOULD be named with .count instead of using pluralisation`?

lmolkova:
@ChrsMark we have a bunch of metrics that use `.active` or some more suitable alternative to `.count`, so I was thinking about 

```diff
- Use count instead of pluralization for UpDownCounters
+ Do not pluralize UpDownCounter names
```

with some examples using `.count` and others and explanation that it's a matter of consistency across conventions.

ChrsMark:
> @ChrsMark we have a bunch of metrics that use `.active` or some more suitable alternative to `.count`, so I was thinking about
> 
> ```diff
> - Use count instead of pluralization for UpDownCounters
> + Do not pluralize UpDownCounter names
> ```
> 
> with some examples using `.count` and others and explanation that it's a matter of consistency across conventions.

Ok, I'm fine with that. I updated the PR accordingly. If there are suggestions about the wording or more examples to add let me know.

docs/general/naming.md
```diff
@@ -288,14 +288,20 @@ Examples:
...
 If the value being recorded represents the count of concepts signified
-by the namespace then the metric should be named `count` (within its namespace).
+by the namespace then the metric SHOULD be named `count` within that namespace.
 
 For example if we have a namespace `system.process` which contains all metrics related
 to the processes then to represent the count of the processes we can have a metric named
-`system.process.count`.
+`system.process.count` instead of `system.processes`.
+
+It's not always necessary to use `count`.
```
braydonk: It feels a bit strange to say a normative `SHOULD` above when we say it's not always necessary down here.

Maybe the above section should say that `UpDownCounter` metric names `MUST NOT` be pluralized, and use `count` as the most common example but then leave this section below suggesting some alternative scenarios.

ChrsMark: Good point! Changed that: [48cb2e9](https://github.com/open-telemetry/semantic-conventions/pull/2317/commits/48cb2e9569bd1a0f32b3207f8d3aef96d057b111)

docs/general/naming.md
```diff
@@ -288,14 +288,19 @@ Examples:
 - `system.paging.faults`, `system.disk.operations`, and `system.network.packets`
   should be pluralized, even if only a single data point is recorded.
 
-#### Use `count` instead of pluralization for UpDownCounters
+#### Do not pluralize UpDownCounter names
 
-If the value being recorded represents the count of concepts signified
-by the namespace then the metric should be named `count` (within its namespace).
+UpDownCounter names MUST NOT be pluralized.
```
lmolkova:
```suggestion
UpDownCounter names SHOULD NOT be pluralized.
```

we already have some precedents on stable metrics, so we can't be too strict

docs/general/naming.md
```diff
@@ -288,14 +288,19 @@ Examples:
...
+
+If the value being recorded represents a count of the concepts defined by the namespace,
+a suffix such as `count`, `active`, etc. SHOULD be used within that namespace.
```
lmolkova: do we really need this guidance? we have `.count`, `.active`, `.used`, `.used_after_last_gc`, `.ready`, `.queued`, `.processed`, ....

I don't think we need to be prescriptive on which adjective to use. 

```suggestion
```

docs/general/naming.md
```diff
@@ -288,14 +288,19 @@ Examples:
...
 
 For example if we have a namespace `system.process` which contains all metrics related
 to the processes then to represent the count of the processes we can have a metric named
-`system.process.count`.
+`system.process.count` instead of `system.processes`.
+
+Similarly, a suffix like `active` SHOULD be used when the metric represents the number of
+concepts in an active state — for example, `cicd.pipeline.run.active`.
```
lmolkova:
```suggestion
```

same here - I don't see why we need a guidance on `active` - if `inflight` or `current` works better, why would someone use `active`?

docs/general/naming.md
```diff
@@ -288,14 +288,19 @@ Examples:
...
-by the namespace then the metric should be named `count` (within its namespace).
+UpDownCounter names MUST NOT be pluralized.
+
+If the value being recorded represents a count of the concepts defined by the namespace,
+a suffix such as `count`, `active`, etc. SHOULD be used within that namespace.
 
 For example if we have a namespace `system.process` which contains all metrics related
 to the processes then to represent the count of the processes we can have a metric named
-`system.process.count`.
+`system.process.count` instead of `system.processes`.
```
lmolkova:
```suggestion
`system.process.count` instead of `system.processes`. Similarly, `cicd.pipeline.run.active` is preferred over the `cicd.pipeline.active_runs`.
```

ChrsMark:
Thank's @lmolkova , your comments should be addressed now, please take another look.

joaopgrassi:
@ChrsMark I think all comments are solved now, so I resolved them. I think we are good to merge this now. CC @lmolkova

PR #2319, opened 2025-06-03
Add k8s Pod volume metrics

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1485

## Changes

This PR adds `k8s.pod.volume.*` metrics:

```
k8s.pod.volume.available
k8s.pod.volume.capacity
k8s.pod.volume.usage
k8s.pod.volume.inode.count
k8s.pod.volume.inode.used # This may not be equal to `inode.count - inode.free` because filesystem may share inodes with other filesystems.
k8s.pod.volume.inode.free
```

Collector ref: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.127.0/receiver/kubeletstatsreceiver/documentation.md#k8svolumeavailable
Collector implementation: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.127.0/receiver/kubeletstatsreceiver/internal/kubelet/volume.go#L16

These metrics are derived from the [VolumeStats](https://pkg.go.dev/k8s.io/kubelet@v0.33.0/pkg/apis/stats/v1alpha1#VolumeStats) field of the [PodStats](https://pkg.go.dev/k8s.io/kubelet@v0.33.0/pkg/apis/stats/v1alpha1#PodStats) of the Kubelet's stats API.

Note: The current implementation reports them as `k8s.volume.*` but I think they should be `k8s.pod.volume.*` to better indicate the scope since there is no concept a K8s Volume outside of the Pod context.

Attributes for volume types that are deprecated are not added here: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40477


dashpole:
Do you think it would make sense to consistently use `used` vs `capacity` everywhere, and omit free/available?

ChrsMark:
> Do you think it would make sense to consistently use `used` vs `capacity` everywhere, and omit free/available?

If I understand the suggestion correctly, it seems that free/available would still be useful/needed?

From https://pkg.go.dev/k8s.io/kubelet@v0.33.0/pkg/apis/stats/v1alpha1#FsStats:

- `UsedBytes`: This may differ from the total bytes used on the filesystem and may not equal CapacityBytes - AvailableBytes.
- `InodesUsed` : This may not equal Inodes - InodesFree because this filesystem may share inodes with other "filesystems"

dashpole:
ah, right

ChrsMark:
However, based on this I think it makes sense to record also the `UsedBytes` as well. I will update the PR to add this.

odubajDT:
Would it be possible to add a `k8s.volume.used`metric? Details here https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40476

ChrsMark:
> Would it be possible to add a `k8s.volume.used`metric? Details here [open-telemetry/opentelemetry-collector-contrib#40476](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40476)

It's done: [5459577](https://github.com/open-telemetry/semantic-conventions/pull/2319/commits/54595774e0f278e0f7be29e9db56a756cbc15e49)

model/k8s/metrics.yaml
```diff
@@ -63,6 +63,112 @@ groups:
...
       - ref: network.io.direction
 
+  # k8s.pod.volume.* metrics
+  - id: metric.k8s.pod.volume.available
+    type: metric
+    metric_name: k8s.pod.volume.available
+    stability: development
+    brief: "Pod volume available"
+    instrument: gauge
+    unit: "By"
```
jinja2, line 159:
```suggestion
    unit: "By"
    entity_associations:
      - k8s.pod
```
I believe we want to add pod id/name to these. I think the entity k8s.namespace also makes sense here.

ChrsMark: Sure! Added.

jinja2 (approved):

odubajDT (approved):

model/k8s/metrics.yaml
```diff
@@ -73,6 +73,124 @@ groups:
       - ref: network.interface.name
       - ref: network.io.direction
 
+  # k8s.pod.volume.* metrics
+  - id: metric.k8s.pod.volume.available
+    type: metric
+    metric_name: k8s.pod.volume.available
+    stability: development
+    brief: "Pod volume available"
+    instrument: gauge
```
dashpole: I would expect these to all be updowncounters, right?

model/k8s/metrics.yaml
```diff
@@ -73,6 +73,124 @@ groups:
...
+      - k8s.pod
+    note: |
+      This metric is derived from the
+      [VolumeStats.AvailableBytes](https://pkg.go.dev/k8s.io/kubelet@v0.33.0/pkg/apis/stats/v1alpha1#VolumeStats) field
+      of the [PodStats](https://pkg.go.dev/k8s.io/kubelet@v0.33.0/pkg/apis/stats/v1alpha1#PodStats) of the
+      Kubelet's stats API.
+    attributes:
+      - ref: k8s.volume.name
+        requirement_level: required
+      - ref: k8s.volume.type
```
dashpole, line 170: volume type seems quite useful, and not particularly high cardinality.  Why is it opt in here?

model/k8s/metrics.yaml
ChrsMark: Right, that'd be more accurate.

model/k8s/metrics.yaml
ChrsMark: No strong preference on this, I just started with the `opt_in` as it's the least opinionated one. I will change it to `recommended`.

ChrsMark:
@dashpole your comments should be addressed now, please take another look when you get the chance.

dashpole (approved):

ChrsMark:
@open-telemetry/specs-semconv-approvers @open-telemetry/specs-semconv-maintainers please take a look

PR #2321, opened 2025-06-04
Add semconv for k8s node condition as metric

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/2077

## Changes

This PR adds the `k8s.node.condition` metric


model/k8s/registry.yaml
```diff
@@ -434,6 +434,43 @@ groups:
...
+          by [K8s documentation](https://v1-32.docs.kubernetes.io/docs/reference/node/node-status/#condition).
+        examples: [ "Ready", "DiskPressure" ]
+        type:
+          members:
+            - id: ready
+              value: 'Ready'
+              brief: |
+                The node is healthy and ready to accept pods. and Unknown if the node controller has not heard from the node in the last node-monitor-grace-period"
+              stability: development
+            - id: ready_unknown
```
dashpole: This doesn't seem to be listed in node conditions.  Is this just because it is the only condition that can be unknown in addition to t/f?

model/k8s/registry.yaml
```diff
@@ -434,6 +434,43 @@ groups:
               value: 'terminating'
               brief: "Terminating namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase)"
               stability: development
+      - id: k8s.node.condition
```
dashpole: Did we consider using `k8s.condition.type` and `k8s.condition.status` (without node/pod namespace)?  I know technically all of the XXXCondition types are separate, but they all follow the same type/status/reason pattern.

ChrsMark: What values would `k8s.condition.status` and `k8s.condition.type` accept?

dashpole: Status would have T/F/unknown.  Type would be open-ended, but we could provide examples from common objects

ChrsMark: I see! 

It seems kube-state-metrics follows the same pattern here: https://github.com/kubernetes/kube-state-metrics/blob/v2.15.0/docs/metrics/cluster/node-metrics.md?plain=1#L14

I will change it like this 👍

dashpole: Out of curiosity, does KSM have reason as a label?  That seems like it would be useful, although it is somewhat higher cardinality

ChrsMark: There are some reason labels for containers' and Pods' statuses but not for Node's conditions: https://github.com/kubernetes/kube-state-metrics/blob/v2.15.0/docs/metrics/workload/pod-metrics.md

For containers' status we have it already in SemConv: https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/metrics.yaml#L80

model/k8s/registry.yaml
ChrsMark: Right, but we can avoid this by using the `k8s.condition.status` with values T/F/unknown. Should be fixed by 
34f1e08.

Also, note that beyond these listed conditions there might other custom ones defined. So it's safer to have the option to track `unknown` status properly.

model/k8s/registry.yaml
dashpole: I would lean towards adding it, but if we are unsure, we can always do that later.

ChrsMark: Good idea! I have added it.

The modeling of this metric is becoming a bit tricky because we have to ensure that metric's name does not collide with a namespace. `k8s.node.condition` cannot be the metric name and a namespace for the attributes at the same time. Let me know what you think about the current suggestion.

dashpole: I still see `k8s.node.condition.*` for attributes.  Did you mean to update that based on this thread?

ChrsMark: Right, eventually I thought that having them under the `k8s.node.condition.*` namespace instead of a generic `k8s.condition.*` is safer. Having generic `k8s.condition.*` attributes across all K8s might be tricky. However we can revisit this during the stabilisation phase and check for all phase/status/condition metrics if there is potential for unifying their attributes?

Note: We had a similar discussion while unifying the `*.cpu.mode` attribute: https://github.com/open-telemetry/semantic-conventions/pull/1026

dashpole: ok, thats fine.

dashpole (approved):

docs/non-normative/k8s-migration.md
```diff
@@ -312,3 +313,20 @@ The changes in their metrics are the following:
...
+[#2077](https://github.com/open-telemetry/semantic-conventions/issues/2077)
+
+The changes in their metrics are the following:
+
+<!-- prettier-ignore-start -->
+
+| Old (Collector) ![changed](https://img.shields.io/badge/changed-orange?style=flat) | New                                                                                                                                                  |
+|------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `k8s.node.condition_*`                                                             | `k8s.node.condition` metric [0,1] with attribute `k8s.node.condition` for the different conditions and `k8s.condition.status` for true/false/unknown |
+
```
```suggestion
| `k8s.node.condition_*`                                                             | `k8s.node.condition.status` metric [0,1] with attribute `k8s.node.condition.type` for the different conditions and `k8s.node.condition.status` for true/false/unknown |

```

model/k8s/metrics.yaml
```diff
@@ -102,6 +102,22 @@ groups:
       The actual accuracy would depend on the instrumentation and operating system.
     instrument: gauge
     unit: "s"
+  - id: metric.k8s.node.condition.status
```
jinja2, line 159: Would implementations be expected to emit a metric with value 0 for all possible condition types not found on a node (kind of like k8s.container.status.state), or should they only emit values for those actually present in status.conditions?

model/k8s/registry.yaml
```diff
@@ -458,6 +458,68 @@ groups:
               value: 'terminating'
               brief: "Terminating namespace phase as described by [K8s API](https://pkg.go.dev/k8s.io/api@v0.31.3/core/v1#NamespacePhase)"
               stability: development
+      - id: k8s.node.condition.type
```
jinja2, line 461: Should we clarify in the documentation that the list of Kubernetes node condition types is not exhaustive? Or is the intention to only allow for these conditions at present? In which case we should probably update this line from earlier in the docs `If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.` Fwiw,  the current collector implementation allows user to pick the conditions to collect with a config option.

Node conditions are extensible and managed k8s environments introduce additional node conditions beyond the k8s defaults. For example, below are the node conditions observed on a GKE node:
```
❯ kubectl get node nodename -o jsonpath='{range .status.conditions[*]}{.type}={.status}:{.reason}{"\n"}{end}' --context gke-lab
NetworkUnavailable=False:RouteCreated
DeprecatedAuthsFieldInContainerdConfiguration=False:DeprecatedAuthsFieldInContainerdConfigurationNotDetected
FrequentKubeletRestart=False:NoFrequentKubeletRestart
DeprecatedMirrorsFieldInContainerdConfiguration=False:DeprecatedMirrorsFieldInContainerdConfigurationNotDetected
SysctlChanged=True:NodeSysctlChange
CorruptDockerOverlay2=False:NoCorruptDockerOverlay2
KernelDeadlock=False:KernelHasNoDeadlock
FrequentDockerRestart=False:NoFrequentDockerRestart
FrequentContainerdRestart=False:NoFrequentContainerdRestart
DeprecatedPullingSchemaV1Image=False:DeprecatedPullingSchemaV1ImageDetected
ReadOnlyRootFileSystem=False:ReadOnlyRootFileSystemNotDetected
Swap=False:SwapNotDetected
DeprecatedConfigsFieldInContainerdConfiguration=False:DeprecatedConfigsFieldInContainerdConfigurationNotDetected
DeprecatedUsingV1Alpha2Cri=False:DeprecatedUsingV1Alpha2CriNotDetected
FrequentUnregisterNetDevice=False:NoFrequentUnregisterNetDevice
DeprecatedOtherContainerdFeatures=False:DeprecatedOtherContainerdFeaturesNotDetected
MemoryPressure=False:KubeletHasSufficientMemory
DiskPressure=False:KubeletHasNoDiskPressure
PIDPressure=False:KubeletHasSufficientPID
Ready=True:KubeletReady
```

model/k8s/metrics.yaml
ChrsMark: Hmm, good point. I think same reasoning of `k8s.container.status.state` applies here too regarding the [missing metrics](https://github.com/open-telemetry/semantic-conventions/blob/d57b2ca0c9a3c5de7253c9a09f797a1afb57b6ad/model/k8s/metrics.yaml#L73-L74).

However I wonder how this can work well when the enum list of states is not exhaustive: https://github.com/open-telemetry/semantic-conventions/pull/2321#discussion_r2138857637

model/k8s/registry.yaml
ChrsMark: Yeap, the list is not exhaustive. That's what the note `If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.` clarifies.  Would we need anything more than this? Usually the mentioned note is enough for this type of enums in SemConv, so not sure if (and how) we should be more explicit here 🤔

model/k8s/metrics.yaml
ChrsMark: @dmitryax @jinja2 regarding our discussion during the K8s SemConv SIG, here is a kube-state-metrics sample:

```console
# HELP kube_node_status_condition [STABLE] The condition of a cluster node.
# TYPE kube_node_status_condition gauge
kube_node_status_condition{node="kind-control-plane",condition="MemoryPressure",status="true"} 0
kube_node_status_condition{node="kind-control-plane",condition="MemoryPressure",status="false"} 1
kube_node_status_condition{node="kind-control-plane",condition="MemoryPressure",status="unknown"} 0
kube_node_status_condition{node="kind-control-plane",condition="DiskPressure",status="true"} 0
kube_node_status_condition{node="kind-control-plane",condition="DiskPressure",status="false"} 1
kube_node_status_condition{node="kind-control-plane",condition="DiskPressure",status="unknown"} 0
kube_node_status_condition{node="kind-control-plane",condition="PIDPressure",status="true"} 0
kube_node_status_condition{node="kind-control-plane",condition="PIDPressure",status="false"} 1
kube_node_status_condition{node="kind-control-plane",condition="PIDPressure",status="unknown"} 0
kube_node_status_condition{node="kind-control-plane",condition="Ready",status="true"} 1
kube_node_status_condition{node="kind-control-plane",condition="Ready",status="false"} 0
kube_node_status_condition{node="kind-control-plane",condition="Ready",status="unknown"} 0
```

Per condition, only one `status` will make the metric to report `1`:

```console
kube_node_status_condition{node="kind-control-plane",condition="DiskPressure",status="true"} 0
kube_node_status_condition{node="kind-control-plane",condition="DiskPressure",status="false"} 1
kube_node_status_condition{node="kind-control-plane",condition="DiskPressure",status="unknown"} 0
```

KSM, does not include a `reason` label but this PR plans to do so as optional. This would make the metrics look like the following since `reason` attribute does not really make sense for when metric is not `1`?

```
kube_node_status_condition{node="kind-control-plane",condition="DiskPressure",status="true"} 0
kube_node_status_condition{node="kind-control-plane",condition="DiskPressure",status="false", reason="KubeletHasNoDiskPressure"} 1
kube_node_status_condition{node="kind-control-plane",condition="DiskPressure",status="unknown"} 0
```

Probably we could explicitly document that `reason` attribute should only be reported for condition/status pairs that make the metric report `1`, or just introduce it under a different metric (similar to [container status reason](https://github.com/open-telemetry/semantic-conventions/blob/d57b2ca0c9a3c5de7253c9a09f797a1afb57b6ad/model/k8s/metrics.yaml#L67-L92))?

/cc @dashpole

dashpole: Can you link to the KSM discussion around how they plan to introduce reason?  It does seem quite tricky

ChrsMark: Hmm, with a quick search I couldn't find any issue in KSM repo proposing adding `reason` for node's condition.

ChrsMark: I suggest we skip the `reason` for now: 937efb9

dashpole: works for me!

model/k8s/registry.yaml
jinja2: Wdyt about adding one of the below suggestions to make this more explicit?

In attributes -
```
The set of possible values for k8s.node.condition.type is not limited to those listed here. Managed Kubernetes environments, or custom controllers MAY introduce additional node condition types. When this occurs, the exact value as reported by the Kubernetes API SHOULD be used.
```

Or in metrics.yaml
```
The list of node condition types is non-exhaustive. Implementations should emit a data point for each observed node condition, using its exact type value as reported by the Kubernetes API.
```

model/k8s/metrics.yaml
jinja2: Thanks for the update! I see we'll be reporting metric per condition+status combo like ksm does.

model/k8s/registry.yaml
ChrsMark: Sure! Added: a6aff07

jinja2 (approved):

ChrsMark:
@open-telemetry/specs-semconv-approvers @open-telemetry/specs-semconv-maintainers this one is approved by the K8s SIG. Could you take a look?

joaopgrassi (approved):

PR #2322, opened 2025-06-04
[chore] properly define entity_associations for k8s metrics

ChrsMark:
## Changes

We can now define how metrics are associated with entities using the `entity_associations` key. This PR updates the k8s metrics properly.

### DIFF
--- a/docs/system/k8s-metrics.md
+++ b/docs/system/k8s-metrics.md
@@ -103,7 +103,7 @@ This metric is [recommended][MetricRecommended].
 
 | Name     | Instrument Type | Unit (UCUM) | Description    | Stability | Entity Associations |
 | -------- | --------------- | ----------- | -------------- | --------- | ------ |
-| `k8s.pod.uptime` | Gauge | `s` | The time the Pod has been running [1] | ![Development](https://img.shields.io/badge/-development-blue) |  |
+| `k8s.pod.uptime` | Gauge | `s` | The time the Pod has been running [1] | ![Development](https://img.shields.io/badge/-development-blue) | `k8s.pod` |
 
 **[1]:** Instrumentations SHOULD use a gauge with type `double` and measure uptime in seconds as a floating point number with the highest precision available.
 The actual accuracy would depend on the instrumentation and operating system.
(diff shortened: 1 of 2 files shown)


jinja2:
I'd like to start a discussion on how we should account for the k8s.namespace.name attribute in our metrics.

Currently, for namespace-scoped Kubernetes objects, the k8sclusterreceiver reports k8s.namespace.name as a resource attribute. I see the need for `k8s.namespace.name` to be reported with namespace-scoped k8s metrics. To ensure consistency, I see a few options to account for this attribute:

1. Add k8s.namespace to entity_associations for relevant metrics.
The [current k8s.namespace entity definition](https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/entities.yaml#L26-L37) in semantic conventions only requires k8s.namespace.name, but we may add uid in the future. If we go this route, does this mean the entity_association would require us to report both the namespace name and (eventually) the uid on these metrics?

2. Update other Kubernetes entities (e.g., Pod, Deployment) to include k8s.namespace.name.
This would ensure the namespace name is always present for namespace-scoped objects, regardless of how metrics are structured.

3. Add k8s.namespace.name as a data attribute on relevant metrics (don't think we should do this).

Wdyt think about this? Is there a preferred pattern for how we should surface k8s.namespace.name for namespace-scoped resources and their metrics? Lmk if a separate issue is preferred to discuss this.

ChrsMark:
Thank's @jinja2 for raising this!

Eventually we'd need to decide on how we define associations between these entities, right? Pod belongs to a Deployment and a Namespace etc.

Maybe @jsuereth @dmitryax can help with that and probably that deserves a separate issue :)?

smith:
> Eventually we'd need to decide on how we define associations between these entities, right? Pod belongs to a Deployment and a Namespace etc.

@ChrsMark most of the relationship modelling will be work that happens in "Phase 2" (see https://github.com/orgs/open-telemetry/projects/85/views/1), and there is no syntax in semconv now for defining these relationships, so for the moment they are implicit.

I had a draft PR where I added some of these relationships and mapped them out, but it was just an exploration of what this might look like once we're working through the relationships: https://github.com/open-telemetry/semantic-conventions/pull/2081

ChrsMark:
Thank's @smith for the info!

@jinja2 given that this work is planed for later, shall we just link K8s metrics to their first level entities, (i.e. Pod's metrics related to Pod etc.), have this PR merged and follow-up once Entities have evolved to cover this?

jinja2:
Sounds good to me!

jinja2 (approved):

joaopgrassi:
@ChrsMark can you please fix the conflicts? Then we can merge this :)

PR #2328, opened 2025-06-06
Create an entity modelling guide

### LINKED ISSUE #2246: [entities] Adding Entity Support to Semantic Conventions
This is a belated tracking issue for entity support in semantic conventions split from https://github.com/open-telemetry/opentelemetry-specification/issues/4436.

Here is the [proposal](https://docs.google.com/document/d/1GmB0kDx9opHxjj2oy6A4h0fqE8UVW-3ebWgp6zhtbVE/edit?tab=t.0) discussed with semconv + entities groups we're executing against.

## Remaining Tasks

- [x] Add `entity_associations` to signals in Semantic Conventions
- [x] Rename `resource` group type to `entity`
- [x] Add support for Descriptive vs. Identifying attributes. (https://github.com/open-telemetry/weaver/pull/735)
- [ ] Update stability requirements for entities 
- [ ] Add Entity Modelling guide in Semantic Conventions contributing guide

jsuereth:
Fixes #2246

## Changes

- Creates new "how to write conventions" section in non-normative guidance
- Adds "resource and entity" guidance to this section.

### DIFF
--- a/docs/non-normative/how-to-write-conventions/resource-and-entities.md
+++ b/docs/non-normative/how-to-write-conventions/resource-and-entities.md
@@ -174,6 +174,14 @@ attributes. Adding for example `process.executable.name`
 attribute to the identity is unnecessary and violates the
 rule of having a [minimally sufficient ID](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.45.0/specification/entities/data-model.md#minimally-sufficient-identity).
 
+Identifying attributes generally form the lifespan of an entity. This is
+important, particularly, for metrics written against an entity.  The lifespan
+determines whether a timeseries remains "connected" between reported points, or
+if it suddenly looks like a drop. It is recommended to select an identity that
+keeps the lifespan "stable" for important alerting and monitoring use cases.
+
+Identifying attributes MUST NOT change during the lifespan of the entity.
+
 #### Multi-Observer Guidance
 
 When choosing Identifying attributes, care should be taken to ensure that


docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+    name: my_entity
+    brief: >
+      A description of my_entity here.
+    attributes:
+      - ref: some.attribute
+        role: identifying
+      - ref: some.other_attribute
+        role: descriptive
+        ...
+```
```
lmolkova, line 41: great idea to include an example!

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+
+Notes:
+
+- You cannot declare an association on an *unstable* resource from a
+  *stable* signal.
+- You can declare multiple associations. These form a "one or many" set,
+  where one or many of the named entities may be associated with the
+  metric. There is *no* requirement on disjointness due to how is-a
+  relationships are modelled.
+
```
lmolkova, line 78: should  we mention that some entities (e.g. `service`) are implicitly associated with everything and don't need to be explicitly listed? Or should we encourage people to add all applicable ones?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+    extends: entity.my_entity
+    attributes:
+      - ref: new.attribute.name
+        requirement_level: opt_in
+        role: descriptive
+```
+
+Notes:
+
+- You cannot change the `name` field of the new entity.
```
lmolkova: nit:

```suggestion
- You cannot change the `name` or `type` fields of the new entity.
```

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+container-based entities are not enough, then new entities should be
+defined.
+
+### When to define an "is-a" relationship vs. extending descriptive attributes?
+
+There are two key rules:
+
+- Default to introducing separate entities with a clear "is a" (or similar) relationship
+- Extend an entity with new descriptive attributes if and only if the following is true:
+  - The extending entity cannot be associated with any telemetry by itself.
```
lmolkova, line 144: I don't understand this sentence. maybe an example would be helpful?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+  - The extending entity doesn’t have any other entities that logically can only be associated with it.
```
lmolkova: don't understand this one either

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+
+### How to define identifying attributes?
+
+The identifying attributes should be minimally sufficient to
+identify an entity within the context of how that entity is
+discovered. For example, when discovering kubernetes entities
+like `k8s.pod`, `k8s.deployment`, the identifying attributes
+should be sufficient to identify these entities within the
+scope of a kubernetes cluster (or more specifically, the
+kubernetes api server where the entities are discovered).
```
lmolkova: nit

```suggestion
kubernetes API server where the entities are discovered).
```

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+
+#### Multi-Observer Guidance
+
+When choosing Identifying attributes, care should be taken to ensure that
+ multiple observers will find the same identifying attribute for the same
+ entity. Generally, entities may be discovered both within OpenTelemetry
+ SDKs and the Collector and should leverage identifying attributes that
+ will be the same between these signal providers.
+
+For example, service.instance.id can be problematic to detect from outside
```
lmolkova: nit

```suggestion
For example, `service.instance.id` can be problematic to detect from outside
```

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+an SDK and inside an SDK consistently. Generally, this can only be
+achieved if some outside source injects a service.instance.id value into
```
lmolkova:
```suggestion
achieved if some outside source injects a `service.instance.id` value into
```

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+the SDK that is externally visible.  An alternative is to have the SDK
+provide a relationship between the service.instance.id and another entity
```
lmolkova:
```suggestion
provide a relationship between the `service.instance.id` and another entity
```

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+that is visible externally. Care should be taken when modelling Entities
+to avoid this problem where possible.
```
lmolkova, line 199: can we add some reasoning/explanation that despite being anti-pattern, we make `service.instance.id` identifiable in https://github.com/open-telemetry/semantic-conventions/pull/2278/files#diff-3d85dfd25c44f916cd9a6c53adf7628b01d37ce93cd2fd5083567af3ee9a64ef due to its usefulness?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,278 @@
...
+
+### How to namespace entities?
+
+Entities should be namespaced around the primary mechanism used to
+identify the Entity. For Example, kubernetes entities use the `k8s`
+namespace, and are primarily discovered using the kubernetes API or
+working within kubernetes.
```
lmolkova: nit, maybe add a link to https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md for "more details"  and maybe we need to make some adjustments there to the naming doc wrt entities?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Great question.

I think we encourage people to add applicable ones. I also wonder if we SHOULD list service on many conventions we have today, to denote we expect SDKs to have service and data to be attached to it as a minimum thing.  WDYT?

Also cc @open-telemetry/specs-entities-approvers for thoughts.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Added the link, and yes we can add new rules there as they evolve.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
lmolkova: let's try explicitly listing it and see if/how we like it

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: This is describing that care that needs to be taken even when using `service.instance.id` to make sure you don't wind up with different identity depending on who's observing it.  It's a bit of a major problem today, and manifests on kubernetes in this way:

- Reading container logs on a VM and trying to identify if those logs came from an OTEL service.
- Reading metrics via prometheus and trying to identify if those metrics came from an OTEL service.
- Accepting OTLP where the SDK did local discovery of OTEL service.

If you're not careful, you can wind up with three different `service.instance.id`s for the same SDK.

See: 

- OpenTelemetry Operator trying to *push down* an ID that will match service.instance id
  - Linked issues from https://github.com/open-telemetry/opentelemetry-operator/issues/3112
  - https://github.com/open-telemetry/opentelemetry-operator/issues/3495
- [Prometheus Mapping for server.instance.id: ](https://github.com/open-telemetry/opentelemetry-specification/issues/3136)](https://github.com/open-telemetry/opentelemetry-specification/issues/3136)


So this guidance is actually helping users understand the "magic" for how `service.instance.id` can work, and why it's horribly confusing when it doesn't.

While we *can* allow more things like `service.instance.id` I don't think we need to or should unless there's no alternative (which there is not for `service.instance.id`).  That's why the guidance in semconv is to have a system of record from which multiple observers can obtain the same identifier.

---

Regarding adding more rationale in this guide - I'll see what I can add.  

Do you think what I typed here makes sense for the guide?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Added an example.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: This one I get.  I can get rid of this when entity is a "signal" type, but effectively this is when you don't expect to report relationships for the entity.  It's just a special case of the first rule, but needs to be clarified prior to entity-as-a-signal (phase 2 for Entities work) being defined in the specification.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,285 @@
...
+
+There are two scenarios where entities should be defined:
+
+- When you are generating a new signal (log, metric, span, etc.)
+  and no existing entity makes sense as the "source".
+- (future) When you need to describe an entity hierarchy from
+  some system of record (e.g. resources in kubernetes, assets
+  in a cloud).
+
+For example, the Mainframe SIG in OpenTelemetry will need to outline a
```
smith: This paragraph seems like it may have been partially changed at some point, since the beginning is talking about mainframes but later it talks about container orchestration. Is this a copy-paste error or am I reading it wrong?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,285 @@
...
+like `k8s.pod`, `k8s.deployment`, the identifying attributes
+should be sufficient to identify these entities within the
+scope of a kubernetes cluster (or more specifically, the
+kubernetes API server where the entities are discovered).
+
+Commonly, a number of attributes of an entity are readily
+available for the telemetry producer to compose an identity. Of
+the available attributes the entity ID should include the
+minimal set of attributes that is sufficient for uniquely identifying that entity. For example a process on a host can be
+uniquely identified by (`process.pid`,`process.start_time`)
```
smith: Is this supposed to be [`process.creation.time`](https://github.com/open-telemetry/semantic-conventions/blob/main/model/process/registry.yaml#L191-L196)? That's not in the entities.yaml for process but it is in the registry.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,285 @@
...
+attribute to the identity is unnecessary and violates the
+rule of having a [minimally sufficient ID](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.45.0/specification/entities/data-model.md#minimally-sufficient-identity).
+
+#### Multi-Observer Guidance
+
+When choosing Identifying attributes, care should be taken to ensure that
+ multiple observers will find the same identifying attribute for the same
+ entity. Generally, entities may be discovered both within OpenTelemetry
+ SDKs and the Collector and should leverage identifying attributes that
+ will be the same between these signal providers.
```
smith: Looks like extra spaces here, maybe some word wrapping thing.

```suggestion
When choosing Identifying attributes, care should be taken to ensure that
multiple observers will find the same identifying attribute for the same
entity. Generally, entities may be discovered both within OpenTelemetry
SDKs and the Collector and should leverage identifying attributes that
will be the same between these signal providers.
```

smith (approved):

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: yes, updated!

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: It's actually two examples, updated.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Ok, rephrased all this and added more, PTAL.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: FYI - I made a long explanation in the comment, but also updated the guide to make this clear.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+
+## Modelling Guide
+
+To define an entity, create a new semantic convention model file, with a group type as `entity`, for example:
+
+`model/{my_domain}/entities.yaml`:
+
+```yaml
+groups:
+  - id: entity.my_entity
```
tigrannajaryan, line 29: Is `id` always prefixed by `entity.`? For example for K8s pods is it `entity.k8s.pod` or `k8s.pod`?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+    type: entity
+    stability: development
+    name: my_entity
+    brief: >
+      A description of my_entity here.
+    attributes:
+      - ref: some.attribute
+        role: identifying
```
tigrannajaryan, line 37: Nit: emphasize that the value of `role` should be either `identifying` or `descriptive` (e.g. that it is an enumeration not a string).

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+    ...
+    entity_associations: 
+      - my_entity
+```
+
+Notes:
+
+- You cannot declare an association on an *unstable* resource from a
+  *stable* signal.
+- You can declare multiple associations. These form a "one or many" set,
```
tigrannajaryan, line 74: Nit: given that `entity_associations` is optional it is "zero or more", right? Just trying to make sure there is no confusion about always requiring an association, which may be implied by "one".

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+  where one or many of the named entities may be associated with the
+  metric. There is *no* requirement on disjointness due to how is-a
```
tigrannajaryan: >There is *no* requirement on disjointness

Please explain. I am not sure I understand what disjointness means here.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+  relationships are modelled.
+
+### Extending an entity
+
+While not recommended for Semantic Conventions, you can define a new
+"view" of an entity that includes additional descriptive attributes. To
```
tigrannajaryan, line 82: I think calling this a "view" is confusing since the newly defined entity has a different entity type. "Extending" sounds about right to me.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+    extends: entity.my_entity
+    attributes:
+      - ref: new.attribute.name
+        requirement_level: opt_in
+        role: descriptive
+```
+
+Notes:
+
+- You cannot change the `name` or `type` fields of the new entity.
```
tigrannajaryan, line 100: Is there a `name` field for entities? I can't see it anywhere in the definitions.

Rant: the `type` field is named confusingly, it is not the type of the entity as it is known in the entity data model, but rather the `type` of the semconv yaml section. I understand the distinction, but it is still confusing. I don't know if there is a reasonable solution to this that doesn't involve a ton of work, so we may have to live with it..

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+- You cannot change the set of identifying attributes.
```
tigrannajaryan, line 101: Why not?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+- You cannot declare an association on an *unstable* resource from a
+  *stable* signal.
+- You can declare multiple associations. These form a "one or many" set,
+  where one or many of the named entities may be associated with the
+  metric. There is *no* requirement on disjointness due to how is-a
+  relationships are modelled.
+
+### Extending an entity
+
+While not recommended for Semantic Conventions, you can define a new
```
tigrannajaryan, line 81: Can you include an example where extending is recommended (or acceptable, if not recommended)?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+
+There are two scenarios where entities should be defined:
+
+- When you are generating a new signal (log, metric, span, etc.)
+  and no existing entity makes sense as the "source".
+- (future) When you need to describe an entity hierarchy from
+  some system of record (e.g. resources in kubernetes, assets
+  in a cloud).
+
+For example, the Mainframe SIG in OpenTelemetry will need to outline a
```
tigrannajaryan: The example is not clear, I don't think it helps. I don't know what Mainframe SIG plans to do and I think vast majority of this document readers don't know either. Is it possible to include an example from a more well-known domain?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+set of entities and relationships to define observed telemetry. Another example:
+If a new clustering solution (e.g. Hashicorp's Nomad) is defined, and existing
+container-based entities are not enough, then new entities should be
+defined.
+
+### When to define an "is-a" relationship vs. extending descriptive attributes?
```
tigrannajaryan, line 138: Is there a way to define "is-a" relationship? I don't see it in this document, is it supposed to be in this doc or somewhere else?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+
+There are two key rules:
+
+- Default to introducing separate entities with a clear "is a" (or similar) relationship
```
tigrannajaryan, line 142: Unclear how to do this, given that the doc doesn't explain how to add "is a" relationship.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+- Extend an entity with new descriptive attributes if and only if the following is true:
+  - The extending entity cannot be associated with any telemetry by itself.
+    *For example, when adding a `windows.process` entity, but no new
```
tigrannajaryan: What's the extended entity in this example? Is it `process`?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+    signals (metrics, logs, spans, etc.) would be added to the entity.*
```
tigrannajaryan: > ... no new signals (metrics, logs, spans, etc.) would be added to the entity

What does this mean? What does "adding a signal to entity" mean?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+  - The extending entity doesn’t have any other entities that logically can
+    only be associated with it.
+    *Note: this only applies to entities as signals with relationships.*
+
+This avoids the complexities of subtyping and ambiguous attribute usage.
```
tigrannajaryan, line 156: Are we calling "extending" a "subtyping"? Are these alias terms or different terms? If aliases can we choose one and stick with one?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+multiple observers will find the same identifying attribute for the same
+entity. Generally, entities may be discovered both within OpenTelemetry
+SDKs and the Collector and should leverage identifying attributes that
+will be the same between these signal providers.
+
+For example, `service.instance.id` can be problematic to detect from outside
+an SDK and inside an SDK consistently. Generally, this can only be
+achieved if some outside source injects a `service.instance.id` value into
+the SDK that is externally visible. An alternative is to have the SDK
+provide a relationship between the `service.instance.id` and another entity
```
tigrannajaryan, line 197: The alternative is not clear. How does having a relationship (which we don't support yet) help here?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+that is visible externally. Care should be taken when modelling Entities
+to avoid this problem where possible.
+
+The choice of `service.instance.id` should be an exception, not the rule, for
```
tigrannajaryan, line 201: >he choice of `service.instance.id` should be an exception, not the rule

I am confused. Is it necessary to say this? How can it be a rule for any other entity that is not a "Service"? Is it implied anywhere that it could be the rule?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+- Reading container logs from a `k8s.node`, where we know the container name
+  and deployment, but can't see into the SDK to understand a chosen instance id.
+
+The OpenTelemetry Operator, and onboarding guides for kubernetes, e.g. leverage
+mechanisms to ensure a `service.instance.id` can be pushed down to SDKs and
+external observers, alleviating this friction for kubernetes.
+
+### How to namespace entities?
+
+Entities should be namespaced around the primary mechanism used to
```
tigrannajaryan:
```suggestion
Entity types should be namespaced around the primary mechanism used to
```

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+  `host`, etc.
+- An Entity can distinguish *identifying* attributes from
+  *descriptive* attributes.
+  - Identifying attributes can be used to identify the entity
+    within some system (See
+    [minimally sufficient id](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.45.0/specification/entities/data-model.md#minimally-sufficient-identity)).
+    For Example, the `k8s.pod.uid` would be considered an
+    identifying attribute for a pod within kubernetes.
+  - *Descriptive* attributes can be used to provide additional labels for
+    entities, but do not uniquely identify that Entity.
```
tigrannajaryan: Nit: descriptive attributes may happen to uniquely identify an Entity. A better phrasing may be "are not necessary to uniquely identify".

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+    For Example, the `k8s.pod.uid` would be considered an
+    identifying attribute for a pod within kubernetes.
+  - *Descriptive* attributes can be used to provide additional labels for
+    entities, but do not uniquely identify that Entity.  
+- A Resource is composed of *multiple* Entities.
+  - Each of the entities within Resource is considered
+    'contributing' to that telemetry.
+  - For Example, today, most SDKs include the service entity,
+    but also another entity, like `k8s.container`, `host`, etc.
+- An Entity may be conceptually similar to another (which we
```
tigrannajaryan, line 270: I think "conceptually similar" is a bit vague. We may want to define "is-a" relationship better. For example a key characteristic of "is-a" relationship is that it is a one-to-one relationship. There is probably other characteristics that can help explain what "is-a" relationship is.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+        ...
+```
+
+Here, the attributes field contains all attributes of the Entity.
+The `role` of each attribute determines if it is identifying or
+descriptive. See
+[How to define identifying attributes?](#how-to-define-identifying-attributes)
+for details on what these mean.
+
+> Note: Declaring Entity relationships is not yet supported.
```
tigrannajaryan, line 49: The document later talks a lot about "is-a" relationship, so this note is confusing. If declaring relationships is not supported then what exactly is supported for relationships? Is it used purely as an explanation mechanism in this document?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: In semantic conventions, yes. We require this.  The oddities of `id` vs. `name` have been a bit of a thorn in our side in the model.  To keep things reasonably namespaced, semconv requires `{signal}.*` as a prefix for a group denoting a signal.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Weaver should fail with an error message denoting this.  Do you think we also need it in the guide?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: The newly defined entity does not have a different type.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: So the design decision for service exists because there is no central source of truth for what is a service. That should be the exception.  Other entities (e.g. k8s, cloud, etc.) should have some "holder" of entity information we can query/inspect that is stable across observers.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: I actually mean both the entity type and its attributes.  Should I clarify both?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Ah right, for when we have options.  I'll update this.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Agreed.  I think in Phase 2 for entities, when we dive into relationships we can refine this description.  For now, this needs to be enough to model existing "entities that will show up in resource".

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Yes.  This is to help folks understand better what a Resource with `container` and `k8s.container` would mean, for example, or another example why a resource might have `k8s.node`, `aws.ecs` and `host` entities.

Another way to phrase this - I think understanding is-a relationship, at a high level, is needed to model entities today.  We can flesh out exactly how to express that distinction in Phase 2.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Welcome to semantic conventions.  `type` predates us by quite some time in the YAML files, and it means the type of the group (metric, log, etc).  `name` means entity type in this case, and yes it's awkward.  We discussed a good bit before choosing what you see today.

See: https://github.com/open-telemetry/weaver/blob/main/schemas/semconv-syntax.md

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Would break stability of telemetry generated.  

Similarly you cannot change the set of attributes on a metric time series, remember that entity identifying attributes contribute to metric identity => no breaking.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: We had a few meetings in Entities SIG trying to come up with a good example.  @dmitryax may remember better, but I don't recall having one I felt strongly was a good example.  The best I came up with as acceptable here was the "windows process" vs. "process" example.  You may create a semconv group id to denote a windows process with descriptive attributes, but really it just reports with entity.type `process`.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Is the second example better?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: It's out of scope for now.  I think we mention it in the data model, and if not we can add it there.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: I think you can have a conceptual is-a relationship without a physical denotation of the is-a relationship.

Also given our principle of remaining open, "is-a" is likely to need to be defined external to an entity anyway.  One entity may not be aware of all the things it can "be".

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: yes

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: It means you define a semantic convention for a metric, span or event that has an entity association.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: The syntax in semconv YAML is `extend` but it's not subtyping at all.  That's where the confusion comes from.

I take it a expectation that folks have read https://github.com/open-telemetry/weaver/blob/main/schemas/semconv-syntax.md or before reading this guide.

Note: Post this PR - I do plan to expand this non-normative how to write conventions to include better links and onboarding.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: It at least lets you model the entity you need, and you can define the relationship in the future once that work lands.

Remember - our goal here is to unblock several semconv SIGs in defining and *stabilizing* attributes for resource. So implicity what we're saying is you'd have two entities reporting on Resource via detectors.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: See @lmolkova's comment - https://github.com/open-telemetry/semantic-conventions/pull/2328#discussion_r2132743735

I think we may actually *want* to require entity associations in semantic conventions going forward.  I'm going to leave this as "one or many" and take this discussion to semconv SIG to resolve.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: A signal can be associated with *all of* the entities.  It's not that if it's associated with one, it can't be associated with the other.  E.g. if I like `service` and `host`, I could be both a service and a host and I would see both.  I.e. the association doesn't have to be disjoint - meaning only one shows up in OTLP.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: That's fine.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: OK. I am confused by the use of word "disjointness". Please consider re-phrasing.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: I don't understand. The example shows a different value for the `id` field. Is that not the entity type?
The extending entity's value is "entity.my_entity_2" in your example and the extended entity's value is "entity.my_entity". Am I misunderstanding what the `id` field denotes?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: OK, please update or resolve this comment once a decision is made.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: Yes, I think it is. I think mainframes are a bit esoteric for most of our audience to be useful as an example.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: OK, makes sense. The phrasing in the doc is a bit hard for me to understand, please consider rewording.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: OK. I am not sure I understand why we mention subtyping then. Is it simply saying "this is NOT subtyping, don't try to interpret extending as subtyping".

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: Oh, I see. You are saying, the "Service" entity is special, don't try to design other entities the same way.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: Yes, please.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: OK, I don't mind using "is-a" to explain things in this doc. However it appears in so many places in this doc that I think it is warranted to clarify somewhere what it really means before we use it. The first reference to "is-a" in the doc is in a sentence about something different, which doesn't make it easier to understand what it is about.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
joaopgrassi: I also would advocate to explicitly list it.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+If a new clustering solution (e.g. Hashicorp's Nomad) is defined, and existing
+container-based entities are not enough, then new entities should be
+defined.
+
+### When to define an "is-a" relationship vs. extending descriptive attributes?
+
+There are two key rules:
+
+- Default to introducing separate entities with a clear "is a" (or similar) relationship
+- Extend an entity with new descriptive attributes if and only if the following is true:
```
joaopgrassi:
```suggestion
- Extend an entity with new descriptive attributes if, and only if, the following is true:
```

docs/non-normative/how-to-write-conventions/resource-and-entities.md
joaopgrassi: Maybe it's just me, but I found the latest version to be still hard to understand 🤔 

>  The extending entity cannot be associated with any telemetry by itself. For example, when adding a windows.process entity, but no new signals (metrics, logs, spans, etc.) would be added to the entity.
>
> The extending entity doesn’t have any other entities that logically can only be associated with it. Note: this only applies to entities as signals with relationships.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: `id` is an internal thing for semantic conventions, it has no meaning in the entity data model.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Yes.  We had a few folks comment repeatedly on subtyping and expecting a full subtyping relationship in entities.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: I'll add it to the FAQ

docs/non-normative/how-to-write-conventions/resource-and-entities.md
lmolkova: I think it's reasonable to require at least one entity association. We might want a more clever policy in the future (e.g. some are associated with `service` others with `host`) but making it required is a good first step forcing semconv authors to start thinking about it.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,301 @@
...
+
+Table of Contents
+
+- [Modelling Guide](#modelling-guide)
+- FAQ
+  - [When to define a new entity](#when-to-define-a-new-entity)
+  - [When to define a new "is-a" relationship](#when-to-define-an-is-a-relationship-vs-extending-descriptive-attributes)
+  - [How to define identifying attributes](#how-to-define-identifying-attributes)
+  - [How to namespace entities](#how-to-namespace-entities)
+- [Background](#background-resource-and-entities)
```
lmolkova: nit: 

```suggestion
<!-- toc -->

<!-- tocstop -->
```

(so it stays up to date)

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Added an "is-a" FAQ

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: So far 3 of 7 maintainers approve :)

I'm comfortable moving forward this direction.

I'll add follow up PRs to require entity associations going forward via policy.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Opened #2360

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: FYI - Added a FAQ on what "is-a" means.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Reworded again, PTAL.  Sorry, I realize I use a lot of math-y terms sometimes.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: added

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: ...ok... I was being to clever not having one of the headers expand.  Fixed.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
lmolkova: I'm still a bit lost. 

What I understand: `windows.process` could be an extension of `process` if and only if we don't expect windows-specific signals associated with it.

But we do expect them and we would want to associate them with `windows.process`. So `windows.process` should be "is-a" `process`. 

Am I missing something? Or is there a more realistic example?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
lmolkova: it makes total sense now - thanks for the explanation!

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: updated

docs/non-normative/how-to-write-conventions/resource-and-entities.md
tigrannajaryan: Which field in yaml defines the value of "entity type"? Is it `name`?

jsuereth: Yes

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: > But we do expect them and we would want to associate them with `windows.process`
It's my understanding we do not.

You're not missing anything.  We actually struggled to come up with an instance where "extension" really makes sense for entity modelling, but we couldn't agree there were NO situations, so we wind up with this documentation.  

I think we have two options going forward:

- Keep the windows process vs. process example assuming we aren't adding signals specific to windows on `windows.process` (which AFAIK is the plan)
- Change this section so that we just disallow modelling something.  Concretely:
  - All Entities in semconv will require a signal association to exists (e.g. a metric, span or event generated where they are the source)
  - We only allow "is-a" relationships to start with in modelling.
  
WDYT?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,331 @@
...
+
+There are two scenarios where entities should be defined:
+
+- When you are generating a new signal (log, metric, span, etc.)
+  and no existing entity makes sense as the "source".
+- (future) When you need to describe an entity hierarchy from
+  some system of record (e.g. resources in kubernetes, assets
+  in a cloud).
+
+For example, ff a new clustering solution (e.g. Hashicorp's Nomad) is
```
ChrsMark:
```suggestion
For example, if a new clustering solution (e.g. Hashicorp's Nomad) is
```

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,331 @@
...
+possible entities that exist in the world. Instead, we are allowing
+overlapping definitions across domains. For example, the `container` and
+`k8s.container` entities exist, and generally every `k8s.container` is
+a `container`, but not every `container` runs in kubernetes.
+
+An "is-a" relationship denotes that one entity is describing the exact
+same system component as another entity, but from a different domain. In
+the above example, `k8s.container` models containers from the kubernetes
+domain, while `container` is a general model for containers, regardless
+of how they are run (e.g. podmon, docker, kubernetes, FAAS, etc.)
```
ChrsMark:
```suggestion
of how they are run (e.g. podman, docker, kubernetes, FAAS, etc.)
```

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,331 @@
...
+
+OpenTelemetry, as an open ecosystem, cannot understand and model all
+possible entities that exist in the world. Instead, we are allowing
+overlapping definitions across domains. For example, the `container` and
+`k8s.container` entities exist, and generally every `k8s.container` is
+a `container`, but not every `container` runs in kubernetes.
+
+An "is-a" relationship denotes that one entity is describing the exact
+same system component as another entity, but from a different domain. In
+the above example, `k8s.container` models containers from the kubernetes
```
ChrsMark, line 129: I'm not sure if definitions here are quite accurate for this example.

To my understanding a `k8s.container` is essentially the declerative definition for running a container within K8s. This might have K8s specific characteristics , attributes etc. This is what the K8s API defines with https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#container-v1-core

While `container` is the system-level container that is actually running on the underlying system, through a container runtime. Runtimes follow the OCI, hence they should have a unified way to represent containers.

In some occasions, we can have a `k8s.container` but not yet a runtime `container` (i.e. during startup/scheduling etc).

`k8s.container.name` is usually completely different from the `container.name`. The way that we can link a `k8s.container` to a runtime `container` is through the `container.id` since it is included in [ContainerStatus. containerID](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#containerstatus-v1-core) in form of `<runtime_type>://<container_id>`

When we collect K8s metrics, we either use `k8s.container.*` namespace or the `container.*` namespace, based on the scope/context of the metric:
- [k8s.container.cpu.limit](https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/metrics.yaml#L539)
- [container.cpu.time](https://github.com/open-telemetry/semantic-conventions/blob/main/model/container/metrics.yaml#L14) (we use this metrics even cpu is collected from within the K8s scope)

Does the above resonate with the original context of this paragraph?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,331 @@
...
+- Default to introducing separate entities with a clear "is a" (or similar) relationship
+- Extend an entity with new descriptive attributes if, and only if, the following is true:
+  - The extending entity cannot be associated with any telemetry by itself.
+    - Example 1: When adding a `windows.process` entity, you do not expect
+      to create any specific process metrics or logs that would be
+      specific to `windows.process`, instead all data is still reported
+      against `process` entity.
+    - Example 2: When adding a `docker.container` entity, you do not
+      expect to create any specific container signals, instead logs,
+      metrics and spans would be reported against the `container` entity.
```
ChrsMark, line 151: What if there are metrics that are `docker` specific and are not covered by the OCI yet? Shouldn't we allow `docker.*` specific metrics in that case?

jsuereth: That's the guideline, yes.  

However, e.g. the system-metrics we have are actually a unification of metrics across linux+windows even if there isn't a standard yet.  So the modelling team COULD still decide to report metrics against container, but abstract what they do to work across docker + podmon, e.g.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Yes.

You're just giving more detail on why a `k8s.container` and `container` are separate actual entities with an `is-a` relationship.

I.e. ` k8s.container` is a declarative kubernetes resource that eventually starts a container.  Knowing that a specific `container` "is a" `k8s.container (i.e. the physical manifestation) is useful.

> When we collect K8s metrics, we either use k8s.container.* namespace or the container.* namespace, based on

This is when you define the attributes.  However, from an OTLP resource standpoint all of these metrics COULD include `k8s.container` and `container` attributes, allowing users to filter/slice by k8s container attributes.

Regarding, e.g. `container.cpu.time` - when collected from k8s scope, do you not report any `container` identifiers and only `k8s.container`?  Would this time diverge from what is reported if you *were* directly monitoring docker/podmon on the node?

ChrsMark: > Regarding, e.g. container.cpu.time - when collected from k8s scope, do you not report any container identifiers and only k8s.container?

In general, `container.id` should be added for `container.*` metrics when collected from K8s scope.

In the current implementation of `kubeletstats` receiver, by-default we [add `k8s.container.name`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9a57b6216ebb6f23125404886fa30f833256fd18/receiver/kubeletstatsreceiver/testdata/scraper/test_scraper_expected.yaml#L1598-L1613) but there is the option to also [add `container.id`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9a57b6216ebb6f23125404886fa30f833256fd18/receiver/kubeletstatsreceiver/testdata/scraper/test_scraper_with_metadata_Container_Metadata_expected.yaml#L3-L22). `container.id` is not added by-default by this receiver because it requires an extra api call (otherwise `k8sattributesprocessor` can be used to add `container.id` to these metrics).

> Would this time diverge from what is reported if you were directly monitoring docker/podmon on the node?

That shouldn't be the case.

jsuereth: > In general, container.id should be added for container.* metrics when collected from K8s scope.

So in this modelling guide `container` and `k8s.container` would be separate entities (connected by "is-a").  You could decide to report K8s states with *just* the `k8s.container` entity or both `k8s.container` and `container`.  If you did not use `container` then `container.id` is not necessary, and you could later reconstruct the "is-a" relationship via an entity signal.

The notion of "telescoping" is important here.  We want to give users the ability to pair down to "minimal" ids if needed, where our default may send more.  
> but there is the option to also [add container.id](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9a57b6216ebb6f23125404886fa30f833256fd18/receiver/kubeletstatsreceiver/testdata/scraper/test_scraper_with_metadata_Container_Metadata_expected.yaml#L3-L22). container.id is not added by-default by this receiver because it requires an extra api call (otherwise k8sattributesprocessor can be used to add container.id to these metrics).

Exactly!  The addition of `container` entity is an optional "telescope" feature for resource (you're extending the telescope to include more identity at the cost of larger signals + a runtime lookup).  This is why `k8s.container` and `container` are separate, so you can do this and because the "source of truth" for each may be different.

Additionally, remember that *not every* `container` is a `k8s.container`.  We can have detection logic and metrics reported without interacting with k8s at all, and we want *consistent* instrumentation and identity in this case.

>> Would this time diverge from what is reported if you were directly monitoring docker/podmon on the node?
>
> That shouldn't be the case.

Shouldn't and "isn't" are different.  Is it the exact same instrumentation or is it different in calculation?

docs/non-normative/how-to-write-conventions/resource-and-entities.md
dmitryax: We have a doc to help differentiate "is-a" relationship versus "a view" https://docs.google.com/document/d/1odZvm1UXQ_9InPsA7AHH9jhgwymWbT0GioUCsvT8f90/edit?tab=t.0. For example, k8s workload is a view on top of k8s deployment, k8s daemonset or k8s statefulset

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,331 @@
...
+"view" of an entity that includes additional descriptive attributes. To
+do so, use the extends field on groups:
+
+`model/{my_other_domain}/entities.yaml`:
+
+```yaml
+groups:
+  - id: entity.my_entity_2
+    type: entity
+    extends: entity.my_entity
```
dmitryax, line 91: This should probably be a list given the examples of views that we have "k8s workload" or "cloud"

docs/non-normative/how-to-write-conventions/resource-and-entities.md
```diff
@@ -0,0 +1,331 @@
...
+    - Example 2: When adding a `docker.container` entity, you do not
+      expect to create any specific container signals, instead logs,
+      metrics and spans would be reported against the `container` entity.
+  - The extending entity doesn’t have any other entities that logically can
+    only be associated with it.
+    *Note: this only applies to entities as signals with relationships.*
+
+This avoids the complexities of subtyping and ambiguous attribute usage.
+
+### How to define identifying attributes?
```
dmitryax, line 158: Should we also say that the set of identifying attributes defines the lifespan of an entity. The values cannot change during the lifespan.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
ChrsMark: > Shouldn't and "isn't" are different. Is it the exact same instrumentation or is it different in calculation?

I don't have a detailed answer right now, but we will need to check it before stabilising the K8s metrics. I expect that `container.cpu.time` and others non-controversial would be the same, however probably the `container.cpu.usage` would need more thought (i.e. kubelet has an opinionated approach for that).

/cc @open-telemetry/semconv-k8s-approvers

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: So weaver actually doesn't support `extends` in that way yet.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Added!

dmitryax (approved):

docs/non-normative/how-to-write-conventions/resource-and-entities.md
ChrsMark: Filed https://github.com/open-telemetry/semantic-conventions/issues/2418 for the above.

ChrsMark (approved):

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: rephrased

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Updated the examples here.

docs/non-normative/how-to-write-conventions/resource-and-entities.md
jsuereth: Rephrased a bit.

PR #2352, opened 2025-06-09
Re-enable github link checking

trask:


Makefile
```diff
@@ -13,7 +13,7 @@ endif
 
 ifeq ($(LYCHEE_GITHUB_TOKEN),)
   LYCHEE_GITHUB_TOKEN_ARG :=
-else:
+else
```
trask, line 16: this was why I concluded that passing `GITHUB_TOKEN` wasn't a solution for other repos 😭

jsuereth (approved):

Makefile
trask: FWIW, I'm really not seeing any consistent difference when testing in spec repo between passing `GITHUB_TOKEN` or not

but `max_concurrency` seems to help a lot (whether or not passing `GITHUB_TOKEN`)

PR #2358, opened 2025-06-11
Add k8s container ports attribute

### LINKED ISSUE #2357: [k8s] Define a semantic convention for k8s container ports
### Area(s)

area:k8s

### Propose new conventions

As part of the K8s SemConv stability work we need to define k8s container ports that will be used by the Opentelemetry Collector in the https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38019

The attribute to be added is:
`k8s.container.ports` following the format of other container resource attributes [here](https://github.com/ChrsMark/semantic-conventions/blob/cbd8872a8b5f87230a02e94d358e8f158f1c58ae/model/k8s/metrics.yaml/)

spurplewang:
Fixes #2357

## Changes

Add k8s container ports attribute

### DIFF
--- a/docs/registry/attributes/deployment.md
+++ b/docs/registry/attributes/deployment.md
@@ -36,8 +36,8 @@ considered to be identifying the same service:
 
 ## Deployment Deprecated Attributes
 
-"Describes deprecated deployment attributes."
+Describes deprecated deployment attributes.
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
-| <a id="deployment-environment" href="#deployment-environment">`deployment.environment`</a> | string | 'Deprecated, use `deployment.environment.name` instead.' | `staging`; `production` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `deployment.environment.name`. |
+| <a id="deployment-environment" href="#deployment-environment">`deployment.environment`</a> | string | Deprecated, use `deployment.environment.name` instead. | `staging`; `production` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `deployment.environment.name`. |
--- a/model/deployment/deprecated/registry-deprecated.yaml
+++ b/model/deployment/deprecated/registry-deprecated.yaml
@@ -2,7 +2,7 @@ groups:
   - id: registry.deployment.deprecated
     type: attribute_group
     brief: >
-      "Describes deprecated deployment attributes."
+      Describes deprecated deployment attributes.
     attributes:
       - id: deployment.environment
         type: string
@@ -11,5 +11,5 @@ groups:
           reason: renamed
           renamed_to: deployment.environment.name
         brief: >
-          'Deprecated, use `deployment.environment.name` instead.'
+          Deprecated, use `deployment.environment.name` instead.
         examples: ['staging', 'production']


model/k8s/registry.yaml
```diff
@@ -524,3 +524,9 @@ groups:
           or
           [K8s ContainerStateTerminated](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#containerstateterminated-v1-core)
         examples: ['ContainerCreating', 'CrashLoopBackOff', 'CreateContainerConfigError', 'ErrImagePull', 'ImagePullBackOff', 'OOMKilled', 'Completed', 'Error', 'ContainerCannotRun']
+      - id: k8s.container.ports
+        type: int[]
+        stability: development
+        brief: >
+          List of ports to expose from the container.
```
jinja2:
```suggestion
          List of ports exposed from the container. See the Kubernetes documentation for [Container ports](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#containerport-v1-core) for more details.
```

model/k8s/entities.yaml
```diff
@@ -60,6 +60,7 @@ groups:
       - ref: k8s.container.name
       - ref: k8s.container.restart_count
       - ref: k8s.container.status.last_terminated_reason
+      - ref: k8s.container.ports
```
jinja2, line 63:
```suggestion
      - ref: k8s.container.ports
        requirement_level: opt_in
```
I think the req level for this should be opt-in.

model/k8s/registry.yaml
```diff
@@ -524,3 +524,9 @@ groups:
...
+        examples: [[10053], [8080, 8081]]
```
jinja2: Should we clarify that the value will be an empty list when there are no ports exposed?

model/k8s/registry.yaml
```diff
@@ -524,3 +524,9 @@ groups:
           or
           [K8s ContainerStateTerminated](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#containerstateterminated-v1-core)
         examples: ['ContainerCreating', 'CrashLoopBackOff', 'CreateContainerConfigError', 'ErrImagePull', 'ImagePullBackOff', 'OOMKilled', 'Completed', 'Error', 'ContainerCannotRun']
+      - id: k8s.container.ports
+        type: int[]
```
ChrsMark: Should we consider making this list of maps instead so as to have the option for adding more attributes like `name`, `protocol` etc if needed in the future?

I think sth like https://github.com/open-telemetry/semantic-conventions/blob/13105b79d2dc47a582e919ac2c9a554c33b999eb/model/gen-ai/events.yaml#L96-L97 could be used as an example.

jinja2: Sounds good to me! We could do something like

```
- id: k8s.container.ports
  type: map[]
  fields:
    - id: port
      type: int
      requirement_level: required
    - id: name
      type: string
      requirement_level: optional
    - id: protocol
      type: string
      requirement_level: optional
```

model/k8s/registry.yaml
```diff
@@ -524,3 +524,9 @@ groups:
           or
           [K8s ContainerStateTerminated](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#containerstateterminated-v1-core)
         examples: ['ContainerCreating', 'CrashLoopBackOff', 'CreateContainerConfigError', 'ErrImagePull', 'ImagePullBackOff', 'OOMKilled', 'Completed', 'Error', 'ContainerCannotRun']
+      - id: k8s.container.ports
```
thompson-tomo: I think this should be under the container namespace.

thompson-tomo (changes requested):

model/k8s/registry.yaml
spurplewang: > I think this should be under the container namespace.



> I think this should be under the container namespace.

Did you mean this field belongs under the model/container section? I partially agree it fits there. However, given it's specific to Kubernetes, and I see other k8s.container.XXX fields already present, I'm wondering what rule guides the placement for such Kubernetes-specific attributes.

jinja2: I think the "container" namespace (without the k8s prefix) is reserved for generic, container runtime-agnostic attributes. In contrast, k8s.container.* attributes are reserved for information that’s specific to Kubernetes’ object model and APIs.

For container ports, if we only want to capture a simple list of exposed ports, it makes sense to propose a generic container.ports attribute. However, in k8s, ports can have additional metadata like name, and we may want to accommodate such enhancements in the future (see the discussion [here](https://github.com/open-telemetry/semantic-conventions/pull/2358/files#r2142015543)). Given that, perhaps it makes sense to keep this under k8s.container.ports if we think future enhancements where we add more structured port information is a use case.

thompson-tomo: I just looked at AWS ECS & gcp compute engine, both of which allow the protocol to be specified with ecs also allowing a name.

Thinking about it some more it might more sense to have a way to describe firewall/routing rules which is what is happening.

ChrsMark: It would make sense to use the `container.*` namespace if we can justify the decision based on an existing unified container spec, like https://opencontainers.org/. I'm not sure if OCI covers this (container ports) already, but with a quick search it seems that it doesn't.

If there is no unified container spec covering this right now, then we should go with the `k8s.container.*` option following the K8s API.

thompson-tomo: Just checked oci and https://github.com/opencontainers/image-spec/blob/64294bd7a2bf2537e1a6a34d687caae70300b0c4/config.md defines the exposed ports including the ability to specify the protocol.

jinja2: If we agree not to introduce the additional fields from the [K8s container port spec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#ports), then container.ports should be sufficient for port/protocol.

I think name might be a useful field to consider. In k8s, a service can reference a container port by its name as `targetPort`, so name could be useful if we want to correlate k8s service attr to container in the future. I don’t feel strongly about this and am just noting a possible future use case.

Is there a use case for host port? Even though it's not in the OCI image spec, tmk most container runtimes will let you specify a host port when running a container.

thompson-tomo: In relation to the Port name & host port, a key thing for me is what scenario they might be used to describe. Based on that identifying where it might fit best. 

In the case here, I am thinking they would be used to describe networking activity hence we could look at using the client/server namespace or alternatively the source/destination namespace.

ChrsMark: I'm not sure OCI Image spec is what we are looking for here. Also if we are going to consume what K8s API provides then that might be a subset of the ports a runtime container actually exposes. In that case the `k8s.container.*` namespace need to be used. 

@spurplewang could you verify what the context is here? 

From https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38019 I understand that the need is for providing the container ports that a K8s Pod spec defines as metadata, right? That might be different from what the runtime container actually exposes eventually and to know that we'd need to query the runtime API (docker, containerd, cri-o etc).

spurplewang: Yes, the container ports defined in k8s pod spec will be used as metadata for metrics.

ChrsMark: In that case I think `k8s.container.ports` is the most accurate for this. 

/cc @open-telemetry/semconv-k8s-approvers

model/k8s/entities.yaml
spurplewang: Added the requirement_level to be opt-in.

model/k8s/registry.yaml
spurplewang: Done.

model/k8s/registry.yaml
spurplewang: Added containerPort as a sub-field as this is the only thing going to be used.

model/k8s/registry.yaml
spurplewang: Added an example for empty list.

model/k8s/registry.yaml
spurplewang: Hi @ChrsMark , I accidently closed this PR. I did reopen it but not sure how to bring it back to `awaiting SIG approval` in `Semantic Conventions Triage`. Could you help navigate how that could be done? Thanks!

ChrsMark: No worries, I fixed that!

model/k8s/registry.yaml
```diff
@@ -687,3 +687,17 @@ groups:
...
         examples: [ 'count/replicationcontrollers' ]
+      - id: k8s.container.ports
+        requirement_level: opt_in
+        type: map[]
+        stability: development
+        brief: >
+          List of ports exposed from the container. See the Kubernetes documentation for [Container ports](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#containerport-v1-core) for more details. 
+          The value will be an empty list when no ports are exposed. 
+        fields:
+          - id: containerPort
```
ChrsMark: Probably that should be `container_port` based on [SemConv guidance](https://github.com/open-telemetry/semantic-conventions/blob/0c4e76a5a604b23de6f0b3dae17b0e8d164b09db/docs/general/naming.md?plain=1#L75)

model/k8s/registry.yaml
```diff
@@ -687,3 +687,17 @@ groups:
...
+            type: int
+            stability: development
+            brief: >
+              Number of port to expose on the pod's IP address. This must be a valid port number, 0 < x < 65536.
+            examples: [[],[10053], [8080, 8081]]
```
ChrsMark: I think this examples should be for the `container_port` field only, so 1 port per example?

See for example https://github.com/open-telemetry/semantic-conventions/blob/13105b79d2dc47a582e919ac2c9a554c33b999eb/model/gen-ai/events.yaml#L96-L97

ChrsMark (commented):
Thank's for reviving this!

It will need an entry to .md file as well similar to https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/system/k8s-metrics.md?plain=1#L283. Then run `make fix` to automatically update the md files.

model/k8s/registry.yaml
spurplewang: Done.

model/k8s/registry.yaml
spurplewang: Done

model/k8s/registry.yaml
```diff
@@ -687,3 +687,17 @@ groups:
...
+        brief: >
+          List of ports exposed from the container. See the Kubernetes documentation for [Container ports](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#containerport-v1-core) for more details. 
+          The value will be an empty list when no ports are exposed. 
+        fields:
+          - id: container_port
+            type: int
+            stability: development
+            brief: >
+              Number of port to expose on the pod's IP address. This must be a valid port number, 0 < x < 65536.
+            examples: [10053, 8080, 8081]
```
jinja2, line 703: Should we include a full example for k8s.container.ports? would it be something like `[{container_port: 10053}, {container_port: 8080}]`?

ChrsMark: This type of example would need to be one level up though. @open-telemetry/specs-semconv-approvers is this type of examples allowed in general for `map[]` attributes?

PR #2362, opened 2025-06-12
Gen ai.embeddings.dimension.count

### LINKED ISSUE #2361: Add gen_ai.embeddings.dimension.count
### Area(s)

gen_ai

### What's missing?

OpenAI and possibly other LLM services have a request parameter called `dimensions` which will impact the number of dimensions the resulting output embeddings will have. This is not currently in the semantic conventions for gen_ai.

Add an attribute called `gen_ai.embeddings.dimension.count` which is added to the client spans and relevant metrics when making a call to the embeddings endpoint.

drewby:
Fixes #2361

## Changes

Adds `gen_ai.embeddings.dimension.count` to `gen_ai.embeddings.client` span and `gen_ai.client.operation.duration` metric. 


xrmx (approved):

model/gen-ai/metrics.yaml
```diff
@@ -66,6 +66,8 @@ groups:
           The `error.type` SHOULD match the error code returned by the Generative AI provider or the client library,
           the canonical name of exception that occurred, or another low-cardinality error identifier.
           Instrumentations SHOULD document the list of errors they report.
+      - ref: gen_ai.embeddings.dimension.count
```
lmolkova: dimension count has high cardinality - it can be any number, we can't have it as a metric dimension, at least not as a recommended one

joaopgrassi: Sounds like a good candidate for a histogram?

lmolkova: I imagine it to be static in practice. I.e. I'd use default (1536 - default for some models), hardcode one, or provide a per-service+model config. I.e. the distribution is probably not interesting, but having it as opt-in attribute sounds fine.

drewby: Do you recommend I remove this metric or make it a histogram. The usage intent is to correlate the requested dimension count to request duration. However, I understand the cardinality constraint. What do you recommend?

lmolkova: the suggestion is to remove the attribute or make it opt-in.

lmolkova: @drewby I took a liberty to update this PR and removed dimensions count from metrics for now. I hope you don't mind. If you feel strong about keeping it, let's discuss it as a follow up.

lmolkova (approved):

drewby:
@lmolkova, I'm not authorized to merge, but seems enough approvers.

PR #2363, opened 2025-06-12
Add database context propagation via `SET CONTEXT_INFO` for SQL Server

XSAM:
Part of #2162, Split from #2236, as this part seems reached the consensus.

https://github.com/XSAM/otelsql/pull/508 is the prototype to demonstrate how `SET CONTEXT_INFO` works. You can run the example on your local Docker environment.

### DIFF
--- a/docs/database/sql-server.md
+++ b/docs/database/sql-server.md
@@ -160,7 +160,9 @@ and SHOULD be provided **at span creation time** (if provided at all):
 
 ### SET CONTEXT_INFO
 
-Instrumentations MAY propagate context using [SET CONTEXT_INFO](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-context-info-transact-sql?view=sql-server-ver16) by setting a string representation of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) before executing a query. This is an opt-in behavior that SHOULD be explicitly enabled by the user.
+Instrumentations MAY propagate context using [SET CONTEXT_INFO](https://learn.microsoft.com/sql/t-sql/statements/set-context-info-transact-sql) by injecting fixed-length part of span context (trace-id, span-id, trace-flags, protocol version) before executing a query. For example, when using W3C Trace Context, only a string representation of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) SHOULD be injected. Context injection SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
+
+Variable context parts (`tracestate`, `baggage`) SHOULD NOT be injected since `CONTEXT_INFO` value length is limited to 128 bytes.
 
 Instrumentations that propagate context MUST execute `SET CONTEXT_INFO` on the same physical connection as the SQL statement (or reuse its transaction).
 


XSAM:
This PR only has sql server part with `SET CONTEXT`. I don't get any objection about this part on the SIG meet, so I am thinking about merge this first to unblock part of #2236.

@trask @lmolkova PTAL. Thanks!

.chloggen/sql-server-context-info-propagation.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add dashbase context propagation via `SET CONTEXT_INFO` for SQL Server
```
arminru:
```suggestion
note: Add database context propagation via `SET CONTEXT_INFO` for SQL Server
```

jsuereth (approved):
As discussed, this works for me now.  The normative language is intentional, and we're moving specification of this sort into semantic conventions

docs/database/sql-server.md
```diff
@@ -152,6 +154,37 @@ and SHOULD be provided **at span creation time** (if provided at all):
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Database Context Propagation
```
lmolkova: Nit

```suggestion
## Context propagation
```

docs/database/sql-server.md
```diff
@@ -152,6 +154,37 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+
+**Status**: [Development][DocumentStatus]
+
+### SET CONTEXT_INFO
+
+Instrumentations MAY make use of [SET CONTEXT_INFO](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-context-info-transact-sql?view=sql-server-ver16) to add text format of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) before executing a query. This is an opt-in behavior that SHOULD be explicitly enabled by the user.
```
lmolkova:
```suggestion
Instrumentations MAY propagate context using [SET CONTEXT_INFO](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-context-info-transact-sql?view=sql-server-ver16) by setting a string representation of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) before executing a query. This is an opt-in behavior that SHOULD be explicitly enabled by the user.
```

docs/database/sql-server.md
```diff
@@ -152,6 +154,37 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+
+`SET CONTEXT_INFO` MUST be executed on the same physical connection as the SQL statement (or reuse its transaction).
```
lmolkova:
```suggestion
Instrumentations that propagate context MUST execute `SET CONTEXT_INFO` on the same physical connection as the SQL statement (or reuse its transaction).
```

docs/database/sql-server.md
lmolkova: can/should we call out that other propagation formats (e.g. b3) can also be used?

docs/database/sql-server.md
```diff
@@ -152,6 +154,37 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
 <!-- endsemconv -->
 
+## Database Context Propagation
+
+**Status**: [Development][DocumentStatus]
+
+### SET CONTEXT_INFO
+
+Instrumentations MAY make use of [SET CONTEXT_INFO](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-context-info-transact-sql?view=sql-server-ver16) to add text format of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) before executing a query. This is an opt-in behavior that SHOULD be explicitly enabled by the user.
+
```
lmolkova, line 164: could we provide some guidance for tracestate and/or baggage? Or should we say that it's not possible to propagate them?

XSAM: We could say it is not possible to propagate them due to the 128-byte length limit.

I want to avoid giving users the illusion that they can propagate variable-length variables, like tracestate and baggage.

docs/database/sql-server.md
XSAM: I feel it would be hard to use other propagation formats due to the 128-byte length limit.

Maybe a callout is not necessary. I don't have a preference here. It is up to you.

docs/database/sql-server.md
lmolkova: sure, let's mention that propagation is limited to traceparent (or similar) due to length limit concerns and instrumentations SHOULD NOT propagate tracestate or baggage

docs/database/sql-server.md
lmolkova: the propagation format is somewhat orthogonal here, e.g. OTel SDK MUST provide B3 propagator https://github.com/open-telemetry/opentelemetry-specification/blob/445a6835d755c1913a9acaa227997952ac9b4ea2/specification/context/api-propagators.md#propagators-distribution  which would be of a fixed length and could look pretty similar to traceparent - `b3: 80f198ee56343ba864fe8b2a57d3eff7-e457b5a2e4d86bd1-1`

I'm not sure how much DB server side cares about the specific format used and if it's configurable today, but I think we should leave a room for configurability.

Left a suggestion in https://github.com/open-telemetry/semantic-conventions/pull/2363/files#r2172842835

docs/database/sql-server.md
```diff
@@ -152,6 +154,37 @@ and SHOULD be provided **at span creation time** (if provided at all):
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Context propagation
+
+**Status**: [Development][DocumentStatus]
+
+### SET CONTEXT_INFO
+
+Instrumentations MAY propagate context using [SET CONTEXT_INFO](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-context-info-transact-sql?view=sql-server-ver16) by setting a string representation of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) before executing a query. This is an opt-in behavior that SHOULD be explicitly enabled by the user.
```
lmolkova:
```suggestion
Instrumentations MAY propagate context using [SET CONTEXT_INFO](https://learn.microsoft.com/sql/t-sql/statements/set-context-info-transact-sql) by injecting fixed-length part of span context (trace-id, span-id, trace-flags, protocol version) before executing a query. For example, when using W3C Trace Context, only a string representation of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) SHOULD be injected. Context injection SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.

Variable context parts (`tracestate`, `baggage`) SHOULD NOT be injected since `CONTEXT_INFO` value length is limited to 128 bytes.
```

docs/database/sql-server.md
lmolkova: added suggestion in https://github.com/open-telemetry/semantic-conventions/pull/2363/files#r2172842835

docs/database/sql-server.md
XSAM: I think it is ok not to stick with B3, as

1. It is not a regular propagation. There is no header we can use, and we cannot even support baggage. So, supporting the `Propagators Distribution` section in the spec is still not possible even if we support B3. The implementation of this propagation cannot and will not use the propagator provided by users; otherwise, `baggage` might be propagated and exceed the length limit.
2. Whatever the propagator (B3 or traceparent) the user is using, the trace_id, span_id and trace_flags are the same in the OTel SDK. This means a user who sticks only with B3 can still produce the same `traceparent` value as the user who sticks with `traceparent`.

Furthermore, this is the format communicated between the SQL client and the SQL server receiver (or other third-party receiver).
B3 is similar to `traceparent` (`00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-00`), but they are not the same. The parser for B3 cannot parse `traceparent`. This means if we allow the B3 value (without the key) to be used, it would be very tricky for the SQL Server receiver to correctly parse the data.

I know we could bring the key with the value like `traceparent=00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-00`, but B3 has a Multiple Header version, which makes the value look like this`X-B3-TraceId=80f198ee56343ba864fe8b2a57d3eff7,X-B3-ParentSpanId=05e3ac9a4f6e3b90,X-B3-SpanId=e457b5a2e4d86bd1,X-B3-Sampled=1`. For that, we have to define a delimiter similar to SQLcommenter, but there is no benefit to it, as the information we push is basically the same.

So, it seems we have two options

- The current plan. I'd prefer this, as it is simple to understand and implement.
- Populate key-value pairs. But the definition will be complicated, as `=` and `,` are included as delimiters. SQL Server receiver will be a sqlcommenter-alike parser to parse the CONTEXT content.

@lmolkova WDYT? Which option do you prefer?

docs/database/sql-server.md
XSAM: Using a different format, such as B3 without the key, would be tricky for the SQL Server receiver to parse since the format is not clear to the parser.

Imagine a SQL Server database that can serve queries from different SQL clients with different propagation settings. And, only one version of sqlserver receiver fetches the SQL Server that needs to handle both B3 and traceparent.

This would create compatibility issues. Also see, https://github.com/open-telemetry/semantic-conventions/pull/2363#discussion_r2172934695

docs/database/sql-server.md
lmolkova: I'd like to avoid being prescriptive on context propagation format used. Ideally context propagation is not  hardcoded into the instrumentation and instrumentation supports custom implementations. 

E.g. instrumentation may take specific DB propagator implementation as a parameter and expect propagator to be aware of 128 bytes limitation. Which header propagator injects is not the instrumentation concern, but there may be default `CONTEXT_INFO` or sql commenter implementations to fall back to. Traceparent could definitely be the  default.

Server and client instrumentations has to use the same format in any case. Regardless of whether we add key name `traceparent=....` or just put a string value, both sides should use the same format. Similarly to HTTP, there is no compatibility between traceparent on the client and b3  on the server.

docs/database/sql-server.md
XSAM: Suggestion applied per https://github.com/open-telemetry/semantic-conventions/pull/2363#discussion_r2173106389.

docs/database/sql-server.md
XSAM: Applied suggestions from https://github.com/open-telemetry/semantic-conventions/pull/2363#discussion_r2172842835

joaopgrassi:
CC @open-telemetry/semconv-db-approvers

lmolkova (approved):

Issue #1084, opened 2024-05-28
Semantic conventions for virtual machine resources


thompson-tomo:
### Area(s)

area:new

### Is your change request related to a problem? Please describe.

I want to know details about the VM resources being utilised in a vendor agnostic manner

### Describe the solution you'd like

I would like a clear definition of semantic conventions which describe a VM provider

- vmhost.provider ie esxi, hyper-v
- vmhost.version
- vmhost.address

Should be checked if any vendor specific VM fields would make sense to be moved.

### Describe alternatives you've considered

Do not have it stsndardised

thompson-tomo:
@gberche-orange I have just made a pr to genericise the cloud platform attributes to be under host (#2367) if accepted this issue can focus on extending the attribute registry with the last 4 mentioned attributes. 

In the end you would have:
- a physical device which is represented by the device namespace, 
- a computing instance which is represented by the host namespace
- a location which is represented by the geo namespace

PR #2367, opened 2025-06-15
#2365 perform the rename of platform/provider attributes

### LINKED ISSUE #2365: Cater for non cloud based hosting
### Area(s)

area:cloud, area:host

### What's missing?

The ability to describe hosting environments which are not cloud based

Move `cloud.platform` & `cloud.provider` to be in the host namespace 

### Alternatives considered

Rename of the cloud namespace to hosting

### Additional Context

The way I see it

- `device` is a physical machine ie server
- `host` is the instance/environment hosting the software within a device
- `service` the software which has been run on a host
- `process` the actual running software

Precursor for
- #2337
- #2347

thompson-tomo:
Fixes #2365 

## Changes

This change seeks to start removing the ambiguity of how to document both hybrid and on-premise deployments. The step taken here is to decouple both the host provider & the host platform from being tied to the cloud namespace to now being directly under host as this is the platform/provider which is facilitating the host instance.

With this change we can now:
* Describe a host ie VM and the platform it is running on be it cloud or on-premise previously this was restricted to cloud
* A platform can be identified based upon on the product being used ie Openshift, hyper-v &/or the commercial name of the platform which is based on the provider


.chloggen/non_cloud_hosting.yaml
```diff
@@ -0,0 +1,28 @@
...
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: cloud, host
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
```
michaelsafyan, line 12: On-premise is sometimes referred to as "private cloud".

Instead of making a breaking change by renaming the keys, could -- instead -- the definition/description of the existing keys simply be made more expansive such as changing the description/definition of `cloud.provider` to make it clear that on-prem environments can also be represented with that key (using a company-specific naming convention for their on-prem environments)?

thompson-tomo: The problem with the current approach is that cloud.platform is an attribute when it should've been a sub-namespace with the attribute being cloud.platform.name. 

As you have pointed out, a hosting platform might be industry specific but based on a common product. with the current solution we have no logical place to put that base product but with this solution we do.

A key thing to point out is that these are not stable attributes so changes can be proposed.

I am aware of the private cloud scenario however it doesn't fit for the following scenario. A customer has a single desktop dell server device which is running esxi to host 2 VM on the 4x internal disks. The services running on the VM are only accessible within the same vlan due to no inbound traffic and users requiring a dedicated client to use them. as such it is Hard to classify that as the cloud due to the lack of a data centre.

jsuereth:
This PR does a lot, and the description doesn't really cover/motivate enough of the breaking changes.

I'd suggest breaking it down as follows:

- How do we want to model VM Entities.  Should there be a generic "vm.image" or do we add "host.image"?  I think this is a topic to open an issue and discuss with @open-telemetry/semconv-system-approvers 
- Should we abstract cloud providers to be platform providers and include "non cloud" things here.  This deserves an issue with motivation and discussion.  Sadly we don't have a set of owners for cloud provider attributes today, those we inherited from the early days of otel and need to sort out abstractions for multi-cloud users.
- A series of tasks (from the above) that are simpler decisions about specific abstractions and whether they still hold up based on decisions made with the first two.

thompson-tomo:
@jsuereth we already have an answer to 1 which is VM = host as per https://opentelemetry.io/docs/specs/semconv/registry/attributes/host/ and this makes sense to me.

For the second one we have #2365 & #1084 with #2365 being a subset of #1084. Perhaps @open-telemetry/semconv-system-approvers provide input/feedback on how to describe the platform providing the virtualised environment is it as here or another approach.

I will adjust the amount of attributes that has changed & that should reduce the volume of breaking changes while also increasing the chances of acceptance.

thompson-tomo:
Changes have been made to remove the breaking changes and focus on adding support for the hosting platform.

trask:

Issue #1106, opened 2024-06-01
Semantic Conventions for Health Check events


thompson-tomo:
### Area(s)

area:new

### Is your change request related to a problem? Please describe.

I want to be able to have health checks of all my different components recorded in  a consistent manner and in the end they would be producing an event

### Describe the solution you'd like

I would like for the introduction of a new attribute group healthCheck which contains the following recommended attributes

- healthCheck,type -> string
- healthCheck.result -> string
- healthCheck.resultDescription -> string
- healthCheck.duration ->float (number of seconds)

Health checks are then able to leverage other semantic conventions ie database to provide information about the database check.

Note i have based it dotnet core health checks but we should check other platforms ie SpringBoot to see what how they model their health checks to ensure we have an agnostic manner to describe the health check

MSNev:
I don't believe that is the domain of the event working group to define the event "definition".

martialblog:
Hi, as discussed over in the issue 3148, I think this is a sensible proposal and I can see some use cases.

I agree that `state_check` might be a better base namespace.

PR #2368, opened 2025-06-15
Adds in lmdb #1107

### LINKED ISSUE #1107: Additional DB systems
### Area(s)

area:db

I want to be able to attach metadata to my health checks which are logged as OTEL events and use the semantic conventions to dictate the keys of this meta data

add the following as DB systems:

- arango
- documentdb
- influxdb
- npgsql
- ravendb

_No response_

### Additional context

https://opentelemetry.io/docs/specs/semconv/attributes-registry/db/

thompson-tomo:
Progresses #1107

## Changes

Adds in lmdb as used by ruby.

## Usage example

https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/lmdb with the span defined in: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/patches/database.rb

### DIFF
--- a/.chloggen/add_db_systems.yaml
+++ b/.chloggen/add_db_systems.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: db
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: This adds ArangoDB, DocumentDB, RavenDB & SurrealDB as db systems.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [1107]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/.chloggen/add_db_systems.yaml
+++ b/.chloggen/add_db_systems.yaml
@@ -10,7 +10,7 @@ change_type: enhancement
 component: db
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: This adds ArangoDB, DocumentDB, RavenDB & SurrealDB as db systems.
+note: This adds lmdb as db systems as used by ruby.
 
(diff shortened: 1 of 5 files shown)


joaopgrassi:
CC @open-telemetry/semconv-db-approvers was there any discussion on the list of db systems? Should we keep it open and not define every single db out there? From the maintainer standpoint this looks fine to me, but I wanted to know the SIGs opinion.

thompson-tomo:
In this case I followed the list of db systems with promoted prometheus exporters https://prometheus.io/docs/instrumenting/exporters/

thompson-tomo:
I agree with not defining all db systems hence why the list focussed on those db systems which are promoted on [prometheus exporters](https://prometheus.io/docs/instrumenting/exporters/). This is because the otel collector, can scrape prometheus metrics and send them via otlp. The only additional ones are the ones required for the dotnet health check project.

lmolkova:
prometheus exporters don't follow OTel semantic conventions and collector receivers can't produce OTel compatible telemetry. 
Such instrumentations are expected to document their schema outside of semconv - 

https://github.com/open-telemetry/semantic-conventions/blob/3b64cb31022feaacb410bfd6e571c1f19b5fbce0/CONTRIBUTING.md?plain=1#L75-L78

If there are database instrumentations that follow OTel conventions, let's document them one-by-one with all the details including how to populate collection/namespace attributes, which status codes map to error, etc, similarly to https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/redis.md.

thompson-tomo:
Doesn't https://github.com/prometheus/prometheus/blob/v2.28.1/docs/configuration/configuration.md#relabel_config as well as the external_labels enable prometheus metrics to be become compliant by setting the required labels and doing the desired transformation within the collector/receiver?

At the same time the dotnet health check project is needing additional db systems to be defined, so that telemetry can be generated. That project will not be instrumenting the client library ie user queries but rather using db.system.name attribute in conjunction with the exceptions namespace to record failing health checks. Also I thought dedicated documents were only needed when the standard definition is being extended/overwritten? If every enum option was being documented the documentation would quickly become overwhelming and hard to maintain.

lmolkova:
if some db system names are needed for OTel-telemetry, such as health checks in .NET, such databases could be added. It'd  be useful to document the health check events along with constants. 

It's also important to follow naming guidance defined here https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#choosing-a-system-name

Please also link existing instrumentations and prototypes.

thompson-tomo:
Scope has been reduced. 

The definition of a health check has been defined in https://github.com/open-telemetry/semantic-conventions/issues/1106 which is the definition used for https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2228/ and a prototype still needing more work documented at: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2387

Issue #2297, opened 2025-05-23
Document NATS attributes similar to other messaging systems


thompson-tomo:
### Area(s)

area:messaging

### What's missing?

Documentation of nats attributes for spans

### Describe the solution you'd like

Creation of a nats messaging document similar to rabbitmq based upon the implementation available at https://github.com/nats-io/nats.net/blob/main/src/NATS.Client.Core/Internal/Telemetry.cs and registration of the messaging.nats.message.subject attribute.

PR #2370, opened 2025-06-15
Add additional messaging systems #2266

### LINKED ISSUE #2266: Additional messaging systems
### Area(s)

area:messaging

### What's missing?

A well known messaging system name which can be used for NATS, SignalR

The list of known messaging systems on [messaging.system](https://opentelemetry.io/docs/specs/semconv/attributes-registry/messaging/#messaging-system) is extended to include:

- SignalR 
- Nats

thompson-tomo:
Fixes #2266 

## Changes

Defines, Nats, SNS & SignalR as messaging systems/

Note nats documentation will be covered with #2297 


model/messaging/registry.yaml
```diff
@@ -192,6 +196,14 @@ groups:
               value: 'kafka'
               brief: 'Apache Kafka'
               stability: development
+            - id: mqtt
```
joaopgrassi, line 195: We should probably not have MQTT here as it is only the protocol, not the messaging system per se.

model/messaging/registry.yaml
```diff
@@ -192,6 +196,14 @@ groups:
...
+              value: 'mqtt'
+              brief: 'MQTT'
+              stability: development
+            - id: nats
```
joaopgrassi: Same here, NATS is not a messaging system.

model/messaging/registry.yaml
```diff
@@ -200,10 +212,26 @@ groups:
               value: 'rocketmq'
               brief: 'Apache RocketMQ'
               stability: development
+            - id: signalr
```
joaopgrassi: I'm not sure if we as a group have decided how to handle websocket and if it would even follow the messaging conventions. I think for now we need to drop it until we figure it out.

model/messaging/registry.yaml
```diff
@@ -200,10 +212,26 @@ groups:
...
+              value: 'signalr'
+              brief: 'SignalR'
+              stability: development
             - id: pulsar
               value: 'pulsar'
               brief: 'Apache Pulsar'
               stability: development
+            - id: beanstalkd
```
joaopgrassi: I feel this falls into the same thing as signalR. I think we discussed to leave in-memory messaging "systems" out of scope for now.

model/messaging/registry.yaml
thompson-tomo: Nats is not just a protocol but a full system of server application, client libraries which communicates over a range of transport protocols.

model/messaging/registry.yaml
thompson-tomo: Yes mqtt is a protocol but for me it becomes a system once you have a Brooker etc.

model/messaging/registry.yaml
thompson-tomo: Websockets is only 1 of 3 protocols supported signalr hence looking for a way to capture that messaging system.

model/messaging/registry.yaml
joaopgrassi: Then we would add the name of this broker/product if not already here, and not the protocol.

thompson-tomo: That would be extremely difficult to achieve in reality as a client library almost always only knows that it is sending a message to a mqtt Brooker and not the product who is acting as the Brooker. Hence having the system as mqtt makes most sense and it eliminates the need for every product which implements mqtt specification to be listed.

thompson-tomo:
Still awaiting review

lmolkova (changes requested):
Some of the systems added are not messaging systems (signalr and mqtt), but more importantly, systems being added should be used within otel instrumentations. 

If there is an otel instrumentation that needs one of them, let's add it individually following [system-specific naming](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#system-specific-naming). In most cases,  it also means adding corresponding markdown file that explains how to apply generic messaging conventions to that system.

thompson-tomo:
I am failing to see how mqtt & signalr are not messaging systems. based on the published definition

> The` system name SHOULD unambiguously identify this specific product or project

Both of those systems fall into the category of a project as they encompass a collection of software products ie clients and servers, which is used in messaging and can be used interchangeably. 

I did consider going microsoft.signalr however that i feel would be misleading as the server or client product would often not be Microsoft but only libraries used in those applications. We can go with microsoft.signalr if that will resolve it.

Note in the case of hosted signalr on azure i would still go with the same messaging system name and set the `peer.service.name` to indicate the peer being used.

For the topic of documentation describing it, there is now pages for signalr & mqtt, nats is seperate issue & solace is added due to already being implemented by the collector.

lmolkova:
mqtt and signalr are transport protocols and messaging conventions cover logical operations. e.g. you can use MQTT or HTTP to communicate with AWS SNS. You can use MQTT (via plugin) or AMQP 0.9.1 to communicate with Rabbit MQ.  Rabbit and SNS are messaging systems while MQTT/HTTP/AMQP are not.
MQTT protocol may be instrumented similarly to HTTP, but it would be a different instrumentation that might have very little intersection with messaging.

Similarly to databases, we would expect messaging operation to cover logical call and include time to connect, authenticate, retry an MQTT call, while mqtt instrumentation would apply on individual mqtt call level and would have a span per connection, retries, etc.  If some batching is done in a hacky ways (multiple messages are sent in one MQTT payload, like it's done in AMQP), we'd expect messaging instrumentation to know about it and apply context propagation to individual messages within the batch that are not accessible on the MQTT level.

I.e. messaging instrumentation applies on the higher level client SDK level, MQTT on the transport level. 

Messaging semconv are built on these principles and changing those (or introducing any other substantial changes) would require messaging SIG to be resurrected.

thompson-tomo:
I am aware that mqtt is a transport protocol, what I am wanting to describe is the mqtt Brooker scenario and usage of Clients such as https://www.nuget.org/packages/MQTTnet. Clients like that have very intersection, would changing the name to mqttBrooker or similar remove the ambiguity.

SignalR supports 3 transport protocols sse, ws & long polling as well as offering things connection handling hence making it similar to sns. The difference being you can easily embed the server component into your own app rather than relying on a hosted solution and the underlying transport can be seamlessly changed. Hence is a messaging system.

docs/messaging/mqtt.md
```diff
@@ -0,0 +1,93 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: SignalR
```
trisch-me:
```suggestion
linkTitle: MQTT
```

model/messaging/spans.yaml
```diff
@@ -333,3 +333,12 @@ groups:
         requirement_level: recommended
       - ref: aws.request_id
         requirement_level: recommended
+
+  - id: messaging.mqtt
+    type: attribute_group
+    stability: development
+    extends: attributes.messaging.trace.minimal
+    brief: "Attributes that exist for SignalR."
```
KalleOlaviNiemitalo: "SignalR" here doesn't look right.

model/messaging/registry.yaml
joaopgrassi: >Hence having the system as mqtt makes most sense and it eliminates the need for every product which implements mqtt specification to be listed.

This goes against the entirety meaning of this attribute. Although these conventions as they are are not stable, they have been around for years, many people relying on them as they are today. That is not to say we can't make breaking changes but if we do what you propose here, I'd feel like it is better to delete the attribute all together then, which I don't agree.

thompson-tomo: Actually it is not that different, in fact it is consistent with how db.system.name works. Look at the postgresql value and the fact that this value would also be used when interacting with Azure Postgresql. We could add a similiar note to this attribute to help.

Key thing is we are describing the usage of a brooker which is described in https://www.emqx.com/en/blog/the-ultimate-guide-to-mqtt-broker-comparison

joaopgrassi:
Hi @thompson-tomo 

We appreciate your efforts and all your recent work in the semconv repo. We have discussed this PR between the semconv maintainers and have decided to not accept it. I understand this may be frustrating but I hope the following conclusion helps with understanding our conclusion.

The messaging conventions cover the [instrumentation of the logical side of the operations](https://github.com/open-telemetry/semantic-conventions/pull/2370#issuecomment-3095249487), thus covering protocol-level aspects is not the goal of them. Additionally, as discussed both in comments here and during our SIG meetings, the `messaging.system.name` attribute should identify the product/system and not the underlying protocol. 

You brought it up that it can happen that an instrumentation may not know which messaging system/product it is talking to, which is a fair and valid point, but so far we have not encountered such situation. No other instrumentation library/author/obsevability vendor has raised this issue with us. 

Nevertheless, you [brought up the MQTT use-case ](https://github.com/open-telemetry/semantic-conventions/pull/2370#issuecomment-3095351535)of a library in .NET [nuget.org/packages/MQTTnet](https://www.nuget.org/packages/MQTTnet). But at first glance of this project, I understand that it is a full-fledged messaging system (with client and broker), based on MQTT:

> MQTTnet is a high performance .NET library for MQTT based communication. It provides a MQTT client and a MQTT server ( broker) and supports the MQTT protocol up to version 5.

So, to me, for the current conventions, if one wants to build instrumentation for such project, we would simply add a new messaging system there for `mqttnet` and add a page for it, similar as we have for other systems as well. 

IF we really come with a use case where a messaging client library being instrumented does not know which messaging system it is talking to, then we need to re-consider things. The point is, we cannot do this in a scope of this PR. It needs proper discussion within the messaging SIG group, which unfortunately is not active right now.

Similarly, with the SignalR topic and general real-time communication, the messaging SIG also needs to decide what to do - if these conventions should also apply to this domain or not. Again, nothing that we can conclude/finish in this PR.

Our suggestion on the next steps are:

- For now, `messaging.system.name` only captures messaging system/products, not protocol values
- Gather real-world example of instrumentation(s) that are unable to populate `messaging.system.name` because they don't know the system they are talking to
  - If we do find, open an issue to track it. Will be added to the messaging SIG backlog
- Open an issue to discuss how we want to model real-time communications, such as with SignalR and other frameworks.

thompson-tomo:
Hi @joaopgrassi 

Thanks for the detailed response however I would like to correct a couple of things.

> I understand that it is a full-fledged messaging system (with client and broker), based on MQTT:

That is not the case, mqttnet is just a dotnet library which can be used to connect to a mqtt compliant Brooker. That Brooker can be built using mqttnet or you can use any other Brooker such as Azure IoT Hub. Hence a clear case of where the client doesn't know the product it is connecting to.

Hence for me it doesn't make sense to define a messaging system for `mqttnet` when that is just the name of the client library.

> Similarly, with the SignalR topic and general real-time communication, the messaging SIG also needs to decide what to do - if these conventions should also apply to this domain or not.

Where is this split of use cases coming from? As I don't see the difference to the already defined messaging systems.

> For now, messaging.system.name only captures messaging system/products

What is the definition a messaging system you are using??? As I was simply following the published guidance of setting the messaging system name "based on instrumentations best knowledge". Key thing is as per the docs the system name is not always the same as the product being interacted with.

PR #2376, opened 2025-06-16
CPU Arch now lives under host.cpu #2374

### LINKED ISSUE #2374: Move host.arch to be under host.cpu.arch
### Area(s)

area:host

### What's missing?

The ability to have the cpu architecture grouped under the host.cpu namespace

The existing host.arch attribute is deprecated and instead advised to use to host.cpu.arch.  This is because the architecture is describing an attribute of the cpu and makes sense especially with entities for which the cpu is one.

thompson-tomo:
Fixes #2374

## Changes

This moves the CPU arch to now be under the host.cpu namespace rather than directly under the host namespace

### DIFF
--- a/.chloggen/move_host_arch.yaml
+++ b/.chloggen/move_host_arch.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: host
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: The CPU architecture now resides under the CPU to support the cpu entity
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [2374]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/docs/registry/attributes/cpu.md
+++ b/docs/registry/attributes/cpu.md
@@ -5,7 +5,8 @@
 
 ## CPU Attributes
 
-Attributes specific to a cpu instance.
+CPU fields can carry data about a specific cpu instance.
+Note: These CPU attributes are typically used to capture system metrics about a CPU. The CPU attributes under host are for describing the cpu in the host.
 
 **Attributes:**
 
--- a/docs/registry/attributes/host.md
+++ b/docs/registry/attributes/host.md
@@ -3,6 +3,9 @@
 
 # Host
 
+- [Host Attributes](#host-attributes)
+- [Deprecated Host Attributes](#deprecated-host-attributes)
+
 ## Host Attributes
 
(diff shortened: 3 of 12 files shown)


braydonk (approved):
I'm in favour of this. We should note that this will probably break some existing resource detectors, namely the one in the Collector (but it will likely need a few changes to come along with other semconv changes we made so I think it's okay).

rogercoll (commented):
I am in favor of moving the arch within the `cpu` namespace, but what is the difference between the `host.cpu.*` attributes and `cpu.*` ones? https://github.com/open-telemetry/semantic-conventions/blob/main/model/cpu/registry.yaml#L4 Should the `host.cpu.*` attributes only be part of `system.*` metrics?

Should we align these two groups?

braydonk:
At the moment, the `host.cpu` attributes are about CPUs attached to the host, but I do like the idea of all of these just being under `cpu` rather than having `cpu` and `host.cpu`.

I'll tag in @jsuereth and @dmitryax to answer this question from the entities perspective. `host.cpu` attributes are not identifying for the `host`, so they'd probably be considered descriptive attributes. Would it be okay for the `host` entity to have descriptive attributes that came from a different root namespace, or is it important that any attributes on an entity (even descriptive ones) share a root namespace?

thompson-tomo:
I was actually think about doing that but decided against it as hopefully entity relationships will help that scenario. What about if a message was added on the cpu.* namespace stating something similar to [geo](https://opentelemetry.io/docs/specs/semconv/registry/attributes/geo/) in relation to nesting it.

I would like to eventually see the scenario where both a device & the host could have cpu objects especially in virtualised environments which is where the cpu entity would come in hence difficult to put it at the root.

thompson-tomo:
So I have been thinking about this some more especially when trying to improve the brief on the page. While also thinking about how a consistent experience can be offered to users of otel and avoid duplication.

What your thought about moving `host.cpu*` & `cpu.*` attributes to be in the `hw.cpu.*` namespace while also taking advantage of #2417 to avoid duplication? I note cpu is already defined as a value for hw.type.

This would mean that the hardware namespace is where all fixed hardware properties are described.

If you are on board  @braydonk / @rogercoll I can make those changes tomorrow.

rogercoll:
I think that even this attribute is still not used in `system` or `k8s` metrics, it could potentially be. That would be the same case as [cpu.mode](https://github.com/open-telemetry/semantic-conventions/issues/840). 

> What your thought about moving host.cpu* & cpu.* attributes to be in the hw.cpu.* namespace while also taking advantage of https://github.com/open-telemetry/semantic-conventions/issues/2417 to avoid duplication? I note cpu is already defined as a value for hw.type.

Although the cpu hw.type is already defined, I find the hw. namespace for cpu attributes redundant. hw. implies a meaningful distinction between hardware and software representations of architecture, but in practice, `cpu.arch` reflects the effective architecture seen by the process/pod/host —whether real, virtualized, or emulated—making the prefix unnecessary. (not a strong opinion though)

thompson-tomo:
Agree we have a couple of scenarios for using the value of host.arch And it could be used more.

I think it would be beneficial to have 2 distinct attributes:
- hw.cpu.arch/hw.processor.arch to describe what sw is running. It would be great to distinguish between physical and virtual hw but that is seperate issue
- app.arch to describe the architecture it has been built for

Ideally eventually we could use a shared enum between them.

jsuereth:
@braydonk - Can you get the @open-telemetry/semconv-system-approvers group to make a decision on this here?
Regarding `cpu` vs `host.cpu` namespace, according to our new namespacing rules, `cpu` is reaosnably unique and this could live in `cpu`.  I defer to System semconv SIG for decision on what they want here.

thompson-tomo:
Don't forget we also have the `hw.cpu.*` namespace.

trask:

PR #2378, opened 2025-06-16
Add stability requirements for entities to have attribute roles.

jsuereth:
splits from #2278

## Changes

- Add more urgently needed controls to ensure stable entities have `identifying` and `descriptive` attributes modeled before stabilizing.
- Retrocatively declares identifying and descriptive attributes for stable entities.

### DIFF
--- a/model/telemetry/entities.yaml
+++ b/model/telemetry/entities.yaml
@@ -8,10 +8,13 @@ groups:
     attributes:
       - ref: telemetry.sdk.name
         requirement_level: required
+        role: identifying
       - ref: telemetry.sdk.language
         requirement_level: required
+        role: identifying
       - ref: telemetry.sdk.version
         requirement_level: required
+        role: descriptive
   - id: entity.telemetry.distro
     name: 'telemetry.distro'
     type: entity
--- a/model/service/entities.yaml
+++ b/model/service/entities.yaml
@@ -8,6 +8,10 @@ groups:
     attributes:
       - ref: service.name
         requirement_level: required
+        role: identifying
       - ref: service.version
+        role: descriptive
       - ref: service.namespace
+        role: identifying
       - ref: service.instance.id
+        role: identifying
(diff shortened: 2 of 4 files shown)


lmolkova (approved):

model/telemetry/entities.yaml
```diff
@@ -8,10 +8,13 @@ groups:
     attributes:
       - ref: telemetry.sdk.name
         requirement_level: required
+        role: identifying
       - ref: telemetry.sdk.language
         requirement_level: required
+        role: identifying
       - ref: telemetry.sdk.version
         requirement_level: required
+        role: descriptive
```
thompson-tomo, line 17: @jsuereth what would happen in the case where 2 dotnet apps were both producing otel telemetry but using different versions of the sdk?

thompson-tomo (commented):
Should we have a rule which prohibits the specifying of the requirement level on an attribute? Or perhaps if we need them for optional identifying attributes in the case of service we only allow them on the identifying attributes & not descriptive.

model/telemetry/entities.yaml
jsuereth: Pretty much the same thing that happens now.

From a metric timeseries identity perspective, the fact they're different "apps" means something else would disambiguate the timeseries.

thompson-tomo: Sorry where I was going with that comment was more about shouldn't the version attribute be also identifying? Given the scenario of 2 apps written in the same language running different sdk versions.

jsuereth: so you don't need to differentiate *application* using `telemetry.sdk`.  You just differentiate which SDK was chosen (which can be name and language).

You'd use a different entity to represent and differentiate application.

jsuereth:
> Should we have a rule which prohibits the specifying of the requirement level on an attribute? Or perhaps if we need them for optional identifying attributes in the case of service we only allow them on the identifying attributes & not descriptive.

Good question @thompson-tomo - We can allow any requirement level for descriptive, but need to limit identifying.  We haven't sorted out all those details but it is follow up work

PR #2380, opened 2025-06-16
move all hardware metrics to their own yaml files

### LINKED ISSUE #1309: Move hardware metrics to the registry
This is a tracking issue for hardware metrics

- [ ] move common attributes https://github.com/open-telemetry/semantic-conventions/pull/1030
- [ ] move `hw.host.`
- [ ] move `hw.battery.`
- [ ] move `hw.cpu.`
- [ ] move `hw.disk_controller.`
- [ ] move `hw.enclosure.`
- [ ] move `hw.fan.`
- [ ] move `hw.gpu.`
- [ ] move `hw.logical_disk.`
- [ ] move `hw.memory.`
- [ ] move `hw.network.`
- [ ] move `hw.physical_disk.`
- [ ] move `hw.power_supply.`
- [ ] move `hw.tape_drive.`
- [ ] move `hw.temperature.`
- [ ] move `hw.voltage.`

trisch-me:
Fix #1309

Continue to move hardware metrics to their own files.

@bertysentry I would need you guidance as it wasn’t possible to move metrics out as is due to restrictions:

We can’t have both `x.y` and `x.y.z` names as declaring `x.y.z` makes `x.y` automatically a namespace. so you can’t use it as leaf node. In our case we have originally `hw.battery.charge` and `hw.battery.charge.limit`. I have changed the latter to the `hw.battery.charge_limit` but I’m not sure if this is a correct way and if it breaks anything?

Another moment is about `hw.status` metric from https://opentelemetry.io/docs/specs/semconv/system/hardware-metrics/#hwbattery---battery-metrics. Do we need to define it separately as I did in this request? It relies on `hw.type` to be the only type (in this case `battery`) and `state` to be the set of specific states for each component and it doesn’t play well together with previous definitions and current restrictions on enums in semconv. We are working on them but before making a final contribution I just want to clarify these points as every other component has the similar set of questions i.e. `hw.cpu.speed` and `hw.cpu.speed.limit` and `hw.status` with another set of values. 

Some of the components have only `hw.status`. And what this actually mean - for example we do define it as `hw.enclosure.* - Enclosure metrics` but later say just `hw.status`. Should it be `hw.enclosure.status`?

cc @jsuereth for the hints on metric solution in this case


model/hardware/battery-metrics.yaml
```diff
@@ -0,0 +1,75 @@
...
+    metric_name: hw.battery.charge_limit
+    stability: development
+    brief: "Lower limit of battery charge fraction to ensure proper operation"
+    instrument: gauge
+    unit: "1"
+    extends: metric.hw.battery
+    attributes:
+      - ref: hw.battery.limit_type
+        requirement_level: recommended
+
```
thompson-tomo: As this would likely be a breaking change I wonder if we actually need it given it seems to be a strange metric. Could we remove it?

model/hardware/registry.yaml
```diff
@@ -104,6 +104,57 @@ groups:
...
+              brief: 'Throttled'
+              stability: development
+            - id: degraded
+              value: 'degraded'
+              brief: 'Degraded'
+              stability: development
+        stability: development
+        brief: >
+          Type of the battery charge limit
+      - id: hw.battery.chemistry
```
thompson-tomo, line 221: Should we not be using iana terminology https://www.iana.org/assignments/battery-technologies/battery-technologies.xhtml? Aka battery.technology

thompson-tomo (changes requested):

bertysentry:
@trisch-me I am surprised by this new requirement that you mentioned:

> We can’t have both x.y and x.y.z names as declaring x.y.z makes x.y automatically a namespace. so you can’t use it as leaf node. In our case we have originally hw.battery.charge and hw.battery.charge.limit. I have changed the latter to the hw.battery.charge_limit but I’m not sure if this is a correct way and if it breaks anything?

My opinion is that it's a shame that this limitation is caused by our tooling only. OTLP itself totally supports having `hw.battery.charge` and `hw.battery.charge.limit`.

The `.limit` suffix has also being used to specify thresholds, like `hw.temperature.limit`. Because of this limitation (namespace vs leaf), we will probably have to rename our temperature metric to `hw.temperature.temperature`, which is... suboptimal.

**Can we add a special status for `*.limit` metrics, to make sure they don't create a namespace? It shouldn't be too complicated.**

If not, my suggestion for battery charge and limit is to go simply with:

* `hw.battery.charge` (indicates the level of charge of the battery)
* `hw.battery.limit` (indicates the charge limits at which the battery can operate properly)

Thank you!

model/hardware/registry.yaml
bertysentry: `hw.battery.chemistry` follows DMTF CIM Model's terminology: https://schemas.dmtf.org/wbem/cim-html/2.31.0/CIM_Battery.html
 which seemed more recent than IANA's MIB definitions. But really both work and designate the same thing.

bertysentry:
@trisch-me WRT "Some of the components have only hw.status. And what this actually mean - for example we do define it as hw.enclosure.* - Enclosure metrics but later say just hw.status. Should it be hw.enclosure.status", that's an excellent question!

When writing the original semconv for hardware, I tried to group metrics per component type (CPU, memory, etc.). Each component had its own status metric, like `hw.enclosure.status` and `hw.physical_disk.status`, etc. But I was asked to refactor that as simply `hw.status` with `hw.type="enclosure|physical_disk|..."`, which I think was a great idea! I now strongly recommend that we keep `hw.status` as a generally applicable metric to any kind of hardware component. It works really great to assess hardware problems, with really simply queries.

*Edit: "I WAS asked", and not "I asked" and then thought it's such a great idea 😅*

braydonk:
> Can we add a special status for *.limit metrics, to make sure they don't create a namespace? It shouldn't be too complicated.

I'm in favour of this or revisiting this whole notion of metric names and namespaces etc.; the original issue for it (https://github.com/open-telemetry/semantic-conventions/issues/50) has been open for a while and the reasoning in favour of it doesn't seem super strong to me relative to the naming gymnastics it requires.

model/hardware/registry.yaml
thompson-tomo: Ok, wasn't aware of the dmtf definition, in that case we should mention the source of the possible values.

docs/hardware/battery.md
```diff
@@ -0,0 +1,231 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Common
```
thompson-tomo: @trisch-me please set link title to Battery.

trisch-me:
I have added a question about metric names to the todays semconv meeting

thompson-tomo (dismissed):

brunobat:
I wonder if we should follow the naming conventions already used by the car battery industry. 
They are pretty precise and are already widely used... This could also open up OpenTelemetry to that field. See: https://en.wikipedia.org/wiki/Battery_management_system#Computation

Examples: [State of charge](https://en.wikipedia.org/wiki/State_of_charge) (SoC), [State of health](https://en.wikipedia.org/wiki/State_of_health) (SoH), charge current limit(CCL)...

model/hardware/battery-metrics.yaml
trisch-me: you mean remove `charge_limit` metric?

thompson-tomo: Correct as I am struggling to see what this metric would actually enable us to see especially as it includes charge limit type.

trisch-me: I'm not sure I can remove the metric as this is not about factual changes but moving metrics to their own files.
@bertysentry could you please check this question?

model/hardware/registry.yaml
trisch-me: added link to description

model/hardware/battery-metrics.yaml
bertysentry: It is not a *"strange metric"*, the `hw.battery.charge.limit` metric represents operational limits of the battery pertaining to the battery device itself.

Examples are many: Li-ion batteries shouldn't be charged at more than 80% or less than 20%. That's 2 limits. Your phone goes into "battery saving mode" when its charge gets below 20%, and automatically gets out of "saving mode" when it reaches 80%. A UPS will trigger a clean shutdown of all connected servers once it reaches 10%, etc.

There are other metrics that could be added later, like charge cycles, and endurance remaining, etc. but these are for future changes!

Thank you!

thompson-tomo: @trisch-me you are already implementing breaking changes to the name/id.

@bertysentry you example makes sense however the brief doesn't convey that especially when you consider the limit type attribute.  Also we don't have the upper limit.

For instance how could a battery which is in critical, throttled, degraded state still be within the range of proper operation.

I would propose that
- the name becomes hw.battery.level.min as the value of the metric is not the battery charge but artificial levels. It will also allow max to be added later in a consistent way.
- The limit type values reviewed as currently there is not one to achieve the use cases you have just mentioned, as not one refers to optimal level but just critical, throttled & degraded.

trisch-me: @thompson-tomo yes, there are might be necessary changes to make it compliant with new requirements we have in tooling.
But this PR should not be delayed because we will discuss material changes. We could discuss and update it after hardware is included fully into semconv

thompson-tomo: Sure, in that case go with the rename to avoid a 2nd breaking change and the rest can be handled via a seperate PR.

bertysentry: @thompson-tomo Yes, you are right: the brief field doesn't properly describe the different use cases I mentioned. The brief should be updated, so the metric doesn't need to be renamed. Description could be changed to:

```yaml
brief: > 
  Represents battery charge level thresholds relevant to device operation and health. 
  Each limit_type denotes a specific charge limit such as the minimum or maximum 
  optimal charge, the shutdown threshold, or energy-saving thresholds. These values 
  are typically provided by the hardware or firmware to guide safe and efficient battery 
  usage.
```

I recommend not using metric names for describing the different thresholds, because we would then need to specify all the different types of thresholds we may have: `hw.battery.charge.min`, `hw.battery.charge.optimal_min`, `hw.battery.charge.shutdown_min`, `hw.battery.charge.save_min`, etc. I don't think this is sustainable. Using an attribute to describe the type of the limit offers more flexibility and is much easier to query: `hw.battery.charge.limit{limit_type="optimal_min|optimal_max|shutdown|saving_min|..."}`

thompson-tomo: > I recommend not using metric names for describing the different thresholds, because we would then need to specify all the different types of thresholds we may have: hw.battery.charge.min, hw.battery.charge.optimal_min, hw.battery.charge.shutdown_min, hw.battery.charge.save_min, etc. I don't think this is sustainable. Using an attribute to describe the type of the limit offers more flexibility 

I agree it is much better to use attributes to distinguish the limit type, the question becomes do we at a later point introduce an additional required attribute min/max or have them as seperate metrics.

Note The metric is currently `hw.battery.charge.limit` but is being changed to `hw.battery.charge_limit` via this PR as it would clash with `hw.battery.charge` within the tooling. ;)

As to me the rename is unavoidable, the Name could become `hw.battery.threshold.min` or `hw.battery.threshold.limit` as they enable maximum chance of future expansion without issues.

model/hardware/fan-metrics.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - id: metric.hw.fan
+    type: attribute_group
+    brief: >
+      Common attributes for fan metrics
+    extends: hardware.attributes.common
+    attributes:
+      - ref: hw.fan.sensor_location
+        requirement_level: recommended
+
+  - id: metric.hw.fan.speed
```
thompson-tomo, line 12: Namespace issue, Should It not become `metric.hw.fan.speed.actual`, that way it can be compared to limit?

model/hardware/cpu-metrics.yaml
```diff
@@ -0,0 +1,38 @@
...
+    metric_name: hw.cpu.speed
+    # value_type: int
+    stability: development
+    brief: "CPU current frequency"
+    instrument: gauge
+    unit: "Hz"
+    extends: metric.hw.cpu
+
+  # TODO: decide on metric name speed.limit is original
+  - id: metric.hw.cpu.speed.limit
```
thompson-tomo, line 28: Maybe name follows battery and become `hw.cpu.threshold.xyz`

model/hardware/power-supply-metrics.yaml
```diff
@@ -0,0 +1,38 @@
+groups:
+  # POWER SUPPLY METRICS
+  - id: metric.hw.power_supply
```
thompson-tomo: Namespace conflict? Should It not become `metric.hw.power_supply.actual`, that way it can be compared to expected/limit?

model/hardware/temperature-metrics.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - id: metric.hw.temperature.common
+    type: attribute_group
+    brief: >
+      Common attributes for temperature metrics
+    extends: hardware.attributes.common
+    attributes:
+      - ref: hw.temperature.sensor_location
+        requirement_level: recommended
+
+  - id: metric.hw.temperature
```
thompson-tomo, line 12: Namespace conflict? Should It not become `metric.hw.temperature.actual`, that way it can be compared to expected/limit?

model/hardware/voltage-metrics.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - id: metric.hw.voltage.common
+    type: attribute_group
+    brief: >
+      Common attributes for voltage metrics
+    extends: hardware.attributes.common
+    attributes:
+      - ref: hw.voltage.sensor_location
+        requirement_level: recommended
+
+  - id: metric.hw.voltage
```
thompson-tomo, line 12: Namespace conflict? Should It not become `metric.hw.voltage.actual`, that way it can be compared to expected/limit?

bertysentry:
@thompson-tomo So there is no way we could consider `.limit` as a "special" metric that doesn't imply a namespace?

I am deeply skeptical of the decision that was made by the TC on this. To allow an hypothetical hierarchical representation of metrics in a JSON payload, we're now trying to invent many workarounds for this artificial limitation, with metrics like `hw.voltage.actual`, or `hw.voltage.current` (which is very very confusing).

The TC probably didn't realize the weird consequences of the decision of preventing a metric from being the prefix of another metric. A `xxx.yyy` metric will prevent any `xxx.yyy.zzz` metric further down the road, resulting in weird decisions like "let's rename `xxx.yyy` to `xxx.yyy.current` or `xxx.yyy.value` so that we can have `xxx.yyy.zzz`").

It will actually create a lot of instability in our semantic conventions!

Is there a way to "appeal" this decision? 😅

thompson-tomo:
When I look at `hw.voltage` that sounds alot like an entity which lacks a descriptor. Does it represent the expected voltage, actual voltage, minimum voltage etc.

model/hardware/fan-metrics.yaml
trisch-me: As discussed during the meeting I have moved everything as is to see if it would break something and to discuss it then separately

trisch-me:
@bertysentry let’s discuss it today during the meeting once more, I have moved everything to the yaml files and we have no problems right now with tooling (but because we don’t check namespace re-usage for the metrics). We can raise it once more to find the appropriate solution

model/hardware/battery-metrics.yaml
trisch-me: small update - I don’t change metric names in this PR. As current tooling allows to have them as is in metrics.

This doesn’t mean they will not change if we will agree on it.

trisch-me:
@bertysentry we have agreed that we merge it as is to make it pure refactoring and extracting the data from md files to the yaml files and discuss that we should come up with a solution later on how we proceed with collision
so I need you factual check if all metrics are ok and I haven’t missed anything

docs/system/hardware-metrics.md
```diff
@@ -47,7 +47,25 @@ Please note that this is an [ongoing process](https://github.com/open-telemetry/
...
 
-## Metric instruments
+## Metric Instruments
+
+### `hw.host.` - Physical host metrics
+
+**Description:** Physical system as opposed to a virtual system or a container.
+Examples: physical server, switch or disk array.
+
+| Name                          | Description                                                                                                                                           | Units | Instrument Type ([*](/docs/general/metrics.md#instrument-types)) | Value Type | Attribute Key(s) | Attribute Values |
```
lmolkova: this is hand-written, which seems to conflict with the PR title that it covers all HW metrics.

model/hardware/battery-metrics.yaml
```diff
@@ -0,0 +1,64 @@
+groups:
+  # BATTERY METRICS
+  - id: metric.hw.battery
```
lmolkova: nit: `metric` prefix is usually used for metrics, not attribute groups


```suggestion
  - id: attributes.hw.battery
```

model/hardware/gpu-metrics.yaml
```diff
@@ -0,0 +1,98 @@
...
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: development
+    brief: "GPU memory used"
+    instrument: updowncounter
+    unit: "By"
+    extends: metric.hw.gpu
+
+  - id: metric.hw.gpu.power
```
bertysentry: `hw.gpu.power` can actually be removed. It was a mistake in the first version of these semconvs. Or we remove it after this is merged! 😉

model/hardware/physical-disk-metrics.yaml
```diff
@@ -0,0 +1,62 @@
...
+    annotations:
+      code_generation:
+        metric_value_type: double
+    stability: development
+    brief: "Endurance remaining for this SSD disk"
+    instrument: gauge
+    unit: "1"
+    extends: metric.hw.physical_disk
+    attributes:
+      - ref: hw.physical_disk.state
```
bertysentry, line 32: We may actually need `hw.physical_disk.endurance.state`

model/hardware/power-supply-metrics.yaml
bertysentry: @thompson-tomo this is the group/namespace definition, not a metric 😉

model/hardware/power-supply-metrics.yaml
```diff
@@ -0,0 +1,42 @@
...
+    brief: "Maximum power output of the power supply"
+    instrument: updowncounter
+    unit: "W"
+    extends: metric.hw.power_supply
+    attributes:
+      - ref: hw.limit_type
+        examples: ["max", "critical", "throttled"]
+        requirement_level: recommended
+
+  - id: metric.hw.power_supply.utilization
```
bertysentry, line 32: Possibly add `hw.power_supply.usage` in Watts:

```yaml
  - id: metric.hw.power_supply.usage
    type: metric
    metric_name: hw.power_supply.usage
    annotations:
      code_generation:
        metric_value_type: int
    stability: development
    brief: "Current power output of the power supply"
    instrument: updowncounter
    unit: "W"
    extends: metric.hw.power_supply
```

model/hardware/battery-metrics.yaml
```diff
@@ -0,0 +1,64 @@
...
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: development
+    brief: "Time left before battery is completely charged or discharged"
+    instrument: gauge
+    unit: "s"
+    extends: metric.hw.battery
+    attributes:
+      - ref: hw.state
```
bertysentry, line 65: I am in favor of having a separate attribute here for battery charging/discharging/idle state: `hw.battery.state`.

model/hardware/registry.yaml
```diff
@@ -92,18 +92,520 @@ groups:
...
               value: 'failed'
               brief: 'Failed'
               stability: development
+            - id: needs_cleaning
+              value: 'needs_cleaning'
+              brief: 'Needs Cleaning'
+              stability: development
+            - id: ok
+              value: 'ok'
+              brief: 'Ok'
```
bertysentry:
```suggestion
              brief: 'OK'
```

model/hardware/registry.yaml
```diff
@@ -92,18 +92,520 @@ groups:
       - id: hw.state
         type:
           members:
-            - id: ok
-              value: 'ok'
-              brief: "Ok"
+            - id: charging
```
bertysentry: `charging` and `discharging` need to be moved to a new `hw.battery.state` attribute (or after this PR).

model/hardware/registry.yaml
```diff
@@ -92,18 +92,520 @@ groups:
...
+              brief: 'Ok'
+              stability: development
+            - id: predicted_failure
+              value: 'predicted_failure'
+              brief: 'Predicted Failure'
+              stability: development
         stability: development
         brief: >
           The current state of the component
+      - id: hw.limit_type
```
bertysentry, line 133: `hw.limit_type` should be able to receive additional "custom" values, vendor-specific. I appreciate that we enforce conventions for values like `critical`, or `degraded`, but instrumentation developers will surely encounter or create new types of limit.

Is it possible to have an open enum?

DMTF's CIM had such open enumerations, where lower values were defined by the DMTF, and OEM could add OEM-specific values... above 16378 (urgl). It would be practical to have a similar way to extend enums (but more elegant 😁)

model/hardware/registry.yaml
```diff
@@ -92,18 +92,520 @@ groups:
...
+              value: 'remaining'
+              brief: "Remaining"
+              stability: development
+        stability: development
+        brief: >
+          State of the physical disk endurance utilization
+      - id: hw.physical_disk.smart_attribute
+        type:
+          members:
+            - id: raw_read_error_rate
```
bertysentry: These are merely examples of SMART attributes. We cannot restrict the value of this attribute.

model/hardware/registry.yaml
bertysentry, line 1: We only need one `hw.vendor` attribute. There is no need to have `hw.cpu.vendor`, `hw.memory.vendor`, `hw.network.vendor`, etc. Same suggestion for `hw.model`, instead of `hw.cpu.model`, `hw.memory.model`, `hw.network.model`, etc.

I will recommend to have only one `hw.serial_number` attribute as well (instead of `hw.cpu.serial_number`, `hw.memory.serial_number`, `hw.network.serial_number`, etc.) because it makes the querying and searching of hardware references much easier.

Querying a metric like `hw.status{hw.state="present"}` will result in a nice table with all hardware components with vendor, model, and serial number. Very easy to match with an asset table.

Other attributes like `hw.firmware_version`, `hw.bios_version`, `hw.driver_version` could also make things easier than having separate attributes for each component type (CPU, disk controller, GPU, network card, etc.)

model/hardware/registry.yaml
thompson-tomo: All enums are open by default.

model/hardware/registry.yaml
thompson-tomo: Already one step ahead of you on vendor, hence have #2420 which I will need to update after this.

model/hardware/registry.yaml
trisch-me: @bertysentry to this comment and the one above - all enums are open. These are the values we provide but they can be extended to any value. 
I can also change enum value to just string with a few examples. Will this better suit the purpose of this attribute?

model/hardware/registry.yaml
trisch-me: @bertysentry so you suggest to have those additional attributes merged all together and not assigned to specific types of hardware? Makes sense, let me rewrite it for this PR. It’s not a material change

thompson-tomo: @trisch-me it would be great if you could take on my proposed attributes in #2420 ie hw.vendor.name as per the guidance

model/hardware/registry.yaml
bertysentry: That's great, thank you for the clarification! 😊

model/hardware/registry.yaml
bertysentry: Oh! I remember this PR now! Thank you, @thompson-tomo! I've added a couple comments there.

As it's done with many existing instrumentation SDKs and APIs, `hw.vendor`, `hw.model` and `hw.serial_number` are good enough.

trisch-me: @bertysentry @thompson-tomo I have consolidated vendor, model and other common attributes so now we have only one in global registry

@thompson-tomo do we really need id there? We used to not add additional complexity until we really need it i.e. there is an instrumentation that uses it.

model/hardware/gpu-metrics.yaml
trisch-me: removed

model/hardware/physical-disk-metrics.yaml
trisch-me: will it break anything if removed? I would propose to move it to additional PR

model/hardware/power-supply-metrics.yaml
trisch-me: should it be `usage` or just `hw.power_supply` as in other similar metrics

model/hardware/battery-metrics.yaml
trisch-me: we are working on enums in tooling so that every specific enum could be derived from some base enum. 
if you are worrying about enum having more values than needed - this should solve this problem

Or you had something else in mind?

model/hardware/registry.yaml
trisch-me: the question still remains open - is it better to leave it as enum or as a sting with just a few examples?

model/hardware/registry.yaml
```diff
@@ -92,18 +93,308 @@ groups:
...
+              value: 'receive'
+              brief: "Receive"
+              stability: development
+            - id: transmit
+              value: 'transmit'
+              brief: 'Transmit'
+              stability: development
+        stability: development
+        brief: >
+          Direction of data transfer
```
thompson-tomo: We should be using [network-io-direction](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/#network-io-direction) rather than a new field

model/hardware/registry.yaml
thompson-tomo: So to help things to progress and hopefully merged I would suggest we
- use hw.model.name rather than hw.model
- use hw.vendor.name rather than hw.vendor
- Do not worry about I'd in this pr

As I can see use cases for attributes such as hw.model.family and that way we are following the guidance on naming 
> When introducing a name describing a certain property of the object, include the property name. For example, use file.owner.name instead of file.owner

model/hardware/registry.yaml
trisch-me: Nice suggestion, I think it makes sense to re-use existing attribute. 
@bertysentry any comments on it?

model/hardware/registry.yaml
trisch-me: I understand that having here “*.name” would help you with your PR later but I’m not convinced that we need id. While `file.owner` is some user which might have internally id I don’t understand what IDs we are talking about for model or vendor and who will be filling them up or using them. Do you have maybe examples?

thompson-tomo: It appears that auto correct changed id to I'd in my last comment and I didn't notice.  Let's discuss id at a later point and here just do *.name

model/hardware/registry.yaml
bertysentry: Yep, great suggestion! I'm in favor of this!

model/hardware/registry.yaml
bertysentry: @thompson-tomo For the sake of this PR, I suggest we keep simple `hw.vendor` and `hw.model`.

I'll make another suggestion in your PR so we can cover all use cases without over-engineering.

thompson-tomo: I don't see how it complicates it especially when the guidance is clear on naming. If we add hw.vendor now, even if we deprecate it later it will stick around for eternity.

model/hardware/physical-disk-metrics.yaml
bertysentry: If there is nothing else referring to `hw.physical_disk.state` then there is no issue to simply rename it. We can do this in another PR if you prefer.

model/hardware/power-supply-metrics.yaml
bertysentry: It's really `usage`, to be compared to `limit` so we can calculate its `utilization`.

model/hardware/battery-metrics.yaml
bertysentry: No, it's really that `hw.state` is about describing the generic health status of a hardware component: OK, degraded, failed, predicting failure, etc.

A battery can be at the same time OK, and either charging or discharging. So we need 2 separate attributes: one for describing the battery health, another one for describing whether it's charging, discharging or idle. These are 2 independent properties.

This can be done in a separate PR.

model/hardware/registry.yaml
bertysentry: String with a few examples is much simpler to maintain for us, quite frankly.

model/hardware/registry.yaml
bertysentry: As being personally involved in the development of a software that implements these semantic conventions, I am in favor of keeping `hw.vendor`, `hw.model`, and `hw.serial_number` for the eternity 😅

trisch-me:
During todays’ semconv meeting we have discussed that we will proceed as is with current changes (not changing anything substantial) and there was a request to create a hardware SIG to discuss any other changes to this domain.

@bertysentry @thompson-tomo and everyone else interested

docs/hardware/battery.md
```diff
@@ -0,0 +1,255 @@
...
+| `needs_cleaning` | Needs Cleaning | ![Development](https://img.shields.io/badge/-development-blue) |
+| `ok` | OK | ![Development](https://img.shields.io/badge/-development-blue) |
+| `predicted_failure` | Predicted Failure | ![Development](https://img.shields.io/badge/-development-blue) |
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Metric: `hw.status` (Battery)
```
lmolkova, line 181: can we avoid defining it multiple times? E.g. could it be defined once in common and if it's important to document some notes in each file, we can keep some notes here and just add a link to the common section?

model/hardware/physical-disk-metrics.yaml
trisch-me: we do have metric `hw.physical_disk.endurance_utilization`, should it be `hw.physical_disk.endurance.utilization` then? I think the best is to move this discussion out of this PR and update it later after PR is merged, if needed

model/hardware/power-supply-metrics.yaml
trisch-me: let’s move this new metric to the new PR too!

trisch-me: btw is it really int? shouldn’t it be double?

model/hardware/battery-metrics.yaml
trisch-me: @bertysentry I am willing to add this change to current PR as it’s actually my fault that I interpreted md file as such.
But I don’t understand one thing - from your words it looks like battery metric should have two different states attributes - global one `hw.state` and battery one `hw.battery.state` with just charging or discharging values. Is it correct?

model/hardware/registry.yaml
trisch-me: as per last meeting we leave these values as is for now with possibility to discuss them later

model/hardware/registry.yaml
trisch-me: Updated

model/hardware/registry.yaml
trisch-me: removed this attribute in favor for `network.io.direction`

model/hardware/battery-metrics.yaml
bertysentry: @trisch-me Yes, this is correct! (and it's not your fault if my explanations and requests are confusing 😅)

trisch-me: @bertysentry So based on the metric (`"Time left before battery is completely charged or discharged”`) I left there only one metric, i.e. `hw.battery.state`. Could you please check if any metric required second global state as well?

model/hardware/power-supply-metrics.yaml
bertysentry: Yeah, I agree it should be double. The range of values is typically \[0 - 2000W].

trisch-me: I added it

model/hardware/battery-metrics.yaml
trisch-me: I think you could just re-check battery metrics once more to make sure everything is correct

trisch-me:
I have also updated github link checker to exclude removed files from check

.github/workflows/reusable-link-check.yml
```diff
@@ -26,7 +26,8 @@ jobs:
         run: |
           merge_base=$(git merge-base origin/${{ github.base_ref }} HEAD)
           # Using lychee's default extension filter here to match when it runs against all files
-          modified_files=$(git diff --name-only $merge_base...${{ github.event.pull_request.head.sha }} \
+          # --diff-filter=d excludes deleted files
```
lmolkova, line 29: is this intentional?

docs/hardware/battery.md
```diff
@@ -0,0 +1,261 @@
...
+| [`hw.model`](/docs/registry/attributes/hardware.md) | string | Descriptive model name of the hardware component | `PERC H740P`; `Intel(R) Core(TM) i7-10700K`; `Dell XPS 15 Battery` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`hw.name`](/docs/registry/attributes/hardware.md) | string | An easily-recognizable name for the hardware component | `eth0` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`hw.parent`](/docs/registry/attributes/hardware.md) | string | Unique identifier of the parent component (typically the `hw.id` attribute of the enclosure, or disk controller) | `dellStorage_perc_0` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`hw.vendor`](/docs/registry/attributes/hardware.md) | string | Vendor name of the hardware component | `Dell`; `HP`; `Intel`; `AMD`; `LSI`; `Lenovo` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
```
thompson-tomo: What benefit do these table bring given attributes are shown under each metric? Could we not remove the table & instead have an entity describing the hw type? The entity can even be associated with the metric

trisch-me: this is autogenerated text made by tooling, it supposed to give you all available information for every metric so you don't need to look around. 
I agree it looks a bit ugly, you can raise this question, but it’s not relevant to this PR

thompson-tomo: The table is the attributes not a list of metrics. 😉 and it is being manually triggered to be added by line 27.

trisch-me: I am not sure I do understand. Those are common attributes supported by these metrics. 
You want just remove this whole block?

thompson-tomo: Yes remove that block as they are also shown on the individual metrics. Also some of those attributes won't be relevant to all metrics hence rather use the attributes on the metric to avoid confusion.

trisch-me: it’s not visible in this PR but this part came from original `hardware-metrics.md` file
As it was 1 file for all the metrics, the beginning had [common attributes](https://github.com/open-telemetry/semantic-conventions/commit/dff7b22a083b7d05e09c95ef078bccf3b9e0a9c3#diff-700b4c150a752bd63389ec6aa7dd0f9846e56f03f5daaf7952e8f86cb578e4baL47-L78) for all metrics and it’s now moved to every single metric file.

Also as per original file these attributes might be used for all metrics. Every additional metric might add specific to that metric attributes.

Due to the nature of our auto generated tables we render all attributes for metrics every time including inherited. This is what I referred to in my first comment and this is what we might to discuss separately.

trisch-me:
@joaopgrassi @trask @AlexanderWert @lmolkova could you please check this PR, thanks

docs/hardware/battery.md
```diff
@@ -0,0 +1,261 @@
...
+
+- [Battery metrics `hw.battery.*`](#battery-metrics-hwbattery)
+  - [Metric: `hw.battery.charge`](#metric-hwbatterycharge)
+  - [Metric: `hw.battery.charge.limit`](#metric-hwbatterychargelimit)
+  - [Metric: `hw.battery.time_left`](#metric-hwbatterytime_left)
+  - [Metric: `hw.status` (Battery)](#metric-hwstatus-battery)
+
+<!-- tocstop -->
+
+## Battery metrics `hw.battery.*`
```
lmolkova, line 19: nit: is the heading necessary given that all metrics in this file are battery metrics?

docs/hardware/cpu.md
```diff
@@ -0,0 +1,276 @@
...
+
+- [CPU metrics `hw.cpu.*`](#cpu-metrics-hwcpu)
+  - [Metric: `hw.errors` (CPU)](#metric-hwerrors-cpu)
+  - [Metric: `hw.cpu.speed`](#metric-hwcpuspeed)
+  - [Metric: `hw.cpu.speed.limit`](#metric-hwcpuspeedlimit)
+  - [Metric: `hw.status` (CPU)](#metric-hwstatus-cpu)
+
+<!-- tocstop -->
+
+## CPU metrics `hw.cpu.*`
```
lmolkova, line 19: nit: is this heading necessary?

docs/hardware/enclosure.md
```diff
@@ -0,0 +1,124 @@
...
+**Status**: [Development][DocumentStatus]
+
+<!-- toc -->
+
+- [Enclosure metrics `hw.enclosure.*`](#enclosure-metrics-hwenclosure)
+  - [Metric: `hw.status` (Enclosure)](#metric-hwstatus-enclosure)
+
+<!-- tocstop -->
+
+## Enclosure metrics `hw.enclosure.*`
```
lmolkova, line 16: nit: also here and in other files

model/hardware/battery-metrics.yaml
```diff
@@ -0,0 +1,72 @@
...
+          optimal charge, the shutdown threshold, or energy-saving thresholds. These values
+          are typically provided by the hardware or firmware to guide safe and efficient battery
+          usage.
+
+  - id: metric.hw.battery.time_left
+    type: metric
+    metric_name: hw.battery.time_left
+    annotations:
+      code_generation:
+        metric_value_type: int
```
lmolkova: nit, maybe 

```suggestion
        metric_value_type: double
```

?

.github/workflows/reusable-link-check.yml
trisch-me: yes, current check [fails](https://github.com/open-telemetry/semantic-conventions/actions/runs/16372025805/job/46262518108) because it can’t find deleted md file and I’m deleting here an old manual md file.

docs/hardware/battery.md
trisch-me: It was copy pasted from old file. I think it plays well with description of components

docs/hardware/battery.md
trisch-me: not every component has status metric and moreover every component might have it’s own set of attribute values i.e. `battery` has `charging` or `discharging`, `tape_drive` has `needs_cleaning ` etc.
Therefore I do add a description of allowed values for each component

docs/system/hardware-metrics.md
trisch-me: hm, this file is deleted in this PR, maybe it has shown you old commit where it was still there?

model/hardware/battery-metrics.yaml
trisch-me: @bertysentry should it be int or double? It is in seconds but I am not sure if we are using fractions there

bertysentry: Good point. I guess it could be theoretically be `double` (if the timer is originally measured in milliseconds), although I don't see any practical use case where one would want such precision for the time left on a battery.

So I suggest we simply keep it consistent with other metrics: if other metrics in semconv (system, http, etc.) use `double` for time metrics, we can use `double` here as well.

trisch-me: updated to double

trisch-me:
any new comments on this? Otherwise I would like it to be merged

PR #2381, opened 2025-06-17
Adds more geo attributes #2371

### LINKED ISSUE #2371: Informative info about location
### Area(s)

area:geo

### What's missing?

The ability to provide more informative information about where somthing is located

Introduction of new attributes

- geo.name -> as per https://www.elastic.co/docs/reference/ecs/ecs-geo#field-geo-name
- geo.timezone -> as per https://www.elastic.co/docs/reference/ecs/ecs-geo#field-geo-timezone this is important when a country has mutliple timezone's
- geo.location -> which is either a user definable string or a kvp option

thompson-tomo:
Fixes #2371 

## Changes

This introduced 2 new defined attributes with both being migrations from ECS:

- geo.location.name which is free text and useful for naming the data centre the software is running in
- geo.timezone for capturing the timezone

To cater for data centres annotations are introduced so that a user can specify any additional information such as rack, cage location.


thompson-tomo:
Awaiting review

thompson-tomo:
@trisch-me another pr which is an ecs migration with only 1 additional field "annotations" added in.

trisch-me:
hey @thompson-tomo thanks for taking care of it.
Please check initial [geo pr](https://github.com/open-telemetry/semantic-conventions/pull/1116) to understand why some fields were not migrated. If there are more arguments to why add them now, let’s discuss them during semconv meeting

thompson-tomo:
So timezone is very useful, for instance I have run into issues where server is 1 timezone, support is in another & end user is in a third. To further compound things the server exited summer time during the investigation.

In relation to the name fields for the codes, these are still documented as being emitted by the collector. If proposal is not to migrate them from ecs is might be a good idea to mark them as such on the ecs website.

My suggestion would be that we put a note on it indicating that the names are for usage in resource detectors etc.

Issue #2346, opened 2025-06-09
App details lack key info


thompson-tomo:
### Area(s)

area:app

### What's missing?

The ability to specify the name, id & version of an application generating telemetry.

### Describe the solution you'd like

New attributes

- `app.id`
- `app.name`
- `app.namespace`
- `app.version`
- `app.type` -> web, GUI, console
- `app.architecture` which is a migration of [`package.architecture`](https://www.elastic.co/docs/reference/ecs/ecs-package#field-package-architecture)

Issue #2345, opened 2025-06-09
Enable applications/services to be grouped together


thompson-tomo:
### Area(s)

area:app, area:cloud, area:cloudfoundry, area:service

### What's missing?

A generic way for applications to be grouped together

### Describe the solution you'd like

Additional attributes to:

- Specify a group/collection that the app/Service belongs to on the cloud platform hence goes in the cloud namespace
- Specify what part of the organisation is responsible for the app/Service via aattribute within the app & service namespace.

These changes can lead to the deprecation of cloudfoundry.org.* & cloudfoundry.space.* namespace as these can be used by any platform.

Issue #2339, opened 2025-06-08
Provide namespace for the module being used


thompson-tomo:
### Area(s)

area:new

### What's missing?

Ability to know what library/module performed the operation be it feature flag evaluation, sending of messages. 

Example would be a feature flag might be defined in Microsoft Feature manager but the evaluation has occured in open feature. In that case the feature flag provider would be MS as that is who provided the details but I want to know that open feature triggered & processed the request.

### Describe the solution you'd like

Introduction of a new collection of attributes:

- `module.id`
- `module.name`
- `module.type`
- `module.version`

### Additional context

Porting ECS -> https://www.elastic.co/docs/reference/ecs/ecs-dll
Needs to cater for
https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/active_model_serializers/README.md

thompson-tomo:
Similar to #2184

Issue #2184, opened 2025-04-25
Semantic Convention for module attribution


michaelsafyan:
### Area(s)

area:code

### What's missing?

There isn't a way to refer to a program "module".

For example, let's suppose I want to record the `__module__` attribute of a Python function, there isn't a good place to do so.

### Describe the solution you'd like

I'd like a `code.module` convention to be added.

michaelsafyan:
There is a draft PR related to Google Gen AI SDK instrumentation that illustrates how this might be used:

 - https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3446/

joaopgrassi:
Closing in favor of #2339

PR #2384, opened 2025-06-17
Define GCP client and server spans.

michaelsafyan:
## Changes

Defines GCP client and server spans, including new attributes:

 - `gcp.project_id`
 - `gcp.resource.name`
 - `gcp.server.service`

This aims to reflect several separate designs/docs internal to Google into OSS, establishing appropriate expectations for GCP telemetry in the community as well as for GCP teams emitting this telemetry.

### DIFF
--- a/.chloggen/gcp_spans.yaml
+++ b/.chloggen/gcp_spans.yaml
@@ -1,4 +1,4 @@
 change_type: enhancement
 component: gcp
 note: Defines GCP spans, including new attributes `gcp.project_id`, `gcp.resource.name`, and `gcp.server.service`.
-issues: []
+issues: [2384]


model/gcp/registry.yaml
```diff
@@ -1,4 +1,37 @@
...
+       examples: ['//somedomain.googleapis.com/projects/some-project/locations/us-central1/instances/12345']
+       note: >
+         The attribute "gcp.resource.name" and the resource attribute "cloud.resource_id"
+         are related but have different purposes. Use "cloud.resource_id" to identify the
+         compute resource where the code is running. Use "gcp.resource.name" to identify
+         the GCP resource that said code is manipulating. For example, if you have code
+         that is running in Cloud Run and the code publishes to a topic in Cloud Pub/Sub,
+         you should use "cloud.resource_id" to refer to the Cloud Run instance where the
+         code is running, and you should use "gcp.resource.name" to identify The
+         Cloud Pub/Sub topic to which your code is actively publishing.
```
thompson-tomo, line 34: Why wouldn't you use the peer namespace or even messaging/db

model/gcp/registry.yaml
```diff
@@ -1,4 +1,37 @@
...
+    attributes:
+     - id: gcp.project_id
+       type: string
+       stability: development
+       brief: Telemetry storage project.
+       examples: ['my-project', 'other-project']
+       note: >
+        This attribute governs the destination project to which telemetry data should be routed.
+        It may differ from the "consumer" project (used for quota, billing) or the "resource" project (the
+        project containing the resource(s) on which the operation being reported was performed).
```
thompson-tomo, line 15: Is this not what service.namespace is for and if not would #2345 solve this?

model/gcp/registry.yaml
```diff
@@ -15,6 +48,18 @@ groups:
...
+    display_name: GCP Server Attributes
+    brief: >
+      Attributes for Google Cloud servers.
+    attributes:
+      - id: gcp.server.service
+        type: string
+        stability: development
+        brief: Identifies the Google Cloud service that is emitting the telemetry.
+        note: Server analogue of 'gcp.client.service'.
+        examples: ['appengine', 'run', 'firestore', 'alloydb', 'spanner']
```
thompson-tomo, line 62: Is this not what `service.*` is for?

model/gcp/spans.yaml
```diff
@@ -0,0 +1,110 @@
...
+- id: span.gcp.server
+  type: span
+  span_kind: server
+  stability: development
+  brief: A server-side span emitted by GCP services.
+  attributes:
+  - ref: rpc.service
+    requirement_level: required
+  - ref: rpc.method
+    requirement_level: required
```
thompson-tomo, line 11: Is this really rpc?

thompson-tomo (changes requested):

model/gcp/registry.yaml
michaelsafyan: The `messaging` and `peer` namespaces may be relevant to a subset of operations but are not universally applicable to all of the operations, and they have different semantics than the conventions here.

Some considerations relevant to this data model:

 - Multiple layers:
   - The customer application code
   - The GCP infrastructure that the application code is running on
   - The GCP services/infrastructure that the customer application code talks to

The `service.name` and `peer.service` generally correspond to the "customer application code" layer with a name that belongs to the customer. The `cloud.resource_id` and `gcp.resource.name` refer to formal concepts owned by GCP and that typically map to underlying infrastructure components that comprise the "service".

model/gcp/registry.yaml
michaelsafyan: No, these are different concepts.

The `service.*` describes what service performed the operation. The `gcp.project_id` attribute governs where to route the telemetry data. Conceptually, the GCP resource hierarchy stores "resources" (like Cloud Trace buckets) under "projects" which are grouped into "folders" and "orgs".

It is possible for an operation to occur in a compute environment hosted in project A, while routing the telemetry data to the Cloud Trace storage bucket in project B. This is often done as a means of centralization (to provide a single pane of glass by routing telemetry from multiple projects to a common observability project).

The `gcp.project_id` parameter is a mandatory attribute required as a control signal to determine where to send the telemetry data. As can be seen from the OTLP migration guide here:

https://cloud.google.com/stackdriver/docs/instrumentation/migrate-to-otlp-endpoints

This is already implemented, and the instructions include:

```
export OTEL_RESOURCE_ATTRIBUTES="gcp.project_id=PROJECT_ID"
```

... in order to make the OTLP exporter compatible with `telemetry.googleapis.com`. This simply aims to better formally document this fact for better discoverability as well as to help clarify the meaning of this project value (in relation to `cloud.account.id` as well as the project which may appear in `cloud.resource_id`).

model/gcp/registry.yaml
michaelsafyan: Not quite, no.

It is important to consider that there are layers:

 - The customer application (using `service.*`)
 - The GCP infrastructure/service that the customer application runs on top of

For example, consider a span emitted from Cloud Run; the expectation would be that `service.*` reports the information about the customer service that is implemented on top / running within Cloud Run. The `gcp.server.service` would provide `run`.

In terms of similarities to existing OTel conventions, it is most similar to `cloud.platform`. However, this resolves the case where one GCP service is implemented on top of another one; `gcp.server.service` can report the higher level service provided, while `cloud.platform` may report the underlying infrastructure it is built on.

model/gcp/spans.yaml
michaelsafyan: Yes. All GCP APIs are defined using RPC protobuf definitions (even if an API gateway layer exposes these via HTTP/REST in addition to gRPC).

model/gcp/registry.yaml
thompson-tomo: Ok, that naming/description wasn't clear to me. Based on the subsequent information ideally we could have either telemetry.namespace.id or gcp.telemetry.namespace.id that way it is clear.

model/gcp/registry.yaml
thompson-tomo: Ok, I am having difficulty seeing where peer is not applicable.

> The service.name and peer.service generally correspond to the "customer application code" layer with a name that belongs to the customer.

Not sure about the "customer application code" part especially when the description states

> If the value was not specified, SDKs MUST fallback to unknown_service: concatenated with process.executable.name, e.g. unknown_service:bash. If process.executable.name is not available, the value MUST be set to unknown_service.

Based on that if pub sub was producing telemetry that would be the service.name.

In relation to peer this actually says
> 	The service.name of the remote service. SHOULD be equal to the actual service.name resource attribute of the remote service if any.

Hence could be pub sub.

model/gcp/registry.yaml
thompson-tomo: Again description is not conveying that additional info. I think my pr https://github.com/open-telemetry/semantic-conventions/pull/2367/files actually would cater for that scenario as it enables you to specify both the platform name and the underlying product used for that service. More than happy to extend the known values there.

model/gcp/spans.yaml
thompson-tomo: Ok then perhaps the span type should be internal then.

model/gcp/spans.yaml
```diff
@@ -0,0 +1,110 @@
...
+      conditionally_required: >
+        Required if the service is not a global service.
+  - ref: cloud.availability_zone
+    requirement_level:
+      conditionally_required: >
+        Required if the service is not a zonal service.
+  - ref: cloud.resource_id
+    requirement_level: required
+    note: This resource should correspond with an AppHub service or workload.
+- id: span.gcp.client
```
thompson-tomo, line 43: Would suggest splitting this into grpc & http spans

model/gcp/registry.yaml
```diff
@@ -1,4 +1,37 @@
 groups:
+  - id: registry.gcp.telemetry_control
+    type: attribute_group
+    display_name: Telemetry Control
+    brief: Attributes for controlling telemetry behavior
+    attributes:
+     - id: gcp.project_id
```
thompson-tomo, line 7: Note as per [naming](https://opentelemetry.io/docs/specs/semconv/general/naming/) the underscore should be avoided.

model/gcp/registry.yaml
michaelsafyan: I think Cloud Pub/Sub may not be the best example here, because you don't do compute on Cloud Pub/Sub.

The goal here is to have a convention that works across GCP services of every kind, including compute platforms like Cloud Run, App Engine, Cloud Functions, Google Kubernetes Engine (GKE), Google Compute Engine (GCE), etc. in addition to services like Cloud Pub/Sub, Vertex AI, and the 150+ cloud services on GCP:

https://cloud.google.com/products

It would require very sophisticated interpretation of the value contained in `peer.service` or `service.name` (and other contextual hints and other attributes) to understand "what is the logical GCP service that is responding to this request" which is answered by `gcp.server.service`. The `gcp.server.service` attribute makes this very easy to understand consistently across any GCP service.

As for `gcp.resource.name`, the resource name is not a "service". It is a named object owned and managed by a service. Some GCP services own the deployment and management of other services, where "resource"s do correspond to a "service" (e.g. resources in Cloud Run include user-deployed services that are deployed on Cloud Run), but not every `gcp.resource.name` is a "service". For example, in the context of Cloud Pub/Sub, an example of a resource might include a topic or a subscription to a topic. Absent `gcp.resource.name` there would be no consistent approach to answering "what GCP resource is being operated on by this operation?". Each specific service may report related information as service-specific attributes (e.g. although not presently reported today, you could imagine Cloud Pub/Sub modelling the topic being operated on with a combination of `gcp.pubsub.topic.id`, `gcp.pubsub.topic.project`, `gcp.pubsub.topic.location`, etc.); however, this makes it difficult to interpret across different GCP services in a consistent fashion (because it requires knowing lots of information about each of the different services to piece together its corresponding formal resource name). The formal resource name is relevant in the context of certain common management APIs, which makes it especially useful to have a very quick and easy way to consistently record and obtain the formal resource name across services.

model/gcp/registry.yaml
michaelsafyan: I'm a little bit confused by this proposal. What are you suggesting that `telemetry.namespace.id` or `gcp.telemetry.namespace.id` represent? And how do you see this addressing the routing goal of `gcp.project_id`?

model/gcp/registry.yaml
michaelsafyan: I don't see how the linked PR would address the goal here. Perhaps I'm missing something. Could you elaborate on how this proposal would make it easy to identify the "logical GCP service being interacted with" in cases where GCP services are built on GCP (e.g. where, say, a service like `industry-specific-service.googleapis.com` is implemented on top of Cloud Run, where `cloud.platform` or your proposed `hosting.platform.name` would reference Cloud Run, because that is the hosting environment)?

model/gcp/spans.yaml
michaelsafyan: I don't think `INTERNAL` would be correct here. The goal is to have a single `SERVER` span representing the end-to-end timing of the computation within GCP, reporting the relevant details of what happened in GCP. The system components of how it is actually implemented are internal details that are not necessarily being reported. While internal details would make sense as `INTERNAL`, the overall span for the end-to-end operation that received the inbound request makes sense to report as `SERVER`.

The important thing here is that GCP abstracts the "transport" layer and conceives of all operations (regardless of the "transport") as well-defined RPC services, even when the "transport" is over HTTP. [Also, side note: it is worth noting that gRPC, itself, is implemented on top of HTTP/2; for that reason, it should not be strange or disallowed for `rpc.*` and `http.*` attributes to exist on the same span].

model/gcp/spans.yaml
michaelsafyan: Why? Aside from the fact that gRPC is implemented on top of HTTP, GCP client and server components generally abstract the "transport" layer. In a lot of cases, the thing being instrumented only knows that it is implementing an RPC service as well as the mapping between the RPC services declaration and HTTP REST/JSON interfaces. Not every component necessarily has access to information about which "transport" implementation is being used.

Separating them would be a hindrance to feasibility of implementation for the instrumentation.

model/gcp/registry.yaml
michaelsafyan: Noted. If this were "proscriptive" (defining what the telemetry SHOULD BE prior to implementation), then I would enthusiastically agree and rename. However, this particular one (`gcp.project_id`) is "descriptive" rather than "proscriptive". I am aiming to document a convention that already exists (but is simply not captured in OTel docs).

Moreover, this "convention" is not merely a "convention" but is actually a hard requirement of the Telemetry API (`telemetry.googleapis.com`), making it somewhat impractical to change.

model/gcp/spans.yaml
thompson-tomo: If we look at the attributes there is a clear split of use cases if grpc these are needed but if http rest is needed, these are needed.

> Not every component necessarily has access to information about which "transport" implementation is being used.

If that is the case, how can you set those required attributes which changes based on if it is grpc or http rest? 

For me if the transport logic is abstracted away, then perhaps the span should be created in the abstraction once it is known. At the same time for me the http span should be an extension of the generic http client span https://opentelemetry.io/docs/specs/semconv/http/http-spans/ and the grpc extends the rpc equivalent https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/ hence more reason to split.

model/gcp/spans.yaml
thompson-tomo: > The goal is to have a single SERVER span representing the end-to-end timing of the computation within GCP

That seems to contradict the definition of a span 
https://opentelemetry.io/docs/specs/otel/trace/api/#spankind in particular 

> callers SHOULD arrange that a single Span does not serve more than one purpose. For example, a server-side span SHOULD NOT be used to describe outgoing remote procedure call. 

Once the above principle is being followed the span kinds should be checked as producer & consumer spans are missing.

> it is worth noting that gRPC, itself, is implemented on top of HTTP/2; for that reason, it should not be strange or disallowed for rpc.* and http.* attributes to exist on the same span

What you are thinking of is the network namespace which captures the protocol etc. Whereas the http namespace focuses the request/response just like grpc.

model/gcp/registry.yaml
thompson-tomo: `Telemetry.namespace.id` would represent the namespace to put the telemetry into which for gcp would be a project. For ELK it would be the space etc.

model/gcp/registry.yaml
thompson-tomo: > It would require very sophisticated interpretation of the value contained in peer.service or service.name (and other contextual hints and other attributes) to understand "what is the logical GCP service that is responding to this request"

I think you misunderstood, peer.service would always identify the service a request is being sent to & service.name is what is generating the telemetry.

If I have an app in gce which publishes a pub sub message. The app span would have a service.name related to the app & peer.service of pubsub. When telemetry is generated in pubsub for the message which was sent the service.name becomes pubsub & peer is not present. Hence no sophisticated interuptation needed.

For resource.name I would recommend looking at how other providers do it & instead strive for interoperability and consistency. Your comment below worries me as we already have attributes such as messaging.destination.name etc.
> you could imagine Cloud Pub/Sub modelling the topic being operated on with a combination of gcp.pubsub.topic.id, gcp.pubsub.topic.project, gcp.pubsub.topic.location, etc.)

model/gcp/registry.yaml
thompson-tomo: In the case I am assuming industry-specific-service functions like run but tailored for an industry. As such the product would be cloud run & the industry specific variant would be the product

model/gcp/registry.yaml
michaelsafyan: Is `telemetry.namespace.id` the type or the value? If it's the value, how do you interpret its meaning? And how would you use this for such a purpose if you are routing telemetry data to multiple systems?

model/gcp/registry.yaml
michaelsafyan: > In the case I am assuming industry-specific-service functions like run but tailored for an industry

That assumption is incorrect. If you go to `cloud.google.com/products` and then click "Industry", you can see a number of industry-specific services like "Cloud Healthcare API", "Recommendations builder", etc.

https://cloud.google.com/products?hl=en&pds=CAo

The point is that GCP may offer "Applications as a Service" (AaaS) services built on top of GCP "Infrastructure as a Service" (IaaS), "Platform as a Service" (PaaS), or "Functions as a Service" (FaaS) offerings.

The current `cloud.platform` (and your proposed `hosting.platform.name`) may be good fit for the lower-level substrate (IaaS/PaaS/FaaS) but doesn't necessarily well-align with the higher level AaaS. I think you were maybe thinking of `host.platform.product` serving this role, but it's a bit unclear what to do when communicating with the management plane of the IaaS/PaaS/FaaS offerings when they are functioning more like a AaaS and are not necessarily the hosting environment. The idea was to have  a single, uniform attribute for the "top-level GCP service being interacted with" regardless of whether it is AaaS, PaaS, IaaS, FaaS, etc.

model/gcp/spans.yaml
michaelsafyan: I think we may be speaking past each other. There is no overloading of Span `Kind`; the `SERVER` span is being used solely for the start of receiving an incoming request to replying with a response to the end user (with outbound requests, if they are being instrumented/sampled, being reported as `CLIENT` spans).

I realize that my previous message might be confusing, because I was using the term "transport" to refer to L7 application layer protocols and not L4 transport layer in the OSI model. What I am trying to say is that idiomatic, language-specific interfaces abstract over the details of the underlying application layer protocol. For GCP services as well as for clients/users of GCP, the distinction of gRPC vs HTTP is not particularly meaningful. Interaction with GCP is typically accomplished via official Google client libraries which abstract over the protocol of choice. Services in GCP are implemented in a way that makes the particular protocol invisible.

model/gcp/spans.yaml
michaelsafyan: For (the majority of) cases where the underlying is not known, it would report both.

The protobuf interface definition language used includes annotations (technically ["extensions"](https://protobuf.dev/programming-guides/proto2/#extensions) in the parlance of the protobuf language) that annotate fields with the details of how those fields are represented in HTTP. Without knowing whether the request at runtime is going over HTTP rather than gRPC transport, it is still possible to know such things as the "url.template" and other HTTP concepts/representations if they were used.

There is likewise a standardized mapping between gRPC status codes and HTTP status codes.

Google documents this publicly here:

 - https://google.aip.dev/127 - "HTTP and gRPC Transcoding"
 - https://google.aip.dev/193 - "Errors"

model/gcp/registry.yaml
thompson-tomo: In the pr for this concept (https://github.com/open-telemetry/semantic-conventions/pull/2403), I ended up going with the name `telemetry.destination.namespace` that way it is clear that this is related to the destination of the telemetry.

model/gcp/registry.yaml
thompson-tomo: As you are talking about applications, I already had raised https://github.com/open-telemetry/semantic-conventions/issues/2346 including a pr.

That way you have host.* to describe where it is running & app.* to describe what it is running in case of an app but in the case of faas you already have the faas namespace for that.

model/gcp/spans.yaml
thompson-tomo: > the `SERVER` span is being used solely for the start of receiving an incoming request to replying with a response to the end user (with outbound requests, if they are being instrumented/sampled, being reported as `CLIENT` spans).

This is only applicable for the client/server model in pubsub this should be producer & subscriber span kind etc.

model/gcp/spans.yaml
thompson-tomo: > For (the majority of) cases where the underlying is not known, it would report both.

How can that be so? The client library should know if it making a grpc request or a http request. If the request is a http to gateway which converts to grpc, that is still a http from the client but would be a grpc from the gateway to the service but that would be a different span.

model/gcp/registry.yaml
michaelsafyan: Left a comment on the PR. I'm uncertain about the feasibility of making this backend use `telemetry.destination.*` instead of `gcp.project_id`, but left some comments to increase possibility/alignment.

model/gcp/registry.yaml
michaelsafyan: Thank you for sharing the `app.*` attributes PR. That is really good to know about.

I think this would still need to be a logically separate attribute, though; for example, the GCP services (including "Application as a Service" applications) can sometimes have multiple instances that are associated with a customer's higher-level concept of an "application". I think it is likely that, if we were to use the `app.*` attributes, that they would refer to higher-level concepts of how  the user organizes their system (and deployed/instantiated services into applications), while `gcp.server.service` would be a lower-level concept than that.

model/gcp/spans.yaml
michaelsafyan: Right. As mentioned earlier, Cloud Pub/Sub is not the focus of this PR. I think it might make sense to first get agreement on the contents of this PR (which applies to the vast majority of Cloud APIs) and then, separately, address the PRODUCER/CONSUMER case for Cloud Pub/Sub.

model/gcp/spans.yaml
michaelsafyan: It is known at some point in both ends, but crucially it may not be known at the point in the code where instrumentation makes the most sense. Both ends use an interface that abstracts this detail or use code generation where the mechanics are delegated to functions whose implementations may vary based on code generation settings. This is why, from a feasibility standpoint, it is important to not require them to be separate.

thompson-tomo: Yes it might make things easier for library authors however this approach can and will lead to users easily misunderstanding the data especially when both set of metrics are being generated. 

As an idea, do we first document a library/module namespace and then use that as the span definition for gcp library calls.

model/gcp/registry.yaml
thompson-tomo: > the GCP services (including "Application as a Service" applications) can sometimes have multiple instances that are associated with a customer's higher-level concept of an "application". I think it is likely that, if we were to use the app.* attributes, that they would refer to higher-level concepts of how the user organizes their system (and deployed/instantiated services into applications), while gcp.server.service would be a lower-level concept than that.

That higher level would/could be expressed as the `app.namespace` which is a collection of apps. that way you can use `app.name`/`app.id` to refer to the running apps.

model/gcp/spans.yaml
thompson-tomo: Ok, to make things clearer I would suggest moving the server span definition to be in the rpc namespace. This would mirror what has been done with messaging/db, have a gcp md doc in rpc namespace and link to it from a gcp cloud provider page.

model/gcp/registry.yaml
michaelsafyan: I think it is more likely that `app.namespace` would represent a user concept (like the [AppHub host project](https://cloud.google.com/app-hub/docs/set-up-app-hub-host-project) in which the `app.name`/`app.id` is defined by the customer), still leaving the need for this separate concept/attribute.

model/gcp/spans.yaml
michaelsafyan: Won't that be more confusing, since it is a blend of HTTP / RPC?

model/gcp/spans.yaml
michaelsafyan: I'm not sure what "library/module namespace" would be appropriate here, since the intention would be for this to be used by many different libraries/modules maintained by GCP (as well as server components in GCP that do not have published names for their internal libraries/modules).

Supposing we did this, though, what library/module namespace would you suggest? And what would be the mechanics of doing that definition?

thompson-tomo: I mean #2184 / #2339 

At a quick sketch, I think we would need things like:
- Name
- Version
- Operation.*

With the name being something like "google cloud client library" 

I would suggest we use one of the above issues to discuss and I can draft a pr if desired

model/gcp/spans.yaml
thompson-tomo: I am just referring to the server span which is just rpc.

model/gcp/registry.yaml
thompson-tomo: I see it as both and it would help to enable maximum compatibility. The 2 use cases could be communicated via documentation.

PR #1075, opened 2024-05-25
[cicd] add initial cicd attributes to registry

adrielp:
Adds proposed CICD common attributes to the registry. 

Fixes #915

## Changes

Please provide a brief description of the changes here.

Note: if the PR is touching an area that is not listed in the [existing areas](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/README.md), or the area does not have sufficient [domain experts coverage](https://github.com/open-telemetry/semantic-conventions/blob/main/.github/CODEOWNERS), the PR might be tagged as [experts needed](https://github.com/open-telemetry/semantic-conventions/labels/experts%20needed) and move slowly until experts are identified.

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [x] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

PR #2387, opened 2025-06-18
Initial implementation of workflows #1688

### LINKED ISSUE #1688: Unified semantic conventions for tasks, workflows, pipelines, jobs

thompson-tomo:
Fixes #1688 

## Changes

This adds in the concept of workflows which are designed to be a broad concept encompassing cicd, cron, faas, deployments etc.


kamphaus:
Since https://github.com/open-telemetry/semantic-conventions/pull/2502 was merged, I assume the requirement for a prototype applies to workflows semconv.

docs/registry/attributes/workflow.md
```diff
@@ -0,0 +1,104 @@
...
+
+---
+
+`workflow.execution.result` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `cancellation` | The task run was cancelled, eg. by a user manually cancelling the task run. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `error` | The task run failed due to an error in the CICD system, eg. due to the worker being killed. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `failure` | The task run did not finish successfully, eg. due to a compile error or a failing test. Such failures are usually detected by non-zero exit codes of the tools executed in the task run. | ![Development](https://img.shields.io/badge/-development-blue) |
```
Cyclonit: There should only  be one out of `error` and `failure` Also all of the descriptions appear to be copy and pasted from CICD.

docs/registry/attributes/workflow.md
```diff
@@ -0,0 +1,104 @@
...
+
+---
+
+`workflow.task.run.result` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `cancellation` | The task run was cancelled, eg. by a user manually cancelling the task run. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `error` | The task run failed due to an error in the CICD system, eg. due to the worker being killed. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `failure` | The task run did not finish successfully, eg. due to a compile error or a failing test. Such failures are usually detected by non-zero exit codes of the tools executed in the task run. | ![Development](https://img.shields.io/badge/-development-blue) |
```
Cyclonit, line 120: Same as above. Duplicate `error` and `failure`.

docs/registry/entities/workflow.md
```diff
@@ -0,0 +1,181 @@
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/entity_namespace.md.j2 -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+# Workflow
```
Cyclonit, line 6: I am missing attributes related to batch processing items. Especially when it comes to metrics, attributes and/or dimensions like `workflow.batch.size`, `workflow.batch.error_count` and `workflow.batch.skip_count` would be helpful. Though given the implied relationship of workflows and tasks, I am unsure where to put the batch information. It could be on either level, depending on whether the workflow iterates over items or tasks first.

docs/registry/entities/workflow.md
```diff
@@ -0,0 +1,181 @@
...
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`workflow.name`](/docs/registry/attributes/workflow.md) | string | Name of the workflow | `Build and deploy application` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`workflow.type`](/docs/registry/attributes/workflow.md) | string | Type of the workflow | `DSL` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`workflow.url.full`](/docs/registry/attributes/workflow.md) | string | The [URL](https://wikipedia.org/wiki/URL) of the workflow, providing the complete address in order to locate and identify the workflow. | `https://github.com/open-telemetry/semantic-conventions/actions/runs/9753949763?pr=1075` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`workflow.version`](/docs/registry/attributes/workflow.md) | string | Version of the workflow | `2.1` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+## Workflow Execution
```
Cyclonit: Unless there is some sort of historic reasoning for this, I'd opt for deciding between the terms `execution` and `run` and using one of them for both workflows and tasks. Imho, adding tasks as an inherent property of workflows is good, but the phrasing should be consistent.

I'd go for `workflow.run.*` and `task.run.*` for sake of simplicity.

docs/registry/entities/workflow.md
```diff
@@ -0,0 +1,181 @@
...
+
+**type:** `workflow.execution`
+
+**Description:** 
+
+**Identifying Attributes:**
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`workflow.execution.id`](/docs/registry/attributes/workflow.md) | string | Unique ID for this execution of the workflow and will be shared across all the task runs. | `5c6e9a4b-69ab-499f-b4a7-4cf5c8720d66` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
Cyclonit: Considering tracing, would the `workflow.execution.id` (or `workflow.run.id` as suggested in another comment), have the same value as the trace id?

Cyclonit (commented):
Though I expect replacing existing conventions for k8s and cicd would be a challenge, I like the overall idea behind this change. I believe it has merit even the decision is made not to replace existing conventions, as there are plenty of other use-cases.

docs/registry/attributes/workflow.md
thompson-tomo: The reason for 2 is to differentiate between when the task completed but returned an error code and when the task failed due to the worker. Briefs have been updated.

docs/registry/attributes/workflow.md
thompson-tomo: See above

docs/registry/entities/workflow.md
thompson-tomo: It was tying to avoid different definitions of sub namespaces. I have adjusted modelling to remove that ambguity.

docs/registry/entities/workflow.md
thompson-tomo: You are correct, I had not factored that in. I have now introduced additional attributes which has been used in the appropriate event/span. Also a new metric has been introduced for tracking ratio of outcomes for iterations.

docs/registry/entities/workflow.md
thompson-tomo: Nope trace id is always an additional property. This id is from the system running the task.

thompson-tomo:
@kamphaus yes a prototype would help. Do you have a suggestion or example that could be used. 

I was thinking about using https://docs.elsaworkflows.io/ as one case and at the same time, it is also allowing alignment with namespaces which might have solutions but might not be as fleashed out.

@Cyclonit i agree it is ambitious but for me it is about having a strong base which can cater for them.

docs/registry/entities/workflow.md
kamphaus: For some history: in cicd we had a discussion about the naming of runs and we landed on run in part because it aligned well with the terminology of CDEvents.
cf. https://github.com/open-telemetry/semantic-conventions/pull/1075#discussion_r1615339395

kamphaus:
I'm not familiar with Elsa Workflow, but if you can implement the proposed semconv in it go for it.
Some thoughts from my side would be evaluating whether Elsa already has otel capability and it's only a matter of some renaming and maybe adding some signals or would it require implementing otel instrumentation for workflows from scratch in Elsa. This might have some considerable impact on the effort needed to implement the prototype.
If it's the latter, maybe you might want to take a look at adapting an existing implementation of the cicd otel semconv like the Jenkins [opentelemetry-plugin](https://github.com/jenkinsci/opentelemetry-plugin) or the [github receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.130.0/receiver/githubreceiver) in the otel collector.

thompson-tomo:
thanks for the feedback, Elsa has recently gained some OTEL span capabilities but no metric/events. I will also keep in mind the other examples.

docs/registry/attributes/workflow.md
```diff
@@ -0,0 +1,126 @@
...
+| <a id="workflow-definition-name" href="#workflow-definition-name">`workflow.definition.name`</a> | string | Name of the workflow | `Build and deploy application` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="workflow-definition-type" href="#workflow-definition-type">`workflow.definition.type`</a> | string | Type of the workflow | `DSL` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="workflow-definition-url-full" href="#workflow-definition-url-full">`workflow.definition.url.full`</a> | string | The [URL](https://wikipedia.org/wiki/URL) of the workflow, providing the complete address in order to locate and identify the workflow. | `https://github.com/open-telemetry/semantic-conventions/actions/runs/9753949763?pr=1075` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="workflow-definition-version" href="#workflow-definition-version">`workflow.definition.version`</a> | string | Version of the workflow | `2.1` | ![Development](https://img.shields.io/badge/-development-blue) |
```
Cyclonit, line 17: What is the reasoning behind putting workflow attributes in the `definition` sub-namespace? Is there any precedence for this in other conventions? It feels quite cumbersome to me.

As an example, `workflow.id` clearly denotes the identity of the workflow, whereas `workflow.execution.id` is a single run. I don't see a need for the `definition` sub-namespace at all.

docs/registry/attributes/workflow.md
```diff
@@ -0,0 +1,126 @@
...
+| <a id="workflow-execution-attempt" href="#workflow-execution-attempt">`workflow.execution.attempt`</a> | int | What number execution attempt is this. [1] | `1` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="workflow-execution-id" href="#workflow-execution-id">`workflow.execution.id`</a> | string | Unique ID for this execution of the workflow and will be shared across all the task runs. | `5c6e9a4b-69ab-499f-b4a7-4cf5c8720d66` | ![Development](https://img.shields.io/badge/-development-blue) |
```
Cyclonit, line 19: Suggestion for the brief for clarity:

> Unique ID for this execution of the workflow. All tasks runs of a workflow execution will share the same execution's id.

Cyclonit, line 19: I understand that there is good reason to have `workflow.execution.id` and `task.run.id` as a separate concept, but would there be some way of mating this with trace and span ids? I feel like the `workflow.execution.id` should be used as trace id and each `task.run.id` should show up as span id somewhere too.

docs/registry/attributes/workflow.md
```diff
@@ -0,0 +1,126 @@
...
+| <a id="workflow-execution-state" href="#workflow-execution-state">`workflow.execution.state`</a> | string | The pipeline run goes through these states during its lifecycle. | `pending`; `executing`; `finalizing` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="workflow-execution-url-full" href="#workflow-execution-url-full">`workflow.execution.url.full`</a> | string | The [URL](https://wikipedia.org/wiki/URL) of the pipeline run, providing the complete address in order to locate and identify the pipeline run. | `https://github.com/open-telemetry/semantic-conventions/actions/runs/9753949763?pr=1075` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="workflow-id" href="#workflow-id">`workflow.id`</a> | string | Unique ID for this execution of the workflow and will be shared across all the task runs. | `5c6e9a4b-69ab-499f-b4a7-4cf5c8720d66` | ![Development](https://img.shields.io/badge/-development-blue) |
```
Cyclonit, line 23: IMHO should be `workflow.execution.id`.

docs/registry/attributes/workflow.md
```diff
@@ -0,0 +1,126 @@
...
+| <a id="workflow-iteration-failed" href="#workflow-iteration-failed">`workflow.iteration.failed`</a> | int | The number items that were processed by this task. | `1` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="workflow-iteration-limit" href="#workflow-iteration-limit">`workflow.iteration.limit`</a> | int | This can be a pre-defined limit of times the task can be run or it can be the batch size. [4] | `10` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="workflow-iteration-skipped" href="#workflow-iteration-skipped">`workflow.iteration.skipped`</a> | int | The number items that were processed by this task. | `2` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="workflow-iteration-success" href="#workflow-iteration-success">`workflow.iteration.success`</a> | int | The number items that were processed by this task. | `7` | ![Development](https://img.shields.io/badge/-development-blue) |
```
Cyclonit, line 29: I see what you are going for, but I don't like the naming of `iteration`. Why not go for this instead?

- `workflow.batch.count`
- `workflow.batch.index`
- `workflow.batch.error_count`
- `workflow.batch.skip_count`
- `workflow.batch.success_count`

This naming is more straightforward and concise. It removes the ambiguity of "limit" (which doesn't tell us whether it means a count, duration or other) and the `_count` suffixes tell us that we are dealing with counters over status enums.

docs/registry/attributes/workflow.md
```diff
@@ -0,0 +1,126 @@
...
+| <a id="workflow-outcome" href="#workflow-outcome">`workflow.outcome`</a> | string | The result of a task run. | `success`; `failure`; `timeout`; `skipped` | ![Development](https://img.shields.io/badge/-development-blue) |
```
Cyclonit: Duplicate of `workflow.execution.outcome`?

docs/registry/attributes/workflow.md
thompson-tomo: Reason is to cater for adhoc tasks where there is no definition defined but rather a task is just run ie deploy service a. Key thing in this case the Workflow is describing what has happened. Perhaps we could do Workflow.instance.id etc to make it clearer.

docs/registry/attributes/workflow.md
thompson-tomo: Yes the tool you would be using would already have them as seperate properties. It could very well be that Workflow.run.id is the trace.id if using remote workers.

docs/registry/attributes/workflow.md
thompson-tomo: Batch doesn't work as some Workflow engines allow loops to be defined hence needing another name. Open to other options

docs/registry/attributes/workflow.md
thompson-tomo: See above and option

docs/registry/attributes/workflow.md
thompson-tomo: Yes by design the result of running a task will have an outcome but it might differ to the outcome of the Workflow.

docs/registry/attributes/workflow.md
Cyclonit: What do you think of `workflow.iteration.index` instead of `workflow.iteration.count`? That would free up `workflow.iteration.count` for the batch use case, where `limit` feels out of place.

In a batch, you have a given number of items you want to process. That total would amount to `workflow.iteration.count`. Out of all of these items `workflow.iteration.error_count` failed, `workflow.iteration.skip_count` were skipped and `workflow.iteration.success_count` succeeded.

`workflow.iteration.limit` would then end up being an upper bound for cases where there is no predeterminable count of items.

docs/registry/attributes/workflow.md
Cyclonit: Hm.. I think this would be clearer with a structure like this:

1. `workflow.*`  
The general definition of the workflow. Such as `workflow.name` or `workflow.id`.
    1. `workflow.run.*`
    A single execution/run/instance of the workflow. Such as `workflow.run.id` or `workflow.run.outcome`. This maps nicely on existing CICD systems and workflow engines. In GitLab, each run of a pipeline would map to a `workflow.run`.
    2. `workflow.task.*`
    The general definition of a single task in a given workflow. Such as `workflow.task.name` or `workflow.task.id`.
        1. `workflow.task.run.*`
        A single run of a task. Such as `workflow.task.run.id`. Always belongs to a `workflow.run.id`. In GitLab, each pipeline job would map to a `workflow.task.run`.

Another option would be to fully extract `task` as its own root namespace. That would alleviate most issues right from the start and allow for adhoc tasks that are not connected a workflow in the first place.

thompson-tomo: The problem is that does not fit the approach elsewhere where the focus is on what are you describing being your base namespace which in this case is the execution of a Workflow. This namespace has additional descriptive attributes attached to it, either from the same, sub or even a different namespace.

Examples include
- Container
- deployment
- Host
- process

I think the key thing is looking at how the attributes would be used which is where the span & metrics come in. As such having Workflow.outcome is clear that it is the overall outcome.

docs/registry/attributes/workflow.md
thompson-tomo: Agree about changing to index but not limit as what we are describing is the limit on the number of iterations.

docs/workflow/workflow-metrics.md
```diff
@@ -0,0 +1,722 @@
...
+| [`error.type`](/docs/registry/attributes/error.md) | string | Describes a class of error the operation ended with. [1] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` If and only if the workflow execution failed. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`host.name`](/docs/registry/attributes/host.md) | string | Name of the host. On Unix systems, it may contain what the hostname command returns, or the fully qualified hostname, or another name specified by the user. | `opentelemetry-test` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`workflow.platform.name`](/docs/registry/attributes/workflow.md) | string | The product being used to co-ordinate the execution of the tasks. | `hangfire`; `k8s`; `quartz` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`container.name`](/docs/registry/attributes/container.md) | string | Container name used by container runtime. | `opentelemetry-autoconf` | `Opt-In` | ![Development](https://img.shields.io/badge/-development-blue) |
```
thompson-tomo: Add workflow.state attribute

docs/workflow/workflow-metrics.md
```diff
@@ -0,0 +1,722 @@
...
+| `quartz` | Quartz | ![Development](https://img.shields.io/badge/-development-blue) |
+
+---
+
+`workflow.state` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `completed` | The completed state spans from when the run has finished finalizing (eg. cleanup of run resources). | ![Development](https://img.shields.io/badge/-development-blue) |
+| `executing` | The executing state spans the execution of any run tasks (eg. build, test). | ![Development](https://img.shields.io/badge/-development-blue) |
```
thompson-tomo: Replace this with active

docs/workflow/workflow-metrics.md
```diff
@@ -0,0 +1,722 @@
...
+| `k8s` | Kubernetes | ![Development](https://img.shields.io/badge/-development-blue) |
+| `quartz` | Quartz | ![Development](https://img.shields.io/badge/-development-blue) |
+
+---
+
+`workflow.state` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `completed` | The completed state spans from when the run has finished finalizing (eg. cleanup of run resources). | ![Development](https://img.shields.io/badge/-development-blue) |
```
thompson-tomo: Remove

Issue #288, opened 2023-08-29
Use consistent suffix for JVM Metrics 


lenin-jaganathan:
While going through https://github.com/open-telemetry/semantic-conventions/blob/4bbb8c907402caa90bc077214e8a2c78807c1ab9/docs/jvm/jvm-metrics.md?plain=1#L15-L38, it seems like the suffices for some of the metrics seem to be inconsistent with one another and are also not so straight forward.

I propose replacing the suffix "usage" (noun) in the following metrics with "used" (verb),

- jvm.memory.usage
- jvm.memory.usage_after_last_gc
- jvm.buffer.memory.usage

All these represent the amount being used. In addition, the description already uses the word "used" to describe these metrics and the JMX methods also use the word "used". Also, the sibling metrics use the verb form. E.g: "jvm.memory.committed"

trask:
hi @lenin-jaganathan!

`.usage` is a common suffix used in otel semconv, check out
* https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#instrument-naming
* https://github.com/open-telemetry/semantic-conventions/issues/211#issuecomment-1650388098

so we would probably need to have that discussion (`usage` -> `used`) more broadly than only JVM metrics

I personally think either nouns or adjectives (when it describes the namespace which is a noun) seem like good metric names, e.g.
* `jvm.memory.committed` - measure "committed memory"
* `jvm.class.loaded` - measures "loaded classes"

trask:
I realized, I don't think we would want to rename `usage` -> `used` generally, e.g. [`system.memory.usage`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemmemoryusage) has dimensions:

* total
* used <--
* free
* shared
* buffers
* cached

@lenin-jaganathan wdyt? do you want to keep this issue open? thx

lenin-jaganathan:
I would still like this to be addressed. There is a subtle difference between `system.memory.usage` and `jvm.memory.usage`.

As you mentioned, the first metric has multiple states - used, free, cached, etc. But the JVM metric is all about using part of it and we have separate metrics for committed, limit, etc. But even in the system metrics use-case, `total` as a dimension is always a questionable choice since we would expect adding up all the dimensions should be total.

trask:
> But even in the system metrics use-case, `total` as a dimension is always a questionable choice since we would expect adding up all the dimensions should be total.

[...]

PR #2392, opened 2025-06-19
Add k8s.{node, pod, container}.filesystem. metrics

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1488

## Changes

This PR adds the following metrics that the Collector already implementes:

```
k8s.node.filesystem.available
k8s.node.filesystem.capacity
k8s.node.filesystem.usage
k8s.pod.filesystem.available
k8s.pod.filesystem.capacity
k8s.pod.filesystem.usage
container.filesystem.available
container.filesystem.capacity
container.filesystem.usage
```


ChrsMark:
Still active. @open-telemetry/semconv-k8s-approvers please take a look.

model/container/metrics.yaml
```diff
@@ -102,3 +102,47 @@ groups:
...
+      In K8s, this metric is derived from the
+      [FsStats.AvailableBytes](https://pkg.go.dev/k8s.io/kubelet@v0.33.0/pkg/apis/stats/v1alpha1#FsStats) field
+      of the [ContainerStats.Rootfs](https://pkg.go.dev/k8s.io/kubelet@v0.33.0/pkg/apis/stats/v1alpha1#ContainerStats)
+      of the Kubelet's stats API.
+  - id: metric.container.filesystem.capacity
+    type: metric
+    metric_name: container.filesystem.capacity
+    stability: development
+    brief: "Container filesystem capacity"
+    instrument: gauge
```
dashpole: should be updowncounters

TylerHelmuth (approved):

docs/non-normative/k8s-migration.md
```diff
@@ -61,6 +61,7 @@ and one for disabling the old schema called `semconv.k8s.disableLegacy`. Then:
   - [K8s Container metrics](#k8s-container-metrics)
   - [K8s ResourceQuota metrics](#k8s-resourcequota-metrics)
   - [K8s Node condition metrics](#k8s-node-condition-metrics)
+  - [K8s Filestystem metrics](#k8s-filestystem-metrics)
```
```suggestion
  - [K8s Filesystem metrics](#k8s-filesystem-metrics)
```

docs/non-normative/k8s-migration.md
```diff
@@ -370,3 +371,28 @@ The changes in their metrics are the following:
 | `k8s.node.condition_*`                                                             | `k8s.node.condition.status` metric [0,1] with attribute `k8s.node.condition.type` for the different conditions and `k8s.node.condition.status` for true/false/unknown |
 
 <!-- prettier-ignore-end -->
+
+### K8s Filestystem metrics
```
```suggestion
### K8s Filesystem metrics
```

model/k8s/metrics.yaml
```diff
@@ -91,6 +91,59 @@ groups:
       - ref: network.interface.name
       - ref: network.io.direction
 
+  # k8s.pod.filesystem.* metrics
+  - id: metric.k8s.pod.filesystem.available
+    type: metric
+    metric_name: k8s.pod.filesystem.available
+    stability: development
+    brief: "Pod filesystem available"
```
joaopgrassi:
```suggestion
    brief: "Pod filesystem available bytes"
```

It's rather unclear what is being measured without having `bytes` somewhere in the bried I'd say.

model/k8s/metrics.yaml
```diff
@@ -91,6 +91,59 @@ groups:
...
+      code_generation:
+        metric_value_type: int
+    entity_associations:
+      - k8s.pod
+    note: |
+      This metric is derived from the
+      [FsStats.CapacityBytes](https://pkg.go.dev/k8s.io/kubelet@v0.33.0/pkg/apis/stats/v1alpha1#FsStats) field
+      of the [PodStats.EphemeralStorage](https://pkg.go.dev/k8s.io/kubelet@v0.33.0/pkg/apis/stats/v1alpha1#PodStats)
+      of the Kubelet's stats API.
+  - id: metric.k8s.pod.filesystem.usage
```
joaopgrassi, line 129: Do you think the `usage` name here is aligned with the [current guidelines](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#instrument-naming)? 

> usage - an instrument that measures an amount used out of a known total (limit) amount should be called entity.usage. For example, system.memory.usage with attribute state = used | cached | free | ... for the amount of memory in a each state. Where appropriate, the sum of usage over all attribute values SHOULD be equal to the limit.

> A measure of the amount consumed of an unlimited resource, or of a resource whose limit is unknowable, is differentiated from usage. For example, the maximum possible amount of virtual memory that a process may consume may fluctuate over time and is not typically known.

Is `k8s.pod.filesystem.capacity` the "limit" here?

model/container/metrics.yaml
```diff
@@ -102,3 +102,56 @@ groups:
     attributes:
       - ref: network.io.direction
       - ref: network.interface.name
+
+  # container.filesystem.* metrics
+  - id: metric.container.filesystem.available
+    type: metric
+    metric_name: container.filesystem.available
+    stability: development
+    brief: "Container filesystem available"
```
joaopgrassi:
```suggestion
    brief: "Container filesystem bytes available"
```

model/k8s/metrics.yaml
joaopgrassi: There was this old discussions, when the JVM metrics were renamed from `usage` to `used`. I wonder if the same applies for these.. https://github.com/open-telemetry/semantic-conventions/issues/288

ChrsMark: > Is k8s.pod.filesystem.capacity the "limit" here?

The same with `k8s.pod.volume.usage` applies here: https://github.com/open-telemetry/semantic-conventions/pull/2319/files#diff-bc357ba83c440c197ef2b49741508f83f85ad2829e7d1abd64b1a84b04135ca5R152

Added the `This may not equal capacity - available` based on https://pkg.go.dev/k8s.io/kubelet@v0.33.0/pkg/apis/stats/v1alpha1#FsStats.

So, there is a limit indeed but it's not necessarily the `k8s.pod.filesystem.capacity`. In this, I guess `.usage` is fine?

joaopgrassi: It seems fine to me!

joaopgrassi (approved):

PR #2403, opened 2025-06-21
Additional info about the telemetry #2345

### LINKED ISSUE #2345: Enable applications/services to be grouped together

thompson-tomo:
Fixes #2345 

## Changes

Adds the ability to specify the organisation which the telemetry is for and the destination of the telemetry,

The organisation telemetry is to bring across a similar functionality in ecs.

### DIFF
--- a/docs/registry/attributes/telemetry.md
+++ b/docs/registry/attributes/telemetry.md
@@ -9,16 +9,23 @@ This document defines attributes for telemetry SDK.
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
-| <a id="telemetry-distro-name" href="#telemetry-distro-name">`telemetry.distro.name`</a> | string | The name of the auto instrumentation agent or distribution, if used. [1] | `parts-unlimited-java` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="telemetry-destination-namespace" href="#telemetry-destination-namespace">`telemetry.destination.namespace`</a> | string | The namespace which the telemetry should be put into. [1] | `public-website` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="telemetry-distro-name" href="#telemetry-distro-name">`telemetry.distro.name`</a> | string | The name of the auto instrumentation agent or distribution, if used. [2] | `parts-unlimited-java` | ![Development](https://img.shields.io/badge/-development-blue) |
 | <a id="telemetry-distro-version" href="#telemetry-distro-version">`telemetry.distro.version`</a> | string | The version string of the auto instrumentation agent or distribution, if used. | `1.2.3` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="telemetry-organisation-id" href="#telemetry-organisation-id">`telemetry.organisation.id`</a> | string | Unique identifier for the organization. | `19e97b12-0fa0-45de-9c59-23ff59d572b9` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="telemetry-organisation-name" href="#telemetry-organisation-name">`telemetry.organisation.name`</a> | string | Name for the Organization or team. | `productX_support` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="telemetry-organisation-role" href="#telemetry-organisation-role">`telemetry.organisation.role`</a> | string | Role of the organisation. | `support`; `hosting` | ![Development](https://img.shields.io/badge/-development-blue) |
 | <a id="telemetry-sdk-language" href="#telemetry-sdk-language">`telemetry.sdk.language`</a> | string | The language of the telemetry SDK. | `cpp`; `dotnet`; `erlang` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
-| <a id="telemetry-sdk-name" href="#telemetry-sdk-name">`telemetry.sdk.name`</a> | string | The name of the telemetry SDK as defined above. [2] | `opentelemetry` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| <a id="telemetry-sdk-name" href="#telemetry-sdk-name">`telemetry.sdk.name`</a> | string | The name of the telemetry SDK as defined above. [3] | `opentelemetry` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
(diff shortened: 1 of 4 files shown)


docs/registry/attributes/telemetry.md
```diff
@@ -9,16 +9,23 @@ This document defines attributes for telemetry SDK.
...
 | <a id="telemetry-sdk-language" href="#telemetry-sdk-language">`telemetry.sdk.language`</a> | string | The language of the telemetry SDK. | `cpp`; `dotnet`; `erlang` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
-| <a id="telemetry-sdk-name" href="#telemetry-sdk-name">`telemetry.sdk.name`</a> | string | The name of the telemetry SDK as defined above. [2] | `opentelemetry` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| <a id="telemetry-sdk-name" href="#telemetry-sdk-name">`telemetry.sdk.name`</a> | string | The name of the telemetry SDK as defined above. [3] | `opentelemetry` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 | <a id="telemetry-sdk-version" href="#telemetry-sdk-version">`telemetry.sdk.version`</a> | string | The version string of the telemetry SDK. | `1.2.3` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 
-**[1] `telemetry.distro.name`:** Official auto instrumentation agents and distributions SHOULD set the `telemetry.distro.name` attribute to
+**[1] `telemetry.destination.namespace`:** For Elastic this would be the space.
```
michaelsafyan, line 24: How would you know which case it is supposed to be?

Suggestion:

Instead of:

 - `telemetry.destination.namespace`

Consider instead:

- `telemetry.destination.type`
- `telemetry.destination.id`

This way, for example, you could have a `type` of things like `elastic.space`, `gcp.project`, etc.

However, I'm still not entirely sure how this works if you are routing to multiple destinations... do you always need to create multiple pipelines to route to more than one destination? And what is the expected behavior for backends that don't recognize this attribute? Or for non-matching values?

thompson-tomo: I don't see a need for the additional/alternative fields as The purpose of the namespace field is to **help** get the telemetry to a particular area of the destination. 

> However, I'm still not entirely sure how this works if you are routing to multiple destinations... 
do you always need to create multiple pipelines to route to more than one destination? 

I don't see it being any different to what you have now or what you have proposed which is also a single attribute.

> And what is the expected behavior for backends that don't recognize this attribute? Or for non-matching values?

It is an opt-in attribute, so if not supported by the backend it would behave no different to what it does now. If directly supported it **can** and is recommended to be used to easily route the telemetry to the most correct location ie project/space. However it could also be used in a collector pipeline.

michaelsafyan: > I don't see it being any different to what you have now or what you have proposed which is also a single attribute.

This proposal is a single shared attribute across all vendors which could be an obstacle to a single data stream being used to route data to multiple vendors.

The other proposal is a vendor-specific routing attribute, allowing other vendors to similarly have vendor-specific attributes, allowing a single data stream to route to multiple vendors.

To illustrate, let's say that instead of `telemetry.destination.*` we had separate vendor attributes like:

   - `gcp.project_id`
   - `elastic.space.id`
   - `datadog.subdomain`
   - ...

It would then be possible to route the same pipeline to multiple different vendors; for example, let's suppose that a user of GCP wanted to record tracing data locally in the project from which it originated as well as to export/route that data to a common Elastic destination for centralized observablity in their organization. If using a single shared attribute (like `telemetry.destination.*`) there would need to be copying of the data to multiple pipelines and transforming the data to mutate the value of `telemetry.destination.*` for each destination; if separate attributes are used for each vendor, then it is possible to take the same data stream and just route it to multiple different exporters without any further transformation.

Vendors could choose whether to support a string or an array of strings, depending on typical usage patterns with the vendor regarding whether -- within the vendor -- users typically route to multiple places (as well as, based on the internal routing capabilities offered by each vendor).

> It is an opt-in attribute, so if not supported by the backend it would behave no different to what it does now. If directly supported it can and is recommended to be used to easily route the telemetry to the most correct location 

In that sense, I would still see there being value in having such an attribute (as a kind of routing hint/tag), but perhaps not having the same purpose/utility as hard unit of tenancy (like `gcp.project_id`); for example, within a Google Cloud project destination, something like `telemetry.destination.*` could be useful to route to a specific storage location for that given customer. But I don't think it could replace `gcp.project_id` (and would likely designate a project sub-resource like a routing rule contained in the project, not the project itself).

thompson-tomo: > This proposal is a single shared attribute across all vendors which could be an obstacle to a single data stream being used to route data to multiple vendors.

Which is solved by using an [otel collector](https://opentelemetry.io/docs/collector/)

> If using a single shared attribute (like telemetry.destination.*) there would need to be copying of the data to multiple pipelines and transforming the data to mutate the value of telemetry.destination.* for each destination; 

Not necessarily and in fact mutation would be very rare as I would have the backend define which space/domain etc stores which namespace. This also means that the end services do not need to know anything destinations but just the general namespace.

> something like telemetry.destination.* could be useful to route to a specific storage location for that given customer. 

If a need for that arose that could be a future addition.

thompson-tomo:
@trisch-me any concerns about this porting of the organisation namespace to semconv in this way?

trisch-me:
@thompson-tomo I think it’s fine because we don’t have inheritance yet. If/when we would have those should be linked to the `organisation` namespace fields

jsuereth:
Two things - 

1. I don't love the `telemetry` namespace here.
2. This should be decided in the new SIG forming here: https://github.com/open-telemetry/community/pull/2837

We want to make sure vendors, cloud providers and the ecosystem are adopting these conventions, which is why we form SIGs to define them and make sure they are compatible across offerings.

PR #2404, opened 2025-06-21
Adds explicit id for faas entities #2393

### LINKED ISSUE #2393: Introduce id for FAAS namespace.
### Area(s)

area:faas

### What's missing?

The ability to have an id for the function when not in a cloud scenario 

Introduction of a faas.id attribute to uniquely identify a function.

I am aware of the cloud.resource_id but that is not usable for non cloud & it makes the entity concept difficult as all cloud resources would be seen as a faas function unless a 2nd/3rd field is used. Whereas single id field  makes things clearer.

thompson-tomo:
Fixes #2393 

## Changes

Adds in an explicit id for the faas entity to enable decoupling of the cloud resource.id

### DIFF
--- a/docs/faas/faas-spans.md
+++ b/docs/faas/faas-spans.md
@@ -48,8 +48,9 @@ If Spans following this convention are produced, a Resource of type `faas` MUST 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
 | [`cloud.resource_id`](/docs/registry/attributes/cloud.md) | string | Cloud provider-specific native identifier of the monitored cloud resource (e.g. an [ARN](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html) on AWS, a [fully qualified resource ID](https://learn.microsoft.com/rest/api/resources/resources/get-by-id) on Azure, a [full resource name](https://google.aip.dev/122#full-resource-names) on GCP) [1] | `arn:aws:lambda:REGION:ACCOUNT_ID:function:my-function`; `//run.googleapis.com/projects/PROJECT_ID/locations/LOCATION_ID/services/SERVICE_ID`; `/subscriptions/<SUBSCRIPTION_GUID>/resourceGroups/<RG>/providers/Microsoft.Web/sites/<FUNCAPP>/functions/<FUNC>` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`faas.id`](/docs/registry/attributes/faas.md) | string | The id of the single function that this runtime instance executes. [2] | `arn:aws:lambda:REGION:ACCOUNT_ID:function:my-function`; `//run.googleapis.com/projects/PROJECT_ID/locations/LOCATION_ID/services/SERVICE_ID`; `/subscriptions/<SUBSCRIPTION_GUID>/resourceGroups/<RG>/providers/Microsoft.Web/sites/<FUNCAPP>/functions/<FUNC>` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`faas.invocation_id`](/docs/registry/attributes/faas.md) | string | The invocation ID of the current function invocation. | `af9d5aa4-a685-4c5f-a22b-444f80b3cc28` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`faas.trigger`](/docs/registry/attributes/faas.md) | string | Type of the trigger which caused this function invocation. [2] | `datasource`; `http`; `pubsub` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`faas.trigger`](/docs/registry/attributes/faas.md) | string | Type of the trigger which caused this function invocation. [3] | `datasource`; `http`; `pubsub` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 
 **[1] `cloud.resource_id`:** On some cloud providers, it may not be possible to determine the full ID at startup,
 so it may be necessary to set `cloud.resource_id` as a span attribute instead.
@@ -69,7 +70,9 @@ The following well-known definitions MUST be used if you set this attribute and 
   This means that a span attribute MUST be used, as an Azure function app can host multiple functions that would usually share
   a TracerProvider.
 
-**[2] `faas.trigger`:** For the server/consumer span on the incoming side,
+**[2] `faas.id`:** This id in the case of cloud based faas systems will be the same value as `cloud.resource_id` however in the case of local systems this should be another local identifier.
+
+**[3] `faas.trigger`:** For the server/consumer span on the incoming side,
 `faas.trigger` MUST be set.
 
 Clients invoking FaaS instances usually cannot set `faas.trigger`,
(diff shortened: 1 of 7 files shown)


thompson-tomo:
Awaiting review

thompson-tomo:
@trisch-me another pr which is a pure ecs migration. :)

model/faas/entities.yaml
```diff
@@ -4,11 +4,20 @@ groups:
...
       - ref: faas.version
       - ref: faas.instance
+  - id: entity.faas.instance
+    type: entity
+    stability: development
+    name: faas.instance
+    brief: >
+      A serverless instance.
+    attributes:
+      - ref: faas.instance
```
trisch-me, line 17: attribute and the entity names are the same?

model/faas/entities.yaml
```diff
@@ -4,11 +4,20 @@ groups:
...
       - ref: faas.max_memory
-      - ref: cloud.resource_id
```
trisch-me, line 14: Should it be removed? Where the id will be located then, why not replacing it with faas.id?

model/faas/entities.yaml
thompson-tomo: Agree not ideal so have split it out to seperate pr

thompson-tomo:
@tylerbenson thanks for feedback, see https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3087 for the update done to the dotnet implementation. Key thing being scope is naming update rather new attributes.

I would suggest looking to stabilise or atleast moving to beta/rc stability for atleast the function & instance namespaces that implementations can be updated in bulk.

docs/faas/faas-spans.md
```diff
@@ -47,9 +47,10 @@ If Spans following this convention are produced, a Resource of type `faas` MUST
...
+| [`cloud.resource_id`](/docs/registry/attributes/cloud.md) | string | Cloud provider-specific native identifier of the monitored cloud resource (e.g. an [ARN](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html) on AWS, a [fully qualified resource ID](https://learn.microsoft.com/rest/api/resources/resources/get-by-id) on Azure, a [full resource name](https://google.aip.dev/122#full-resource-names) on GCP) [1] | `arn:aws:lambda:REGION:ACCOUNT_ID:function:my-function`; `//run.googleapis.com/projects/PROJECT_ID/locations/LOCATION_ID/services/SERVICE_ID`; `/subscriptions/<SUBSCRIPTION_GUID>/resourceGroups/<RG>/providers/Microsoft.Web/sites/<FUNCAPP>/functions/<FUNC>` | `Recommended` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `faas.function.id`. |
```
Oberon00, line 50: I am strongly against deprecating cloud.resource_id. There should be no harm in setting it in addition to faas.function.id if on a cloud, and generic cloud backend logic not aware of faas.function.id would break and require extra code after this change.

Besides, we originally had faas.id but explicitly removed it in favor of the more generic cloud.resource_id. Reverting such an explicit decision IMHO requires a stronger motivation than this issue + PR currently provides (or at least documents). See https://github.com/open-telemetry/opentelemetry-specification/pull/3188

For example, it is not clear why one can't set cloud.resource_id even outside a classic cloud, e.g. by adding a special cloud.provider (like "on-premise").

model/faas/spans.yaml
```diff
@@ -26,6 +27,9 @@ groups:
           call to invoke the lambda, which is often HTTP).
       - ref: faas.invocation_id
       - ref: cloud.resource_id
+        deprecated:
```
Oberon00, line 30: I am strongly against deprecating cloud.resource_id. There should be no harm in setting it in addition to faas.function.id if on a cloud. Generic cloud backend logic not aware of faas.function.id would break and require extra code after this change.

Besides, we originally had faas.id but explicitly removed it in favor of the more generic cloud.resource_id. Reverting such an explicit decision IMHO requires a stronger motivation than this issue + PR currently provides (or at least documents). See https://github.com/open-telemetry/opentelemetry-specification/pull/3188

For example, it is not clear why one can't set cloud.resource_id even outside a classic cloud, e.g. by adding a special cloud.provider (like "on-premise").

(EDIT: moved  this comment from md to yaml)

model/faas/entities.yaml
```diff
@@ -1,14 +1,18 @@
...
+        role: identifying
+        requirement_level: required
       - ref: faas.name
+        role: descriptive
         requirement_level: required
       - ref: faas.version
+        role: identifying
       - ref: faas.instance
       - ref: faas.max_memory
-      - ref: cloud.resource_id
```
Oberon00, line 14: Even more important than the (optional alternative) cloud resource ID on the span would be keeping this one to not cause breakage. See https://github.com/open-telemetry/semantic-conventions/pull/2404#discussion_r2333052506

model/faas/entities.yaml
```diff
@@ -1,14 +1,18 @@
...
     stability: development
-    name: faas
+    name: faas.function
     brief: >
-      A serverless instance.
+      A serverless function.
     attributes:
+      - ref: faas.function.id
+        role: identifying
+        requirement_level: required
```
Oberon00, line 11: While semantically desirable, in practice we cannot require this as resource attribute I believe, see https://github.com/open-telemetry/opentelemetry-specification/issues/1261#issuecomment-1198285795

model/faas/entities.yaml
thompson-tomo: Entities are not currently able to be attached to signals via sdk hence this doesn't lead to a Breakage.

model/faas/entities.yaml
thompson-tomo: So a key part of entities is reproducible & consistent identity hence we need something to identify a function. There is nothing to say aws has to produce the produce the entity and take a look at cicd examples etc.

model/faas/entities.yaml
Oberon00: Can you point me to more information on that? My understanding was that the entity is just a logical description of a set of resource attributes that together describe that entity. So I thought it is attached via resource attributes.

model/faas/spans.yaml
thompson-tomo: This is a general span and what I would suggest the cloud attributes are added to this page https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/ alongside the aws specific attributes.

model/faas/entities.yaml
Oberon00: Actually I see it referenced in `docs/resource/faas.md` as `<!-- semconv entity.faas.function -->` so I think it actually is the resource?

model/faas/spans.yaml
Oberon00: The cloud attributes are not aws specific.

model/faas/entities.yaml
Oberon00: Why can't we use cloud.resource_id for that consistent identity? It seems you think cloud.resource_id is AWS-specific? That is not the case.

model/faas/entities.yaml
thompson-tomo: The draft got closed as it went stale https://github.com/open-telemetry/opentelemetry-specification/pull/4565 and the resource id would be handled via the cloud entity https://opentelemetry.io/docs/specs/semconv/registry/entities/cloud/ which also provides the cloud provider etc.

Oberon00: In any case the changed entity file seems to be relevant already right now in released SDKs due to the cross-reference from the resource semconvs IIUC

model/faas/spans.yaml
thompson-tomo: Actually the latest approach is to have those pages show the full definition of signals rather than technology specific attributes. See db pages as an example.

Oberon00: Probably we are talking past each other. My point is that from backend perspective I would still prefer to keep having cloud.resource_id set on spans that had it before. IIUC the deprecation here would deprecate that.

model/faas/entities.yaml
thompson-tomo: Cloud attributes are designed for cloud providers be it aws, gcp, azure etc as per the description of the attributes.

Hence this solution is consistent with other namespaces and it works regardless of where it is deployed to.

Oberon00: But I think this is not specific to faas and probably also not only to resource_id. Any deployed "thing" that uses cloud.resource_id could use an ID even if it is deployed somewhere else. So maybe pursuing the pseudo cloud provider value like "on-premise"  would be worth it?

thompson-tomo: Just did a search of the repo but the only attribute reference to cloud.resource_id not within the cloud or faas namespace I could find is an azure resource log event. 

I am aware that they are common concept hence see having it on an implementation description page for lamdba as a fitting option especially as that way we can also add other things specific for aws including other cloud attributes. It is almost similar to thread I'd, it's something always there when running in the cloud but we don't document it on all the signals/traces.

joaopgrassi (changes requested):
I understand the goal, but I'm failing to see the motive. Do we have users asking for this? Are there "non cloud"/"on prem" instrumentations/implementations today that are lacking a way to encode the "faas" identity? 

Given there was [extensive discussion from the FaaS SIG](https://github.com/open-telemetry/opentelemetry-specification/pull/3188) in the past where it was decided to actually switch to `cloud.resource_id` from `faas.id` AND this would break back-ends and ultimately users of such backends I think we should not accept this without a proper SIG and the proper stabilization process, where we have grace periods where instrumentations emit both attributes until everyone adapts.

model/faas/entities.yaml
thompson-tomo: So the attribute in question is still provided via the cloud entity along with other attributes ie cloud.platform which is used by cloud faas platforms hence it is not needed also on faas entity. A resource is made of 1 or more entities, hence this should not be an issue.

PR #2406, opened 2025-06-21
Add more service info #2391

### LINKED ISSUE #2391: Insight into what a service is providing
### Area(s)

area: service

### What's missing?

The ability to specify what roles the service is performing in the scope of the entire solution as well as it's current state

Introduction of additional service level attributes

- service.roles[]
- service.state

thompson-tomo:
Fixes #2391

## Changes

Introduces the roles array to services so that data can be filtered based upon the role the service is providing. At the same time the service can now have a state.

## Usages
State is a migration of https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/windowsservicereceiver/documentation.md and roles is coming across from ECS.

### DIFF
--- a/model/service/entities.yaml
+++ b/model/service/entities.yaml
@@ -1,17 +1,35 @@
 groups:
+  - id: entity.service.intance
+    type: entity
+    name: service.instance
+    brief: >
+      A service instance.
+    stability: development
+    attributes:
+      - ref: service.name
+        requirement_level: required
+        role: descriptive
+      - ref: service.version
+        role: descriptive
+      - ref: service.namespace
+        role: descriptive
+      - ref: service.instance.id
+        role: identifying
   - id: entity.service
     type: entity
     name: service
     brief: >
-      A service instance.
+      A service which can generate telemetry.
     stability: stable
     attributes:
       - ref: service.name
         requirement_level: required
         role: identifying
       - ref: service.version
-        role: descriptive
+        role: identifying
       - ref: service.namespace
         role: identifying
-      - ref: service.instance.id
-        role: identifying
+      - ref: service.roles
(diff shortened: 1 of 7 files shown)


thompson-tomo:
Awaiting review

model/service/registry.yaml
```diff
@@ -69,3 +69,27 @@ groups:
...
+        type: string[]
+        stability: development
+        brief: >
+          Roles of a service node.
+        note: |
+          This allows for distinction between different running roles of the same service.
+
+          In the case of Kibana, the service.node.role could be ui or background_tasks or both.
+
+          In the case of Elasticsearch, the service.node.role could be master or data or both.
```
KalleOlaviNiemitalo: Do you mean `service.roles` rather than `service.node.role`?

model/service/entities.yaml
```diff
@@ -1,17 +1,35 @@
 groups:
+  - id: entity.service.intance
```
KalleOlaviNiemitalo: This should be `entity.service.instance` rather than `entity.service.intance`.

model/service/entities.yaml
```diff
@@ -1,17 +1,35 @@
...
+      A service which can generate telemetry.
     stability: stable
     attributes:
       - ref: service.name
         requirement_level: required
         role: identifying
       - ref: service.version
-        role: descriptive
-      - ref: service.namespace
         role: identifying
```
KalleOlaviNiemitalo: `service.version` changed from descriptive to identifying.  Is that by design?

thompson-tomo:
@KalleOlaviNiemitalo thanks for picking up the copy paste issue as this is an ecs migration.

PR #2407, opened 2025-06-21
More informative container runtime #2342

### LINKED ISSUE #2342: Extended details about container runtime
### Area(s)

area: container

### What's missing?

I want to be able to see what runtime is being used to run my container

Introduction of additional attributes to store information about the runtime being used to run the container 

- `container.runtime.name` ie containerd
- `container.runtime.description` 
- `container.runtime.verson` ie 2.0.0

Deprecation of

- `container.runtime`

### Additional context

Connected to
- #2340
- #2341

thompson-tomo:
Fixes #2342 

## Changes

This converts the `container.runtime` attribute to be a sub-namespace so that the version & description of the runtime can also be supplied.


model/container/entities.yaml
```diff
@@ -7,17 +7,47 @@ groups:
       A container instance.
     attributes:
       - ref: container.name
+        role: descriptive
```
ChrsMark: `role` is pretty fresh concept, right? This should be handled in a different standalone PR I think.

thompson-tomo: It has been around for a little while as it is a part of the entity topic which now generates docs. I added them in to avoid the need to come back and add the role.

ChrsMark: Has it been decided already which attributes will be descriptive and which identifying for container entities? If not, then that's a discussion need to happen first in a different issue. Mixing such a discussion with a change that's trying to address a specific issue on its own does not sound a good idea to me. Unless there is a specific reason for this.

thompson-tomo: Sure I can restrict my usage of identifying role to just the entities I have added and reduce the descriptive on the other entity.

model/container/registry.yaml
```diff
@@ -19,12 +19,26 @@ groups:
           [identify Docker containers](https://docs.docker.com/engine/containers/run/#container-identification).
           The UUID might be abbreviated.
         examples: ['a3bf90e006b2']
-      - id: container.runtime
+      - id: container.runtime.name
```
ChrsMark, line 22: Could you add this in the [non-normative migration guidance](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-migration.md#summary-of-changes)? It affects the [k8sclusterreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9f1f674d7af62267f641c0b631d1326815ff80d1/receiver/k8sclusterreceiver/metadata.yaml#L180-L188) and we try to keep track of such changes to prepare the users.

thompson-tomo: Sure, I think https://github.com/open-telemetry/semantic-conventions/commit/9be5dcdf8fbd9946139a0429aa41f651558b1723 should now cover it.

ChrsMark (approved):

docs/non-normative/k8s-migration.md
```diff
@@ -370,3 +371,18 @@ The changes in their metrics are the following:
 | `k8s.node.condition_*`                                                             | `k8s.node.condition.status` metric [0,1] with attribute `k8s.node.condition.type` for the different conditions and `k8s.node.condition.status` for true/false/unknown |
 
 <!-- prettier-ignore-end -->
+
+### Container Runtime
+
+The container runtime has become more desriptive with changes introduced to semantic conventions
```
```suggestion
The container runtime has become more descriptive with changes introduced to semantic conventions
```

model/container/registry.yaml
```diff
@@ -19,12 +19,26 @@ groups:
...
+          The version of the runtime of this process, as returned by the runtime
+          without modification.
+        examples: '1.0.0'
+      - id: container.runtime.description
+        type: string
+        stability: development
+        brief: >
+          An additional description about the runtime, for example
+          a specific vendor customization of the runtime environment.
+        examples: ''
```
lmolkova: is there a realistic example?

model/container/registry.yaml
```diff
@@ -19,12 +19,26 @@ groups:
...
+        stability: development
+        brief: >
+          The version of the runtime of this process, as returned by the runtime
+          without modification.
+        examples: '1.0.0'
+      - id: container.runtime.description
+        type: string
+        stability: development
+        brief: >
+          An additional description about the runtime, for example
```
lmolkova: Is there a scenario where it's necessary? Can we provide guidance on how different instrumentations would collect it so it's useful and consistent?

model/container/registry.yaml
thompson-tomo: Have now provided a realistic example.

model/container/registry.yaml
thompson-tomo: Wording has been updated and is complemented by the example. 

An alternative option if want consistency we could replace `container.runtime.description` with `container.runtime.api.version` & `container.runtime.api.type` ie cri.

jinja2 (approved):

PR #2421, opened 2025-06-24
Adds in mac address for traffic analysis #2416

### LINKED ISSUE #2416: Capture source/destination MAC Address
### Area(s)

area:source, area:destination

### What's missing?

The ability to capture the mac address of the network traffic

Introduction of the following attributes

- `source.mac`
- `destination.mac`

### Additional context

As per ECS https://www.elastic.co/docs/reference/ecs/ecs-destination & https://www.elastic.co/docs/reference/ecs/ecs-source

thompson-tomo:
Fixes #2416

## Changes

Adds in attributes to source & destination namespaces to support packet tracing based on ecs.
Added fields are:

- *.mac

### DIFF
--- a/model/destination/registry.yaml
+++ b/model/destination/registry.yaml
@@ -23,18 +23,6 @@ groups:
         stability: development
         brief: 'Destination port number'
         examples: [3389, 2888]
-      - id: destination.interface.id
-        stability: development
-        type: int
-        brief: Destination Interface Id
-        note: Interface ID as reported by an observer (typically SNMP interface ID).
-        examples: [10]
-      - id: destination.interface.name
-        stability: development
-        type: string
-        brief: Destination Interface Name
-        note: Interface name as reported by the system.
-        examples: ["eth0"]
       - id: destination.mac
         stability: development
         type: string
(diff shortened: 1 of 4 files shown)


thompson-tomo:
Awaiting review

thompson-tomo:
@trisch-me could u review as migration of ecs fields. Also is interface.alias needing to be migrated given that we already have *.zone which seems to be the same.

model/destination/registry.yaml
```diff
@@ -23,3 +23,41 @@ groups:
...
+      - id: destination.interface.name
+        stability: development
+        type: string
+        brief: Destination Interface Name
+        note: Interface name as reported by the system.
+        examples: ["eth0"]
+      - id: destination.mac
+        stability: development
+        type: string
+        brief: 'Destination mac address'
```
trisch-me:
```suggestion
        brief: Destination mac address
```

model/source/registry.yaml
```diff
@@ -23,3 +23,41 @@ groups:
...
+      - id: source.interface.name
+        stability: development
+        type: string
+        brief: Source Interface Name
+        note: Interface name as reported by the system.
+        examples: ["eth0"]
+      - id: source.mac
+        stability: development
+        type: string
+        brief: 'Source mac address'
```
trisch-me:
```suggestion
        brief: Source mac address
```

trisch-me:
Hey @thompson-tomo 
From ECS standard only `mac` is correct field adoption under source or destination.

All other fields are located under `observer` ingress/egress and putting those under source/destination is semantically ambiguous as it might be seen as source/destination host configuration where in reality it’s observer configuration, the device that received traffic

thompson-tomo:
So on looking at the description of host it mentions that it could be a switches hence I am not sure if we really need a dedicated namespace for observer.

I would rather use host to describe the computing instance which is running the observation software. The software can then be described by the app/Service namespace.

PR #2422, opened 2025-06-24
More OS properties #1258

### LINKED ISSUE #1258: Introduction of OS platform attribute
### Area(s)

area:os

I want to easily see all my logs from linux machines, just like i can with ECS.

Introduction of a new field -> os.platform which is a mapping of highlevel OS groups ie Linux, Windows etc

_No response_

### Additional context

https://www.elastic.co/guide/en/ecs/current/ecs-os.html

thompson-tomo:
Fixes #1258

## Changes


docs/registry/attributes/os.md
```diff
@@ -10,13 +10,57 @@ The operating system (OS) on which the process represented by this resource is r
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
 | <a id="os-build-id" href="#os-build-id">`os.build_id`</a> | string | Unique identifier for a particular build or compilation of the operating system. | `TQ3C.230805.001.B2`; `20E247`; `22621` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="os-classification" href="#os-classification">`os.classification`</a> | string | A classification of the OS which corresponds to the upstram distro. | `rhel`; `fedora`; `arch` | ![Development](https://img.shields.io/badge/-development-blue) |
```
LikeTheSalad:
```suggestion
| <a id="os-classification" href="#os-classification">`os.classification`</a> | string | A classification of the OS which corresponds to the upstream distro. | `rhel`; `fedora`; `arch` | ![Development](https://img.shields.io/badge/-development-blue) |
```

docs/registry/attributes/os.md
```diff
@@ -10,13 +10,57 @@ The operating system (OS) on which the process represented by this resource is r
...
+| `slackware` | Slackware | ![Development](https://img.shields.io/badge/-development-blue) |
+| `ubuntu` | Ubuntu | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[1]:** This would usually be OS by apple
+
+---
+
+`os.family` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
```
LikeTheSalad: I'm not too familiar with the Apple ecosystem, but I'm wondering what value we should provide on `os.family` for `iOS`, for example? Is `macosx` generic enough to be an `os.family` value that covers it all from Apple? - On the same iOS note, I'm guessing its value for `os.classification` should be `darwin`?

thompson-tomo: correct about Darwin. in relation to the family for ios, perhaps i update the message to be something like

> `os.family` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, the `os.type` should be used

LikeTheSalad: I see. Though I'm now concerned about potential misunderstandings of the exact differences between `family` and `type`. Based on how they're described here, they seem to overlap a lot. Can you elaborate on what kind of information we might need on `family` that cannot be added to `type`?

thompson-tomo: So from the extended list of os types, the only os types which doesn't have a family is ios, tvos, watch os etc. Based on that, it probs make more sense to either
- rename macosx to AppleOS
- Add AppleOS

thompson-tomo: Have added apple to the family enum which should make things clearer and also an entity to show usage  of the new attributes.

model/os/registry.yaml
```diff
@@ -75,6 +75,22 @@ groups:
...
+        stability: development
+        brief: 'OS family (such as redhat, debian, freebsd, windows).'
+        note: 'Theses values could also be an os.name'
+        examples:
+          - ['android']
+          - ['rhel']
+          - ['fedora']
+          - ['arch']
+          - ['darwin']
+          - ['ubuntu']
```
LikeTheSalad, line 96: I'm trying to understand the difference between `type` and `family`, and I think I got an idea that `type` would be the highest level definition of an OS, and `family` would be a subcategory of a `type` that covers a lot of OS types underneath. If that's correct, I'm confused about this item `ubuntu`, wouldn't it have to be `debian` instead?

model/os/registry.yaml
```diff
@@ -75,6 +75,22 @@ groups:
...
         examples: ['iOS', 'Android', 'Ubuntu']
+      - id: os.family
+        type: string[]
+        stability: development
+        brief: 'OS family (such as redhat, debian, freebsd, windows).'
+        note: 'Theses values could also be an os.name'
+        examples:
+          - ['android']
+          - ['rhel']
+          - ['fedora']
```
LikeTheSalad: Wouldn't this be part of the `rhel` family?

model/os/registry.yaml
```diff
@@ -75,6 +75,22 @@ groups:
...
+          - ['arch']
+          - ['darwin']
+          - ['ubuntu']
+          - ['opensuse']
+          - ['gentoo']
+          - ['slackware']
+          - ['android']
```
LikeTheSalad: This entry is duplicated.

model/os/registry.yaml
```diff
@@ -75,6 +75,22 @@ groups:
         stability: development
         brief: 'Human readable operating system name.'
         examples: ['iOS', 'Android', 'Ubuntu']
+      - id: os.family
+        type: string[]
```
LikeTheSalad, line 85: Why an array? I would expect only one family per OS, so it would also have to have predefined values, as with `type`.

thompson-tomo: On linux it is an array https://man.linuxreviews.org/man5/os-release.5.html. Note the identifier would be ID_LIKE.

I would have liked to use an enum as well but am blocked due to https://github.com/open-telemetry/weaver/issues/329

model/os/registry.yaml
thompson-tomo: As per https://man.linuxreviews.org/man5/os-release.5.html centos would report rhel fedora

model/os/registry.yaml
thompson-tomo: Will remove it.

model/os/registry.yaml
thompson-tomo: So the discussion in the semconv meeting was that type should effectively refer to the base ie unix & windows. There is a discussion in the linked PR.

If os was Ubuntu server, then the family would be Ubuntu & debian hence another reason for an array. This way unbuntu mate would also be the same

model/os/registry.yaml
```diff
@@ -70,11 +70,31 @@ groups:
...
         type: string
         stability: development
         brief: 'Human readable operating system name.'
-        examples: ['iOS', 'Android', 'Ubuntu']
+        examples: ['Centos Stream', 'Arch Linux', 'Ubuntu']
+      - id: os.family
+        type: string[]
+        stability: development
+        brief: 'A collection of `os.id` that this os is like.'
+        note: 'This relationship is to only be desribed from the derivitative of the parent ie for Fedora this would be RHEL but RHEL would not have fedora.'
```
KalleOlaviNiemitalo: A few typos here: `desribed`, `derivitative`.

model/os/registry.yaml
```diff
@@ -70,11 +70,33 @@ groups:
           Human readable (not intended to be parsed) OS version information,
           like e.g. reported by `ver` or `lsb_release -a` commands.
         examples: ['Microsoft Windows [Version 10.0.18363.778]', 'Ubuntu 18.04.1 LTS']
+      - id: os.id
+        type: string
+        stability: development
+        brief: 'A string to uniquely identify the os without the version'
+        note: 'On linux this is to be the id from the os.release file'
```
braydonk:
```suggestion
        note: 'On Linux, this is the `ID` field from the `os-release` file.'
```

braydonk: We should also be clear here about what our expectations for other platforms would be. What should this be for Windows and Mac? Should this be populated at all on non-linux? If it's linux only, we may want to consider renaming to something like `os.release.*` to be more specific about attributes that come from the `os-release` file and don't have equivalents on other platforms.

model/os/registry.yaml
```diff
@@ -70,11 +70,33 @@ groups:
           Human readable (not intended to be parsed) OS version information,
           like e.g. reported by `ver` or `lsb_release -a` commands.
         examples: ['Microsoft Windows [Version 10.0.18363.778]', 'Ubuntu 18.04.1 LTS']
+      - id: os.id
+        type: string
+        stability: development
+        brief: 'A string to uniquely identify the os without the version'
```
braydonk:
```suggestion
        brief: 'A string to uniquely identify the os without the version.'
```

thompson-tomo:
I don't see a need to separate them given that all properties except for os.family applies to windows os & a different attribute os.variant is the only one which may or may not apply to macos.

What I do agree with is in general there needs to be a better way to describe examples of entities especially with attribute groups not in v2 and certainly won't be in the current form.

model/os/registry.yaml
thompson-tomo: Alot more notes have been added explaining where values can be sourced from if not provided by the framework.

thompson-tomo:
Have been thinking about it some more and the usage of notes to describe the sources of values seems to have produced a fractured reading experience with lots of small tables. So I have explored using annotations to generate a table to present the sources. PR currently contains both options so looking for feedback if the table is a good option then notes will be cleaned up.

trask:

Issue #2324, opened 2025-06-04
Clarify OS resource attributes for Linux


smaddock:
### Area(s)

area:os

### What's missing?

This issue is a continuation from a [Slack conversation](https://cloud-native.slack.com/archives/C05CTFE9U4A/p1748963627842699), trying to get clarity around the intent of the `os` attribute definitions. (I understand they are all still in "development" status and subject to change.) The existing descriptions are ambiguous, so I was going more off of the example values, but even then there were questions for Linux at least of OS versions vs kernel builds and whether to get the most accurate result or the most consistent result. I checked 5 current linux distros (2 Debian-based, 2 Red Hat-based, and Alpine) for common paths and tools, and have attached a summary of their outputs here.

[alpine3.txt](https://github.com/user-attachments/files/20597694/alpine3.txt)
[centos9stream.txt](https://github.com/user-attachments/files/20597695/centos9stream.txt)
[debian12.txt](https://github.com/user-attachments/files/20597698/debian12.txt)
[fedora41.txt](https://github.com/user-attachments/files/20597697/fedora41.txt)
[ubuntu24lts.txt](https://github.com/user-attachments/files/20597696/ubuntu24lts.txt)

cc @braydonk 

### Describe the solution you'd like

Additional guidance, preferably in the semantic convention specs themselves, of what the intent of each `os` attribute is for. Perhaps recommendations on where to pull each value from.

joaopgrassi:
I looked into it a bit more, and maybe we can discuss whether this makes sense or not. For the current fields we have:

- `os.name`: For linux use the `NAME` from `/etc/os-release`

- `os.description`: For Linux use the `PRETTY_NAME` from `/etc/os-release`

- `os.version`: More complicated. 
  - Catch all attempt: `VERSION_ID` from `/etc/os-release`. May not be accurate, missing point releases. Could also try `DISTRIB_DESCRIPTION` from `/etc/lsb-release`
  - Distro-specific files: `/etc/debian_version`, `/etc/redhat-release`, `/etc/centos-release`, `cat /etc/arch-release`, `cat /etc/alpine-release`


Then If we want to also record the kernel version, maybe then we introduce a new field for it. In ECS there's `os.kernel` https://www.elastic.co/docs/reference/ecs/ecs-os#field-os-kernel but I think that name is not clear it's a version. Maybe `os.kernel_version` or something is better.

smaddock:
[...]

Issue #66, opened 2023-05-31
Add kernel related semantic conventions


mx-psi:
I would like to add three new resource-level semantic conventions to the `os.*` namespace that include information about the kernel used by the system. Roughly, they could be defined as follows:

| Attribute  | Type | Description  | Examples  | Requirement Level |
|---|---|---|---|---|
| `os.kernel.name` | string | The kernel name, e.g. on a Unix system, the output of `uname -s`. | `Linux` | Recommended |
| `os.kernel.release` | string | The kernel release, e.g. on a Unix system, the output of `uname -r` | `5.19.0-42-generic` | Recommended | 
| `os.kernel.version` | string | The kernel version, e.g. on a Unix system, the output of `uname -v` | `#43~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Apr 21 16:51:08 UTC 2` | Recommended |

These would be useful to identify a resource as coming from a machine with a specific kernel version, which can be useful in debugging and troubleshooting systems.

It's been a while since I have contributed to semantic conventions so I have a few open questions, please bear with me :)

### Open questions

1. Elastic Common Schema already defines `os.kernel` as *"Operating system kernel version as a raw string."* and gives `4.4.0-112-generic` as an example value (so, effectively what I called here `os.kernel.release`). I think we should use the more specific `os.kernel.release` for this instead, since `os.kernel` seems more like a namespace to me, and there's several meanings for 'version'. Is that okay?
2. Should `os.kernel.name` have a pre-defined set of values? If so, what values are in scope?
3. Are the values as described sufficiently cross-platform? At a minimum, do they have a reasonable equivalent on Windows and macOS?
4. Is the proposed description well-specified enough? Are `uname` output and flags consistent enough to use it as a standard? Based on the description of `host.id` it sounds like it should be okay, but I am happy to consider a different definition if something is suggested.

trisch-me:
Hi @mx-psi, I have checked we are using os.kernel along in security in around 100 different integrations. 
Saying that I like this proposal to have `os.kernel` as a namespace. For us it would be a breaking change to use `os.kernel.release` instead of just `os.kernel`. I'm checking now on how big would be this breakage for us

trisch-me:
Hey @mx-psi 
[...]

PR #2423, opened 2025-06-24
Refocus of os types #1257

### LINKED ISSUE #1257: Add additional os.types
### Area(s)

area:os

I want to ensure my OS is accurately reported

The following OS's are added to the strongly typed OS'S:
- android
- Browser
- ios
- MacCatalyst
- TVOS
- WASI
- WatchOS

_No response_

### Additional context

https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem?view=net-8.0

thompson-tomo:
Closes #1257 

## Changes

Deprecate all existing os types and instead use the field to describe the type of the OS ie windows or Unix. The existing types should be the `os.name` and/or `unix.kernel.name` with the later handled via #2641

### DIFF
--- a/model/os/registry.yaml
+++ b/model/os/registry.yaml
@@ -19,7 +19,7 @@ groups:
             - id: linux
               value: 'linux'
               brief: "Linux"
-              deprecated: 
+              deprecated:
                 reason: renamed
                 renamed_to: unix
               stability: development


model/os/registry.yaml
```diff
@@ -17,11 +17,11 @@ groups:
               stability: development
             - id: linux
               value: 'linux'
-              brief: "Linux"
```
joaopgrassi, line 20: I'm not sure I agree with this change. The attribute is called `os.type`, so to me, `Linux` for a "type" of OS is correct. If we go by this change, then `Microsoft Windows` here is also wrong then, because that's analogous to referring to linux distros as `Linux`. 

We already have `os.name` to describe what you added here, for example `Ubuntu`.

joaopgrassi (changes requested):
Blocking as to this is not simply adding more OS types. It is changing the behavior of the attribute entirely. 

The links you added to the issue description, for example from ECS https://www.elastic.co/docs/reference/ecs/ecs-os#field-os-type also have a `os.type` which is pretty much the same as what we have today.

model/os/registry.yaml
thompson-tomo: i disagree with that summary as if the type is supposed to be some generic term then why do we have values such as dragonflybsd, freebsd etc. Hence the current solution is a mismatch of items which are not equals. Here is another example of issues with the solution: https://github.com/open-telemetry/semantic-conventions/issues/2324

For instance solaris is an OS which can be deployed onto hw but Linux is not an actual OS but rather a family of os's. This pain is addressed by https://github.com/open-telemetry/semantic-conventions/pull/2422 which introduces more fields

In regards to windows it is not wrong as that is an OS which can be used.

In relation to the ecs example they use the clear options of ios, macos but in the current solution these are bundled under darwin which again is not consistent with the approach taken with the bsd.

joaopgrassi: Sure, the fact that today there are "wrong" values for type does not mean that we should add yet another thing. Maybe we should just fix and remove the "wrong" values for type, and use the `os.name` for what you are proposing here instead.

thompson-tomo: I did consider that however if you look at the examples for linux, 

name are items such as following:

- Debian GNU/Linux
- Ubuntu 24.04.2 LTS
- Fedora Linux
- CentOS Stream
- Alpine Linux

and before you suggest using those names as description. That field should be the pretty name as that is in all cases as descriptive if not more so.

- Debian GNU/Linux 12 (bookworm)
- Ubuntu 24.04.2 LTS but potentially could be Ubuntu 24.04.2 LTS (Noble Numbat)
- Fedora Linux 41 (Cloud Edition)
- CentOS Stream 9
- Alpine Linux v3.22

The way i look at it is like a car. if Someone asks me:
- what **type** of car i have, i would say. "Golf" 
- what is the name of that car (pointing to mine), I would tell them Golf GTI 
- to describe that car (pointing to mine), i would tell them 2023 VM Golf GTI

The same analogy can be made for phones, hence the changes being proposed in conjunction with the other PR. This approach also complements entities and allows for more use cases especially in metrics.

joaopgrassi: I think we probably should do a overhaul into this entirely. Your proposoal is also not compatible with ECS, https://www.elastic.co/docs/reference/ecs/ecs-os#field-os-type. There they specifically say type:

> Use the os.type field to categorize the operating system into one of the broad commercial families.

Which I agree with. Maybe we just align completely with ECS here. @open-telemetry/semconv-system-approvers @open-telemetry/specs-semconv-approvers

thompson-tomo: I would say that the changes are compliant with that definition as The only values not brought across from ecs is linux & Unix which is replaced by the commercial name of the os family. In fact this change mirrors the approach done for os by apple where ios & macos were listed.

joaopgrassi: In my mind, what could makes sense is something like this:

### `os.type` 

Really the type of the OS and nothing more. We don't need many values here.

- `linux`
- `unix`
- `nt`

### `os.family`

The family of OSs that are under each type above. [For ex this for Linux distros](https://community.nethserver.org/uploads/db8506/original/2X/b/b56cdb4a1e3a82020f27d0b9ef36bb6b8553712d.jpeg)

- `Debian`
- `Arch`
- `Red Hat`
- `Fedora`
- `BSD`
- `Darwin` https://en.wikipedia.org/wiki/Darwin_(operating_system)
- `AOSP` (https://en.wikipedia.org/wiki/Android_(operating_system)#Development)
- `Windows NT`
- `Solaris`

### `os.name`

The name of the os which are inside the family above. 

- `Ubuntu`
- `Arch Linux`
- `CentOS`
- `Linux Mint`
- `macOS Sequoia`
- `iOS`
- `watchOS`
- `Android`


I think this way we have a good compromise of OK classification vs not having a lot of fields. We have to remember that these will end up as attributes which costs money for people in the end, so the questions is always how much info is really needed here.

thompson-tomo: That is very similar to what I have already proposed with #the #2422 ;)

What i would suggest now is:
- deprecate Os.type to avoid confusion via this issue
- I switch family & classification values around in #2422
- Use what you define as os.type here to either be ~~os.class~~, os.classification or os.base. I actually think os.base is the most fitting & clear name.

joaopgrassi: Will have to have a further discussion about getting rid of `os.type`. Although I also do not like it, as it's confusing (`os.base` is much better) but then we drift even more from ECS.

thompson-tomo: Agree about drifting from ecs but I feel this would be the easiest to follow as a reader as all options would result in a difference in behaviour/conflict.

thompson-tomo: Suggestions from semconv meeting has been implemented and is inline with above comments.

thompson-tomo:
@joaopgrassi can you review and unblock as implementation is following what was discussed in the semconv.

model/os/registry.yaml
```diff
@@ -11,45 +11,63 @@ groups:
...
+              deprecated: Should use `unix` instead'
+              stability: development
+            - id: unix
+              value: 'unix'
+              brief: "Unix"
               stability: development
             - id: darwin
               value: 'darwin'
               brief: "Apple Darwin"
+              deprecated: Should be set as the `os.family` instead.
```
joaopgrassi: I guess we need to instruct which type to use for these that don't have a replacement now. May not be obvious for everyone so better be specific.

model/os/registry.yaml
```diff
@@ -11,45 +11,63 @@ groups:
...
         type:
           members:
+            - id: windowsnt
+              value: 'windowsnt'
+              brief: "Microsoft Windows NT based OS"
+              stability: development
             - id: windows
               value: 'windows'
               brief: "Microsoft Windows"
+              deprecated: Should use `windowsnt` instead'
```
joaopgrassi: Why should and not just Use.. instead? Same for the Linux one

docs/resource/os.md
```diff
@@ -18,7 +18,7 @@ In case of virtualized environments, this is the operating system as it is obser
 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
-| [`os.type`](/docs/registry/attributes/os.md) | string | The operating system type. | `windows`; `linux`; `darwin` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`os.type`](/docs/registry/attributes/os.md) | string | The operating system type. | `windowsnt`; `windows`; `linux` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
```
joaopgrassi: `windows` should not be showing here in the examples, right? I think it may be coming from some other model file.

thompson-tomo: Actually examples do not filter out deprecated values hence we have #2497 to address it.

model/os/registry.yaml
thompson-tomo: Will update once weaver is updated as we can then have structured deprecation.

Also keep in mind currently deprecated enum members are not rendered #2496

model/os/registry.yaml
thompson-tomo: The usage of a Structured deprecation will remove this phrasing.

model/os/registry.yaml
thompson-tomo: Deprecated enum members will now show once #2623 is merged. I still need to do the structured deprecation but that will break build.

docs/resource/os.md
joaopgrassi: I sent a PR to address it https://github.com/open-telemetry/semantic-conventions/pull/2624

model/os/registry.yaml
braydonk: I still think it's valuable to avoid the word "Should". I'm unsure if this is a specific rule in this repo, but I generally try to avoid usage of normative language words like "should", "must", "may" if I'm not going to capitalize it.

model/os/registry.yaml
```diff
@@ -59,6 +77,15 @@ groups:
             - id: zos
               value: 'zos'
               brief: "IBM z/OS"
+              deprecated: "Use `os.name` instead"
+              stability: development
+            - id: wasm
+              value: 'wasm'
+              brief: "WASM OS"
```
braydonk: What is this referring to? Is there a WASM-based operating system? I can't seem to find anything about it.

model/os/registry.yaml
```diff
@@ -11,45 +11,63 @@ groups:
       - id: os.type
         type:
           members:
+            - id: windowsnt
+              value: 'windowsnt'
```
braydonk: Is this stylized `windowsnt` string based on some industry standard? Feels like a major breakage and the value of breaking is unclear to me.

model/os/registry.yaml
```diff
@@ -59,6 +77,15 @@ groups:
             - id: zos
               value: 'zos'
               brief: "IBM z/OS"
+              deprecated: "Use `os.name` instead"
```
braydonk: `zos` is a real unique operating system type, I don't think we should be removing it.

model/os/registry.yaml
```diff
@@ -11,45 +11,63 @@ groups:
...
               value: 'windows'
               brief: "Microsoft Windows"
+              deprecated: Should use `windowsnt` instead'
               stability: development
             - id: linux
               value: 'linux'
               brief: "Linux"
+              deprecated: Should use `unix` instead'
+              stability: development
+            - id: unix
```
braydonk, line 26: I'm not sold on putting all unixes together. I think there is still value in separating `linux`, `darwin`, and `bsd` at least. While it's true they are all unix-likes, they are unique enough and are treated differently enough in the minds of most users to deserve their own attributes. I think moving all the `*bsd` attributes to `os.name` is good so we can keep that, but I think this should restore `linux`, `darwin`, and `bsd`.

model/os/registry.yaml
thompson-tomo: It is defined in the rust & dotnet docs. For reference https://doc.rust-lang.org/std/env/consts/constant.FAMILY.html

And also an article on it https://opensource.microsoft.com/blog/2025/03/26/hyperlight-wasm-fast-secure-and-os-free/

model/os/registry.yaml
thompson-tomo: It is because Microsoft has multiple distinct windows os platforms https://learn.microsoft.com/en-us/dotnet/api/system.platformid?view=net-9.0

model/os/registry.yaml
thompson-tomo: So looking at implementations and language api's the common functionality is being able to determine the OS type which is what is being expressed by this attribute. You can then look at the os.name &/or os.family to get that specific info, you are after.

model/os/registry.yaml
thompson-tomo: From my research it is a Unix os.

model/os/registry.yaml
mx-psi: From your link all of the other Windows OS platforms are referred as "no longer in use". 

While it is theoretically possible that someone will implement these Windows platforms in the future or that there will be other Windows-based platforms, the truth is that 'Windows' is synonymous with Windows NT today for our users, and I think `windows` is much less confusing so I think we should not make this change

model/os/registry.yaml
mx-psi: I think the previous state is more useful than just a 'unix' value. It may be too specific today, but I don't think we should jump to the other extreme.

model/os/registry.yaml
mx-psi: Unless we have a specific use case where this would be used today I don't think we should add this value. It's debatable whether this is an OS in the first place so I would keep it out.

model/os/registry.yaml
```diff
@@ -59,6 +77,15 @@ groups:
...
               brief: "IBM z/OS"
+              deprecated: "Use `os.name` instead"
+              stability: development
+            - id: wasm
+              value: 'wasm'
+              brief: "WASM OS"
+              stability: development
+            - id: itron
+              value: 'itron'
+              brief: "Industrial TRON"
```
mx-psi: I suppose this refers to https://en.wikipedia.org/wiki/ITRON_project? My stance with this is the same as the WASM value (though this is more clearly an OS): we should not add this unless we have a real use case today.

thompson-tomo: In the rust resource detector, this would be one of the enum values coming from the sdk.

model/os/registry.yaml
thompson-tomo: Rust, dotnet & java resource detector would all be using sdk which returns this flag

model/os/registry.yaml
thompson-tomo: You have os.name/os.type to be more specific. And in fact this is what I was told to implement in the sem conv meeting as I had originally extended type to include alot of other common os but that was not wanted due to it becoming a never ending list. At the same time we also needing to migrate the other properties from ecs, hence the request was to focus type on the minimal items and provide the other properties.

Also this approach here would nicely complement the solution I have proposed in #66 as if type is Unix, then you can also provide details of the Unix.kernel which could be Linux, Darwin etc.

model/os/registry.yaml
mx-psi: Does the OTel Rust library have official support for ITRON platforms? The Rust compiler has [Tier 3 support](https://doc.rust-lang.org/nightly/rustc/platform-support/kmc-solid.html) for this target which means:

> Tier 3 targets are those which the Rust codebase has support for, but which the Rust project does not build or test automatically, so they may or may not work. Official builds are not available.

I don't think this counts as a real use case

model/os/registry.yaml
mx-psi: I don't think WASM is the right value to return in an OS field. To make a comparison, we wouldn't consider JVM to be an OS, but there are significant conceptual similarities between both (they both are instructions for a particular VM that is meant to be platform-agnostic).

I think the right call if you are running in WASM and are unable to fetch what the OS is is to not set this field

thompson-tomo: It's interesting that so many languages do classify it as an OS. To avoid holding this PR up, I will split out wasm.

thompson-tomo: Been split out to separate pr

model/os/registry.yaml
thompson-tomo: To further support the simplification proposed here #2641 has now been created to express the kernel type ie linux, Darwin. 

That pr provides another way to maintain the same granularity while introducing consistent attribute usage and further migration of ecs.

model/os/registry.yaml
thompson-tomo: Been split off.

trask:

PR #2424, opened 2025-06-24
Add more thread details #2409

### LINKED ISSUE #2409: Encorporate thread details into process
### Area(s)

area:thread

### What's missing?

Conventions for capturing thread details

A sub namespace with process to describe the thread

- process.thread.id
- process.thread.background
- process.thread.name
- process.thread.pooled
- process.thread.priority

### Additional context 
https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread?view=net-9.0#properties

thompson-tomo:
Fixes #2409

## Changes

Introduces more details about the thread being described.

### DIFF
--- a/model/thread/registry.yaml
+++ b/model/thread/registry.yaml
@@ -17,3 +17,21 @@ groups:
         brief: >
           Current thread name.
         examples: main
+      - id: thread.background
+        type: boolean
+        stability: development
+        brief: >
+          The thread is a background thread.
+        examples: [true, false]
+      - id: thread.pooled
+        type: boolean
+        stability: development
+        brief: >
+          The thread belongs to the managed thread pool.
+        examples: [true, false]
+      - id: thread.priority
+        type: string
+        stability: development
+        brief: >
+          The scheduling priority of a thread.
+        examples: [ "Lowest", "BelowNormal", "Normal", "AboveNormal", "Highest" ]
--- a/.chloggen/TEMPLATE copy.yaml
+++ b/.chloggen/TEMPLATE copy.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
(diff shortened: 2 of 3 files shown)


model/thread/registry.yaml
```diff
@@ -17,3 +17,21 @@ groups:
...
+        stability: development
+        brief: >
+          The thread belongs to the managed thread pool.
+        examples: [true, false]
+      - id: thread.priority
+        type: string
+        stability: development
+        brief: >
+          The scheduling priority of a thread.
+        examples: [ "Lowest", "BelowNormal", "Normal", "AboveNormal", "Highest" ]
```
joaopgrassi: These seem to be all .NET examples. Maybe list ones for Java, which are integers? https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#setPriority-int-

thompson-tomo: Sure will add a note mentioning it should be as per obtained via sdk & change the example to include a number & text description.

thompson-tomo: Changes made and pushed up.

joaopgrassi:
Is there any background in why you want these fields added? Is there instrumentation work wanting these fields? I tried to look at [ECS](https://www.elastic.co/docs/reference/ecs/ecs-process) and there it is also not defined, which makes me wonder how relevant these really are. CC @open-telemetry/semconv-system-approvers

thompson-tomo:
I see them as useful metadata to provide additional context about why things might have taken as long as they did.

The priority would be useful addition to https://opentelemetry.io/docs/specs/semconv/runtime/go-metrics/#metric-goscheduleduration for example and also metrics about locks

jsuereth:
Is there an instrumentation prototype with these new fields being defined we can look at?

thompson-tomo:
So the fields have been updated to reflect what is needed for the collector with links in issue to some of them.

PR #630, opened 2024-01-08
Clarify that service.* conventions apply to all telemetry sources

jack-berg:
The question of whether the `service.*` conventions are applicable only to web services is an important one that has come up several times. If the answer is yes, then everything that isn't a web service needs its own version of `service.name`, `service.instance.id`, `service.namespace`, `service.version` to uniquely define the thing producing telemetry. We've discussed this at length several times and while we don't have anything written down yet in the spec / semantic-conventions, the actions we've taken (i.e. rejecting alternatives like `telemetry.source`, `app.name`, etc) confirm that the `service.*` attributes are applicable to _all_ telemetry services, not some narrower subset that some people consider a web service.

This PR aims to clarify this to avoid repeating the same discussion.

Some PRs, issues that are related:

- https://github.com/open-telemetry/oteps/issues/244
- https://github.com/open-telemetry/opentelemetry-specification/issues/3791
- https://github.com/open-telemetry/semantic-conventions/pull/557
- open-telemetry/opentelemetry-specification#2050
- open-telemetry/opentelemetry-specification#2111
- open-telemetry/opentelemetry-specification#2115
- open-telemetry/opentelemetry-specification#2192
- open-telemetry/oteps#194

PR #2430, opened 2025-06-25
Add more app details #2346

### LINKED ISSUE #2346: App details lack key info

thompson-tomo:
Fixes #2346

## Changes

Adds in functionality to record the details of the app which the telemetry is coming from to enable support for non service based scenarios ie Android App.

### DIFF
--- a/model/app/registry.yaml
+++ b/model/app/registry.yaml
@@ -42,18 +42,6 @@ groups:
           (so the empty/unspecified namespace is simply one more valid namespace).
           Zero-length namespace string is assumed equal to unspecified namespace.
         examples: ["Shop"]
-      - id: app.roles
-        type: string[]
-        brief: >
-          What role this app can perform.
-        note: >
-          This doesn't correspond to the roles being performed as a role can be disabled.
-
-          To discover the active roles you can look at `service.roles`
-        examples:
-          - ["ui", "background_tasks"]
-          - ["background_tasks", "console"]
-        stability: development
       - id: app.installation.id
         type: string
         stability: development
--- a/.chloggen/add_app_details.yaml
+++ b/.chloggen/add_app_details.yaml
@@ -10,7 +10,7 @@ change_type: enhancement
 component: app
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Adds more app attributes to cater for non service based telemetry.
+note: Adds more app attributes to describe the running app ie the name & version.
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 # The values here must be integers.
--- a/docs/registry/attributes/app.md
+++ b/docs/registry/attributes/app.md
@@ -19,14 +19,13 @@ Describes attributes related to client-side applications (e.g. web apps or mobil
 | <a id="app-jank-threshold" href="#app-jank-threshold">`app.jank.threshold`</a> | ![Development](https://img.shields.io/badge/-development-blue) | double | The minimum rendering threshold for this jank, in seconds. | `0.016`; `0.7`; `1.024` |
 | <a id="app-name" href="#app-name">`app.name`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | Logical name of the app. [4] | `shoppingcart` |
 | <a id="app-namespace" href="#app-namespace">`app.namespace`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | A namespace for `app.name`. [5] | `Shop` |
-| <a id="app-roles" href="#app-roles">`app.roles`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string[] | What role this app can perform. [6] | `["ui", "background_tasks"]`; `["background_tasks", "console"]` |
 | <a id="app-screen-coordinate-x" href="#app-screen-coordinate-x">`app.screen.coordinate.x`</a> | ![Development](https://img.shields.io/badge/-development-blue) | int | The x (horizontal) coordinate of a screen coordinate, in screen pixels. | `0`; `131` |
 | <a id="app-screen-coordinate-y" href="#app-screen-coordinate-y">`app.screen.coordinate.y`</a> | ![Development](https://img.shields.io/badge/-development-blue) | int | The y (vertical) component of a screen coordinate, in screen pixels. | `12`; `99` |
-| <a id="app-screen-id" href="#app-screen-id">`app.screen.id`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | An identifier that uniquely differentiates this screen from other screens in the same application. [7] | `f9bc787d-ff05-48ad-90e1-fca1d46130b3`; `com.example.app.MainActivity`; `com.example.shop.ProductDetailFragment`; `MyApp.ProfileView`; `MyApp.ProfileViewController` |
-| <a id="app-screen-name" href="#app-screen-name">`app.screen.name`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The name of an application screen. [8] | `MainActivity`; `ProductDetailFragment`; `ProfileView`; `ProfileViewController` |
+| <a id="app-screen-id" href="#app-screen-id">`app.screen.id`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | An identifier that uniquely differentiates this screen from other screens in the same application. [6] | `f9bc787d-ff05-48ad-90e1-fca1d46130b3`; `com.example.app.MainActivity`; `com.example.shop.ProductDetailFragment`; `MyApp.ProfileView`; `MyApp.ProfileViewController` |
+| <a id="app-screen-name" href="#app-screen-name">`app.screen.name`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The name of an application screen. [7] | `MainActivity`; `ProductDetailFragment`; `ProfileView`; `ProfileViewController` |
 | <a id="app-version" href="#app-version">`app.version`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The version string of the app. The format is not defined by these conventions. | `2.0.0`; `a01dbef8a` |
-| <a id="app-widget-id" href="#app-widget-id">`app.widget.id`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | An identifier that uniquely differentiates this widget from other widgets in the same application. [9] | `f9bc787d-ff05-48ad-90e1-fca1d46130b3`; `submit_order_1829` |
-| <a id="app-widget-name" href="#app-widget-name">`app.widget.name`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The name of an application widget. [10] | `submit`; `attack`; `Clear Cart` |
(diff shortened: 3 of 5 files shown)


thompson-tomo:
Awaiting review

model/app/registry.yaml
```diff
@@ -6,6 +6,75 @@ groups:
...
+        stability: development
+      - id: app.name
+        type: string
+        brief: >
+          Logical name of the app.
+        note: >
+          MUST be the same for all installations of the app.
+        examples: ["shoppingcart"]
+        stability: development
+      - id: app.version
```
LikeTheSalad, line 25: It seems to me that `app.name` and `app.version` are duplicates of `service.name` and `service.version` values.

model/app/registry.yaml
```diff
@@ -6,6 +6,75 @@ groups:
...
+        type: string
+        brief: >
+          The version string of the app. The format is not defined by these conventions.
+        examples: ["2.0.0", "a01dbef8a"]
+        stability: development
+      - id: app.namespace
```
LikeTheSalad, line 31: I know that this term is important for Android apps, although I don't see the value elsewhere. If we want it, it's probably better to put it into an `android`-specific namespace. WRT the `app` namespace, it seems to me that `app.id` makes more sense for different platforms.

model/app/registry.yaml
```diff
@@ -6,6 +6,75 @@ groups:
       Describes attributes related to client-side applications (e.g. web apps or mobile apps).
     stability: development
     attributes:
+      - id: app.id
+        type: string
+        brief: >
+          A unique identifier for the app.
+        note: >
+          MUST remain the same across all versions. For mobile applications this should correspond to the package id.
```
LikeTheSalad: I believe that "package id" might be Android-specific; I think they call it "bundle ID" on iOS, not sure about other platforms. Probably, we're better off not mentioning any platform-specific term?

model/app/registry.yaml
```diff
@@ -6,6 +6,75 @@ groups:
...
+              value: "service"
+              brief: The application runs in the backgroung without any ui.
+              stability: development
+            - id: web
+              value: "web"
+              brief: The application is interacted with via a web browser.
+              stability: development
+        brief: Describes the type of application which generated the telemetry
+        stability: development
+      - id: app.roles
```
LikeTheSalad: Tbh I don't find the value of adding `app.type` and `app.roles`, if anything, they might make this PR more difficult to merge because it sounds like things like "what enum values should be present in `app.type`" can be reduced to subjective opinions, leading to discussions around the lines of: "what is a service", "what can we call a web app that runs embedded in a mobile app", "what if different types can be set to the same app", "what overlapping issues can happen when a type is `service` and a role is `background_task` given that the service type's description has the word 'background' in it", with the last question going back again to "what is a service"... 😅

model/app/registry.yaml
thompson-tomo: The namespace is designed to be a Grouping of apps similar to how it is used elsewhere. 

An example would be if a product had both server & client apps, the app namespace would refer to the product hence shared across all apps.

model/app/registry.yaml
thompson-tomo: In some cases they will be the same however in other cases they will differ.

For instance you could have a server application which is comprised of multiple services, app provides a way to group the services into a versioned bundle.

model/app/registry.yaml
thompson-tomo: Agree use case is limited with the introduction to roles which was added at a later point in time. Will  likely remove it now.

model/app/registry.yaml
thompson-tomo: Have adjusted the wording to cater for other platforms.

breedx-splk (changes requested):
Hey @thompson-tomo 
There is a LOT lot lot of history around using `service` vs. `app` for client/mobile applications. And when I say a LOT, I mean an excessively long and drawn out history. On the client/mobile side, we have long ago conceded and agreed to use `service.name` (and friends) instead of having explicit `app.*` equivalents. 

Please see this PR and all of the related content linked from there: https://github.com/open-telemetry/semantic-conventions/pull/630

thompson-tomo:
@breedx-splk this doesn't change that, a service is still what is used to describe what is generating the telemetry. An app is 1 or more service's bundled together into an independently versioned package.

breedx-splk:
> @breedx-splk this doesn't change that, a service is still what is used to describe what is generating the telemetry. An app is 1 or more service's bundled together into an independently versioned package.

Oh interesting. Well, that's not what we've previously considered to be an "app"... so it would appear that you're trying to completely redefine something that we have already established. Look at line 14 of `app.md` where it clearly says:

```
An app used directly by end users — like mobile, web, or desktop.
```

I don't think we want to change that.

thompson-tomo:
No I am not redefining it, if you install a desktop application ie visual studio 2022 that would be considered 1 app. That app is made up of multiple services. The same is the case for web apps where you might have a frontend & backend services but it should still be considered 1 app.

thompson-tomo:
These changes are now even more relevant due to the introduction of app.build_id and the associated entity. Without the introduction of app.name etc the usability of app.build_id to uniquely identify an app, is highly unlikely especially given the description of that attribute.

thompson-tomo:
@breedx-splk Let me try to highlight the difference but complementary nature of service and app namespace/entity with the focus on the relationships.

Starting from a signal:

- A trace is captured by a service.
- There can be 1 or more instances of this service.
- These service instances are running with a process.
- A process requires a process executable to run.
- A process executable belongs to an app.
- An app has to be installed.
- An app installation occurs on to a host machine.

Or looking at it the other way (starting from the host):

- A host has 1 or more apps installed on it.
- Each app consists of 1 or more executables which when run creates a process
- Each process provides 1 or more telemetry services  with each instance of the service identifiable.

These definitions enable the identification of the app based on just the process executable which cannot be said for the service. Also the app namespace can also be used for capturing name + version of what is installed on a machine ie database servers.

Based on the above you can hopefully now see the usage for having both.

breedx-splk:
/sigh

This is still frustrating...although I think I have a slightly better understanding why and how we got here.

It still seems to me that @thompson-tomo is trying to introduce a new concept of "app" -- not new as in "this idea didn't ever exist before" but new in that it has not been previously codified in opentelemetry spec or semconv. @thompson-tomo is attempting to weave together this new idea of an "app" by taking a yarn from attributes and a thread from resources and viola...twist them together, and look, here, we have something that fits their mental model of an "app".

I pointed this out previously above, with regard to this statement:

> An app is 1 or more service's bundled together into an independently versioned package.

This is a new classification/grouping idea that you have introduced in this PR. Please correct me if you can find evidence of this existing elsewhere in spec/semconv...and no, the existence of an attribute with `app.` prefix doesn't count.

In your new message, with the two bulleted lists, I think you once again have tried to introduce an entirely new concept to otel under the guise of just adding app details. I could nitpick thru those lists, but I don't think it's worth doing....I just think that these 3 points are part of a conceptual model that isn't here yet or hasn't yet been written down in spec/semconv:

> * A process executable belongs to an app. (does it? can't an exec just be a service? why imply an app?)
> * An app has to be installed. (executables and services are installed, why is an app installed?)
> * Each app consists of 1 or more executables which when run creates a process (I'm unaware of any other art here that describes an app as consisting of 1 or more executables).

In the otel client world, this problem of incorrectly/strangely/inaccurately using `service.name` as the identifier for a mobile or web application goes back AT LEAST 4 years. I feel like maybe I understated the gravity of the history around this topic....at least it wasn't acknowledged. Don't believe me about the history? Check the SIG notes: https://docs.google.com/document/d/16Vsdh-DM72AfMg_FIt9yT9ExEWF4A_vRbQ3jRNBe09w/edit?tab=t.0#heading=h.yplevr950565 

```
2021-11-10
* Semantic conventions 
  * [...]
  * App.name
    * JS sdk-for-web sends service.name=unknown_service, it should instead send app.name
      * [jkwatson] why is JS sending any “service” namespace for browser? It’s not a service! [answer, I suspect: there isn’t yet a namespace for “app” or whatever, so they used what they could]
```

Still don't believe me? [Here's a cool PR from 2021 ](https://github.com/open-telemetry/opentelemetry-specification/pull/2115/files#diff-c454bdf5d340aa2bed2cb92d32707bc4c6710ebf1c8738f21700092f3e316686R102)that literally attempts to add `app.name` and `app.namespace`...sound familiar?  It should, because that's what this PR is also attempting to add. That PR was closed. 

But time passes, and the discussion got rebooted in 2023 (SIG notes):

```
2023-10-03
* [Surbhi A] [service.name](https://github.com/open-telemetry/semantic-conventions/blob/2c065be5e66bbaf00898b1a740bb3b3e47b8acc6/docs/resource/README.md#service) should be set to bundleId for unique identification instead of application name?
  * PR/issues related to app.name vs service.name
    * https://github.com/open-telemetry/opentelemetry-specification/pull/2115
I’ve requested that this PR be reopened as it was closed as stale
    * https://github.com/open-telemetry/opentelemetry-specification/pull/2111 
  * AI: reopen discussion in [#2115](https://github.com/open-telemetry/opentelemetry-specification/pull/2115)
    *Alternative - clarify that service.name / version applies to client apps
```

And there was so much thrash around this topic that the otel leadership stepped in to clarify:
https://github.com/open-telemetry/semantic-conventions/pull/630
...but the gravity/significance of that PR was summarily dismissed by a comment above. Meh.

Or maybe all of this history and previous discussion is just unimportant and this is trying to do something new or reboot the same discussions again? If so, let's be clear and straightforward and say this. If you're trying to introduce a new concept, you should be straightforward about that as well...but to me it sure feels like the old discussion coming back to life. 🧟🧟🧟🧟🧟🧟🧟🧟

thompson-tomo:
So this app concept is not new and is in fact already in semconv:

- https://opentelemetry.io/docs/specs/semconv/registry/entities/cloudfoundry/#cloudfoundry-app
- https://opentelemetry.io/docs/specs/semconv/registry/entities/gcp/#gcp-apphub-application
- https://opentelemetry.io/docs/specs/semconv/registry/entities/heroku/#heroku which will be updated via https://github.com/open-telemetry/semantic-conventions/pull/2843

If you take a look at the definition of [`service.version`](https://opentelemetry.io/docs/specs/semconv/registry/entities/service/) it states "The version string of the service API or implementation." hence service is talking about the api.

> A process executable belongs to an app. (does it? can't an exec just be a service? why imply an app?)

You would already have process being related to the process executable. The process is what is related to the service and in facts creates instances of the service.

> An app has to be installed. (executables and services are installed, why is an app installed?)

On a technical level yes executables and services are installed on to a machine as part of an installation but that is done as part of the installation process for an app/program.

> Each app consists of 1 or more executables which when run creates a process (I'm unaware of any other art here that describes an app as consisting of 1 or more executables).

App is a synnom for a program and if you look at your computer when you install an app/program what will happen is that 1 or more executables are installed.

I think google has a good definition showing this https://cloud.google.com/app-hub/docs/overview hence each micro-service is an executable.

I no where state or believe that service shouldn't be used in fact it is the opposite it should still be used in all cases and can be complemented/supplemented by app. And in fact it is about accurately capturing the details of what app the end user is using.

I do think the description of the entities could be improved.

trask:
@thompson-tomo it sounds like this doesn't have the support of the client instrumentation / browser / android SIGs

closing, but feel free to discuss with them

PR #2431, opened 2025-06-25
Improve otel.sdk.span.ended metric

JonasKunz:
TBD

## Changes

### DIFF
--- a/docs/otel/sdk-metrics.md
+++ b/docs/otel/sdk-metrics.md
@@ -95,7 +95,18 @@ For spans with `recording=false`: If implementations decide to record this metri
 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
+| [`otel.span.parent.origin`](/docs/registry/attributes/otel.md) | string | Determines whether the span has a parent span, and if so, whether it is a remote parent | `NONE`; `LOCAL`; `REMOTE` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`otel.span.sampling_result`](/docs/registry/attributes/otel.md) | string | The result value of the sampler for this span | `DROP`; `RECORD_ONLY`; `RECORD_AND_SAMPLE` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+---
+
+`otel.span.parent.origin` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `LOCAL` | The span has a parent and the parent's span context isRemote() is false | ![Development](https://img.shields.io/badge/-development-blue) |
+| `NONE` | The span does not have a parent, it is a root span | ![Development](https://img.shields.io/badge/-development-blue) |
+| `REMOTE` | The span has a parent and the parent's span context isRemote() is true | ![Development](https://img.shields.io/badge/-development-blue) |
 
 ---
 
--- a/model/otel/metrics.yaml
+++ b/model/otel/metrics.yaml
@@ -24,6 +24,7 @@ groups:
       For spans with `recording=false`: If implementations decide to record this metric, they MUST also record `otel.sdk.span.live`.
     attributes:
       - ref: otel.span.sampling_result
+      - ref: otel.span.parent.origin
 
   - id: metric.otel.sdk.processor.span.queue.size
     type: metric
--- a/model/otel/registry.yaml
+++ b/model/otel/registry.yaml
@@ -39,6 +39,23 @@ groups:
               stability: development
         brief: "The result value of the sampler for this span"
         stability: development
+      - id: otel.span.parent.origin
+        type:
+          members:
+            - id: none
+              value: NONE
+              brief: 'The span does not have a parent, it is a root span'
+              stability: development
+            - id: local
+              value: LOCAL
+              brief: The span has a parent and the parent's span context isRemote() is false
+              stability: development
+            - id: remote
+              value: REMOTE
+              brief: The span has a parent and the parent's span context isRemote() is true
+              stability: development
+        brief: "Determines whether the span has a parent span, and if so, whether it is a remote parent"
+        stability: development
   - id: registry.otel.scope
     type: attribute_group
     display_name: OTel Scope Attributes
--- a/docs/registry/attributes/otel.md
+++ b/docs/registry/attributes/otel.md
@@ -14,9 +14,20 @@ Attributes reserved for OpenTelemetry
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
+| <a id="otel-span-parent-origin" href="#otel-span-parent-origin">`otel.span.parent.origin`</a> | string | Determines whether the span has a parent span, and if so, whether it is a remote parent | `NONE`; `LOCAL`; `REMOTE` | ![Development](https://img.shields.io/badge/-development-blue) |
 | <a id="otel-span-sampling-result" href="#otel-span-sampling-result">`otel.span.sampling_result`</a> | string | The result value of the sampler for this span | `DROP`; `RECORD_ONLY`; `RECORD_AND_SAMPLE` | ![Development](https://img.shields.io/badge/-development-blue) |
 | <a id="otel-status-code" href="#otel-status-code">`otel.status_code`</a> | string | Name of the code, either "OK" or "ERROR". MUST NOT be set if the status code is UNSET. | `OK`; `ERROR` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 | <a id="otel-status-description" href="#otel-status-description">`otel.status_description`</a> | string | Description of the Status if it has a value, otherwise not set. | `resource not found` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+
+---
+
+`otel.span.parent.origin` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `LOCAL` | The span has a parent and the parent's span context isRemote() is false | ![Development](https://img.shields.io/badge/-development-blue) |
+| `NONE` | The span does not have a parent, it is a root span | ![Development](https://img.shields.io/badge/-development-blue) |
+| `REMOTE` | The span has a parent and the parent's span context isRemote() is true | ![Development](https://img.shields.io/badge/-development-blue) |
 
 ---
 


dashpole:
>  This means that otel.sdk.span.ended can simply be recorded from the tracer starting the span and the span implementation can remain stateless because it doesn't have to track the end() call.

Do you have a java example/prototype I can see?

model/otel/metrics.yaml
```diff
@@ -8,7 +8,8 @@ groups:
     unit: "{span}"
     note: |
       For spans with `recording=true`: Implementations MUST record both `otel.sdk.span.live` and `otel.sdk.span.ended`.
-      For spans with `recording=false`: If implementations decide to record this metric, they MUST also record `otel.sdk.span.ended`.
+      For spans with `recording=false`: Implementations MAY omit recording `otel.sdk.span.live` (e.g. for performance or simplicity reasons).
```
dashpole:
```suggestion
      Spans with `recording=false` are not considered live for purposes of calculating `otel.sdk.span.live`.
```

dashpole: I would prefer having a consistent meaning of live, instead of making it optional with a MAY.  I would also consider updating the description to say "The number of created, **recording** spans for which the end..."

model/otel/metrics.yaml
```diff
@@ -21,9 +22,12 @@ groups:
     unit: "{span}"
     note: |
       For spans with `recording=true`: Implementations MUST record both `otel.sdk.span.live` and `otel.sdk.span.ended`.
-      For spans with `recording=false`: If implementations decide to record this metric, they MUST also record `otel.sdk.span.live`.
+      For spans with `recording=false`: If implementations decide to not record `otel.sdk.span.live`, they MUST record `otel.sdk.span.ended` on span start for those spans instead.
+      In other words, for spans with `recording=false` implementations may treat them as ended directly after they are started from a metrics perspective.
+      This ensures that `otel.sdk.span.ended` is guaranteed to provide insights about the total span counts and sampling rates, even if spans are not recorded.
```
dashpole:
```suggestion
      Spans with `recording=false` are considered ended immediately after Start to ensure SDKs do not need to track state for non-recording spans.
```
I would love to simplify the wording here, but you don't need to take this verbatim.

docs/otel/sdk-metrics.md
```diff
@@ -91,14 +92,27 @@ This metric is [recommended][MetricRecommended].
...
-For spans with `recording=false`: If implementations decide to record this metric, they MUST also record `otel.sdk.span.live`.
+For spans with `recording=false`: If implementations decide to not record `otel.sdk.span.live`, they MUST record `otel.sdk.span.ended` on span start for those spans instead.
+In other words, for spans with `recording=false` implementations may treat them as ended directly after they are started from a metrics perspective.
+This ensures that `otel.sdk.span.ended` is guaranteed to provide insights about the total span counts and sampling rates, even if spans are not recorded.
 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
+| [`otel.span.parent.origin`](/docs/registry/attributes/otel.md) | string | Determines whether the span has a parent span, and if so, whether it is a remote parent | `NONE`; `LOCAL`; `REMOTE` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
AlexanderWert:
```suggestion
| [`otel.span.parent.origin`](/docs/registry/attributes/otel.md) | string | Determines whether the span has a parent span, and if so, [whether it is a remote parent](https://opentelemetry.io/docs/specs/otel/trace/api/#isremote) | `NONE`; `LOCAL`; `REMOTE` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```

docs/otel/sdk-metrics.md
```diff
@@ -91,14 +92,27 @@ This metric is [recommended][MetricRecommended].
...
 | [`otel.span.sampling_result`](/docs/registry/attributes/otel.md) | string | The result value of the sampler for this span | `DROP`; `RECORD_ONLY`; `RECORD_AND_SAMPLE` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 
 ---
 
+`otel.span.parent.origin` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `LOCAL` | The span has a parent and the parent's span context isRemote() is false | ![Development](https://img.shields.io/badge/-development-blue) |
```
AlexanderWert:
```suggestion
| `LOCAL` | The span has a parent and the parent's span context [isRemote()](https://opentelemetry.io/docs/specs/otel/trace/api/#isremote) is false | ![Development](https://img.shields.io/badge/-development-blue) |
```

docs/otel/sdk-metrics.md
```diff
@@ -91,14 +92,27 @@ This metric is [recommended][MetricRecommended].
...
+| `NONE` | The span does not have a parent, it is a root span | ![Development](https://img.shields.io/badge/-development-blue) |
+| `REMOTE` | The span has a parent and the parent's span context isRemote() is true | ![Development](https://img.shields.io/badge/-development-blue) |
```
AlexanderWert:
```suggestion
| `REMOTE` | The span has a parent and the parent's span context [isRemote()](https://opentelemetry.io/docs/specs/otel/trace/api/#isremote) is true | ![Development](https://img.shields.io/badge/-development-blue) |
```

AlexanderWert (approved):

JonasKunz:
>> This means that otel.sdk.span.ended can simply be recorded from the tracer starting the span and the span implementation can remain stateless because it doesn't have to track the end() call.
>
> Do you have a java example/prototype I can see?

@dashpole [here](https://github.com/open-telemetry/opentelemetry-java/pull/7430) is the PR for the java implementation. However, there we in fact do record `spans.ended` in the `end()` methods for non-recording spans.

I also just noticed that it in fact isn't really possible to return a noop-singleton span implementation even for non-recording spans: Even those spans need to correctly return their `SpanContext`, so it is not possible to use a shared singleton.

This and [your comment](https://github.com/open-telemetry/semantic-conventions/pull/2431#discussion_r2172087791) got me thinking: Why not just ditch all the `MAY`s and say that implementations must simply always collect `spans.ended` and `spans.live`, even for non-recording spans? I don't think that this imposes a measurable performance overhead and also shouldn't be difficult to implement due to what I said above. 

WDYT?

model/otel/metrics.yaml
JonasKunz: We decided to omit recording a `spans.started` metric because it can be computed via `spans.ended+spans.live`. Ditching the `MAY` here means that this is no longer possible, at least for non-recording spans.

Please see [my other comment](https://github.com/open-telemetry/semantic-conventions/pull/2431#issuecomment-3018119566) for a proposed solution here

dashpole: `spans.ended+spans.live` is still the number of started spans, though, since non-recording spans are considered ended.

dashpole:
> Why not just  all the MAYs and say that implementations must simply always collect spans.ended and spans.live, even for non-recording spans? I don't think that this imposes a measurable performance overhead and also shouldn't be difficult to implement due to what I said above.

I believe that is explicitly prohibited by the spec:  https://github.com/open-telemetry/opentelemetry-specification/blob/445a6835d755c1913a9acaa227997952ac9b4ea2/specification/trace/api.md?plain=1#L734.  I don't think your implementation is permitted by the spec.

> The remaining functionality of `Span` MUST be defined as no-op operations.
Note: This includes `End`, so as an exception from the general rule,
it is not required (or even helpful) to end such a Span.

JonasKunz:
Oh wow, great find! We totally overlooked this when initially defining the metrics.

I'd then suggest to do the following:
 * Remove `spans.ended` and add `spans.started` instead
    * `spans.started` MUST be recorded for all spans
 * Change `spans.live` metric that it MUST NOT be recorded for non-recording spans
 
 This way the metrics are simple to define and don't have that wierdness that `spans.ended` also counts actually non-ended non-recording spans.
 
 The number of ended spans can still be derived via `spans.started - spans.live`, which then will count non-recording spans as ended immediately.
 
 EDIT:
 
 After reading the spec again, I don't think my implementation violates the spec. The spec you linked is about `Wrapping a SpanContext in a Span`. This is handled by [PropagatedSpan](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/PropagatedSpan.java) in java.
 In contrast to that, the spec states that [spans created by tracers MUST be ended](https://github.com/open-telemetry/opentelemetry-specification/blob/445a6835d755c1913a9acaa227997952ac9b4ea2/specification/trace/api.md?plain=1#L425), independent of whether they are recording or not. So it kind of gives a use-case to track the `spans.live` even for non-recording spans: A growing `spans.live` value identifies a problem in the instrumentation, even when not recording.
 
 So I think there are two paths forward:
  * (a) We say that `span.live` is required even for non-recording spans. In that case we can safely keep `spans.ended` around, because it doesn't become "weird" for non-recording spans
  * (b) We don't report `span.live` for non-recording spans. In that case we should replace `span.ended` with `span.started` to avoid the "weirdness" for non-recording spans
  
  I would prefer (a), but I don't have a strong opinion. At least in Java it's easy to implement as shown by my PR and gives the data one less edge case. Also I think the performance overhead of the `UpDownCounter` update is acceptable for non-recording spans.

dashpole:
I think that might work. The one thing i'm not sure of is whether we needed spans.ended to help make sense of the processor metrics.  It came from https://github.com/open-telemetry/semantic-conventions/pull/1631#discussion_r1913585907, which was a suggestion by @jmacd.

@jmacd do you have any concerns with measuring spans.started instead of spans.ended?

dashpole:
I agree there is a use-case to track live spans.  The question is whether SDKs can/should implement it.  It looks like SDKs MAY use a NonRecordingSpan described in the API when a decision not to record is made: https://github.com/open-telemetry/opentelemetry-specification/blob/445a6835d755c1913a9acaa227997952ac9b4ea2/specification/trace/sdk.md#sdk-span-creation

> A non-recording span MAY be implemented using the same mechanism as when a Span is created without an SDK installed or as described in [wrapping a SpanContext in a Span](https://github.com/open-telemetry/opentelemetry-specification/blob/445a6835d755c1913a9acaa227997952ac9b4ea2/specification/trace/api.md#wrapping-a-spancontext-in-a-span).

If they do that, they won't be able to implement the live metric.  There is a bit of gray area as well.  Go doesn't explicitly use the NonRecordingSpan from the API, since the type is kept private, but instead defines a different private nonRecordingSpan in the SDK.  I would assume the non-recording spec applies to the copy we have in our SDK, but you could make the case it shouldn't apply.

In general, i'm still concerned with changing a call that is expected to be on the hot path, and is currently a no-op (ending a non-recording span), and changing it to a call to the metrics API (including constructing attributes, etc).  I prefer (b)

JonasKunz:
> I think that might work. The one thing i'm not sure of is whether we needed spans.ended to help make sense of the processor metrics. It came from https://github.com/open-telemetry/semantic-conventions/pull/1631#discussion_r1913585907, which was a suggestion by @jmacd.

We wanted the ability to have both the number of started and ended spans, but decided to collect just one of them because they can be derived from each other with the help of `spans.live`:

E.g. `increase(spans_started{sampling_result="SAMPLED"}[5m]) - delta(spans_live{sampling_result="SAMPLED"}[5m]) ` will give you the number of ended, sampled spans in the last five minutes.

> In general, i'm still concerned with changing a call that is expected to be on the hot path, and is currently a no-op (ending a non-recording span), and changing it to a call to the metrics API (including constructing attributes, etc). I prefer (b)

Then let's just go with (b). It's easier to add spans.live for non-recording spans later than removing it from the spec if we ever feel the need for it.

model/otel/registry.yaml
```diff
@@ -39,6 +39,23 @@ groups:
               stability: development
         brief: "The result value of the sampler for this span"
         stability: development
+      - id: otel.span.parent.origin
+        type:
+          members:
+            - id: none
+              value: NONE
```
lmolkova: does it have to be in SCREAMING_CASE?


```suggestion
              value: none
```

also is it necessary to set this attribute for root spans at all?

model/otel/registry.yaml
```diff
@@ -39,6 +39,23 @@ groups:
...
+              brief: 'The span does not have a parent, it is a root span'
+              stability: development
+            - id: local
+              value: LOCAL
```
lmolkova:
```suggestion
              value: local
```

model/otel/registry.yaml
```diff
@@ -39,6 +39,23 @@ groups:
...
+              brief: The span has a parent and the parent's span context [isRemote()](https://opentelemetry.io/docs/specs/otel/trace/api/#isremote) is false
+              stability: development
+            - id: remote
+              value: REMOTE
```
lmolkova:
```suggestion
              value: remote
```

model/otel/registry.yaml
JonasKunz: > also is it necessary to set this attribute for root spans at all?

Yes IMO it's important to be explicit instead of inferring information from the absence of an attribute.
In this concrete example, let's imagine we decide to later add `otel.span.parent.origin` to `spans.live`.
It will then be impossible for users to distinguish if they are receiving "old" `spans.live` data without the attribute or if all their spans are root-spans. This problem does not occur with the explicit `none` value. Also this comes at close to no cost, as it doesn't increase the cardinality to have `none` vs the absence of the attribute.

PR #2444, opened 2025-06-27
Add metric value type everywhere (as code-generation annotation)

### LINKED ISSUE #591: Metrics semconv should include value type of instrument in yaml
Metric-specific properties (in yaml) [are limited](https://github.com/open-telemetry/build-tools/blob/259a02fb341d9fae89f8fdaf9c55fdb8e92c4428/semantic-conventions/syntax.md?plain=1#L135) to `metric_name`,  `instrument`, and `unit`:

We never document the [value type](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#timeseries-model) (long or float) in yaml or markdown files.

As a result code generation cannot really generate metric definitions and we leave a wiggle room for instrumentation to decide on (and change) the value type.

Also, we manually include requirement level into md.

We should 
- extend metric yaml schema to include
  - value type (and make it a part of [stability guarantees](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#semantic-conventions-stability))
  --requirement level-- moved to #796
- update existing metric definitions in the spec

lmolkova:
Fixes #591 
This PR adds `value_type` (int or double) to all metric definitions and updates template to render it in the metrics table.

All MD changes are auto-generated, all substantial changes are in yaml


model/system/metrics.yaml
```diff
@@ -29,6 +31,7 @@ groups:
   - id: metric.system.cpu.logical.count
     type: metric
     metric_name: system.cpu.logical.count
+    value_type: int
```
lmolkova: /cc @open-telemetry/semconv-system-approvers is this right?

model/k8s/metrics.yaml
```diff
@@ -27,6 +29,7 @@ groups:
   - id: metric.k8s.pod.cpu.usage
     type: metric
     metric_name: k8s.pod.cpu.usage
+    value_type: double
```
lmolkova: @open-telemetry/semconv-k8s-approvers is this right?

model/k8s/metrics.yaml
```diff
@@ -122,6 +131,7 @@ groups:
   - id: metric.k8s.node.allocatable.cpu
     type: metric
     metric_name: k8s.node.allocatable.cpu
+    value_type: double
```
lmolkova: @open-telemetry/semconv-k8s-approvers is this and other cpu-related changes in this file correct?

model/aspnetcore/metrics.yaml
```diff
@@ -63,6 +66,7 @@ groups:
   - id: metric.aspnetcore.rate_limiting.request_lease.duration
     type: metric
     metric_name: aspnetcore.rate_limiting.request_lease.duration
+    value_type: int
```
trask:
```suggestion
    value_type: double
```

trask (approved):
I also have similar question as you about when cpu count should allow for millicores and when it shouldn't

E.g. `jvm.cpu.count` is captured using https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#availableProcessors-- and so can only be an `int`, but there could always be a new Java API in the future that is container aware and allows capturing in millicores (e.g. https://openjdk.org/jeps/8182070)

although I think it's not a breaking change to switch between int and double, since there's no difference in the OTLP output, so future proofing may not be so important?

model/system/metrics.yaml
braydonk: To my knowledge this is correct. Semi-related to @trask's comment below, but none of the raw system APIs we would use to instrument this metric will produce anything but an integer count of CPU cores. This is different for container environments like k8s though (and I see they are correctly marked as `double` within this PR).

model/aspnetcore/metrics.yaml
lmolkova: thanks!

kamphaus (approved):

templates/registry/markdown/metric_table.j2
```diff
@@ -2,7 +2,7 @@
 {% import 'notes.j2' as notes %}
 {% import 'metric_macros.j2' as metrics %}
 {% macro print_entities(associations) %}{%- for e in associations %}{%if loop.first == false %}; {% endif %}`{{ e | trim }}`{%- endfor %}{% endmacro %}
-{% macro generate(group) %}| Name     | Instrument Type | Unit (UCUM) | Description    | Stability | Entity Associations |
+{% macro generate(group) %}| Name     | Instrument Type | Value Type | Unit (UCUM) | Description    | Stability | Entity Associations |
 | -------- | --------------- | ----------- | -------------- | --------- | ------ |
```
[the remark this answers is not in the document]
joaopgrassi: Good bot!

xrmx (approved):

lmolkova:
based on the SIG call discussion:
- metric value type is explicitly called out as something that can change - https://github.com/open-telemetry/opentelemetry-specification/blob/ac137b349b006dc926da91c6a6619217737921cc/specification/versioning-and-stability.md?plain=1#L260
- we can still have a hint defined for codegen reasons - this is what `annotations` are for


so, I repurposed this PR to use `annotations.code_generation.metric_value_type`

E.g.

```yaml
  - id: metric.system.process.count
    type: metric
    metric_name: system.process.count
    annotations:
      code_generation:
        metric_value_type: int
```

model/system/metrics.yaml
lmolkova: thanks!

resolving this since we're not going to require types to match, it's just a hint for codegen and can be ignored. - https://github.com/open-telemetry/semantic-conventions/pull/2444#issuecomment-3029374240

model/http/metrics.yaml
```diff
@@ -156,15 +180,16 @@ groups:
         requirement_level: required
       - ref: server.port
         requirement_level: required
-        brief: >
```
trisch-me: is this (and others similar in this file) intentional change?

model/http/metrics.yaml
```diff
@@ -141,6 +162,9 @@ groups:
   - id: metric.http.client.open_connections
     type: metric
     metric_name: http.client.open_connections
+    annotations:
+      code_generation:
+        metric_value_type: int
```
antonfirsov, line 167: Does `int` cover all integer types? This uses `long` in the .NET implementation (not sure why we decided to do so).

lmolkova: all ints are signed int64 in OTel

model/http/metrics.yaml
lmolkova: oops, great catch!

jerbly:
See https://github.com/open-telemetry/weaver/issues/817#issuecomment-3039466462 - all is not lost. Since `annotations` are part of the resolved registry they are also passed through to custom rego policies you can provide to `live-check`. So, it is still possible to check telemetry aligns with model.

PR #2447, opened 2025-06-28
Add copilot instructions for auto-review

lmolkova:
Adds [instructions](https://docs.github.com/en/copilot/how-tos/custom-instructions/adding-repository-custom-instructions-for-github-copilot) for review/coding agent based on [contributing.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) and [general guidance](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/general).

Tried on a few PRs, it seems to produce some reasonable advice, but definitely can be tuned further.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,23 +1,33 @@
-# Copilot instructions for OpenTelemetry Semantic Conventions
+# Copilot Instructions for OpenTelemetry Semantic Conventions
 
 ## Overview
-These instructions guide AI assistants in reviewing OpenTelemetry Semantic Conventions
-pull requests to ensure consistency, quality, and adherence to established standards.
+These instructions guide AI assistants in reviewing pull requests for the
+OpenTelemetry Semantic Conventions repository. Follow these guidelines to
+ensure consistency, quality, and adherence to established standards.
 
-**CRITICAL REVIEW CRITERIA**: New areas require special scrutiny and must meet these requirements:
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,95 @@
...
+- **Template attributes**: Dynamic names (last segment only)
+
+## Event Conventions
+- MUST have unique, low-cardinality event name
+- SHOULD have attributes for context, SHOULD NOT use body
+
+## Error Handling Conventions
+- **Spans**: Set status to Error, populate `error.type`, set description when helpful
+- **Metrics**: Include `error.type` attribute for filtering and analysis
+- **Exceptions**: Record as span events or log records using SDK APIs
```
trask, line 96:
```suggestion
- **Exceptions**: Record as exception events
```

.github/copilot-instructions.md
```diff
@@ -0,0 +1,95 @@
...
+- **Consistency**: Same `error.type` across spans and metrics for same operation
+
+## Common Issues to Flag
```
trask, line 99: nice

trask (approved):
Good idea! Looking forward to seeing how it works / iterating on it.

.github/copilot-instructions.md
lmolkova: good catch! let's remove this line - conventions don't really include exception attributes anyway.

joaopgrassi (approved):
Let's see how good this is!

Issue #2455, opened 2025-06-30
SemConv 1.35.0 breaks codegen (enum members collisions between _ and .)


lmolkova:
⚠ Please DO NOT release semconv artifacts based on v1.35.0 ⚠ 

https://github.com/open-telemetry/semantic-conventions/pull/1698 renamed a bunch of `cloud.platform` members from `azure_` to `azure.`  which causes collisions similar to https://github.com/open-telemetry/semantic-conventions/issues/1118 when generating code - the constant/enum member name is exactly the same for `azure_aks` and `azure.aks`.

E.g. on JS

```
> tsc --build tsconfig.json tsconfig.esm.json tsconfig.esnext.json

src/experimental_attributes.ts:1448:14 - error TS2451: Cannot redeclare block-scoped variable 'CLOUD_PLATFORM_VALUE_AZURE_AKS'.

1448 export const CLOUD_PLATFORM_VALUE_AZURE_AKS = "azure.aks" as const;
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/experimental_attributes.ts:1483:14 - error TS2451: Cannot redeclare block-scoped variable 'CLOUD_PLATFORM_VALUE_AZURE_AKS'.

1483 export const CLOUD_PLATFORM_VALUE_AZURE_AKS = "azure_aks" as const;
```

We'll need to:

- [ ] Support `annotation` mechanism to exclude old values from codegen similar to attributes - it does not apply to enum members today - https://github.com/open-telemetry/weaver/pull/812
- [ ] Add policy preventing such collisions in future - https://github.com/open-telemetry/semantic-conventions/pull/2456
- [ ] Update semconv and annotate problematic enum members
- [ ] Release 1.35.1 patch

lmolkova:
/cc @open-telemetry/android-maintainers @open-telemetry/cpp-maintainers @open-telemetry/dotnet-maintainers @open-telemetry/erlang-maintainers @open-telemetry/go-maintainers @open-telemetry/java-maintainers @open-telemetry/javascript-maintainers @open-telemetry/php-maintainers @open-telemetry/python-maintainers @open-telemetry/ruby-maintainers @open-telemetry/rust-maintainers @open-telemetry/swift-maintainers 

⚠ Please DO NOT release SemConv artifacts based on v1.35.0 ⚠

PR #2456, opened 2025-07-01
Member collisions policies: id, value, const name

### LINKED ISSUE #2448: Duplicate members are not being identified
### Area(s)

schema, tooling, docs

### What's the problem?

Uniqueness of members is not being checked which is allowing duplicates.

An error is thrown when attempting to compile markdown hence scenario is identified prior to commiting code. Uniqueness should be limited to the attribute which they are used. 

### Additional context 
Observed when doing a manual review of https://github.com/open-telemetry/semantic-conventions/commit/8e19db8412095757c5ae920d907b6b6f2e20c4ef where clickhouse was mistakenly duplicated.

lmolkova:
Fixes #2448

Related to #2455

Blocked on https://github.com/open-telemetry/weaver/pull/812


policies/registry.rego
```diff
@@ -89,10 +89,82 @@ deny contains attr_registry_violation(description, group.id, "") if {
...
+    description := sprintf("Member with id '%s' is already defined on the attribute '%s' in the group '%s'. Member id must be unique.", [member.id, attr.id, group.id])
+}
+
+# check that member values do not collide within the same attribute
+deny contains attr_registry_violation(description, group.id, attr.id) if {
+    group := input.groups[_]
+    startswith(group.id, "registry.")
+    attr := group.attributes[_]
+    member := attr.type.members[_]
+    not is_property_set(member, "deprecated")
```
lmolkova, line 125: I can never understand the difference between `not obj.prop` vs `obj.prop == null` vs this hack. For whatever reason this works and `not obj.prop`,  `obj.prop == null` don't 😿

lmolkova:
replaced by #2477

Issue #2443, opened 2025-06-27
HTTP: The definition of `server.address` differs across various metrics


antonfirsov:
### Area(s)

HTTP

### What's the problem?

The HTTP client request metrics spec (e.g. for [`http.client.request.duration`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpclientrequestduration)) and the [HTTP client span](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-client-span) spec defines `server.address` as 

> Host identifier of the ["URI origin"](https://www.rfc-editor.org/rfc/rfc9110.html#name-uri-origin) HTTP request is sent to.

while in the spec of other metrics (e.g. [`http.client.connection.duration`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpclientconnectionduration)), and the attribute registry [for "Server"](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/server.md) it is defined as

> Server domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name.

This leads to a problem deciding whether the Host header should take precedence over the URI origin. The current wording seems to _slightly_ suggest that request span (and metrics) should use the URI origin while the connection metrics should use the Host header but this doesn't seem to be neither intentional nor clear.

Context: in https://github.com/dotnet/runtime/issues/117049 a user reported a problem with .NET's `HttpClient` metrics which currently report the URI origin instead of the Host header. While changing this to report the Host header instead seems better from practical POV, the expected behavior is not clear from the current state of the semconv spec.

### Describe the solution you'd like

- Uniformize the specs or make it clear in comments that the expected attribute values differ intentionally across various metrics/spans.
- In either case, it should be clear if the Host header has precedence over the URI origin (for each metric).
- Personally, I would prefer the Host header whenever available since it adds the host information back when an IP address is being used in the URI. There is `network.peer.address` for IP addresses anyways.

lmolkova:
@antonfirsov is it the case when `HttpRequestMessage.RequestUri.Host` is different than the `Host` header?  I would imagine when it gets to sending physical request, the `HttpRequestMessage.RequestUri.Host` is actually ignored, is this right?

[...]

PR #2463, opened 2025-07-01
Replace confusing URI origin with more specific guidance

lmolkova:
Related to #2443

### DIFF
--- a/docs/http/http-metrics.md
+++ b/docs/http/http-metrics.md
@@ -559,11 +559,6 @@ the `server.address` SHOULD match the host component of the request target.
 In all other cases, `server.address` SHOULD match the host component of the
 `Host` header in HTTP/1.1 or the `:authority` pseudo-header in HTTP/2 and HTTP/3.
 
-If an HTTP client request is explicitly made to an IP address, e.g. `Host` header
-is set to `x.x.x.x:8080` and request target is not in the absolute-form,
-then `server.address` SHOULD be the IP address `x.x.x.x`.
-A DNS lookup SHOULD NOT be used.
-
 **[3] `server.port`:** In the case of HTTP/1.1, when the [request target](https://www.rfc-editor.org/rfc/rfc9112.html#name-request-target)
 is passed in its [absolute-form](https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2.2),
 the `server.port` SHOULD match the port component of the request target.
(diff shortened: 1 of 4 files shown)


docs/http/http-metrics.md
```diff
@@ -552,9 +552,22 @@ HTTP method names are case-sensitive and `http.request.method` attribute value M
...
-**[2] `server.address`:** If an HTTP client request is explicitly made to an IP address, e.g. `http://x.x.x.x:8080`, then `server.address` SHOULD be the IP address `x.x.x.x`. A DNS lookup SHOULD NOT be used.
+**[2] `server.address`:** In HTTP/1.1, when the [request target](https://www.rfc-editor.org/rfc/rfc9112.html#name-request-target)
+is passed in its [absolute-form](https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2.2),
+the `server.address` SHOULD match the host component of the request target.
 
-**[3] `server.port`:** When observed from the client side, and when communicating through an intermediary, `server.port` SHOULD represent the server port behind any intermediaries, for example proxies, if it's available.
+In all other cases, `server.address` SHOULD match the host component of the
+`Host` header in HTTP/1.1 or the `:authority` pseudo-header in HTTP/2 and HTTP/3.
+
+If an HTTP client request is explicitly made to an IP address, e.g. `http://x.x.x.x:8080`, then
```
antonfirsov: This wording goes against our goal with https://github.com/open-telemetry/semantic-conventions/issues/2443 and https://github.com/dotnet/runtime/issues/117049.

If a request goes to `http://x.x.x.x`, but the `Host` header is set, `x.x.x.x` will be only used to make the connection at transport level. At the HTTP level the IP information does not appear anymore, and `Host` is the _Server domain name_ users want to see in `server.address`.

joaopgrassi (approved):

docs/http/http-metrics.md
```diff
@@ -552,9 +552,24 @@ HTTP method names are case-sensitive and `http.request.method` attribute value M
...
+is passed in its [absolute-form](https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2.2),
+the `server.address` SHOULD match the host component of the request target.
 
-**[3] `server.port`:** When observed from the client side, and when communicating through an intermediary, `server.port` SHOULD represent the server port behind any intermediaries, for example proxies, if it's available.
+In all other cases, `server.address` SHOULD match the host component of the
+`Host` header in HTTP/1.1 or the `:authority` pseudo-header in HTTP/2 and HTTP/3.
+
+If an HTTP client request is explicitly made to an IP address, e.g. `Host` header
+is set to `x.x.x.x:8080` and request target is not in the absolute-form,
+then `server.address` SHOULD be the IP address `x.x.x.x`.
```
antonfirsov:
```suggestion
If the the host component of the `Host` header or the `:authority` pseudo-header is set to an IP address `x.x.x.x` and request target is not in the absolute-form,
then `server.address` SHOULD be the IP address `x.x.x.x`.
```

> If an HTTP client request is explicitly made to an IP address

I find this part confusing. Sending an HTTP request to `http(s)://x.x.x.x[...]` means that we are _explicitly asking_ the HTTP client to connect to the `x.x.x.x` address at the transport level which can be interpreted that the "request is explicitly made to an IP address". (While knowing the hostname associated with `x.x.x.x`, and specifying it in the Host header).

trask: If this sentence is confusing, should we just remove it? I think it doesn't add anything that isn't specified above?

trask (approved):

lmolkova:
@antonfirsov could you please take another look?

Issue #2382, opened 2025-06-17
Improve navigational structure


thompson-tomo:
Having spent a fair amount of time using the otel semantic convention documentation recently, I have a couple of ideas which I feel could help people, learn about otel and increase automation potential.

## Navigational structure of website
Currently when you go to https://opentelemetry.io/docs/specs/semconv/ within the navigation structure you are presented both background information ie General and samples ie ci cd with no clear seperation.

### idea
Move the documentation for events, metrics & spans to be in corresponding registries that way if you expand the metric registries you can see everything that has defined metrics. This would make the registries your key repository of samples and what is available. We could also offer a namespace registry with documents similar to https://opentelemetry.io/docs/specs/semconv/cicd/ with the key thing being the linking to entries in other registries ie metrics.

PR #2464, opened 2025-07-02
[chore] Create an event registry #2382

thompson-tomo:
Progresses #2382

## Changes

This introduces an event registry to reside alongside the attributes registry and the soon to be released entities registry. This provides pure code generated documentation for events which reduces maintaince effort and makes things more discoverable. 


templates/registry/markdown/event_namespace.md.j2
```diff
@@ -0,0 +1,36 @@
...
+<!-- markdownlint-disable -->
+{%- import 'stability.j2' as stability %}
+{%- import 'requirement.j2' as requirement %}
+{%- import 'attribute_table.j2' as at %}
+{%- import 'body_field_table.j2' as body_table %}
+{%- import 'event_macros.j2' as event %}
+
+# {{ ctx.namespace | title_case | acronym }}
+
+{%- if ctx.entities | length > 0 %}
```
jsuereth: Why are entities in the event namespace?

thompson-tomo: I was hoping to get this index page to show all the signal types grouped by type but was stuck on the filter hence my other comment (https://github.com/open-telemetry/semantic-conventions/pull/2464#discussion_r2202453950). That way a user could go to the rpc index page and see all attributes, events, log records, metrics & spans.

docs/registry/events/README.md
```diff
@@ -0,0 +1,112 @@
...
+> The following registry overview is a work in progress.
+
+Currently, the following namespaces exist:
+<table>
+<thead>
+<tr>
+  <th>Namespace</th>
+  <th>Event</th>
+  <th>Stability</th>
+</tr>
```
thompson-tomo, line 23: Would be great to see this as a table with a count of items per signal type which is linked to the section especially if we combine based on namespace.

jsuereth:
A few thoughts on this:

- Can you comment on why each event is its own markdown file? I think this is likely the right way forward but want to be explicit as we move forward on signals.  It is not what we did for Entity.
- Can you add a link to the Event registry from the `registry/README.md` file?
- Can you modify this to pull in every `event` without using `startswith`.  I think you had at one time split logrecord vs. event.  We do not want that distinction going forward, so this should just grab all `group.type=event`.

thompson-tomo:
Thanks for the feedback @jsuereth 

> Can you comment on why each event is its own markdown file? I think this is likely the right way forward but want to be explicit as we move forward on signals. It is not what we did for Entity.

Reason for 1 page per signal definition is so that we could have a namespace index page which would hopefully list everything in that namespace be it events, log records, metrics, spans etc. If we have 1 page it quickly grows out of hand and this approach is inline with the proposal to update the metric layout.

> Can you add a link to the Event registry from the registry/README.md file?

Yes can do

> Can you modify this to pull in every event without using startswith. I think you had at one time split logrecord vs. event. We do not want that distinction going forward, so this should just grab all group.type=event.

If possible I would get rid of the starts with altogether and just do filtering on the last mapping and is what is required for the general namespace page.

thompson-tomo:
@jsuereth Following on from the semconv meeting, I will do the following:
- Raise a seperate pr to introduce the namespace index page as seen here so it can be used for entities.
- Add entities to the namespace index page & dedicated pages
- Allow this PR to focus on adding events to the registry
- Work through existing event pages inspecting what has been migrated

docs/registry/namespaces/gen-ai/event-gen-ai-assistant-message.md
```diff
@@ -0,0 +1,102 @@
+<!-- NOTE: THIS FILE IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/entity_namespace.md.j2 -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+# Event Gen AI Assistant Message
+
+## Gen AI Assistant Message
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
```
lmolkova, line 10: this (and all other gen_ai) event is not really usable without the rest of the GenAI convention it follows at least today  

To make the usable:

1. we need to express event requirement level in the yaml. 
2. find a way to link from here to the actual convention which consist of a group of events, spans and metrics.
3. Find ways to bring these notes https://github.com/thompson-tomo/semantic-conventions/blob/42a320708d1a5bbdaad4cb955765b8f9ef926d24/docs/gen-ai/gen-ai-events.md?plain=1#L20-L39 - without them someone who discovers this even without the convention might not know how to migrate.

When it comes to GenAI events, I don't see how event registry is helpful. It breaks coherent document in multiple pieces without solving any problem.

thompson-tomo: > we need to express event requirement level in the yaml.

There is already a weaver issue for this https://github.com/open-telemetry/weaver/issues/406

> find a way to link from here to the actual convention which consist of a group of events, spans and metrics.

When #2518 is merged in the template used can add a link to the namespace page (currently commented out). Hence solving this. Also Grouping ties into https://github.com/open-telemetry/weaver/issues/406

> Find ways to bring these notes https://github.com/thompson-tomo/semantic-conventions/blob/42a320708d1a5bbdaad4cb955765b8f9ef926d24/docs/gen-ai/gen-ai-events.md?plain=1#L20-L39 - without them someone who discovers this even without the convention might not know how to migrate.

Agree and was already captured as https://github.com/open-telemetry/weaver/issues/802

lmolkova: can we talk about the benefits we get from event registry in MD and moving event definitions away from the corresponding conventions?

thompson-tomo: > benefits we get from event registry in MD

I no longer see it as an event registry, originally I did when first raising PR. It is now adding events to the namespace registry alongside the other signals. The benefits would be things like readability, discoverability & consistency.

> moving event definitions away from the corresponding conventions

This I don't get, when #2518 is merged you will have a single gen-ai page which is automatically generated which contains tables describing all of the events & entities which are defined. This will grow to include metrics, spans & attributes once their definitions can be fully defined in yaml and tooling can handle it.

What I do agree with is the way in which implementations are defined, can and should be improved. I will see if I can come with an interim solution for gen-ai.

lmolkova: The notion of namespace is not clear and I'd recommend prototyping and socializing the end result before working on any new proposals. I don't believe it's productive to work on it without getting feedback and buy-in from @open-telemetry/specs-semconv-approvers

thompson-tomo: i was actioning what was discussed in the last semconv meeting where it was socialized the end result of having a page showcasing the events available and being able to select one to see more details. It was explicitly mentioned that this index page would grow to include the other signals.

PR #2465, opened 2025-07-02
[chore] Create a span registry #2382

thompson-tomo:
Progresses #2382 

## Changes

This introduces a span registry to reside alongside the attributes registry and the soon to be released entities registry. This provides pure code generated documentation for events which reduces maintenance effort and makes things more discoverable.


jsuereth:
Until we have a unique way to reference spans, we're not ready for a registry.

@lmolkova and the tooling SIG have been working on a proposal where we will have a distinguishing identifier for `Span` that we could use for a registry.  I think we should keep this PR in draft until that work lands.  Anything we do now may increase friction in how we organize and report Spans.

thompson-tomo:
Agree but I also suspect we will also need a way to distinguish between a span definition and a more descriptive definition of that span based upon an instrument using that definition.

Yes both should have auto-generated docs, but they should be organised differently.

PR #2470, opened 2025-07-03
[chore] Add in Attribute Bundle pages #2382

thompson-tomo:
Progresses #2382

## Changes

Adds in a log record registry to enable easy discovery of log record definitions


thompson-tomo:
Currently blocked by #2471 after which build should be resolved.

jsuereth:
Commented on the Event registry - We don't want logs vs. records split in semconv.  Let's drop this PR and update the Event PR to include all event group types.

thompson-tomo:
As per the filter these are looking at attribute groups rather than events.

What we do need is a way to capture a group/bundle of attributes which can be attached to log records, events or spans. 

If we take the example of https://opentelemetry.io/docs/specs/semconv/cloudevents/cloudevents-spans/ that documentation is not a span but rather a grouping of attributes which could be added to a signal.

Let me rework this tomorrow.

Issue #199, opened 2023-07-20
Idiomatic way to implement a desktop app heartbeat


riverar:
Previously asked at https://github.com/open-telemetry/community/discussions/1598

Is there a documented pattern or generally accepted method of implementing a _heartbeat_ in a desktop application? That is, a periodic message send to a collector for the purpose of determining an active session count? Or perhaps a general pattern of reporting back Boolean states?

I was going down the path of using a Counter with a 0|1 value and just letting that contact a Collector periodically, but wanted to double-check.

My notes thus far:
* Metrics reader can be configured non-cumulative for fire once
   https://github.com/open-telemetry/opentelemetry-dotnet/issues/2524

* On-going work related to establishing conventions for service uptime
   https://github.com/open-telemetry/oteps/pull/185

Prior art:
* Microsoft App Insights Live Metrics https://azure.microsoft.com/en-us/blog/live-metrics-stream/

jsuereth:
We're looking at kicking off a working group around self-observability of our SDKs/APIs.  I think a general health check / heartbeat recommendations should go along with that.

WDYT @carlosalberto @yurishkuro ?   cc @reyang

yurishkuro:
Not seeing a strong motivation for SDK heartbeat. If SDK produces a metric indicating how much data it exports, that is equivalent to a heartbeat.

On the other hand, this issue seems to be about the application's heartbeat, not the SDKs. Either way, the motivation is not clear to me. I am assuming this is all independent of a health check endpoint an application may provide.

jsuereth:
My question was whether, as part of SDK self-o11y health, we provide general health metric support.

joaopgrassi:
This was closed by [mistake by the stale bot.](https://github.com/open-telemetry/semantic-conventions/pull/771) Re-opening

decko:
There is some place where the discussion about this feature is happening right now?

joaopgrassi:
@decko https://github.com/open-telemetry/semantic-conventions/pull/2472/files here :)

PR #2472, opened 2025-07-03
[chore] Non-normative guidance for status metrics

### LINKED ISSUE #1554: Create guideline for modeling state and phase as metrics

braydonk:
Fixes #1554 

## Changes

This PR adds non-normative guidance for designing status metrics. This is a common metric pattern that has confused contributors in the past, so this document exists to clear up some misconceptions and outline a unified design plan.

### DIFF
--- a/docs/non-normative/status-metrics.md
+++ b/docs/non-normative/status-metrics.md
@@ -1,0 +1,89 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Status Metrics
+aliases: [status-metrics]
+--->
+
+# State Metrics
+
+**Status**: [Development][DocumentStatus]
+
+<!-- toc -->
+
+<!-- tocstop -->
+
+## Definition
+
+For the purposes of this document, a "status metric" is a metric used to
+represent something being in a particular state at a given time from a closed
+set of distinct possible state values.
+
+## Design
+
+To represent a metric of this nature, the design will look like the following:
+
+```yaml
+id: metric.example.status
+type: metric
+metric_name: example.status
+stability: development
+brief: "The current state."
+note: |
+  A timeseries is produced for every possible value of `example.state`. The
+  value of this metric is 1 for a state if it is currently in said state, and
+  is 0 for all other states.
+instrument: updowncounter
+unit: "1"
+attributes:
+- ref: example.state
+```
+
+### Naming
+
+This pattern represents a unique naming challenge. You could reasonably call the
+metric and attribute both "status" or "state" as the two words are relatively
+interchangeable in common parlance.
+
+This problem necessitates choosing words to represent two things:
+
+* The noun representing one of the set of possible values (the attribute name)
+* The adjective describing which of the values is currently active (the metric
+  name)
+
+The general recommendation is to use the word "state" for the attribute and
+"status" for the metric. This is derived from common turns of phrase for each
+word, respectively:  
+"What **state** is X in?"  
+"What is the **current status** of X?"  
+
+The english-language semantics of this are heavily debatable, thus it is
+recommended for the sake of consistency that this pattern be adopted to avoid
+inconsistent naming across similarly designed metrics.
+
+The exception to this would be if a particular system has only one word that
+would be used for both the noun and the adjective in question. An example would
+be in Kubernetes, where the word "phase" equally represents both and there
+is no acceptable alternative. In this case, using a metric name suffix is
+recommended to avoid the naming clash, i.e. naming the attribute `k8s.phase` and
+the metric `k8s.phase.current`.
+
+### Instrument
+
+The metric is instrumented as an `UpDownCounter` rather than a `Gauge`. This is
+a deliberate choice, as it is a reasonable use case to count objects that are
+in a particular state. Since the metric value is either `0` or `1` for a given
+state attribute value, this means you can do a simple sum aggregation to count
+instances of particular states.
+
+### Why not Resource Attributes?
+
+It is a common mistake to attach a "state" as a Resource Attribute. This is
+not recommended for two reasons:
+
+* Resource is intended to be immutable, thus adding an attribute like "state"
+  that will almost certainly change during the resource's lifetime breaks the
+  immutability and will create what are essentially new timeseries each time the
+  attribute would update.
+* Making the "state" an attribute on Resource would make it impossible to track
+  changes in state over time.
+
+[DocumentStatus]: https://opentelemetry.io/docs/specs/otel/document-status


docs/non-normative/status-metrics.md
```diff
@@ -0,0 +1,95 @@
...
+The exception to this would be if a particular system has only one word that
+would be used for both the noun and the adjective in question. An example would
+be in Kubernetes, where the word "phase" equally represents both and there
+is no acceptable alternative. In this case, using a metric name suffix is
+recommended to avoid the naming clash, i.e. naming the attribute `k8s.phase` and
+the metric `k8s.phase.current`.
+
+### Instrument
+
+The metric is instrumented as an `UpDownCounter` rather than a `Gauge`. This is
```
joaopgrassi: We should probably link to the recommendation on how to record consistent UpDownCounters: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#consistent-updowncounter-timeseries

docs/non-normative/status-metrics.md
```diff
@@ -0,0 +1,95 @@
...
+The english-language semantics of this are heavily debatable, thus it is
+recommended for the sake of consistency that this pattern be adopted to avoid
+inconsistent naming across similarly designed metrics.
+
+The exception to this would be if a particular system has only one word that
+would be used for both the noun and the adjective in question. An example would
+be in Kubernetes, where the word "phase" equally represents both and there
+is no acceptable alternative. In this case, using a metric name suffix is
+recommended to avoid the naming clash, i.e. naming the attribute `k8s.phase` and
+the metric `k8s.phase.current`.
```
joaopgrassi: Looking, I see there's some metrics where the name + attributes are the same, thus already going against the recommendation here. For ex:

https://github.com/open-telemetry/semantic-conventions/blob/83b55440146a1978699bc79b1c88b8a73085478c/model/k8s/metrics.yaml#L79, https://github.com/open-telemetry/semantic-conventions/blob/83b55440146a1978699bc79b1c88b8a73085478c/model/k8s/metrics.yaml#L92

Could those be somehow adapted to not clash like it is proposed here? Would be unfortunate that we publish the guidance while we itself already do not follow it, unless we can give reasons.

joaopgrassi (commented):
Thank you for putting this together!

joaopgrassi:
This can also close this old issue #199. CC @riverar

docs/non-normative/status-metrics.md
ChrsMark: I agree! We have several metrics of this type in K8s already:

- https://github.com/open-telemetry/semantic-conventions/blob/83b55440146a1978699bc79b1c88b8a73085478c/model/k8s/metrics.yaml#L159
- https://github.com/open-telemetry/semantic-conventions/blob/83b55440146a1978699bc79b1c88b8a73085478c/model/k8s/metrics.yaml#L604
- https://github.com/open-telemetry/semantic-conventions/blob/83b55440146a1978699bc79b1c88b8a73085478c/model/k8s/metrics.yaml#L79
- https://github.com/open-telemetry/semantic-conventions/blob/83b55440146a1978699bc79b1c88b8a73085478c/model/k8s/metrics.yaml#L92

I wonder if we should not enforce the wording here but only the modeling instead. Maybe  using `.state` can just come as a generic suggestion rather than "stricter" guidance?

braydonk: I can loosen the guidance here. But isn't it pretty problematic to have an attribute that is the exact same name as a metric? They are un-clashed in tooling by their Weaver identifiers, but couldn't this cause problems or at least confusion for people interpreting the conventions?

docs/non-normative/status-metrics.md
```diff
@@ -0,0 +1,95 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Status Metrics
+aliases: [status-metrics]
+--->
+
+# State Metrics
```
lmolkova: wdyt about moving it to https://github.com/open-telemetry/semantic-conventions/tree/main/docs/non-normative/how-to-write-conventions ? 

I believe @jsuereth wanted to consolidate all the guidance there and maybe we'll move it altogether to a different place (some of the meta-guidance is or could be normative)

lmolkova: could you please add a link to it from https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/how-to-define-semantic-conventions.md#defining-metrics ?

docs/non-normative/status-metrics.md
```diff
@@ -0,0 +1,95 @@
...
+
+**Status**: [Development][DocumentStatus]
+
+<!-- toc -->
+
+- [Definition](#definition)
+- [Design](#design)
+  - [Naming](#naming)
+  - [Instrument](#instrument)
+  - [Why not Resource Attributes?](#why-not-resource-attributes)
```
lmolkova:
```suggestion
  - [Why not Entity Attributes?](#why-not-resource-attributes)
```

?

docs/non-normative/status-metrics.md
```diff
@@ -0,0 +1,95 @@
...
+  A timeseries is produced for every possible value of `example.state`. The
+  value of this metric is 1 for a state if it is currently in said state, and
+  is 0 for all other states.
+instrument: updowncounter
+unit: "1"
+attributes:
+- ref: example.state
+```
+
+### Naming
```
lmolkova: could you summarize it to one sentence and add it (with link to this doc) to https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#metrics ?

docs/non-normative/status-metrics.md
```diff
@@ -0,0 +1,95 @@
...
+in a particular state. Since the metric value is either `0` or `1` for a given
+state attribute value, this means you can do a simple sum aggregation to count
+instances of particular states.
+
+### Why not Resource Attributes?
+
+It is a common mistake to attach a "state" as a Resource Attribute. This is
+not recommended for two reasons:
+
+* Resource is intended to be immutable, thus adding an attribute like "state"
```
lmolkova: does it still apply considering entities and descriptive attributes being mutable?

docs/non-normative/status-metrics.md
ChrsMark: I haven't seen this confusing people so far (in PRs' reviews). But if people in general think it's an issue I'd fine changing it.

docs/non-normative/status-metrics.md
joaopgrassi: Yes I was also thinking the same when I reviewed but wasn't sure. 🤔

docs/non-normative/status-metrics.md
TylerHelmuth: I was confused when reviewing some of the new k8s metrics that include a required attribute that exactly matches the metric name. I prefer the guidance suggested in the PR.

TylerHelmuth: I'll propose it also makes translating to/from OTLP just a tiny bit harder.

When converting from the highly-structured OTLP format to other, non-otlp formats, being able to distinguish between a metric's name and the metric's datapoint's attributes is helpful. In OTLP we have the structure to distinguish between a metric's name and a datapoint attribute name, but in other formats we do have that structure. 

Specifically, having the values match makes it hard to use the metric's name as a key and the datapoint attribute name as a key in a flat structure.

OTLP of course allows metric names and datapoint attributes to match, and we would never change that, but in our semantic conventions I like requiring metric name and datapoint attribute name to be different. I think it provides the most flexibility for to/from translations without putting a big restriction on ourselves.

braydonk: I have opened an issue to get clarification around this in general here: https://github.com/open-telemetry/semantic-conventions/issues/2476

riverar:
> This can also close this old issue #199. CC @riverar

Can you elaborate on this? I read the document and didn't really come away with any clarity with regards to my very specific desktop app monitoring scenario. Maybe I missed something?

ChrsMark:
Hey @braydonk, any updates on this? It's currently blocking related progress i.e. https://github.com/open-telemetry/semantic-conventions/pull/2488, so it'd be nice to move this guidance forward.

joaopgrassi:
> > This can also close this old issue #199. CC @riverar
> 
> Can you elaborate on this? I read the document and didn't really come away with any clarity with regards to my very specific desktop app monitoring scenario. Maybe I missed something?

Well, in your original issue you described the need to observe/monitor the state/heartbeat of an app. This now described a state metric which basically can be used to report a boolean state 1/0/on/off using an UpDownCounter.

riverar:
@joaopgrassi Agreed. Took a look at it again this morning with fresh eyes and it makes sense to me. Thank you!

braydonk:
Summer busy-ness and other fires to put out has forced me to push this off for longer than I meant to. I still intend to come back and clean this up based on review comments.

docs/non-normative/status-metrics.md
ChrsMark: > In this case, using a metric name suffix is
recommended to avoid the naming clash, i.e. naming the attribute `k8s.phase` and
the metric `k8s.phase.current`.

Would this be allowed? I remember there was a restriction on using namespaces as standalone metrics/attributes, has this changed? 
@joaopgrassi maybe you know more on this?

ChrsMark: @TylerHelmuth we discussed this during the last System SemConv SIG (Sept 4th) with @open-telemetry/semconv-system-approvers and @jsuereth. The group in general is not convinced that this distinction should be enforced so as to support flattening fields which is not necessarily a pure OTel use-case/need. When transforming, I assume users can easily add a `metric.*` prefix to their metrics and/or `attribute.*` to the attributes to avoid such collisions.

Distinguishing them will make our lives hard on finding legit names for metrics like `k8s.container.status.reason` which are already long.

TylerHelmuth: Ok thanks for discussing the issue. I don't want to hold this up any longer.

ChrsMark: No worries.

In that case @braydonk do you think we can adjust the guidance accordingly? Or is there anything else missing to discuss here?

docs/non-normative/status-metrics.md
braydonk: Done in [ec07dec](https://github.com/open-telemetry/semantic-conventions/pull/2472/commits/ec07decaa0eb626a794485c8722dff8229cd9572)

docs/non-normative/status-metrics.md
braydonk: Changed the guidance in [ec07dec](https://github.com/open-telemetry/semantic-conventions/pull/2472/commits/ec07decaa0eb626a794485c8722dff8229cd9572) let me know what you think.

docs/non-normative/status-metrics.md
braydonk: I moved this document to that folder in [ec07dec](https://github.com/open-telemetry/semantic-conventions/pull/2472/commits/ec07decaa0eb626a794485c8722dff8229cd9572)

braydonk: I think this got moved since this comment; where is this now for me to add a link?

docs/non-normative/status-metrics.md
braydonk: Done in [ec07dec](https://github.com/open-telemetry/semantic-conventions/pull/2472/commits/ec07decaa0eb626a794485c8722dff8229cd9572) (but I forgot to regenerate the TOC that's a followup commit)

docs/non-normative/status-metrics.md
braydonk: Rewrote this section to target entities instead of resource in [ec07dec](https://github.com/open-telemetry/semantic-conventions/pull/2472/commits/ec07decaa0eb626a794485c8722dff8229cd9572)

docs/non-normative/how-to-write-conventions/status-metrics.md
```diff
@@ -0,0 +1,87 @@
...
+The metric is
+[instrumented as an `UpDownCounter`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#consistent-updowncounter-timeseries)
+rather than a `Gauge`. This is a deliberate choice, as it is a reasonable use 
+case to count objects that are in a particular state. Since the metric value is
+either `0` or `1` for a given state attribute value, this means you can do a
+simple sum aggregation to count instances of particular states.
+
+### Should it be an Entity Attribute?
+
+It is possible to make the status a descriptive attribute on an entity (**not**
```
braydonk, line 56: I feel okay about this but not 100% confident. @jsuereth @dmitryax any opinions here?

jsuereth: Yes - status would be part of the description of an entity.

We plan to allow descriptive attributes to elevate to metrics for systems that don't support a CMDB e.g. but want entity-state as a timeseries (similar to kube-state metrics)

ChrsMark (approved):
Thank's for putting this together!

docs/non-normative/how-to-write-conventions/status-metrics.md
```diff
@@ -0,0 +1,87 @@
...
+an identifying attribute). The primary upside to making a state metric is for
+tracking state changes over time; as an entity attribute, the history of state
+changes would not necessarily be tracked.
+
+### Naming
+
+If there is particular verbiage for this sort of thing within the instrumentation
```
joaopgrassi:
```suggestion
If there is already well-known naming conventions within the instrumentation
```
Maybe? The doc is very well written but "for this sort of thing" caught me off guard here :D

joaopgrassi (approved):
non blocking comment

docs/non-normative/how-to-write-conventions/status-metrics.md
```diff
@@ -0,0 +1,87 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Status Metrics
+aliases: [status-metrics]
```
lmolkova: curious why is this necessary?

lmolkova (approved):

docs/non-normative/how-to-write-conventions/status-metrics.md
braydonk: Yeah good point, sometimes I write a little too rushed and dip into informal language. :smile:

Reworded in [5e2b502](https://github.com/open-telemetry/semantic-conventions/pull/2472/commits/5e2b50298c5237c8cc1623e6b0982cea753594d9)

docs/non-normative/how-to-write-conventions/status-metrics.md
braydonk: I actually don't know; it was in the document I copied from https://github.com/open-telemetry/semantic-conventions/blob/5ee549b1ce30fe11fcb9b7e3bd35ebfb363f467f/docs/non-normative/http-migration.md?plain=1#L2

It looked needed so I left it, but I can remove it if we don't think it's needed.

docs/non-normative/how-to-write-conventions/status-metrics.md
```diff
@@ -0,0 +1,87 @@
...
+recommended naming scheme.  
+This naming scheme necessitates choosing words to represent two things:
+
+* The noun representing one of the set of possible values (the attribute name)
+* The adjective describing which of the values is currently active (the metric
+  name)
+
+The general recommendation is to use the word "state" for the attribute and
+"status" for the metric. This is derived from common turns of phrase for each
+word, respectively:
```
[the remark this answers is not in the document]
braydonk: [5e2b502](https://github.com/open-telemetry/semantic-conventions/pull/2472/commits/5e2b50298c5237c8cc1623e6b0982cea753594d9)

docs/non-normative/how-to-write-conventions/status-metrics.md
```diff
@@ -0,0 +1,87 @@
...
+changes would not necessarily be tracked.
+
+### Naming
+
+If there is particular verbiage for this sort of thing within the instrumentation
+context, i.e. `k8s` having words like `phase` or `status`, those words should
+always be chosen.
+
+If there are not any naming options immediately available, you may use this
+recommended naming scheme.
```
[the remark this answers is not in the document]
braydonk: This is a Markdown thing, it has to be two spaces.

docs/non-normative/how-to-write-conventions/status-metrics.md
```diff
@@ -0,0 +1,87 @@
...
+unit: "1"
+attributes:
+- ref: example.state
+```
+
+### Instrument
+
+The metric is
+[instrumented as an `UpDownCounter`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#consistent-updowncounter-timeseries)
+rather than a `Gauge`. This is a deliberate choice, as it is a reasonable use
```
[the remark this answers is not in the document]
braydonk: [5e2b502](https://github.com/open-telemetry/semantic-conventions/pull/2472/commits/5e2b50298c5237c8cc1623e6b0982cea753594d9)

docs/non-normative/how-to-write-conventions/status-metrics.md
```diff
@@ -0,0 +1,87 @@
...
+instrument: updowncounter
+unit: "1"
+attributes:
+- ref: example.state
+```
+
+### Instrument
+
+The metric is
+[instrumented as an `UpDownCounter`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#consistent-updowncounter-timeseries)
```
kamphaus, line 48: Given the compatibility spec https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#sums would `UpDownCounter` instrumentation require that the metric name when exported to Prometheus have the `_total` suffix?

docs/non-normative/how-to-write-conventions/status-metrics.md
```diff
@@ -0,0 +1,87 @@
...
+rather than a `Gauge`. This is a deliberate choice, as it is a reasonable use
+case to count objects that are in a particular state. Since the metric value is
+either `0` or `1` for a given state attribute value, this means you can do a
+simple sum aggregation to count instances of particular states.
```
kamphaus, line 52: I have seen state metrics where the metric value corresponds to an enum (eg. https://longhorn.io/docs/1.9.1/monitoring/metrics/ `longhorn_volume_state`).

Should such enum metrics be explicitly excluded from the definition in `docs/non-normative/how-to-write-conventions/status-metrics.md`?

docs/non-normative/how-to-write-conventions/status-metrics.md
dashpole: No.  Only monotonic sums get the _total suffix.  Non-monotonic sums (i.e. updowncounters) are converted to prometheus gauges.

docs/non-normative/how-to-write-conventions/status-metrics.md
jsuereth: I think it's only needed (aliases) if you've moved a document so URL links aren't broken.

You can remove aliases

docs/non-normative/how-to-write-conventions/status-metrics.md
jsuereth: We currently don't support `enum` (or string-value) metrics in our metric data model.

I think such metrics need an encoding in OTLP today (in GCP we use an `UpDownCounter` where the value is a label with a known key and the value is always `1`).  I wouldn't outlaw them, but it is something we probably need a way to deal with in OTLP (and the metric data model) itself at some point.

docs/non-normative/how-to-write-conventions/status-metrics.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Should it be an Entity Attribute?
+
+It is possible to make the status a descriptive attribute on an entity (**not**
+an identifying attribute). The primary upside to making a state metric is for
+tracking state changes over time; as an entity attribute, the history of state
+changes would not necessarily be tracked.
```
jsuereth: TBD, honestly.  I'd rephrase this:

Using a metric timeseries database to track entity state changes is a known use case for Entities.  Explicitly defining these state metrics helps metric-centric solutions understand how to interact with Entities.

docs/non-normative/how-to-write-conventions/status-metrics.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Naming
+
+If there is already well-known naming conventions within the instrumentation
+context, i.e. `k8s` having words like `phase` or `status`, those words should
+always be chosen.
+
+If there are not any naming options immediately available, you may use this
+recommended naming scheme.  
+This naming scheme necessitates choosing words to represent two things:
```
jsuereth, line 69: @dashpole WDYT of this vs. the `*_info` convention being used in OTEL<->Prometheus?

docs/non-normative/how-to-write-conventions/status-metrics.md
braydonk: What should I rephrase it to? Or should I remove the section altogether? It's kind of a chicken and egg problem; I wrote the guidance in the first place to explain why these things shouldn't be Resource Attributes and to come up with an alternate guidance, but if Entities are built to address this use case then is there ever still a reason to use a metric as described in this document?

docs/non-normative/how-to-write-conventions/status-metrics.md
kamphaus: Can't such enum metrics already be represented in otlp as `Gauge`?
The issue would be how to define the mapping enum to/from int value in SemConv.

docs/non-normative/how-to-write-conventions/status-metrics.md
jsuereth: Apologies, my comment was the thing to say here: "Using a metric timeseries database to track entity state changes is a known use case for Entities. Explicitly defining these state metrics helps metric-centric solutions understand how to interact with Entities."

That's why we want to allow this use case.  So it's not "why entity vs. metric", it's "both"

docs/non-normative/how-to-write-conventions/status-metrics.md
jsuereth: You can use Gauge or UpDownCounter.  The choice of UpDownCounter is so that spatial-aggregation defaults to SUM (vs. Gauge where it would just "pick one").

In both cases the mapping enum is the issue.

docs/non-normative/how-to-write-conventions/status-metrics.md
braydonk: Took a stab at rewording, does this sound right? Are there any additional qualifying details I might be able to add here?

jsuereth (approved):

docs/non-normative/how-to-write-conventions/status-metrics.md
jsuereth: Yeah - that's fine.  Eventually we want a generic way to take Entity-State and turn it into metrics for metric systems, so you COULD call that out, but what's written is fine.

trask (approved):

SylvainJuge:
When building "JMX state metrics" in Java instrumentation we took inspiration from the [`hw.status`](https://opentelemetry.io/docs/specs/semconv/hardware/common/#metric-hwstatus) metric defined in hardware semantic conventions, once this recommendation is merged it would be beneficial to also rename `hw.status` to `hw.state` to prevent future inconsistencies.

I am not aware of any other similar status metric elsewhere, but my knowledge is quite limited here so maybe an exhaustive review would be needed to identify other candidates for alignment.

braydonk:
> Once this recommendation is merged it would be beneficial to also rename hw.status to hw.state to prevent future inconsistencies.

`*.status` as the metric name and `*.state` as the attribute name is actually what is written as the guidance here; `hw.status` is matching the written guidance here. The upcoming metric `process.status` will use this as well. Other examples of this metric pattern exist in `k8s.*` but there is lots of existing verbiage there that is used instead.

SylvainJuge:
> `*.status` as the metric name and `*.state` as the attribute name is actually what is written as the guidance here; `hw.status` is matching the written guidance here. The upcoming metric `process.status` will use this as well. Other examples of this metric pattern exist in `k8s.*` but there is lots of existing verbiage there that is used instead.

My bad, I should have paid more attention while reading it. The good news is that there is nothing to change then, maybe as a follow-up adding a link from `hw.status` metric to this recommendation would be beneficial.

docs/non-normative/how-to-write-conventions/status-metrics.md
joaopgrassi: we have enough approvals, should we hold off merging by this comment or not? @jsuereth @dashpole

dashpole: Info isn't a perfect 1:1 with status metrics.  Many status metrics are not info (e.g. kube_pod_status_ready).  Others are better represented as prometheus stateset type (1's and 0's).  Info is also a broader category that encompasses things like resource or scope attributes.

I think we should use _info metrics when it makes sense in the conversion spec (e.g. target_info, or entity_foo_info), but should move forward with having our own conventions around status metrics.

PR #2484, opened 2025-07-05
Split multi use gen-ai spans in to 2 distinct spans #2467

thompson-tomo:
Progresses open-telemetry/semantic-conventions-genai#31

## Changes

In a couple of scenarios spans either had 2 name definitions or 2 span kinds. Each of these were for different use cases, hence these have been split into 2 spans.

## Reason for the change

This is a blocker for moving to code generated span registries as a span can only have 1 type, hence should be separate definitions.


thompson-tomo:
This pr has been split to only include the gen-ai part. @open-telemetry/semconv-genai-approvers can you review

model/gen-ai/spans.yaml
```diff
@@ -97,14 +97,35 @@ groups:
...
+      This span is to be used when the models is running in a different process or when
+      the GenAI call happens over instrumented protocol such as HTTP.
+      In other cases, [Embedded Inference](/docs/gen-ai/gen-ai-spans.md#embedded-inference) span should be used.
+
       **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.request.model}`.
       Semantic conventions for individual GenAI systems and frameworks MAY specify different span name format
       and MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.37.0/specification/trace/api.md#span).
+    extends: attributes.gen_ai.inference.client
+    attributes:
+      - ref: gen_ai.system
```
gyliu513: `gen_ai.system` now renamed to `gen_ai.provider.name` at https://github.com/open-telemetry/semantic-conventions/pull/2046

Also I can see the only difference of `span.gen_ai.embedded.internal` and `span.gen_ai.inference.client` is one is using `gen_ai.system` which has been deprecated and another is using `gen_ai.provider.name`, what is the use case here? Thanks

lmolkova (changes requested):
I don't believe that splitting spans by kind improves end user or instrumentation author experience. Now we duplicate huge span tables and the only difference between them is the kind.

Moreover, we know for a fact that client/internal is frequently a preference - e.g. Spring AI uses internal kind for external model calls - https://github.com/spring-projects/spring-ai/issues/1174, more context https://github.com/open-telemetry/semantic-conventions/issues/1315

model/gen-ai/spans.yaml
thompson-tomo: It will be an issue due to rebasing as the rename occurred after this PR was prepared.

thompson-tomo: Has been fixed

thompson-tomo:
So i have fixed the issue introduced by rebasing as identified by @gyliu513 and the internal span now no longer has the server attributes with the logic being if you know that you are talking to a sever then you aren't internal. Is there any other attributes which only apply when using a server based gen-ai model? For instance are usage tokens still applicable?

@lmolkova i am aware of the ambiguity around span king usage etc and believe that is addressed by using the phrase Should rather than must.  I think this scenario is different as the documentation is clearly talking about 2 different kinds of activities, hence splitting them and now removing the non applicable server attributes. If there is others happy to move them to create a bigger difference in definitions.

lmolkova:
I still don't understand which problem this PR is trying to solve and how is it solved by introducing `embedded inference` span in addition to `inference` and `embeddings`. I'm going to close it.

Issue #2320, opened 2025-06-03
Document otel.scope.schema_url used by non-OTLP exporters 


pellared:
we should document these in semantic-conventions as a follow-up - https://github.com/open-telemetry/semantic-conventions/blob/2bc97890c1ad82232745b4dd74fd3144476b6a5c/docs/registry/attributes/otel.md#otel-scope-attributes

_Originally posted by @lmolkova in https://github.com/open-telemetry/opentelemetry-specification/pull/4505#discussion_r2124199652_

pellared:
@lmolkova, can you please move this to https://github.com/open-telemetry/semantic-conventions?

joaopgrassi:
@pellared will you be working on this? Just so I know if I can assign it to you or not :)

pellared:
I can give it a try.

pellared:
I created https://github.com/open-telemetry/semantic-conventions/pull/2489

Issue #1550, opened 2024-07-22
[Logs] Reporting logger name, log bridge name and version


pellared:
**What are you trying to achieve?**

I would like a convention on how the log bridges should report:
- logger name
- log bridge [package] name
- log bridge [package] version

Thanks to this backends would get telemetry of the log bridges in use that can help with troubleshooting and provide information what log bridges are popular.

Currently instrumentation scope name is supposed to be used as a "logger name" per:
[opentelemetry-specification/specification/logs/bridge-api.md](https://github.com/open-telemetry/opentelemetry-specification/blob/2eca069034b3012ccef0378aa6fa61a5391c14d1/specification/logs/bridge-api.md#L74-L76)

>  For log sources which define a logger name (e.g. Java  [Logger Name](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String)))  the Logger Name should be recorded as the instrumentation scope name.

An initial proposal for the change was [here](https://github.com/open-telemetry/opentelemetry-specification/pull/3855#discussion_r1488101227), but it was postponed until now.

Slack thread before the issue was created: https://cloud-native.slack.com/archives/C041APFBYQP/p1721231579496729

## Proposal A

Add `bridge.name` and `bridge.version` instrumentation scope attributes to report the log bridge package name and version.
Use logger's instrumentation scope name parameter via Bridge API to report the logger name.

Benefits:
- Does not require changing the Bridge API specification. 

Drawbacks:
- Hard to implement bridges for log libraries which do not have a concept of a "logger name"  or they do not use them by default.
- Most SDKs does not support instrumentation scope attributes. Related PR: https://github.com/open-telemetry/opentelemetry-specification/pull/4146
- Acquiring a logger on hot-path may be less performant 
- High cardinality of instrumentation scope name. More [here](https://github.com/open-telemetry/semantic-conventions/issues/1550).

## Proposal B

Use logger's instrumentation scope parameters via Bridge API to report the log bridge package name and version.
Add `logger.name` log attribute to report the logger name.

Benefits:

- Similar pattern to instrumentation libraries where library name and version is reported as instrumentation scope.
- Since there is no need to obtain different Loggers for each name, performance hit in hot path can be avoided
- Not all logging libraries have a concept of a logger name. But all bridges should have some name.

Drawbacks:

[...]

PR #2489, opened 2025-07-07
Add otel.scope.schema_url to non-OTLP exporters

pellared:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/2320

Related to (follows):
- https://github.com/open-telemetry/opentelemetry-specification/pull/4505

The [Instrumentation Scope Attributes for non-OTLP formats](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/mapping-to-non-otlp.md#instrumentation-scope-attributes) already says what exporters can do with scope attributes.

### Changes

Add `otel.scope.schema_url` attribute to non-OTLP exporters that represents an instrumentation scope schema URL.

### Remarks

Schema URL is defined as part of instrumentation scope in the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/instrumentation-scope.md

However, in OTLP it is always on a "higher" level for reasons unknown to me (maybe it was historically introduced before the concept of the instrumentation scope?):  https://github.com/open-telemetry/opentelemetry-proto/blob/c30610041736aa5c0077b156f27b09e878b797ea/opentelemetry/proto/trace/v1/trace.proto#L58-L64 

We could consider also naming it `otel.schema_url`. However, with the pre-work of https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/instrumentation-scope.md and https://github.com/open-telemetry/opentelemetry-specification/pull/4505 I prefer `otel.scope.schema_url`.


pellared:
@open-telemetry/specs-semconv-maintainers, how do we want to mitigate the following validation error?

```
Violation: Attribute 'otel.scope' name is used as a namespace in the following attributes '["otel.scope.name", "otel.scope.schema_url", "otel.scope.version"]'.
```

See: https://github.com/open-telemetry/semantic-conventions/actions/runs/16116431161/job/45471273064?pr=2489

Should we make it `otel.scope.attr` or `otel.scope.attribute`?

CC @open-telemetry/prometheus-interoperability, as this is related to https://github.com/open-telemetry/opentelemetry-specification/pull/4505

ArthurSens:
@pellared, I've read all the links provided, but I'm still feeling a bit lost regarding the problem you're trying to solve here. Are we facing a name collision with other attributes set by semantic conventions?

model/otel/registry.yaml
```diff
@@ -71,6 +71,17 @@ groups:
...
+        type: string
+        brief: The schema URL of the instrumentation scope.
+        examples: ['https://opentelemetry.io/schemas/1.31.0']
+        stability: stable
+      - id: otel.scope
+        type: template[string]
+        brief: >
+          The attributes of the instrumentation scope - (`InstrumentationScope.Attributes` in OTLP).
+        examples: ["someval", "55"]
+        stability: stable
```
thompson-tomo: My understanding of https://github.com/open-telemetry/oteps/blob/main/text/0201-scope-attributes.md might not be 100% but I get the impression that scope attributes are the same as span/metric attributes hence we would not need this attribute?

thompson-tomo (commented):
Not sure if the otel.scope<key> attribute is what is meant by
> We will create a section for Scope attributes' semantic conventions in the specification.

I would more expect a registry of scopes in semconv which show the name, version & attributes. This would tie into showing of instrumentations which now sounds like scopes

model/otel/registry.yaml
pellared: These are not the same. Scope attributes are e.g. adding some metadata which is common for all spans/metrics/logs emitted by a tracer/meter/logger. Here is one example: https://github.com/open-telemetry/semantic-conventions/issues/1550

pellared:
> Are we facing a name collision with other attributes set by semantic conventions?

@ArthurSens, it looks there the semconv has a rule to not create attributes (even templatized) which is a "namespace" of other attributes. If my understanding is correct, then for Prometheus and OpenMetrics Compatibility we should consider renaming changing `otel_scope_[attribute]` to e.g. `otel_scope_attr_[key]` or `otel_scope_attribute_[key]`.

pellared:
Undrafting as this needs to be reviewed/discussed even though it is not ready to be merged.

ArthurSens:
> > Are we facing a name collision with other attributes set by semantic conventions?
> 
> @ArthurSens, it looks there the semconv has a rule to not create attributes (even templatized) which is a "namespace" of other attributes. If my understanding is correct, then for Prometheus and OpenMetrics Compatibility we should consider renaming changing `otel_scope_[attribute]` to e.g. `otel_scope_attr_[key]` or `otel_scope_attribute_[key]`.

[Prometheus has committed to supporting `otel_scope_[attribute]`](https://github.com/prometheus/prometheus/pull/16730), changing this now is a breaking change for us 😕

aknuds1:
> it looks there the semconv has a rule to not create attributes (even templatized) which is a "namespace" of other attributes. If my understanding is correct, then for Prometheus and OpenMetrics Compatibility we should consider renaming changing otel_scope_[attribute] to e.g. otel_scope_attr_[key] or otel_scope_attribute_[key].

@pellared Could you explain in more detail the reason for why we'd have to rename the otel_scope_* labels? The above is a bit too vague for me. It would be painful to make this change in Mimir and Grafana Cloud :/

pellared:
> Could you explain in more detail the reason for why we'd have to rename the otel_scope_* labels?

I am not sure if we have to do it. It just looks against semconv policies but I do not think that we need to follow them in Prometheus and OpenMetrics Compatibility. I think this is one of the reasons why we have a separate compatibility document.

I think that the proposal here is mostly for other exporters.

I would be happy if @open-telemetry/specs-semconv-maintainers share their opinion.

aknuds1:
> I am not sure if we have to do it. It just looks against semconv policies but I do not think that we need to follow them in Prometheus and OpenMetrics Compatibility. I think this is one of the reasons why we have a separate compatibility document.

@pellared Thanks. In retrospect, I realized this feature hasn't yet been enabled in Grafana Cloud, so the change would at the current stage actually not be painful.

@ArthurSens Do you think the [feature PR](https://github.com/prometheus/prometheus/pull/16730) should be marked do-not-release, until we figure this out?

ArthurSens:
> @ArthurSens Do you think the [feature PR](https://github.com/prometheus/prometheus/pull/16730) should be marked do-not-release, until we figure this out?

If there are risks that we need to change the OTel<->Prometheus spec, it would be awesome if we could delay the release. At the same time, Robert is saying that we won't need to change the spec...?

model/otel/registry.yaml
```diff
@@ -71,6 +71,17 @@ groups:
         brief: The version of the instrumentation scope - (`InstrumentationScope.Version` in OTLP).
         examples: ['1.0.0']
         stability: stable
+      - id: otel.scope.schema_url
+        type: string
+        brief: The schema URL of the instrumentation scope.
+        examples: ['https://opentelemetry.io/schemas/1.31.0']
+        stability: stable
```
lmolkova: any reason to start with `stable` and not `development` ?

dashpole:
It does seem like this protects against a common problem.  If I define a scope attribute with the key "name" or "version" (which seems plausible), it collides with the scope name or scope version.  I don't have a strong opinion on `attr` vs `attribute`.  Is there precedence for using `attr` as an abbreviation in semconv? Otherwise `attribute` seems most appropriate.

model/otel/registry.yaml
```diff
@@ -71,6 +71,17 @@ groups:
...
+      - id: otel.scope
```
lmolkova: could you remind me why we need to prefix attribute names with `otel.scope` ? why not just use whatever was provided in the `InstrumentationScope.Attributes` ?

lmolkova: oh, I see there is a discussion on it in the comments - https://github.com/open-telemetry/semantic-conventions/pull/2489#issuecomment-3045954740

lmolkova:
I believe we can/should fix collision policy for template attributes - the name of the attribute is `otel.scope.{key}` and not `otel.scope`, so there is no collision in practice, at least, no collision we can prevent.

To @dashpole's point

> It does seem like this protects against a common problem. If I define a scope attribute with the key "name" or "version" (which seems plausible), it collides with the scope name or scope version. 

If someone provides a scope with a name and also a `name` attribute on that scope, the value of us preventing this collisions is quite low.

But I don't see a reason to have attribute names prefixed with `otel.scope.`. It seems counterintuitive and unnecessary. 

If user provided attribute with the same name on the instrumentation scope and on the actual telemetry item, I don't really think we need to put any extra effort disambiguating those - this is probably a wrong usage pattern.

dashpole:
Thus far we've tried to preserve OTel's nested structure (resource, scope, datapoint) when translating to and from Prometheus, so the original intent was to be able to be able to distinguish scope attributes from datapoint attributes.  It does also significantly increase the likelihood of collisions. Do we actually document that users shouldn't set the same scope and datapoint attributes?

lmolkova:
> Thus far we've tried to preserve OTel's nested structure (resource, scope, datapoint) when translating to and from Prometheus, so the original intent was to be able to be able to distinguish scope attributes from datapoint attributes.

thanks for the reminder! Is it actually useful or is there a future where something is optimized based on this?

> It does also significantly increase the likelihood of collisions. 

This I don't agree with - the source of the data is OTel and OTel recommends namespacing everything - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#recommendations-for-application-developers. 

So good otel instrumentation should not put `name` on scope and `name` on data, it would put something like `foo.name` on the scope attribute and `bar.name` on the telemetry item.

If they put the same name on both, they want to pass the same thing twice - this seems redundant and ambigous. And while it's allowed, it makes no sense.

model/otel/registry.yaml
joaopgrassi: Both of you are correct :) Scope attributes are just as well a bag of attributes, same as on the attributes on each signal. And what @pellared said is also true, the scope applies to all signals it emits. 

~What I think the confusion here is and what I also don't get is why do we need the prefix `otel.scope.<key>` for scope attributes? I think they should just be added as is, without being under the "otel defined" prefix.~

I should have read the entire discussion: https://github.com/open-telemetry/semantic-conventions/pull/2489#issuecomment-3046160797 here's the reason why.

joaopgrassi:
> This I don't agree with - the source of the data is OTel and OTel recommends namespacing everything - [`main`/docs/general/naming.md#recommendations-for-application-developers](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#recommendations-for-application-developers).
> 
> So good otel instrumentation should not put `name` on scope and `name` on data, it would put something like `foo.name` on the scope attribute and `bar.name` on the telemetry item.
> 
> If they put the same name on both, they want to pass the same thing twice - this seems redundant and ambigous. And while it's allowed, it makes no sense.

I also agree here with @lmolkova. If the deal is because of prometheus compatibility, we already have that defined in the other document: https://github.com/open-telemetry/opentelemetry-specification/blob/ac137b349b006dc926da91c6a6619217737921cc/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope. So I'm not sure this needs to be defined in semconv.

model/otel/registry.yaml
thompson-tomo: @pellared sorry I wasn't too clear, I meant scope attribute names are the same as span attribute names. Note it appears the reason for them to be different is no longer the case.

thompson-tomo:
> Do we actually document that users shouldn't set the same scope and datapoint attributes?

@dashpole this is actually already defined as a part of https://github.com/open-telemetry/oteps/blob/main/text/0201-scope-attributes.md#attribute-value-precedence as it mentions what takes precedence etc.

At the same time the back ends are already being informed if the attribute is from the scope or data point, hence I don't see a need to also distinguish it on the attribute names.

model/otel/registry.yaml
pellared: No, we can start with development.

model/otel/registry.yaml
pellared: I am not sure if this is the same discussion. Do you propose to export the scope attributes "as they are" without adding `otel.scope` attributes? I think that adding `otel.scope` is important so that an receiver can correctly convert it back to an OTel-like data model (e.g. see the Prometheus compatibility).

Please resolve the comment if it is no longer relevant.

You can also check: https://github.com/open-telemetry/semantic-conventions/pull/2489#discussion_r2192579062

pellared:
I remember that when I originally proposed https://github.com/open-telemetry/opentelemetry-specification/pull/4505 I did not wanted to add any `_attr_` infix as it would feel very unnatural, awkward, and unnecessary.  For similar reasons, I feel that adding `attribute` to a attribute name is non-sense and it does not provide any value for the end-user.

policies/attribute_name_collisions.rego
```diff
@@ -41,6 +41,9 @@ deny contains attr_registry_collision(description, name) if {
     name := attribute_names[i].name
     prefix := attribute_names[i].namespace_prefix
 
+    # ignore otel.scope attribute
+    name != "otel.scope"
```
pellared: > I believe we can/should fix collision policy for template attributes 

@lmolkova, @joaopgrassi, do you have any other preference how to make changes in the collision policy?

pellared:
I think this PR is ready for an actual review.

CC @open-telemetry/prometheus-interoperability

model/otel/registry.yaml
pellared: @thompson-tomo, I am not sure if I follow you. Do you mean that `otel.scope` prefix should not be added? This is important to be able for the receiver to know that this is a "scope-level" attribute so that it is possible to make further transformations without loss. So for instance, thanks to the `otel_scope_` prefix in `prometheusreceiver` it is possible to map these attributes to scope attributes instead of metrics attributes.

thompson-tomo: Correct I am wondering if it should be added.
Would it be correct to say that otel.scope.<key> would never be sent via otlp as the key of an attribute and it is just used added by a target being scraped and removed by the prometheusreceiver?

model/otel/deprecated/entities-deprecated.yaml
```diff
@@ -12,3 +12,7 @@ groups:
         requirement_level: recommended
       - ref: otel.scope.version
         requirement_level: recommended
+      - ref: otel.scope.schema_url
+        requirement_level: recommended
+      - ref: otel.scope
+        requirement_level: recommended
```
thompson-tomo: Why are we adding attributes to a deprecated signal? These should not be there.

model/otel/registry.yaml
pellared: Yes. 

>  removed by the prometheusreceiver

The `otel.scope.` prefix is removed but the attribute is put in the correct place (scope attribute, instead of metric attribute).

thompson-tomo (changes requested):
Still discussing the attributes topic but found another small thing.

model/otel/deprecated/entities-deprecated.yaml
pellared: Thanks. I have not noticed that it is deprecated (sic!)  :sweat_smile: 

https://github.com/open-telemetry/semantic-conventions/pull/2489/commits/dad5f064374576349c2cf379a645d3ed7edda249

model/otel/registry.yaml
joaopgrassi: Since we already have the two `name/version` defined for non-otlp receivers I guess it makes sense to also add this `otel.scope.<key>` template attribute, so the entire scope can be re-constructed. I think I understand the need now.

model/otel/registry.yaml
thompson-tomo: In that case, my proposal would be to leave it out of semconv as it won't actually be transmitted with any otel signal and instead put it in the docs that this needing to bevdefined. That being said if @lmolkova / @joaopgrassi see benefit of having it in, with a big note/brief explaining the special usage of this attribute.

policies/attribute_name_collisions.rego
joaopgrassi: Maybe we just exclude template attributes from the namespace check? https://github.com/open-telemetry/semantic-conventions/blob/8638e7b4b9ff38da58559e04ef63172cca04ed0a/policies/attribute_name_collisions.rego#L39

But we will probably need to add the `type` property to https://github.com/open-telemetry/semantic-conventions/blob/8638e7b4b9ff38da58559e04ef63172cca04ed0a/policies/attribute_name_collisions.rego#L6

pellared: WDYT of creating a separate issue to tackle this (as it does not seem trivial to me)?
We could leave a hyperlink to the created issue in a TODO comment.
The alternative could be to do this in a separate PR before this one is merged.

joaopgrassi: Yes, it has to be either done as part of this PR or before, otherwise we can never merge this as the check will fail. I'd prefer to not go with the current state and implement the proper exclusion.

pellared: > I'd prefer to not go with the current state and implement the proper exclusion.

@joaopgrassi, are you able to help and create a PR for the proper exclusion?

joaopgrassi: Let's see what @lmolkova thinks, I have to confess I'm not so up to speed with all the rego rules, haven't written one myself yet :D. But yeah if that's the direction we agree I can do that sure :)

ArthurSens:
We managed to revert the feature in Prometheus before it made it into a release: https://github.com/prometheus/prometheus/pull/16842

It should be safe to change the spec

pellared:
> It should be safe to change the spec

Can you please how do we want to change the spec?
What do you changes do you have in mind, the ones from this PR or some follow-up in Spec repo?

pellared:
OTel Logs SIG meeting notes (from my memory):

For most use-cases non-OTLP exporters could simply put scope attributes to attributes/tags without any modifications.

However, for places where OTel's nested structure (resource, scope, datapoint) when translating is important adding an `otel.scope.` prefix can be allowed (it would be especially handy for Collector receivers). Probably this should be an opt-in behavior for the non-OTLP exporters.

It looks like we do not need to describe anything in the Semantic Conventions.

_@lmolkova, @trask, @austinlparker, feel free to edit my comment if I missed anything or made any mistake._

pellared:
Maybe the existing description in [Instrumentation Scope Attributes for non-OTLP formats](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/mapping-to-non-otlp.md#instrumentation-scope-attributes) is good enough?

**I changed this PR so it only adds `otel.scope.schema_url` which is not controversial and I also plan create a similar PR for (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/mapping-to-non-otlp.md#instrumentation-scope-attributes.**

policies/attribute_name_collisions.rego
pellared: Removed in https://github.com/open-telemetry/semantic-conventions/pull/2489/commits/bdd6fc1f1190a62123392770f0b640311faa1d0c

model/otel/registry.yaml
pellared: > my proposal would be to leave it out of semconv

We had the same discussion during Logs SIG meeting.

I decided the decrease the scope of this PR.

See: https://github.com/open-telemetry/semantic-conventions/pull/2489/commits/bdd6fc1f1190a62123392770f0b640311faa1d0c

model/otel/registry.yaml
pellared: I decided to remove `otel.scope.<key>` from this PR.

model/otel/registry.yaml
```diff
@@ -71,6 +71,11 @@ groups:
         brief: The version of the instrumentation scope - (`InstrumentationScope.Version` in OTLP).
         examples: ['1.0.0']
         stability: stable
+      - id: otel.scope.schema_url
+        type: string
+        brief: The schema URL of the instrumentation scope.
+        examples: ['https://opentelemetry.io/schemas/1.31.0']
+        stability: stable
```
pellared: Do we want/need to make it "development" first?

model/otel/registry.yaml
```diff
@@ -71,6 +71,11 @@ groups:
         brief: The version of the instrumentation scope - (`InstrumentationScope.Version` in OTLP).
         examples: ['1.0.0']
         stability: stable
+      - id: otel.scope.schema_url
+        type: string
+        brief: The schema URL of the instrumentation scope.
```
pellared, line 76: Schema URL is defined as part of instrumentation scope in the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/instrumentation-scope.md

However, in OTLP it is always on a "higher" level for reasons unknown to me (maybe it was historically introduced before the concept of the instrumentation scope?):  https://github.com/open-telemetry/opentelemetry-proto/blob/c30610041736aa5c0077b156f27b09e878b797ea/opentelemetry/proto/trace/v1/trace.proto#L58-L64 

We could consider also naming it `otel.schema_url`. However, with the pre-work of https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/instrumentation-scope.md and https://github.com/open-telemetry/opentelemetry-specification/pull/4505 I prefer `otel.scope.schema_url`.

ArthurSens:
> > It should be safe to change the spec
> 
> Can you please how do we want to change the spec? What do you changes do you have in mind, the ones from this PR or some follow-up in Spec repo?

What I meant here is that we were able to revert our commitment to translating Scope attributes to `otel_scope_[attr]` before it gets released in Prometheus. This means we can adjust the spec, e.g., using `otel_scope_attr_[attr]` instead **_if_** we so desire and that won't be a breaking change in Prometheus. And by spec, I mean the [Prometheus and OpenMetrics compatibility](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1).

model/otel/registry.yaml
pellared: I added it to the PR description.

pellared:
> What I meant here is that we were able to revert our commitment to translating Scope attributes to otel_scope_[attr] before it gets released in Prometheus. This means we can adjust the spec, e.g., using otel_scope_attr_[attr] instead if we so desire and that won't be a breaking change in Prometheus. And by spec, I mean the [Prometheus and OpenMetrics compatibility](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1).

@ArthurSens, with the conversation we had I think it is find to keep `otel_scope_[attr]`. There is a question whether users would not simply prefer `[attr]` instead. This could be potentially configurable. In my opinion this can be discussed separately by @open-telemetry/prometheus-interoperability. I am sorry for my bothersome way of working...

ArthurSens:
> @ArthurSens, with the conversation we had I think it is find to keep `otel_scope_[attr]`. There is a question whether users would not simply prefer `[attr]` instead. This could be potentially configurable. In my opinion this can be discussed separately by @open-telemetry/prometheus-interoperability. I am sorry for my bothersome way of working...

No problems at all! I'm glad we have the chance to align, please don't hesitate to keep tagging us when things aren't clear

model/otel/registry.yaml
lmolkova: yes, we do - https://github.com/open-telemetry/semantic-conventions/blob/37d19be4c609decf6cfec1ba4a89846e3ab5e0fb/docs/general/how-to-define-semantic-conventions.md?plain=1#L32

pellared:
```suggestion
        stability: development
```

dashpole (approved):

Issue #2390, opened 2025-06-19
Specify Relational attributes of an entity


thompson-tomo:
### Area(s)

docs, schema

### What's missing?

A way to indicate which attributes on an entity are reference attributes to other entities.

### Describe the solution you'd like

When specifying an attribute as being a part of an entity we should be able to specify an attribute as having a relational role which indicates that it needs to match the identifying attributes of another entity.

bertysentry:
+1

I agree with the principle. In hardware semantic conventions, we have `hw.parent.id` and `hw.parent.type` to specify a parent of the monitored entity, with `hw.parent.id` that has to match with `hw.id`, and `hw.parent.type` with `hw.type` of the parent entity.

The above is imperfect as it suppose unicity in { `hw.id`, `hw.type` }, which may not apply to other entities in other areas of semantic conventions (like databases, services, etc.).

It would be great to have a way to properly reference a "related" entity (parent, member-of, etc.) so that we don't invent various ways of doing this in various semconvs.

joaopgrassi:
@open-telemetry/entities-maintainers was this ever triaged by you? I added to the project board but would like to confirm.

PR #2492, opened 2025-07-07
[chores] check collisions for metric names

### LINKED ISSUE #2419: [chore] Update policy to check names for metrics/events
Currently our policy is checking attribute names to make sure they are not colliding with namespaces i.e. there shouldn't be `domain.namespace.x` and `domain.namespace` names because first one implies a namespace, so it’s not clear if `domain.namespace` is a leaf node or a namespace.

We should also enforce the same rule for metrics and events names

trisch-me:
Fixes #2419

This PR adds policy to check metric names for namespace collision.

There are existing collisions in hardware namespace. Those are added as exception for now. 

Related: #2380 #50 

FYI @bertysentry @lmolkova @jsuereth

### DIFF
--- a/policies/metrics_collisions.rego
+++ b/policies/metrics_collisions.rego
@@ -23,6 +23,7 @@ deny contains metrics_registry_collision(description, name) if {
 
 
     exceptions = {
+        # TODO: https://github.com/open-telemetry/semantic-conventions/issues/3058
         # legacy hardware metrics that are known to cause collisions
         "hw.battery.charge", "hw.cpu.speed", "hw.fan.speed", "hw.temperature", "hw.voltage"
     }


policies/entities_collisions.rego
```diff
@@ -0,0 +1,47 @@
...
+import rego.v1
+
+# Data structures to make checking things faster.
+entities_names := { obj |
+    group := input.groups[_]
+    group.type = "entity"
+    obj := { "name": group.name, "const_name": to_const_name(group.name), "namespace_prefix": to_namespace_prefix(group.name)}
+}
+
+# check that matric names do not collide with namespaces
```
lmolkova:
```suggestion
# check that entities names do not collide with namespaces
```

policies/entity_collisions.rego
```diff
@@ -0,0 +1,42 @@
...
+import rego.v1
+
+# Data structures to make checking things faster.
+entities_names := { obj |
+    group := input.groups[_]
+    group.type = "entity"
+    obj := { "name": group.name, "namespace_prefix": extract_entity_namespace_prefix(group.name)}
+}
+
+# check that matric names do not collide with namespaces
```
lmolkova:
```suggestion
# check that entity names do not collide with namespaces
```

policies/metrics_collisions.rego
```diff
@@ -0,0 +1,43 @@
...
+import rego.v1
+
+# Data structures to make checking things faster.
+metric_names := { obj |
+    group := input.groups[_];
+    group.type = "metric"
+    obj := { "name": group.metric_name, "namespace_prefix": extract_metric_namespace_prefix(group.metric_name)}
+}
+
+# check that matric names do not collide with namespaces
```
lmolkova:
```suggestion
# check that metric names do not collide with namespaces
```

policies/entity_collisions.rego
```diff
@@ -0,0 +1,42 @@
...
+deny contains entities_registry_collision(description, name) if {
```
lmolkova: could we add some tests for entities and metrics?

policies/metrics_collisions.rego
```diff
@@ -0,0 +1,43 @@
...
+deny contains metrics_registry_collision(description, name) if {
+    some i
+
+    name := metric_names[i].name
+    prefix := metric_names[i].namespace_prefix
```
lmolkova, line 21: let's exclude deprecated metrics - it should cover most, if not all, of the existing collisions.

Same on entities - it's ok to have collision with a deprecated thing

trisch-me: true, missed that attribute, thanks!

trisch-me:
After adding the deprecation check it’s now only conflicting entities present

```yaml

Diagnostic report:

Violation: Entity with name 'cicd.pipeline' is used as a namespace in the following entities '["cicd.pipeline.run"]'.
  - Category         : namespace_collision
  - SemConv group    : 
  - SemConv attribute: cicd.pipeline

Violation: Entity with name 'host' is used as a namespace in the following entities '["host.cpu"]'.
  - Category         : namespace_collision
  - SemConv group    : 
  - SemConv attribute: host

Violation: Entity with name 'process' is used as a namespace in the following entities '["process.runtime"]'.
  - Category         : namespace_collision
  - SemConv group    : 
  - SemConv attribute: process
```

policies/entity_collisions.rego
```diff
@@ -0,0 +1,48 @@
...
+
+        other.name != name
+        startswith(other.name, prefix)
+    ]
+    count(collisions) > 0
+    description := sprintf("Entity with name '%s' is used as a namespace in the following entities '%s'.", [name, collisions])
+}
+
+
+entities_registry_collision(description, entity_name) = violation if {
```
jsuereth: So `entity_type` is *intended* to be a namespace for entities.

The is for an entity type `k8s.pod`, we expect attributes to be `k8s.pod.*` generally.  We aren't making this a hard requirement yet, but that's the intention.

I realized, that should be updated in the [modelling guide](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/how-to-write-conventions/resource-and-entities.md#how-to-namespace-entities).

jsuereth (changes requested):

policies/entity_collisions.rego
trisch-me: but here are 2 entity names are colliding. Not entity and their attributes
Does your comment still apply then? So we expect to have entity `k8s.pod` and other entities _under_ this one as well?

thompson-tomo:
My 2 cents on the entities aspect is that it should be ok for entities to be nested. Ideally we could enforce that a parent entity must also contain the identifying attributes of the child with the role being relational or something similar. That is what I was thinking with #2390. 

As it stands there is no way to express a relationship between metrics and no/very low interest in doing so I would block it until that changes.

policies/entity_collisions.rego
jsuereth: We haven't yet decided how to do parent-child relationships with entities.  It is possible we have overlap like this, and it's actually relatively common so far in entities defined across semconv, as you see in your policy.

I think we should open a ticket to discuss how we want to model entity type names, but I wouldn't include it in this policy yet until we've discussed.

trisch-me: I agree with you and that was just an idea during one of the semconv to implement this check.
But I’m not sure even for metrics if we should have this restriction or not (after merging hardware PR it will start to fail) as we haven’t discussed yet

lmolkova:
Semconv SIG discussions:
- let's remove entities
- let's allow collisions with deprecated metrics
- let's add exception list for current HW metrics and create an issues to clean them up

policies/entity_collisions.rego
trisch-me: removed entities

policies/entity_collisions.rego
trisch-me: tests added

trisch-me:
@jsuereth @lmolkova as discussed - I have removed check for entities. 
metrics are updated and tests are added

lmolkova:
@trisch-me could you please update the PR description?

policies/metrics_collisions.rego
```diff
@@ -0,0 +1,60 @@
+package after_resolution
```
lmolkova, line 1: leaving a comment there - please update PR description

trisch-me:
@lmolkova thanks, updated

Issue #2480, opened 2025-07-04
`make check` (specifically the `normalized-link-check` target) fails on macOS


trentm:
tl;dr: See my "open question" section at the bottom.

> [...] The `normalized-link-check` target fails on macOS, at least for me, because the `model/[a-z]*` shell pattern with the `/bin/sh` shell *does* include "model/README.md" for some reason I don't understand. With macOS `grep` one could instead use `--exclude model/README.md`, but I couldn't get that to work with the grep in a `ubuntu:latest` container, so I gave up on that. 

 _Originally posted by @trentm in [#2214](https://github.com/open-telemetry/semantic-conventions/issues/2214#issuecomment-2859970247)_

... time passes ...

I hit this again today (`make check` failing on macOS), forgot that I had hit this before, re-learned where and why this is failing, re-tried to find a solution.

# the issue

Here is the Makefile target that fails on macOS (for me at least):

```makefile
.PHONY: normalized-link-check
# NOTE: we check `model/[a-z]*` to avoid `model/README.md`, which contains
# valid occurrences of `../docs/`.
normalized-link-check:
  @if grep -R '\.\./docs/' docs model/[a-z]*; then \
    echo "\nERROR: Found occurrences of '../docs/'; see above."; \
    echo "       Remove the '../docs/' from doc page links referencing doc pages."; \
    exit 1; \
  else \
    echo "Normalized-link check passed."; \
  fi
```

The issue (for me) is that the `grep -R '\.\./docs/' docs model/[a-z]*` does *not* skip "model/README.md" as intended on macOS for me. It *does* skip that file in CI and when running on Linux, e.g.:

```sh
% docker run -ti --rm --volume ./:/src/ ubuntu:latest /bin/sh
# cd src
# grep -R '\.\./docs/' docs model/[a-z]*
#
```

# the cause

I believe the cause of the issue is this: `model/[a-z]*` is using shell expansion. Shell expansion of that using the default macOS shell (`zsh`) will exclude "model/README.md" as intended:

```sh
% echo $SHELL
/bin/zsh
% echo model/[a-z]*
[...]

PR #2493, opened 2025-07-07
[chore] Fix 'make check' on macOS

trentm:
This fixes the breakage of `make normalized-link-check` on macOS by
carefully finding an invocation of `grep ...` that searches docs/ and
model/ but skips "model/README.md" and works with `sh` and `grep` on
macOS and linux.

Fixes: https://github.com/open-telemetry/semantic-conventions/issues/2480

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,7 +114,7 @@ misspell-correction:	$(MISSPELL)
 	$(MISSPELL) -w $(ALL_DOCS)
 
 .PHONY: normalized-link-check
-# NOTE: Seach "model/*/**" rather than "model" to skip `model/README.md`, which
+# NOTE: Search "model/*/**" rather than "model" to skip `model/README.md`, which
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -114,10 +114,10 @@ misspell-correction:	$(MISSPELL)
 	$(MISSPELL) -w $(ALL_DOCS)
 
 .PHONY: normalized-link-check
-# NOTE: we check `model/[a-z]*` to avoid `model/README.md`, which contains
-# valid occurrences of `../docs/`.
+# NOTE: Seach "model/*/**" rather than "model" to skip `model/README.md`, which
```
lmolkova:
```suggestion
# NOTE: Search "model/*/**" rather than "model" to skip `model/README.md`, which
```

lmolkova (approved):

PR #2495, opened 2025-07-07
Add SQL commenter as context propagation for databases

XSAM:
Part of https://github.com/open-telemetry/semantic-conventions/issues/2162, Replace https://github.com/open-telemetry/semantic-conventions/pull/2236. All previous commits are copied to this PR.

This PR focused solely on the SQL commenter, as https://github.com/open-telemetry/semantic-conventions/pull/2363 has been merged.

https://github.com/XSAM/otelsql/pull/512 is the prototype that demonstrates the SQL commenter with the `text attributes` propagator.

## Changes

Add SQL commenter as context propagation for databases


XSAM:
cc @lmolkova @pellared @Kielek @zacharycmontoya @nenadnoveljic @trask

frzifus:
cc @simonpasquier

docs/database/database-spans.md
```diff
@@ -478,6 +480,32 @@ Semantic conventions for individual database systems or specialized instrumentat
 MAY specify a different `db.query.summary` format as long as produced summary remains
 relatively short and its cardinality remains low comparing to the `db.query.text`.
 
+## Context propagation
+
+**Status**: [Development][DocumentStatus]
+
+### SQL commenter
+
+Instrumentations MAY propagate context using [SQL commenter](https://google.github.io/sqlcommenter/spec/) by injecting comments into SQL queries before execution. Context injection SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
```
pellared:
```suggestion
Instrumentations MAY propagate context using [SQL commenter](https://google.github.io/sqlcommenter/spec/) by injecting comments into SQL queries before execution. Context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
```

pellared (approved):

docs/database/database-spans.md
```diff
@@ -478,6 +480,32 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+Instrumentations MAY propagate context using [SQL commenter](https://google.github.io/sqlcommenter/spec/) by injecting comments into SQL queries before execution. Context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
+
+The instrumentation implementation MAY choose to either **append** the comment to the end of the query or **prepend** the comment at the beginning of the query, depending on the specific database system's requirements or preferences.
+
+The instrumentation SHOULD allow users to pass a list of propagators to overwrite the default propagators.
+
+The instrumentation MAY provide the following propagators that users can use to overwrite the default propagators:
+
+- `service.name` propagator, a propagator that can inject the `service.name` into carrier.
```
zacharycmontoya: Super small nit:

```suggestion
- `service.name` propagator, a propagator that can inject the `service.name` resource attribute into carrier.
```

zacharycmontoya (approved):

nenadnoveljic (approved):

docs/database/database-spans.md
```diff
@@ -478,6 +480,32 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+### SQL commenter
+
+Instrumentations MAY propagate context using [SQL commenter](https://google.github.io/sqlcommenter/spec/) by injecting comments into SQL queries before execution. Context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
+
+The instrumentation implementation MAY choose to either **append** the comment to the end of the query or **prepend** the comment at the beginning of the query, depending on the specific database system's requirements or preferences.
+
+The instrumentation SHOULD allow users to pass a list of propagators to overwrite the default propagators.
+
+The instrumentation MAY provide the following propagators that users can use to overwrite the default propagators:
```
Kielek: There is a chance that I missed something, but it is not obvious what are `default propagators` here.

docs/database/database-spans.md
```diff
@@ -478,6 +480,32 @@ Semantic conventions for individual database systems or specialized instrumentat
 MAY specify a different `db.query.summary` format as long as produced summary remains
 relatively short and its cardinality remains low comparing to the `db.query.text`.
 
+## Context propagation
+
+**Status**: [Development][DocumentStatus]
+
+### SQL commenter
+
+Instrumentations MAY propagate context using [SQL commenter](https://google.github.io/sqlcommenter/spec/) by injecting comments into SQL queries before execution. Context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
```
cheempz: Do we leave it to each language/instrumentation the level (and naming) of opt-in configurability, e.g. enable feature for all DBs, certain DBs, only if not prepared statements, etc. 

And similarly no prescription on how users can pass in the list of propagators?

docs/database/database-spans.md
pellared: I assume the globally set propagator as usual

docs/database/database-spans.md
```diff
@@ -478,6 +480,32 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+The instrumentation implementation MAY choose to either **append** the comment to the end of the query or **prepend** the comment at the beginning of the query, depending on the specific database system's requirements or preferences.
```
lmolkova: can we pick one as a default to recommend? 
we can say something like


```suggestion
The instrumentation implementation SHOULD **append** the comment to the end of the query. Semantic conventions for individual database systems MAY specify different format depending on the specific database system's requirements or preferences.
```

docs/database/database-spans.md
```diff
@@ -478,6 +480,32 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+The instrumentation SHOULD allow users to pass a list of propagators to overwrite the default propagators.
```
lmolkova: why a list? I believe in OTel you always configure one propagator, but it may be a composite one

docs/database/database-spans.md
```diff
@@ -478,6 +480,32 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+The instrumentation MAY provide the following propagators that users can use to overwrite the default propagators:
+
+- `service.name` propagator, a propagator that can inject the `service.name` resource attribute into carrier.
```
lmolkova: I don't think database instrumentation should care about this level of detail. It knows how to call a propagator, but what it injects is not relevant to them.

I would rewrite this section assuming that service name propagator is provided by the application or some contrib component

docs/database/database-spans.md
lmolkova: why instrumentation should 'provide' any propagators, it can use some propagator provided to it? DB instrumentation can't provide `service.name` propagator, at least, because it can't access resource attributes. 

Something outside of DB instrumentation would need to provide such propagator

docs/database/database-spans.md
XSAM: You are right. One propagator should be enough.

docs/database/database-spans.md
XSAM: > some contrib component

I wonder how could that work in the real world implementation, as a instrumentation itself could already be a contrib component.

Did you mean we need to specify a `service.name` propagator in language contrib, like https://github.com/open-telemetry/opentelemetry-go-contrib?

docs/database/database-spans.md
XSAM: > why instrumentation should 'provide' any propagators

If instrumentation does not provide the `service.name` propagator, then that probably means the user would need to write a propagator, which could be very inconvenient.

> DB instrumentation can't provide service.name propagator, at least, because it can't access resource attributes.

It actually can by using user specified value. See https://github.com/XSAM/otelsql/pull/512/files#diff-65caf616dc77b09c44c1fc0eba33369464655b1318bdedb0051351b635ebf881R139

```go
otelsql.WithSQLCommenter(true,
	propagation.NewCompositeTextMapPropagator(
		otelsql.ServiceNamePropagator{ServiceName: "otelsql-example"},
		propagation.TraceContext{}, // Optional, if you want to propagate trace context
	),
)
```

I know it is ugly, but since there is no entity provider yet, it is what we have.

> Something outside of DB instrumentation would need to provide such propagator

I don't know how to write a section to recommend/suggest a `service.name` propagator while it cannot be provided by the instrumentation. There is no example I found in the semantic convention states like this. Could you help with this?

docs/database/database-spans.md
nenadnoveljic: Can we prepend by default to avoid truncation? Discussed [here](https://github.com/open-telemetry/semantic-conventions/pull/2236#discussion_r2078566667).

XSAM: I would slightly prefer append by default. Maybe it is better up to instrumentation author's preference to decide whether to use append or prepend.

Another discussion [here](https://github.com/open-telemetry/semantic-conventions/pull/2236#discussion_r2103582514).

docs/database/database-spans.md
XSAM: I think it is up to DB instrumentation to decide how to apply comments to a certain query (like skip the comment injection or not)

> And similarly no prescription on how users can pass in the list of propagators?

It is up to the language or instrumentation preference on how to leave an option for users to opt into the custom propagator.

docs/database/database-spans.md
XSAM: I feel it is a never-ending discussion. Maybe we should leave it to the instrumentation author's preference. Could we not define a default one for now?

docs/database/database-spans.md
XSAM: I have split the definition of the `service-name` propagator to another section, and instrumentation does not need to care about this detail.

It would be the language SDK contrib's concern to implement such a propagator, just like the AWS propagator. For instance, in Go, we have [aws x-ray propagator](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/propagators/aws/xray/propagator.go) in contrib.

Context: https://cloud-native.slack.com/archives/C041APFBYQP/p1752514146312029?thread_ts=1752509024.937119&cid=C041APFBYQP

docs/database/database-spans.md
XSAM: `service-name` propagator is moved out of the db instrumentation.

docs/database/database-spans.md
```diff
@@ -478,6 +481,46 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+  SELECT * FROM songs /*traceparent='00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01',tracestate='congo%3Dt61rcWkgMzE%2Crojo%3D00f067aa0ba902b7'*/
+  ```
+
+### `service-name` propagator
+
+The `service-name` propagator is a propagator that can inject the `service.name` resource attribute into carrier.
+
+SDK contrib MAY provide this propagator for databases context propagation.
+
+The `service-name` propagator need to be manually configured by the user to know the `service.name` value if OpenTelemetry does not have a way to automatically fetch the `service.name` value.
```
brettmc: I don't think users should need to configure the service name, since there are many sources it can come from (environment, declarative config, manually configured SDK, ...) - and, if it's just a user-provided value then there is no guarantee that it's the same value as `service.name` as defined in resource attributes.
Likewise, SDK contrib shouldn't try to guess the service name for the same reason. It's tedious and error-prone to duplicate checking different possible sources of `service.name` in the correct order.

I don't think that the resource attributes are exposed in any spec-approved way. It seems to me that if `service.name` is accepted as a valid value for context propagation, then there should be an agreed way to retrieve it. A couple of options could be:
- add `service.name` to [SpanContext](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.46.0/specification/overview.md#spancontext)
- add `TracerProvider.GetResource` or `TracerProvider.GetServiceName`.

What about `service.namespace` and `service.version`, since `service.name` might not be uniquely identifying by itself?

XSAM: > I don't think users should need to configure the service name, since there are many sources it can come from

That means the propagator needs to access the OTel SDK. But, it should be ok if this propagator lives in the contrib.

However, this is not a guarantee that it is the same value as `service.name`, as users might have a different way to retrieve their own `service.name` instead of using some automatic discoveries. And, I don't think you could include every automatic discovery in a propagator. The easiest way (when there is no entity provider API) might be ask users.

> if it's just a user-provided value then there is no guarantee that it's the same value as service.name as defined in resource attributes.

Users are responsible for providing the correct value.

> What about service.namespace and service.version, since service.name might not be uniquely identifying by itself?

Good question. I feel we might need a generic propagator to inject. Maybe a `attributes` propagator that can inject a list of attributes?

> then there should be an agreed way to retrieve it.

We could probably let the `attributes` propagator (I would like to use `TextMapPropagator`, but it is reserved in the API) retrieve the user-defined key list from the carrier to make it more legitimate.

An example in Go

```go
type AttributesPropagator struct {
	Attributes map[string]string
}

var _ propagation.TextMapPropagator = AttributesPropagator{}

func (p AttributesPropagator) Inject(ctx context.Context, carrier propagation.TextMapCarrier) {
	for k, v := range p.Attributes {
		carrier.Set(k, v)
	}
}

func (p AttributesPropagator) Extract(ctx context.Context, carrier propagation.TextMapCarrier) context.Context {
	for _, k := range carrier.Keys() {
		if _, ok := p.Attributes[k]; ok {
			p.Attributes[k] = carrier.Get(k)
		}
	}
	return ctx
}

func (p AttributesPropagator) Fields() []string {
	var keys []string
	for k := range p.Attributes {
		keys = append(keys, k)
	}
	return keys
}
```

@pellared Do you think `AttributesPropagator` is a suitable name that we could put in Go contrib?

docs/database/database-spans.md
XSAM: Discussion about using a generic propagator to push a list of string key-values. https://github.com/open-telemetry/semantic-conventions/pull/2495#discussion_r2209133620

docs/database/database-spans.md
brettmc: > That means the propagator needs to access the OTel SDK

Contrib code can depend on the API, and if the TracerProvider API had an additional method to expose resource attributes, then the "globally configured tracer provider" could be used to retrieve this info? Likewise, [SpanContext](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.46.0/specification/trace/api.md#spancontext) is an API construct and so contrib code may use it. SpanContext is the key to context propagation, which is why I suggest making `service.name` available there, if it is to be used for a different form of context propagation.

docs/database/database-spans.md
XSAM: Discussion about having a generic propagator to push a list of string key-values. https://github.com/open-telemetry/semantic-conventions/pull/2495#discussion_r2209133620

docs/database/database-spans.md
XSAM: > if the TracerProvider API had an additional method to expose resource attributes, then the "globally configured tracer provider" could be used to retrieve this info?

> Likewise, [SpanContext](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.46.0/specification/trace/api.md#spancontext) is an API construct and so contrib code may use it

That means we need to change the spec of TracerProvider API or SpanContext. I am not sure about this, as

1. It is not just about trace, it is more about resource attributes. Using a trace signal is strange in this case.
2. There is an ongoing proposal of [EntityProvider API](https://github.com/open-telemetry/opentelemetry-specification/pull/4316/files#diff-34ec96a19fd0549b33086a024122b849b868c4e5532bb11d7336f02765f3ad40R146), which has `Get Resource` operation that we can use.

pellared: I think it can be also acceptable for an experimental service name propagator to accept Resource in its constructor. I think that it is a fair trade-off before EntityProvider API is available. In future, changing the dependency from Resource to EntityProvider should not be hard to deal with. 

Adding some attributes propagator can also work.

Whether these propagators can be hosted in Contrib repositories is a SIG decision.

docs/database/database-spans.md
XSAM: I have moved the custom propagator out of this PR, so we can focus on the SQL commenter itself without any add-ons.

docs/database/database-spans.md
XSAM: Per discussion on SIG, I have moved the custom propagator out of this PR, so we can focus on the SQL commenter itself without any add-ons.

I will create another PR introducing the Attributes (or TextAttributes) propagator after aligning with the spec SIG.

docs/database/database-spans.md
XSAM: Per semconv SIG, though it is hard to pick which way, it would be good to still have a default behavior to give instrumentation a specific direction.

For that, I prefer append, and it is also the format defined in the SQL commenter doc. https://google.github.io/sqlcommenter/spec/

PR has updated.

docs/database/database-spans.md
```diff
@@ -478,6 +480,34 @@ Semantic conventions for individual database systems or specialized instrumentat
 MAY specify a different `db.query.summary` format as long as produced summary remains
 relatively short and its cardinality remains low comparing to the `db.query.text`.
 
+## Context propagation
+
+**Status**: [Development][DocumentStatus]
+
+### SQL commenter
+
+Instrumentations MAY propagate context using [SQL commenter](https://google.github.io/sqlcommenter/spec/) by injecting comments into SQL queries before execution. Context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
```
trask: this creates an odd situation where we reference out to an external specification, and that external specification says it has been donated to OpenTelemetry: https://github.com/google/sqlcommenter#sqlcommenter

docs/database/database-spans.md
```diff
@@ -478,6 +480,34 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+If the propagator does not provide any value, the instrumentation SHOULD NOT inject any comment into the query.
+
+**Examples:**
+
+- For a query `SELECT * FROM songs` with W3C TraceContext propagator:
+
+  ```sql
+  SELECT * FROM songs /*traceparent='00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01',tracestate='congo%3Dt61rcWkgMzE%2Crojo%3D00f067aa0ba902b7'*/
+  ```
```
trask, line 501: I think we need to call out the performance implications of doing this: https://github.com/google/sqlcommenter/issues/284

docs/database/database-spans.md
XSAM: And, the OpenTelemetry repo says

> Please continue using https://github.com/google/sqlcommenter for now.

https://github.com/open-telemetry/opentelemetry-sqlcommenter

XSAM: Could we keep using `https://google.github.io/sqlcommenter/spec/`, as it at least presents the spec on the page while others don't?

The https://github.com/google/sqlcommenter#sqlcommenter also says

> We will continue to maintain this repository until opentelemetry SQLCommenter feature stabilizes.

docs/database/database-spans.md
```diff
@@ -478,6 +480,34 @@ Semantic conventions for individual database systems or specialized instrumentat
...
 
+## Context propagation
+
+**Status**: [Development][DocumentStatus]
+
+### SQL commenter
+
+Instrumentations MAY propagate context using [SQL commenter](https://google.github.io/sqlcommenter/spec/) by injecting comments into SQL queries before execution. Context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
+
+The instrumentation implementation SHOULD **append** the comment to the end of the query. Semantic conventions for individual database systems MAY specify different format depending on the specific database system's requirements or preferences. For instance, the instrumentation MAY **prepend** the comment at the beginning of the query.
```
trask: does the sqlcommenter spec not say anything about prepending or appending?

XSAM: From https://google.github.io/sqlcommenter/spec/, it says

> The final comment SHOULD be affixed to the final SQL statement in the format
> ```
> <SQL STATEMENT> /*<ATTRIBUTE_KEY_VALUE_PAIRS>*/
> ```

Which means appending.

docs/database/database-spans.md
XSAM: Updated on https://github.com/open-telemetry/semantic-conventions/pull/2495/commits/b9aabc88d7a401adc9d92ca25174bc8367fe443f

docs/database/database-spans.md
```diff
@@ -478,6 +480,36 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+**Examples:**
+
+- For a query `SELECT * FROM songs` with W3C TraceContext propagator:
+
+  ```sql
+  SELECT * FROM songs /*traceparent='00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01',tracestate='congo%3Dt61rcWkgMzE%2Crojo%3D00f067aa0ba902b7'*/
+  ```
+
+  Note that adding high cardinality comments, like `traceparent` and `tracestate`, to queries can impact the performance for some database systems, like MySQL. See [this related issue](https://github.com/google/sqlcommenter/issues/284) for more details.
```
cristiangreco: Hi! 👋 The current wording might sound a bit scary unless you actually open the linked issue. I'm wondering if maybe we could say more explicitly "like prepared statements in MySQL"?

nenadnoveljic: Severely affected databases are Oracle and SQL Server for both prepared and non-prepared statements.

XSAM: Updated on https://github.com/open-telemetry/semantic-conventions/pull/2495/commits/f6f5994392edbc4c213f83397e01f877fa544d8e

XSAM: Updated on https://github.com/open-telemetry/semantic-conventions/pull/2495/commits/f6f5994392edbc4c213f83397e01f877fa544d8e

XSAM:
@open-telemetry/semconv-db-approvers PTAL

docs/database/database-spans.md
```diff
@@ -478,6 +480,39 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+**Status**: [Development][DocumentStatus]
+
+### SQL commenter
+
+Instrumentations MAY propagate context using [SQL commenter](https://google.github.io/sqlcommenter/spec/) by injecting comments into SQL queries before execution. Context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
+
+The instrumentation implementation SHOULD **append** the comment to the end of the query. Semantic conventions for individual database systems MAY specify different format depending on the specific database system's requirements or preferences. For instance, the instrumentation MAY **prepend** the comment at the beginning of the query.
+
+The instrumentation SHOULD allow users to pass a propagator to overwrite the global propagator.
```
trask: FWIW, there's some prior art in Java Instrumentation of pairing a specific instrumentation to a specific propagator:

https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-java-instrumentation%20AwsXrayPropagator.getInstance()&type=code

Maybe it's ok to pair this behavior directly to the w3c propagator (only supporting w3c trace context)?

XSAM: Do you mean that instead of using the global propagator by default, we should use the `W3C propagator` as the default one?

I don't have a preference here. What makes the `W3C propagator` the default, a better choice than the global propagator?

trask: Good point

XSAM:
@trask, I find all comments are resolved. Is that okay for you to approve this PR?

docs/database/database-spans.md
```diff
@@ -478,6 +480,39 @@ Semantic conventions for individual database systems or specialized instrumentat
 MAY specify a different `db.query.summary` format as long as produced summary remains
 relatively short and its cardinality remains low comparing to the `db.query.text`.
 
+## Context propagation
+
+**Status**: [Development][DocumentStatus]
+
+### SQL commenter
+
+Instrumentations MAY propagate context using [SQL commenter](https://google.github.io/sqlcommenter/spec/) by injecting comments into SQL queries before execution. Context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
```
lmolkova: nit:

```suggestion
Instrumentations MAY propagate context using [SQL commenter](https://google.github.io/sqlcommenter/spec/) by injecting comments into SQL queries before execution. SQL commenter-based context propagation SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
```

it should be fine to propagate context using other means if database supports it (e.g. db uses HTTP as a transport)

docs/database/database-spans.md
```diff
@@ -478,6 +480,39 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+The instrumentation implementation SHOULD **append** the comment to the end of the query. Semantic conventions for individual database systems MAY specify different format depending on the specific database system's requirements or preferences. For instance, the instrumentation MAY **prepend** the comment at the beginning of the query.
```
lmolkova:
```suggestion
The instrumentation implementation SHOULD **append** the comment to the end of the query. Semantic conventions for individual database systems MAY specify different format depending on the specific database system's requirements or preferences.
```

it feels confusing to first tell to append, then tells that it's overridable and then allow to prepend. I feel the last sentence is not really necessary, is it?

docs/database/database-spans.md
```diff
@@ -478,6 +480,39 @@ Semantic conventions for individual database systems or specialized instrumentat
...
+
+The instrumentation SHOULD allow users to pass a propagator to overwrite the global propagator.
+
+If the propagator does not provide any value, the instrumentation SHOULD NOT inject any comment into the query.
```
lmolkova:
```suggestion
If the propagator is not provided, the instrumentation SHOULD NOT inject any comment into the query.
```
?

lmolkova (approved):

docs/database/database-spans.md
XSAM: This prevents injecting empty comments like `/**/` when the propagator sets nothing.

So, we won’t see a query like this `SELECT * FROM Users /**/`

lmolkova: Perhaps you meant something like this then

```suggestion
If the context is empty or invalid, the propagator SHOULD NOT inject any comment into the query.
```

? 

But this sentence immediately follows "The instrumentation SHOULD allow users to pass a propagator to overwrite the global propagator.", it seems it intends to talk about propagator not being provided by the user.

docs/database/database-spans.md
XSAM:
```suggestion
The instrumentation implementation SHOULD **append** the comment to the end of the query. Semantic conventions for individual database systems MAY specify different format depending on the specific database system's requirements or preferences. For instance, the instrumentation could **prepend** the comment at the beginning of the query as an alternative format.
```

You are right. But maybe we could alter the sentence a little bit to make it clear, so we could keep an example of what different format could look like. WDYT?

lmolkova: I really don't understand why it's necessary, but if you feel it's important to keep it, may this would be less confusing?

> Semantic conventions for individual database systems MAY specify different format, which may include different position, encoding, or schema, depending on the specific database system's requirements or preferences.

docs/database/database-spans.md
XSAM: We can use this code as the example. In this example, even the propagator provides nothing, the comments will be injected.

https://github.com/XSAM/otelsql/blob/370716b2f8c317607879035937ba65cfc751dc5a/commenter.go#L66

```
return fmt.Sprintf("%s /*%s*/", query, cc.Marshal())
```

Since it is instrumentation's job to actually inject the comment, the instrumentation should be the one to be regulated for this, not the propagator.

docs/database/database-spans.md
XSAM: Yes, this would be better. Updated on https://github.com/open-telemetry/semantic-conventions/pull/2495/commits/c0a6f4b38a0794b87dac4c41b74649e0b2304f01

docs/database/database-spans.md
XSAM: I updated this part on https://github.com/open-telemetry/semantic-conventions/pull/2495/commits/c0a6f4b38a0794b87dac4c41b74649e0b2304f01. I think it is clear now.

XSAM: @lmolkova Is this good now?

lmolkova: sorry, I still don't get it - propagator injects the context, not the instrumentation. Propagators check if the context is valid before attempting to inject it. So I'd still suggest to change to remove this sentence or change it to.

```diff
- If the propagator produces no output (empty context), the instrumentation SHOULD NOT inject an empty comment into the query.
+ Propagator SHOULD NOT inject invalid/empty context.
```

XSAM: @lmolkova Let me elaborate on this. The propagator injects data from the context to the carrier during the `inject` call. This is the method we will use during the SQL comments injection. And, **the propagator does not inject any context during this.**

After the data is transferred to the carrier, the instrumentation can use the data from the carrier to inject comments into SQL. During this injection, the instrumentation might just use a simple approach to put carrier data in a string `/* <carrier-data> */`. When the `<carrier-data>` is empty, then we get `/* */`, which is not ideal, as empty comments do not have meaningful meaning. The ideal output of this is that the instrumentation should skip the SQL injection under such a situation and leave the SQL untouched.

I think we can update this sentence to this.

```
The instrumentation SHOULD NOT inject an empty comment into the query.
```

This does not mention the propagator. I hope this can reduce the confusion.

lmolkova: Is there a propagator that attempts to inject invalid context into a carrier? The propagator implementations I'm aware of don't do it.

E.g. [python](https://github.com/open-telemetry/opentelemetry-python/blob/b06cf803605dd323ad134a9a434a204bf7fb150f/opentelemetry-api/src/opentelemetry/trace/propagation/tracecontext.py#L101), [java](https://github.com/open-telemetry/opentelemetry-java/blob/f984d201802a2417df707d87329cf3bb601be569/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java#L102), [dotnet](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Context/Propagation/TraceContextPropagator.cs#L98). Baggage propagators follow the same principle of not injecting empty/invalid stuff - not even passing it to the setter. So instrumentation would never even see the setter called.

I could not find language in the spec that would tell them not to, but the should be a feature of the propagator - the common piece to filter out invalid context. If we need to explicitly document it, let's document it in the spec and for all propagators.

XSAM: > Is there a propagator that attempts to inject invalid context into a carrier?

My point is more like the propagator may not doing the injecting at all, instead of injecting empty stuff.

Users might implement their own propagator, not calling the setter at all is allowed.

> So instrumentation would never even see the setter called.

Exactly. And when this happens, the instrumentation may produce empty comment like I demonstrated in my previous comment.

**In short, what I was trying to do is prevent instrumentation injecting useless (empty) comment to sql query, we shouldn’t concern about how it is triggered.** It can be triggered by any component during the injecting, like propagator, carrier, and the instrument itself.

lmolkova: what I'm concerned about

- instrumentation should not inject context without a propagator
- propagator implementation should be responsible to validate context before injecting it and not inject empty things
- carrier should assume it's not called unless there is a context to propagate
- having it all documented in the right place for all propagators, not just DB SQL commenter ones.

If some custom propagator does not abide by these rules and injects invalid context, that's on them and it would be a small inefficiency. We should not ask every single instrumentation in the world to be mindful of that poorly implemented propagator and make a perf-impact (however small) on everybody.

XSAM: > instrumentation should not inject context without a propagator

Yes.

> propagator implementation should be responsible to validate context before injecting it and not inject empty things

Yes, but keep in mind that not injecting anything is common for a propagator. Take baggage as an example, if there is no data to inject, the baggage propagator does nothing.

https://github.com/open-telemetry/opentelemetry-go/blob/22c6da030791487544f85c8d002263f5dd4d8ee8/propagation/baggage.go#L23-L28

It won't even call the carrier

> carrier should assume it's not called unless there is a context to propagate

Yes.

> having it all documented in the right place for all propagators, not just DB SQL commenter ones.

I am not sure what should be documented.

> If some custom propagator does not abide by these rules and injects invalid context

I was not trying to resolve the situation where the propagator injects invalid context (I consider doing nothing from the propagator as not an invalid behavior)

Take baggage propagator as an example, when calling the injecting method:

- the`bStr` in the implementation can be empty if there is no baggage to propagate. https://github.com/open-telemetry/opentelemetry-go/blob/22c6da030791487544f85c8d002263f5dd4d8ee8/propagation/baggage.go#L24
- Then, it won't call the setter from the carrier.
- So, we have an empty carrier. And, this is a pretty common situation.
- In this case, the carrier in this code will be empty https://github.com/XSAM/otelsql/blob/370716b2f8c317607879035937ba65cfc751dc5a/commenter.go#L60
- Then, in this code https://github.com/XSAM/otelsql/blob/370716b2f8c317607879035937ba65cfc751dc5a/commenter.go#L66
  - `cc.Marshal()` (carrier to string) gonna produce empty string
  - And, this line `fmt.Sprintf("%s /*%s*/", query, cc.Marshal())` would make the SQL like `SELECT * FROM Users` become `SELECT * FROM Users /* */`, because `fmt.Sprintf` will always inject comments into the SQL even if there is no content to inject. What I am trying to tell the author of the instrumentation is, in this case, the instrumentation should skip injecting, so we can get `SELECT * FROM Users` without being touched.

@lmolkova Do you think the spec can enforce this requirement to prevent injecting `/* */` in the SQL? If so, how am I supposed to write to meet the writing style in the spec for this enforcement? If not, I will consider this to be out of the scope of the enforcement of the spec and remove this line.

lmolkova: The standard propagator behavior is to actually inject things into the carrier and the carrier to be something that goes over the wire, not for carrier to be intermediary property bag.

E.g. if you look at [HTTP go](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/bdba5c83b829a938a76d41aed6ef47427cbe1f47/instrumentation/net/http/otelhttp/transport.go#L129) - there is no additional step that copies the carrier over. It's very surprising to ask instrumentation to actually serialize carrier.

Converting it to sql commenter, I would expect the setter to take care of `fmt.Sprintf("%s /*%s*/", query, cc.Marshal())` or a special SQL commenter propagator to do it, but not ask all the instrumentations to do it.

XSAM: This line is removed and updated on https://github.com/open-telemetry/semantic-conventions/pull/2495/commits/a1d1d79471b674e8f18a99bf58d085145feba1db.

XSAM:
Not sure why the CLA bot is complaining about a non-existent account. I guess I will need to rebase the branch to address this.

XSAM:
Sorry for accidentally triggering so many reviewing groups... 🥲

XSAM:
@open-telemetry/specs-semconv-maintainers, all comments are resolved. Could we merge this PR?

PR #2499, opened 2025-07-09
Better link checking

trask:
This will hopefully help a lot with all the sporadic link failures on PRs which are confusing to contributors.

Link checks on PR are now done in two passes:

- Check all links in the modified files
- Check local links across all files (this will catch if the PR removes an anchor or file that another file in this repo was linking to)

In order to catch external links that randomly go poof, there's now a daily job that will create an issue if there's a failure (and will autoclose the daily job if it passes the next day).


.github/scripts/link-check.sh
trask, line 1: this is almost a direct copy from Java Contrib, maybe we can centralize in the future, but first step is to try it out in repo where we have lots of problems and see if it helps

.github/workflows/reusable-link-check.yml
trask, line 1: same

.github/workflows/reusable-workflow-notification.yml
trask, line 1: same

laurit (approved):

jsuereth (approved):

lmolkova (approved):
Thank you!

PR #2501, opened 2025-07-09
Move how-to-contribute information into its own directory.

jsuereth:
Moves all "how to contribute" information into the same directory in anticipation of fleshing out more guides on providing semantic conventions.

- Move `how-to-contribute.md` from "General" semantic conventions into its own directory.
- Move Entity & Resource modeling guide into the new directory.


thompson-tomo:
@jsuereth would it be possible to have this folder to be located under the [contributing/](https://opentelemetry.io/docs/contributing/) folder on the website? That way contributing is grouped together.

jsuereth:
> @jsuereth would it be possible to have this folder to be located under the [contributing/](https://opentelemetry.io/docs/contributing/) folder on the website? That way contributing is grouped together.

Yes we coudl move this contributing

arminru (approved):

jsuereth:
@thompson-tomo I looked into CONTRIBUTING.md and github recommendations.

Technically, they recommend a `contributing.MD` file not a directory and either a `docs/` directory or using `.github`.  `contributing/` as a directory isn't really a convention.

Additionally, I think `how-to-write-contribute` directory, for now, is more attention grabbing, as we anticipate relying on links from `CONTRIBUTING.md` anyway.

thompson-tomo:
@jsuereth apologies it appears that it wae missed that the website I was referring to was https://opentelemetry.io/ (link was included). Hencr I was meaning if we could have this folder of docs within the contributing folder on the website rather than under specs. 

Note the link I provided https://opentelemetry.io/docs/contributing/

jsuereth:
> @jsuereth apologies it appears that it wae missed that the website I was referring to was https://opentelemetry.io/ (link was included). Hencr I was meaning if we could have this folder of docs within the contributing folder on the website rather than under specs.
> 
> Note the link I provided https://opentelemetry.io/docs/contributing/

Sorry for misunderstanding.  Yes - we should figure out how to update website docs.  That'll need to be a separate task, as we want the ownership of the contributing to be in this repo, but the exposure of the documents in the website.  We can open a separate task.

Issue #2228, opened 2025-05-05
JSON RPC: define attributes on invalid requests


louis-cl:
The [jsonrpc conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/json-rpc.md) require the presence of `rpc.method`.

In some invalid request scenarios, it is impossible to determine the method trying to be invoked.
Like from the [jsonrpc specification examples](https://www.jsonrpc.org/specification#examples):
```
--> {"jsonrpc": "2.0", "method": "foobar, "params": "bar", "baz]
<-- {"jsonrpc": "2.0", "error": {"code": -32700, "message": "Parse error"}, "id": null}
```

**Should there be a convention for the telemetry in this scenario?**

I can imagine a few options:
1. Don't generate metrics for this scenario
2. Break convention and don't send `rpc.method`
3. Send a special value for `rpc.method` like `""` 
4. Make `rpc.method` conditionally required

I see value in counting how many jsonrpc errors with code `-32700` my server returns.
I would be erring on option 3 to respect the convention.

**Extra thoughts on option 3**

Why `""` ? the jsonspec requires a null on these errors:
> If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.

And the semantic convention maps it to empty string
> Use empty string in case of null value

It felt reasonable to follow the same results for when rpc.method can't be detected.

thompson-tomo:
@louis-cl why wouldn't your method be foobar? There is nothing specifying that it needs to be a valid in fact using an invalid one would be useful as you could see which methods are generating errors.

louis-cl:
> [@louis-cl](https://github.com/louis-cl) why wouldn't your method be foobar? There is nothing specifying that it needs to be a valid in fact using an invalid one would be useful as you could see which methods are generating errors.

The JSON request is invalid, notice the lack of `"` after foobar or the missing `}`.
If I try the request part from the example I quote `{"jsonrpc": "2.0", "method": "foobar, "params": "bar", "baz]`  in [jsonlint.com](https://jsonlint.com/) I get this error:

```
Error: Parse error on line 1:
..."method": "foobar, "params": "bar", "baz
-----------------------^
Expecting 'EOF', '}', ':', ',', ']', got 'undefined'
```
Detecting `foobar` as method would require some smart heuristics in parsing.
Maybe the scenario is clearer with this other invalid request: `{"jsonrpc": "2.0", "bar", "baz]`

[...]

Issue #2063, opened 2025-04-08
JSON RPC: deprecate duplicate attributes


lmolkova:
The following [JSON RPC attributes](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/json-rpc.md) look like a duplicates of common properties:

- `rpc.jsonrpc.error_message` - should be reported as a span status description or `error.message` on logs/events.
- `rpc.jsonrpc.version` - should be reported as `network.protocol.version`

thompson-tomo:
I would also put rpc.jsonrpc.error_code into that category by using error.type for the error code.

lmolkova:
related to https://github.com/open-telemetry/semantic-conventions/issues/2967

lmolkova:
Being addressed in https://github.com/open-telemetry/semantic-conventions/pull/2920

PR #2503, opened 2025-07-10
explicit json rpc span definitions #2228

### LINKED ISSUE #2762: Remove rpc.system attribute from rpc framework pages
Since they only every use a single value which is specified at top of page already.

thompson-tomo:
Related to #2228
Closes #2762

## Changes

This rewrites the jsonrpc page to provide the span definitions rather than just listing the attributes as requested in #2720

This also defines jsonrpc as a system which had been documented but wasn't part of the model.


jsuereth:
We should hold on RPC related changes until the RPC semconv SIG is formed.

thompson-tomo:
As this came about due to a bug, what can be done to progress it? Would changing the rpc.method conditionally required message to indicate only for json-rpc & put back rpc.service allow this to be progressed sooner?

model/rpc/spans.yaml
```diff
@@ -87,11 +87,24 @@ groups:
...
-  - id: rpc.jsonrpc.attributes
-    type: attribute_group
+  - id: span.rpc.jsonrpc.client
+    type: span
     stability: development
-    brief: 'Tech-specific attributes for [JSON RPC](https://www.jsonrpc.org/).'
+    brief: This span represents an outgoing Remote Procedure Call (RPC).
+    note: |
+      Remote procedure calls can only be represented with these semantic conventions
+      when the names of the called service and method are known and available.
```
KalleOlaviNiemitalo: IMO the name of the called service should not be required for JSON-RPC, which has no such thing in the protocol.  Rather, the span name should be specified such that the service name can be omitted or replaced with some placeholder.

Even if both the JSON-RPC server and the JSON-RPC client set `rpc.service` attributes in their telemetry, there is no reason to expect them to be using the same service name.

.chloggen/rpc-json_rpc_span_def.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: rpc
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: JsonRPC now has it's own span definition.
```
KalleOlaviNiemitalo: "JSON-RPC", "its".

docs/rpc/json-rpc.md
```diff
@@ -10,26 +10,198 @@ The Semantic Conventions for [JSON-RPC](https://www.jsonrpc.org/) extend and ove
...
+`rpc.system` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
 
-**[1] `rpc.method`:** This is always required for jsonrpc. See the note in the general RPC conventions for more information.
+| Value  | Description | Stability |
+|---|---|---|
+| `apache_dubbo` | Apache Dubbo | ![Development](https://img.shields.io/badge/-development-blue) |
+| `connect_rpc` | Connect RPC | ![Development](https://img.shields.io/badge/-development-blue) |
+| `dotnet_wcf` | .NET WCF | ![Development](https://img.shields.io/badge/-development-blue) |
+| `grpc` | gRPC | ![Development](https://img.shields.io/badge/-development-blue) |
+| `java_rmi` | Java RMI | ![Development](https://img.shields.io/badge/-development-blue) |
```
KalleOlaviNiemitalo: `jsonrpc` itself is missing from this table.  And I don't think the table is even necessary on the page that is specifically for JSON-RPC.

model/rpc/spans.yaml
thompson-tomo: This is a pure copy paste of the base span with the only change being the system name. Suggest discussing this via seperate issue.

docs/rpc/json-rpc.md
thompson-tomo: Good spot, will add it. 

I have already raised https://github.com/open-telemetry/semantic-conventions/pull/2739 for a proposal to auto hide the table.

thompson-tomo: System has been added to the enum

model/rpc/spans.yaml
```diff
@@ -102,11 +117,37 @@ groups:
...
+
+      `rpc.system` MUST be set to `"jsonrpc"` and SHOULD be provided **at span creation time.**
+
+      **Span name:** refer to the [Span Name](./rpc-spans.md#span-name) section.
+
+      **Span kind** MUST be `SERVER`.
+    attributes:
+      - ref: rpc.system
+        examples: jsonrpc
```
trask: let's follow database semconv for now and hide this

trask: similar to https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/mysql.md

(for handling sampling relevant question)

model/rpc/spans.yaml
lmolkova: rpc.service does not make sense for JSON RPC, so we need to do yaml refactoring allowing to omit it here.

model/rpc/spans.yaml
```diff
@@ -102,11 +75,35 @@ groups:
...
-          This is always required for jsonrpc. See the note in the general
-          RPC conventions for more information.
+      - ref: rpc.jsonrpc.version
+        requirement_level:
+          conditionally_required: If other than the default version (`1.0`)
+      - ref: rpc.jsonrpc.request_id
+        requirement_level: recommended
+      - ref: rpc.jsonrpc.error_code
+        requirement_level:
+          conditionally_required: If response is not successful.
```
lmolkova: why so? if response code is available, we should set it. `error.type` is used when response is not successful

```suggestion
          conditionally_required: when available
```

model/rpc/spans.yaml
```diff
@@ -102,11 +75,35 @@ groups:
...
+
+  - id: span.rpc.jsonrpc.server
+    type: span
+    stability: development
+    extends: rpc.server
+    span_kind: server
+    brief: This span represents an incoming Remote Procedure Call (RPC).
+    note: |
+      Remote procedure calls can only be represented with these semantic conventions
+      when the names of the called service and method are known and available.
```
lmolkova: this is redundant, right? by saying that `rpc.method` is required, we're also saying that these conventions can only be used when it's applicable.

model/rpc/spans.yaml
thompson-tomo: So this is what the existing message was see https://github.com/thompson-tomo/semantic-conventions/blob/fix/%232228_MethodRequirement/model/rpc/spans.yaml#L72 aka the client span

model/rpc/spans.yaml
thompson-tomo: would that not be considered a breaking change given a json-rpc reader is currently told

> The Semantic Conventions for [JSON-RPC](https://www.jsonrpc.org/) extend and override the [RPC spans](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/) 

And is then presented a list of attributes which are overriden.

hence i focussed on not making breaking changes in this pr. Perhaps we could make it opt-in for json-rpc that way non-breaking and do a bigger refactor with all the breaking changes in a next version after the stability note is changed.

model/rpc/spans.yaml
thompson-tomo: that is true and stems from the original description of the span which had this. Hence have removed it from json-rpc and will raise issue (#2765) to discuss the general page.

model/rpc/spans.yaml
thompson-tomo: We also have #2063 to tackle deprecating the duplicates.

lmolkova: there is no need to copy it over if it does not make sense

model/rpc/spans.yaml
lmolkova: we're going o make tons of breaking changes with rpc semconv. they are underspecified now. no need to make things worse by specifying wrong thing

model/rpc/spans.yaml
thompson-tomo: Ok have updated both instances of the message.

model/rpc/spans.yaml
thompson-tomo: Ok change is made.

.chloggen/rpc-json_rpc_span_def.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: rpc
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: JSON-RPC now has its own span definition's.
```
trask:
```suggestion
note: JSON-RPC now has its own span definition.
```

trask (approved):

thompson-tomo:
@lmolkova could you take a look at this & let me know if everything is resolved and can hopefully be merged soon.

lmolkova (approved):

PR #2504, opened 2025-07-10
add known methods for declarative configuration

zeitlinger:
Relates https://github.com/open-telemetry/opentelemetry-java/issues/7468

### DIFF
--- a/docs/http/http-spans.md
+++ b/docs/http/http-spans.md
@@ -179,7 +179,8 @@ If the HTTP request method is not known to instrumentation, it MUST set the `htt
 If the HTTP instrumentation could end up converting valid HTTP request methods to `_OTHER`, then it MUST provide a way to override
 the list of known HTTP methods. If this override is done via environment variable, then the environment variable MUST be named
 OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS and support a comma-separated list of case-sensitive known HTTP methods
-(this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults).
+If this override is done via declarative configuration, then the list key MUST be named `known_methods` (under `general`/`http`).
+In either case, this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults.
 
 HTTP method names are case-sensitive and `http.request.method` attribute value MUST match a known HTTP method name exactly.
 Instrumentations for specific web frameworks that consider HTTP methods to be case insensitive, SHOULD populate a canonical equivalent.
@@ -480,7 +481,8 @@ If the HTTP request method is not known to instrumentation, it MUST set the `htt
 If the HTTP instrumentation could end up converting valid HTTP request methods to `_OTHER`, then it MUST provide a way to override
 the list of known HTTP methods. If this override is done via environment variable, then the environment variable MUST be named
 OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS and support a comma-separated list of case-sensitive known HTTP methods
-(this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults).
+If this override is done via declarative configuration, then the list key MUST be named `known_methods` (under `general`/`http`).
+In either case, this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults.
 
 HTTP method names are case-sensitive and `http.request.method` attribute value MUST match a known HTTP method name exactly.
 Instrumentations for specific web frameworks that consider HTTP methods to be case insensitive, SHOULD populate a canonical equivalent.
--- a/docs/registry/attributes/http.md
+++ b/docs/registry/attributes/http.md
@@ -51,7 +51,8 @@ If the HTTP request method is not known to instrumentation, it MUST set the `htt
 If the HTTP instrumentation could end up converting valid HTTP request methods to `_OTHER`, then it MUST provide a way to override
 the list of known HTTP methods. If this override is done via environment variable, then the environment variable MUST be named
 OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS and support a comma-separated list of case-sensitive known HTTP methods
-(this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults).
+If this override is done via declarative configuration, then the list key MUST be named `known_methods` (under `general`/`http`).
+In either case, this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults.
 
 HTTP method names are case-sensitive and `http.request.method` attribute value MUST match a known HTTP method name exactly.
 Instrumentations for specific web frameworks that consider HTTP methods to be case insensitive, SHOULD populate a canonical equivalent.
--- a/model/http/registry.yaml
+++ b/model/http/registry.yaml
@@ -92,7 +92,8 @@ groups:
           If the HTTP instrumentation could end up converting valid HTTP request methods to `_OTHER`, then it MUST provide a way to override
           the list of known HTTP methods. If this override is done via environment variable, then the environment variable MUST be named
           OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS and support a comma-separated list of case-sensitive known HTTP methods
-          (this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults).
+          If this override is done via declarative configuration, then the list key MUST be named `known_methods` (under `general`/`http`).
+          In either case, this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults.
 
           HTTP method names are case-sensitive and `http.request.method` attribute value MUST match a known HTTP method name exactly.
           Instrumentations for specific web frameworks that consider HTTP methods to be case insensitive, SHOULD populate a canonical equivalent.
--- a/docs/database/elasticsearch.md
+++ b/docs/database/elasticsearch.md
@@ -66,7 +66,8 @@ If the HTTP request method is not known to instrumentation, it MUST set the `htt
 If the HTTP instrumentation could end up converting valid HTTP request methods to `_OTHER`, then it MUST provide a way to override
 the list of known HTTP methods. If this override is done via environment variable, then the environment variable MUST be named
 OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS and support a comma-separated list of case-sensitive known HTTP methods
-(this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults).
+If this override is done via declarative configuration, then the list key MUST be named `known_methods` (under `general`/`http`).
+In either case, this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults.
 
 HTTP method names are case-sensitive and `http.request.method` attribute value MUST match a known HTTP method name exactly.
 Instrumentations for specific web frameworks that consider HTTP methods to be case insensitive, SHOULD populate a canonical equivalent.
--- a/docs/http/http-metrics.md
+++ b/docs/http/http-metrics.md
@@ -100,7 +100,8 @@ If the HTTP request method is not known to instrumentation, it MUST set the `htt
 If the HTTP instrumentation could end up converting valid HTTP request methods to `_OTHER`, then it MUST provide a way to override
 the list of known HTTP methods. If this override is done via environment variable, then the environment variable MUST be named
 OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS and support a comma-separated list of case-sensitive known HTTP methods
-(this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults).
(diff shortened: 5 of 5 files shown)


zeitlinger:
@jack-berg I didn't see any mention of other declarative configuration properties in this repo. Should they be added?

jack-berg:
> @jack-berg I didn't see any mention of other declarative configuration properties in this repo. Should they be added?

This is an unsolved problem. For now, I did a one time analysis of the semantic conventions repo and modeled of the concepts that are supposed to be configurable as properties in the declarative configuration schema: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/instrumentation.json#L50-L128

@lmolkova has previously expressed the idea that the `semantic-conventions` repo itself could somehow directly model the configuration schema. Then, `opentelemetry-configuration` could take a dependency on `semantic-conventions` and translate / merge the configuration schema into its JSON schema. 

I like this idea but for now its just an idea.

zeitlinger:
> that the `semantic-conventions` repo itself could somehow directly model the configuration schema. Then, `opentelemetry-configuration` could take a dependency on `semantic-conventions` and translate / merge the configuration schema into its JSON schema.

I also like that.
I'm wondering if we should continue with the current approach until we've figured it out - to not create a split setup.

lmolkova:
related: https://github.com/open-telemetry/semantic-conventions/issues/1450

How I see it:
- when opt-in metric/attribute is defined, there should be a way to express which config option it's controlled with. this option should be the same across instrumentations
- it should be possible to enable/disable a tracer/meter/logger (or even span/metric/event) based on a stable language-agnostic property that goes over the wire - semconv and config should work together
- stability opt-in env var should be a config option
- ...

I don't believe anyone is working on it, but I would be exited to see any detailed suggestions/proposals in this space.

docs/registry/attributes/http.md
```diff
@@ -51,7 +51,8 @@ If the HTTP request method is not known to instrumentation, it MUST set the `htt
 If the HTTP instrumentation could end up converting valid HTTP request methods to `_OTHER`, then it MUST provide a way to override
 the list of known HTTP methods. If this override is done via environment variable, then the environment variable MUST be named
 OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS and support a comma-separated list of case-sensitive known HTTP methods
-(this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults).
+If this override is done via declarative configuration, then the list key MUST be named `known_methods` (under `general`/`http`).
```
lmolkova: I think we should include [`instrumentation/development`](https://github.com/open-telemetry/opentelemetry-configuration/blob/d6b127882f41db92b26017c736a52ceb6f2b4451/examples/kitchen-sink.yaml#L855) here? 

plus given it's in development, but appears on the stable attribute we need to express that the name is in development.

E.g. like here https://github.com/open-telemetry/semantic-conventions/blob/463b7f5a3d173bd6d0677834a9ef17da1dac7b2d/docs/http/http-spans.md?plain=1#L115

lmolkova:
@jack-berg what are your thoughts on semconv eventually owning the instrumentation config schema - it'd be difficult to write and implement conventions otherwise having config defined separately.

Our long-term vision is that codegen will get smarter and would generate whole spans/events/etc according to configuration, so we'd need to define/refer to config properties formally here in semconv and would interact with config API to retrieve values.

jack-berg:
> @jack-berg what are your thoughts on semconv eventually owning the instrumentation config schema - it'd be difficult to write and implement conventions otherwise having config defined separately.

I'm supportive. We can find a way to make `opentelemetry-configuration` depend on `semantic-conventions` and transform / merge the schema.

zeitlinger:
I propose we make this transformation separate from adding the property now - to avoid delaying the feature.

I've created https://github.com/open-telemetry/opentelemetry-configuration/pull/244

docs/registry/attributes/peer.md
```diff
@@ -9,4 +9,16 @@ Operations that access some remote service.
...
+
+```yaml instrumentation/development:
+  general:
+    peer:
+      service_mapping:
+        - peer: 1.2.3.4
+          service: FooService
+        - peer: 2.3.4.5
+          service: BarService
+```
```
KalleOlaviNiemitalo: This looks like there should be a line break between ```` ```yaml ```` and `instrumentation/development:`.

model/peer/registry.yaml
```diff
@@ -13,3 +13,17 @@ groups:
...
+          ```yaml
+          instrumentation/development:
+            general:
+              peer:
+                service_mapping:
+                  - peer: 1.2.3.4
+                    service: FooService
+                  - peer: 2.3.4.5
+                    service: BarService
+          ```
```
KalleOlaviNiemitalo: Here in the YAML source, you do have a line break between ```` ```yaml ```` and `instrumentation/development:`.  Does the YAML parser swallow it, then?

docs/registry/attributes/http.md
zeitlinger: I'll do "(under `instrumentation`/development`/`general`/`http`)."

How does that sound?

model/peer/registry.yaml
zeitlinger: looks good in the generated files

zeitlinger: fixed

docs/http/http-spans.md
```diff
@@ -644,6 +649,38 @@ and SHOULD be provided **at span creation time** (if provided at all):
 
 `http.route` MUST be provided at span creation time if and only if it's already available. If it becomes available after span starts, instrumentation MUST populate it anytime before span ends.
 
+## Capturing HTTP headers in traces
+
```
lmolkova, line 653:
```suggestion
**Status:** ![Development](https://img.shields.io/badge/-development-blue)

```

model/peer/registry.yaml
```diff
@@ -13,3 +13,18 @@ groups:
           of the remote service. SHOULD be equal to the actual `service.name`
           resource attribute of the remote service if any.
         examples: "AuthTokenCache"
+        note: >
+          It MUST be possible to configure the peer service name in declarative configuration as follows:
```
lmolkova: do we know which component would be responsible to use this configuration? This requirement should be actionable for this component. E.g. specialized span or log processor would match `network.protocol.address` to `peer` and populate `peer.service` on all spans based on the `service`.

In the current form, this requirement is only actionable for declarative config schema.

lmolkova: please add  `Development` status for this note

docs/http/http-spans.md
```diff
@@ -644,6 +649,38 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+In addition to the attributes defined in this document, HTTP client and server instrumentations MUST also allow
```
lmolkova: I don't think we can use MUST here

1. there are dozens of instrumentations out there that would become non-conformant.
2. we never require configurability and we can't until it's stable and adopted by language SDKs

We already define request and response attributes - search for `http.request | response.header.<key>` and we say that instrumentations MAY provide config to capture them. I don't think we can be stronger than that.

lmolkova (commented):
Thanks! I left some comments. A few additional thoughts:

- this is a first precedent documenting declarative config in semantic conventions, which I support
- we need to find way to record config options formally in YAML schema at least in some cases, but it's not trivial and we need to collect more precedents to start formalizing it. I don't want to block this PR on that. But we should leave some TODOs around options being introduced and link https://github.com/open-telemetry/semantic-conventions/issues/1450 as tracking issue
- I'd like to hear what other @open-telemetry/specs-semconv-maintainers think. @zeitlinger could you please bring this topic up on the next semconv call? Thanks!

docs/registry/attributes/peer.md
```diff
@@ -9,4 +9,17 @@ Operations that access some remote service.
...
+
+```yaml
+instrumentation/development:
+  general:
+    peer:
+      service_mapping:
+        - peer: 1.2.3.4
+          service: FooService
+        - peer: 2.3.4.5
+          service: BarService
```
lmolkova, line 29: is this format defined in the declarative config? I'd use exact attribute names instead.

```suggestion
        - network.peer.address: 1.2.3.4
          peer.service: FooService
        - network.peer.address: 2.3.4.5
          peer.service: BarService
```

but also, it could be much more generic  and allow to do things like "if attribute A matches value B, then set attribute C with value D"

docs/registry/attributes/peer.md
```diff
@@ -9,4 +9,17 @@ Operations that access some remote service.
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
-| <a id="peer-service" href="#peer-service">`peer.service`</a> | string | The [`service.name`](/docs/resource/README.md#service) of the remote service. SHOULD be equal to the actual `service.name` resource attribute of the remote service if any. | `AuthTokenCache` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="peer-service" href="#peer-service">`peer.service`</a> | string | The [`service.name`](/docs/resource/README.md#service) of the remote service. SHOULD be equal to the actual `service.name` resource attribute of the remote service if any. [1] | `AuthTokenCache` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[1] `peer.service`:** It MUST be possible to configure the peer service name in declarative configuration as follows:
+
+```yaml
+instrumentation/development:
```
lmolkova, line 22: why would it be an instrumentation responsibility at all BTW? There are a lot of HTTP / RPC/ etc instrumentations producing spans that may benefit from `peer.service`.

This enrichment can be done in the single processor (collector or per-language) and dozens of instrumentations would never bother.

model/peer/registry.yaml
trask: I like the wording for known_methods, something like 

> If this is done via declarative configuration, then then it MUST

model/http/registry.yaml
```diff
@@ -92,7 +92,9 @@ groups:
           If the HTTP instrumentation could end up converting valid HTTP request methods to `_OTHER`, then it MUST provide a way to override
           the list of known HTTP methods. If this override is done via environment variable, then the environment variable MUST be named
           OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS and support a comma-separated list of case-sensitive known HTTP methods
-          (this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults).
+          If this override is done via declarative configuration, then the list key MUST be named `known_methods`
+          (under `instrumentation`/development`/`general`/`http`).
+          In either case, this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults.
```
trask: should we split `known_methods` by client and server?

docs/registry/attributes/peer.md
zeitlinger: just added the following description:

          - `peer` is host, and optionally the and path of the remote service that is used to identify the service 
             instance, e.g., `example.com:443`, `1.2.3.4:8080`, `example.com`, `example.com:443/fooService`.
          - `service` is the name of the remote service, e.g., `AuthService` that will be translated to the
            `peer.service` attribute.

1. `network.peer.address` would be wrong, because it can contain the port and path as well
2. declarative config uses `_` style 
3. `it could be much more generic` 
  - this has to be implemented in every SDK, so we try to limit such complex features
  - otel collector has an attributes processor where you can already do this if needed
4. `peer_service` would still be an option

model/peer/registry.yaml
zeitlinger: 1. yes, it should only apply to declarative config, because this is a use case that is too complicated for env vars
2. the responsible component is typically a shared library in the instrumentation part
3. specialized log processor: that does not work because you typically don't have access to the URL for HTTP [server duration](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration)

zeitlinger: the whole attribute has `Development` stability

docs/registry/attributes/peer.md
zeitlinger: the SDK does not have access to the URL - see above

model/http/registry.yaml
zeitlinger: good idea

zeitlinger:
@trask @lmolkova please check again :smile:

model/peer/registry.yaml
```diff
@@ -13,3 +13,23 @@ groups:
           of the remote service. SHOULD be equal to the actual `service.name`
           resource attribute of the remote service if any.
         examples: "AuthTokenCache"
+        note: >
+          If this is done via declarative configuration, then it MUST be possible to configure the peer service name:
+
+
+          - `peer` is host, and optionally the and path of the remote service that is used to identify the service
```
thompson-tomo, line 20: Should this not read as:

>  `peer` is **the** host, and optionally the ~~and~~ path of the remote service that is used to identify the service instance

thompson-tomo, line 20: Should we not mention port as well in the description given examples show port included?

PR #2506, opened 2025-07-10
Add missing brief on `ios.state` attribute and add policies to require briefs

lmolkova:
We've accidentally removed brief on `ios.state` which leads to codegen issues like https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2896.

I believe we should guarantee brief on attributes to be either set explicitly or auto-generated in some edge cases (#1419) so that codegen should not need to fallback. 

So adding a policy to require briefs on attributes and groups.
We have  A LOT of enum members without briefs, so not adding a check at least for the time being.

### DIFF
--- a/docs/registry/attributes/android.md
+++ b/docs/registry/attributes/android.md
@@ -33,7 +33,7 @@ This document defines attributes that represents an occurrence of a lifecycle tr
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
-| <a id="android-state" href="#android-state">`android.state`</a> | string | Deprecated. Use `android.app.state` body field instead. | `created`; `background`; `foreground` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Use `android.app.state` body field instead. |
+| <a id="android-state" href="#android-state">`android.state`</a> | string | Deprecated. Use `android.app.state` attribute instead. | `created`; `background`; `foreground` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Use `android.app.state` attribute instead. |
 
 ---
 
--- a/model/system/metrics.yaml
+++ b/model/system/metrics.yaml
@@ -149,7 +149,7 @@ groups:
       code_generation:
         metric_value_type: double
     stability: development
-    brief: ""
+    brief: TODO.
     instrument: gauge
     unit: "1"
     attributes:
@@ -183,7 +183,7 @@ groups:
       code_generation:
         metric_value_type: double
     stability: development
-    brief: ""
+    brief: TODO.
     instrument: gauge
     unit: "1"
     attributes:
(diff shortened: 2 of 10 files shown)


docs/registry/attributes/android.md
```diff
@@ -33,7 +33,7 @@ This document defines attributes that represents an occurrence of a lifecycle tr
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
-| <a id="android-state" href="#android-state">`android.state`</a> | string | Deprecated. Use `android.app.state` body field instead. | `created`; `background`; `foreground` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Use `android.app.state` body field instead. |
+| <a id="android-state" href="#android-state">`android.state`</a> | string | Deprecated. Use `android.app.state` attribute instead. | `created`; `background`; `foreground` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Use `android.app.state` attribute instead. |
```
lmolkova, line 36: @open-telemetry/semconv-mobile-approvers could you please check this and a similar correction on the ios.state

model/system/metrics.yaml
```diff
@@ -154,7 +154,7 @@ groups:
       code_generation:
         metric_value_type: double
     stability: development
-    brief: ""
+    brief: TODO
```
lmolkova: created https://github.com/open-telemetry/semantic-conventions/issues/2507 to update

Kielek (approved):
Thank you!

model/system/metrics.yaml
thompson-tomo: pr raised which completes the TODO: https://github.com/lmolkova/semantic-conventions/pull/13

trisch-me: @thompson-tomo it would be better to make this PR against main repo so we all could verify and approve it

thompson-tomo: The pr is targeting this branch, so if @lmolkova merges my pr the changes gets added to this PR hence can be reviewed.

thompson-tomo: Note if it was against main, then a bunch of merge conflicts would occur as we both have changed the same line.

lmolkova: We try to avoid bundling independent changes together. The briefs on system metrics can be added to main independently from this change and would need to be reviewed by the system approvers. I'd recommend sending them as a separate PR and polishing wording with system semconv sig

breedx-splk (approved):
LGTM

policies_test/metric_brief_format_test.rego
```diff
@@ -26,25 +26,6 @@ test_metric_brief_without_period if {
     ]}
 }
 
-test_metric_without_brief if {
```
lmolkova, line 29: those are now covered with `test_fails_on_group_without_brief` in yaml_schema_test.rego

PR #2508, opened 2025-07-11
Add ASP.NET Core identity metrics

### LINKED ISSUE #2509: Add ASP.NET Core Identity metrics
ASP.NET Core is adding new metrics in the next version for Identity. These should be documented in the semantic conventions.

JamesNK:
Fixes #2509

## Changes

Adds documentation for metrics added to ASP.NET Core identity.

* [x] Link to instrumentation: https://github.com/dotnet/aspnetcore/pull/62078


JamesNK:
What's the right way to have `_OTHER` values? They're creating an error.

thompson-tomo:
@JamesNK Have u tried removing the underscore from the Id? Note fields which are enums are not restricted to those known values but in fact other values can be sent hence reducing the need for _other.

.chloggen/2508.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: all
```
lmolkova: nit


```suggestion
component: aspnetcore
```

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
         stability: stable
-        brief: ASP.NET Core exception middleware handling result
+        brief: ASP.NET Core exception middleware handling result.
         examples: ["handled", "unhandled"]
+      - id: aspnetcore.identity.user_type
+        type: string
+        brief: The full name of the identity user type.
+        stability: stable
+        examples: ["Contoso.ContosoUser"]
+      - id: aspnetcore.identity.result
```
lmolkova, line 91: do you expect this list to grow? specifically - are we going to have  more success statuses? 

The suggestion (that matches what we usually do in semconv including HTTP) is to 
- remove `aspnetcore.identity.result`
- rename `aspnetcore.identity.result_error_code` to `aspnetcore.identity.error_code` and populate it when there was no exception but there was another error. It'd be similar to `http.response.status_code`
- populate `error.type` whenever error happens - exception or [`IdentityError.Code`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.identityerror.code?view=aspnetcore-9.0#microsoft-aspnetcore-identity-identityerror-code) indicates an error. It'd be consistent with http when we put unsuccessful status codes on `error.type`

This way it's trivial to distinguish errors from non-errors and you can still distinguish exceptions from responses with error.

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+              brief: "Identity operation was successful."
+              stability: stable
+            - id: failure
+              value: 'failure'
+              brief: "Identity operation failed."
+              stability: stable
+        stability: stable
+        brief: The result of the identity operation.
+        examples: ["success", "failure"]
+      - id: aspnetcore.identity.result_error_code
```
lmolkova: WDYT about 

```suggestion
      - id: aspnetcore.identity.error_code
```

?

it looks more readable

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+              stability: stable
+            - id: set_passkey
+              value: 'set_passkey'
+              brief: "Identity user passkey set."
+              stability: stable
+            - id: remove_passkey
+              value: 'remove_passkey'
+              brief: "Identity user passkey removed."
+              stability: stable
+            - id: _OTHER
```
lmolkova:
```suggestion
            - id: other
```

id and name don't have to be the same

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+              stability: stable
+            - id: change_email
+              value: 'change_email'
+              brief: "The token is for changing the user email address."
+              stability: stable
+            - id: two_factor
+              value: 'two_factor'
+              brief: "The token is for changing user two factor settings."
+              stability: stable
+            - id: _OTHER
```
lmolkova:
```suggestion
            - id: other
```

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+        examples: ["Identity.Application"]
+      - id: aspnetcore.identity.sign_in.type
+        type:
+          members:
+            - id: password
+              value: 'password'
+              brief: "Sign in with password."
+              stability: stable
+            - id: two_factor_recovery_code
+              value: 'two_factor_recovery_code'
```
lmolkova, line 331: nit: given it's limited to AspNetCore, we can totally use PascalCase and enum member names used in .NET, i.e. `TwoFactorRecoveryCode`.

No strong preference from my side as long as it's consistent within the attribute/AspNetCore.

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+              brief: "Sign in with a previously registered third-party login."
+              stability: stable
+            - id: passkey
+              value: 'passkey'
+              brief: "Sign in with passkey."
+              stability: stable
+        stability: stable
+        brief: The authentication type.
+        examples: ["password", "two_factor"]
+      - id: aspnetcore.identity.sign_in.is_persistent
```
lmolkova: we try to avoid using `is_*` in other semconv (but we have a few stable occurrences in .NET already - `aspnetcore.routing.is_fallback`,  `aspnetcore.request.is_unhandled`). 

Given we already use `is_` pattern in other ASP.NET Core metrics and it mirrors ASP.NET Core API I'd rather stay consistent within ASP.NET Core.

/cc @trask  - I think you had concerns with `is_*` in the past.

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
               brief: "Exception handling didn't run because the request was aborted."
               stability: stable
         stability: stable
-        brief: ASP.NET Core exception middleware handling result
+        brief: ASP.NET Core exception middleware handling result.
         examples: ["handled", "unhandled"]
+      - id: aspnetcore.identity.user_type
+        type: string
+        brief: The full name of the identity user type.
+        stability: stable
```
lmolkova: are we going to them stable in Nov? If so, lets add them as `development` now and change to stable when they actually ship.

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+        requirement_level:
+          conditionally_required: if no exception was thrown.
+      - ref: aspnetcore.identity.result_error_code
+        requirement_level:
+          conditionally_required: if an error was set on a failed identity result.
+      - ref: error.type
+        brief: The full name of exception type.
+        requirement_level:
+          conditionally_required: if and only if an error has occurred.
+        examples: ['System.OperationCanceledException']
```
lmolkova: If we can get rid of `aspnetcore.identity.result` and record error.type with all errors , it becomes

 
```suggestion
      - ref: aspnetcore.identity.error_code
        requirement_level:
          conditionally_required: if an error was set on a failed identity result.
      - ref: error.type
        brief: `IdentityError.Code` or the full name of exception type.
        requirement_level:
          conditionally_required: if and only if an error has occurred.
        examples: ['System.OperationCanceledException']
```

(without losing any details)

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+  - id: metric.aspnetcore.identity.user.delete
+    type: metric
+    metric_name: aspnetcore.identity.user.delete
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: stable
+    brief: The number of users deleted.
+    instrument: counter
+    unit: "{count}"
```
lmolkova: nit


```suggestion
    unit: "{user}"
```

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+              brief: "Identity user passkey removed."
+              stability: stable
+            - id: other
+              value: '_OTHER'
+              brief: "Any update type that the instrumentation has no prior knowledge of."
+              stability: stable
+        stability: stable
+        brief: The user update type.
+        examples: ["update", "user_name", "reset_password"]
+      - id: aspnetcore.identity.user.password_result
```
lmolkova: wdyt about 

```suggestion
      - id: aspnetcore.identity.password_check_result
```

I assume it doesn't matter if it's happened on the user-related API?

It'd also match the API name [Microsoft.AspNetCore.Identity.PasswordVerificationResult](https://github.com/dotnet/aspnetcore/blob/main/src/Identity/Extensions.Core/src/PasswordVerificationResult.cs)

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+  - id: metric.aspnetcore.identity.user.create
+    type: metric
+    metric_name: aspnetcore.identity.user.create
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: stable
+    brief: The number of users created.
+    instrument: counter
+    unit: "{count}"
```
lmolkova: nit

```suggestion
    unit: "{user}"
```

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+  - id: metric.aspnetcore.identity.user.update
+    type: metric
+    metric_name: aspnetcore.identity.user.update
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: stable
+    brief: The number of user updates.
+    instrument: counter
+    unit: "{count}"
```
```suggestion
    unit: "{update}"
```

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+  - id: metric.aspnetcore.identity.user.check_password
+    type: metric
+    metric_name: aspnetcore.identity.user.check_password
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: stable
+    brief: The number of check password attempts. Only checks whether the password is valid and not whether the user account is in a state that can log in.
+    instrument: counter
+    unit: "{count}"
```
```suggestion
    unit: "{check}"
```

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+  - id: metric.aspnetcore.identity.sign_in.sign_out_principal
+    type: metric
+    metric_name: aspnetcore.identity.sign_in.sign_out_principal
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: stable
+    brief: The number of calls to sign out user principals.
+    instrument: counter
+    unit: "{count}"
```
```suggestion
    unit: "{sign_out}"
```

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+        examples: ['System.OperationCanceledException']
+
+  - id: metric.aspnetcore.identity.sign_in.check_password
+    type: metric
+    metric_name: aspnetcore.identity.sign_in.check_password
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: stable
+    brief: The number of check password attempts. Checks that the account is in a state that can log in and that the password is valid using the UserManager.CheckPasswordAsync method.
```
lmolkova: >  UserManager.CheckPasswordAsync

is this right? 

I assumed that `sign_in` metrics track  https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.signinmanager-1 and `user` metrics track usermanager, is this correct?

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+    unit: "{count}"
+    note: |
+      Meter name: `Microsoft.AspNetCore.Identity`; Added in: ASP.NET Core 10.0
+    extends: aspnetcore.common.identity.metrics.attributes
+    attributes:
+      - ref: aspnetcore.identity.sign_in.result
+        requirement_level:
+          conditionally_required: if no exception was thrown.
+      - ref: error.type
+        brief: The full name of exception type.
```
lmolkova, line 398: usual question on whether failed login (without exception) should be an error and then whether error.type should be populated.

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+  - id: metric.aspnetcore.identity.sign_in.sign_in_principal
+    type: metric
+    metric_name: aspnetcore.identity.sign_in.sign_in_principal
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: stable
+    brief: The number of calls to sign in user principals.
+    instrument: counter
+    unit: "{count}"
```
```suggestion
    unit: "{sign_in}"
```

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+      - ref: aspnetcore.identity.sign_in.result
+        requirement_level:
+          conditionally_required: if no exception was thrown.
+      - ref: error.type
+        brief: The full name of exception type.
+        requirement_level:
+          conditionally_required: if and only if an error has occurred.
+        examples: ['System.OperationCanceledException']
+
+  - id: metric.aspnetcore.identity.sign_in.sign_in_principal
```
lmolkova: wdyt about 

```suggestion
  - id: metric.aspnetcore.identity.sign_in.sign_in
```

and  `aspnetcore.identity.sign_in.sign_out`

It shorter, it matches the API, where the principal is not mentioned at all (and it sent me into a huge rabbit hole because of my ignorance on identity topics).

model/aspnetcore/metrics.yaml
JamesNK: I use `{count}` in these metrics as the unit.

Would a more descriptive unit be better? For example, `{user}` for this method

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+      - ref: aspnetcore.identity.user.update_type
+        requirement_level: required
+      - ref: aspnetcore.identity.result
+        requirement_level:
+          conditionally_required: if no exception was thrown.
+      - ref: aspnetcore.identity.result_error_code
+        requirement_level:
+          conditionally_required: if an error was set on a failed identity result.
+      - ref: error.type
+        brief: The full name of exception type.
```
lmolkova: I *think* this should include error code returned in IdentityResult.Error(s)

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+    unit: "{count}"
+    note: |
+      Meter name: `Microsoft.AspNetCore.Identity`; Added in: ASP.NET Core 10.0
+    extends: aspnetcore.common.identity.metrics.attributes
+    attributes:
+      - ref: aspnetcore.identity.user.password_result
+        requirement_level:
+          conditionally_required: if no exception was thrown.
+      - ref: error.type
+        brief: The full name of exception type.
```
lmolkova, line 248: usual comment on whether it should include failed identity results (e.g. aspnetcore.identity.user.password_result = failed)

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+      Meter name: `Microsoft.AspNetCore.Identity`; Added in: ASP.NET Core 10.0
+    extends: aspnetcore.common.identity.metrics.attributes
+    attributes:
+      - ref: aspnetcore.identity.token_purpose
+        requirement_level: required
+      - ref: aspnetcore.identity.token_verified
+        requirement_level:
+          conditionally_required: if no exception was thrown.
+      - ref: error.type
+        brief: The full name of exception type.
```
lmolkova, line 274: a slightly different question on whether invalid token should be treated as an error

lmolkova (commented):
Thank you!

I left a bunch of minor naming suggestions. The key question is what should be counted as errors.  I assume `IdentityError` is an error and the code should appear on `error.type`. I'd argue that at least some of other results should be populated as errors (e.g. `aspnetcore.identity.user.password_result = failure`)

model/aspnetcore/metrics.yaml
lmolkova: yep, more precise unit would be better (where it makes sense)

JamesNK: Lower down you propose `aspnetcore.identity.user.update` have a unit of `{update}`. Should this metric unit be `{create}` instead of user? User and update seems inconsitent.

JamesNK: Or should `aspnetcore.identity.user.update` have a unit of `{user}`?

I think these should to be consistent:
- `aspnetcore.identity.user.create`
- `aspnetcore.identity.user.update`
- `aspnetcore.identity.user.delete`

model/aspnetcore/metrics.yaml
JamesNK: This comment is here to try and explain the difference between `aspnetcore.identity.sign_in.check_password` and `aspnetcore.identity.user.check_password`.

Checking the password on SignManager will incremented both counters because the sign in manager uses the user manager internally.

JamesNK:
I need to investigate more for feedback on changing/removing some results and using `error.type` more.

model/aspnetcore/registry.yaml
JamesNK: > do you expect this list to grow? specifically - are we going to have more success statuses?

No.

> remove `aspnetcore.identity.result`

A failure result doesn't have to have an error code. I don't think `aspnetcore.identity.result` can be removed and replaced with the presence of `error.type` because there won't always be a value to put in `error.type`.

> rename aspnetcore.identity.result_error_code to aspnetcore.identity.error_code and populate it when there was no exception but there was another error. It'd be similar to http.response.status_code

OK

> populate error.type whenever error happens - exception or [IdentityError.Code](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.identityerror.code?view=aspnetcore-9.0#microsoft-aspnetcore-identity-identityerror-code) indicates an error. It'd be consistent with http when we put unsuccessful status codes on error.type

OK

model/aspnetcore/registry.yaml
JamesNK: We use this format in most places already so we'll keep it.

model/aspnetcore/metrics.yaml
JamesNK: I don't think this will work. See https://github.com/open-telemetry/semantic-conventions/pull/2508#discussion_r2206213073

model/aspnetcore/metrics.yaml
JamesNK: No, I don't think so

model/aspnetcore/metrics.yaml
JamesNK: I don't think it is an error for the password to be wrong.

JamesNK:
Updated. I used most feedback. I notably I think `aspnetcore.identity.result` is still required.

JamesNK:
Note: Don't merge this until .NET 10 preview 7 ships.

JamesNK:
I fixed the generated markdown files by copying output from build. Is it possible to run the tooling on Windows to generate markdown? Or is WSL recommended?

docs/dotnet/dotnet-aspnetcore-metrics.md
```diff
@@ -310,4 +323,816 @@ Meter name: `Microsoft.AspNetCore.RateLimiting`; Added in: ASP.NET Core 8.0
...
+should be prepared for `error.type` to have high cardinality at query time when no
+additional filters are applied.
+
+If the operation has completed successfully, instrumentations SHOULD NOT set `error.type`.
+
+If a specific domain defines its own set of error identifiers (such as HTTP or gRPC status codes),
+it's RECOMMENDED to:
+
+- Use a domain-specific attribute
+- Set `error.type` to capture all errors, regardless of whether they are defined within the domain-specific set or not.
```
lmolkova: I believe this note is generic and not useful here, can we remove it? It could be achieved with adding `note: ""` when attribute is referenced. I'll leave another comment on the yaml

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +141,298 @@ groups:
...
+        requirement_level:
+          conditionally_required: if no exception was thrown.
+      - ref: aspnetcore.identity.error_code
+        requirement_level:
+          conditionally_required: if an error was set on a failed identity result.
+      - ref: error.type
+        brief: The full name of exception type or the identity error code.
+        requirement_level:
+          conditionally_required: if and only if an error has occurred.
+        examples: ['System.OperationCanceledException', 'PasswordMismatch']
```
lmolkova, line 171: to override generic note

```suggestion
        examples: ['System.OperationCanceledException', 'PasswordMismatch']
        note: ""
```

(also on other metrics)

docs/dotnet/dotnet-aspnetcore-metrics.md
```diff
@@ -21,6 +21,19 @@ This article defines semantic conventions for ASP.NET Core metrics.
   - [Metric: `aspnetcore.rate_limiting.queued_requests`](#metric-aspnetcorerate_limitingqueued_requests)
   - [Metric: `aspnetcore.rate_limiting.request.time_in_queue`](#metric-aspnetcorerate_limitingrequesttime_in_queue)
   - [Metric: `aspnetcore.rate_limiting.requests`](#metric-aspnetcorerate_limitingrequests)
+- [Identity](#identity)
+  - [Metric: `aspnetcore.identity.user.create`](#metric-aspnetcoreidentityusercreate)
```
lmolkova: I should have asked before, but is there a reason we report them as counters and not duration histograms?

lmolkova (approved):
LGTM with some nits.

It'd be great to document (here in the comments) the reason why we report counters and not duration histograms.

docs/dotnet/dotnet-aspnetcore-metrics.md
JamesNK: They're counters because I don't think people care about the durations of identity methods. They're unlikely to be on the hot path of an app.

lmolkova:
@JamesNK 

>I fixed the generated markdown files by copying output from build. Is it possible to run the tooling on Windows to generate markdown? Or is WSL recommended?

wsl is recommended. It's possible to install weaver https://github.com/open-telemetry/weaver/releases (or use docker image - `otel/weaver`) and run something like these two commands 

```ps
weaver registry update-markdown --registry .\model --param registry_base_url=/docs/registry/ --target markdown --templates .\templates --future .\docs
```

and 

```ps
weaver registry generate --registry .\model --templates .\templates markdown .\docs\registry\
```

which would be the same as `make table-generation attribute-registry-generation`

docs/dotnet/dotnet-aspnetcore-metrics.md
lmolkova: We can still create histograms with 1-2 buckets by default, the benefit is that it's comparable to counters in terms of volume, but users can change bucket boundaries when they care about durations. Not blocking here, FJYI for the future

lmolkova:
@open-telemetry/semconv-dotnet-approver @open-telemetry/dotnet-contrib-approvers @open-telemetry/dotnet-approvers PTAL

docs/dotnet/dotnet-aspnetcore-metrics.md
```diff
@@ -21,6 +21,19 @@ This article defines semantic conventions for ASP.NET Core metrics.
   - [Metric: `aspnetcore.rate_limiting.queued_requests`](#metric-aspnetcorerate_limitingqueued_requests)
   - [Metric: `aspnetcore.rate_limiting.request.time_in_queue`](#metric-aspnetcorerate_limitingrequesttime_in_queue)
   - [Metric: `aspnetcore.rate_limiting.requests`](#metric-aspnetcorerate_limitingrequests)
+- [Identity](#identity)
```
Kielek, line 24: Do you have a plan to mark ASP.NET Core versions supporting given metrics, or it should be done on implementer side?
I suppose in this case it will be .NET10+.

@JamesNK, any plans to extend instrumentation package in the contrib with new activity source?

JamesNK: I don't understand the question.

UserManager and SignInManager support metrics. If you change the user provider for example then you still get metrics, but if you override methods on the managers then you'll lose them and need to readd to your new method.

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +142,310 @@ groups:
     attributes:
       - ref: aspnetcore.rate_limiting.result
         requirement_level: required
+
+  # identity
+  - id: metric.aspnetcore.identity.user.create
```
JamesNK: Should these counter names be puralized? I noticed other counters in this file are puralized, e.g. `aspnetcore.routing.match_attempts`.

Pluralized versions of counters:
* aspnetcore.identity.user.create -> aspnetcore.identity.user.creates
* aspnetcore.identity.user.update -> aspnetcore.identity.user.updates
* aspnetcore.identity.user.delete -> aspnetcore.identity.user.deletes
* aspnetcore.identity.user.check_password -> aspnetcore.identity.user.check_password_attempts
* aspnetcore.identity.user.generate_token -> aspnetcore.identity.user.generated_tokens
* aspnetcore.identity.user.verify_token -> aspnetcore.identity.user.verify_token_attempts
* aspnetcore.identity.sign_in.authenticate -> aspnetcore.identity.sign_in.authentications
* aspnetcore.identity.sign_in.check_password -> aspnetcore.identity.sign_in.check_password_attempts
* aspnetcore.identity.sign_in.sign_in -> aspnetcore.identity.sign_in.sign_ins
* aspnetcore.identity.sign_in.sign_out -> aspnetcore.identity.sign_in.sign_outs
* aspnetcore.identity.sign_in.remember_two_factor -> aspnetcore.identity.sign_in.remembered_two_factor_clients
* aspnetcore.identity.sign_in.forget_two_factor -> aspnetcore.identity.sign_in.forgotten_two_factor_clients

What do you think @lmolkova?

docs/dotnet/dotnet-aspnetcore-metrics.md
Kielek: Sorry, let me rephrase it.
Do you think that this document should contain information about ASP.NET Core version supporting given set of versions or it should be done only on .NET documentation side.
As I understand, Identity _space_ will be supported only by ASP.NET Core 10+.

model/aspnetcore/metrics.yaml
lmolkova: good catch! in my mental model, the name of these (new) metrics match the API. E.g. `aspnetcore.identity.user.create` matches `Microsoft.AspNetCore.Identity.UserManager.CreateAsync` almost precisely. 

If we pluralize them (`aspnetcore.identity.user.creates` ) or use adjectives (`aspnetcore.identity.user.created`) we'll deviate from the API and it will be a loss.

Other metrics in this file, if I understand correctly, are not reporting specific API calls and don't really have this problem. Moreover, if we designed them today, we'd recommend [not to pluralize updowncounters](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#do-not-pluralize-updowncounter-names), e.g. we'd suggest using `aspnetcore.rate_limiting.request_lease.active` instead of `aspnetcore.rate_limiting.active_request_leases`. 

I agree pluralization feels somewhat inconsistent. My soft preference is to stay close to API when metric is tight to it - that's something users can easily understand and it's an easy principle to follow to minimize naming bikeshed.

lmolkova: thinking about it a bit more in combination with https://github.com/open-telemetry/semantic-conventions/pull/2508#discussion_r2211907232, assuming we'd ever want to report duration histograms, we'd use a name like `aspnetcore.identity.user.create.duration` and at this point we'd regret having `aspnetcore.identity.user.create` as a counter (because it looks like a namespace).  

Names like `aspnetcore.identity.user.create.count` would be very precise and pluralization would not make sense.

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+        type:
+          members:
+            - id: update
+              value: 'update'
+              brief: "Identity user updated."
+              stability: development
+            - id: user_name
+              value: 'user_name'
+              brief: "Identity user name updated."
+              stability: development
```
thompson-tomo, line 120: Should this be change_user_name?

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+            - id: two_factor
+              value: 'two_factor'
+              brief: "The token is for changing user two factor settings."
+              stability: development
+            - id: other
+              value: '_OTHER'
+              brief: "Any token purpose that the instrumentation has no prior knowledge of."
+              stability: development
+        stability: development
+        brief: The result from checking the password.
```
thompson-tomo: Fix

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+              value: 'requires_two_factor'
+              brief: "User requires two factory authentication to sign in."
+              stability: development
+            - id: failure
+              value: 'failure'
+              brief: "Sign in failed."
+              stability: development
+        stability: development
+        brief: Whether the sign in result was success or failure.
+        examples: ["password", "two_factor"]
```
thompson-tomo: Thoughts about putting some/all of these in an access-control.authentication/iam.authentication namespace?

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+              value: 'success'
+              brief: "Token verification was successful."
+              stability: development
+            - id: failure
+              value: 'failure'
+              brief: "Token verification failed."
+              stability: development
+        stability: development
+        brief: The result of token verification.
+        examples: ["success", "failure"]
```
thompson-tomo, line 317: Could this not leverage sign_in.result?

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+      - id: aspnetcore.identity.authentication_scheme
+        type: string
+        stability: development
+        brief: The authentication scheme to sign in with.
+        examples: ["Identity.Application"]
```
thompson-tomo: Thoughts about putting this in an access-control.authentication/iam.authentication  namespace?

model/aspnetcore/registry.yaml
```diff
@@ -81,5 +81,302 @@ groups:
...
+              value: 'two_factor'
+              brief: "The token is for changing user two factor settings."
+              stability: development
+            - id: other
+              value: '_OTHER'
+              brief: "Any token purpose that the instrumentation has no prior knowledge of."
+              stability: development
+        stability: development
+        brief: The result from checking the password.
+        examples: ["success", "failure"]
```
thompson-tomo: Thoughts about putting some/all of these in an access-control/iam namespace?

docs/dotnet/dotnet-aspnetcore-metrics.md
JamesNK: Each metric has a note that says:

> Meter name: `Microsoft.AspNetCore.Identity`; Added in: ASP.NET Core 10.0

docs/dotnet/dotnet-aspnetcore-metrics.md
JamesNK: I've changed create/update/delete to histograms.

model/aspnetcore/registry.yaml
JamesNK: I want to keep them specific to ASP.NET Core and not worry about a cross platform standard.

JamesNK:

lmolkova:
LGTM!

could we please get an approval from @open-telemetry/semconv-dotnet-approver ?

model/aspnetcore/registry.yaml
trask: I think it's generally unnecessary in semconv attribute names, but ok with it here if you prefer for consistency with already stabilized attributes

model/aspnetcore/metrics.yaml
trask: I think ok to resolve this based on the
- https://github.com/open-telemetry/semantic-conventions/pull/2508/files#r2206439317
- https://github.com/open-telemetry/semantic-conventions/pull/2508/files#r2206230514

PR #2514, opened 2025-07-13
(k8s): add CPU limit/request utilization metrics for pod and container

ChrsMark:
Part of https://github.com/open-telemetry/semantic-conventions/issues/1489 and https://github.com/open-telemetry/semantic-conventions/issues/1032

## Changes

This PR adds the following metrics:

- `k8s.pod.cpu.limit_utilization`
- `k8s.pod.cpu.request_utilization`
- `k8s.container.cpu.limit_utilization`
- `k8s.container.cpu.request_utilization`

These metrics are already use by the Collector: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.129.0/receiver/kubeletstatsreceiver/documentation.md


ChrsMark:
@open-telemetry/semconv-k8s-approvers please take a look

ChrsMark:
@jinja2 @skl @dmitryax I have removed the pod related metric for now until we have the discussion on how/if these will be implemented. I filed an issue in the Collector contrib so we can revisit the current implementation first: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42485

Apart from this, the current PR is ready for review so please take a look.

model/k8s/metrics.yaml
```diff
@@ -1142,6 +1142,34 @@ groups:
...
+    metric_name: k8s.container.cpu.request_utilization
+    annotations:
+      code_generation:
+        metric_value_type: double
+    stability: development
+    brief: "The ratio of container CPU usage to its CPU request."
+    entity_associations:
+      - k8s.container
+    note: >
+      The value range is [0.0,1.0]. A value of 1.0 means the container is using 100% of its CPU request.
```
skl: Container CPU usage can be greater than the CPU requests, so the range is unbounded

ChrsMark: Ah right, good catch! Removed the note, thank's!

model/k8s/metrics.yaml
```diff
@@ -1142,6 +1142,32 @@ groups:
...
+    metric_name: k8s.container.cpu.limit_utilization
+    annotations:
+      code_generation:
+        metric_value_type: double
+    stability: development
+    brief: "The ratio of container CPU usage to its CPU limit."
+    entity_associations:
+      - k8s.container
+    note: >
+      The value range is [0.0,1.0]. A value of 1.0 means the container is using 100% of its CPU limit.
```
jinja2:
```suggestion
      The value range is [0.0,1.0]. A value of 1.0 means the container is using 100% of its CPU limit. If the CPU limit is not set, this metric SHOULD NOT be emitted for that container.
```
I think we should make this additional clarification. Same for request

ChrsMark:
@open-telemetry/specs-semconv-approvers @open-telemetry/specs-semconv-maintainers could you please take a look?

PR #2521, opened 2025-07-15
Adds in time in queue metric and doc fixes #2516

### LINKED ISSUE #2516: The http metrics for dotnet should be generated and not statically defined in md
### Area(s)

docs, area:dotnet, area:http

### What's missing?

Auto generated metric definitions

On the https://github.com/open-telemetry/semantic-conventions/blob/main/docs/dotnet/dotnet-http-metrics.md page the client metrics should be auto-generated as stability status can be generated. Also the server metrics should be auto generated following conventions.

thompson-tomo:
Fixes #2516

## Changes

This migrates the time in queue metric to yaml and gets the dotnet http metrics to be generated


model/http/metrics.yaml
```diff
@@ -143,6 +143,18 @@ groups:
       header. For requests using transport encoding, this should be the compressed size.
     extends: metric_attributes.http.client.experimental
 
+  - id: metric.http.client.request.time_in_queue
+    type: metric
+    metric_name: http.client.request.time_in_queue
+    annotations:
+      code_generation:
+        metric_value_type: int
```
JamesNK: I haven't looked at the source code to confirm, but all other duration types are double.
```suggestion
        metric_value_type: double
```

model/http/metrics.yaml
```diff
@@ -143,6 +143,18 @@ groups:
...
+    stability: development
+    brief: "The amount of time requests spent on a queue waiting for an available connection."
```
JamesNK:
```suggestion
    brief: "The amount of time requests spent in a queue waiting for an available connection."
```

docs/dotnet/dotnet-http-metrics.md
```diff
@@ -43,152 +43,270 @@ Notes:
...
+|---|---|---|---|---|---|
+| [`http.request.method`](/docs/registry/attributes/http.md) | string | HTTP request method. [1] | `GET`; `POST`; `HEAD` | `Required` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`server.address`](/docs/registry/attributes/server.md) | string | Server domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name. [2] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Required` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`server.port`](/docs/registry/attributes/server.md) | int | Server port number. [3] | `80`; `8080`; `443` | `Required` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`error.type`](/docs/registry/attributes/error.md) | string | Describes a class of error the operation ended with. [4] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` If request has ended with an error. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
antonfirsov: This metric currently doesn't emit `error.type`. This can be changed if needed.

> `java.net.UnknownHostException`

Such examples should not appear in the docs of a .NET-specific instrument.

docs/dotnet/dotnet-http-metrics.md
```diff
@@ -43,152 +43,270 @@ Notes:
...
+| [`http.response.status_code`](/docs/registry/attributes/http.md) | int | [HTTP response status code](https://tools.ietf.org/html/rfc7231#section-6). | `200` | `Conditionally Required` If and only if one was received/sent. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
antonfirsov: `http.response.status_code` doesn't make sense for this metric. We don't know the status code at the point when it is emitted.

docs/dotnet/dotnet-http-metrics.md
```diff
@@ -43,152 +43,270 @@ Notes:
...
+| [`network.protocol.name`](/docs/registry/attributes/network.md) | string | [OSI application layer](https://wikipedia.org/wiki/Application_layer) or non-OSI equivalent. [5] | `http`; `spdy` | `Conditionally Required` [6] | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
antonfirsov: Currently, this instrument does not emit `network.protocol.name`. It can be changed if we want to.

docs/dotnet/dotnet-http-metrics.md
```diff
@@ -43,152 +43,270 @@ Notes:
...
+| [`network.protocol.version`](/docs/registry/attributes/network.md) | string | The actual version of the protocol used for network communication. [7] | `1.0`; `1.1`; `2`; `3` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`url.scheme`](/docs/registry/attributes/url.md) | string | The [URI scheme](https://www.rfc-editor.org/rfc/rfc3986#section-3.1) component identifying the used protocol. | `http`; `https` | `Opt-In` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`url.template`](/docs/registry/attributes/url.md) | string | The low-cardinality template of an [absolute path reference](https://www.rfc-editor.org/rfc/rfc3986#section-4.2). [8] | `/users/{id}`; `/users/:id`; `/users?id={id}` | `Opt-In` | ![Development](https://img.shields.io/badge/-development-blue) |
```
antonfirsov: This would be nice to have but, currently unimplemented, and I don't expect it to happen anytime soon.

antonfirsov (commented):
This reults in unintentional changes in the spec. If there are some changes that are desired, they should be discussed first, and an issue has to be open against `dotnet/runtime`. Note that even if decide to eventually take them, there is no guarantee the change will happen in the next release. This can result in a state where the spec and the implementation is out of sync for a couple releases which would be unfortunate.

cc @lmolkova

thompson-tomo:
@antonfirsov i agree the last thing I want is for an implementation to deviate from the spec which is why it is important that all specs are in yaml and docs generated based on specs. Today I have planned to go through the diff reviewing the changes seeing what the actual changes are seeing what if any changes are needed & where.

docs/dotnet/dotnet-http-metrics.md
thompson-tomo: Will split this off to avoid blocking this PR. Have raised https://github.com/dotnet/runtime/issues/117695 for it to be added and have raised a corresponding spec pr to add it.

thompson-tomo:
Just went through comparing the changes and have made a couple of changes to avoid un-intended changes. 

@open-telemetry/semconv-http-approvers i have set the stability to RC for metrics which have been released but not declared stable

antonfirsov:
cc @karelz @mihazupan @manickap

docs/dotnet/dotnet-http-metrics.md
```diff
@@ -152,43 +168,77 @@ If the HTTP request method isn't known, it sets the `http.request.method` attrib
...
+| `http.client.active_requests` | UpDownCounter | `{request}` | Number of active HTTP requests. | ![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid) |  |
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`server.address`](/docs/registry/attributes/server.md) | string | Server domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name. [1] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Required` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`server.port`](/docs/registry/attributes/server.md) | int | Server port number. [2] | `80`; `8080`; `443` | `Required` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`url.template`](/docs/registry/attributes/url.md) | string | The low-cardinality template of an [absolute path reference](https://www.rfc-editor.org/rfc/rfc3986#section-4.2). [3] | `/users/{id}`; `/users/:id`; `/users?id={id}` | `Conditionally Required` If available. | ![Development](https://img.shields.io/badge/-development-blue) |
```
thompson-tomo: A difference but requirement makes it ok I think.

thompson-tomo:
The missing metric definition has been split out to #2760 hence this PR now is focussed on just auto generating definitions and bumping stability to rc.

PR #2522, opened 2025-07-15
model/profile: add symbolization information

florianl:
## Changes

Add Profiling specific attributes that describe the symbolization level of a Mapping.

FYI: @open-telemetry/profiling-approvers 


docs/general/profiles.md
```diff
@@ -49,6 +50,55 @@ They may be used in any Profiles record they apply to.
 | `rust` | [Rust](https://wikipedia.org/wiki/Rust_(programming_language)) | ![Development](https://img.shields.io/badge/-development-blue) |
 | `v8js` | [V8JS](https://wikipedia.org/wiki/V8_(JavaScript_engine)) | ![Development](https://img.shields.io/badge/-development-blue) |
 
+---
+
+`profile.mapping.symbolization` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+
+| Value  | Description | Stability |
+|---|---|---|
+| `lines_inline` | Full, most-desired level of symbolization. [1] | ![Development](https://img.shields.io/badge/-development-blue) |
```
rockdaboot: Do we need the inline flag here?
The inline information is documented to be available on the location level:
```
  // Multiple line indicates this location has inlined functions,
  // where the last entry represents the caller into which the
  // preceding entries were inlined.
  //
  // E.g., if memcpy() is inlined into printf:
  //    line[0].function_name == "memcpy"
  //    line[1].function_name == "printf"
```
And even if we want it, would it be possible to have `symbols` together with `inline` in the future or with other agent than ebpf-profiler?

docs/general/profiles.md
```diff
@@ -49,6 +50,55 @@ They may be used in any Profiles record they apply to.
...
+| `lines_no_inline` | Limited debug information is available. [2] | ![Development](https://img.shields.io/badge/-development-blue) |
+| `none` | No symbolization was attempted for the mapping. [3] | ![Development](https://img.shields.io/badge/-development-blue) |
+| `symbols` | Limited symbol information is available. [4] | ![Development](https://img.shields.io/badge/-development-blue) |
+| `unspecified` | Unknown or unassigned. [5] | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[1]:** All of function and file names, line numbers and inline frames are
+available. This level indicates that full debug information was
```
rockdaboot: Can we define "full"? Does it include `Line.column` or would that be even "fuller"?

aalexand:
I think we agreed [here](https://docs.google.com/document/d/19UqPPPlGE83N37MhS93uRlxsP1_wGxQ33Qv6CDHaEp0/edit?tab=t.0#bookmark=id.2rqvdd5azlr) to add these as `pprof.*` boolean attributes equivalent to the current pprof's proto attributes since use cases other than pprof's were unclear to us and for pprof keeping the current semantics is the easiest for the migration.

florianl:
@aalexand thanks for reminding me about the `pprof` group. I did forget about this :see_no_evil: updated the PR accordingly

But I don't recall from the recoding, that we agreed on a 1:1 mapping from the google/pprof has_* fields. To me, the suggestions you made in https://github.com/open-telemetry/opentelemetry-proto/pull/595, made sense as they add information about the quality of the symbol information. Using the 1:1 mapping from the google/pprof has_* fields does not allow the backend to make a sound decision on the quality of the received information.

model/profile/common.yaml
```diff
@@ -6,3 +6,5 @@ groups:
     attributes:
       - ref: profile.frame.type
         requirement_level: recommended
+      - ref: profile.pprof.mapping
```
aalexand: I noticed that this `profile.pprof.mapping` attribute is under `profile.frame group` which seems a little odd.

model/profile/registry.yaml
```diff
@@ -78,3 +78,36 @@ groups:
                 [Rust](https://wikipedia.org/wiki/Rust_(programming_language))
               value: "rust"
               stability: development
+  - id: registry.profile.pprof
+    type: attribute_group
```
aalexand: Does `attribute_group` type allow holding multiple values in a single attribute? With the `has_*` values a mapping can have several of them set.

model/profile/common.yaml
florianl: Created dedicated group with `profile.pprof` and moved `profile.pprof.mapping` to it - see https://github.com/open-telemetry/semantic-conventions/pull/2522/commits/d1708e792880b7b769bb6752317fd863dd2f9321

model/profile/registry.yaml
florianl: To my knowledge, a single attribute, like `profile.pprof.mapping`, can not have multiple values, like `has_filenames` and `has_functions`. There can be onlye one value per attribute. Consequently, we have to set an attribute per value.

I also asked in [#otel-semantic-conventions](https://cloud-native.slack.com/archives/C041APFBYQP/p1752821911097279) about this.

aalexand: To follow up on this: I'm not super familiar with the syntax of semantic conventions, but right now it feels like we define an attribute `profile.pprof.mapping` that can have one of `has_functions` / `has_filenames` / `has_line_numbers` / `has_inline_frames` values, while to be an equivalent to the pprof format's corresponding boolean fields, we need to be able to set each of these values independently.

aalexand:
Given https://github.com/open-telemetry/opentelemetry-proto/pull/690 and the discussion in the Slack, what do you think about making adding the is_folded attribute a part of this PR as well? I think it might be easier and more convenient to review adding all of these at once. This will help ensure consistency in naming and typing etc.

florianl:
Cherry-picked and so added the is_folded change here as well.

model/profile/registry.yaml
thompson-tomo: Unfortunately you can't have an enum array but you can do a string array.

thompson-tomo:
Hmm, I am not sure about this proposal here. Is it the idea that the collector would transform a profile message to become a span hence wanting attribute to represent the low level message properties? 

Keep in mind the idea of adding attributes to the registry is so that they can be the key & sometimes value used in the attributes property of the message ie https://github.com/open-telemetry/opentelemetry-proto/blob/8654ab7a5a43ca25fe8046e59dcd6935c3f76de0/opentelemetry/proto/profiles/v1development/profiles.proto#L114

If you are wanting to define a profile message hence using attributes to indicate if that message should have functions for instance I would recommend you raise an issue weaver listing what are the properties of a profile.

It might help to define how these attributes would be used,

model/profile/registry.yaml
florianl: string array is also not possible if well-known values are defined.

florianl: I have updated the PR so that it now uses dedicated `profile.pprof.mapping.has_*` attributes of type boolean.

model/profile/common.yaml
```diff
@@ -6,3 +6,19 @@ groups:
     attributes:
       - ref: profile.frame.type
         requirement_level: recommended
+
+  - id: profile.pprof
+    type: attribute_group
+    brief: >
+      Describes the relationship between pprof and a Profile.
+    attributes:
+      - ref: profile.pprof.mapping.has_functions
```
beaubelgrave: Is this only available for pprof? If we want a semantic meaning of a mapping has functions within it (IE: It's executable or was hit during sampling) why limit it to pprof? I see a previous discussion happened about this, but I am missing context.

Our sampling profilers will emit out OTel profiling data, and I'm not sure if these will be used broadly or only within pprof translation cases?

florianl: The idea of these attributes is to provide compatibility between pprof and OTel profiling.
Originally, these attributes were dedicated fields in messages of pprof. As their counter part in OTel profiling no longer exists, the idea is to have them as attributes in SemConv so that conversions between OTel profiling and pprof (and vice versa) can happen.

aalexand (approved):

aalexand:
@jhalliday @christos68k to take a look also

model/profile/common.yaml
christos68k: These fields are mostly useful to processing pipelines using pprof. I think that besides naming, we're neither over-specifying nor restricting usage here, so any producer could fill these in if deemed useful for the problem at hand.

docs/registry/attributes/profile.md
```diff
@@ -29,3 +32,15 @@ Describes the origin of a single frame in a Profile.
 | `ruby` | [Ruby](https://wikipedia.org/wiki/Ruby_(programming_language)) | ![Development](https://img.shields.io/badge/-development-blue) |
 | `rust` | [Rust](https://wikipedia.org/wiki/Rust_(programming_language)) | ![Development](https://img.shields.io/badge/-development-blue) |
 | `v8js` | [V8JS](https://wikipedia.org/wiki/V8_(JavaScript_engine)) | ![Development](https://img.shields.io/badge/-development-blue) |
+
+## Pprof specific attributes for the Profiling signal
+
+Attributes specific to pprof that helps to convert Profiling signals.
```
christos68k:
```suggestion
Attributes specific to pprof that help convert from pprof to Profiling signal.
```

docs/registry/attributes/profile.md
```diff
@@ -29,3 +32,15 @@ Describes the origin of a single frame in a Profile.
...
+| <a id="profile-pprof-mapping-has-filenames" href="#profile-pprof-mapping-has-filenames">`profile.pprof.mapping.has_filenames`</a> | boolean | Indicates that there are filenames related to this mapping. |  | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="profile-pprof-mapping-has-functions" href="#profile-pprof-mapping-has-functions">`profile.pprof.mapping.has_functions`</a> | boolean | Indicates that there are functions related to this mapping. |  | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="profile-pprof-mapping-has-inline-frames" href="#profile-pprof-mapping-has-inline-frames">`profile.pprof.mapping.has_inline_frames`</a> | boolean | Indicates that there are inlined frames related to this mapping. |  | ![Development](https://img.shields.io/badge/-development-blue) |
```
christos68k: Nit: For consistency in terminology, we've so far used "inlined functions" but "inline frames".

```suggestion
| <a id="profile-pprof-mapping-has-inline-frames" href="#profile-pprof-mapping-has-inline-frames">`profile.pprof.mapping.has_inline_frames`</a> | boolean | Indicates that there are inline frames related to this mapping. |  | ![Development](https://img.shields.io/badge/-development-blue) |
```

model/profile/registry.yaml
```diff
@@ -78,3 +78,38 @@ groups:
...
+    display_name: Pprof specific attributes for the Profiling signal
+    brief: >
+      Attributes specific to pprof that helps to convert Profiling signals.
```
christos68k:
```suggestion
      Attributes specific to pprof that help convert from pprof to Profiling signal.
```

model/profile/common.yaml
```diff
@@ -6,3 +6,19 @@ groups:
     attributes:
       - ref: profile.frame.type
         requirement_level: recommended
+
+  - id: profile.pprof
+    type: attribute_group
+    brief: >
+      Describes the relationship between pprof and a Profile.
```
christos68k: Why not use the same brief description as in registry below, I think it's more specific and less vague:

```suggestion
      Attributes specific to pprof that help convert from pprof to Profiling signal.
```

model/profile/registry.yaml
```diff
@@ -78,3 +78,38 @@ groups:
...
+        type: boolean
+        stability: development
+      - id: profile.pprof.mapping.has_line_numbers
+        brief: >
+          Indicates that there are line numbers related to this mapping.
+        type: boolean
+        stability: development
+      - id: profile.pprof.mapping.has_inline_frames
+        brief: >
+          Indicates that there are inlined frames related to this mapping.
```
christos68k:
```suggestion
          Indicates that there are inline frames related to this mapping.
```

christos68k (approved):

model/profile/registry.yaml
```diff
@@ -78,3 +78,38 @@ groups:
                 [Rust](https://wikipedia.org/wiki/Rust_(programming_language))
               value: "rust"
               stability: development
+  - id: registry.profile.pprof
+    type: attribute_group
+    display_name: Pprof specific attributes for the Profiling signal
+    brief: >
+      Attributes specific to pprof that help convert from pprof to Profiling signal.
+    attributes:
+      - id: profile.pprof.mapping.has_functions
```
lmolkova: any reason to prefix with`profile` ? 

Following 

https://github.com/open-telemetry/semantic-conventions/blob/1cf2476ae5e518225a766990a28a6d5602bd5a30/docs/general/naming.md?plain=1#L95-L97

I'd suggest `pprof.mapping.*` for this and other attributes.

model/profile/registry.yaml
```diff
@@ -78,3 +78,38 @@ groups:
...
+        brief: >
+          Indicates that there are line numbers related to this mapping.
+        type: boolean
+        stability: development
+      - id: profile.pprof.mapping.has_inline_frames
+        brief: >
+          Indicates that there are inline frames related to this mapping.
+        type: boolean
+        stability: development
+      - id: profile.pprof.location.is_folded
```
lmolkova: In other conventions, we usually skip `is_` unless we're representing an external API or concept that has it, so the naming suggestion would be

```suggestion
      - id: profile.pprof.location.folded
```

docs/registry/attributes/profile.md
```diff
@@ -29,3 +32,15 @@ Describes the origin of a single frame in a Profile.
 | `ruby` | [Ruby](https://wikipedia.org/wiki/Ruby_(programming_language)) | ![Development](https://img.shields.io/badge/-development-blue) |
 | `rust` | [Rust](https://wikipedia.org/wiki/Rust_(programming_language)) | ![Development](https://img.shields.io/badge/-development-blue) |
 | `v8js` | [V8JS](https://wikipedia.org/wiki/V8_(JavaScript_engine)) | ![Development](https://img.shields.io/badge/-development-blue) |
+
+## Pprof specific attributes for the Profiling signal
```
lmolkova: nit, maybe 

```suggestion
## pprof Attributes
```

docs/registry/attributes/profile.md
```diff
@@ -29,3 +32,15 @@ Describes the origin of a single frame in a Profile.
...
+
+Attributes specific to pprof that help convert from pprof to Profiling signal.
```
lmolkova: It seems like these attributes are used when converting pprof to OTel profiling signal. Is there more information on this conversion - something we can link to? Or could we describe it in general? It does not have to be in the PR, but creating a work item and adding TODO here might help readers understand it's coming and a way to monitor progress.

joaopgrassi: Right, something like we have for Prometheus <> OTLP: https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#prometheus-metric-points-to-otlp

model/profile/registry.yaml
florianl: As the brief introduction says, these attributes exist to help converting pprof to Profiling signal and vice versa and so are directly related on how the protocol should be handled. In the scope of other signals, like logs, traces, events or metrics, these attributes are not applicable.

@lmolkova do you still prefer to use `pprof.*` over `profile.pprof.*`?

model/profile/registry.yaml
florianl: These attributes are directly tied to an external API/protocol.
https://github.com/google/pprof/blob/5eec23c92c42b9ccc56b658247e66e35662e2582/proto/profile.proto#L204-L209

docs/registry/attributes/profile.md
florianl: > Is there more information on this conversion - something we can link to?

For the moment, there is just the convertibility promise as stated in https://github.com/open-telemetry/opentelemetry-proto/issues/567#issuecomment-2286565449. With the release of OTel proto v1.8.0, we want to contribute a pprof <-> OTel profiling processor. But this is still work in progress. And for this processor these attributes are essential, otherwise information will get lost in the conversion of pprof <-> OTel profiling.

florianl: Added a comment around convertibility in https://github.com/open-telemetry/semantic-conventions/pull/2522/commits/a519a7547635b54cd4e43dc6fe90c50114a4b1c0

model/profile/registry.yaml
lmolkova: SemConv SIG meeting notes: profle prefix is not necessary - we don't include signals into attribute name.

florianl: @lmolkova  updated the namespace with https://github.com/open-telemetry/semantic-conventions/pull/2522/commits/aab31a24cedb5c5d8de594af9d4f61e3c60e1062.

PR #2179, opened 2025-04-24
Using attributes for chat history on gen_ai spans and events

lmolkova:
Fixes #2010, #2179, #1913, #1621, #1912

Depends on #2588 

Key points:

1. System message is recorded as `gen_ai.system.instructions` attribute. It's not necessarily a part of chat history (when dealing with server-side agents)
2. Prompt is recorded as `gen_ai.input.messages` attribute, format is defined in json schema (for now until we have tooling to define complex attributes). If uploading to an external storage, it's replaced with `gen_ai.input.messages_ref`
3. Completion is recorded as `gen_ai.output.messages` (or `gen_ai.output.messages_ref`)
4. These attributes can be recorded on spans or on the new event `gen_ai.client.inference.operation.details` which captures all the details of the inference operation so that it can be useful if span is sampled out and/or if queries between spans and logs backends is problematic.

Depends on #2046 (`gen_ai.system` rename and migration plan)

Issue #1556, opened 2024-11-07
Support for Multi modal inputs and generations


karthikscale3:
### Area(s)

area:gen-ai

### Is your change request related to a problem? Please describe.

This relates to the support for multi modal content - images/audio etc. (inputs and generation). The current GenAI spec does not have support for multi modal inputs and generations as there needs to be additional considerations around the increase the size of the span. As a result, an external blob storage is being considered for storage and hyperlinking the stored location back to the span. 

The main task relates to:
- Replace content attribute in event with a similar structure to OpenAI - array of content

Some additional comments and open questions discussed during the SIG meeting:
- We should add a hook in the instrumentation ([simiar to OpenLLMetry instrumentation](https://github.com/traceloop/openllmetry/blob/85a9e82a570263c78445db281ff8c21f58423bed/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py#L351))
- Semantic convention for references - https://github.com/open-telemetry/semantic-conventions/pull/1521
- What metadata should we stored with the blob that is stored in the bucket?

### Describe the solution you'd like

The main task relates to:
- Replace content attribute in event with a similar structure to OpenAI - array of content

lmolkova:
Related to https://github.com/open-telemetry/semantic-conventions/issues/963

alexmojaki:
Superset of this proposal: https://github.com/open-telemetry/semantic-conventions/issues/1913

alexmojaki:
<details><summary>Examples from provider docs</summary>
<p>

https://platform.openai.com/docs/guides/pdf-files?api-mode=responses

```python
{
    "type": "input_file",
    "file_url": "https://www.berkshirehathaway.com/letters/2024ltr.pdf",
}
{
    "type": "input_file",
    "file_id": file.id,
}
{
    "type": "input_file",
    "filename": "draconomicon.pdf",
    "file_data": f"data:application/pdf;base64,{base64_string}",
}
```

https://platform.openai.com/docs/guides/images-vision

```python
{
    "type": "input_image",
    "image_url": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg",
}
{
    "type": "input_image",
    "file_id": file_id,
}
{
    "type": "input_image",
[...]

PR #2528, opened 2025-07-16
New attributes in execute-tool and llm span for Single and Multi-Agent traceability

ShipraJain01:
Fixes #

## Changes

This PR extends OpenTelemetry semantic conventions to capture of inputs, outputs, and tool definitions in GenAI Single-Agent and Multi-Agent systems.

###Proposed Additions:
Add tool definitions and execution details for Single and Multi-Agent systems.
  - Add 'gen_ai.tool_definitions' attribute to 'gen_ai.invoke_agent' span
  - Add 'gen_ai.tool_definitions' attribute to 'attributes.gen_ai.inference.client' attribute group. 
  - Add 'gen_ai.tool.call.arguments' and 'gen_ai.tool.call.result' attributes to 'gen_ai.execute_tool' span.

Prototype in Semantic Kernel: [Link](https://github.com/ShipraJain01/semantic-kernel/commit/36daa3644d28fe813c71fa268baf33aa20866acb#diff-d3cae2f7f8f407d2d6143bd58643c49ff2e1b98c7eba4bae514f98a753cb9e17)

Prototype in Langgraph: [Link](https://github.com/nagkumar91/langchain-azure/tree/main/libs/azure-ai/langchain_azure_ai/callbacks/tracers)


model/gen-ai/registry.yaml
```diff
@@ -103,6 +103,20 @@ groups:
         type: string
         brief: The name of the GenAI model a request is being made to.
         examples: 'gpt-4'
+
+      - id: gen_ai.model.version
```
lmolkova: from the offline discussion: the value is the same as `gen_ai.response.model` and it's not known at the request time.

model/gen-ai/registry.yaml
```diff
@@ -230,6 +244,12 @@ groups:
         type: string
         brief: Name of the tool utilized by the agent.
         examples: ["Flights"]
+      - id: gen_ai.tool.version
+        stability: development
+        type: string
+        brief: Version of the tool utilized by the agent.
+        note: "include version if tool schema exposes it."
```
lmolkova: are there any tool schemas that expose it or any real-life scenario when we can populate it?

I don't believe it's available in MCP tool, OpenAI tool definition, Semantic Kernel or LangChain.

model/gen-ai/registry.yaml
```diff
@@ -251,6 +271,63 @@ groups:
             Client-side operations are actions taken on the user's end or within the client application.
           Datastore: A tool used by the agent to access and query structured or unstructured external data for retrieval-augmented tasks or knowledge updates.
         examples: ['function', 'extension', 'datastore']
+
+      - id: gen_ai.tool.call.arguments
+        stability: development
+        type: any
+        brief: Parameters passed to the tool call.
+        note: >
+          TODO: we need to decide what to do with well-known function call params
```
lmolkova: Are we going to use this attribute to record OpenAI [file_search params](https://platform.openai.com/docs/api-reference/responses/object) and similar? 

Image

or is it limited to client-side function execution?

model/gen-ai/registry.yaml
```diff
@@ -251,6 +271,63 @@ groups:
...
+          (eg. filesearch) - do we want to define a format and normalize across
+          models/frameworks or leave it as is and record a complex object.
+          If we choose to support standardized formats, the complex object can carry a type discriminator along with the value, with the following guidance:
+          - For `text` type: Plain text string containing the tool's response
+          - For `json` type: Valid JSON string that can be parsed into an object or array
+          - For `table` type: JSON-stringified array of objects (recommended), Markdown table, or CSV-formatted string
+          - For `html` type: Valid HTML string containing structured markup
+          - For `image` type: Base64-encoded image string
+          - For `image_url` type: URL string pointing to an externally hosted image
+          - For `chart` type: Either base64-encoded chart image or code/data structure to render the chart
```
lmolkova: are we going to use this attribute to record details about file_search and other server-side functions?

If not and it's limited to client-side function execution, we won't be able to populate any of those. 
We can record the output object, but we have no means to tell it's an image_url or table. It's just a complex object of unknown nature.

lmolkova:
@ShipraJain01 thank you! could you please sign the CLA ? I believe it's recommended to use individual contributor path for MS employees.

model/gen-ai/registry.yaml
ShipraJain01: we should be able to accept all kinds of tools.

model/gen-ai/registry.yaml
ShipraJain01: You are right that from telemetry layer's perspective, especially for server-side functions like file_search , we often don't have understanding of result's content structure (e.g., whether it's a table, image, or HTML). The intent here is to provide guidance, not impose structure.

We’re suggesting that when teams do know the nature of the tool output (either because it’s client-side, simulated, or they're enriching it), they can still record it under a common attribute (gen_ai.tool.call.result) with a known type (via a discriminator inside the complex object).

This allows for:
Standardization where possible, improving downstream processing (e.g., analytics or trace visualization).

Flexibility, so tools like file_search, image_generation, or custom internal tools can still store raw outputs in the same attribute even when the type isn't known at collection time.

We’re not expecting all implementations to support full content introspection — just giving a path for those who can.

model/gen-ai/spans.yaml
```diff
@@ -284,6 +317,15 @@ groups:
...
           recommended: if available
+      - ref: gen_ai.tool.type
+        requirement_level:
+          recommended: if available
+      - ref: gen_ai.tool.call.arguments
+        requirement_level:
+          conditionally_required: if available
+      - ref: gen_ai.tool.call.result
+        requirement_level:
+          conditionally_required: opt_in
```
91pavan: A lot of these is already being captured by `gen_ai.choice` and `gen_ai.tool.message` event, if i'm not wrong. Can't we reuse that? It also has `function definitions`. 

https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aichoice

model/gen-ai/spans.yaml
```diff
@@ -89,6 +89,33 @@ groups:
...
+          Due to potential size and sensitive content, population should be opt-in.
+      
+      - ref: gen_ai.agent.child_agents
+        requirement_level:
+          conditionally_required: if applicable
+        note: |
+          Information about child agents utilized during execution.
+          This attribute should be populated when the agent orchestrates or delegates
+          to other specialized agents as part of its workflow.
+
```
91pavan: Would it make more sense if the agent parent-child relationships can be captured via OTel `SpanContext`? The parent-child relationship can be implicit from `traceId`, `spanId`, `parentSpanId` tuple?  This would anyway capture the agent hierarchies or delegation chains in a better manner.

This could aid in capturing metadata (which needs to be done manually), but capturing this info can be somewhat hard to generalize among frameworks, from our research.  It can also be very hard to serialize in dynamic graphs, and you may lose ordering & task lineage.

lmolkova: to add to @91pavan point - correlated telemetry would show what agents were de-facto called in scope of this agent invocation and would include agents called through tools / remote agents called via RPC / HTTP.

In generic case, the info about sub-agents is not available and capturing it under `gen_ai.agent` could be misleading, especially when it'd coexist with other agents called via mcp/tools. 

It might make sense under `gen_ai.orchestrator.agent_definitions` or similar and then would be unambiguous.  We'd only populate it on orchestrator spans.

model/gen-ai/spans.yaml
lmolkova: yes, but there are caveats. Let's assume there is a bug in the code and 
- tool output is not passed over to the agent correctly
- it's post-processed and then added to the chat history in modified form
- the agent is never called again because of some error (e.g. because tool output is in unexpected format)

Telemetry is most critical in edge cases like these. 

I agree it's overly verbose though. Perhaps we should consider a mode when tool args/outputs are not recorded, but all the chat history is on to be the one we'd recommend unless users want it all? 

From semconv perspective, we don't document this level of details, we just say that the attribute/event is opt-in.

model/gen-ai/registry.yaml
```diff
@@ -251,6 +253,172 @@ groups:
...
+              "body": [
+                {
+                  "type": "text",
+                  "content": "Plan a trip to Bali for 5 days for a family of 4."
+                }
+              ]
+            }
+          ]
+
+      - id: gen_ai.agent.invocation_output
```
lmolkova: do you think it'd converge with https://github.com/open-telemetry/semantic-conventions/blob/edb8ec8b329b75d7ea87aaaff88175583231feae/docs/gen-ai/non-normative/models.ipynb if/when https://github.com/open-telemetry/semantic-conventions/pull/2179 gets in? Is it intentional that the data structure is different between agents and non-agents?

model/gen-ai/registry.yaml
lmolkova: the questions is if we need to document the type of `gen_ai.tool.call.arguments` for `file_search`. If we're going to capture functions like file_search, we'll need to decide which of the properties above go to arguments and how.

E.g. for file_search call with `queries = ["foo", "bar"]`, I'd expect something like  `gen_ai.tool.call.arguments = {"queries" : ["foo", "bar"]}`, `gen_ai.tool.call.result = [{"file_id": "file1", "filename": "file1.md", "score" : 42, "text"  : "something important"}]`.   

We don't have to document all possible tools right away, but we need to define if we do this and then document the approach with an example or two.

model/gen-ai/registry.yaml
lmolkova: User applications are most usually instrumented by a generic instrumentation. I don't think we can rely on this attribute being present or that user applications would populate it consistently. It's rarely the case. It'd be easier if we moved all suggestions we'd have for application-code to a separate PR - this would allow us to focus on the 80%+ of the user apps and the experience we can provide out of the box.

model/gen-ai/spans.yaml
ShipraJain01: thanks for the feedback. Let me look at `gen_ai.orchestrator.agent_definitions` proposal.

model/gen-ai/registry.yaml
ShipraJain01: Yes, `gen_ai.agent.invocation_output` should get covered in #2179 PR @lmolkova.

model/gen-ai/registry.yaml
ShipraJain01: @lmolkova :  If we're only instrumenting client-side tools, then for server-side tools like file_search, you don’t have visibility into the content of the response. so you won’t be able to say if it's a table, image, or just some HTML. I can remove the details there and keep the `gen_ai.tool.call.result` as type any.

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -193,16 +197,61 @@ Semantic conventions for individual GenAI systems and frameworks MAY specify dif
 | [`gen_ai.operation.name`](/docs/registry/attributes/gen-ai.md) | string | The name of the operation being performed. [1] | `chat`; `generate_content`; `text_completion` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`gen_ai.provider.name`](/docs/registry/attributes/gen-ai.md) | string | The Generative AI provider as identified by the client or server instrumentation. [2] | `openai`; `gcp.gen_ai`; `gcp.vertex_ai` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`error.type`](/docs/registry/attributes/error.md) | string | Describes a class of error the operation ended with. [3] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` if the operation ended in an error | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.agent.child_agents`](/docs/registry/attributes/gen-ai.md) | any | Information about child agents utilized during the agent invocation. [4] | `[
```
TaoChenOSU: This may have overlap with the newly proposed `gen_ai.tool.definitions` attribute to the span. When an agent delegates a task to a sub agent, it's usually done via a tool call.

docs/gen-ai/gen-ai-spans.md
```diff
@@ -68,12 +68,57 @@ client or when the GenAI call happens over instrumented protocol such as HTTP.
...
 | [`error.type`](/docs/registry/attributes/error.md) | string | Describes a class of error the operation ended with. [3] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` if the operation ended in an error | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
-| [`gen_ai.conversation.id`](/docs/registry/attributes/gen-ai.md) | string | The unique identifier for a conversation (session, thread), used to store and correlate messages within this conversation. [4] | `conv_5j66UpCpwteGg4YSxUnt7lPY` | `Conditionally Required` when available | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`gen_ai.output.type`](/docs/registry/attributes/gen-ai.md) | string | Represents the content type requested by the client. [5] | `text`; `json`; `image` | `Conditionally Required` [6] | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.agent.child_agents`](/docs/registry/attributes/gen-ai.md) | any | Information about child agents utilized during the agent invocation. [4] | `[
```
TaoChenOSU: Is "agent" applicable in a model span?

docs/gen-ai/gen-ai-spans.md
```diff
@@ -68,12 +68,57 @@ client or when the GenAI call happens over instrumented protocol such as HTTP.
...
-| [`server.port`](/docs/registry/attributes/server.md) | int | GenAI server port. [8] | `80`; `8080`; `443` | `Conditionally Required` If `server.address` is set. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.tool.definitions`](/docs/registry/attributes/gen-ai.md) | any | The list of tool definitions available to the GenAI agent or model. [9] | `[{"type":"function","name":"get_current_weather","description":"Get the current weather in a given location","parameters":{"type":"object","properties":{"location":{"type":"string","description":"The city and state, e.g. San Francisco, CA"},"unit":{"type":"string","enum":["celsius","fahrenheit"]}},"required":["location","unit"]}}]` | `Conditionally Required` opt_in | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`server.port`](/docs/registry/attributes/server.md) | int | GenAI server port. [10] | `80`; `8080`; `443` | `Conditionally Required` If `server.address` is set. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.agent.invocation_input`](/docs/registry/attributes/gen-ai.md) | any | The input messages provided to the agent invocation. [11] | `[
```
TaoChenOSU: Is "agent" applicable in a model span?

docs/gen-ai/gen-ai-spans.md
```diff
@@ -68,12 +68,57 @@ client or when the GenAI call happens over instrumented protocol such as HTTP.
...
+    "role": "user", 
+    "body": [
+      {
+        "type": "text",
+        "content": "Plan a trip to Bali for 5 days for a family of 4."
+      }
+    ]
+  }
+]` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.agent.invocation_output`](/docs/registry/attributes/gen-ai.md) | any | The output messages generated by the agent invocation. [12] | `[
```
TaoChenOSU: Is "agent" applicable in a model span?

model/gen-ai/spans.yaml
ShipraJain01: adopting gen_ai.orchestrator.agent_definitions. Pushing change in next PR

model/gen-ai/spans.yaml
ShipraJain01: This change is proposed to capture `execute_tool` span's input and output aptly. Taking reference from PRs #2179 which highlights issues in using events for input and output and emphasizes the use of attributes.

docs/gen-ai/gen-ai-agent-spans.md
ShipraJain01: @TaoChenOSU: calling agent as a tool, is it not framework specific? like OpenAI SDK? I have seen the opportunity to register sub-agent to main agent which get considered as agents itself (eg: Azure AI Services). Adding Agent definitions within Tool definition would lead to confusing schema which would differ for both `tool` vs `agent`. Isn't it.

model/gen-ai/registry.yaml
ShipraJain01: removed the version in latest commit

model/gen-ai/registry.yaml
ShipraJain01: removed this param in new commit

model/gen-ai/registry.yaml
ShipraJain01: @lmolkova : let me add the example here

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -77,8 +77,9 @@ Semantic conventions for individual GenAI systems and frameworks MAY specify dif
...
-| [`server.port`](/docs/registry/attributes/server.md) | int | GenAI server port. [5] | `80`; `8080`; `443` | `Conditionally Required` If `server.address` is set. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
-| [`server.address`](/docs/registry/attributes/server.md) | string | GenAI server address. [6] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.tool.definitions`](/docs/registry/attributes/gen-ai.md) | any | The list of tool definitions available to the GenAI agent or model. [5] | `[{"type":"function","name":"get_current_weather","description":"Get the current weather in a given location","parameters":{"type":"object","properties":{"location":{"type":"string","description":"The city and state, e.g. San Francisco, CA"},"unit":{"type":"string","enum":["celsius","fahrenheit"]}},"required":["location","unit"]}}]` | `Conditionally Required` opt_in | ![Development](https://img.shields.io/badge/-development-blue) |
```
TaoChenOSU: Does this only apply to local functions? i.e. do we need to record the code interpreter if it's available and enabled? For example, the OpenAI code interpreter tool: https://platform.openai.com/docs/assistants/tools/code-interpreter

docs/gen-ai/gen-ai-agent-spans.md
TaoChenOSU: In this documentation by OpenAI: https://openai.github.io/openai-agents-python/handoffs/, it states that "Handoffs are represented as tools to the LLM. So if there's a handoff to an agent named Refund Agent, the tool would be called transfer_to_refund_agent."

I am curious how Azure AI realizes the connected agent feature underneath.

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -193,16 +197,61 @@ Semantic conventions for individual GenAI systems and frameworks MAY specify dif
...
+| [`gen_ai.tool.definitions`](/docs/registry/attributes/gen-ai.md) | any | The list of tool definitions available to the GenAI agent or model. [10] | `[{"type":"function","name":"get_current_weather","description":"Get the current weather in a given location","parameters":{"type":"object","properties":{"location":{"type":"string","description":"The city and state, e.g. San Francisco, CA"},"unit":{"type":"string","enum":["celsius","fahrenheit"]}},"required":["location","unit"]}}]` | `Conditionally Required` opt_in | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`server.port`](/docs/registry/attributes/server.md) | int | GenAI server port. [11] | `80`; `8080`; `443` | `Conditionally Required` If `server.address` is set. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.agent.invocation_input`](/docs/registry/attributes/gen-ai.md) | any | The input messages provided to the agent invocation. [12] | `[
+  {
+    "role": "system",
+    "body": [
```
TaoChenOSU: Should this be "content" instead of "body"? "content" would match the existing input&output event schema.

docs/gen-ai/gen-ai-agent-spans.md
ShipraJain01: I understand. However, I see OpenAI's Hand-off calling agent calls as tool is a special case. Not all frameworks have adopted this representation. It would depend how the child_agent's span is initiated. we can assume 2 options - 

1.  parent invoke_agent span > execute_tool span > child invoke_agent span? 
2. parent invoke_agent span > child invoke_agent span. 

 I would propose to use `gen_ai.orchestrator.agent_definitions` as eventually we are calling invoke_agent span for the child agent and its upto frameworks what path amongst 1 or 2 they choose to implement in underlying core engine.

docs/gen-ai/gen-ai-spans.md
ShipraJain01: No, my bad. This got added mistakenly. Fixing it in next PR

model/gen-ai/registry.yaml
ShipraJain01: Adding appropriate example in next PR upgrade.

docs/gen-ai/gen-ai-spans.md
ShipraJain01: Thanks for pointing out. Mistakenly added it here. Fixing in next PR

docs/gen-ai/gen-ai-spans.md
ShipraJain01: Thanks for pointing out. Mistakenly added it here. Fixing in next PR

docs/gen-ai/gen-ai-agent-spans.md
ShipraJain01: Same structure can work for code interpreter as well. Let me add an example.

docs/gen-ai/gen-ai-agent-spans.md
ShipraJain01: We are trying to follow either Open AI format (which says body) or Google's (which says parts). I am converging this parameter to gen_ai.input.messages and gen_ai.output.messages in #2179 PR. So adopting 'parts' terminology

docs/gen-ai/gen-ai-spans.md
Cirilla-zmh: Perhaps you should avoid using line breaks in markdown tables — they can cause markdown formatting errors. You can refer to this commit for an example: https://github.com/open-telemetry/semantic-conventions/pull/2179/commits/50f95437276cbfcfcd20ccdff00454240d491f14

docs/gen-ai/gen-ai-spans.md
```diff
@@ -68,12 +68,57 @@ client or when the GenAI call happens over instrumented protocol such as HTTP.
...
 | [`gen_ai.request.seed`](/docs/registry/attributes/gen-ai.md) | int | Requests with same seed value more likely to return same result. | `100` | `Conditionally Required` if applicable and if the request includes a seed | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`server.port`](/docs/registry/attributes/server.md) | int | GenAI server port. [8] | `80`; `8080`; `443` | `Conditionally Required` If `server.address` is set. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.tool.definitions`](/docs/registry/attributes/gen-ai.md) | any | The list of tool definitions available to the GenAI agent or model. [9] | `[{"type":"function","name":"get_current_weather","description":"Get the current weather in a given location","parameters":{"type":"object","properties":{"location":{"type":"string","description":"The city and state, e.g. San Francisco, CA"},"unit":{"type":"string","enum":["celsius","fahrenheit"]}},"required":["location","unit"]}}]` | `Conditionally Required` opt_in | ![Development](https://img.shields.io/badge/-development-blue) |
```
Cirilla-zmh: `gen_ai.tool.definitions` is a more detailed and sensitive attribute that shouldn't be collected by default. Do you think we should have an attribute named `gen_ai.tool.names` solely for capturing the tool names provided to model?

In most cases, the other content in `tool.definitions` is static and can be easily retrieved through association. Therefore, I believe it is reasonable to capture only the tool names, and this attribute could be marked as recommended.

docs/gen-ai/gen-ai-spans.md
Cirilla-zmh: Sorry for any possible misunderstandings. Have you considered separating the agent span definitions from the inference? Or do you think the current structure is reasonable as it is?

docs/gen-ai/gen-ai-agent-spans.md
Cirilla-zmh: +1

docs/gen-ai/gen-ai-spans.md
```diff
@@ -306,14 +366,42 @@ by the application code.
...
 
 **[1] `error.type`:** The `error.type` SHOULD match the error code returned by the Generative AI provider or the client library,
 the canonical name of exception that occurred, or another low-cardinality error identifier.
 Instrumentations SHOULD document the list of errors they report.
 
+**[2] `gen_ai.tool.call.arguments`:** TODO: we need to decide what to do with well-known function call params (eg. filesearch) - do we want to define a format and normalize across models/frameworks or leave them as is and record a complex object.
+> [!WARNING] > This attribute may contain sensitive information.
+It's expected to be an object - in case a serialized string is available to the instrumentation, the instrumentation SHOULD do the best effort to deserialize it to an object.
+
+**[3] `gen_ai.tool.call.result`:** TODO: we need to decide what to do with well-known function call results
```
Cirilla-zmh: I believe we won't be able to reach an agreement on those multimodal results until #1556 is resolved. Perhaps we should focus on the text results for now.

Cirilla-zmh (commented):
Thank you! Tool arguments and results are important attributes that we are also carefully addressing. We're more than happy to engage in discussions!

docs/gen-ai/gen-ai-spans.md
ShipraJain01: thanks for highlighting it @Cirilla-zmh . I just fixed this issue in my latest push :)

ShipraJain01: thanks for highlighting it @Cirilla-zmh . I just fixed this issue in my latest push..

docs/gen-ai/gen-ai-spans.md
ShipraJain01: @Cirilla-zmh, Updated the changes in latest PR. Also included samples for function call, file search and code interpreter as per the comments received from other reviewers.

model/gen-ai/registry.yaml
```diff
@@ -251,6 +253,219 @@ groups:
...
+                            {
+                              "type": "text",
+                              "content": "The weather in Paris is currently rainy with a temperature of 57°F."
+                            }
+                          ],
+                          "finish_reason": "stop"
+                        }
+                      ]
+
+      - id: gen_ai.orchestrator.agent_definitions
```
gyliu513, line 373: Can we split this to two ?

- gen_ai.orchestrator.agent_registry, this will represent which agents are registered to the orchestrator, and this should be used when creating agent
- gen_ai.orchestrator.agents_invoked: This is used at run time when agents are being called by the orchestrator

gyliu513:
@PRATIBHA-Moogi ^^

docs/gen-ai/gen-ai-spans.md
ShipraJain01: @Cirilla-zmh  thanks for the feedback. Few point of views - 

1. gen_ai.tool.definitions attribute explicitly states "Opt-in by default" and "Instrumentations MAY provide a way to enable populating this attribute." This means it's already designed to be opt-in rather than collected by default.
2. Adding a separate gen_ai.tool.names attribute might create confusion with the existing gen_ai.tool.name (singular) and would duplicate information already available in gen_ai.tool.definitions (when enabled).Flexibility for different use cases:
3. It basically gives flexibility for users who need to audit/debug tool configurations in that case they can opt-in to collect gen_ai.tool.definitions, else skip storing this information. Additionally, I would take reference from already merged PR [https://github.com/open-telemetry/semantic-conventions/pull/2179] and would recommend using gen_ai.tool.definitions_ref if uploading to an external storage.

Cirilla-zmh: @ShipraJain01 

> Adding a separate gen_ai.tool.names attribute might create confusion with the existing gen_ai.tool.name (singular) and would duplicate information already available in gen_ai.tool.definitions (when enabled).Flexibility for different use cases:

You're right. The naming needs more discussion. However, what I meant is that capturing the list of tools sent to the model is necessary—and ideally, this should be collected by default, similar to how stop sequences are handled. This is because, for an LLM client, the tool list provided in each call may vary, and having this information can be valuable for troubleshooting.

On the other hand, the detailed definitions of the tools are more verbose and can remain optional by default, as you mentioned. :)

ShipraJain01: @Cirilla-zmh : We can manage it even in current form right. Only maintain 'type' and 'name' key value pair to avoid adding sensitive 'parameters' definition. Tool definitions is complex object but no fixed schema. So end user can achieve both scenarios thru same attribute. See example below - 

              {
                "type": "function",
                "name": "get_current_weather",
                "description": "Get the current weather in a given location",
                "parameters": {
                  "type": "object",
                  "properties": {
                    "location": {
                      "type": "string",
                      "description": "The city and state, e.g. San Francisco, CA"
                    },
                  },
                }
              }

vs 

{
  "type": "function",
  "name": "get_current_weather",
}

model/gen-ai/registry.yaml
ShipraJain01: @gyliu513 : great point. We did discuss this during designing the attributes and also with extended OTel Community. Ideally, we would need to provide the list of agents registered (or available) for Orchestrator agent to select from. That's what is captured in 'gen_ai.orchestrator.agent_definitions'. 

Agents invoked should be derived using child 'invoke_agent' spans created during runtime. Separately having an attribute to store this information seemed redundant.

.chloggen/2528.yaml
```diff
@@ -0,0 +1,7 @@
+change_type: enhancement
+component: gen-ai
+note: | 
+  Add new attributes to 'invoke_agent' and 'execute_tool' span to capture input and output definition for Multi-Agentic systems.
```
alexmojaki, line 4: This only says Multi-Agentic, but it applies to single agents too, right?

model/gen-ai/spans.yaml
```diff
@@ -305,6 +315,14 @@ groups:
       - ref: gen_ai.tool.type
         requirement_level:
           recommended: if available
+
+      # New attributes added to execute_tool span
```
alexmojaki:
```suggestion
```

model/gen-ai/spans.yaml
```diff
@@ -305,6 +315,14 @@ groups:
...
+      - ref: gen_ai.tool.call.arguments
+        requirement_level:
+          conditionally_required: if available
+      - ref: gen_ai.tool.call.result
```
alexmojaki, line 315: I think these should be omitted if the user has not opted in to including content, they should be treated similarly to input/output messages.

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+        stability: development
+        type: any
+        brief: Definitions of agents available to the orchestrator for multi-agent workflows.
+        note: |
+          This attribute contains an array of objects representing agent definitions that are available to the orchestrator for potential use in multi-agent workflows.
+          
+          The structure is an array where each object contains:
+          
+          - 'agent_id': Unique identifier for the agent
+          - 'name': Human-readable name of the agent
```
alexmojaki, line 383: I imagine these should correspond to `gen_ai.agent.id` and `gen_ai.agent.name`, right?

But then here there's `role` which looks similar to `gen_ai.agent.description` but not quite the same. That doesn't feel right. Are applications expected to have both of these?

I think just `id` makes more sense than `agent_id`.

model/gen-ai/spans.yaml
```diff
@@ -276,6 +277,15 @@ groups:
         requirement_level: opt_in
       - ref: gen_ai.output.messages
         requirement_level: opt_in
+      - ref: gen_ai.tool.definitions
+        requirement_level: opt_in      
+      - ref: gen_ai.orchestrator.agent_definitions
+        requirement_level:
+          conditionally_required: if applicable 
+        note: |
+          Information about child agents utilized during execution.
```
alexmojaki, line 282: This note overrides the original longer note, making that one less likely to be seen. Is that intentional?

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+        note: >
+          The value of this attribute matches provider-specific tool definition
+          format.
+          It's expected to be an object - in case a serialized string is available
+          to the instrumentation, the instrumentation SHOULD do the best effort to
+          deserialize it to an object.
+
+          Since this attribute could be large, it's NOT RECOMMENDED to populate
+          it by default. Instrumentations MAY provide a way to enable
+          populating this attribute.
```
alexmojaki, line 341: More specifically, I would usually expect this to be treated the same as `gen_ai.system_instructions`. They're very similar and I imagine most users would want to populate either both or neither.

model/gen-ai/spans.yaml
```diff
@@ -276,6 +277,15 @@ groups:
         requirement_level: opt_in
       - ref: gen_ai.output.messages
         requirement_level: opt_in
+      - ref: gen_ai.tool.definitions
```
alexmojaki, line 276: In Pydantic AI, we record the tool definitions on the LLM spans. We don't currently record them on the agent spans. We want to, but one complication is that the tool definitions can actually change over the course of the agent run. That doesn't mean that it isn't possible to record something meaningful here, just wanted to point this out.

I do think it's strange though to be adding this to the agent span but not the LLM span.

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+              ],
+              "execution_time_ms": 3427
+            }
+
+      - id: gen_ai.tool.definitions
+        stability: development
+        type: any
+        brief: The list of tool definitions available to the GenAI agent or model.
+        note: >
+          The value of this attribute matches provider-specific tool definition
```
alexmojaki, line 333: I can see that there's value in recording the raw provider-specific data. But there's also value in a normalized conventional format that telemetry consumers can understand. So I think some discussion is needed here before we just go ahead with the one option. For example we might want to eventually have both, in which case we would need two names. We don't have to have that now, but should we maybe make this name more specific so that there's more 'room' to add another normalized attribute later?

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+        examples:
+          - 'For function: {"temperature_range": {"high": 75, "low": 60}, "conditions": "sunny"}'
+          - 'For file_search: {"file_id": "file1", "filename": "file1.md", "text": "something important"}'
+          - |
+            For code_interpreter with multiple outputs:
+            {
+              "outputs": [
+                {
+                  "type": "logs",
+                  "logs": "Analysis complete. Generated 3 visualizations and summary report."
```
alexmojaki:
```suggestion
                  "logs": "Analysis complete. Generated visualization and summary report."
```

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+      - id: gen_ai.tool.definitions
+        stability: development
+        type: any
+        brief: The list of tool definitions available to the GenAI agent or model.
+        note: >
+          The value of this attribute matches provider-specific tool definition
+          format.
+          It's expected to be an object - in case a serialized string is available
+          to the instrumentation, the instrumentation SHOULD do the best effort to
+          deserialize it to an object.
```
alexmojaki:
```suggestion
          It's expected to be an array of objects - in case a serialized string is available
          to the instrumentation, the instrumentation SHOULD do the best effort to
          deserialize it to an array.
```

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+              }
+            ]
+      # Add these new attributes to the registry.gen_ai group in registry.yaml
+
+      - id: gen_ai.orchestrator.agent_definitions 
+        stability: development
+        type: any
+        brief: Definitions of agents available to the orchestrator for multi-agent workflows.
+        note: |
+          This attribute contains an array of objects representing agent definitions that are available to the orchestrator for potential use in multi-agent workflows.
```
alexmojaki, line 378: The span is already by definition about invoking a specific agent, not e.g. asking a 'router' to select an agent. So I imagine the way that the existence of other agents becomes relevant is that the agent being invoked might call some 'handover' tool to switch to a different agent. In that case, the agent will already be provided all this information in tool definitions, making this a bit redundant. Is that right?

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+          
+          The structure is an array where each object contains:
+          
+          - 'agent_id': Unique identifier for the agent
+          - 'name': Human-readable name of the agent
+          - 'role': Classification of the agent (e.g., "specialist", "coordinator", "validator")
```
91pavan, line 384: One of the feedback I received in my now closed PR (#2551), was that name is sufficient and the role can be inferred from the name itself. As most users never do anything custom, less attributes the better IMO.

model/gen-ai/registry.yaml
gyliu513: @ShipraJain01 so you want to use `gen_ai.orchestrator.agent_definitions` to cover both, including the agent registry and agents invoked at run time?

model/gen-ai/spans.yaml
```diff
@@ -305,6 +315,14 @@ groups:
       - ref: gen_ai.tool.type
         requirement_level:
           recommended: if available
+
+      # New attributes added to execute_tool span
+      - ref: gen_ai.tool.call.arguments
+        requirement_level:
+          conditionally_required: if available
```
lmolkova, line 314: I believe it should be opt-in (see @alexmojaki comment below on content -
https://github.com/open-telemetry/semantic-conventions/pull/2528/files#r2300510710)

.chloggen/2528.yaml
ShipraJain01: yes @alexmojaki . Please allow me to edit it to Single and Multi-Agent systems

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+        brief: Parameters passed to the tool call.
+        note: >
+          > [!WARNING]
+          > This attribute may contain sensitive information.
+
+          It's expected to be an object - in case a serialized string is available
+          to the instrumentation, the instrumentation SHOULD do the best effort to
+          deserialize it to an object.
+        examples: 
+          - 'For function_call: {"location": "San Francisco?", "date": "2025-10-01"}'
```
lmolkova, line 271: please remove free-form text from the examples - we'd like to be able to validate examples and free form text would be a problem.

It's ok to keep it as yaml comments

model/gen-ai/registry.yaml
ShipraJain01: @gyliu513, well it would only cover definitions i.e. what's available in agent registry for Orchestrator agent to choose from. Agents invoked at runtime need not be captured in separate attribute as it would be duplicate information. That information can be derived by reading the child invoke_agent spans within orchestrator spans. So we decided to not capture this info separately.

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+          - 'For file_search: {"queries": ["semantic conventions for HTTP attributes", "Find all files containing OpenTelemetry database semantic conventions"]}'
```
lmolkova, line 272: We don't have concensus on how built-in tool would be represented in the chat history, so I'm not sure we're ready to define how they'd look on the `gen_ai.tool.call.arguments` attribute. Also, the execute tool span can't really be reported for the built-in tools (on the client side), so is there a reason to provide these examples at all?

model/gen-ai/registry.yaml
ShipraJain01: sure makes sense @alexmojaki. Let me change it to 'id'

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+      - id: gen_ai.tool.call.result
+        stability: development
+        type: any
+        brief: The result returned by the tool call (if any and if execution was successful).
+        note: |
+          > [!WARNING]
+          > This attribute may contain sensitive information.
+
+        examples:
+          - 'For function: {"temperature_range": {"high": 75, "low": 60}, "conditions": "sunny"}'
```
lmolkova, line 297: see another comment on free-form text in examples, let's replace it with yaml comments

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+          - 'For file_search: {"file_id": "file1", "filename": "file1.md", "text": "something important"}'
+          - |
+            For code_interpreter with multiple outputs:
```
lmolkova, line 300: since execute tool span can't really be reported for built-in tools on the client side, do we really need these examples?

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+                    }
+                  },
+                  "required": [
+                    "location",
+                    "unit"
+                  ]
+                }
+              }
+            ]
+      # Add these new attributes to the registry.gen_ai group in registry.yaml
```
lmolkova, line 371: is this comment still relevant?

model/gen-ai/spans.yaml
```diff
@@ -276,6 +277,15 @@ groups:
         requirement_level: opt_in
       - ref: gen_ai.output.messages
         requirement_level: opt_in
+      - ref: gen_ai.tool.definitions
+        requirement_level: opt_in      
+      - ref: gen_ai.orchestrator.agent_definitions
```
lmolkova, line 278: What are the other properties orchestrator we envision going forward? 

What if the orchestrator does not know it's role and would delegate via mcp tools and not to the agents directly, would it be a problem?

model/gen-ai/spans.yaml
ShipraJain01: @alexmojaki, 

In theory tools are execution engine available to agents who in turn pass these tool definitions to LLM to utilize appropriate tools. Hence it made more sense to capture this at Agent Span level, instead of LLM span or tool span. 

Another very important reason for which we captured tool definitions at Agent span is to avoid duplication, coz there can be multiple LLM calls that may happen during an agent's span. Passing it all at Agent's span keeps it central, de-duped.

model/gen-ai/registry.yaml
ShipraJain01: @alexmojaki : I agree. In this PR, I intend to capture tool definitions at agent span. Let me remove model, since I believe having this information at model could lead to duplication of passing same definitions many times.

model/gen-ai/registry.yaml
ShipraJain01: @lmolkova: I provided the examples based on previous comments from reviewers. If we need more discussion around built-in tools, please allow me to keep the example only for 'function_call'

model/gen-ai/registry.yaml
gyliu513: OK, then should this attribute be in `create agent` but not `invoke agent` span?

model/gen-ai/registry.yaml
ShipraJain01: Nopes, I added these based on previous comments from reviewers. Since we are taking built-in tools separately, please allow me to remove these examples and retain only function_call

model/gen-ai/spans.yaml
alexmojaki: 1. Again, the tool definitions sent with each LLM request can change over the course of the agent run, in which case recording on the LLM span makes a lot more sense than on the agent span.
2. LLM spans can happen outside an agent run, wouldn't we record tool definitions then?

Anyway, this doesn't need to be a blocker, just something to keep in mind.

model/gen-ai/registry.yaml
alexmojaki: @ShipraJain01 that doesn't relate to what I said in this comment at all. This isn't about model vs agent, it's about the real raw value ('provider-specific tool definition') vs a normalized/canonical/consistent representation for telemetry purposes.

model/gen-ai/registry.yaml
ShipraJain01: no, removing in next PR update

model/gen-ai/registry.yaml
ShipraJain01: Sure @lmolkova . converting it to yaml comments.

model/gen-ai/spans.yaml
ShipraJain01: @lmolkova : for now delegation via mcp tools is not mainstream. As per definition of Orchestrator in multi-agent framework, its the central decision making agent which controls delegation. I cant forsee if orchestrator wont know its role, even if orchestrator chooses to delegate via mcp tools.

What do you think?

model/gen-ai/registry.yaml
ShipraJain01: Doing it in next PR update.

model/gen-ai/spans.yaml
ShipraJain01: let me convert the requirement level to opt_in to allow users the flexibility

model/gen-ai/spans.yaml
ShipraJain01: @lmolkova : fair point. please allow me to change the requirement level to opt-in

model/gen-ai/spans.yaml
lmolkova: SIG discussion:
- let's add it as LLM span attribute as well as agent one
- having it in both places involves some duplication, but we don't know yet how/if it's bad

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,155 @@ groups:
...
+        type: any
+        brief: Definitions of agents available to the orchestrator for multi-agent workflows.
+        note: |
+          This attribute contains an array of objects representing agent definitions that are available to the orchestrator for potential use in multi-agent workflows.
+          
+          The structure is an array where each object contains:
+          
+          - 'agent_id': Unique identifier for the agent
+          - 'name': Human-readable name of the agent
+          - 'role': Classification of the agent (e.g., "specialist", "coordinator", "validator")
```
lmolkova, line 384: linking to related discussion - https://github.com/open-telemetry/semantic-conventions/pull/2685

model/gen-ai/registry.yaml
ShipraJain01: @gyliu513 : as per the decision in today's SIG call, we will capture this attribute more in details in separate PR.

model/gen-ai/spans.yaml
ShipraJain01: Adding tool definitions in LLM Spans as well

model/gen-ai/registry.yaml
ShipraJain01: Yes, got more clarity during our call. As per the decision in the call, we are adding this attribute to LLM and agent spans

alexmojaki: again, not what this is about. we didn't discuss this comment in the call.

model/gen-ai/registry.yaml
ShipraJain01: As discussed in SIG call, we are going to capture this attribute in separate PR only capturing behaviour of orchestrator agents

model/gen-ai/registry.yaml
ShipraJain01: As discussed in SIG call, moving this attribute to separate PR

model/gen-ai/spans.yaml
ShipraJain01: As discussed in SIG call, moving this attribute to separate PR

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -199,11 +199,12 @@ Semantic conventions for individual GenAI systems and frameworks MAY specify dif
...
 | [`gen_ai.data_source.id`](/docs/registry/attributes/gen-ai.md) | string | The data source identifier. [5] | `H7STPQYOND` | `Conditionally Required` if applicable. | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`gen_ai.output.type`](/docs/registry/attributes/gen-ai.md) | string | Represents the content type requested by the client. [6] | `text`; `json`; `image` | `Conditionally Required` [7] | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.orchestrator.agent_definitions`](/docs/registry/attributes/gen-ai.md) | any | Definitions of agents available to the orchestrator for multi-agent workflows. [6] | [<br>&nbsp;&nbsp;{<br>&nbsp;&nbsp;&nbsp;&nbsp;"agent_id": "agent_weather_001",<br>&nbsp;&nbsp;&nbsp;&nbsp;"name": "WeatherAgent",<br>&nbsp;&nbsp;&nbsp;&nbsp;"role": "specialist"<br>&nbsp;&nbsp;},<br>&nbsp;&nbsp;{<br>&nbsp;&nbsp;&nbsp;&nbsp;"agent_id": "agent_calendar_002",<br>&nbsp;&nbsp;&nbsp;&nbsp;"name": "CalendarAgent",<br>&nbsp;&nbsp;&nbsp;&nbsp;"role": "coordinator"<br>&nbsp;&nbsp;}<br>] | `Conditionally Required` if applicable | ![Development](https://img.shields.io/badge/-development-blue) |
```
TaoChenOSU, line 201: nit: What is the definition of `role` for a sub agent?

docs/gen-ai/gen-ai-spans.md
```diff
@@ -370,18 +370,26 @@ by the application code.
...
+| [`gen_ai.tool.call.arguments`](/docs/registry/attributes/gen-ai.md) | any | Parameters passed to the tool call. [3] | `For function_call: {"location": "San Francisco?", "date": "2025-10-01"}`; `For file_search: {"queries": ["semantic conventions for HTTP attributes", "Find all files containing OpenTelemetry database semantic conventions"]}`; For code_interpreter with code:<br>{<br>&nbsp;&nbsp;"type": "code",<br>&nbsp;&nbsp;"content": "import pandas as pd\ndf = pd.read_csv('/tmp/data.csv')\nprint(df.describe())",<br>&nbsp;&nbsp;"file_ids": ["file-abc123", "file-xyz789"]<br>}; For code_interpreter with file instruction:<br>{<br>&nbsp;&nbsp;"type": "file",<br>&nbsp;&nbsp;"content": "Analyze this CSV file and create a summary report with visualizations",<br>&nbsp;&nbsp;"file_ids": ["file-data001"]<br>} | `Conditionally Required` if available | ![Development](https://img.shields.io/badge/-development-blue) |
```
TaoChenOSU, line 367: nit: file search and code interpreter can be service side tools, in which case client SDKs will have to very deliberately retrieve the inputs to those tools for instrumentation purposes. Maybe we can point it out here?

TaoChenOSU (approved):

model/gen-ai/registry.yaml
ShipraJain01: Re-reading your comment @alexmojaki, my bad. 

As per SIG discussion we are adding 'gen_ai.tool.definitions' in both LLM and Agent span. 

Your point here is to name it differently. In this attribute we want to capture 'provider specific tool definitions', I can think of below possible naming options. Please share your thoughts - 

gen_ai.raw_tool.definitions
gen_ai.original_tool.definitions
gen_ai.source_tool.definitions
gen_ai.provider_tool.definitions

docs/gen-ai/gen-ai-agent-spans.md
ShipraJain01: @TaoChenOSU : this topic requires more discussion. As per decision in today's SIG call, I am taking it in separate PR.

docs/gen-ai/gen-ai-agent-spans.md
ShipraJain01: Built-in tools will be discussed in separate PR. removing code-interpreter example from this PR.

docs/gen-ai/gen-ai-spans.md
ShipraJain01: @Cirilla-zmh: please suggest if above explanation sounds reasonable.

docs/gen-ai/gen-ai-agent-spans.md
ShipraJain01: this attribute is removed from the PR as we are adoption 'gen_ai.input.messages' and 'gen_ai.output.messages' which got merged from following PR - 

https://github.com/open-telemetry/semantic-conventions/pull/2179

model/gen-ai/registry.yaml
alexmojaki: @lmolkova are there any general guidelines around recording real and/vs normalized values?

model/gen-ai/registry.yaml
ShipraJain01: @alexmojaki: As per decision in SIG call today, removing this attribute from current PR. Would create separate PR for this one.

model/gen-ai/registry.yaml
ShipraJain01: @alexmojaki , @lmolkova: I was wondering what if we leave this attribute with same naming convention - gen_ai.tool.definitions (in description I am highlighting that its provider-specific tool definition to avoid any confusion for user). 

In future if we want to introduce new attributes, we choose new name accordingly like - 

gen_ai.tool_definitions.normalized
gen_ai.tool_definitions.canonicalized

etc..

thoughts?

model/gen-ai/spans.yaml
ShipraJain01: As per SIG decision, we are removing this attribute in this PR. To be taken in another one.

docs/gen-ai/gen-ai-spans.md
ShipraJain01: @TaoChenOSU : agree. For now we are separating the scope of service side / built-in tools from this PR which requires broader discussion. 
https://github.com/open-telemetry/semantic-conventions/issues/2585

model/gen-ai/registry.yaml
ShipraJain01: @alexmojaki: could you please double click on the similarity between the 2 attributes - 'gen_ai.system_instructions' vs 'gen_ai.tool_definitions' since these attributes differ in purpose. 

eg: 
System Instructions = behavioral context for the model. Tells the model “how” to act.
Tool Definitions = capability context for the model / Agent. Tells the model/Agent “what” it can do.

model/gen-ai/registry.yaml
alexmojaki: i would probably go with `gen_ai.tool_definitions.raw` for this and `gen_ai.tool_definitions.normalized` in the future. the name `gen_ai.tool_definitions` on its own implies that it's not a namespace and there shouldn't be attribute names starting with it. and i would think that 'normalized' is the default expectation, not 'raw'.

model/gen-ai/registry.yaml
ShipraJain01: @alexmojaki : Since the content in this attribute can grow in size, in PR description I suggested to follow similar approach to use 'gen_ai.tool.definitions_ref' as described in PR - https://github.com/open-telemetry/semantic-conventions/pull/2179

Also we are still to define common approach in OTEL to record references to externally stored content. Once that is figured, it would have to be applied to gen_ai.system.instructions , gen_ai.input.messages, gen_ai.output.messages and gen_ai.tool.definitions

Until then I can refer to [Recording content on attributes](/docs/gen-ai/gen-ai-spans.md#recording-content-on-attributes) section in description of this attribute. 

What are your thoughts?

model/gen-ai/registry.yaml
ShipraJain01: @lmolkova: could you share your feedback on using `gen_ai.tool_definitions.raw` as the changed name for `gen_ai.tool_definitions`? for now I have changed the name as per Alex's suggestion.

model/gen-ai/registry.yaml
alexmojaki: ok, no need to block anything here.

docs/gen-ai/gen-ai-spans.md
Cirilla-zmh: > We can manage it even in current form right. Only maintain 'type' and 'name' key value pair to avoid adding sensitive 'parameters' definition.

Sure! Great proposal. A configuration item should be defined to control the behavior.

ShipraJain01: @Cirilla-zmh ,
Could you elaborate what you have in mind for `configuration item should be defined to control the behavior.` ?

ShipraJain01:
> [CLA Missing ID](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/638591290/2528/#/?version=2) [CLA Not Signed](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/638591290/2528/#/?version=2)
> 
> * ✅ login: ShipraJain01 / name: ShipraJain01  ([6d9a13e](https://github.com/open-telemetry/semantic-conventions/commit/6d9a13ec0dadeb071c89ec0fa68abe4753ae81d9), [8e8ede2](https://github.com/open-telemetry/semantic-conventions/commit/8e8ede24f11e4484483c82ffd3beb660205ca898), [6e5b1d8](https://github.com/open-telemetry/semantic-conventions/commit/6e5b1d84ddf52f16b372140db4f5eefd345d06dc))
> * ✅ login: ShipraJain01 / name: shiprajain01  ([7ac0088](https://github.com/open-telemetry/semantic-conventions/commit/7ac00887d4b3e56ffa052247f426aeae1c4f92fc), [e98d4db](https://github.com/open-telemetry/semantic-conventions/commit/e98d4dbc163582cabda3577b9b42f90dc3bd26b2), [a5ccf25](https://github.com/open-telemetry/semantic-conventions/commit/a5ccf257e052fbf2de827c208a1b914d3dc799b9), [f9af46b](https://github.com/open-telemetry/semantic-conventions/commit/f9af46b5ee3952fbe65f7421598764ca34b332fe), [00b9369](https://github.com/open-telemetry/semantic-conventions/commit/00b9369c17c12c9ffa76747fcc317751f68e0e78), [f356a4a](https://github.com/open-telemetry/semantic-conventions/commit/f356a4ab189937265f00249f0eea7dbddbb6ed32))
> * ❌ The email address for the commit ([eb7e0f4](https://github.com/open-telemetry/semantic-conventions/commit/eb7e0f4e4b37c144c7374a950eb112724688c5dd), [507acc5](https://github.com/open-telemetry/semantic-conventions/commit/507acc5696977e85ba009f20e27f363a51ba3937), [6d6896e](https://github.com/open-telemetry/semantic-conventions/commit/6d6896e5912339e758ef066c7a22cd1f81084931), [bdb4106](https://github.com/open-telemetry/semantic-conventions/commit/bdb410666cf1f4bd169cd0a6bd23c528fe4c39de), [4c93be1](https://github.com/open-telemetry/semantic-conventions/commit/4c93be167708a32a5a90e5368608869b3cd3c6cb), [f89a054](https://github.com/open-telemetry/semantic-conventions/commit/f89a0546789c7df6cbb205af5d11fd61aefe3dc2), [eedfecc](https://github.com/open-telemetry/semantic-conventions/commit/eedfecc424b9c836f8dece8cf2a81a2c2cf38c88)) is not linked to the GitHub account, preventing the EasyCLA check. Consult [this Help Article](https://linuxfoundation.atlassian.net/wiki/spaces/LP/pages/160923756/Missing+ID+on+Commit+but+I+have+an+agreement+on+file) and [GitHub Help](https://help.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user) to resolve. (To view the commit's email address, add .patch at the end of this PR page's URL.) For further assistance with EasyCLA, [please submit a support request ticket](https://jira.linuxfoundation.org/servicedesk/customer/portal/4).

ShipraJain01:
Unintentionally closed the PR. Reopening.

model/gen-ai/registry.yaml
lmolkova: I don't envision us ever recording both - raw and normalized on the same span/event. It's super-wasteful and doesn't seem to make much sense. 

Assuming we use raw definitions today and in the future decide to make type more restrictive, we'd rather do it in-place on the same attribute. 

Given we're not stabilizing yet, I'd rather keep `gen_ai.tool_definitions` as is - figuring out how strict we want to be would be a blocker to stabilization. Without some research on different formats and use-cases I would be hesistant to stablize any form of `gen_ai.tool_definitions*`, so I don't see what we're gaining by adding `.raw` or similar suffix.

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,93 @@ groups:
...
+            }
+        
+      - id: gen_ai.tool_definitions.raw
+        stability: development
+        type: any
+        brief: The list of source system tool definitions available to the GenAI agent or model.
+        note: >
+          The value of this attribute matches source system tool definition format.
+         
+          It's expected to be an array of JSON objects - in case a serialized string is available
```
```suggestion
          It's expected to be an array of objects where each object represents a tool definition. In case a serialized string is available
```

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,93 @@ groups:
             Client-side operations are actions taken on the user's end or within the client application.
           Datastore: A tool used by the agent to access and query structured or unstructured external data for retrieval-augmented tasks or knowledge updates.
         examples: ['function', 'extension', 'datastore']
+
+      - id: gen_ai.tool.call.arguments
+        stability: development
+        type: any
+        brief: Parameters passed to the tool call.
+        note: >
```
lmolkova, line 263:
```suggestion
        note: |
```

(otherwise formatting is broken)

model/gen-ai/spans.yaml
```diff
@@ -119,6 +119,8 @@ groups:
         requirement_level: required
       - ref: gen_ai.request.top_k
         requirement_level: recommended
+      - ref: gen_ai.tool_definitions.raw
```
lmolkova: I think it should go to the `attributes.gen_ai.inference.client` group and it would populate it on all inference spans, agent spans, and also details event (which mirrors inference span).

lmolkova (commented):
LGTM, the only concern is discussion around`gen_ai.tool_definitions.raw`

docs/gen-ai/gen-ai-spans.md
Cirilla-zmh: @ShipraJain01 
Yes. The naming of configuration shouldn't be defined in semantic conventions. But we may document like: 
https://github.com/open-telemetry/semantic-conventions/blob/570283b62fb14364606c8150f0c9843183555082/docs/gen-ai/gen-ai-spans.md?plain=1#L429-L430

In python instrumentation the option may be named `OTEL_INSTRUMENTATION_GENAI_CAPTURE_DETAILED_TOOL_DEFINITION`.

model/gen-ai/registry.yaml
alexmojaki: ok, i'm fine with dropping the suffix

ShipraJain01: Great! I will retain `gen_ai.tool_definitions`

model/gen-ai/registry.yaml
ShipraJain01: Done

docs/gen-ai/gen-ai-spans.md
ShipraJain01: @Cirilla-zmh : In yesterday's SIG call and from offline feedback, we have concluded to use `gen_ai.tool_definitions` attribute with opt-in requirement level with following documentation to the user which covers your suggestion - `The value of this attribute matches source system tool definition format. It's expected to be an array of objects where each object represents a tool definition. In case a serialized string is available to the instrumentation, the instrumentation SHOULD do the best effort to deserialize it to an array. Since this attribute could be large, it's NOT RECOMMENDED to populate it by default. Instrumentations MAY provide a way to enable populating this attribute.`

model/gen-ai/spans.yaml
ShipraJain01: added the `gen_ai.tool_definitions` attribute to `attributes.gen_ai.inference.client`. Pushing the change in next PR. Thanks @lmolkova.

.chloggen/2528.yaml
```diff
@@ -0,0 +1,8 @@
+change_type: enhancement
+component: gen-ai
+note: | 
+  Add new attributes to 'invoke_agent', 'execute_tool' and llm inference spans to capture input and output definition for Single and Multi-Agent systems.
+  - Add 'gen_ai.tool_definitions.raw' to 'gen_ai.invoke_agent' span and 'gen_ai.inference' span.
```
lmolkova:
```suggestion
  - Add 'gen_ai.tool_definitions' to 'gen_ai.invoke_agent' span and 'gen_ai.inference' span. 
```

.chloggen/2528.yaml
```diff
@@ -0,0 +1,8 @@
+change_type: enhancement
+component: gen-ai
+note: | 
+  Add new attributes to 'invoke_agent', 'execute_tool' and llm inference spans to capture input and output definition for Single and Multi-Agent systems.
```
lmolkova:
```suggestion
  Add tool definitions and execution details:
```

model/gen-ai/registry.yaml
lmolkova: maybe a commit was lost, I don't see the change

model/gen-ai/registry.yaml
```diff
@@ -253,6 +255,93 @@ groups:
...
+        - |
+            {
+              "temperature_range": {
+                "high": 75,
+                "low": 60
+              },
+              "conditions": "sunny"
+            }
+        
+      - id: gen_ai.tool_definitions.raw
```
```suggestion
      - id: gen_ai.tool_definitions
```

it seems some changes got lost

model/gen-ai/registry.yaml
ShipraJain01: @lmolkova : repushed the changes again, its visible in latest commit. kindly review.

.chloggen/2528.yaml
ShipraJain01: @lmolkova : repushed the changes again, its visible in latest commit. kindly review.

.chloggen/2528.yaml
ShipraJain01: @lmolkova : made the changes in latest push. Please review

model/gen-ai/registry.yaml
ShipraJain01: @lmolkova : repushed the changes again, its visible in latest commit. kindly review.

alexmojaki (approved):

lmolkova:
I believe this one should be closed in favor of #2702

Issue #2530, opened 2025-07-17
Add ASP.NET Core authentication and authorization metrics


JamesNK:
ASP.NET Core is adding new metrics in the next version for authentication and authorization. These should be documented in the semantic conventions.

PR #2531, opened 2025-07-17
Add ASP.NET Core AuthN and AuthZ metrics

JamesNK:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/2530

## Changes

* [x] Links to the prototypes or existing instrumentations (when adding or changing conventions) - https://github.com/dotnet/aspnetcore/blob/main/src/Http/Authentication.Core/src/AuthenticationMetrics.cs


model/aspnetcore/registry.yaml
```diff
@@ -83,3 +83,50 @@ groups:
...
+              stability: development
+        stability: development
+        brief: The result of the authentication operation.
+        examples: ["success", "failure"]
+      - id: aspnetcore.authentication.scheme
+        type: string
+        stability: development
+        brief: The identifier that names a particular authentication handler.
+        examples: ["Cookies", "Bearer", "Identity.Application"]
+      - id: user.is_authenticated
```
lmolkova: this friend belongs in the https://github.com/open-telemetry/semantic-conventions/blob/main/model/user/registry.yaml.

The problem is that `user` namespace here is way too broad - https://github.com/open-telemetry/semantic-conventions/issues/1172. We're still trying to figure out what should be there and it won't be stable in foreseeable future. 

If the goal is to stabilize metrics defined here in Nov, I'd recommend creating `aspnetcote.identity.is_authenticated` attribute (or `dotnet.identity.is_authenticated` given that `ClaimsPrincipal` and `Identity` are in `System.Security` namespace). This can be stabilized.

It'd likely collide with identity conventions we'll eventually have to define here, but it's might be years away.

model/aspnetcore/registry.yaml
```diff
@@ -83,3 +83,50 @@ groups:
...
+        type: boolean
+        stability: stable
+        brief: A value that indicates whether the user is authenticated.
+        examples: [true]
+      - id: aspnetcore.authorization.policy
```
lmolkova, line 124:
```suggestion
      - id: aspnetcore.authorization.policy.name
```

(to disambiguate between name, version, type and other possible properties)

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +148,104 @@ groups:
...
+      - ref: user.is_authenticated
+        requirement_level: required
+      - ref: aspnetcore.authorization.policy
+        requirement_level:
+          conditionally_required: if a policy is specified.
+      - ref: aspnetcore.authorization.result
+        requirement_level:
+          conditionally_required: if no exception was thrown.
+      - ref: error.type
+        brief: The full name of exception type.
```
lmolkova, line 320: Is it possible that no exception is thrown but aspnetcore.authorization.result == failure? whould error.type be set? let's document it

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +148,104 @@ groups:
...
+  - id: metric.aspnetcore.authorization.attempts
+    type: metric
+    metric_name: aspnetcore.authorization.attempts
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: development
+    brief: The total number of requests for which authorization was attempted.
+    instrument: counter
+    unit: "{request}"
```
lmolkova: nit, maybe

```suggestion
    unit: "{attempt}"
```

?

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +148,104 @@ groups:
...
+  - id: metric.aspnetcore.authentication.sign_outs
+    type: metric
+    metric_name: aspnetcore.authentication.sign_outs
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: development
+    brief: The total number of times a principal is signed out with a scheme.
+    instrument: counter
+    unit: "{request}"
```
lmolkova: nit, maybe

```suggestion
    unit: "{sign_out}"
```

?

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +148,104 @@ groups:
...
+  - id: metric.aspnetcore.authentication.challenges
+    type: metric
+    metric_name: aspnetcore.authentication.challenges
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: development
+    brief: The total number of times a scheme is challenged.
+    instrument: counter
+    unit: "{request}"
```
```suggestion
    unit: "{challenge}"
```

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +148,104 @@ groups:
...
+    instrument: histogram
+    unit: "s"
+    note: |
+      Meter name: `Microsoft.AspNetCore.Authentication`; Added in: ASP.NET Core 10.0
+    extends: aspnetcore.common.authentication.metrics.attributes
+    attributes:
+      - ref: aspnetcore.authentication.result
+        requirement_level: required
+
+  - id: metric.aspnetcore.authentication.challenges
```
lmolkova, line 241: wrt pluralization question here https://github.com/open-telemetry/semantic-conventions/pull/2508#discussion_r2214490007

a consistent approach could  be to follow `aspnetcore.authentication.*.count` format

docs/dotnet/dotnet-aspnetcore-metrics.md
```diff
@@ -21,6 +21,14 @@ This article defines semantic conventions for ASP.NET Core metrics.
   - [Metric: `aspnetcore.rate_limiting.queued_requests`](#metric-aspnetcorerate_limitingqueued_requests)
   - [Metric: `aspnetcore.rate_limiting.request.time_in_queue`](#metric-aspnetcorerate_limitingrequesttime_in_queue)
   - [Metric: `aspnetcore.rate_limiting.requests`](#metric-aspnetcorerate_limitingrequests)
+- [Authentication](#authentication)
+  - [Metric: `aspnetcore.authentication.authenticate.duration`](#metric-aspnetcoreauthenticationauthenticateduration)
+  - [Metric: `aspnetcore.authentication.challenges`](#metric-aspnetcoreauthenticationchallenges)
+  - [Metric: `aspnetcore.authentication.forbids`](#metric-aspnetcoreauthenticationforbids)
+  - [Metric: `aspnetcore.authentication.sign_ins`](#metric-aspnetcoreauthenticationsign_ins)
+  - [Metric: `aspnetcore.authentication.sign_outs`](#metric-aspnetcoreauthenticationsign_outs)
```
lmolkova, line 34: a typical OTel way would be to define one duration metric like `aspnetcore.authentication.operation.duration` and add an `aspnetcore.authenthication.operation.name = authenticate | challenge | sign_in` to distunguish. 

This works reasonably well here since these metrics share pretty much the same set of attributes. `aspnetcore.authentication.result` could  be added with `recommended: when applicable` or similar level.

This simplifies end user experience - user can render one metric and then discover all the operations they use instead of looking into the docs on how each operation is called.

There could be pros and cons to this approach, but worth considering.

model/aspnetcore/registry.yaml
```diff
@@ -83,3 +83,50 @@ groups:
...
+              value: 'success'
+              brief: "Authorization was successful."
+              stability: development
+            - id: failure
+              value: 'failure'
+              brief: "Authorization failed."
+              stability: development
+        stability: development
+        brief: The result of calling the authorization service.
+        examples: ["success", "failure"]
```
thompson-tomo, line 142: Thoughts about putting some/all of these in an access-control.authorization/iam.authorization namespace?

model/aspnetcore/registry.yaml
```diff
@@ -83,3 +83,50 @@ groups:
...
+      - id: aspnetcore.authentication.scheme
+        type: string
+        stability: development
+        brief: The identifier that names a particular authentication handler.
+        examples: ["Cookies", "Bearer", "Identity.Application"]
+      - id: user.is_authenticated
+        type: boolean
+        stability: stable
+        brief: A value that indicates whether the user is authenticated.
+        examples: [true]
```
thompson-tomo: Thoughts about putting some/all of these in an access-control.authentication/iam.authentication namespace?

model/aspnetcore/registry.yaml
JamesNK: Changed to `aspnetcore.user.is_authenticated`

model/aspnetcore/registry.yaml
JamesNK: Not needed because an auth policy is only identified by the name.

model/aspnetcore/registry.yaml
JamesNK: I'd prefer to keep them specific to aspnetcore and not worry about defining cross framework standards.

model/aspnetcore/metrics.yaml
JamesNK: No it wouldn't be set.

JamesNK:

JamesNK:
@trask @lmolkova I resolved conflicts created by merging other PRs.

This PR is ready to merge.

(last PR for now)

lmolkova (approved):

Issue #2532, opened 2025-07-17
Add ASP.NET Core memory pool metrics


JamesNK:
ASP.NET Core is adding new metrics in the next version for memory pool. These should be documented in the semantic conventions.

PR #2533, opened 2025-07-17
Add ASP.NET Core memory pool metrics

JamesNK:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/2532

## Changes

* [x] Links to the prototypes or existing instrumentations (when adding or changing conventions) - https://github.com/dotnet/aspnetcore/blob/main/src/Shared/Buffers.MemoryPool/PinnedBlockMemoryPoolMetrics.cs


model/aspnetcore/registry.yaml
```diff
@@ -83,3 +83,8 @@ groups:
         stability: stable
         brief: ASP.NET Core exception middleware handling result
         examples: ["handled", "unhandled"]
+      - id: aspnetcore.memory_pool.owner
+        type: string
+        stability: development
+        brief: The name of the part of the app using the pool instance.
```
lmolkova: is there some public API or something documented in .NET that it maps to? Looking into [current implementation](https://github.com/dotnet/aspnetcore/blob/1f0c98bb585e0add5e45cb2d42ae40a76ec38fb1/src/Shared/Buffers.MemoryPool/PinnedBlockMemoryPool.cs#L102), it doesn't seem it's reported yet

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +142,60 @@ groups:
     attributes:
       - ref: aspnetcore.rate_limiting.result
         requirement_level: required
+
+    # memory_pool
+  - id: metric.aspnetcore.memory_pool.current_memory
+    type: metric
+    metric_name: aspnetcore.memory_pool.current_memory
```
lmolkova: I believe we use `used` or `usage` in similar metrics. 

```suggestion
    metric_name: aspnetcore.memory_pool.used # or current
```

`Usage` usually has state-like attribute with values (free | cached | etc) and sums up to `limit`, so here `used` is more suitable.

E.g. [`jvm.memory.used`, `jvm.memory.committed`,  `metric.jvm.memory.limit`](https://github.com/open-telemetry/semantic-conventions/blob/8b1d5d4bfe50fd4630e24141c89b0bf09ebf076f/model/jvm/metrics.yaml#L12). 

Somewhat relevant guidance - https://github.com/open-telemetry/semantic-conventions/blob/1000be4d897fafbc8554b373267992e47c1b86ac/docs/general/naming.md#instrument-naming.

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +142,60 @@ groups:
...
+        metric_value_type: int
+    stability: development
+    brief: Number of bytes that are currently pooled by the pool.
+    instrument: updowncounter
+    unit: "By"
+    note: |
+      Meter name: `Microsoft.AspNetCore.MemoryPool`; Added in: ASP.NET Core 10.0
+    extends: aspnetcore.common.memory_pool.metrics.attributes
+
+  - id: metric.aspnetcore.memory_pool.total_allocated
```
lmolkova: Following previous suggestion on current

```suggestion
  - id: metric.aspnetcore.memory_pool.allocated # or limit
```
it does not seem like `total` is meaningful here anyway

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +142,60 @@ groups:
...
+        metric_value_type: int
+    stability: development
+    brief: Total number of allocations made by the pool.
+    instrument: counter
+    unit: "By"
+    note: |
+      Meter name: `Microsoft.AspNetCore.MemoryPool`; Added in: ASP.NET Core 10.0
+    extends: aspnetcore.common.memory_pool.metrics.attributes
+
+  - id: metric.aspnetcore.memory_pool.evicted_memory
```
lmolkova: Consistent with suggestions in previous comments

```suggestion
  - id: metric.aspnetcore.memory_pool.evicted
```

it does not seem like `_memory` is essential

model/aspnetcore/metrics.yaml
```diff
@@ -135,3 +142,60 @@ groups:
...
+    brief: Total number of bytes that have been evicted.
+    instrument: counter
+    unit: "By"
+    note: |
+      Meter name: `Microsoft.AspNetCore.MemoryPool`; Added in: ASP.NET Core 10.0
+    extends: aspnetcore.common.memory_pool.metrics.attributes
+
+  - id: metric.aspnetcore.memory_pool.total_rented
+    type: metric
+    metric_name: aspnetcore.memory_pool.total_rented
```
```suggestion
    metric_name: aspnetcore.memory_pool.rented
```

model/aspnetcore/metrics.yaml
JamesNK: I used `pooled` instead of used. Used doesn't work in the context of a memory pool because it could be confused with current pooled and rented memory.

model/aspnetcore/registry.yaml
JamesNK: Adding it here: https://github.com/dotnet/aspnetcore/pull/63032

I don't think there will be docs that discuss this list. People using the metrics can see the owners when they view their app's telemetry.

JamesNK:

noahfalk (approved):

KalleOlaviNiemitalo:
Is there an invariant:

aspnetcore.memory_pool.pooled = aspnetcore.memory_pool.allocated - aspnetcore.memory_pool.rented - aspnetcore.memory_pool.evicted

JamesNK:
No, because rented is total. Memory could be returned and rented many times.

Allocated: 50 bytes
Rented: 500 bytes
Evicted: 0 bytes

Pooled does not equal -450 bytes.

One interesting piece of data is how much an application has currently leased. You can calculate the currently leased memory with this:

```
leased = allocated - pooled - evicted
```

JamesNK:
@trask I resolved conflicts created by merging https://github.com/open-telemetry/semantic-conventions/pull/2508.

This PR is ready to merge.

lmolkova (approved):

julealgon:
Hey @JamesNK , is there any particular reason why these metrics are not documented here?
- https://learn.microsoft.com/en-us/aspnet/core/log-mon/metrics/built-in?view=aspnetcore-10.0

JamesNK:
Never got around to it. AI could probably copy it across quickly

julealgon:
> Never got around to it. AI could probably copy it across quickly

Thanks for clarifying. I was wondering if there was a chance this was actually not pushed as part of .NET 10 and that was maybe why. Good to hear that's not the case.

PR #2534, opened 2025-07-17
Used memory should be based on operating system's internals

rogercoll:
## Changes

This PR clarifies the expected value of the `used` memory attribute on Linux systems, specifying that it should be derived from `MemAvailable` for a more accurate representation of memory usage.

On Linux, using MemAvailable breaks the assumption that the sum of all `system.memory.state` values equals `system.memory.limit`. However, this assumption is only valid because the current used value is **derived** indirectly from the limit and other memory states (total - free - buffers - cached), which can misrepresent actual memory pressure (with some [issues](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7417)). If the current used state is derivable, I recommend redefining it to reflect a more meaningful value that cannot be derived—specifically one based on MemAvailable—and making system.memory.limit enabled by default.

While the proposed used value can still be calculated from the optional [system.memory.limit](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemmemorylimit) and [system.linux.memory.available](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemlinuxmemoryavailable) metrics, doing so requires awareness of the current used state’s inaccuracy and explicit user configuration. This change aims to promote a more accurate and meaningful metric for memory usage in Linux environments.

Context: https://github.com/shirou/gopsutil/issues/1873
Repository to test the different formulas: https://github.com/rogercoll/linux-memory-metrics

### DIFF
--- a/model/system/registry.yaml
+++ b/model/system/registry.yaml
@@ -22,8 +22,9 @@ groups:
             - id: used
               value: 'used'
               stability: development
-              brief: >
-                Actual used virtual memory in bytes. This is calculated based on the operating system.
+              brief: Actual used virtual memory in bytes.
+              note: >
+                Calculation based on the operating system metrics.
                 On Linux, this corresponds to "MemTotal - MemAvailable" from /proc/meminfo, which
                 more accurately reflects memory in active use by applications compared to older
                 formulas based on free, cached, and buffers. If MemAvailable is not available,
--- a/docs/registry/attributes/system.md
+++ b/docs/registry/attributes/system.md
@@ -69,7 +69,9 @@ Describes System Memory attributes
 | `buffers` | buffers | ![Development](https://img.shields.io/badge/-development-blue) |
 | `cached` | cached | ![Development](https://img.shields.io/badge/-development-blue) |
 | `free` | free | ![Development](https://img.shields.io/badge/-development-blue) |
-| `used` | Actual used virtual memory in bytes. This is calculated based on the operating system. On Linux, this corresponds to "MemTotal - MemAvailable" from /proc/meminfo, which more accurately reflects memory in active use by applications compared to older formulas based on free, cached, and buffers. If MemAvailable is not available, a fallback to those older formulas may be used. | ![Development](https://img.shields.io/badge/-development-blue) |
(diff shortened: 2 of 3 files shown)


braydonk (approved):

model/system/registry.yaml
```diff
@@ -22,6 +22,11 @@ groups:
             - id: used
               value: 'used'
               stability: development
+              brief: >
+                Actual used virtual memory in bytes. This is calculated based on the operating system.
+                On Linux, this corresponds to "MemTotal - MemAvailable" from /proc/meminfo, which
+                more accurately reflects memory in active use by applications. If MemAvailable is not
```
mx-psi: The "more" here implies a comparison with something else, but that something else is not explicitly mentioned. Should it?

rogercoll: Good point, tried to clarify to what's being compared with these words: https://github.com/open-telemetry/semantic-conventions/pull/2534/commits/9b2636a61ba484bcbfb938b09fd2826f71624793

wdyt?

rogercoll:
Opened this issue to work on the fix for breaking the aggregation over the states: https://github.com/open-telemetry/semantic-conventions/issues/2538

cc @open-telemetry/semconv-system-approvers @dmitryax

model/system/registry.yaml
mx-psi: makes sense, thx!

model/system/registry.yaml
```diff
@@ -22,6 +22,12 @@ groups:
             - id: used
               value: 'used'
               stability: development
+              brief: >
```
lmolkova: nit: could you please keep brief concise and move additional info to the note?

lmolkova (approved):

model/system/registry.yaml
rogercoll: Fixed: https://github.com/open-telemetry/semantic-conventions/pull/2534/commits/ae308aef63c6f84f42a346e2280c5874594cf5c4

PR #2544, opened 2025-07-19
Add in the collector defined NTP metrics

### LINKED ISSUE #2543: Document NTP Collector Metrics
### Area(s)

area:new

### Propose new conventions

I want the OTEL collectors related to NTP to be defined in semantic conventions with changes being designed either renames or switching of data types.

thompson-tomo:
Fixes #2543

## Changes

This brings across the NTP metrics/attributes definitions from the otel collector ntp recievers.

## Deviation from current implementation:

- Rename of `leap.status` to `ntp.leapsec.mode`
- Rename of `ntp.host` to `ntp.servers` and changing type to array of strings


jsuereth:
@atoulme Could you do a review of this as owner of NTP receiver in collector?  Is this component in state of development and adoption to formalize semantic conventions?

atoulme:
I propose we do this first in the collector reusing the work I did here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40698

Once we are sure we have compatibility, we can bring over the yaml file to semantic-conventions. It might be possible to make semantic-conventions import the file from the collector repository, or transfer it here in semantic-conventions and import it in the collector.

thompson-tomo:
I am not sure about that approach as it doesn't seem to encourage reusage/interoperability/discoverability and results in duplication.

What my thinking is to explore my proposal in my comment (https://github.com/open-telemetry/weaver/issues/406) of creating a package definitions which contains a range of signals. You can use that package definition to generate the metadata.yaml file. Key thing is that semantic conventions should be the driver of the definition of what can/should be implemented.

lmolkova:
Do we actually control which metrics are collected by the NTP receiver in collector? I.e. do we control the attributes, can we define new metrics or are they coming from the external source and the best we can do is to apply our naming guidance to what's already there? 

It it's the latter, these metrics don't belong in semantic conventions and their definitions should remain in the collector.

https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#which-semantic-conventions-belong-in-this-repo

thompson-tomo:
Yes we are in control of the definition of the metrics.

PR #2549, opened 2025-07-22
Convert html table to markdown

### LINKED ISSUE #2536: HTML Links in entities/README.md break page generation
### Area(s)

_No response_

### What's the problem?

I am currently working on https://github.com/open-telemetry/opentelemetry.io/pull/7337

The newly introduced overview page for entities unfortunately breaks the process because of the HTML links used in `templates/registry/markdown/entity_readme.md.j2` as hugo is not able to auto parse them. 

Since the latest release is done already I will add a work around but it would be good to make sure that no HTML links to markdown files are used (anchors seem to be fine) 

_No response_

trask:
Resolves #2536

I think the result is the same for now, but let's wait for @jsuereth in case he had future plans for this being html...

### DIFF
--- a/docs/registry/entities/README.md
+++ b/docs/registry/entities/README.md
@@ -16,269 +16,81 @@ The entities registry is the catalog of available Entities.
 > Relationships and signal associations are a work in progress.
 
 Currently, the following namespaces exist:
-<table>
-<thead>
-<tr>
-  <th>Namespace</th>
-  <th>Entity</th>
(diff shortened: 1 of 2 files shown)


jsuereth:
Originally, I was trying to have nested tables / blended cells which is why I had HTML. I think I abandoned that, so this is fine.

lmolkova (approved):

arminru (approved):

PR #2551, opened 2025-07-22
feat: New GenAI agent 'role'span attribute, single-agent metrics

91pavan:
New Proposal after feedback from SIG during our weekly meeting. 

### What?

This PR adds two new attributes to the GenAI semantic conventions:

`gen_ai.request.logprobs` (boolean): Captures whether log probabilities are requested for each token in the response. Set as conditionally required when available in the request and not false.

`gen_ai.agent.role` (string): Describes the role of an agent in the operation context (e.g., "researcher", "coder", "assistant"). Conditionally required when provided by the application.

Additionally, this PR:

Adds two new agent-specific metrics: `gen_ai.client.agent.end_to_end_completion_duration` and `gen_ai.client.active_agents`

### Why?

- Agents are frequently used to perform reasoning, planning, verification, or refinement tasks, which may naturally involve looping, retrying, or revisiting.
- An agent might be re-invoked based on a condition. Tools like ReAct naturally involves repeated use of the same agent.
- `gen_ai.client.active_agents` tracks the number of active AI agents currently running in a system. It follows the established OpenTelemetry pattern for monitoring active resource usage.


gyliu513:
@PRATIBHA-Moogi ^^

model/gen-ai/registry.yaml
```diff
@@ -225,6 +225,76 @@ groups:
...
+              brief: 'Task performed by the agent'
+              stability: development
+            - id: agent
+              value: "agent"
+              brief: 'Agent performing the task'
+              stability: development
+            - id: tool
+              value: "tool"
+              brief: 'Tool used by the agent'
+              stability: development
```
gyliu513: not sure if we really need an entity or not, this may make the semantic convention more complex. @91pavan can you share some detail for what is the benefit of introducing the general entity instead of defining different semantic convention for each entity type?

We already have semantic convention for tool and agent, I think we need to add task and workflow in future.

@lmolkova ^^

91pavan: Thanks for the feedback @gyliu513! 
IMO, the benefit of the unified entity approach is that it provides a consistent attribution model across all components in single agent / multi-agent workflows. Rather than having separate attribute sets for each entity type, this allows:

* Unified querying - You can filter spans by entity type and consistently get name, input, output, and outcome regardless of whether it's a task, agent, tool, or workflow type span. Because each agent would spin up multiple entity types as child spans in order to complete a particular opertation.
* Simplified instrumentation
* Future extensibility - New entity types can be added without introducing new attribute namespaces in the future

However, I understand the concern about overlap with existing conventions. Would you prefer an approach where:

- We keep entity-specific conventions (like existing `tool/agent` ones) for detailed attributes
- Use `gen_ai.agent.entity.*` only for cross-cutting concerns like outcome tracking and chain attribution

Or alternatively, should we focus this PR on just adding the missing task and workflow entity types to existing conventions and defer the entity model?

model/gen-ai/spans.yaml
```diff
@@ -260,7 +260,34 @@ groups:
       - ref: gen_ai.data_source.id
         requirement_level:
           conditionally_required: if applicable.
-
+      - ref: gen_ai.agent.entity.type
```
zhirafovod: Should it be `gen_ai.entity.type`? i.e. `gen_ai.entity.type=agent` or `gen_ai.entity.type=tool`

model/gen-ai/spans.yaml
```diff
@@ -260,7 +260,34 @@ groups:
...
+        requirement_level:
+          conditionally_required: if applicable.
+        note: |
+          Type of the entity involved in the span, such as `task`, `agent`, `tool`, or `workflow`.
+          Useful for understanding role-specific behavior within a chain.
+      - ref: gen_ai.agent.entity.name
```
zhirafovod: DITO `gen_ai.entity.name`

model/gen-ai/spans.yaml
91pavan: Oh sorry, I think you're right. A small typo. Let me fix it

model/gen-ai/registry.yaml
```diff
@@ -225,6 +225,76 @@ groups:
         type: string
         brief: Free-form description of the GenAI agent provided by the application.
         examples: ["Helps with math problems", "Generates fiction stories"]
+      - id: gen_ai.agent.entity.type
```
ShipraJain01: @91pavan: Is entity type synonym to span type? can we not use span type itself?

91pavan: IMO, the key difference could be the granularity and scope:

Span name/type typically represents the operation (e.g., "llm.chat", "tool.call", "agent.execute")
Entity type is intended to classify the role within the agent workflow (task, agent, tool, workflow etc)

91pavan: But you are right, it could be derived from the span name field.

model/gen-ai/registry.yaml
PRATIBHA-Moogi: @91pavan I agree with @gyliu513 here that use of gen_ai.agent.entity is not really necessary here. Span.kind can hold the value of entity being task, agent, tool.  Right now at agent invocation level we have gen_ai.agent.xx attributes to be applied based on current version of semantic conventions, similarly at tool invocation level we have gen_ai.tool.xx attributes to be applied. So expressing this way newer set of attributes could be redundant set of information.  Since spans are stitched together based on a unique trace id, you can have a e2e view of all spans belonging to Agent, Tool invocation.

model/gen-ai/registry.yaml
```diff
@@ -225,6 +225,76 @@ groups:
...
+              stability: development
+            - id: failure_technical
+              value: "failure_technical"
+              brief: 'The entity failed due to a technical issue, e.g., an exception or timeout.'
+              stability: development
+        brief: >
+          Execution outcome classification for the entity, such as success, partial_success,
+          failure_logical (e.g., no valid answer), or failure_technical (e.g., exception, timeout).
+          This helps in understanding the success or failure of the entity's execution.
+        examples: ['success', 'partial_success', 'failure_logical', 'failure_technical']
```
thompson-tomo: Should these not be gen_ai.request.* & gen_ai.response.* so that they are alongside other attributes?

model/gen-ai/registry.yaml
```diff
@@ -225,6 +225,76 @@ groups:
...
+          Type of the entity involved in the span, such as task, agent, tool, or workflow.
+          Useful for understanding role-specific behavior within a chain.
+        examples: ['task', 'agent', 'tool', 'workflow']
+      - id: gen_ai.agent.entity.name
+        stability: development
+        type: string
+        brief: >
+          Identifier or name of the entity, such as agent name (e.g., "researcher") or an internal task (e.g., "summarize_input").
+          This is often used to identify the agent or task being performed.
+        examples: ["researcher", "summarize_input"]
```
thompson-tomo: Would avoid using the term entity as it could lead to confusion given entity is an OTEL term.

model/gen-ai/registry.yaml
```diff
@@ -225,6 +225,76 @@ groups:
...
+              brief: 'Tool used by the agent'
+              stability: development
+            - id: workflow
+              value: "workflow"
+              brief: 'Workflow executed by the agent'
+              stability: development
+        brief: >
+          Type of the entity involved in the span, such as task, agent, tool, or workflow.
+          Useful for understanding role-specific behavior within a chain.
+        examples: ['task', 'agent', 'tool', 'workflow']
```
thompson-tomo: As an alternative would gen-ai.operation.category or gen-ai.operation.component or even gen-ai.operation.mode work as it is describing what is performing the operation.

model/gen-ai/registry.yaml
91pavan: Hi @PRATIBHA-Moogi, thanks for the feedback. May I know where it says SpanKind can hold arbitrary values? Searching through the docs, they mention that it can hold either `SERVER`, `CLIENT`, `INTERNAL`, `PRODUCER` or `CONSUMER`. 

https://opentelemetry.io/docs/specs/otel/trace/api/#spankind

IMO, we should still need some sort of an indication to differentiate the task spans vs a workflow span vs an agent span. 

As per one of the suggestions below, I'm wondering if `gen-ai.operation.category` or `gen-ai.operation.component` will work as it is describing what is performing the operation.

model/gen-ai/registry.yaml
91pavan: Thanks for the feedback @thompson-tomo. Given that OTel generally logs any input/output content as `events`, I'm wondering if I should do the same for `workflow` and `task` types.

model/gen-ai/registry.yaml
91pavan: Got it, thank you.

model/gen-ai/registry.yaml
thompson-tomo: Agree about using events and keeping it consistent being ideal. But I do know that there has been push back to Events and was being discussed moving to spans. There is a couple of issues floating around about it.

model/gen-ai/registry.yaml
91pavan: Will take this up as part of a future PR. Thanks!

model/gen-ai/registry.yaml
```diff
@@ -225,6 +230,11 @@ groups:
         type: string
         brief: Free-form description of the GenAI agent provided by the application.
         examples: ["Helps with math problems", "Generates fiction stories"]
+      - id: gen_ai.agent.role
```
lmolkova, line 233: where would it come from? are there prototypes of the instrumentations that collect it?

model/gen-ai/metrics.yaml
```diff
@@ -105,3 +105,25 @@ groups:
     unit: "s"
     stability: development
     extends: metric_attributes.gen_ai
+  - id: metric.gen_ai.client.agent.end_to_end_completion_duration
```
lmolkova, line 108: how is this metric different than `gen_ai.client.operation.duration` with operation name `invoke_agent`?

model/gen-ai/metrics.yaml
```diff
@@ -105,3 +105,25 @@ groups:
...
+    metric_name: gen_ai.client.agent.end_to_end_completion_duration
+    annotations:
+      code_generation:
+        metric_value_type: double
+    brief: 'End-to-end duration of the agent execution.'
+    instrument: histogram
+    unit: "s"
+    stability: development
+    extends: metric_attributes.gen_ai
+  - id: metric.gen_ai.client.active_agents
```
lmolkova: I like the idea, do we have evidence that people are generally running more than one agent per service?

Also, how would someone report it? E.g. if I use server-side agent like AWS bedrock or Azure AI agents, when would I increment/decrement it ? If I run client-side agent, when would I do this?

model/gen-ai/metrics.yaml
```diff
@@ -105,3 +105,25 @@ groups:
...
+    type: metric
+    metric_name: gen_ai.client.active_agents
```
lmolkova: nit: we [avoid pluralizing updowncounter names](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#do-not-pluralize-updowncounter-names)

```suggestion
    metric_name: gen_ai.client.agent.active
```

model/gen-ai/metrics.yaml
91pavan: Hi @lmolkova, thanks for your input. 

The difference would be in the scope and granularity. 

`gen_ai.client.operation.duration` tracks a individual agent execution request, but `gen_ai.client.agent.end_to_end_completion_duration` would try and capture the duration of the potential multi-step agent process. 

- An orchestrator (e.g., planner agent or controller) may route multiple steps to the same agent.
- Tools like reAct, chain-of-thought involve  repeated use of the same agent.
- Fault tolerance scenarios.

But largely, it depends on the design choices adopted by the agentic application.

model/gen-ai/registry.yaml
91pavan: This would be an optional parameter set by the agent app developer. 

https://github.com/agntcy/observe/pull/35/files#diff-f46dc645316aa1840b5b1dfd7c24cd7f058809a363728b99a6adba1bf33d426eR201

Sample trace:
https://cloud.langfuse.com/project/cmdmzj18e00l0ad07sh6gcau8/traces/b3b3fdbed390328dab262788ba6f4753

Sample trace screenshot:
Screenshot 2025-07-31 at 08 38 46

model/gen-ai/metrics.yaml
91pavan: I believe it would hard to track agents that are managed on the cloud provider side (i.e., server-side agents) since we have no lifecycle visibility. At best, we can only do operation metrics. 

But client-side agents, it's possible since we will have control over their lifecycle events. 

In a decorator based agentic observability SDK approach, when a suitable method or the class which is annotated with the suitable `agent` decorator is invoked, we could bump up the `agent.active` counter, and when the execution is finished (when the `finally` block is called in the SDK, we can decrease the count). Just my initial thoughts.

```
def multi_agent_workflow(*agent_roles):
    """Decorator for functions that manage multiple agents"""
    def decorator(func):
        def wrapper(*args, **kwargs):
            # Increment counters for all agents
            for role in agent_roles:
                active_agents_counter.add(1, {"gen_ai.agent.role": role})
            
            try:
                result = func(*args, **kwargs)
            finally:
                # Decrement counters for all agents
                for role in agent_roles:
                    active_agents_counter.add(-1, {"gen_ai.agent.role": role})
            return result
        return wrapper
    return decorator


# Usage
@multi_agent_workflow("researcher", "writer", "reviewer")
def content_pipeline(topic):
    # All three agents are tracked as active during execution
    research = research_agent.gather_info(topic)
    draft = writer_agent.create_content(research)
    final = reviewer_agent.review(draft)
    return final

```

There could be other ways we can also track, e.g., active context manager.

model/gen-ai/registry.yaml
PRATIBHA-Moogi: @91pavan Traceloop/Openllmetry SDK provides extra set of [span.kind](https://www.traceloop.com/docs/openllmetry/contributing/semantic-conventions) such as Tool, Task, Workflow etc. You may have a look at it. Those attributes can be brought into Semantic Conventions here. But We can surely keep them under gen-ai.operation.category or we can introduce gen-ai.operation.span.kind to capture high level of Agents semantic concepts such as Task, Workflow, Tool  etc..

model/gen-ai/registry.yaml
lmolkova: it means some manual level of instrumentation and most users never do anything custom. Besides this, how it is different from the [gen_ai.agent.name](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md#gen-ai-agent-name)?

model/gen-ai/metrics.yaml
lmolkova: Do you envision this metric to be reported by user application? 
I'm not sure if auto-instrumentations would be able to know what's e2e. It seems you're looking for something one level higher than GenAI - some workflow or business-process and I don't think gen_ai is the right namespace for it.

model/gen-ai/metrics.yaml
lmolkova: it seems you're looking for decorator-based solution for gen-ai workflow that requires users to do some work to instrument them. 
To me it sounds like an extra layer on top of GenAI - business processes/workflows do not really depend on who does sub-tasks - AI, humans or traditional code and workflow telemetry doesn't need to be limited to GenAI. So I'd look for a more generic solution.

thompson-tomo: As a thought would https://github.com/open-telemetry/semantic-conventions/pull/2387 help with a generic solution for workflows. 

It would provide you with a Workflow, definitions, execution & tasks.

91pavan:
will create specific PRs, this one seems a bit all over the place!

PR #2552, opened 2025-07-22
Add client-side jank event

breedx-splk:
This adds a new event called `app.jank` that can be used to report when an application experiences slow rendering. 
It is a restart of #2157, intended to be simpler and without the need to discuss why metrics aren't a good fit for client-side reporting of jank. It also omits the activity name and classifier (slow/frozen) to hopefully reduce contention and allow us to build experimental semconv incrementally.

In pseudocode, the jank event looks like this:

```json5
{
  event_name: app.jank,
  timestamp: <standard log/event timestamp>
  attributes: {
    app.jank.count: 3,
    app.jank.period: 5.0,
    app.jank.threshold_ms: 16
  }
}
```

That's it.

The idea is that when RUM (real user montoring) users are looking at a session timeline, it can be helpful to have some indication when slow renders happen, because it influences user interaction behavior. This event is not intended to be a continuous signal like a metric, and it should only be reported for small intervals when jank is detected. 

This semconv also does not specify behavioral aspects of instrumentation. For example, some implementations might report more than one jank event for the same period, but with different thresholds. This is purposefully not covered by this semantic convention. It is purely intended to define the event and specify which attributes are expected on the event. 

Another reminder for reviewers, because this is not always obvious: Most of the content (everything between html style comments) in the docs/markdown files is auto-generated from the yaml, so comments about that should probably be comments about the source yaml.


model/app/registry.yaml
```diff
@@ -31,6 +31,24 @@ groups:
...
         examples:
           - 2ab2916d-a51f-4ac8-80ee-45ac31a28092
+      - id: app.jank.count
+        type: int
+        stability: development
+        brief: A count of the number of frame renders that experienced jank.
+        note: >
+          Depending on platform limitations, the value provided
+          MAY be approximation.
+        examples: [ 9, 42 ]
```
thompson-tomo: Would this be better suited as event.count as it is describing the number of times the event occurred within quick succession? Obviously needs a note saying that client should not be aggregating the values but rather forwarding the values from the source.

model/app/events.yaml
```diff
@@ -37,3 +37,20 @@ groups:
...
+    note: >
+      Jank happens when the UI is rendered slowly enough for the user to
+      experience some disruption or sluggishness.
+    attributes:
+      - ref: app.jank.count
+        requirement_level: required
+      - ref: app.jank.threshold_ms
+        requirement_level: required
+      - ref: app.jank.period
+        requirement_level: required
```
thompson-tomo: Requirement levels seems too strong. For me 1 & 3 should be opt-in, 2 should be recommended.

model/app/registry.yaml
```diff
@@ -31,6 +31,24 @@ groups:
...
+      - id: app.jank.threshold_ms
+        stability: development
+        type: int
+        brief: The minimum rendering threshold for this type of jank, in milliseconds.
+        examples: [ 16, 700, 1024 ]
+      - id: app.jank.period
+        stability: development
+        type: double
+        brief: The time period, in seconds, for which this jank is being reported.
+        examples: [ 0.5, 1.0, 10.0 ]
```
thompson-tomo: Would this be better suited as event.period as it is describing the number of times the event occurred within quick succession during that period? Obviously needs a note saying that client should not be aggregating the values but rather forwarding the values from the source. Also we should have an event.starttime attribute

model/app/events.yaml
```diff
@@ -37,3 +37,20 @@ groups:
         requirement_level: opt_in
       - ref: app.screen.coordinate.y
         requirement_level: opt_in
+  - id: event.app.jank
+    stability: development
+    type: event
+    name: app.jank
```
thompson-tomo, line 43: Would append "report" to the name to indicate it is a report.

model/app/registry.yaml
breedx-splk: It's not a count of the number of times the event happened, it is a count of the number of frames. Those are different things. I would consider `app.jank.frames.count` if that's prefrable.

Attributes are namespaced, so your suggestion would indicate that this is a generic `count` attribute within the `event` namespace, which really isn't an attribute namespace afaik (is it defined elsewhere?). My preference is to keep it within the `app.jank` namespace because this specifically a count related to jank.

model/app/events.yaml
breedx-splk: I'll concede that `required` is probably too strict for all. I'll try this first pass with `recommended` then and we can go from there. I'm not convinced we should use opt-in for these yet.

model/app/registry.yaml
breedx-splk: I disagree for the same reasons above about attribute namespaces. This is a jank-specific attribute, not an event specific attribute. And no, this will not include a description of instrumentation behavior...it's just a description of the data, its type, and what it means. 

All events have a timestamp, so no, we should not add another timestamp attribute.

model/app/events.yaml
breedx-splk: It's not a report. It's an event. I'm not aware of any other convention that names events with a `.report` suffix and I think it's unnecessary.

model/app/registry.yaml
bidetofevil: `app.jank.frames_count` or ` app.jank.frames` is preferable to me because `frames` doesn't seem like a concept that needs further breaking down, so using a namespace might not be necessary?

model/app/events.yaml
bidetofevil: I think without something to contextualize the event, it's hard to tell if it's actually a problem just based on the data. 

That said, if someone has a standard definition internally and just want to fire the event without additional context, they can omit all of them - they have the knowledge of what they are dealing with when they get one of these, and we shouldn't prevent that.

I'd say `recommend` for all three?

model/app/registry.yaml
bidetofevil: This is the period in which the dropped frames were collected, enough to fire a jank event, so this feels correctly named.

In terms of start time, for the context of this event, it doesn't make sense. The event is triggered when a certain chunk of time contained enough frame drops to constitute a reportable issue. The start/end times are arbitrary and not necessarily when the jank started happening (which could be at any point during the chunk of time)

Besides, if you have the period and the time the event is logged, you can probably figure out when it started, so folks can log this themselves if they want.

If this proves to be useful in the future, we can add it then - now, it seems a bit of an overkill.

model/app/events.yaml
bidetofevil: I think `jank` is clear enough of an noun to leave as is. In a way, all telemetry are reports of some sort :-)

docs/app/app.md
```diff
@@ -80,6 +82,42 @@ Use this event to indicate that visual application component has been clicked, t
...
+The event name MUST be `app.jank`.
+
+This event indicates that the application has detected substandard UI rendering performance.
+
+Jank happens when the UI is rendered slowly enough for the user to experience some disruption or sluggishness.
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`app.jank.count`](/docs/registry/attributes/app.md) | int | A count of the number of frame renders that experienced jank. [1] | `9`; `42` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`app.jank.period`](/docs/registry/attributes/app.md) | double | The time period, in seconds, for which this jank is being reported. | `0.5`; `1.0`; `10.0` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
bidetofevil: Why double? It feels like having fractions of a second is overkill in term of granularity, and making it a int seems good enough

docs/app/app.md
```diff
@@ -80,6 +82,42 @@ Use this event to indicate that visual application component has been clicked, t
...
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `app.jank`.
+
+This event indicates that the application has detected substandard UI rendering performance.
+
+Jank happens when the UI is rendered slowly enough for the user to experience some disruption or sluggishness.
```
bidetofevil, line 106: Do we want to clarify the methodology of how one of these are generated to clarify what the attributes means?

e.g. This event should be reported when the instrumentation detects enough frame drops within a certain period of time that would constitute something noticeable to the user. How that is surfaced is unspecified, but the event should describe the parameters if they are not clearly defined and understood in a particular use case.

This is non-blocking. Feel free to ignore this if it doesn't jive with what you want

docs/app/app.md
breedx-splk: I thought it would allow more flexibility in scheduling/polling...but I don't feel strongly about that. I'll relax it to int and if that doesn't work we can reconsider. 👍🏻

docs/app/app.md
breedx-splk: I appreciate that, but as I said in the original description, 
> This semconv also does not specify behavioral aspects of instrumentation.

That was intentional, to try and keep this smaller in scope and limit the amount of api-type bikeshedding.

model/app/registry.yaml
thompson-tomo: yes, `event` is already a defined namespace (https://opentelemetry.io/docs/specs/semconv/registry/attributes/event/). The event is the jank occurring but because those events are being aggregated, it makes sense to put that count on the event.

model/app/events.yaml
thompson-tomo: > if someone has a standard definition internally and just want to fire the event without additional context, they can omit all of them 

I am coming from the standard OTEL definition of events which is that they are sent as soon as they are generated, happy for it to be recommendedd only if the event name is changed.

model/app/events.yaml
thompson-tomo: but it is a report as you are aggregating the events to form a report/summary which can then be emitted as an event. Hence the suffix is needed. 

the app.jank event should not have aggregation.

model/app/registry.yaml
thompson-tomo: as stated in the other comment, you are wanting instrumentation to aggregate the events which come in and then emit them based on thresholds.

> this will not include a description of instrumentation behavior
We need to ensure that this is being instrumented in a consistent manner.

Start time does make sense especially as you are aggregating events to form a report. The event time cannot be reliably used as there can be a delay between event being sent and period elapsed.

docs/app/app.md
thompson-tomo: As per the pr description a prototype is needed

model/app/registry.yaml
LikeTheSalad: +1 on the mention of frames in the name. I like the one that @bidetofevil mentioned: `app.jank.frames_count`.

bidetofevil (approved):
LGTM. I would do the s/sount/[frame_count|frames] rename but that's only a slight preference and is non-blocking.

I have a nit that I will follow up separately 😅

model/app/registry.yaml
breedx-splk: ok, let's try with `app.jank.frame_count`. I looked for other `_count` examples and all but one used a singular noun before the `_count`.

model/app/registry.yaml
```diff
@@ -31,6 +31,23 @@ groups:
...
+        brief: A count of the number of frame renders that experienced jank.
+        note: >
+          Depending on platform limitations, the value provided MAY be approximation.
+        examples: [ 9, 42 ]
+      - id: app.jank.threshold_ms
+        stability: development
+        type: int
+        brief: The minimum rendering threshold for this type of jank, in milliseconds.
+        examples: [ 16, 700, 1024 ]
+      - id: app.jank.period
```
lmolkova, line 46: is there a way we can be consistent in units between `app.jank.threshold_ms` and `app.jank.period` ? Given that we default to seconds on all metrics, is there a reasonable path with `app.jank.threshold` in seconds?

model/app/registry.yaml
```diff
@@ -31,6 +31,23 @@ groups:
           More information about Android identifier best practices can be found [here](https://developer.android.com/training/articles/user-data-ids).
         examples:
           - 2ab2916d-a51f-4ac8-80ee-45ac31a28092
+      - id: app.jank.frame_count
+        type: int
+        stability: development
+        brief: A count of the number of frame renders that experienced jank.
```
lmolkova: nit: would this simplified version make sense?

```suggestion
        brief: A number of frame renders that experienced jank.
```

model/app/registry.yaml
breedx-splk: I guess if we're trying to universally kinda standardize on seconds, then sure. I didn't think it was as important for event attributes, especially in cases like this where it will almost always be sub-second, but I guess I can make that change.

model/app/registry.yaml
```diff
@@ -31,6 +31,23 @@ groups:
...
+        stability: development
+        brief: A number of frame renders that experienced jank.
+        note: >
+          Depending on platform limitations, the value provided MAY be approximation.
+        examples: [ 9, 42 ]
+      - id: app.jank.threshold
+        stability: development
+        type: int
+        brief: The minimum rendering threshold for this jank, in seconds.
+        examples: [ 16, 700, 1024 ]
```
KalleOlaviNiemitalo: The example thresholds 700 seconds and 1024 seconds seem impractically long.

breedx-splk: Thank you, good catch.

lmolkova:
@breedx-splk could you please add links to instrumentations/prototypes to the PR description? thanks!

model/app/registry.yaml
```diff
@@ -31,6 +31,23 @@ groups:
...
+          Depending on platform limitations, the value provided MAY be approximation.
+        examples: [ 9, 42 ]
+      - id: app.jank.threshold
+        stability: development
+        type: double
+        brief: The minimum rendering threshold for this jank, in seconds.
+        examples: [ 0.016, 0.700, 1.024 ]
+      - id: app.jank.period
+        stability: development
+        type: int
```
lmolkova: nit: is there a chance it needs to be double?

jsuereth:
@breedx-splk Can you add a link to the prototype in the PR description?

model/app/registry.yaml
breedx-splk: yeah thanks, when switching from int ms to seconds, it does need to be float. I'll fix it. 👍🏻

breedx-splk:
> @breedx-splk Can you add a link to the prototype in the PR description?

Sure. I hope to have something worked up this week. It's kinda funny, because we currently use a zero-duration span in Android, and part of the goal of having a semconv was to be able to switch to an actual event....so 🐔 🥚 situation. But I think I have a way forward.

breedx-splk:
> @breedx-splk Can you add a link to the prototype in the PR description?

Ok here https://github.com/open-telemetry/opentelemetry-android/pull/1175
There's more than one thing happening there, but the main event generator is this class here: https://github.com/open-telemetry/opentelemetry-android/pull/1175/files#diff-cecbb46cb90250501402cd2fd1405454f7fa168746edd2bc1681902e08e692a2R21

lmolkova (approved):

Issue #2555, opened 2025-07-24
Ensure all enum members have brief


thompson-tomo:
Following on from #2507 a number of enum members don't have briefs, as such, they should be added in an orderly manner so that rules could be added to enforce it at a later point in time.

- #2556
- #2557
- #2558
- #2559
- #2560
- #2636
- #2637
- #2638
- #2639
- #2640
- #2650
- #2651 
- #2652 
- #2654 
- ....

lmolkova:
I do not support any further progress on this issue with reasoning outlined in https://github.com/open-telemetry/semantic-conventions/pull/2560#pullrequestreview-3129014040 - duplicating member value in different casing does not add value to otel conventions.

thompson-tomo:
Reasoning was based on the assumption that only the first letter was being capitalised however in many cases that is not the case.

lmolkova:
The core question is what new information is brought. Neither different casing nor AI generated obvious version help much - they don't add new information.

Ideally enum attribute has docs or full explanation on what it is and pointers to external standards/docs where possible values are defined. Enum members may have pointer to external docs or well thought-out explanation that add something non-obvious.

[Code is a liability, less is more, No code is the best way to write secure and reliable applications. Write nothing; deploy nowhere.](https://github.com/kelseyhightower/nocode)

thompson-tomo:
For me it is better to be more explicit in descriptions & documentation rather than assuming that everyone will come to a common understanding as if you don't that leads to issues.

Anyway most of the members in the pr's have been updated to use either the correctly stylised proper noun for names or to be a self contained explanation sourced from official docs where possible.

PR #2560, opened 2025-07-24
[chore] Set enum briefs for Database

thompson-tomo:
Progresses #2555

## Changes

### DIFF
--- a/docs/database/database-metrics.md
+++ b/docs/database/database-metrics.md
@@ -442,8 +442,8 @@ This metric is [required][MetricRequired].
 
 | Value  | Description | Stability |
 |---|---|---|
-| `idle` | idle | ![Development](https://img.shields.io/badge/-development-blue) |
-| `used` | used | ![Development](https://img.shields.io/badge/-development-blue) |
+| `idle` | Idle | ![Development](https://img.shields.io/badge/-development-blue) |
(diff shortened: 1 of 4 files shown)


thompson-tomo:
Awaiting review

joaopgrassi (approved):

thompson-tomo:
I agree that having more detailed briefs would be ideal however that is a more longer term wish. This pr is about bringing consistency to the definitions as a large number of enums already follows this pattern while simplifing codegen/documentation.

lmolkova:
> This pr is about bringing consistency to the definitions as a large number of enums already follows this pattern while simplifing codegen/documentation.

this can be achieved with jinja filter that renders brief in sentence case. There is no need to duplicate things in the yaml.

thompson-tomo:
So looking at the diff, there is a number of scenarios where it is not just the first letter being changed and also sentence case doesn't apply in all cases ie proper nouns hence is useful to have them explicitly defined.

lmolkova:
It should be possible to do this in jinja as well. Replace `.`, `_` and change in casing with a space. I'm not suggesting to do it. The value of this change is still questionable. 

Authors of semconv that feel that brief is necessary and beneficial would add it. There is no point in forcing or faking it.

Issue #2010, opened 2025-03-19
Capture GenAI prompts and completions as events or attributes


lmolkova:
The GenAI SIG has been discussing how to capture prompts and completion for a while and there are several issues that are blocked on this discussion (#1913, #1883, #1556) 

What we have today on OTel semconv is a set of structured events exported separately from spans. This was implemented in #980 and discussed in #834 and #829. The motivation to use events was to
- overcome size limits on attribute values by using event body
- have a signal that supports structured body and attributes
- have a clear 1:1 relationship between event name and structure (as opposed to polymorphic types or arrays or heterogeneous objects)
- making it possible and easier to consume individual events and events in general without spans
- having verbosity controls

Turns out that:
- after ~9 month events are not adopted by GenAI-focused tracing tools and their external instrumentation libs including [Arize](https://docs.arize.com/arize/llm-tracing/tracing/semantic-conventions), [Traceloop](https://www.traceloop.com/docs/openllmetry/contributing/semantic-conventions#llm-foundation-models), [Langtrace](https://github.com/Scale3-Labs/langtrace-trace-attributes/blob/main/schemas/llm_span_attributes.json) - all these providers use attributes to capture prompts and completions.
- These backends consume prompts and completions along with spans and don't envision separating them and they store and visualize this data along with spans

So, the GenAI SIG is re-litigating this decision taking into account backend feedback and other relevant issues: #1621, #1912, https://github.com/open-telemetry/opentelemetry-specification/issues/4414

-------------------

**The fundamental question is whether this data belongs on the span or is a separate thing useful without a span.**

How it can be useful without a span:
- audit logs - https://cloud-native.slack.com/archives/C06KR7ARS3X/p1742322601090389?thread_ts=1741895340.932419&cid=C06KR7ARS3X - we could capture them on the request-response payloads where they are not unified/filtered/altered in other ways. But also audit logs have different deliver guarantees/storage/retention needs than telemetry
- some applications don't use tracing and rely on logs
- to be useful without a span, events should probably contain more details - endpoint, model used, input parameters, etc - it's not the case today

Are prompts/completions are point-in-time telemetry?
- they don't really have a timestamp - prompts are input parameters, completion comes at the end of non-streaming call, buffered completion comes at the end of the streaming call. (#1701, https://github.com/open-telemetry/semantic-conventions/issues/1621#issuecomment-2552335590)
- Streaming chunks, if captured at all, would have timestamps (#1964)

[...]

PR #2563, opened 2025-07-24
Gen AI Evaluation Event

singankit:
Fixes #

## Changes


model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,42 @@ groups:
...
+        brief: Metadata associated with the evaluation input.
+        examples: ['{\"requestId\": \"fab3ee5d-a3c6-4c47-b3de-901bf02fa045\"}']
+        note: >
+          The structure is specific to the evaluator. If the metadata is structured, it
+          is RECOMMENDED to provide it in a structured form using language-specific API. It
+          can also be captured as a JSON string when structured API is not available. If
+          metadata properties contain any sensitive information such as prompts or
+          completions, corresponding properties MUST NOT be recorded by default. Instrumentations
+          MAY provide a way to override this behavior and record sensitive information in
+          the metadata if user explicitly allows it.
```
thompson-tomo: Could this not be gen_ai.request.metadata so that it sits alongside the other request attributes?

singankit: Aim of this attribute is to capture metadata related to input used for evaluation. Have a similar property for output metadata. Open to suggestions if there are better ways to capture it.  `gen_ai.request.metadata` seems generic to capture metadata related to input and output used for evaluation.

thompson-tomo: I would also change the output attribute to gen_ai.response.metadata.

On further review and comparison to existing semconv for gen ai, it might be beneficial to have 2 events. 1 for request & 1 for reply similar to the others. Also you could then use the body to include the metadata.

singankit: GenAI Sem Conv is moving away from using events for capturing input/output. More details can be found in this [issue](https://github.com/open-telemetry/semantic-conventions/issues/2010).

(But evaluation cannot be captured as attributes since attributes cannot be added after span closes, and evaluation **usually**(not always) happens after traces are complete.)

Regarding your point of capturing input/output for evaluation, please let me know if I am formulating your concern correctly?
- Capture input/output for evaluation as separate events and have a generic metadata field that can be used for input and output event.

If I understood your feedback correctly. The focus of the PR is to capture the final evaluation score and associate it with a span via an event. Traces/span on how this score was calculated (input, any processing logic like transformation) can be an enhancement in a future PR. For prompt based evaluation (llm-as-a-judge), score generation can still be captured as a GenAI Span similar to chat completion, but that is out of scope of this PR.

thompson-tomo: thanks, i am aware of that discussion. As the intention is to move away from events, should you not be creating a new root span for the evaluation which can then have a link to the original span? That way you can use attributes on the span and cater for both use case.

My original point was and still stands is that the attribute for metadata should reside in the same sub-namespace as other request and response attributes.

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,57 @@ groups:
 
           Additional output format details may be recorded in the future in the
           `gen_ai.output.{type}.*` attributes.
+      - id: gen_ai.evaluation.name
+        stability: development
+        type: string
+        brief: The qualified name of the evaluator used to evaluate the GenAI response.
+        examples: ["Relevance", "IntentResolution"]
```
lmolkova, line 486: nit: the brief mentions that the name should be qualified, but examples are not following the recommendation

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,57 @@ groups:
...
+        note: >
+          This attribute identifies the evaluation name.
+      - id: gen_ai.evaluation.score
+        stability: development
+        type: double
+        brief: The evaluation score for the GenAI response.
```
lmolkova: maybe

```suggestion
        brief: The evaluation score returned by the evaluator
```

just suggesting to makit it a bit more broad

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,57 @@ groups:
...
+        examples: [4.0]
+      - id: gen_ai.input.metadata
```
lmolkova: do we have a real-life example on where/how it would be useful?

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,57 @@ groups:
...
+          can also be captured as a JSON string when structured API is not available. If
+          metadata properties contain any sensitive information such as prompts or
+          completions, corresponding properties MUST NOT be recorded by default. Instrumentations
+          MAY provide a way to override this behavior and record sensitive information in
+          the metadata if user explicitly allows it.
+        examples: ['{\"Perplexity\": 1.335}']
+        requirement_level: recommended
+      - id: gen_ai.evaluation.reasoning
+        stability: development
+        type: string
```
lmolkova, line 504: do we know for sure that string is and will be enough?

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,57 @@ groups:
...
+        brief: A free-form reasoning for the assigned score provided by the evaluator.
+        examples: ["The response is factually accurate but lacks sufficient detail to fully address the question."]
+      - id: gen_ai.evaluation.metadata
+        stability: development
+        type: string
+        brief: Additional metadata associated with the evaluation.
+        examples: ['{\"evaluator_version\": \"1.2.0\", \"gen_ai.thread.id\": \"thread_ggguJ0iZXRPjUnCy9vT9Fdvs\"}']
```
lmolkova: all the gen_ai attributes should be top-level ones, we should not hide them under `gen_ai.evaluation.metadata` - we want people to aggregate and filter by them  without checking nested objects.

Things like evaluator version would be good candidates for top-level attributes.
Is there a practical reason we need this property bag right away?

model/gen-ai/events.yaml
```diff
@@ -313,3 +313,88 @@ groups:
...
+
+  - id: event.gen_ai.evaluation.result
+    name: gen_ai.evaluation.result
+    type: event
+    stability: development
+    brief: >
+      This event describes a generic GenAI response evaluation result.
+    extends: gen_ai.common.event.attributes
+    attributes:
+      - id: gen_ai.evaluation.name
```
lmolkova: you can  only declare attributes inside attribute groups, not inside signal definitions. Please run `make check-policies` locally (in WSL if you use windows).

once defined in the registry, it's possible to reference it here like other attributes

lmolkova (commented):
Are there publicly available prototypes of the code emitting evaluation results?  Please link them in the PR description

model/gen-ai/registry.yaml
singankit: Thanks @thompson-tomo for the feedback. I have moved the metadata attribute to the gen_ai namespace and using spans to capture evaluation results seemed a better approach and PR reflects that too. Please review and share your feedback.

model/gen-ai/registry.yaml
singankit: Have updated the description to avoid this confusion. Please share your feedback.

model/gen-ai/registry.yaml
singankit: I have moved this as a span attribute and added an example. Please share your feedback.

model/gen-ai/registry.yaml
singankit: So far string seems sufficient to capture it. In future can a field be updated to support multiple types ?

model/gen-ai/registry.yaml
singankit: Have moved metadata as a top-level attribute during gen_ai namespace.

model/gen-ai/events.yaml
singankit: Ran this after making changes for using span instead of events for evaluation results.

dmontagu:
One issue I have with creating a separate span for tracking each evaluation score is that it makes it harder (at least with the way we index spans..) to write at least a couple classes of queries for specific cases:

* Find all cases where the total number of tokens for the task is above a threshold and a score is below a threshold (or vice versa)
    * This would require comparing information about token usage (or some other attribute, in the abstract) from the task execution span with the evaluation score from the evaluation score span
* Find all cases where score A is above a threshold and score B is below a threshold
    * This would require comparing two different evaluation score spans

It would work way better for us if there was a way that we could, while complying with the semantic conventions, put all this information as attributes of a single span so that we can query it at once. I guess we can do that _in addition_ to complying with the semantic convention, this just sticks out to me as an unfortunate aspect of this design.

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,62 @@ groups:
 
           Additional output format details may be recorded in the future in the
           `gen_ai.output.{type}.*` attributes.
+      - id: gen_ai.evaluation.name
+        stability: development
+        type: string
+        brief: The name of the evaluation used for the GenAI response.
+        examples: ["Relevance", "IntentResolution"]
+        note: >
+          This attribute identifies the evaluation name.
```
lmolkova: nit

```suggestion
```

since it does not seem to convey anything new not mentioned in the brief

model/gen-ai/registry.yaml
lmolkova: I still don't see where it would be useful. Do we have a real-life example? There are multiple ways to record request-id things, one of them is e.g. [`azure-service-request-id`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/azure.md#azure-service-request-id). 

Most importantly - why would someone use a property bag here instead of top-level attributes?

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,62 @@ groups:
...
+        examples: ['{\"requestId\": \"fab3ee5d-a3c6-4c47-b3de-901bf02fa045\"}']
+        note: >
+          The structure is specific to the implementation. If the metadata is structured, it
+          is RECOMMENDED to provide it in a structured form using language-specific API. It
+          can also be captured as a JSON string when structured API is not available. If
+          metadata properties contain any sensitive information such as prompts or
+          completions, corresponding properties MUST NOT be recorded by default. Instrumentations
+          MAY provide a way to override this behavior and record sensitive information in
+          the metadata if user explicitly allows it.
+      - id: gen_ai.output.metadata
```
lmolkova: same question here - what are the real-life use-cases for it. Can we recommend using top-level attributes instead?

model/gen-ai/registry.yaml
lmolkova: it can be, yes, but if we envision the need to support other cases, I'd consider `any` type instead

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,62 @@ groups:
...
+      - id: gen_ai.evaluation.reasoning
+        stability: development
+        type: string
+        brief: A free-form reasoning for the assigned score provided by the evaluator.
+        examples: ["The response is factually accurate but lacks sufficient detail to fully address the question."]
+      - id: gen_ai.metadata
+        stability: development
+        type: string
+        brief: Metadata associated with Gen AI operation.
+        examples: ['{\"gen_ai.thread.id\": \"thread_ggguJ0iZXRPjUnCy9vT9Fdvs\"}']
```
lmolkova: this should be a top-level attribute, not a property in the metadata where it can't be correlated with other telemetry easily

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,62 @@ groups:
...
+          MAY provide a way to override this behavior and record sensitive information in
+          the metadata if user explicitly allows it.
+        examples: ['{\"Perplexity\": 1.335}']
+        requirement_level: recommended
+      - id: gen_ai.evaluation.reasoning
+        stability: development
+        type: string
+        brief: A free-form reasoning for the assigned score provided by the evaluator.
+        examples: ["The response is factually accurate but lacks sufficient detail to fully address the question."]
+      - id: gen_ai.metadata
```
lmolkova: we have input metadata, output metadata and just metadata - I suggest we remove all of them until we have a real-life use-case that can't be addressed with top-level attribute

model/gen-ai/spans.yaml
```diff
@@ -309,3 +309,57 @@ groups:
         requirement_level: required
       - ref: aws.bedrock.knowledge_base.id
         requirement_level: recommended
+
+  - id: span.gen_ai.evaluation.result.internal
+    type: span
+    stability: development
+    span_kind: internal
+    brief: >
+      Describes GenAI evaluation span - a request to evaluate a GenAI response or output.
```
```suggestion
      This span captures the process and the result of evaluating GenAI output for quality, accuracy, or other characteristics.
```

model/gen-ai/spans.yaml
```diff
@@ -309,3 +309,57 @@ groups:
...
+    note: |
+      `gen_ai.operation.name` SHOULD be `evaluation`.
+
+      **Span name** SHOULD be `evaluation {gen_ai.evaluation.name}`.
+
+      This span is typically used when instrumenting evaluation frameworks or services
+      that assess the quality, accuracy, or other characteristics of GenAI outputs.
```
lmolkova: do we need this sentence? I left a suggestion on the brief that probably conveys this information.

singankit:
> One issue I have with creating a separate span for tracking each evaluation score is that it makes it harder (at least with the way we index spans..) to write at least a couple classes of queries for specific cases:
> 
> * Find all cases where the total number of tokens for the task is above a threshold and a score is below a threshold (or vice versa)
>   
>   * This would require comparing information about token usage (or some other attribute, in the abstract) from the task execution span with the evaluation score from the evaluation score span
> * Find all cases where score A is above a threshold and score B is below a threshold
>   
>   * This would require comparing two different evaluation score spans
> 
> It would work way better for us if there was a way that we could, while complying with the semantic conventions, put all this information as attributes of a single span so that we can query it at once. I guess we can do that _in addition_ to complying with the semantic convention, this just sticks out to me as an unfortunate aspect of this design.

Thank you for your detailed feedback. I agree that having all relevant metrics as attributes on a single span would simplify querying and analysis, however not fully sure if this approach is flexible to cover various different scenario especially with async evaluations

To clarify with a concrete scenario:

- On Day 1, I select 3 evaluation metrics for my GenAI application.
- On Day 5, I realize that 3 metrics are not sufficient and decide to add 2 more, bringing the total to 5 metrics.

If I need to retroactively compute the 2 new metrics on existing traces that already contain the original 3 metrics, would the recommended approach be to generate new spans for these additional metrics? Or is there a preferred way to update the original span with the new evaluation results, while still adhering to the semantic conventions?

A similar situation could arise if some evaluation metrics are computed asynchronously or by a downstream service at different times. In such cases, would each metric (or set of metrics computed together) necessarily require a separate span, or is there flexibility to consolidate them as attributes on the original span?

Thanks again for your insights. I’m keen to understand the best practices for handling evolving needs of asynchronous evaluation workflows in line with the semantic conventions.

- Would evaluation scores as events help for this case? If yes I can start a new issue to discuss if evaluations should be emitted as events in addition to span

model/gen-ai/spans.yaml
```diff
@@ -309,3 +309,57 @@ groups:
...
+      - ref: gen_ai.operation.name
+        requirement_level: required
+      - ref: gen_ai.evaluation.name
+        requirement_level: required
+      - ref: gen_ai.evaluation.score
+        requirement_level:
+          conditionally_required: if evaluation completed successfully
+      - ref: gen_ai.evaluation.label
+        requirement_level:
+          conditionally_required: if evaluation completed successfully
```
lmolkova: what would I set here if there is only a number? e.g. relevance is `3.5` - what would be the text here? 

BTW I think it can't be required (even if eval is successful)

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,62 @@ groups:
...
+          This attribute identifies the evaluation name.
+      - id: gen_ai.evaluation.score
+        stability: development
+        type: double
+        brief: The evaluation score returned by the evaluator.
+        examples: [4.0]
+      - id: gen_ai.evaluation.label
+        stability: development
+        type: string
+        brief: Human readable label for evaluation.
```
lmolkova, line 495: Following current naming and description it can be absolutely anything, and it's also a required (when successful) attribute, which is not a great combination.

could we elaborate what it means? what would someone put here? 

Is it a text representation of the score where the number is not helpful/applicable? would you use them at the same time or are they mutually exclusive?  Should we do  `gen_ai.evaluation.score.number` and `gen_ai.evaluation.score.label | text | etc` to make name accurate?

docs/gen-ai/gen-ai-spans.md
```diff
@@ -351,6 +352,98 @@ Datastore: A tool used by the agent to access and query structured or unstructur
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+### Evaluation span
```
lmolkova: getting back to @samuelcolvin question on the SIG call 8/5 - what if there is one eval call but it returns multiple scores - with the current design we can't accommodate this - we'd need to have an array of evals and a complex attribute for this. 

From UX perspective, I believe we need to focus on two scenarios:

1. Show a list of eval scores next to GenAI telemetry similar to what AI-focused tools do today. For this case it's not that important how this data is recorded - UX will make it user-friendly

2. Let users build dashboards and alerts based on evaluation scores. For this case, it's much easier to export one eval result as one telemetry item. Then I can write a query like `select avg(attributes["gen_ai.eval.score"]) from events where event_name == "gen_ai.eval.result"` rather than expand lists in span attributes, parse and aggregate across them.

I wonder what @alexmojaki thinks since he brought up "why it should be event and not the span" question in the past. It seems we have an argument on why it could be an event.

Some options to help us satisfy batch-eval case:

1. evals are reported as arrays on spans - see above on the cons (tl;dr - hard to query, alert on)
    - 1.5 - report evals as metrics as well - we'd have a metric reported per eval method. It'd solve dashboard/alerting problem for backends that support metrics well. AFAIK most newish AI-focused vendors are not good friends with metrics.
   
2. report span per each eval method/score - this feels really artificial, but it allows us to solve the problem of two contexts - the one where eval happens and the other one where the original genai output was produced.

3. report the process of evaluation as span, but results as events. This has a small problem - what events should be children of - the eval span (and then it's hard to correlate them to GenAI output being evaluation) or the GenAI span they're evaluating (and then it's hard to correlate them to the eval span). 
    - pros: 
      - consistent with user evals which is an event
      - consistent with what we'd expect external eval service to produce - we won't expect eval-as-a-service to share much of internal telemetry on how it runs evals, but results are a must.
      - works well for multiple-evals-with-one-call cases
    - cons: 
      - two telemetry items even when there is 1 eval 
      - finding eval span from eval result requires finding all back-links to the parent

image

I think we need to keep talking about it. What may be helpful is if we try to prioritize UX we want to provide and also try to show how this proposal applies to common eval frameworks..  

I'm also really curious to know if it's common to ask LLM to provide multiple evals with one call. If it's a common case, we really need to find a way to solve it.

dmontagu:
> If I need to retroactively compute the 2 new metrics on existing traces that already contain the original 3 metrics, would the recommended approach be to generate new spans for these additional metrics? Or is there a preferred way to update the original span with the new evaluation results, while still adhering to the semantic conventions?

For the most part, I typically think of traces as static/immutable after their root span closes (I know that's not technically the case, and there are reasonable scenarios where that is explicitly avoided, but still). I think we should just stick to that mental model here — if you want to "mutate" an evaluation run/experiment/whatever-you-want-to-call-it, my personal feeling is that that should happen in some application-layer logic, and just rely on OTel for a static record of what happened when it happened. For example, you could create a new evaluation run where you just copy the old run's outputs for the metrics you've already computed, and compute new ones where you want.

I would personally have no problem just generating a new trace for the updated evaluation results (even if that meant copying execution data from an older trace or otherwise referencing it via span links or something), I don't need to extend an old trace. Maybe others feel differently, but just sharing my opinion.

I'll note that the idea of adding additional metrics also feels somewhat awkward because, I can always add new metrics to an old trace, but what about redefining an existing metric? Presumably that opens more of a can of worms about what it means to "overwrite" old spans? I personally feel it's better to avoid the whole issue by not encouraging this pattern. Just my 2c

model/gen-ai/registry.yaml
singankit: Idea if this attribute was to have a place to capture any metadata related to evaluation input that cannot be captured by existing attributes. If it does not add value right now , I can remove it and once I have more concrete examples we can resurface this discussion.

model/gen-ai/registry.yaml
singankit: Looking for suggestions on:
1. Keep it string until we find a need to support other cases.
2. Make it `any` now.
Open to both, if needed I can make this change.

model/gen-ai/registry.yaml
singankit: The idea for label is to have an easy way for users to understand the outcome of eval. 

For example, I created a chat bot where any harmful content generated by bot should result in fail for say violence (a metric I measure, assuming it has likert scale 1-5). Even 1 is a `fail`. For someone in gaming industry threshold might be higher where 0-3 are acceptable (`pass`) and 4-5 score as (`fail`) outcome. 

Another Example:
For groundedness can I show label as `grounded` and `not grounded` instead of say score of 0 and 1 respectively.

This attribute being required is an error on my side. I shall fix it.

model/gen-ai/registry.yaml
singankit: Same as for metadata.input .If it does not add value right now , I can remove it and once I have more concrete examples we can resurface this discussion.

model/gen-ai/registry.yaml
singankit: Reason I did not put it as top level property is because threads are specific to assistants API from OpenAI and it is not there in Responses API. I treat it as an agentic framework specific attribute. Open to suggestion on best way to handle framework specific attributes in OTEL

model/gen-ai/spans.yaml
singankit: Thanks removed it.

model/gen-ai/spans.yaml
singankit: This is an error on my side. Have updated it to be `requirement_level: recommended`

model/gen-ai/registry.yaml
singankit: Looking for suggestion on what would be the best attribute to capture information like evaluator used to produce evaluation score ?

singankit:
> > If I need to retroactively compute the 2 new metrics on existing traces that already contain the original 3 metrics, would the recommended approach be to generate new spans for these additional metrics? Or is there a preferred way to update the original span with the new evaluation results, while still adhering to the semantic conventions?
> 
> For the most part, I typically think of traces as static/immutable after their root span closes (I know that's not technically the case, and there are reasonable scenarios where that is explicitly avoided, but still). I think we should just stick to that mental model here — if you want to "mutate" an evaluation run/experiment/whatever-you-want-to-call-it, my personal feeling is that that should happen in some application-layer logic and just rely on OTel for a static record of what happened when it happened. For example, you could create a new evaluation run where you just copy the old run's outputs for the metrics you've already computed and compute new ones where you want.
> 
> I would personally have no problem just generating a new trace for the updated evaluation results (even if that meant copying execution data from an older trace or otherwise referencing it via span links or something), I don't need to extend an old trace. Maybe others feel differently but just sharing my opinion.
> 
> I'll note that the idea of adding additional metrics also feels somewhat awkward because, I can always add new metrics to an old trace, but what about redefining an existing metric? Presumably that opens more of a can of worms about what it means to "overwrite" old spans? I personally feel it's better to avoid the whole issue by not encouraging this pattern. Just my 2c

The proposal in this PR is consistent with your feedback on traces as static/immutable. Proposal says keep the `span being evaluated` untouched but add a link in `evaluation span` which helps identify span that is being evaluated. Example below shows it. This has flexibility to add more evals as needed.


<table>
  <tr>
    <th>Span Being Evaluated</th>
    <th>Evaluation Span</th>
  </tr>
  <tr>
    <td>
      <pre>
     {
    "name": "chat gpt-4o",
    "context": {
        "trace_id": "0xeb0fdf5670975fea194b2eef13e789c6",
        "span_id": "0x63e929946253cf52",
        "trace_state": "[]"
    },
    "kind": "SpanKind.CLIENT",
    "parent_id": null,
    "start_time": "2025-08-04T23:45:55.959766Z",
    "end_time": "2025-08-04T23:45:57.512342Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "gen_ai.operation.name": "chat",
        "gen_ai.system": "openai",
        "gen_ai.request.model": "gpt-4o",
        "server.address": "anksing1rpeastus2.openai.azure.com",
        "_MS.sampleRate": 100.0,
        "gen_ai.response.model": "gpt-4o-2024-11-20",
        "gen_ai.response.finish_reasons": [
            "stop"
        ],
        "gen_ai.response.id": "chatcmpl-C0zAK9AqI5MpwVScha6p1Dm5BgO7R",
        "gen_ai.usage.input_tokens": 15,
        "gen_ai.usage.output_tokens": 114
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "telemetry.sdk.language": "python",
            "telemetry.sdk.name": "opentelemetry",
            "telemetry.sdk.version": "1.36.0",
            "service.name": "unknown_service"
        },
        "schema_url": ""
    }
}
      </pre>
    </td>
    <td>
      <pre>
      {
    "name": "evaluation relevance",
    "context": {
        "trace_id": "0x6ebb9835f43af1552f2cebb9f5165e39",
        "span_id": "0x89829115c2128845",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": null,
    "start_time": "2025-08-04T23:48:35.592833Z",
    "end_time": "2025-08-04T23:48:35.592833Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "_MS.sampleRate": 100.0,
        "gen_ai.operation.name": "evaluation",
        "gen_ai.evaluation.name": "relevance",
        "gen_ai.evaluation.score": 4,
        "gen_ai.evaluation.label": "Pass",
        "gen_ai.evaluation.reasoning": "Response is relevant to the query."
    },
    "events": [],
    "links": [ // Added Links
        {
            "context": {
                "trace_id": "0xeb0fdf5670975fea194b2eef13e789c6",
                "span_id": "0x63e929946253cf52",
                "trace_state": "[]"
            },
            "attributes": {
                "gen_ai.operation.name": "evaluation"
            }
        }
    ],
    "resource": {
        "attributes": {
            "telemetry.sdk.language": "python",
            "telemetry.sdk.name": "opentelemetry",
            "telemetry.sdk.version": "1.36.0",
            "service.name": "unknown_service"
        },
        "schema_url": ""
    }
}
      </pre>
    </td>
  </tr>
</table>

> what about redefining an existing metric?
- Would this be a good candidate for metric versioning? For version 2 there can a new evaluation span that links to `span being evaluated`?

Appreciate you bring up the scenarios and use cases. It will help shape this work. :)

docs/gen-ai/gen-ai-spans.md
singankit: Thanks @lmolkova for sharing the options.

Options 3 seems most flexible and covers the requirements for both UI (scores along traces + easy to query for dashboarding). It also provided the flexibility to be able to capture the traces for evaluation (probably done during evaluation development phase, optional) if needed without enforcing to create a span. 
I have put together this [document](https://docs.google.com/document/d/1TDuqfULZYY31HHylspeecj0LN_-KCOrw3hz2XKXPekc/edit?usp=sharing) which compares some of the AI Platforms that and how/what is captured as part of evaluation. Hope this can help make a decision. I can add more data as needed, please share any feedback.

Cirilla-zmh: Thank you for this PR!

I'd like to share some practices from Alibaba that might be helpful:

- Evaluation results can vary significantly, so we don't report them as metrics.
  - Besides the evaluation categories listed in your [document](https://docs.google.com/document/d/1TDuqfULZYY31HHylspeecj0LN_-KCOrw3hz2XKXPekc/edit?usp=sharing), we also evaluate other aspects such as 'intention'. Evaluation results can sometimes be just a few descriptive words.
- Spans are typically immutable once stored, so we can't directly insert evaluation results into the LLM span.
- Including evaluation process spans may be important for online evaluations, but we don't believe it's necessary for UX improvements.
  - Actually, we provide post-evaluations after spans are saved. Evaluation events are then created and linked to the corresponding spans via `trace_id` and `span_id`. In this workflow, evaluation process spans don't seem to provide significant value.

alexmojaki: I say option 3. The eval results are more important, and an event works well there, especially when it's a human annotation and there is no process to make a span for. The event can be a child of the thing being evaluated so that its location is consistent while allowing the eval process span to be optional. It can also have attributes pointing to the eval span to make that correlation easier. And the span can have attributes pointing back to the eval score events, or even just have copies of the scores on it directly in a more complex format.

singankit: As per discussion in SIG on 8/12 option 3 is best solution so far. However there are some details to be figured out on how span will be created in addition to events. To continue that discussion creating this [issue](https://github.com/open-telemetry/semantic-conventions/issues/2626). In this PR as per the aggrement will go ahead with Evaluation Results as span events on span being evaluated.

model/gen-ai/registry.yaml
lmolkova: The Assistant API thread matches otel's `gen_ai.conversation.id` semantically. The only difference is the terminology and we can document (separately from this PR) that in OpenAI Assistant case, thread id is captured under `gen_ai.conversation.id` attribute. 

Attributes that are specific to the model provider or framework would look like this https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/openai.md and if we wanted to capture them here they'd go to the top-level attributes, not property bags under other attributes.

model/gen-ai/registry.yaml
lmolkova: > would be the best attribute to capture information like evaluator used to produce evaluation score ?

My understanding is the `gen_ai.evaluation.name` identifies the evaluator. Isn't it the case?

We'd also capture evaluator info in instrumentation scope - the scope name matches evaluator class or artifact name, evaluator version is recorded in the scope version.

The key point - let's try to define things we need we know for sure as top-level attributes. If we need property bag, let's discuss why we need it.

model/gen-ai/registry.yaml
lmolkova: I don't have a strong opinion, let's keep it as string for now

model/gen-ai/registry.yaml
lmolkova: So how would you feel about `gen_ai.evaluation.score.number | value` (instead of `gen_ai.evaluation.score`) and `gen_ai.evaluation.score.description | text`?

this way it's clear that both describe evaluation score, but in different ways? 

`label` is very ambiguous and being in `gen_ai.evaluation` namespace can be property of anything including evaluator itself.

model/gen-ai/registry.yaml
singankit: Thanks will mark it as resolved.

model/gen-ai/registry.yaml
singankit: The challenge I face with `gen_ai.evaluation.score.number` and `gen_ai.evaluation.score.description` is that `gen_ai.evaluation.score.description` feels like the description of the score (this is my interpretation) as you would put in the [rubric ](https://github.com/Azure/azure-sdk-for-python/blob/78693776b7bdd953d08f63bb7fb63787f230f7e9/sdk/evaluation/azure-ai-evaluation/azure/ai/evaluation/_evaluators/_relevance/relevance.prompty#L60)while defining a metric especially for llm-as-a-judge case. However, that is not what I want this property to capture. 

Can this a possible third option?

- Drop `gen_ai.evaluation.label`
- `gen_ai.evaluation.score` ---> `gen_ai.evaluation.value` with type `Any`

model/gen-ai/registry.yaml
singankit: > My understanding is the `gen_ai.evaluation.name` identifies the evaluator. Isn't it the case?

Most of the time yes, but not always. For example, I have a evaluator that measures `violence` called `ViolenceEvaluator`. It takes in content and tells me if it is violent or non-violent on scale of `1-5`. I want to run this evaluator on `input` that comes to my gen-ai App AND on output of the gen-ai app. So I call them `input_violence` and `output_violence` though I use the same evaluator for both.


> We'd also capture evaluator info in instrumentation scope - the scope name matches evaluator class or artifact name, evaluator version is recorded in the scope version.

Is this the [instrumentation scope](https://github.com/open-telemetry/semantic-conventions/blob/6649e7ac3268040831212402fe52ab1702b883f6/model/otel/registry.yaml#L59) ? If this can cover the evaluator name and version that should suffice. Then I do not see a need for a metadata attribute as of now. I can remove it.

model/gen-ai/registry.yaml
singankit: Thanks for the feedback, sounds like a better approach. However, I will wait to make this change in Microsoft specific markdown file after this PR.

docs/gen-ai/gen-ai-events.md
```diff
@@ -519,6 +520,120 @@ Semantic conventions for individual systems MAY specify a different type for arg
...
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `gen_ai.evaluation.result`.
+
+This event captures the process and the result of evaluating GenAI output for quality, accuracy, or other characteristics.
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`gen_ai.evaluation.name`](/docs/registry/attributes/gen-ai.md) | string | The name of the evaluation used for the GenAI response. | `Relevance`; `IntentResolution` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.operation.name`](/docs/registry/attributes/gen-ai.md) | string | The name of the operation being performed. [1] | `chat`; `generate_content`; `text_completion` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
```
alexmojaki: This is about the operation being *evaluated*, right? So it's expected to be the same on the original span?

docs/gen-ai/gen-ai-events.md
```diff
@@ -519,6 +520,120 @@ Semantic conventions for individual systems MAY specify a different type for arg
...
+| [`error.type`](/docs/registry/attributes/error.md) | string | Describes a class of error the operation ended with. [2] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` if the operation ended in an error | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.evaluation.score`](/docs/registry/attributes/gen-ai.md) | double | The evaluation score returned by the evaluator. | `4.0` | `Conditionally Required` if evaluation completed successfully | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.evaluation.label`](/docs/registry/attributes/gen-ai.md) | string | Human readable label for evaluation. | `relevant`; `not_relevant`; `correct`; `incorrect`; `pass`; `fail` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
alexmojaki: Why is one conditionally required and the other recommended? are these not very similar?

docs/gen-ai/gen-ai-events.md
```diff
@@ -519,6 +520,120 @@ Semantic conventions for individual systems MAY specify a different type for arg
...
+| [`gen_ai.evaluation.reasoning`](/docs/registry/attributes/gen-ai.md) | string | A free-form reasoning for the assigned score provided by the evaluator. | `The response is factually accurate but lacks sufficient detail to fully address the question.` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
alexmojaki: This is likely to not be present - is 'recommended' right in that case?

docs/gen-ai/gen-ai-events.md
```diff
@@ -519,6 +520,120 @@ Semantic conventions for individual systems MAY specify a different type for arg
...
+| [`gen_ai.metadata`](/docs/registry/attributes/gen-ai.md) | string | Metadata associated with Gen AI operation. [3] | `{\"evaluator_version\": \"1.2.0\"}` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.provider.name`](/docs/registry/attributes/gen-ai.md) | string | The Generative AI provider as identified by the client or server instrumentation. [4] | `openai`; `gcp.gen_ai`; `gcp.vertex_ai` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.usage.input_tokens`](/docs/registry/attributes/gen-ai.md) | int | The number of tokens used in the GenAI input (prompt). [5] | `100` | `Recommended` if evaluation was performed by a model | ![Development](https://img.shields.io/badge/-development-blue) |
```
alexmojaki: it feels a bit weird that IIUC gen_ai.operation.name and gen_ai.provider.name are copies of values from the evaluated span but these usage attributes which would also be present on the evaluated span are not.

docs/gen-ai/gen-ai-events.md
```diff
@@ -519,6 +520,120 @@ Semantic conventions for individual systems MAY specify a different type for arg
...
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `gen_ai.evaluation.result`.
+
+This event captures the process and the result of evaluating GenAI output for quality, accuracy, or other characteristics.
```
alexmojaki: Is it the process, or just the result?

model/gen-ai/registry.yaml
alexmojaki: I like the third option, have a single consistent attribute name, and it also supports booleans.

docs/gen-ai/gen-ai-events.md
```diff
@@ -519,6 +520,120 @@ Semantic conventions for individual systems MAY specify a different type for arg
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Event: `gen_ai.evaluation.result`
```
alexmojaki: Is the event meant to be a child of an evaluated span? If so I think that should be mentioned.

model/gen-ai/registry.yaml
singankit: Sharing this [document](https://docs.google.com/document/d/1TDuqfULZYY31HHylspeecj0LN_-KCOrw3hz2XKXPekc/edit?tab=t.0#heading=h.7c094agw3rnn) which shows some competition and how evaluation results look like. cc @lmolkova

singankit: As per SIG 8/19 discussion will update `gen_ai.evaluation.score` ---> `gen_ai.evaluation.score.value` and `gen_ai.evaluation.label` ----> `gen_ai.evaluation.score.label`. This approach gives room to expand in future to something   `score.category` where number possibly does not make sense.

model/gen-ai/registry.yaml
singankit: Removed the metadata attribute since evaluator_id can be captured via instrumentation scope and thread_id can be mapped to conversation.id in MS specific attributes.

model/gen-ai/registry.yaml
singankit: > As per SIG 8/19 discussion will update `gen_ai.evaluation.score` ---> `gen_ai.evaluation.score.value` and `gen_ai.evaluation.label` ----> `gen_ai.evaluation.score.label`. This approach gives room to expand in future to something `score.category` where number possibly does not make sense.

PR is updated to reflect this change.

docs/gen-ai/gen-ai-events.md
singankit: To maintain consistency with existing events like gen_ai.assistant.message, which do not explicitly indicate their parent (e.g., chat completion), I’ve followed the same pattern. However, I’m open to incorporating that detail if it improves clarity or aligns better with the overall structure. Happy to adjust.

docs/gen-ai/gen-ai-events.md
singankit: To maintain consistency with existing events like gen_ai.assistant.message, which do not explicitly indicate their parent (e.g., chat completion), I’ve followed the same pattern. However, I’m open to incorporating that detail if it improves clarity or aligns better with the overall structure. Happy to adjust.

docs/gen-ai/gen-ai-events.md
alexmojaki: it's not mentioned much in those specs, you're right, it probably should have been although those events are being removed. but if you search https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/ and https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/ for 'parent', there are mentions.

docs/gen-ai/gen-ai-events.md
alexmojaki: is this reply meant to be here?

singankit: My bad. That was for a different comment. I have removed `gen_ai.operation.name` from the event this came here by mistake when moving things around from span.yaml

docs/gen-ai/gen-ai-events.md
singankit: Have updated the description to reflect is should be parented to GenAI operation span. Please review and share feedback.

docs/gen-ai/gen-ai-events.md
singankit: Addressed

docs/gen-ai/gen-ai-events.md
singankit: operation.name got added here by mistake have removed it.

docs/gen-ai/gen-ai-events.md
singankit: Idea was label can or cannot be provided. Hence recommended. However, with score.value and score.label I have made them conditionally required if applicable

docs/gen-ai/gen-ai-events.md
singankit: Addressed.

model/gen-ai/registry.yaml
```diff
@@ -332,3 +332,23 @@ groups:
...
+        stability: development
+        type: double
+        brief: The evaluation score returned by the evaluator.
+        examples: [4.0]
+      - id: gen_ai.evaluation.score.label
+        stability: development
+        type: string
+        brief: Human readable label for evaluation.
+        examples: ["relevant", "not_relevant", "correct", "incorrect", "pass", "fail"]
+      - id: gen_ai.evaluation.reasoning
```
aabmass: nit: reasoning could be confused with reasoning LLM? I don't want to bike shed the names but wdyt of "explanation" which Arize uses?

aabmass (commented):
I don't have a strong background in evals so would prefer to see some other approvals before ✅ this, but LGTM.

model/gen-ai/events.yaml
```diff
@@ -313,3 +313,37 @@ groups:
                     arguments field.
                   examples: ['{\"location\": \"Paris\"}']
                   requirement_level: opt_in
+
+  - id: event.gen_ai.evaluation.result
+    name: gen_ai.evaluation.result
+    type: event
+    stability: development
+    brief: >
+      This event captures the result of evaluating GenAI output for quality, accuracy, or other characteristics and are parented to GenAI operation span being evaluated.
```
aabmass: It's kind of implicit, but it might be useful to explain the meaning of this event's timestamp. I imagine it would be when evaluation finished, which could be way after the span's end time.

model/gen-ai/registry.yaml
singankit: I like this suggestion. However after thinking about it I can sway either way. Difference I see is that for non-reasoning model user would need to have explicit instructions to give chain-of-thoughts but for reasoning model it would do it automatically in a way.

model/gen-ai/events.yaml
singankit: Added content to description to reflect it. Please review and share your feedback.

model/gen-ai/registry.yaml
singankit: Updated to explanation.

docs/gen-ai/gen-ai-events.md
singankit: Usage attributes on the event will be for the evaluation performed, which is yes different from usage attributes on span being evaluated. Span being evaluated can/cannot be using LLM and depending on it attributes like token usage may or may not exist on span being evaluated.

model/gen-ai/registry.yaml
```diff
@@ -474,3 +474,23 @@ groups:
                   "finish_reason": "stop"
                 }
               ]
+      - id: gen_ai.evaluation.name
+        stability: development
+        type: string
+        brief: The name of the evaluation used for the GenAI response.
```
lmolkova: Can we be more specific?

```suggestion
        brief: The name of the evaluation metric used for the GenAI response.
```

metric | method | etc

model/gen-ai/registry.yaml
```diff
@@ -474,3 +474,23 @@ groups:
...
+        examples: ["Relevance", "IntentResolution"]
+      - id: gen_ai.evaluation.score.value
+        stability: development
+        type: double
+        brief: The evaluation score returned by the evaluator.
+        examples: [4.0]
+      - id: gen_ai.evaluation.score.label
+        stability: development
+        type: string
+        brief: Human readable label for evaluation.
```
lmolkova, line 495: could you please add a note and explain what the label is? 

E.g.

```suggestion
        brief: Human readable label for evaluation.
        note: > 
          This attribute represents an interpretation of evaluation score when its applicable. For example, score value of 1 could mean that content is relevant when using one evaluator which measures score in range [-1, 1] and not relevant for another evaluator which measures it in the range [0, 5].
          This attribute SHOULD have low cardinality. Possible values depend on the evaluation metric and evaluator used. Implementations that report this attribute SHOULD document possible values.
```

I used my imagination and poor English to write this note, so please correct it and adjust to the real-life needs.

model/gen-ai/events.yaml
```diff
@@ -10,3 +10,38 @@ groups:
       This event is opt-in and could be used to store input and output details
       independently from traces.
     extends: attributes.gen_ai.inference.client
+
+  - id: event.gen_ai.evaluation.result
+    name: gen_ai.evaluation.result
+    type: event
+    stability: development
+    brief: >
+      This event captures the result of evaluating GenAI output for quality, accuracy, or other characteristics and are parented to GenAI operation span being evaluated. This event can be created after the GenAI operation span being evaluated has completed.
```
lmolkova:
```suggestion
      This event captures the result of evaluating GenAI output for quality, accuracy, or other characteristics and are parented to GenAI operation span being evaluated. 
```

Is this sentence useful? Someone could report eval event without ever reporting GenAI spans.

lmolkova: where did we end up on the relationship discussion ?

I.e. do we expect this event to be a child of GenAI span? What to do when it's not feasible?

I think it's ok to postpone this discussion, but wanted to make sure that's what we decided.

One thing that would be useful is to also reference `gen_ai.response.id` attribute (with level `recommended: when available`) so that some correlation mechanism exists right away regardless of trace-context.

singankit: > where did we end up on the relationship discussion ?
> 
> I.e. do we expect this event to be a child of GenAI span? What to do when it's not feasible?
> 
> I think it's ok to postpone this discussion, but wanted to make sure that's what we decided.

This is to be discussed as part of this [issue](https://github.com/open-telemetry/semantic-conventions/issues/2626).


> One thing that would be useful is to also reference `gen_ai.response.id` attribute (with level `recommended: when available`) so that some correlation mechanism exists right away regardless of trace-context.

I shall add this attribute.

singankit: Have added the response.id attribute

singankit: Addressed it. Thanks for feedback.

model/gen-ai/registry.yaml
singankit: Thanks for the write up it helped a lot. Have added it to the note section

model/gen-ai/registry.yaml
singankit: Went with metric

model/gen-ai/events.yaml
singankit: Have updated the issue to include this question/ask.

docs/gen-ai/gen-ai-events.md
```diff
@@ -209,4 +210,53 @@ section for more details.
...
+| [`error.type`](/docs/registry/attributes/error.md) | string | Describes a class of error the operation ended with. [1] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` if the operation ended in an error | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.evaluation.score.label`](/docs/registry/attributes/gen-ai.md) | string | Human readable label for evaluation. [2] | `relevant`; `not_relevant`; `correct`; `incorrect`; `pass`; `fail` | `Conditionally Required` if gen_ai.evaluation.score.value is not provided | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.evaluation.score.value`](/docs/registry/attributes/gen-ai.md) | double | The evaluation score returned by the evaluator. | `4.0` | `Conditionally Required` if gen_ai.evaluation.score.label is not provided | ![Development](https://img.shields.io/badge/-development-blue) |
```
zhirafovod: if the value is not in [0,1] range - probably we need to provide `genai.evaluation.score.range`, i.e. 
`gen_ai.evaluation.score.range="[0,5]"`
(inspired by https://platform.openai.com/docs/guides/graders#model-graders)

docs/gen-ai/gen-ai-events.md
```diff
@@ -209,4 +210,53 @@ section for more details.
...
+the canonical name of exception that occurred, or another low-cardinality error identifier.
+Instrumentations SHOULD document the list of errors they report.
+
+**[2] `gen_ai.evaluation.score.label`:** This attribute provides a human-readable interpretation of the evaluation score produced by an evaluator. For example, a score value of 1 could mean "relevant" in one evaluation system and "not relevant" in another,depending on the scoring range and evaluator. The label SHOULD have low cardinality and be easily understood by users. Possible values depend on the evaluation metric and evaluator used; implementations SHOULD document the possible values.
+
+**[3] `gen_ai.usage.input_tokens`:** The total number of input tokens used by the model during the evaluation.
+
+**[4] `gen_ai.usage.output_tokens`:** The total number of output tokens used by the model during the evaluation.
+
+---
```
zhirafovod, line 247: @singankit Can we add optional attributes `gen_ai.input.messages`, `gen_ai.output.messages` and `gen_ai.system_instructions`  to the event? And also to support multiple evaluations on the same event? 
The rationale is to have the data and the evaluations on the same data frame, so it can be easily put into idempotent table and ease searching by this data on the backend.

docs/gen-ai/gen-ai-events.md
```diff
@@ -209,4 +210,53 @@ section for more details.
...
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-capture -->
+<!-- markdownlint-disable -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `gen_ai.evaluation.result`.
+
+This event captures the result of evaluating GenAI output for quality, accuracy,or other characteristics and are parented to GenAI operation span being evaluated.
```
lmolkova:
```suggestion
This event captures the result of evaluating GenAI output for quality, accuracy, or other characteristics and SHOULD be parented to GenAI operation span being evaluated when possible.
```

It's not possible when doing sync evaluations

docs/gen-ai/gen-ai-events.md
lmolkova: I like the idea, but would suggest to do it incrementally as a follow up.

docs/gen-ai/gen-ai-events.md
lmolkova: can it be added as a follow-up? It's questionable on what input and output messages represent - should it include the evaluation prompt when LLM used as a judge?

model/gen-ai/events.yaml
```diff
@@ -10,3 +10,42 @@ groups:
...
+    stability: development
+    brief: >
+      This event captures the result of evaluating GenAI output for quality, accuracy,or other characteristics
+      and are parented to GenAI operation span being evaluated.
+    attributes:
+      - ref: gen_ai.evaluation.name
+        requirement_level: required
+      - ref: gen_ai.evaluation.score.value
+        requirement_level:
+          conditionally_required: if gen_ai.evaluation.score.label is not provided
```
lmolkova:
```suggestion
          conditionally_required: if applicable
```

I don't think these are mutually exclusive, `if gen_ai.evaluation.score.label is not provided` creates  this impression

model/gen-ai/events.yaml
```diff
@@ -10,3 +10,42 @@ groups:
...
+      - ref: gen_ai.evaluation.score.label
+        requirement_level:
+          conditionally_required: if gen_ai.evaluation.score.value is not provided
```
lmolkova:
```suggestion
          conditionally_required: if applicable
```

model/gen-ai/registry.yaml
```diff
@@ -474,3 +474,28 @@ groups:
...
+        examples: ["relevant", "not_relevant", "correct", "incorrect", "pass", "fail"]
+        note: >
+          This attribute provides a human-readable interpretation of the evaluation score produced by an evaluator.
+          For example, a score value of 1 could mean "relevant" in one evaluation system and "not relevant" in another,depending on the scoring range and evaluator.
```
```suggestion
          For example, a score value of 1 could mean "relevant" in one evaluation system and "not relevant" in another, depending on the scoring range and evaluator.
```

docs/gen-ai/gen-ai-events.md
singankit: Thanks @zhirafovod. Can this be discussed as part of this [issue](https://github.com/open-telemetry/semantic-conventions/issues/2626) ? Sounds like a good candidate for cover as part of this issue.

model/gen-ai/registry.yaml
```diff
@@ -474,3 +474,28 @@ groups:
...
+          The label SHOULD have low cardinality and be easily understood by users.
```
lmolkova: I think human-readable in the brief covers it. "easily understood" is not really adding much, wdyt? 

```suggestion
          The label SHOULD have low cardinality.
```

docs/gen-ai/gen-ai-events.md
singankit: Thanks @zhirafovod for the feedback. Created a new issue[](https://github.com/open-telemetry/semantic-conventions/issues/2670) to capture this feedback. Would appreciate to get your opinion on it and get the conversation started.

lmolkova (approved):

docs/gen-ai/gen-ai-events.md
```diff
@@ -209,4 +210,53 @@ section for more details.
...
+| [`gen_ai.evaluation.score.label`](/docs/registry/attributes/gen-ai.md) | string | Human readable label for evaluation. [2] | `relevant`; `not_relevant`; `correct`; `incorrect`; `pass`; `fail` | `Conditionally Required` if applicable | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.evaluation.score.value`](/docs/registry/attributes/gen-ai.md) | double | The evaluation score returned by the evaluator. | `4.0` | `Conditionally Required` if applicable | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.evaluation.explanation`](/docs/registry/attributes/gen-ai.md) | string | A free-form explanation for the assigned score provided by the evaluator. | `The response is factually accurate but lacks sufficient detail to fully address the question.` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
srtux, line 234: Could there be issues where the explanation is too long to fit in an attribute? Is there an option to use messages_ref similar to #2179 ?

docs/gen-ai/gen-ai-events.md
alexmojaki: OK but I still find it weird to put `gen_ai.usage.input_tokens` here at all. It should be on the span describing the evaluation process. Same for `gen_ai.response.id`.

docs/gen-ai/gen-ai-events.md
alexmojaki: I think that could be a followup, but I would be very surprised if the explanation was that long.

docs/gen-ai/gen-ai-events.md
singankit: Evaluation can produce multiple evaluation metrics; this will result in one evaluation span but multiple evaluation result for each metric. This makes it easier to put on the evaluation result event, rather than on evaluation process span since latter can require an attribute to capture tokens for each metric

alexmojaki: OK, that was a mistake. What I was thinking is just that if you use an LLM to evaluate, then there will inherently often be an LLM request span which will have this info. Not that this usage should be recorded on the outer generic 'evaluation process' span.

If a single LLM request is used to produce multiple scores and thus multiple events, it'd be weird for the same usage to be recorded multiple times on different events. Adding up all the usage attributes across some set of evaluation events won't be meaningful.

singankit: > If a single LLM request is used to produce multiple scores and thus multiple events, it'd be weird for the same usage to be recorded multiple times on different events. Adding up all the usage attributes across some set of evaluation events won't be meaningful.

Agree if one LLM call produces multiple scores then it is not possible to break down token usage for each score. In that case it will be misleading to report same token score in each evaluation result event. (In my opinion we should not do this). For cases where one service call calculates multiple scores, and service gives info about token usage for each if applicable. This info can be put on each evaluation event, and aggregated (sum) info can be put on evaluation process span if needed.

> What I was thinking is just that if you use an LLM to evaluate, then there will inherently often be an LLM request span which will have this info. Not that this usage should be recorded on the outer generic 'evaluation process' span.

This is definitely true if user controls both. Putting a table below for good comparison and how having it on event help most of those scenarios

| Scenario | Description | Token Data on Event | Token Data on Span(Per Score) |
|----------|-------------|--------------------|--------------------|
| **One LLM Call, One Metric** | Single LLM API to calculate evaluator score| Y | Y |
| <br>**One LLM Call, Multiple Metrics**</br> <br> **One Service Call, Multiple Metrics**</br> | Single LLM API call to calculate multiple evaluator score | Not possible | Not possible (Only Total token usage) |
| **One Service Call, Multiple Metrics + Token Usage** | Service endpoint that makes multiple LLM calls internally, tracking service-level metrics plus detailed token usage | Y | Not possible (Only total token usage) |

lmolkova: is there a strong scenario behind reporting token usage (caused by evaluation) on the evaluation result? I can see the logic behind @alexmojaki question - we don't report duration of the evaluation process or parameters used on the result event, so why should we report usage?

usage, along with perf data would be reported on the spans-about-eval-process - this would be consistent.

I don't have a strong objection to reporting it either, but I wonder why is it important.

singankit: Motivation is similar to any other LLM operation ,to know what it costs. But yes I agree there can be other usage data too that potentially a user would want to see. I am open to removing it for now and continue this discussion as part of this [issue](https://github.com/open-telemetry/semantic-conventions/issues/2626) to make progress on this PR @alexmojaki  @lmolkova

zhirafovod:
Looks good to me overall

docs/gen-ai/gen-ai-events.md
singankit: PR is update to remove token usage attribute on evaluation result event. Have updated this [issue](https://github.com/open-telemetry/semantic-conventions/issues/2626) to capture this dicussion.

model/gen-ai/events.yaml
```diff
@@ -10,3 +10,32 @@ groups:
...
+        requirement_level: required
+      - ref: gen_ai.evaluation.score.value
+        requirement_level:
+          conditionally_required: if applicable
+      - ref: gen_ai.evaluation.score.label
+        requirement_level:
+          conditionally_required: if applicable
+      - ref: gen_ai.evaluation.explanation
+        requirement_level: recommended
+      - ref: gen_ai.response.id
```
alexmojaki, line 33: What does this refer to? Is it a copy of the value on the parent span?

nirga (approved):

model/gen-ai/events.yaml
lmolkova: sig discussion: let's update brief and tell:
- it's used as a correlation mechanism when span id is not available
- it's the id of the original completion being evaluated

singankit: Have updated the PR to update note for this attribute.

singankit:
Thank you all for the valuable feedback and thoughtful discussion that helped bring this PR to a merge-ready state.
@lmolkova ,could you please proceed with merging the PR at your convenience? It has received the required approvals.

model/gen-ai/events.yaml
```diff
@@ -10,3 +10,36 @@ groups:
...
+        requirement_level:
+          recommended: when available
+        note: |
+          The `gen_ai.response.id` is the unique identifier assigned to the specific
```
alexmojaki:
```suggestion
          The unique identifier assigned to the specific
```

model/gen-ai/events.yaml
```diff
@@ -10,3 +10,36 @@ groups:
...
       independently from traces.
     extends: attributes.gen_ai.inference.client
+
+  - id: event.gen_ai.evaluation.result
+    name: gen_ai.evaluation.result
+    type: event
+    stability: development
+    brief: >
+      This event captures the result of evaluating GenAI output for quality, accuracy,or other characteristics
+      and SHOULD be parented to GenAI operation span being evaluated when possible.
```
alexmojaki: I think mention `gen_ai.response.id` when this is not possible.

docs/gen-ai/gen-ai-events.md
```diff
@@ -223,7 +223,7 @@ section for more details.
 
 The event name MUST be `gen_ai.evaluation.result`.
 
-This event captures the result of evaluating GenAI output for quality, accuracy,or other characteristics and SHOULD be parented to GenAI operation span being evaluated when possible.
+This event captures the result of evaluating GenAI output for quality, accuracy,or other characteristics. This event SHOULD be parented to GenAI operation span being evaluated when possible  or set `gen_ai.response.id` when span id is not available.
```
alexmojaki:
```suggestion
This event captures the result of evaluating GenAI output for quality, accuracy, or other characteristics. This event SHOULD be parented to GenAI operation span being evaluated when possible  or set `gen_ai.response.id` when span id is not available.
```

model/gen-ai/events.yaml
singankit: Thanks for the feedback have added it. Please share your review.

model/gen-ai/events.yaml
```diff
@@ -33,9 +34,9 @@ groups:
         requirement_level:
           recommended: when available
         note: |
-          The `gen_ai.response.id` is the unique identifier assigned to the specific
+          The unique identifier assigned to the specific
           completion being evaluated. This attribute helps correlate the evaluation
-          event with the corresponding operation when span id is not available.
```
alexmojaki: why was 'available' removed?

alexmojaki (approved):

model/gen-ai/events.yaml
singankit: Not sure why it got removed (maybe I missed pushing some changes), latest iteration has it. Please review.

docs/gen-ai/gen-ai-events.md
singankit: Have update the description as per the suggestion here.

model/gen-ai/events.yaml
singankit: Updated.

docs/gen-ai/gen-ai-events.md
```diff
@@ -209,4 +210,51 @@ section for more details.
...
+|---|---|---|---|---|---|
+| [`gen_ai.evaluation.name`](/docs/registry/attributes/gen-ai.md) | string | The name of the evaluation metric used for the GenAI response. | `Relevance`; `IntentResolution` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`error.type`](/docs/registry/attributes/error.md) | string | Describes a class of error the operation ended with. [1] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` if the operation ended in an error | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.evaluation.score.label`](/docs/registry/attributes/gen-ai.md) | string | Human readable label for evaluation. [2] | `relevant`; `not_relevant`; `correct`; `incorrect`; `pass`; `fail` | `Conditionally Required` if applicable | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`gen_ai.evaluation.score.value`](/docs/registry/attributes/gen-ai.md) | double | The evaluation score returned by the evaluator. | `4.0` | `Conditionally Required` if applicable | ![Development](https://img.shields.io/badge/-development-blue) |
```
srtux, line 233: Can we also include a  confidence interval value along with the eval score double value?

For example a score of `0.399` with a CI of `+/- 0.032`

singankit: @srtux Is the suggestion to have another attribute that captures confidence interval ?

singankit: Can this be added as a follow-up? I can update this [issue](https://github.com/open-telemetry/semantic-conventions/issues/2626) to capture it.

alexmojaki: I agree with making this a followup, no reason for it to block this PR. But that issue about spans isn't related.

srtux: Fine to add as a follow-up

singankit: Created a new [issue ](https://github.com/open-telemetry/semantic-conventions/issues/2687)to capture this feedback and have a follow up.

PR #2564, opened 2025-07-26
Introduce standalone in-memory db span def #2467

thompson-tomo:
Progresses open-telemetry/semantic-conventions-genai#31

## Changes
Introduce a dedicated span definition for in-memory db with the internal kind set. This removes need for notes on the general span.

There has been no change in attributes except for the new in-memory db span not having server attributes due to not being relevant.

## Reason for the change
This is a blocker for moving to code generated span registries as a span can only have 1 type, hence should be separate definitions. At the same time is a blocker for code generation as a system cannot process the split nature.


thompson-tomo:
Awaiting feedback after changes based on last changes to provide differences.

thompson-tomo:
Awaiting feedback given in-memory was already implied in the docs, this makes it explicit.

model/database/spans.yaml
```diff
@@ -111,6 +127,9 @@ groups:
     stability: stable
     brief: This span describes database client call.
     note: |
+      This span is for db calls where the db runs in a different process to the client or when the database calls happen over an
+      instrumented protocol such as HTTP. If the db call is occurring in-memory then the [db.inmemory.internal](/docs/database/inmemory.md) span should be used.
```
KalleOlaviNiemitalo, line 131: This is a bit confusing because `inmemory.md` mentions the `db.inmemory.internal` name only in an HTML comment so the reader is left wondering whether the instrumentation should place the `db.inmemory.internal` name in an attribute or something.

docs/database/inmemory.md
```diff
@@ -0,0 +1,225 @@
...
+
+# Semantic conventions for database client spans
+
+**Status**: [Stable][DocumentStatus], Unless otherwise specified.
+
+## Span definition
+
+### In-memory Process
+
+It is expected that `db.system.name` is set to inmemory.
```
KalleOlaviNiemitalo, line 13: SQLite runs in the same process and it already has `sqlite` as `db.system.name`.

Is this `inmemory` intended only for databases that keep their data completely in memory and not in any disk file?

(I'm commenting on the Markdown file because I did not find this text in any YAML file.)

model/database/spans.yaml
thompson-tomo: Good point will update it.

docs/database/inmemory.md
thompson-tomo: Correct intended for pure in-memory providers.

Might be worthwhile to also add a sqlite page.

PR #2567, opened 2025-07-29
Add in SNS as messaging system #2266

thompson-tomo:
Progresses #2266 

## Changes

### DIFF
--- a/docs/messaging/messaging-metrics.md
+++ b/docs/messaging/messaging-metrics.md
@@ -74,7 +74,7 @@ of `[ 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
 | [`messaging.operation.name`](/docs/registry/attributes/messaging.md) | string | The system-specific name of the messaging operation. | `send`; `receive`; `ack` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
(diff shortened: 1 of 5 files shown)


joaopgrassi:
I have not used SNS, so I'm not any close to being an expert. @thompson-tomo have you used it? Did you review the current messaging span/trace structure conventions? Does it apply to SNS well enough that it can be referred in the list of systems? Is there instrumentations/prototypes already for it following these conventions? 

All of these questions need to be answered before we add the new system here I'm afraid. We want to be more strict of what we add, since we will have to maintain it later on.

thompson-tomo:
@joaopgrassi yes there is already instrumentations. In fact there is already a page for it https://opentelemetry.io/docs/specs/semconv/messaging/sns/ and on that page it lists sns specific attributes for sns.

What was missing was it being listed as a system.

joaopgrassi (approved):

joaopgrassi:
ah right, completely missed that. Not sure why was this marked as blocked then?

docs/messaging/messaging-metrics.md
```diff
@@ -450,6 +453,7 @@ the broker doesn't have such notion, the destination name SHOULD uniquely identi
 | Value  | Description | Stability |
 |---|---|---|
 | `activemq` | Apache ActiveMQ | ![Development](https://img.shields.io/badge/-development-blue) |
+| `aws.sns` | Amazon Simple Notification Service (SNS) | ![Development](https://img.shields.io/badge/-development-blue) |
```
jsuereth, line 456: This really sticks out with the other `_` now.

@lmolkova / @joaopgrassi Is the plan to move these all to be `.` based in the future?  Is there a tracking ticket for that?

joaopgrassi: I created one https://github.com/open-telemetry/semantic-conventions/issues/2668

PR #2570, opened 2025-07-29
Add in ArrangoDB #1107

thompson-tomo:
Progresses #1107

## Changes

This adds ArrangoDB as a db system, so that it can be used in telemetry.

## Usage Example

- Health Check of the connection to the db https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2228. The definition of a health check event is much like an exception/event and as mentioned in the semconv meeting, 28/7/25, if attributes are not listed they are implicitly opt-in. As such this is just like if you were to put db.system.name on an exception it is beneficial. In some health checks it is even possible to check the health of db's not just sever connection etc hence even more reason to leverage the db namespace to capture that data.

### DIFF
--- a/.chloggen/add_arrango.yaml
+++ b/.chloggen/add_arrango.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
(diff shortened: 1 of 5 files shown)


thompson-tomo:
Still Awaiting for review.

joaopgrassi:
@open-telemetry/semconv-db-approvers can you please take a look?

thompson-tomo:
awaiting feedback from @open-telemetry/semconv-db-approvers

PR #2575, opened 2025-07-31
Add instance entity

thompson-tomo:
## Changes

Enable instance to be able to be an entity which required rename.

### DIFF
--- a/.chloggen/add_instance_entity.yaml
+++ b/.chloggen/add_instance_entity.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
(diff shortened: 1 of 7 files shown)


thompson-tomo:
@tylerbenson thanks for feedback, see https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3087 for the update done to the dotnet implementation. Key thing being scope is naming update rather new attributes.

I would suggest looking to stabilise or atleast moving to beta/rc stability for atleast the function & instance namespaces that implementations can be updated in bulk.

joaopgrassi:
I will close this PR as there's no current SIG active for FaaS and no clear code owners to review/work on a migration path for this.

PR #2577, opened 2025-07-31
[chore] Describe configurable nature of attributes

thompson-tomo:
## Changes

This adds a table summarising the ability to include/exclude an attribute from telemetry based upon requirement level. This table is based upon the paragraphs below the table.

### DIFF
--- a/docs/general/attribute-requirement-level.md
+++ b/docs/general/attribute-requirement-level.md
@@ -7,10 +7,11 @@
 
 <!-- toc -->
 
-- [Required](#required)
-- [Conditionally Required](#conditionally-required)
-- [Recommended](#recommended)
-- [Opt-In](#opt-in)
+- [Requirement Levels](#requirement-levels)
+  - [Required](#required)
+  - [Conditionally Required](#conditionally-required)
+  - [Recommended](#recommended)
+  - [Opt-In](#opt-in)
 - [Performance suggestions](#performance-suggestions)
 
 <!-- tocstop -->
@@ -22,13 +23,6 @@ requirement levels for attributes defined in semantic conventions._
 
 Attribute requirement levels apply to the
 [instrumentation library](https://opentelemetry.io/docs/concepts/glossary/#instrumentation-library).
-
-The following attribute requirement levels are specified:
-
-- [Required](#required)
-- [Conditionally Required](#conditionally-required)
-- [Recommended](#recommended)
-- [Opt-In](#opt-in)
 
 The requirement level for an attribute is specified by semantic conventions
 depending on attribute availability across instrumented entities, performance,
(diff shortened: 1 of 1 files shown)


thompson-tomo:
Awaiting feedback as table provides a summary of the text already published.

docs/general/attribute-requirement-level.md
```diff
@@ -48,7 +42,25 @@ For example, [Database semantic convention](../database/README.md) references
...
+The below table seeks to summarise the behaviour of attributes on signals,
+by providing a quick overview.
+In the scenario where an instrumentation has not implemented an opt-in or recommended attribute,
+the configuration option should either not exist or it should operate as a noop.
+The same is the case for conditionally required attributes where the conditions or instructions are not satisfied.
+
+| Level | Included by Default | Included via Config | Excluded via Config |
+| --- | --- | --- | --- |
+| [Required](#required) | Yes | n/a | No |
+| [Conditionally Required](#conditionally-required) | Yes [^1] | No [^1] | No [^1] |
```
joaopgrassi: Hum I'm not sure this table adds any value to this document. The terms about "excluded/included via config" are only understandable when one reads the full text below.

I personally don't mind to read the more extended text and I don't think they are long that a summary table like this is needed.

I do like the organization of it on a sub heading like you have now though.

thompson-tomo: See one such example of what it is trying to address https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2962

Is your suggestion to replace the foot notes with more verbose content in the cells?

joaopgrassi: I see, so the table is not a summarization of the following text, but an overview of how each level can be configured/not configured?

I think then it makes sense, but I would move the table section after all the text. Then the flow of a reader is:

- I learn what each req level mean
- I learn how they are configured

Does it make sense?

docs/general/attribute-requirement-level.md
```diff
@@ -48,7 +42,25 @@ For example, [Database semantic convention](../database/README.md) references
...
-## Required
+## Requirement Levels
+
+The below table seeks to summarise the behaviour of attributes on signals,
+by providing a quick overview.
+In the scenario where an instrumentation has not implemented an opt-in or recommended attribute,
+the configuration option should either not exist or it should operate as a noop.
+The same is the case for conditionally required attributes where the conditions or instructions are not satisfied.
+
+| Level | Included by Default | Included via Config | Excluded via Config |
```
joaopgrassi:
```suggestion
| Level | Included by default | Can be included via config | Can be disabled via config |
```

Maybe?

joaopgrassi: or since we use Opt-in opt-out, maybe we use that instead of "included/disabled"?

thompson-tomo: So I went with included/excluded terminology to be consistent declarative config see https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml

docs/general/attribute-requirement-level.md
thompson-tomo: So column 2 could be seen as a summary as it is summarising the inclusion behaviour with column 3 & 4 about if it can be changed via Config.

Note the table is replacing a list of the requirement levels.

docs/general/attribute-requirement-level.md
```diff
@@ -48,7 +42,25 @@ For example, [Database semantic convention](../database/README.md) references
...
 `Conditionally Required` level on it.
 
-## Required
+## Requirement Levels
+
+The below table seeks to summarise the behaviour of attributes on signals,
+by providing a quick overview.
+In the scenario where an instrumentation has not implemented an opt-in or recommended attribute,
+the configuration option should either not exist or it should operate as a noop.
+The same is the case for conditionally required attributes where the conditions or instructions are not satisfied.
```
joaopgrassi: I'm not sure this text helps with anything. I feel it carries too much information and by the end of reading I feel users will get more confused. Maybe we can just remove it altogether and simply leave the table as is. 

If not, I'd try to make it shorter and avoid things as "the table below seeks to summarise". It shouldn't seek, it either does or doesn't 😅. If it doesn't then we have a problem and that should be fixed.

thompson-tomo: Sure I have made changes to the text including removing most of the text. Let me know if you are happy with the changes.

Issue #2587, opened 2025-08-01
Define resources/entities for Azure Resource Detector


thompson-tomo:
### Area(s)

area:azure

### What's missing?

Any entries or resources which describe an azure computing instance

### Describe the solution you'd like

Addition of the required attributes to the registry.

1 or more definitions of an azure computing instance which can be used by the resource detectors.

PR #2589, opened 2025-08-03
Add azure computing attributes #2587

thompson-tomo:
Progresses #2587 

## Changes

Adds in 2 attributes for usage by the resource detectors.

Implementation is already done in the java & dotnet contrib project.

Only change which has occurred is app.service has become app_service to follow guidance


thompson-tomo:
@trask as you are a component owner of the java resource detector which use these could you take a look and advise if you have concerns about the renaming of 1 attribute based on conventions.

trask:
cc @zeitlinger

zeitlinger:
@thompson-tomo the java implementation may be changed - because it's not stable yet

thompson-tomo:
Thanks, hopefully we can get these 2 attributes reviewed soon given that they are already implemented in multiple languages.

joaopgrassi:
I'm not against this change, but I have a general question on these: files on `model/azure` don't have a code owner assigned. I feel these should be reviewed by Azure/MS folks as they most likely have the ultimate decision on how to name things. I will mark this as blocked until that is figured it out. We should aim to not have not-owned conventions as much as possible.

thompson-tomo:
So these attributes are used in multiple otel contrib projects and the property names are direct mapping to Ms defined properties. The question is on sub-namespaces and as such how can this be progressed? @joaopgrassi

thompson-tomo:
How can this be progressed given it used by many otel implementations?

joaopgrassi:
I have no authority to say if this is correct or not, have not many experience with azure. To progress, is the usual we have been discussing: Get either azure folks to own/put their reviews, or gather end users with real experience to ensure this makes sense.

lmolkova:
Semconv SIG discussion:
There is currently no formal (or informal) SIG to work on Azure resource detection, so while there is a value in defining attributes, we'd want to see a group of owners who can define Azure resources wholistically and drive them so stability. 

So closing this one, until this group is formed.

PR #2590, opened 2025-08-03
Improve cicd worker description

### LINKED ISSUE #1191: Discuss guidance around runner system attributes in distributed tracing
# Overview

Discuss and add guidance around setting attributes in traces when multiple runners are tied to a single pipeline.

> Note: Removed `...runner.{name, id}` from pull request #1075 which can be added back as part of this issue.

Reference:
* https://github.com/open-telemetry/semantic-conventions/pull/1075#discussion_r1651185234

kamphaus:
Fixes #1191

## Changes

This PR improves the description of `cicd.worker` entities.

* Links to the prototypes or existing instrumentations (when adding or changing conventions) → NA


model/cicd/entities.yaml
```diff
@@ -24,6 +24,12 @@ groups:
     name: cicd.worker
     brief: >
       A CICD worker is a component of the CICD system that performs work (eg. running pipeline tasks or performing sync).
+
+      A single pipeline run may be distributed across multiple workers. Each span representing a pipeline run or a task run
+      should be linked to the worker that performed the corresponding work.
+
+      When a pipeline run involves several workers, its task run spans may reference different `cicd.worker` resources.
+      The pipeline run's parent span may instead reference the CICD controller as the `cicd.worker` resource.
```
lmolkova: Should span relationship be documented on the spans? any reason to document them on entities?

kamphaus: Usually the tasks executing on a worker emit several spans. By setting the worker as an entity it's more efficient than setting the worker attributes on each span.

lmolkova: sorry, it's not about attributes - the paragraph 

>  A single pipeline run may be distributed across multiple workers. Each span representing a pipeline run or a task run  should be linked to the worker that performed the corresponding work.

talk about relationships between CI/CD spans, which should be part of span docs.

Also if by `linked` you mean associated with the entity, please use this wording - links between spans are something different

kamphaus: Thanks for the clarification.

> by linked you mean associated with the entity, please use this wording - links between spans are something different

Agreed, I have changed this as requested: 4cf970096841e55d9deb2c8243a0eb7b52693c8c

> should be part of span docs

While I wrote here about spans I think the clarification for the worker description regarding a pipeline run executing across multiple workers is applicable to more than just spans.
It would also apply to any logs of a pipeline run emitted on each worker or the controller (as well as metrics, cf #1184).
As such I think we should keep this in the worker entities description.

I will have to think about making this description more generic...

lmolkova (approved):

model/cicd/entities.yaml
kamphaus: > I will have to think about making this description more generic

Done: 5d721b52f696793aebae441924a7dd90c71a7d7e and 9cc63f949484015e8e979284c598c9df142b8882

adrielp (approved):

PR #2591, opened 2025-08-04
App build id attribute

LikeTheSalad:
Fixes #

## Changes

### DIFF
--- a/model/app/registry.yaml
+++ b/model/app/registry.yaml
@@ -62,3 +62,8 @@ groups:
           A widget is an application component, typically an on-screen visual
           GUI element.
         examples: [ "submit", "attack", "Clear Cart" ]
+      - id: app.build_id
+        type: string
+        stability: development
+        brief: 'Unique identifier for a particular build or compilation of the application.'
+        examples: ['6cff0a7e-cefc-4668-96f5-1273d8b334d0', '9f2b833506aa6973a92fde9733e6271f', 'my-app-1.0.0-code-123']
--- a/docs/registry/attributes/app.md
+++ b/docs/registry/attributes/app.md
@@ -9,6 +9,7 @@ Describes attributes related to client-side applications (e.g. web apps or mobil
 
 | Attribute | Type | Description | Examples | Stability |
 |---|---|---|---|---|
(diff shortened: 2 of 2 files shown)


thompson-tomo (approved):

bidetofevil (approved):
Looks good. This is useful to distinguish between different pre-release app release that have the same public version build-specific information like deobfuscation mapping files or build configuration.

.chloggen/add_app_build_id.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: app
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Defines a new build_id attribute.
```
breedx-splk:
```suggestion
note: Defines a new app.build_id attribute.
```

model/app/registry.yaml
```diff
@@ -62,3 +62,8 @@ groups:
           A widget is an application component, typically an on-screen visual
           GUI element.
         examples: [ "submit", "attack", "Clear Cart" ]
+      - id: app.build_id
```
lmolkova, line 65: should we reference this attribute on app-related events/entities?

LikeTheSalad: I'm not sure I follow. Where can I add such a reference?

thompson-tomo: @LikeTheSalad perhaps add it to the `app` entity as a starting point.

LikeTheSalad: > @LikeTheSalad perhaps add it to the `app` entity as a starting point.

Right, I guess my question was more about _how_ to do so, I'm not familiarized with entities yet. I've pushed [this commit](https://github.com/open-telemetry/semantic-conventions/pull/2591/commits/c851bdac41d70518438917b06c30dea8fe09e30b) where I think I do it, please let me know if there's something else missing, @lmolkova.

thompson-tomo: Looks correct to me, as it was added to the  model & docs generated.

lmolkova: @LikeTheSalad, it looks good to me, but is there a chance to specify if it's identifying or descriptive? (The modeling guide is here https://github.com/open-telemetry/semantic-conventions/blob/main/docs/how-to-write-conventions/resource-and-entities.md). We don't specify roles for other app attributes, so it's not blocking.

lmolkova (approved):

model/app/registry.yaml
LikeTheSalad: Got it, thank you for the details. I've set an identifying role to it.

PR #2163, opened 2025-04-22
Add `gen_ai.conversation.id` attribute

gyliu513:
Fixes #2024

## Changes

Please provide a brief description of the changes here.

Note: if the PR is touching an area that is not listed in the [existing areas](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/README.md), or the area does not have sufficient [domain experts coverage](https://github.com/open-telemetry/semantic-conventions/blob/main/.github/CODEOWNERS), the PR might be tagged as [experts needed](https://github.com/open-telemetry/semantic-conventions/labels/experts%20needed) and move slowly until experts are identified.

## Merge requirement checklist

* [ ] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [ ] [schema-next.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/schema-next.yaml) updated with changes to existing conventions.

PR #2594, opened 2025-08-04
feat: introduce new span attribute to track sessionId across individual traces together

91pavan:
Accompanying google doc that goes into much more detail: https://docs.google.com/document/d/1VaMgwHXGhBaLBslAgiUbsZEGD0CObYBWS4jxMzEd7_I/edit?usp=sharing

### DIFF
--- a/model/gen-ai/registry.yaml
+++ b/model/gen-ai/registry.yaml
@@ -251,6 +251,13 @@ groups:
             Client-side operations are actions taken on the user's end or within the client application.
           Datastore: A tool used by the agent to access and query structured or unstructured external data for retrieval-augmented tasks or knowledge updates.
         examples: ['function', 'extension', 'datastore']
+      - id: gen_ai.session.id
+        stability: development
+        type: string
+        brief: The unique identifier for the GenAI session.
+        note: >
+          Groups multiple related traces under one workflow or user interaction — typically long-lived, multi-step, multi-agent processes.
+        examples: ["session_5j66UpCpwteGg4YSxUnt7lPY"]
       - id: gen_ai.data_source.id
         stability: development
         type: string
--- a/model/gen-ai/spans.yaml
+++ b/model/gen-ai/spans.yaml
@@ -231,6 +231,9 @@ groups:
       - ref: gen_ai.agent.description
         requirement_level:
           conditionally_required: If provided by the application.
+      - ref: gen_ai.session.id
+        requirement_level:
+          conditionally_required: if applicable.
 
   - id: span.gen_ai.invoke_agent.client
     type: span
@@ -258,6 +261,9 @@ groups:
         requirement_level:
           conditionally_required: when available
       - ref: gen_ai.data_source.id
+        requirement_level:
+          conditionally_required: if applicable.
+      - ref: gen_ai.session.id
         requirement_level:
           conditionally_required: if applicable.
 
@@ -289,6 +295,9 @@ groups:
       - ref: gen_ai.tool.type
         requirement_level:
           recommended: if available
+      - ref: gen_ai.session.id
+        requirement_level:
+          conditionally_required: if applicable.
       - ref: error.type
         requirement_level:
           conditionally_required: "if the operation ended in an error"


lmolkova (changes requested):
We need to figure out the relationship with browser/client app [`session.id`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/session.md) first.

Based on the current definition, the gen_ai.session.id seems to be the same, so we should rather re-use existing attribute. 

It might also be useful to discuss with the browser SIG on how to propagate the session into the backend  - [Browser	Thursday at 8:30 PT	Google Doc	#otel-browser	calendar-browser](https://github.com/open-telemetry/community)

model/gen-ai/spans.yaml
```diff
@@ -289,6 +295,9 @@ groups:
       - ref: gen_ai.tool.type
         requirement_level:
           recommended: if available
+      - ref: gen_ai.session.id
+        requirement_level:
+          conditionally_required: if applicable.
```
thompson-tomo: Rename missed.

model/gen-ai/spans.yaml
```diff
@@ -260,6 +263,9 @@ groups:
       - ref: gen_ai.data_source.id
         requirement_level:
           conditionally_required: if applicable.
+      - ref: gen_ai.session.id
+        requirement_level:
+          conditionally_required: if applicable.
```
thompson-tomo: Rename missed.

model/gen-ai/registry.yaml
```diff
@@ -251,6 +251,13 @@ groups:
             Client-side operations are actions taken on the user's end or within the client application.
           Datastore: A tool used by the agent to access and query structured or unstructured external data for retrieval-augmented tasks or knowledge updates.
         examples: ['function', 'extension', 'datastore']
+      - id: session.id
+        stability: development
+        type: string
+        brief: The unique identifier for the GenAI session.
+        note: >
+          Groups multiple related traces under one workflow or user interaction — typically long-lived, multi-step, multi-agent processes.
+        examples: ["session_5j66UpCpwteGg4YSxUnt7lPY"]
```
thompson-tomo, line 262: Should not need to be defined. Might be worthwhile to copy brief/note on to the ref's.

model/gen-ai/registry.yaml
```diff
@@ -251,6 +251,13 @@ groups:
             Client-side operations are actions taken on the user's end or within the client application.
           Datastore: A tool used by the agent to access and query structured or unstructured external data for retrieval-augmented tasks or knowledge updates.
         examples: ['function', 'extension', 'datastore']
+      - id: session.id
```
gyliu513, line 256: I think we already have `gen_ai.conversation.id` https://github.com/open-telemetry/semantic-conventions/blob/main/model/gen-ai/registry.yaml#L208-L212, the `gen_ai.conversation.id` is for a conversation (session, thread), used to store and correlate messages within this conversation.

91pavan: Hi @gyliu513, thanks for the feedback. 

Correct me if I'm wrong, but doesn't `conversation.id` only cover - 
* co-relating a sequence of LLM message turns (e.g., prompt → response → user follow-up).
* and which is useful in chat-style agent interactions, like LangChain or ChatGPT sessions.
* and that lives within a single trace.

It does not fully cover the broader concept of a workflow session in multi-agent systems. `session.id` is useful to represent and measure cross-agent workflows, particularly for multi-agent orchestration and observability.

gyliu513: Actually, we I propose this PR, I was hoping the `conversational id` should able to cover both, for both single agent and multi-agent, the only requirement is the instrumentation code need to set the `conversatonal id` for their agent frameworks.  Can you help review https://github.com/open-telemetry/semantic-conventions/pull/2163#discussion_r2062704201 as well? Thanks

91pavan:
> We need to figure out the relationship with browser/client app [`session.id`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/session.md) first.
> 
> Based on the current definition, the gen_ai.session.id seems to be the same, so we should rather re-use existing attribute.
> 
> It might also be useful to discuss with the browser SIG on how to propagate the session into the backend - [Browser Thursday at 8:30 PT Google Doc #otel-browser calendar-browser](https://github.com/open-telemetry/community)

Thanks @lmolkova. Renamed it accordingly. 

As far as PoC is concerned, we have implemented session propagation for multiple protocols in Observe SDK (https://github.com/agntcy/observe).

model/gen-ai/registry.yaml
91pavan: Thanks @gyliu513. I see you already tried proposing it earlier :) 

I fully agree that `gen_ai.conversation.id` is broadly applicable across LLM client frameworks and user-facing applications — not just agentic systems.

IMO, the concept of a conversation — while essential — does not fully address more general agent (multi-agent) orchestration workflows where:

* agents operate asynchronously or in parallel
* execution may includes temporal gaps
* there is no conversational structure or direct user interaction.

gyliu513: > there is no conversational structure or direct user interaction.

@91pavan can you show a use case here? Thanks

91pavan: Use Case: Travel Itinerary Planning with Multi-Agent Delegation

User Query:

> “Plan me a 5-day trip to Japan with a focus on food and culture.”

This launches a **multi-agent workflow** spanning multiple domains and phases. Some steps involve brief user conversations, others are fully autonomous:

---

### Multi-Agent Execution Plan

| Agent                 | Task Description                                                 | User Chat? |
|-----------------------|----------------------------------------------------------------|-------------------------|-------------|
| **Planner Agent**       | Breaks task into subgoals (flights, hotels, meals, etc.)           | ✅ Yes      |
| **Flight Agent**        | Searches for and books flights                                      | ❌ No       |
| **Hotel Agent**         | Recommends and books accommodations                               | ❌ No       |
| **Food Guide Agent**    | Maps top culinary spots                      | ✅ Yes      |
| **Budget Agent**        | Monitors and optimizes for budget constraints across agents         | ❌ No       |

---

> All agent traces share a common attribute:  
> `gen_ai.session.id = "xyz"`

---

### Case for `session.id`

- **Better Correlation:**
  - Links both conversational and non-conversational spans into one travel planning session.
  - Even when agents act silently in the background, they’re associated with the same user intent.

- **Async Execution Support:**
  - Tasks are completed at different times (e.g., hotel search might retry 3x).
  - Still part of one cohesive travel session.

- **Cross-Agent Coordination:**
  - Budget Agent watches cost across flight, hotel, food — requires shared context. They could be communicating via an agentic protocol like A2A for example.

- **Analytics Example Queries:**
  - *“How long does an average travel planning session take?”*
  - *“Which agents fail most often?”*

---

###  Why `gen_ai.conversation.id` Could Falls Short

- Only the Planner and Food Guide interact with the user.
- Other agents operate headlessly — no message threads or conversation structure.
- `conversation.id` can only track chat turns, not full orchestration.

91pavan:
For GenAI use cases, the session here can originate in the client (browser or app) or be created server-side, but it will follow the same semantics as the existing `session.id`.

I agree it’s worth aligning with the Browser SIG on session propagation.

model/gen-ai/registry.yaml
91pavan: done

model/gen-ai/registry.yaml
91pavan: @gyliu513 any feedback? Thanks.

aabmass:
@91pavan would you mind summarizing the discussion from SIG today?

91pavan:
Sure thing.

- The concept of sessions is needed, where there could be `n` multi-turn conversations within a single e2e run.
- Figure out if the existing `gen_ai.conversation.id` attribute would be sufficient to group together traces belonging to a single session.
- Figure out if we can use `workflow.id` instead, since `session.id` could be specific to browsers. But `session` as a concept is pretty generic. Workflows and conversations can be modeled on top of sessions. In multi-agent systems, `session.id` gives a stable boundary for all traces that happen under one “user/system session.”

91pavan:
The existing `gen_ai.conversation.id` attribute has some limitations for session grouping:

- Scope limitation: Designed for single conversation flows, not multi-conversation sessions
- Semantic mismatch: A session containing multiple conversations would require multiple conversation IDs
- Hierarchy confusion: Sessions should logically contain conversations, not the reverse

Rationale for `session.id` - 

- Clear hierarchy: `session.id` > `gen_ai.conversation.id` > individual spans
- Generic concept: Sessions exist across all computing contexts, not just browsers
- Stable boundary: Provides the consistent grouping that's needed for multi-agent systems

model/gen-ai/registry.yaml
thompson-tomo: Appears not to be resolved. These briefs/notes should be added to the ref and the attribute removed from here

91pavan:
Related Issue - https://github.com/open-telemetry/semantic-conventions-genai/issues/51

91pavan:
reopen

Issue #2602, opened 2025-08-09
Reduce the number of uncategorized deprecations


thompson-tomo:
Currently there is a few deprecations which are like the below:

> reason: uncategorized
> note: Use `xyz` attribute instead.

However these should be replaced with a rename deprecation.

lmolkova:
this should be resolved now

PR #2605, opened 2025-08-09
[chore] Use rename deprecations for http #2602

thompson-tomo:
Progresses #2602

## Changes

This uses a structured rename deprecation for the


docs/registry/attributes/http.md
```diff
@@ -114,9 +114,9 @@ Describes deprecated HTTP attributes.
...
 | <a id="http-method" href="#http-method">`http.method`</a> | string | Deprecated, use `http.request.method` instead. | `GET`; `POST`; `HEAD` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `http.request.method`. |
-| <a id="http-request-content-length" href="#http-request-content-length">`http.request_content_length`</a> | int | Deprecated, use `http.request.header.content-length` instead. | `3495` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `http.request.header.content-length`. |
+| <a id="http-request-content-length" href="#http-request-content-length">`http.request_content_length`</a> | int | Deprecated, use `http.request.header.content-length` instead. | `3495` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `http.request.header`. |
```
arminru, line 117: Unrelated to this PR but it would be good if the template was changed to use the same format string as in the Attribute column where it's rendered as [`http.response.header.<key>`](https://github.com/thompson-tomo/semantic-conventions/blob/chore/%232602_HttpDeprecations/docs/registry/attributes/http.md#http-response-header), WDYT?

arminru (approved):

docs/registry/attributes/http.md
thompson-tomo: Have raised https://github.com/open-telemetry/semantic-conventions/issues/2615 to discuss this as I agree with you that this should be improved.

model/http/deprecated/registry-deprecated.yaml
```diff
@@ -49,16 +49,16 @@ groups:
         brief: 'Deprecated, use `http.request.header.content-length` instead.'
         stability: development
         deprecated:
-          reason: uncategorized
-          note: Replaced by `http.request.header.content-length`.
+          reason: renamed
+          renamed_to: http.request.header
```
trask, line 53: is there validation that prevents using `renamed_to: http.request.header.content-length` here?

model/http/deprecated/registry-deprecated.yaml
```diff
@@ -49,16 +49,16 @@ groups:
...
         examples: 3495
       - id: http.response_content_length
         type: int
         brief: 'Deprecated, use `http.response.header.content-length` instead.'
         stability: development
         deprecated:
-          reason: uncategorized
-          note: Replaced by `http.response.header.content-length`.
+          reason: renamed
+          renamed_to: http.response.header
```
lmolkova, line 61: it's not just a rename, the type is changing from int to string, so it's intentionally uncategorized

model/http/deprecated/registry-deprecated.yaml
thompson-tomo: Yes there is #2615

lmolkova:
as mentioned in https://github.com/open-telemetry/semantic-conventions/pull/2605#discussion_r2271270806, this is a legit use of uncategorized, the rename deprecation only applies when attribute can be renamed without any additional transformations.

Issue #2597, opened 2025-08-05
Revive `event.name` attribute


pellared:
### Area(s)

area:otel

### What's missing?

**Scenario 1: exporting event records for non-OTLP exporters**

Non-OTLP exporters (e.g. a console/stdout exporter) need to emit [EventName](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-eventname) .

**Scenario 2: emitting event records using logging libraries**

Application developers that are already using some existing logging libraries (like `log4j`) may not want to additionally depend on OTel Logs API in places where they want to emit application-specific event records. Most logging libraries do not have a "event name" or "event ID" concept.

### Describe the solution you'd like

Undeprecate and change the description of `event.name` in https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/event.md

**Scenario 1:**
The non-OTLP exporters (like stdout exporter) can add `event.name` attribute when exporting event records.

**Scenario 2:** 
Users of non-OTel logging libraries can add `event.name` attribute/field when emitting logs. Then the SDK or Collector can be configured with a log record processor that sets the `EventName` on log records that contain the `event.name` attribute.
Prior art:
- https://github.com/open-telemetry/opentelemetry-go/pull/7128

Other attribute keys that were considered:

- `otel.event.name` - Add `otel.event.name` attribute to the [OTel attributes registry](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/otel.md). However, given the feedback it seems that there is a preference to use `event.name` instead. I tried to summarize here: https://github.com/open-telemetry/semantic-conventions/issues/2597#issuecomment-3163015578
- `otel.event_name` - this was the original proposal, @trask gave feedback that having `event.` scopes it under "event" signal/domain

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

pellared:
@open-telemetry/specs-semconv-maintainers, I would appreciate it if we could reach an agreement on the attribute key (name) in the form of a comment on this issue, so that I can merge:
- https://github.com/open-telemetry/opentelemetry-go/pull/7128

lmolkova:
[...]

PR #2606, opened 2025-08-09
[chore] Desiginate event.name as obsoleted

thompson-tomo:
## Changes

This marks event.name as obsoleted rather than uncategorised,

### DIFF
--- a/model/event/deprecated/registry-deprecated.yaml
+++ b/model/event/deprecated/registry-deprecated.yaml
@@ -9,7 +9,7 @@ groups:
         type: string
         stability: development
         deprecated:
-          reason: uncategorized
+          reason: obsoleted
           note: >
             Replaced by EventName top-level field on the LogRecord.
         brief: >


thompson-tomo:
This PR was designed to not change the definition of any attribute but rather is a documentation chore to increase consistency. Yes it might only be short lived change if the attribute is undeprecated but that is likely to take longer.

joaopgrassi:
I'd also rather wait to the resolution of #2597 before merging this. Let's not make extra work for ourselves if it will be brought back.

thompson-tomo:
@lmolkova as part of #2607 I did a search for uncategorised to see which ones were not set. I then went through setting them with explicit ones when it was clear what it should be based on definition in weaver. In the case here there was no replacement attribute hence fitted nicely in obsoleted.

lmolkova:
There is a replacement - it's the event_name property on the proto. It's not an attribute though.

thompson-tomo:
Three is not a replacement attribute as we are talking about attributes. It is obsoleted as now it is not needing to be added as an attribute as the message property can be directly used.

pellared:
Closing per https://github.com/open-telemetry/semantic-conventions/pull/2715

pellared:
@thompson-tomo, this can be reopened per https://github.com/open-telemetry/semantic-conventions/pull/2914 and https://github.com/open-telemetry/semantic-conventions/issues/2913.

PR #2607, opened 2025-08-09
[chore] Introduce additional requirement levels #2601

### LINKED ISSUE #2601: Improve consistency in migrating to new semconv
We should introduce a new attribute requirement level of migrate to indicate when an attribute should be migrated to a replacement.

To go alongside this change the attributes requirement doc should be updated to include this new level.

Key points in description would be:
- Stable instrumentations should continue emitting unless opted not to via exclusion or env variable
- unStable instrumentations should not continue emitting unless opted to via inclusion or env variable
- Can be removed from stable upon major version bump
- Can be removed from unstable upon bumping of deployment level

thompson-tomo:
Fixes #2601

## Changes

This defines 2 new attribute requirement levels migrate & remove which is used in conjunction with deprecations. With this change we can now see which attributes should be removed and which ones are needing to be migrated.

Note this is also needed for potential code generation as we are needing a way to support side by side.


docs/general/attribute-requirement-level.md
```diff
@@ -112,6 +120,75 @@ particularly expensive to retrieve or might pose a security or privacy risk.
...
+
+The type of instrumentation helps to determine how the attribute should be handled, see below.
+
+### Stable Instrumentation
+
+Should continue emitting the attribute unless:
+
+* User has set the domain ie `database` via the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable.
+* User has excluded the attribute via explicit configuration
+* The instrumentation bumps it's major version but will continue providing security patches for
```
KalleOlaviNiemitalo: "its" without apostrophe.

docs/general/attribute-requirement-level.md
```diff
@@ -112,6 +120,75 @@ particularly expensive to retrieve or might pose a security or privacy risk.
...
+designed to help support a allowing for a phased rollout of the stable semantic conventions.
+Under no circumstances should this attribute be added to an existing instrumentation.
+
+The type of instrumentation helps to determine how the attribute should be handled, see below.
+
+### Stable Instrumentation
+
+Should continue emitting the attribute unless:
+
+* User has set the domain ie `database` via the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable.
```
KalleOlaviNiemitalo: "e.g." seems more appropriate than "ie", as "database" is an example.

Also applies to the "long-term unstable" section below.

docs/general/attribute-requirement-level.md
```diff
@@ -112,6 +120,75 @@ particularly expensive to retrieve or might pose a security or privacy risk.
 These should therefore only be enabled explicitly by a user making an informed
 decision.
 
+## Migrate
+
+The migrate requirement level is reserved for deprecated attributes and is
+designed to help support a allowing for a phased rollout of the stable semantic conventions.
```
KalleOlaviNiemitalo: Is a word missing near "a allowing"?

docs/general/attribute-requirement-level.md
```diff
@@ -112,6 +120,75 @@ particularly expensive to retrieve or might pose a security or privacy risk.
...
+* User has excluded the attribute via explicit configuration
+* The instrumentation bumps it's major version but will continue providing security patches for
+the previous major version for at least 6 months.
+
+Removal can occur when the major version is bumped provided previous major version will/has received 6 months of security patches from the time the replacement attribute is introduced.
+
+### Long-term Unstable Instrumentation
+
+> [!NOTE]
+> Example's of long term unstable instrumentation, would be the OpenTelemetry Contrib packages as
```
KalleOlaviNiemitalo: "Examples" without an apostrophe.

docs/general/attribute-requirement-level.md
```diff
@@ -112,6 +120,75 @@ particularly expensive to retrieve or might pose a security or privacy risk.
...
+Under no circumstances should this attribute be added to an existing instrumentation.
+
+The type of instrumentation helps to determine how the attribute should be handled, see below.
+
+### Stable Instrumentation
+
+Should continue emitting the attribute unless:
+
+* User has excluded the attribute via explicit configuration
+* The instrumentation bumps it's major version but will continue providing security patches for
```
KalleOlaviNiemitalo: "its" without an apostrophe.

docs/general/attribute-requirement-level.md
```diff
@@ -112,6 +120,76 @@ particularly expensive to retrieve or might pose a security or privacy risk.
 These should therefore only be enabled explicitly by a user making an informed
 decision.
 
+## Migrate
```
jsuereth: This is an interesting concept.  I believe we were more thinking for things like this there would be a "major version release" that would have its own sub-migration, vs piecemeal / fine-grained migration.

Curious what other @open-telemetry/specs-semconv-maintainers + @open-telemetry/weaver-maintainers think there.

thompson-tomo: This approach focuses on improving the rendering of deprecated attributes by explaining what to do with them but more significantly reduces the ambiguity in the current solution. You could still have tagged releases with sub-migration.

docs/messaging/azure-messaging.md
```diff
@@ -63,8 +63,11 @@ The following additional attributes are defined:
...
 | [`messaging.message.conversation_id`](/docs/registry/attributes/messaging.md) | string | Message [correlation Id](https://learn.microsoft.com/azure/service-bus-messaging/service-bus-messages-payloads#message-routing-and-correlation) property. | `MyConversationId` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`messaging.message.id`](/docs/registry/attributes/messaging.md) | string | A value used by the messaging system as an identifier for the message, represented as a string. | `452a7c7c7c7048c2f887f61572b18fc2` | `Recommended` If span describes operation on a single message. | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`messaging.message.payload_compressed_size_bytes`](/docs/registry/attributes/messaging.md) | int | The compressed size of the message payload in bytes. [10] | `2048` | `Remove` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Obsoleted. |
```
lmolkova: this approach prioritizes migration over greenfield instrumentation - for the latter it pollutes conventions with attributes instrumentation does not need to know about. I'd rather prioritize greenfield development and keep conventions clean

docs/general/attribute-requirement-level.md
lmolkova: it does not look like a real requirement level - it does not appear in the yaml schema and is not visible to code generation. I don't suggest to make it a requirement level though.

The goal seems to be to 
- render a 'past' version of conventions 
- provide migration instructions

That's what we're trying to do with [migration guides](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/db-migration.md). 

I'd be happy to see them automated (it's possible now) and separate from the general conventions which tell how telemetry should look following the current version and nothing else.

docs/messaging/azure-messaging.md
thompson-tomo: Actually it is all about making it easy for a user by providing a single spot to view the attributes rather than having to browse through many different versions to determine if it is even a valid attribute. What has now been done is de-emphasised them by splitting them into seperate tables which provides additional benefits.

docs/general/attribute-requirement-level.md
thompson-tomo: Ok, based on that feedback i have changed it from a requirement level to a deprecation action which is now rendered in a separate table and requirement levels are not being touched.

The goal is to provide a single spot for a user to go to get an explaination of the attributes on their traces, metrics etc without needing to jump through many different documents. It can be used by new contributors to oss projects to know quickly know which attributes are from the spec and which one's have been added by other developers not following the spec.

It also helps to address the ambguity in how to update the signal definition being emitted when wanting to follow the published guidance which is especially important when the definitions being updated to are still not stable. Hence the clear definition of the actions.

As it currently stands if you were wanting to update an elasticsearch instrumentation to the latest spec from pre < 1.20 you end up in a massive grey area as:
- the migration guide says the OTEL_SEMCONV_STABILITY_OPT_IN variable is only when migrating to stable, 
- the elasticsearch page has no warning on it and is defined as experimental, 
- the database span page has a warning that you need the variable if upgrading from <= 1.24 so you could upgrade to 1.25 (also not stable) using the variable for the changed attributes but that introduces new grey area. Then any subsequent updates if not changing to a stable definition you don't need the stability opt in variable to control them.

With the approach here you can see which attributes need to be controlled via the stability environment variable, which one's dont and how the instrumentation stability impacts the behaviour.

The migration guide seems alot more like a changelog/revision history when you read it.

lmolkova: we can shape migration guides to provide any details.

with this proposal, we'd have more confusion:
- current and previous versions of semconv in the same page
- all previous versions squashed together into 'past attributes'
- no details on non-attribute parts of the past versions.

I don't see how it reduces any ambiguity around elasticsearch - it does not even change anything on its page. If there is a problem there, we better address it first without introducing new notions.

With current state in main:
- users can check past version via tags, e.g. https://github.com/open-telemetry/semantic-conventions/tree/v1.36.0. They should have schema-url in their telemetry to tell them which version they are in
- users can check the current version on main
- they can read the migration guide to learn how to get from version vPast to vCurrent and explore which attributes were used by the conventions in the past.

What I'm still suggesting is to build a better vision on what's included in the migration guides and automate their  generation instead of squashing them into one page in actual conventions. 

I would support adding a link to migration guide from the 'actual' semconv page

thompson-tomo: No where has it been stated by me that this would remove the migration guides, in fact I stated you could still have them.

> all previous versions squashed together into 'past attributes'

This is No different to what you have now in migration guide for db for example. 

> no details on non-attribute parts of the past versions.

Migration Guide could & would still exist to cover those additional things.

> I don't see how it reduces any ambiguity around elasticsearch - it does not even change anything on its page.

That's because the signals don't currently contain the past attributes which would need for codegen etc. Once they are added to the model, the docs get updated. I have only added a couple of attributes to verify it

> If there is a problem there, we better address it first without introducing new notions.

It is a general problem which can be reproduced across the conventions and not restricted to the db namespace, but also applies to messaging etc.

> They should have schema-url in their telemetry to tell them which version they are in

I have never encountered the schema-url being set in the implementations i have seen.

> What I'm still suggesting is to build a better vision on what's included in the migration guides and automate their generation instead of squashing them into one page in actual conventions.

I agree with automating and improving the migration guides but I don't see it as an either or scenario but rather they complement each other. As such I would like to see both of them with the deprecation actions also appearing in the migration guides as a way to describe how to make the upgrade as that is where the grey area is.

PR #2610, opened 2025-08-10
[db] Add context propagation for Oracle

sincejune:
## Changes

This PR introduces context propagation for Oracle database (similiar to #2363)

### Note
An alternative approach involves using [DBMS_SESSION.SET_CONTEXT](https://docs.oracle.com/en/database/oracle/oracle-database/23/arpls/DBMS_SESSION.html#GUID-395C622C-ED79-44CC-9157-6A320934F2A9) to set context. However, this PR opted for the `V$SESSION.ACTION` solution for two primary reasons:
1. The performance overhead of `V$SESSION.ACTION` is very low, especially when set via `Connection.setClientInfo()` in JDBC, as the information is batched with the next database call. In contrast, `DBMS_SESSION.SET_CONTEXT` typically requires an extra network round-trip.
2. It is simpler to use, unlike `DBMS_SESSION.SET_CONTEXT` which needs [CREATE CONTEXT](https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/CREATE-CONTEXT.html) permission to create the context namespace.

Any feedback is welcome!

  * https://github.com/signalfx/splunk-otel-java/pull/2405

### DIFF
--- a/docs/database/oracledb.md
+++ b/docs/database/oracledb.md
@@ -149,7 +149,7 @@ and SHOULD be provided **at span creation time** (if provided at all):
 
 ### V$SESSION.ACTION
 
-Instrumentations MAY propagate context using [V$SESSION.ACTION](https://docs.oracle.com/en/database/oracle/oracle-database/23/refrn/V-SESSION.html) by injecting fixed-length part of span context (trace-id, span-id, trace-flags, protocol version) before executing a query. For example, when using W3C Trace Context, only a string representation of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) SHOULD be injected. Context injection SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
+Instrumentations MAY propagate context with a fixed-length, 64 byte value using [V$SESSION.ACTION](https://docs.oracle.com/en/database/oracle/oracle-database/23/refrn/V-SESSION.html) by injecting part of span context (trace-id, span-id, trace-flags, protocol version) before executing a query. For example, when using W3C Trace Context, only a string representation of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) SHOULD be injected. Context injection SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
 
 Variable context parts (`tracestate`, `baggage`) SHOULD NOT be injected since `V$SESSION.ACTION` value length is limited to 64 bytes.
 


docs/database/oracledb.md
```diff
@@ -141,6 +143,38 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
 
+## Context propagation
+
+**Status**: [Development][DocumentStatus]
+
+### V$SESSION.ACTION
+
+Instrumentations MAY propagate context using [V$SESSION.ACTION](https://docs.oracle.com/en/database/oracle/oracle-database/23/refrn/V-SESSION.html) by injecting fixed-length part of span context (trace-id, span-id, trace-flags, protocol version) before executing a query. For example, when using W3C Trace Context, only a string representation of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) SHOULD be injected. Context injection SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
+
+Variable context parts (`tracestate`, `baggage`) SHOULD NOT be injected since `V$SESSION.ACTION` value length is limited to 64 bytes.
```
KalleOlaviNiemitalo, line 154: <https://docs.oracle.com/en/database/oracle/oracle-database/23/arpls/DBMS_APPLICATION_INFO.html#GUID-90DA860F-BFBE-4539-BA00-2279B02B8F26> says:

> Names longer than 32 bytes are truncated.

Is that an error in the Oracle documentation and is the limit actually 64?

sincejune: Upon local testing, I confirmed that the actual limitation for `V$SESSION.ACTION` is 64 bytes, as defined by its `VARCHAR2(64)` type in the table's definition: https://docs.oracle.com/en/database/oracle/oracle-database/23/refrn/V-SESSION.html

KalleOlaviNiemitalo: I saw the `V$SESSION.ACTION` column definition, but I suspected the case might be similar to the `@cookie` parameter of [sp_setapprole](https://learn.microsoft.com/sql/relational-databases/system-stored-procedures/sp-setapprole-transact-sql) in Microsoft SQL Server, where the parameter is defined as varbinary(8000) but only 50 bytes are currently used.

Thank you for testing.

KalleOlaviNiemitalo: I filed feedback about the length limit on the Oracle DBMS_APPLICATION_INFO documentation page.

nenadnoveljic: The field limit is 64 bytes, as documented [here](https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/V-SESSION.html). The 32-byte limit only applies to `DBMS_APPLICATION_INFO`. But that shouldn't be used outside PL/SQL anyway, since it adds an extra round trip. It's better to use the language-level API instead, like [this example](https://github.com/DataDog/dd-trace-java/pull/8090/files#diff-70e53115fc8165dd83c5e9614b45b9db27fe412f8f2e354dd6185e2f9113422aR96).

lmolkova:
@sincejune thanks for the contribution! It seems you have some prototypes, please link to them in the PR description. 

Are there some docs that recommend using V$SESSION.ACTION for debugging/observability? My brief chat with AI confirms this, but I'd be great to link to some official documentation that mentions it. 

Also checking with other folks who worked on DB conventions - I remember some of you had experience with oracle @trask @alanwest @matt-hensley in case you have context/thoughts.

nenadnoveljic:
[doc reference](https://docs.oracle.com/en/database/oracle/oracle-database/23/arpls/DBMS_APPLICATION_INFO.html) that `ACTION` is intended for app instrumentation

docs/database/oracledb.md
```diff
@@ -141,6 +143,38 @@ and SHOULD be provided **at span creation time** (if provided at all):
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+## Context propagation
+
+**Status**: [Development][DocumentStatus]
+
+### V$SESSION.ACTION
+
+Instrumentations MAY propagate context using [V$SESSION.ACTION](https://docs.oracle.com/en/database/oracle/oracle-database/23/refrn/V-SESSION.html) by injecting fixed-length part of span context (trace-id, span-id, trace-flags, protocol version) before executing a query. For example, when using W3C Trace Context, only a string representation of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) SHOULD be injected. Context injection SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
```
matt-hensley: Having the second statement ("SHOULD NOT") be the only reference to the 64 byte limit may make it easy to overlook. 

Quick rewording to call the limit out earlier for clarity:
```suggestion
Instrumentations MAY propagate context with a fixed-length, 64 byte value using [V$SESSION.ACTION](https://docs.oracle.com/en/database/oracle/oracle-database/23/refrn/V-SESSION.html) by injecting part of span context (trace-id, span-id, trace-flags, protocol version) before executing a query. For example, when using W3C Trace Context, only a string representation of [`traceparent`](https://www.w3.org/TR/trace-context/#traceparent-header) SHOULD be injected. Context injection SHOULD NOT be enabled by default, but instrumentation MAY allow users to opt into it.
```

matt-hensley:
Read through docs and this usage of `V$SESSION.ACTION` makes sense.

Only question is the 64 vs 32 byte discrepancy in the docs that are being resolved 👍

docs/database/oracledb.md
sincejune: Resolving this conversation because the 32-byte limitation does not apply to the `V$SESSION.action` column, which has a length limit of 64 bytes.

sincejune:
> Read through docs and this usage of `V$SESSION.ACTION` makes sense.
> 
> Only question is the 64 vs 32 byte discrepancy in the docs that are being resolved 👍

I believe we are fine with the 32-byte limitation, as it does not actually apply to the `V$SESSION.action` column we are using.

PR #2617, opened 2025-08-12
feat: add semantic conventions for gce instance labels

paulojmdias:
Relates to https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41919

## Changes

This change introduces a new templated attribute for Google Compute Engine instance labels.
The attribute prefix gcp.gce.instance.labels.<key> follows the existing GCP/GCE naming scheme and aligns with current semantic conventions for similar resources (e.g., container.label).


ridwanmsharif (approved):
Makes sense to me from the GCP side of things. Will wait for semconv approvers for a review

model/gcp/registry.yaml
```diff
@@ -195,3 +195,13 @@ groups:
...
+      - id: gcp.gce.instance.labels
+        type: template[string]
+        stability: development
+        brief: >
+          GCE instance labels, `<key>` being the label name and the value being the label value.
+        note: >
+          For example, a GCE instance label `env` with value `prod` SHOULD be recorded as the
+          `gcp.gce.instance.labels.env` attribute with value `"prod"`.
+          The `<key>` MUST be the exact GCE instance label key.
+        examples: [ 'prod' ]
```
thompson-tomo: Would prefer to see this in the host namespace given  the reusability potential.

thompson-tomo:
I feel that we should add this to the host namespace for maximum interoperability and work towards phasing out the dedicated gce namespace.

thompson-tomo (commented):
I feel that we should add this to the host namespace for maximum interoperability and work towards phasing out the dedicated gce namespace as the need is currently very low.

model/gcp/registry.yaml
jsuereth: This is against the direction we're taking in the entities SIG

jsuereth:
This seems highly related to a new SIG forming - https://github.com/open-telemetry/community/pull/2837

I'm supportive of adding these labels now, but I also think we should track the long term direction here, as there may be other places we want these labels.

model/gcp/registry.yaml
```diff
@@ -195,3 +195,13 @@ groups:
         brief: >
           The hostname of a GCE instance. This is the full value of the default or [custom hostname](https://cloud.google.com/compute/docs/instances/custom-hostname-vm).
         examples: ['my-host1234.example.com', 'sample-vm.us-west1-b.c.my-project.internal']
+      - id: gcp.gce.instance.labels
```
jsuereth, line 324: This will need to be attached to a signal (see `entities.yaml`). These should be descriptive attributes on the `gcp.gce.instance`

paulojmdias: Done.

I agree that this should probably discuss a long-term solution, as this data exists in multiple cloud providers, and I see a huge benefit in standardizing it.

model/gcp/registry.yaml
paulojmdias: Understood, either way, I also updated the example to be something more agnostic than the `environment`

model/gcp/gce-entities.yaml
```diff
@@ -10,3 +10,13 @@ groups:
         requirement_level: recommended
       - ref: gcp.gce.instance.hostname
         requirement_level: recommended
+      - ref: gcp.gce.instance.labels
```
jsuereth, line 13: Nit: can you add `role: descriptive` to this?

paulojmdias: Sure! Added 👍

paulojmdias: @jsuereth, can you re-open the PR?

dashpole:
@paulojmdias can you reopen this when you get the chance?

paulojmdias:
Thanks for reopen @dashpole. I couldn't reopen it 🙏

paulojmdias:
@dashpole Conflicts fixed 👍

dashpole (approved):

dashpole (approved):

lmolkova (approved):

jsuereth (approved):

PR #2619, opened 2025-08-12
feat: add semantic conventions for gce instance group manager

### LINKED ISSUE #905: gcp.gce: Add instance group name
### Area(s)

area:gcp

On GCE, [instance groups](https://cloud.google.com/compute/docs/instance-groups) can be used to organize the configuration and lifecycle of multiple VM instances as a single entity. I'd like to associate the instance group name with telemetry from its individual instances to help with grouping, in the same way that a cluster or namespace helps to group k8s pods.

A new attribute `gcp.gce.instance.group` that is *optionally* populated on GCP instances. It can be retrieved as the value of the [`created-by` instance metadata field](https://cloud.google.com/compute/docs/instance-groups/getting-info-about-migs#checking_if_a_vm_instance_is_part_of_a_mig).

* Fetching instance group at query time -- doesn't work if the instance no longer exists
* Using a label or naming convention -- fragile

### Additional context

_No response_

paulojmdias:
Fixes #905 

## Changes

This PR adds semantic conventions for Google Compute Engine (GCE) Managed Instance Group (MIG) attributes, aligning with the existing definitions in the [GCP GCE resource detection processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/internal/gcp/metadata.yaml#L74-L85).

New attributes
	•	gcp.gce.instance_group_manager.name – The IGM name managing this VM.
	•	gcp.gce.instance_group_manager.zone – Zone of a zonal IGM (set only for zonal MIGs).
	•	gcp.gce.instance_group_manager.region – Region of a regional IGM (set only for regional MIGs).

These attributes enable consistent resource labeling for VMs managed by MIGs, improving filtering, grouping, and correlation across telemetry data.


model/gcp/registry.yaml
```diff
@@ -195,3 +195,23 @@ groups:
         brief: >
           The hostname of a GCE instance. This is the full value of the default or [custom hostname](https://cloud.google.com/compute/docs/instances/custom-hostname-vm).
         examples: ['my-host1234.example.com', 'sample-vm.us-west1-b.c.my-project.internal']
+      - id: gcp.gce.instance_group_manager.name
+        type: string
+        stability: development
+        brief: >
+          The name of the Instance Group Manager (IGM) that manages this VM, if any.
+        examples: ['web-igm', 'my-managed-group']
```
thompson-tomo, line 329: Would be great if this was `host.cluster.name` as that way it could be applicable for a range of use cases/providers.

model/gcp/registry.yaml
```diff
@@ -195,3 +195,23 @@ groups:
...
+          The zone of a **zonal** Instance Group Manager (e.g., `us-central1-a`).
+          Set this **only** when the IGM is zonal.
+        examples: ['us-central1-a', 'europe-west1-b']
+      - id: gcp.gce.instance_group_manager.region
+        type: string
+        stability: development
+        brief: >
+          The region of a **regional** Instance Group Manager (e.g., `us-central1`).
+          Set this **only** when the IGM is regional.
+        examples: ['us-central1', 'europe-west1']
```
thompson-tomo, line 343: Would one or both of these be better suited in the host.platform.* namespace as proposed in #2367 with the product being an option for gce?

thompson-tomo (commented):
Would be great to see these in the host namespace to enable maximum interoperability given that these attributes are common across platforms.

model/gcp/registry.yaml
jsuereth: This is a very different concept then cluster.  I disagree with this suggestion.

model/gcp/registry.yaml
```diff
@@ -195,3 +195,23 @@ groups:
         brief: >
           The hostname of a GCE instance. This is the full value of the default or [custom hostname](https://cloud.google.com/compute/docs/instances/custom-hostname-vm).
         examples: ['my-host1234.example.com', 'sample-vm.us-west1-b.c.my-project.internal']
+      - id: gcp.gce.instance_group_manager.name
```
jsuereth, line 324: Thanks for adding this attributes!

I think we should also define an entity at the same time.  

Something like `entities.yaml`:

```yaml
groups:
- id: entities.gcp.gce.instance_group_manager
  type: entity
  name: gcp.gce.instance_group_manager
  brief: A GCE instance group manager.
  stability: development
  attributes:
  - ref: gcp.gce.instance_group_manager.name
    role: identity
    requirement_level: required
  - ref: gcp.gce.instance_group_manager.zone
    role: identity
    requirement_level:
      conditionally_required: >
        When the instance group is zonal, then zone MUST be filled out.
  - ref: gcp.gce.instance_group_manager.region
    role: identity
    requirement_level:
      conditionally_required: >
        When the instance group is regional, then region MUST be filled out.
```

model/gcp/registry.yaml
jsuereth: The issue here is that an instance group manager may not be in the same region or zone as the VM itself, so re-using `host` is problematic.  That context may not have made it onto this PR but there was a doc shared with Entity SIG + Semconv that led to the design you see today.

model/gcp/registry.yaml
paulojmdias: @jsuereth added in the following [commit](https://github.com/open-telemetry/semantic-conventions/pull/2619/commits/1cec0c829ef832c69307fc7fe6947d88a2597c24). PTAL when you have a chance

dashpole:
Thank you for doing this @paulojmdias!

dashpole:
cc @quentinmit

paulojmdias:
> Thank you for doing this @paulojmdias!

you're welcome 🙏

PR #2622, opened 2025-08-13
feat: add openstack nova as cloud entities

paulojmdias:
Relates to:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39117
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41973

## Changes

This PR introduces a new `cloud.provider` and `cloud.platform` values for OpenStack (openstack) and OpenStack Compute (Nova) (openstack_nova).
These additions align with the new OpenStack Nova resource detection processor in the Collector and ensure consistent attribute naming across all resources.

### DIFF
--- a/.chloggen/feat_39117.yaml
+++ b/.chloggen/feat_39117.yaml
@@ -14,4 +14,4 @@ note: "Add Openstack as a cloud provider in the registry"
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 # The values here must be integers.
-issues: []
+issues: [2622]


model/cloud/registry.yaml
```diff
@@ -32,6 +32,10 @@ groups:
               value: 'ibm_cloud'
               brief: 'IBM Cloud'
               stability: development
+            - id: 'openstack'
+              value: 'openstack'
+              brief: 'Openstack'
+              stability: development
```
thompson-tomo, line 38: Openstack is not a provider.

model/cloud/registry.yaml
```diff
@@ -263,6 +267,10 @@ groups:
               value: 'ibm_cloud_openshift'
               brief: Red Hat OpenShift on IBM Cloud
               stability: development
+            - id: openstack_nova
+              value: 'openstack_nova'
+              brief: Openstack Compute (Nova)
+              stability: development
```
thompson-tomo, line 273: Check id to be compliant with latest conventions.

paulojmdias:
@thompson-tomo, nothing against it, but I also think you are missing the `openstack` in #2367 

Let's wait for the reviewers also 👍

thompson-tomo:
Agree and that is because I had never heard of openstack prior to this PR.

joaopgrassi:
I think this is unfortunately blocked on https://github.com/open-telemetry/semantic-conventions/pull/2367. We need to define how to model such scenarios as it does not fit well with our current cloud provider attribute.

paulojmdias:
@joaopgrassi does this make sense to remove the `Stale` label to avoid being automatically closed?

thompson-tomo:
@paulojmdias how about I add openstack to my pr and that way this PR can be closed.

lmolkova:
Semconv SIG discussion:
- it seems Nova resource detection is only available in the collector 
- cloud.provider and platform are open enums and it's ok to put arbitrary constants in them 
- we'd recommend documenting such constants in collector unless there is another implementation (eg. in language-contrib) and then we'd be happy to accept contribution

I'm going to close this PR as not planned, but feel free to reopen if you believe it should be documented in semconv right away.

PR #2621, opened 2025-08-13
[chore] Adjust logic to be compatible with weaver 0.17

thompson-tomo:
## Changes

The templates are now compatible with weaver 0.17 with the main change being related to handling of deprecated  on enum members

Note: if the PR is touching an area that is not listed in the [existing areas](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/README.md), or the area does not have sufficient [domain experts coverage](https://github.com/open-telemetry/semantic-conventions/blob/main/.github/CODEOWNERS), the PR might be tagged as [experts needed](https://github.com/open-telemetry/semantic-conventions/labels/experts%20needed) and move slowly until experts are identified.

## Merge requirement checklist

* [ ] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [ ] Links to the prototypes or existing instrumentations (when adding or changing conventions)

PR #2624, opened 2025-08-13
Not render deprecated enum members in examples

### LINKED ISSUE #2497: Deprecated enum members included in examples for attribute
### Area(s)

tooling, docs

### What's the problem?

If an enum value is deprecated similar to https://github.com/open-telemetry/semantic-conventions/commit/0ab3f5678c82496338e6be39e8324fede86a698e than the value should not be included as an example.

Only non deprecated values should be considered when generating example values for attributes

joaopgrassi:
Fixes #2497

## Changes

Changed the examples macro to filter out enum/mapping members that are deprecated. I tested out the changes from https://github.com/open-telemetry/semantic-conventions/pull/2423/files#diff-fa59105573b579adae0af74f93808bd513aca06bdf4b3587d7bfcc0f662890f8 and the deprecated OS's do not show up anymore in the examples column in the table.

* [] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).

### DIFF
--- a/templates/registry/markdown/examples_macros.j2
+++ b/templates/registry/markdown/examples_macros.j2
@@ -11,7 +11,7 @@
 {%- elif item.examples is sequence %}{{ print_examples(item.examples) }}
 {%- else %}`{{ item.examples | trim }}`
 {%- endif %}{%- elif item.type is mapping %}
-{%- for e in item.type.members | selectattr("deprecated", "none") %}{% if loop.index0 < 3 %}{% if loop.first == false %}; {% endif %}`{{ e.value | trim }}`{% endif %}{%- endfor %}
+{%- for e in item.type.members if e.deprecated is not defined or e.deprecated is none %}{% if loop.index0 < 3 %}{% if loop.first == false %}; {% endif %}`{{ e.value | trim }}`{% endif %}{%- endfor %}
 {%- elif item.type == "enum" -%}
-{%- for e in item.members | selectattr("deprecated", "none") %}{% if loop.index0 < 3 %}{% if loop.first == false %}; {% endif %}`{{ e.value | trim }}`{% endif %}{%- endfor %}
+{%- for e in item.members if e.deprecated is not defined or e.deprecated is none %}{% if loop.index0 < 3 %}{% if loop.first == false %}; {% endif %}`{{ e.value | trim }}`{% endif %}{%- endfor %}
 {%- endif %}{% endmacro %}


templates/registry/markdown/examples_macros.j2
```diff
@@ -11,7 +11,7 @@
 {%- elif item.examples is sequence %}{{ print_examples(item.examples) }}
 {%- else %}`{{ item.examples | trim }}`
 {%- endif %}{%- elif item.type is mapping %}
-{%- for e in item.type.members %}{% if loop.index0 < 3 %}{% if loop.first == false %}; {% endif %}`{{ e.value | trim }}`{% endif %}{%- endfor %}
+{%- for e in item.type.members | selectattr("deprecated", "none") %}{% if loop.index0 < 3 %}{% if loop.first == false %}; {% endif %}`{{ e.value | trim }}`{% endif %}{%- endfor %}
 {%- elif item.type == "enum" -%}
-{%- for e in item.members %}{% if loop.index0 < 3 %}{% if loop.first == false %}; {% endif %}`{{ e.value | trim }}`{% endif %}{%- endfor %}
+{%- for e in item.members | selectattr("deprecated", "none") %}{% if loop.index0 < 3 %}{% if loop.first == false %}; {% endif %}`{{ e.value | trim }}`{% endif %}{%- endfor %}
```
thompson-tomo: This will break as soon as weaver is updated. See #2621 on the solution to support both versions.

thompson-tomo (commented):
Thanks @joaopgrassi but unfortunately this will not work once weaver is updated due to the change in depreciation definition. As such suggest implementing a solution similar to #2621 which works for both.

templates/registry/markdown/examples_macros.j2
joaopgrassi: how will it break though? Is `none` different than `undefined`? 🤔

thompson-tomo: Not sure, perhaps it is related to the change from a string to an object or changes with null property handling. 
It could be that none requires property to be there whereas undefined caters for it not being. I just know what was needed to update semconv to the latest weaver.

jsuereth: You can probably use `| reject("is_deprecated")` here

joaopgrassi: I think I got it now, seems to work.

PR #2634, opened 2025-08-15
Add in Itron as OS type

thompson-tomo:
Progresses #1257

## Changes

Adds in itron as a os type. Itron is defined as an os by rust hence the rust resource detector will be reporting this value: https://doc.rust-lang.org/std/env/consts/constant.FAMILY.html

### DIFF
--- a/.chloggen/add_itron.yaml
+++ b/.chloggen/add_itron.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: os
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add in Itron as an os.type as per rust instrumentation.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [1257]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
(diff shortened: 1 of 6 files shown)


jsuereth:
@cijothomas Would you mind weighing in on this from the rust side?  Is this an os.type you already report?

cijothomas:
> @cijothomas Would you mind weighing in on this from the rust side? Is this an os.type you already report?

We have a detector that sets the os.type attribute [here](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-resource-detectors/src/os.rs#L14C22-L14C79), but the possible values are from this list: https://doc.rust-lang.org/std/env/consts/constant.OS.html

There is another std const (https://doc.rust-lang.org/std/env/consts/constant.FAMILY.html) for OS Family, that does have "itron". Maybe the detector should use OSFamily for os.type..?

@gruebel Do you have any thoughts on this?

lalitb:
> Maybe the detector should use OSFamily for os.type..?

OS.Family seems to be broader bucket, won't cover all the OS types from specs (e.g. both linux and macOS/darwin are treated as UNIX by OS.Family). I believe we can have combination of both OS.Family and OS.Type. something like:
```rust
use std::env::consts::{OS, FAMILY};

let os_type = match FAMILY {
    // special handling for wasm and itron
    "wasm" => "wasm",
    "itron" => "itron",
    // everything else: keep using the specific OS string
    _ => OS,
};

```

lalitb:
but going through the comment - https://github.com/open-telemetry/semantic-conventions/pull/2635#pullrequestreview-3129157328

I think the Itron also falls in same reasoning. From the [wiki](https://en.wikipedia.org/wiki/ITRON_project) - this is a specification which can be used to develop RTOS, and so not a specific OS type.

thompson-tomo:
I agree `os.family` is a broader bucket which is actually what we are after. Note with #2423 the refocusing of that attribute will be done which will result in options of Unix & windows nt. Afterwards https://doc.rust-lang.org/std/env/consts/constant.OS.html would be used as the name which most of that list appears to be ie watchos.

In terms of Linux vs darwin that would actually be addressed via #2461

PR #2635, opened 2025-08-15
Add in WASM as OS

thompson-tomo:
Progresses #1257

## Changes

Adds in Wasm as an OS as it provided by frameworks such as rust, dotnet, java etc,

### DIFF
--- a/.chloggen/Add_Wasm_as_os.yaml
+++ b/.chloggen/Add_Wasm_as_os.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: os
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add in Wasm as an os so that it can be reported by resource detectors.
+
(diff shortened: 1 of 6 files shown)


thompson-tomo:
@jsuereth as you had asked for in semconv meeting `os.type` has been refocused via #2463 to be just 2 options. The implementation of wasm would be specified via `os.name`

jsuereth:
Can you answer the other questions?

Which metrics or spans that rely on os.name now would apply to wasm?

Also would be good for @open-telemetry/semconv-system-approvers to chime in.

thompson-tomo:
> Which metrics or spans that rely on os.name now would apply to wasm?

Wasm for me would be used just like the other values which is in resources/entities. I can't find any metrics/span that rely on them. In the case here this value is based on the method to implement the resource detector providing this value.

mx-psi:
I agree that this is not the right choice, I left a comment here: https://github.com/open-telemetry/semantic-conventions/pull/2423#discussion_r2277064256 summarizing my thoughts. 

I don't think we should add this just because Rust decided to put it in their list, it seems more of an ad-hoc decision that conflates a compiler target with an OS. It would be good IMO however to provide guidance regarding what to do in a situation like this (you are running on WASM, you don't know anything about the host machine). I think the right call here is just not to set `os.type`

thompson-tomo:
It is not just rust but dotnet also identifies wasm. 

> It would be good IMO however to provide guidance regarding what to do in a situation like this (you are running on WASM, you don't know anything about the host machine).

I don't see how not knowing about the host is connected to this PR given that os & host are seperate entities.

> I think the right call here is just not to set os.type.

Currently the guidance is to use a custom value if the value is not defined as a known value, hence this PR aims to reduce custom values by looking at the language api's to see what should be defined.

PR #2641, opened 2025-08-15
Add in unix kernel #66

### LINKED ISSUE #66: Add kernel related semantic conventions

thompson-tomo:
Fixes #66

## Changes

This introduces the unix kernel to capture that information while enabling additional streamlining of the os namespace.


thompson-tomo:
Awaiting feedback and this is tied into the optimisations of the OS namespace.

joaopgrassi:
CC @open-telemetry/semconv-system-approvers

PR #2648, opened 2025-08-18
Introduce metadata about SIGs and markdown generation

### LINKED ISSUE #2697: [Project management] Add table listing all current SIG/areas/groups present in semconv
Today, it is not clear which conventions are being worked on and what is the status of such work and who owns it. We should make it possible to list all current areas, and their sigs along with metadata about the status of the SIG/work. 

We should be able to look at a page and easily find out:

- Which SIGs exist
- Which SIGs are active and looking for help
- Which SIGs/areas are inactive and not accepting changes at the moment

This should help not only contributors with more transparency and aligning expectations but also will allow maintainers to build automation around issue/PR triage. For example, an ultimate goal would be to automatically tag PRs/Issues on areas that are currently not being worked on/active.

joaopgrassi:
Fixes #2697 

## Changes

Very early on draft of an attempt to encode SIG metadata and render it in a markdown file. The goal is to make which SIGs exist and what are their statuses more transparent to all users.


internal/tools/scripts/update-sig-tables.py
```diff
@@ -0,0 +1,90 @@
...
+
+for sig in data['sigs']:
+    name = sig['name']
+    project = sig['project']
+    board = sig['board']
+    directories = sig['directories']
+    notes = sig['notes']
+
+    owners = ",<br/>".join(
+        [f"[{owner['name']}](https://github.com/orgs/open-telemetry/teams/{owner['github']})"
```
joaopgrassi: TODO: need to consider that this can be either a team user or "just" a normal user. Not sure if we will allow such thing to be considered as a SIG owner though.

sigs-schema.json
```diff
@@ -0,0 +1,84 @@
...
+            "items": {
+              "type": "string",
+              "enum": [
+                "needs_staffing",
+                "needs_prototyping",
+                "not_accepting_changes",
+                "accepting_contributions"
+              ]
+            },
+            "description": "A list of predefined labels representing the SIG's states or needs."
```
joaopgrassi:
```suggestion
            "description": "A list of predefined labels representing the SIG's state or needs."
```

thompson-tomo:
@svrnm I suspect that this information would be better suited in the community section of the website/repo rather than duplicated into the sem conv repo.

joaopgrassi:
> @svrnm I suspect that this information would be better suited in the community section of the website/repo rather than duplicated into the sem conv repo.

No, we specifically discussed and we want to have this as part of semconv. Not everyone knows even the community repo exists, so we want to encode this info here for transparency. Additionally it will serve as the base for some automation we are planning to do around issue/PR triage/management.

thompson-tomo:
> Not everyone knows even the community repo exists

Agree that is why I mentioned the website as well.

I am just mindful of creating yet another location/source of info. As such the definition reside in community repo and the otelbot syncs it to repos?

sigs.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - name: "semconv-k8s-approvers"
+        github: semconv-k8s-approvers
+    project: "https://github.com/open-telemetry/community/blob/main/projects/k8s-semconv.md"
+    board: "https://github.com/orgs/open-telemetry/projects/114/views/1"
+    areas:
+      - area:k8s
+      - area:container
+    labels:
+      - accepting_contributions
+      - needs_prototyping
```
lmolkova: do you think we'd have SIG-specific labels?  trying to think of a scenario and only some super-advanced things come up.  So wonder if we can postpone adding those until we need it?

UPDATE: ah, I see it's used in the `status` - perhaps we could rename it as such?

PS: not sure what `needs_prototyping` mean - all SIGs need prototyping, no?

SIGS.md
lmolkova, line 1: nit: maybe `sigs` or `SIGs` ?

Also, should we render it like inside the readme? Or at least link from readme?

SIGS.md
```diff
@@ -0,0 +1,6 @@
+<!-- sigs -->
+| Name | Owners | Project | Board | Areas | Status | Directories | Notes |
```
lmolkova: I wonder if we need directories here - I'd be happy to help automating codeowners generation and this is where this info would be necessary.

lmolkova (commented):
thanks A LOT for putting it together! This builds the foundation that allows us to automate triage process and be upfront with community wrt contribution process.

SIGS.md
joaopgrassi: Yeah I was not sure what's best. One downside of doing in the general readme is that it may become long. I can go either way, no preference :)

SIGS.md
joaopgrassi: Wasn't our idea to use the YAML here as context for AI agents to automate PRs? Like, "I see you touched these files, these match these sigs, I wil  let them know/apply the labels". Or do you envision a mix between this AND the codeowners for the AI agent?

joaopgrassi:
> > Not everyone knows even the community repo exists
> 
> Agree that is why I mentioned the website as well.
> 
> I am just mindful of creating yet another location/source of info. As such the definition reside in community repo and the otelbot syncs it to repos?

The target audience of the otel website is for people instrumenting their apps or people that generate telemetry. In contrast, the target audience for this are semconv contributors. The idea is that this is not even rendered in the website. same way we don't render other contents outside of the `docs` folder in the website

sigs.yaml
joaopgrassi: Yeah that's true. I wanted to somehow encode that when someone sends a PR adding conventions, but hasn't linked any implementation/prototyping we would flag that automatically. But I guess you are right, that's probably always our expectations so may not need this label.

The idea I had (which now makes less sense) is when conventions are marked stable, the sig may not need prototyping but that's wrong xD

SIGS.md
lmolkova: I mean it should definitely be in the yaml and should be used by automations, just checking if it needs to be in this markdown as well

joaopgrassi: ah right I see. I guess doesn't hurt to get the quick overview on where to find stuff?

SIGS.md
joaopgrassi: I will keep it as a separated file and add a link from the main readme.

sigs.yaml
joaopgrassi: I removed `needs_prototyping` for now and added instead `active` and `inactive`.

internal/tools/scripts/update-sig-tables.py
joaopgrassi: Done. Now there's a `type` for owner, which can be used to add both team or individual users

trask (approved):
> This builds the foundation that allows us to automate triage process and be upfront with community wrt contribution process.

👍

cc @danielgblanco for awareness

SIGs.md
```diff
@@ -0,0 +1,33 @@
...
+SIGs have several responsibilities and goals, but their primary focus is to
+define and guide semantic conventions toward stabilization. They are also tasked
+with the long-term maintenance of these conventions, serving as domain experts
+for pull request (PR) reviews and issue triage.
+
+This document provides an overview of all SIGs involved in the
+semantic conventions project. In particular, the table below highlights
+the owners of each SIG and their current status.
+
+To learn more about SIGs for example, how to kick-start a new one, see the
```
lmolkova: wonder if we can 
- add a warning that this table is under development
- merge this PR 
- divide and conquer filling SIG details

WDYT?

lmolkova (approved):
LGTM

checking with @open-telemetry/semconv-k8s-approvers if they have any feedback 

Context: we're using k8s SIG as an example here and will add details for other SIGs as well. 
The yaml will be used to automate triage (e.g. close/comment on the PRs for inactive areas), but the automation will come as a follow up.

lmolkova:
> > Not everyone knows even the community repo exists
> 
> Agree that is why I mentioned the website as well.
> 
> I am just mindful of creating yet another location/source of info. As such the definition reside in community repo and the otelbot syncs it to repos?

you're more than welcome to help build this automation

SIGs.md
joaopgrassi: I'm gonna add all the SIGs we have today, so we should be good!

joaopgrassi: Done in [9f55485](https://github.com/open-telemetry/semantic-conventions/pull/2648/commits/9f55485f38a40bd3c6996b62721bc1de36d433fa)

joaopgrassi:
Pinging all SIG users to have a chance in validating/adding more input 

-  @open-telemetry/semconv-system-approvers
-  @open-telemetry/semconv-cicd-approvers
- @open-telemetry/semconv-security-approvers
- @open-telemetry/semconv-browser-approvers
- @open-telemetry/semconv-client-approvers

sigs.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+  - name: "Semantic Conventions: CI/CD"
+    owner:
+      - name: "semconv-cicd-approvers"
+        github: semconv-cicd-approvers
+        type: team-user
+    project: "https://github.com/open-telemetry/community/blob/main/projects/ci-cd.md"
+    board: "https://github.com/orgs/open-telemetry/projects/79"
+    areas:
+      - area:cicd
```
kamphaus: In the CICD SIG we have also worked with the following areas:
- artifact
- deployment
- test
- vcs

kamphaus (commented):
The CODEOWNERS file mentions other approver groups (eg. feature-flags).
Should those be listed as well?

sigs.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+  - name: "Semantic Conventions: GenAI"
+    owner:
+      - name: "semconv-genai-approvers"
+        github: semconv-genai-approvers
+        type: team-user
+    project: "https://github.com/open-telemetry/community/blob/main/projects/gen-ai.md"
+    board: "https://github.com/orgs/open-telemetry/projects/82"
+    areas:
+      - area:gen-ai
```
lmolkova:
```suggestion
      - area:gen-ai
      - area:openai
```

sigs.yaml
```diff
@@ -0,0 +1,134 @@
...
+  - name: "Semantic Conventions: Client Instrumentation"
+    owner:
+      - name: "semconv-client-approvers"
+        github: semconv-client-approvers
+        type: team-user
+    project: "https://github.com/open-telemetry/community/blob/main/projects/client-instrumentation.md"
+    board: "https://github.com/orgs/open-telemetry/projects/19"
+    areas:
+      - area:client
+      - area:event
```
lmolkova: is this accurate? I think `area:event` is owned by the @open-telemetry/semconv-log-approvers.

wdyt @open-telemetry/semconv-client-approvers ?

sigs.yaml
```diff
@@ -0,0 +1,134 @@
...
+        github: semconv-messaging-approvers
+        type: team-user
+    project: "https://github.com/open-telemetry/community/blob/main/projects/currently-inactive/messaging-semconv.md"
+    board: "https://github.com/orgs/open-telemetry/projects/20"
+    areas:
+      - area:messaging
+    labels:
+      - inactive
+      - needs_staffing
+    notes: "The SIG is inactive. If you have interest in working towards messaging stabilization, reach out to maintainers to discuss activating the SIG again"
```
lmolkova: Maybe 

```suggestion
    notes: "The SIG is inactive. If you have interest in working towards messaging stabilization, consider proposing a project update in the https://github.com/open-telemetry/community"
```

or something? I feel if someone asks me, I'd tell them that + advertise on slack

sigs.yaml
```diff
@@ -0,0 +1,134 @@
+sigs:
+  - name: "Semantic Conventions: System"
+    owner:
+      - name: "semconv-system-approvers"
+        github: semconv-system-approvers
+        type: team-user
```
lmolkova: BTW, do we have a case for individual being a user? I think we mention @tigrannajaryan for schema transformations and logs in codeowners, but this is now owned by the tooling approvers and log approvers. I'd rather stick to group being the default and stop supporting individuals.

sigs.yaml
```diff
@@ -0,0 +1,134 @@
+sigs:
+  - name: "Semantic Conventions: System"
```
lmolkova: Looking into https://github.com/open-telemetry/semantic-conventions/blob/main/.github/CODEOWNERS (whihc I hope we'd eventually populate from this), I think we're missing a few:

- inactive HTTP, DB, JVM, Feature Flags (bugs and bugfixes are welcome, features need a new project) - I can see how this can be a default, but having them along with the status listed could be helpful - allows to document relevant areas and provide personalized notes - both would be used by the future automation
- .NET - active, but driven by .NET runtime, contributions are welcome if they are aligned with .NET runtime features
 - mainframes - active (?)
 - logs - active
 - profiling - active

maybe a few more.

sigs.yaml
joaopgrassi: I basically looked at their project board and picked labels based on issues/PRs there. I was also not sure though.

sigs.yaml
joaopgrassi: I didn't have it it in the initial draft but thought it would be good to make it flexible. If we don't need it, I can remove it again

joaopgrassi:
> The CODEOWNERS file mentions other approver groups (eg. feature-flags). Should those be listed as well?

yeah I wasn't sure if I wanted to add the "finished" sigs as part of this PR (would be too big) but I guess I could.

sigs.yaml
joaopgrassi: Done!

sigs.yaml
joaopgrassi: Removed.

sigs.yaml
joaopgrassi: Ok, I added the above - but .NET, Mainframe, Logs, Profiling are all missing either the project board or project document in the community repo (or both). I'm not even sure if those were even properly defined. 

I didn't want to leave the project/board fields as optional, to really enforce consistency and well defined process. For now for those I just used `N/A`. Sucks but.. oh well.

kamphaus (approved):

SIGs.md
```diff
@@ -0,0 +1,49 @@
+# Special Interest Groups (SIGs)
+
+Semantic conventions in OpenTelemetry are organized into distinct areas,
+with each area being overseen by a Special Interest Group (SIG). A SIG is
```
lmolkova:
```suggestion
with some of them being overseen by Special Interest Group (SIG).
```

we actually have full coverage, right?

SIGs.md
```diff
@@ -0,0 +1,49 @@
...
+composed of multidisciplinary contributors who bring expertise in both the
+relevant domain and OpenTelemetry itself.
+
+SIGs have several responsibilities and goals, but their primary focus is to
+define and guide semantic conventions toward stabilization. They are also tasked
+with the long-term maintenance of these conventions, serving as domain experts
+for pull request (PR) reviews and issue triage.
```
lmolkova: maybe something around these lines?
There is a push in the commuity repo to have smaller, well-scoped projects and permanent SIGs are seen more as an exception than a common scenario. E.g. we had HTTP or JVM  SIG but they had achieve their goals and are dissolved now.

```suggestion
define and guide semantic conventions toward stabilization. SIGs that achieve their goal 
may be dissolved, but the codeowner group stays serving as domain experts
for pull request (PR) reviews and issue triage.
```

SIGs.md
```diff
@@ -0,0 +1,49 @@
...
+
+This document provides an overview of all SIGs involved in the
```
lmolkova:
```suggestion
This document provides an overview of all SIGs and codeowner groups involved in the
```

sigs.yaml
```diff
@@ -0,0 +1,217 @@
+sigs:
```
lmolkova: nitpicking and we can always rename it later: those are not really sigs, but components/domains. SIGs come and go, we can form a new DB SIG to improve conventions for semantic/vector search it'd have new project description and board, but it would not change the entry here, codeownership group, etc.

sigs.yaml
lmolkova: yeah, I think we overlooked the lightweight-process (no sig ever) and already stable (SIG does not exist anymore) cases. added some suggestions to the text

sigs.yaml
joaopgrassi: Makes sense. But I'm not sure about calling it components/domains. Do you have a suggestion? What about reusing the term `areas`?

joaopgrassi:
Not sure why the link check fails for the https://github.com/orgs/open-telemetry/projects/45. It works for me all the time 🤔. It seems it is a bit slow, but other than that I'm not sure why it does not like it.

danielgblanco:
> Not sure why the link check fails for the https://github.com/orgs/open-telemetry/projects/45.

I don't think this is related, but that project was closed in https://github.com/open-telemetry/community/pull/2931. The GH project was not being used, or updated, by SIG maintainers.

sigs.yaml
danielgblanco: Maybe worth calling it `projects`? (note: we're thinking of renaming these "projects" to something else like "initiative" because the word "project" is way too overused... however this is what we have now).

I've tried to add some clarifications [here](https://github.com/open-telemetry/community/blob/main/project-management.md#relation-to-sigs-github-projects-and-roadmap) about the relationship between projects and SIGs.

In particular, and since I'm seeing a 1:1 between SIG and project/board in the structure below: a project is necessary to create a new SIG, with a set of deliverables, but a project is not a SIG. A project needs a somewhat short/mid-term scope, and after that the SIG may continue to exist to take on new work, or be dissolved.

We want to encourage shorter projects, with tighter scopes, that allow the community to get better insights into the current focus, and what they can expect in the mid-term. Also, when people sign up to contribute they may just want to commit to the next 6-12 months and cannot commit to e.g. 2 years of work.

For instance, the CI/CD semconv SIG below is nearing completion of their [Phase 1](https://github.com/orgs/open-telemetry/projects/79) project and moving to [Phase 2](https://github.com/open-telemetry/community/pull/2884). Each one of those has a set of deliverables.

If the intention of this list is showing the current areas of focus of SemConv SIGs, maybe `projects` would be more appropriate?

danielgblanco: Worth noting, the current `sigs.yml` in `community` now has an extra [field](https://github.com/open-telemetry/community/blob/7771f05d31542a3d1d6c10d8a3323b4a7808a4ee/sigs.yml#L175) called `roadmapProjectIds` which is used to populate https://github.com/orgs/open-telemetry/projects/158. This will represent the current/recent projects being worked on by a given SIG.

Is the expectation that those align with the projects listed here? Will it be the responsibility of the SIGs to keep this (as well as `community` list up to date).

joaopgrassi: Thanks @danielgblanco for the insights. I think project then giving your details here don't really match the goal of the list here. 

I tend to think it aligns more with what @lmolkova referred above: 

> SIGs come and go, we can form a new DB SIG to improve conventions for semantic/vector search it'd have new project description and board, but it would not change the entry here, codeownership group, etc.

Would also apply to projects, as you said they are short-lived and come and go. But here the "area/domain" always stays.

If it's OK to everyone I will rename it to `areas` then. It also aligns with the `area:{}` labels we have :)

joaopgrassi: > Worth noting, the current `sigs.yml` in `community` now has an extra [field](https://github.com/open-telemetry/community/blob/7771f05d31542a3d1d6c10d8a3323b4a7808a4ee/sigs.yml#L175) called `roadmapProjectIds` which is used to populate [OpenTelemetry Roadmap](https://github.com/orgs/open-telemetry/projects/158). This will represent the current/recent projects being worked on by a given SIG.
> 
> Is the expectation that those align with the projects listed here? Will it be the responsibility of the SIGs to keep this (as well as `community` list up to date).

@danielgblanco same as above - the project comes and goes so I don't think we need to also use the `id` here. We want to refer to the SemConv area and its "current status". For ex, I see the flow:

- area is inactive, in need for stabilization
- community decides to start a new stabilization project (e.g., db, rpc)
- project is created, this yaml here is updated to reflect the new status (link to project is added, labels are changed to `active`, `accepting_contributions` etc)
- project is done, closed, yaml here is updated to inactive again


The goal is to use this yaml as the base for automation/copilot. For ex if the yaml says a area is "inactive", it may add certain labels to issues/PRs so contributors know this isn't being considered as a priority. Maintainers then will be required as part of triage process to look at those specific items and make the final decision.

danielgblanco: Thanks @joaopgrassi. That makes sense. Reading that back, this list then summarises the different areas, and for each of them the current effort to stabilise it, which should map to an OTel "project".

joaopgrassi: We will also be adding somewhere a link to this roadmap board, filtering for items on the semconv repo. That's another PR coming

sigs.yaml
lmolkova: let's keep what we have and update areas as a follow up

joaopgrassi:
@lmolkova I renamed the top level property to `areas`. Changed then field with the areas to `labels` and the other to `status`. I think this looks more "future proof" now and detached from SIGs

jsuereth (approved):

PR #2651, opened 2025-08-19
[chore] Set enum briefs for System

thompson-tomo:
Progresses #2555

## Changes

Sets the brief property for enum members to informative details about the member using info from man page or in the case of proper noun the name stylised correctly.

### DIFF
--- a/docs/registry/attributes/system.md
+++ b/docs/registry/attributes/system.md
@@ -35,9 +35,11 @@ Describes Filesystem attributes
 
 | Value  | Description | Stability |
 |---|---|---|
-| `free` | Filesystem which is Free. | ![Development](https://img.shields.io/badge/-development-blue) |
-| `reserved` | Filesystem which has been Reserved. | ![Development](https://img.shields.io/badge/-development-blue) |
-| `used` | Filesystem has been Used. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `free` | Filesystem storage space that could have have data written to it. | ![Development](https://img.shields.io/badge/-development-blue) |
+| `reserved` | Filesystem storage space that has been set aside for a dedicated purpose. [1] | ![Development](https://img.shields.io/badge/-development-blue) |
+| `used` | Filesystem storage space that has data written to it. | ![Development](https://img.shields.io/badge/-development-blue) |
+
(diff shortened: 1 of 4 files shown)


thompson-tomo:
@ChrsMark take a look at the diff & you will see the markdown has not just a casing change of the first letter especially for file system type.

braydonk:
Most of the PR is sentence casing though. I think that comment misses the point of @lmolkova's concern.  
A description of `Free` for the enum value `free` doesn't accomplish anything meaningful in my eyes. As a viewer I would look at this and think "why did they bother writing a description if that's all it's going to say?"  
The only way I would accept this PR is if the Weaver or SemConv maintainers forced us into it for schema/tooling reasons.

thompson-tomo:
Whereas I see it the other way, why didn't the author put in a brief of just the value properly formatted as a starting point given it is done that way in other namespaces. But yes I would have preferred to put in more verbose briefs but I expected that to be a more time consuming review process hence went for a hopeful quick win with just formatting improvements.

braydonk (commented):
I'm not willing to allow this PR as proposed. It feels like a mechanical attempt at adding briefs to the ends of improving something in the markdown rendering, and not in favour of actually improving the availability of information. Much of what's in this PR is either sentence-casing, or stating something in more words with no new information (and worst-case muddying up the actual meaning of the enum value).

I'm happy to look at and accept proposals that improve the actual quality of the information presented. Adding words that carry no new information doesn't feel right to me; even if the markdown is bad without them, what I care about is the actual information being added.

thompson-tomo:
Thanks for that feedback, I have since gone and further improved the scope of information being included to be even more informative and complete using authorities sources. In some cases notes have  also been added.

The only one still just casing change is when the value represents a recognised name of a standard in which case it follows the naming of the standard.

The change also benefits code generation as it can now provide a description of the value.

If you feel any of them need further improvement it would be good to know which ones.

joaopgrassi:
These brief changes have been blocked for a while and bring mixed feelings of usefulness. In a effort to clean up our vast backlog I'm closing it.

thompson-tomo:
@joaopgrassi there has never been any feedback after switching to informative briefs which provides additional information.

Issue #2655, opened 2025-08-19
Increase ability to describe processes


thompson-tomo:
### Area(s)

area:process

### What's missing?

Dedicated entity for process executable and increase attribute usage for the process entity.

### Describe the solution you'd like

Increase descriptive attributes of process entity and split out process executable entity

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #2657, opened 2025-08-19
Add in process executable entity #2655

thompson-tomo:
Progresses #2655

## Changes


model/process/entities.yaml
```diff
@@ -36,3 +36,15 @@ groups:
...
       - ref: process.runtime.version
       - ref: process.runtime.description
+
+  - id: entity.process.executable
+    type: entity
+    stability: development
+    name: process.executable
+    brief: The executable which is being run by the process.
+    attributes:
+      - ref: process.executable.name
```
jsuereth, line 70: Can you select which ones of these will be identifying or descriptive?

thompson-tomo: Hmm, should the identity attribute values change if the executable is updated to a new version? If so we would probs need an additional attribute for version or everything becomes identity.

jsuereth: That's a question for how this entity will be used and tracked in the system.

As I think this entity is MOST relevant to @open-telemetry/profiling-maintainers  - I'd like to get there take on what "identity" for an executable should be.

Do you want to group profiles by an executable by "concept" or "by its version"?  I suspect it's the later...

thompson-tomo: It could be used as either but what I will do is define it without a version (this PR) and create a seperate pr to introduce a new attribute and add it to the entity as identifying as I see a couple of options to address it.

thompson-tomo: Seperate PR for adding an attribute for executable version https://github.com/open-telemetry/semantic-conventions/pull/2673.

thompson-tomo:
Sig have approved and requested extension has been done. Awaiting feedback.

model/process/entities.yaml
thompson-tomo: @jsuereth could you take a look as the role has been added as requested without defining any new attributes.

braydonk: It's worth noting that it's not uncommon but possible for an executable to be renamed while a process is running. Based on [the instrumentation instructions](https://github.com/open-telemetry/semantic-conventions/blob/56b613e44a93601d354490287964451642cda311/model/process/registry.yaml#L68) for this attribute, someone `mv`ing an executable during a process's runtime will result in a new name being retrieved. I haven't tried on Windows, but I'm pretty sure that renaming an executable while a process running will result in a new value from `GetProcessImageFileNameW`, so the same would apply if I'm remembering correctly (needs validation).

How to identity files as entities is a broader question in general. On Linux at least, I think the way to generally identify a file would be inode and creation time, but Windows would need some different semantics.

Is it of particular importance that `process.executable` is its own entity? Perhaps it would be easier if these attributes were instead added as descriptive attributes on the process entity.

jsuereth: Isn't this the opposite that the build id is identifying but the name may not be?

@felixge @florianl - I know you were using these attributes.  Could you help discuss the identifying vs. descriptive nature of attributes?

christos68k: In profiling, [process.executable.build_id.htlhash](https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/#process-executable-build-id-htlhash) is used to uniquely identify executables when 1-1 accuracy is needed (e.g. symbolization) but for other cases (e.g. visualization of flamegraphs) one may use `process.executable.name`, which introduces fuzziness (e.g. multiple different executables, which could be entirely unrelated functionality-wise, with the same executable name collapsed under one entity).

jsuereth: It sounds to me like we want the executable to be identified by the build-id, and the name is a descriptive attirbute which can be used for grouping.

@braydonk / @christos68k  - I think this is worth further discussion

thompson-tomo: > It's worth noting that it's not uncommon but possible for an executable to be renamed while a process is running. Based on the instrumentation instructions for this attribute, someone mving an executable during a process's runtime will result in a new name being retrieved. I haven't tried on Windows, but I'm pretty sure that renaming an executable while a process running will result in a new value from GetProcessImageFileNameW, so the same would apply if I'm remembering correctly (needs validation).

@braydonk Actually if the path to the executable is all that is changed and the file name renames the same the value of the name attribute wouldn't change.

I am more than happy to introduce a new attribute to use as the identifying attribute.

Howabout I set the name to descriptive in this PR or even leave it unset & we use #2673 for defining a new identifying attribute.

braydonk: I think it's important that the profiling SIG make the decision for what the identifying attribute(s) should be. Based on my discussion with them we at least agreed on the name not being identifying so we'll start with that, and it seems like the attribute @christos68k mentioned above is likely the best candidate, but it sounded like the Profiling SIG had other things to work through before making the final decision if I understood correctly.

thompson-tomo: Sure I have made the name attribute to be descriptive and the htlhash to not have a role so that we can discuss seperately if htlhash is the identity and the description is updated. Let me know if this PR nneeds any additional changes.

trask:

PR #2662, opened 2025-08-20
Fix release workflow 2

trask:


Makefile
```diff
@@ -322,7 +322,6 @@ NEXT_SEMCONV_VERSION ?= next
 generate-schema-next:
 	mkdir -p $(TOOLS_DIR)/bin
 	$(DOCKER_RUN) --rm \
-	$(DOCKER_USER_IS_HOST_USER_ARG) \
```
trask, line 325: @michaelsafyan this was causing error

>   × Cache directory not created: Permission denied (os error 13)

seems ok to remove podman support here since this target is only used by the release workflow, but we'd accept a more proper if you want to look into it

PR #2663, opened 2025-08-20
Document 'gcp.apphub_destination.*' attributes.

### LINKED ISSUE #2649: Minor doc update for "Google Cloud AppHub"
The [Google Cloud AppHub](https://opentelemetry.io/docs/specs/semconv/resource/cloud-provider/gcp/apphub/) page should also document that every `gcp.apphub.*` entry has a corresponding `gcp.apphub_destination.*` attribute for reporting the destination/peer AppHub application/workload/service on the other end of a `CLIENT` span.

michaelsafyan:
Fixes #2649

## Changes

Updates the markdown docs related to GCP AppHub to document destination attributes.


model/gcp/registry.yaml
```diff
@@ -172,6 +172,139 @@ groups:
...
+              stability: development
+            - id: low
+              value: "LOW"
+              brief: Low impact.
+              stability: development
+        stability: development
+        brief: >
+          Criticality of a destination workload indicates its importance to the business.
+        note: >
+          [See AppHub type enum](https://cloud.google.com/app-hub/docs/reference/rest/v1/Attributes#type)
```
lmolkova: md readability/brevity suggestion:

```suggestion
        brief: >
          Criticality of a destination workload indicates its importance to the business as specified in [AppHub type enum](https://cloud.google.com/app-hub/docs/reference/rest/v1/Attributes#type)
```

(notes are much less discoverable and having link to the original concept in the brief is nice)

model/gcp/registry.yaml
```diff
@@ -172,6 +172,139 @@ groups:
...
+              stability: development
+            - id: development
+              value: "DEVELOPMENT"
+              brief: Development environment.
+              stability: development
+        stability: development
+        brief: >
+          Environment of a destination workload is the stage of a software lifecycle.
+        note: >
+          [See AppHub environment type](https://cloud.google.com/app-hub/docs/reference/rest/v1/Attributes#type_1)
```
lmolkova: similar to prev comment - readability suggestion
```suggestion
        brief: >
          Environment of a destination workload is the stage of a software lifecycle as provided in the [AppHub environment type](https://cloud.google.com/app-hub/docs/reference/rest/v1/Attributes#type_1)
```

model/gcp/registry.yaml
```diff
@@ -172,6 +172,139 @@ groups:
...
         note: >
           [See AppHub type enum](https://cloud.google.com/app-hub/docs/reference/rest/v1/Attributes#type)
+  - id: registry.gcp.apphub_destination
+    type: attribute_group
+    display_name: GCP - AppHub Destination Attributes
+    brief: >
+      This document defines attributes AppHub will apply to destination resources in GCP.
+      See [AppHub overview](https://cloud.google.com/app-hub/docs/overview).
+    attributes:
+      - id: gcp.apphub_destination.application.container
```
lmolkova, line 182: subjective comment (not blocking):

why not `gcp.apphub.destination.application.container` here and also `.` in other attributes? wouldn't you want to have them in the same namespace ? 
In my mental model, namespaces align with the ownership model. Anything under `gcp.apphub.*` is owned by a certain group. It's ok for this group to own `gcp.apphub_destination` as well, but it feels more natural to say "everything under `gcp.apphub.*` is about GCP AppHub and is owned by XYZ team ant Google." rather than having a set of namespaces that might grow in the future.

model/gcp/registry.yaml
```diff
@@ -172,6 +172,139 @@ groups:
...
+        brief: >
+          The GCP zone or region where the destination application is defined.
+        examples: ['us-central1']
+      - id: gcp.apphub_destination.application.id
+        type: string
+        stability: development
+        brief: >
+          The name of the destination application as configured in AppHub.
+        examples: ['my-application']
+      - id: gcp.apphub_destination.service.id
```
lmolkova, line 200: is `id` a part of apphub terminology? you use name in the brief below and I'd suggest `gcp.apphub?destination.service.name` - this'd align with  `service.name` all otel users should have experience with

model/gcp/registry.yaml
```diff
@@ -172,6 +172,139 @@ groups:
...
+        type: string
+        stability: development
+        brief: >
+          The name of the destination service as configured in AppHub.
+        examples: ['my-service']
+      - id: gcp.apphub_destination.service.environment_type
```
lmolkova, line 206: the rule of thumb is to use `.` whenever it makes sense. It seems type is a property of env and (in theory) there could be multiple other properties, thus `.` makes sense

```suggestion
      - id: gcp.apphub_destination.service.environment.type
```

model/gcp/registry.yaml
```diff
@@ -172,6 +172,139 @@ groups:
...
+              stability: development
+            - id: development
+              value: "DEVELOPMENT"
+              brief: Development environment.
+              stability: development
+        stability: development
+        brief: >
+          Environment of a destination service is the stage of a software lifecycle.
+        note: >
+          [See AppHub environment type](https://cloud.google.com/app-hub/docs/reference/rest/v1/Attributes#type_1)
```
lmolkova: would it be possible to merge note into the brief (see other comments on readability)?

model/gcp/registry.yaml
```diff
@@ -172,6 +172,139 @@ groups:
...
+          Criticality of a destination service indicates its importance to the business.
+        note: >
+          [See AppHub type enum](https://cloud.google.com/app-hub/docs/reference/rest/v1/Attributes#type)
+      - id: gcp.apphub_destination.workload.id
+        type: string
+        stability: development
+        brief: >
+          The name of the destination workload as configured in AppHub.
+        examples: ['my-workload']
+      - id: gcp.apphub_destination.workload.environment_type
```
lmolkova, line 256: let's not define multiple attributes for the same thing. 
According to docs the same enum is shared

> Environment of the Application, Service, or Workload

so, can we define one and call it `gcp.apphub?destination.environment.type` or similar?

model/gcp/registry.yaml
michaelsafyan: Done.

model/gcp/registry.yaml
michaelsafyan: Done.

model/gcp/registry.yaml
michaelsafyan: Done.

model/gcp/registry.yaml
michaelsafyan: There are two reasons that it is done this way:

  1. **Implementation/safety**: the instrumentation automatically takes a protocol buffer of type `AppHub` and projects its fields either under the prefix `gcp.apphub` or under the prefix `gcp.apphub_destination`. Nothing prevents this protobuf from developing a new field named `destination`, which would create a conflict on `gcp.apphub.destination`. It was safer to project into a disjoint prefix.
 
 2. **Consistency**:  In Cloud Logging, `LogEntry` has fields `AppHub apphub` and `AppHub apphub_destination`.  The naming here reflects the naming used in Cloud Logging's `LogEntry` protobuf.

model/gcp/registry.yaml
michaelsafyan: Yes, `id` is part of the `AppHub` terminology.

The word `name` is a bit overloaded in the context of GCP. In GCP land, `name` rarely refers to a simple name/ID, but rather typically refers to the "REST resource name" (e.g. `projects/some-project/locations/some-location/resourceTypes/resourceTypeId`). It can also refer to the "fully qualified name" (the REST resource name prefixed with `//somegcpservice.googleapis.com/`).

The [REST API Reference for AppHub](https://cloud.google.com/app-hub/docs/reference/rest/v1alpha/projects.locations.applications#Application) uses the term "name" to refer to that longer name, while using "id" or "application-id" to refer to the last path segment (what is reported here).

Please also note that `gcp.apphub.service.id` is already defined; see:
https://opentelemetry.io/docs/specs/semconv/resource/cloud-provider/gcp/apphub/

Divergent naming here would break the rule that "for every attribute `gcp.apphub.*` there is an equivalent field `gcp.apphub_destination.*` corresponding to the destination application". We need this rule to hold for simplicity of implementation with respect to automatic conversation to/from unstructured attributes in OTLP and highly structured fields in an `AppHub` protocol buffer. Additionally, having consistent rules like that make it easier for downstream consumption, even for those not using such a solution, simply by increasing predictability of the attribute naming between source and destination application information.

model/gcp/registry.yaml
michaelsafyan: While this would be a reasonable suggestion in the initial design (when `gcp.apphub.service.environment_type` was defined), renaming here would break the rule that "for every attribute `gcp.apphub.*` there is an equivalent  `gcp.apphub_destination.*` field corresponding to the destination application".

Given that [`gcp.apphub.service.environment_type`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gcp/) already exists, I would like to push back on this request.

model/gcp/registry.yaml
michaelsafyan: For similar reasons as the other comments, I need to push back.

I concur that there is a lot of overlap between the "service" and "workload" concepts in AppHub. Were I in charge, I would have made a single definition with "SERVICE" and "WORKLOAD" as enum types that are independent of the overall structure.

While it would have been ideal to have:

```
     message AppHub {
        AppHubApplication application
        ApplicationComponentType component_type  // WORKLOAD or SERVICE
        string id
        CriticalityType criticality_type
        EnvironmentType environment_type
     } 
```

What actually exists in AppHub is:

```
   message AppHub {
       AppHubApplication application
       oneof component_type {
         AppHubService service
         AppHubWorkload workload
       }
    }
```

... with `AppHubService` and `AppHubWorkload` having substantially similar fields. This organizational structure is carried through to the [currently defined AppHub attributes](https://opentelemetry.io/docs/specs/semconv/resource/cloud-provider/gcp/apphub/) (perhaps under the assumption that these fields may take on divergent meaning or values in the two contexts in the future).

The desired rule "for every `gcp.apphub.*` attribute, there is an equivalent gcp.apphub_destination.* attribute corresponding to the destination application" ties my hands a bit in terms of improving this organization.

michaelsafyan:
Thanks for the review, @lmolkova . I know that I pushed back on a number of the requests here... please let me know if you are OK with me resolving those conversations or if you still consider these to be blockers. Thanks!

model/gcp/registry.yaml
lmolkova: One the first point, having `apphub.destination` along with `apphub_destination` would not be great anyway. 
the second point makes sense to me though.

michaelsafyan:
Friendly ping, @lmolkova . May I resolve the comment threads above?

model/gcp/registry.yaml
lmolkova: For experimental conventions in this repo, we usually try to make them right while we can.

I see a few paths forward:
- let the current approach go in, create an issue to resolve minor naming and duplication concerns prior to stability. 
- don't introduce new problems for `apphub_destination`, but create an issue for `apphub` and update it as a blocker for stability
- fix existing **experimental** apphub attributes right away

model/gcp/registry.yaml
lmolkova: thanks for the context, I've left some suggestions in the https://github.com/open-telemetry/semantic-conventions/pull/2663#discussion_r2324085371 - the tl;dr - we should fix experimental things while we can - that's why we have them experimental. 

I understand it might take time, so I'm not going to block on this, but I'd consider it to be a blocker for stability.

lmolkova (approved):
Please create an issue to track naming and duplication concern (or consider fixing them right away for `apphub`)

michaelsafyan:
Thank you so much for your understanding and for the approval stamp.

I've filed https://github.com/open-telemetry/semantic-conventions/issues/2740 to track this in OTel SemConv and will shortly file internal bugs referencing that issue with the source of these attributes to see what we can do to get these into the desired shape.

PR #2673, opened 2025-08-22
Adds in process executable version attribute #2655

thompson-tomo:
Progresses #2655

## Changes

This takes the approach of introducing a version attribute for the executable to enable the entity to be version specific.

Alternatives names would be:

* process.executable.build_id.file
* process.executable.file_version

The implemented option was chosen to make it complementary to other instances of name/version tuple.


model/process/registry.yaml
```diff
@@ -76,6 +76,12 @@ groups:
           be set to the target of `proc/[pid]/exe`. On Windows, can be set to the
           result of `GetProcessImageFileNameW`.
         examples: ['/usr/bin/cmd/otelcol']
+      - id: process.executable.version
+        type: string
+        stability: development
+        brief: >
+          The file version of the process executable.
+        examples: ['/usr/bin/cmd/otelcol']
```
KalleOlaviNiemitalo: That doesn't look like a version.

thompson-tomo: Thanks for spoting that

model/process/registry.yaml
```diff
@@ -76,6 +76,12 @@ groups:
           be set to the target of `proc/[pid]/exe`. On Windows, can be set to the
           result of `GetProcessImageFileNameW`.
         examples: ['/usr/bin/cmd/otelcol']
+      - id: process.executable.version
```
christos68k, line 79: What value would one use for this attribute for a typical ELF binary on Linux? 

Also, isn't [process.executable.build_id](https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/#process-executable-build-id-gnu) enough to cover the versioning use-case? (The GNU build ID may not always be present which is why profiling has defined a separate hashing scheme for [process.executable.build_id.htlhash](https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/#process-executable-build-id-htlhash) which is specified [here](https://opentelemetry.io/docs/specs/otel/profiles/mappings/#algorithm-for-processexecutablebuild_idhtlhash))

thompson-tomo: Ideally this attribute is set by the executable using language apis/constants I am not across elf so not aware of how to get this value from outside the executable on Linux.

No it's not. a build id differs to version with version being a human readable value which doesn't change with each build wheras the build is unique to that build. Hence why we have the differentiation in other namespaces.

Let me flip you question, how can you express the version of a dotnet executable running on Windows without introducing a new attribute?

christos68k: I've forgotten most of what I knew about Windows environments and I don't remember if file versioning is an OS-wide concept there, but on Linux, file versioning is arbitrary and not specified/exposed in ELF structures. On the filesystem front, filesystem-based versioning is also not widely supported.

This means that on Linux, there wouldn't typically be a straightforward way for a processor (outside of the executable process itself) to fill in this version value. One potential partial implementation that comes to mind is to first find the package that the executable belongs to and then query the OS package registry to retrieve the package version (and that's not a lightweight workflow but comes with lots of potential issues attached). 

A more realistic scenario is (as you mentioned) to fill in this value via the SDK, where the version is explicitly passed-in by the developer.

KalleOlaviNiemitalo: On Windows, file version numbers are typically stored in a [VERSIONINFO resource](https://learn.microsoft.com/windows/win32/menurc/versioninfo-resource) embedded in each Portable Executable file.  This structure stores both a `FILEVERSION` as four integers and a `FileVersion` as a (potentially localized) string, as well as corresponding fields for the product version.  I think the OpenTelemetry semantic conventions should not specify which of those fields is used; it won't be a problem if different instrumentations make different choices there.

KalleOlaviNiemitalo: On the other hand, if the Windows kernel or a process crashes due to an unhandled exception and Windows makes a crash dump, then IIRC the dump will contain the four-component FILEVERSION but not necessarily the string-format FileVersion.  So if the crash dump is ever analysed by something that emits OpenTelemetry signals, then it could be useful to have the same FILEVERSION value available in both the crash dump and in any other telemetry about the process before it crashed.  In that scenario though, one would also be interested in the file versions of the shared libraries loaded into the process, and `process.executable.version` does not sound like the correct attribute for those.

thompson-tomo: Agree I wouldn't want to suggest the lookup option and instead for an externally observed process we leave it out if there is no reliable way to get it.

I feel in sem conv we need to specify a preferred source to maximise compatibility between scraped metrics & self observed.

thompson-tomo: > In that scenario though, one would also be interested in the file versions of the shared libraries loaded into the process, and process.executable.version does not sound like the correct attribute for those.

That is correct, executable is for what the process is running, what you are looking for would hopefully be addressed via #2339

thompson-tomo:
Still looking for further feedback on naming of the attribute as it ties into defining the entity into the best way.

Issue #2334, opened 2025-06-08
recording of deployment events


thompson-tomo:
### Area(s)

area:deployment

### What's missing?

The ability to record that a deployment has happened and where to.

### Describe the solution you'd like

Introduction of the below attributes 

- `deployment.type` ie registry, installation
- `deployment.target.name` could be deployment.environment.name if usage loosened up
- `deployment.target.path` similar to previous

Introduce a new event which uses the new attributes plus others

### Implementation

- #2674
- #2675

### Additional context 
Consider #1404 in implementation

PR #2674, opened 2025-08-22
Introduce the deployment target #2334

thompson-tomo:
Progresses #2334

## Changes

This allows capturing of where files are being deployed to while also enabling the same attributes to be used to correlate the telemetry generated by that service.

This is particularly useful if you have multiple machines all running multiple of the same processes as the deployment.target.name would designate the machine whereas the deployment.target.path is where the files are hence providing full granularity.


lmolkova (changes requested):
Defining new semantic conventions usually means doing a lot of research on existing technologies and practices and working with subject matter experts to reach consensus.

Right now, we don’t have an active group focused on deployments, so there’s no effective way to review this PR.

If such a group existed, they would first research deployment tools and the information they expose. Then, by considering multiple UX scenarios, they would work toward a common set of conventions, prototyping them with at least two tools (for example, Kubernetes and a CI/CD pipeline).

Since no active group is working on deployments at the moment, I’m going to block this PR. You’re welcome to propose a [new project](https://github.com/open-telemetry/community/blob/main/project-management.md)  and look for others interested in deployments to continue this work.

Related to https://github.com/open-telemetry/semantic-conventions/pull/2675#discussion_r2298282067

[UPDATE] The deployment context is included in the new CI/CD SIG project proposal - https://github.com/open-telemetry/community/pull/2884, it may be a good opportunity to collaborate.

thompson-tomo:
How do you define when a pr is introducing new conventions vs enhancements to existing conventions? 

I would have seen this as an enhancement given that we are already capturing deployment tasks and this just emhances this task to also capture where it is being deployed to. But yes I also see this being part of cicd hence my examples.

joaopgrassi:
> How do you define when a pr is introducing new conventions vs enhancements to existing conventions?
> 
> I would have seen this as an enhancement given that we are already capturing deployment tasks and this just emhances this task to also capture where it is being deployed to. But yes I also see this being part of cicd hence my examples.

We are working towards making this clearer, but the gist of it is: We (maintainers) have very limited bandwidth to review PRs and the backlog of things to review/discuss just increases. 

To try to organize and prioritize things, we want to focus on a set of limited areas we want to either add as new (and drive to stabilization) or stabilize existing conventions. In this case, `deployment.` is not new, but it has been here for some time without changes/without a proper sig working on it. So this also then classifies as we can't accept work on it right now, as it's not the current priority (unless a new SIG forms and commits to drive it to stabilization). 

This is what I'm mainly working on - making what areas are the priority/focus clear to everyone.

trask:
Discussed in SIG meeting, let's draft for now until CI/CD phase 2 launches

PR #2675, opened 2025-08-22
Add deployment type #2334

thompson-tomo:
Progresses #2334

## Changes

Records the type of deployment occuring so that provisioning of a new instance running the software can be differentiated from adding the files to a store.

### DIFF
--- a/model/deployment/registry.yaml
+++ b/model/deployment/registry.yaml
@@ -31,6 +31,20 @@ groups:
         brief: >
           The status of the deployment.
         stability: development
+      - id: deployment.type
+        stability: development
+        brief: >
+          The type of deployment which is occuring.
+        type:
+          members:
+            - id: publish
+              value: 'publish'
+              brief: The software is being published to a registry/store.
+              stability: development
+            - id: provision
+              value: 'provision'
+              brief: The software is being provisioned on a machine aka the target.
+              stability: development
       - id: deployment.environment.name
         type: string
         stability: development
--- a/.chloggen/add_deployment_type.yaml
+++ b/.chloggen/add_deployment_type.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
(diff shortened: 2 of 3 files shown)


model/deployment/registry.yaml
```diff
@@ -31,6 +31,20 @@ groups:
...
+          The type of deployment which is occurring.
+        type:
+          members:
+            - id: publish
+              value: 'publish'
+              brief: The software is being published to a registry/store.
+              stability: development
+            - id: provision
+              value: 'provision'
+              brief: The software is being provisioned on a machine aka the target.
```
joaopgrassi, line 46: I'm having a hard time finding the difference between these two. Could you elaborate more? I'm also having a hard time finding how this attribute would be useful/used in reality. Can you maybe add use-case for it, or any instrumentation already using it?

thompson-tomo: A scenario is within your cicd pipeline you could publish your artifact's to a registry/store for instance docker hub if building a container or could be provisioning a container instance running the artifact's.

lmolkova: I don't believe there is industry-wide notion of deployment type that is also known and available inside ci/cd pipeline. It's not clear how existing tooling would populate this attribute. The `type` is also a very broad term that could mean a lot of things, including [k8s deployment strategy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy) or similar [Azure DevOps thing](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/deployment-jobs?view=azure-devops).

Defining semantic conventions usually means researching how things are done in multiple environments, finding commonalities, ensuring information is available and then documenting how to capture new things in these different environments. 

Currently it seems we're inventing something new that is not widely applicable. It goes against best practices documented in this repo

https://github.com/open-telemetry/semantic-conventions/blob/846028d6164593f599659f1f7117aa54a9e3bd7c/docs/how-to-write-conventions/README.md?plain=1#L127-L129

Issue #2677, opened 2025-08-24
Increase visibility to sampling relevant attributes


thompson-tomo:
Currently within span definition pages, it is not obvious which attributes should be set at creation time to assist with head sampling. As such sampling relevant attributes should be clearly identifiable to a reader by appearing under their own heading before any general attributes. This is to indicate that they are first.

Note currently they are appearing in a paragraph below the notes and as such they are blending into the notes.

PR #2678, opened 2025-08-24
[chore] Increase emphasis of sampling attributes

### LINKED ISSUE #2667: Clarify expected behaviour for `db.query.parameter.<key>`
Context: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3015#discussion_r2289930932

When creating `db.query.parameter.<key>` resource attributes, should `key` be processed before being set, or should it match the user-specified name verbatim?

For example: if a user has a query parameter named `@firstName` with a value `"John"`, what is the intended end result?

Is it literal as-is:

```text
db.query.parameter.@firstName=John
```

or lower-cased:

```text
db.query.parameter.@firstname=John
```

or even removing the `@` prefix as well:

```text
db.query.parameter.firstname=John
```

or something else?

My interpretation was that it should be used verbatim:

https://github.com/open-telemetry/semantic-conventions/blob/0089e3b1ef8f8df2663f0d6b9ea544751258c413/docs/database/database-spans.md?plain=1#L216-L217

thompson-tomo:
Fixes #2667

## Changes

All changes have been contained the jinja templates with an explaination of the changes below:

- Introduce collapsable Title for Attributes table
- Split attribute table based on if for sampling or not
- Removal of sampling paragraph
- Rename Attribute Key column to just be Key rather than Attribue
- Rename Examples column to Examples Values to indicate it is the value of the attribute
- Rename Description to Summary as it is a brief summary of the attribute

These changes also support #2607 which adds an additional attribute table


KalleOlaviNiemitalo:
This should link to <https://github.com/open-telemetry/semantic-conventions/issues/2677>, not <https://github.com/open-telemetry/semantic-conventions/issues/2667>.

If an attribute is added after the span is **created**, and thus is not available to head sampling, then does it matter whether it is added before or after the span is **started**?

thompson-tomo:
Thanks for spoting that the wrong issue was referenced.

In relation to adding attributes before/after span is started, my understanding is there would be no difference in functionality as I am not aware of the start action triggering any functionality. But not 100% sure.

KalleOlaviNiemitalo:
At <https://opentelemetry.io/docs/specs/otel/trace/api/>, I don't see any requirement for OpenTelemetry API to provide a way to create a Span without starting it immediately.  Because of that, I think the semantic conventions don't need to give any recommendations about adding attributes before vs. after starting a span.

In .NET though, the [Activity](https://learn.microsoft.com/dotnet/api/system.diagnostics.activity?view=netstandard-2.0-pp) API supports the distinction between creating a span and starting it.  One can set an [ActivityListener.ActivityStarted](https://learn.microsoft.com/dotnet/api/system.diagnostics.activitylistener.activitystarted?view=netstandard-2.0-pp) callback, which can read the attributes of the span and export them to real-time monitoring.  (Not over OTLP, which only deals with finished spans.)

thompson-tomo:
This is not a new recommendation, and in fact the wording was simply moved along with the  attributes it applies to. 

Also the schema when writing sem conv, allows for attributes to be indicated that they should be provided at creation time to support head sampling based on attributes.

Note the creating of a span with attributes and starting it via the 1 call should be the seen as creating it, adding tags & then starting.

KalleOlaviNiemitalo:
Yes I saw you did not modify YAML files in this pull request.  I just meant there is no need to additionally replace the two levels (attributes at span creation for sampling / attributes added later) with three levels (at span creation / before starting / after starting).

docs/registry/entities/vcs.md
```diff
@@ -18,7 +18,10 @@
 > :warning: This entity definition contains attributes without a role.
 > Stable Entities MUST NOT have attributes without a defined role.
 
-| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+<details open>
+<summary><b>General Attributes:</b></summary>
```
joaopgrassi: Few questions in general:

1. Why we need the collapsible thingy? 
2. That makes us have to use HTML tags here? We had problems already with the OTel website when using HTML, I wonder if we should just avoid it altogether and stick with markdown only

joaopgrassi: Also we should avoid as much as possible having to manually add stuff to markdown. If we really need these HTML tags, we should make them part of the jinja templates.

thompson-tomo: It's collapsible as when we add in past/deprecated attributes (seperate PR) that table would be collapsed by default to have focus on current attributes. Having a mismatch of formating/styling is not ideal for me.

The html issue was related to links and in fact we are already using collapsible div on a couple of pages. These tags in markdown is quite commonly used.

They are already part of the jinja template and same with splitting the tables. No manual changes to markdown were down it was all based on template changes and regerating the docs.

jsuereth:
So in generally I'm a fan of calling out sampling relevant in a far better fashion.  I think highlighting this makes a lot of sense.

What I'm not certain of is the incarnation here.  This trades off sampling-relevant being highlighted at the expense of not being able to see all attributes you'd expect on a span in one table.

Maybe, instead, we should render "Sampling Relevant" in the attribute table when we render Span semantic conventions?

thompson-tomo:
Yes I did think about that however if we merge the tables, we lose the descriptive paragraph which provide context about what it means.

If we merge the tables what could the heading be? The only option I can think of is "Value Captured" with options of "Span Creation" & "Any time".

joaopgrassi:
Why not just add a new column to the existing table with a little badge? Or just add a badge under the attribute name, in case it's relevant for sampling?  For ex

image


The last column could also be just a "other" column that we just add stuff about it, like sampling relevant is somewhat metadata about the attribute.

> Yes I did think about that however if we merge the tables, we lose the descriptive paragraph which provide context about what it means.
> 
> If we merge the tables what could the heading be? The only option I can think of is "Value Captured" with options of "Span Creation" & "Any time".

what descriptive paragraph you mean?

thompson-tomo:
@joaopgrassi as stated above what could the heading be? That is where the difficulty in extending the existing table lies and using text or a badge is just styling.

In the example you provided, The heading should not be also the value of the cell but rather true/false used. In the case here most of the registry would then then end up with Sampling Relevant = false as sampling relevant is sparsely set and that would be conveying the wrong message.

> what descriptive paragraph you mean?

The below paragraph is taken from the existing implementation and was brought across without any changes.

> The following attributes can be important for making sampling decisions and SHOULD be provided <b>at span creation time</b> (if provided at all).

That paragraph is why the "value captured" heading seems to be the best fit as the default value would be any-time unless it has been marked as sampling relevant in which case the value becomes "Span Creation". That matches the paragraph and removes the need for the paragraph as it clearly tells the implementor when they should be capturing this value.

JamesNK:
I think you could add the `sampling relevent` badge under the requirement level column text. It seems the most related column.

Then on the requirement level page - https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/ - you could talk about how to handle values that are relevent to sampling.

joaopgrassi:
> I think you could add the `sampling relevent` badge under the requirement level column text. It seems the most related column.
> 
> Then on the requirement level page - [opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) - you could talk about how to handle values that are relevent to sampling.

Yes that also sounds like a good option to me. 

About how to name the column, if we decide to go that route. I think that's easier than introducing all the changes the current PR has. Also, we could like I said just either leave the column without a name and just use it as a bag of stuff. Or call it metadata. 

and 

> in the case here most of the registry would then then end up with Sampling Relevant = false as sampling relevant is sparsely set and that would be conveying the wrong message.

Of course for the ones that are not sampling relevant, we would just not render anything. That's how it is done today. We don't say "the attributes not part of this list are not sampling relevant". So if we decide to add another column, not rendering anything for the ones not relevant achieves the same status quo of today.

thompson-tomo:
yes it is the most relevant of the existing columns but is better suited as a separate column to preserve that each column has a single purpose which in the case of requirement level it is describing the requirement of implementing it.

> Also, we could like I said just either leave the column without a name and just use it as a bag of stuff. Or call it metadata.

With this approach you have no context or awareness of what it actually means hence is confusing and not informative.

> So if we decide to add another column, not rendering anything for the ones not relevant achieves the same status quo of today.

When i look at a table with empty cells it conveys incomplete information. Hence I have gone and added it as seperate column "Value Captured" as this provides a clear definition of what this column is conveying to both an implementator and a user. There is a docment describing the value captured reasons. With this the implementator knows when they should record it and the reader knows when the value should have been captured which becomes relevant for long running spans.

docs/cicd/cicd-metrics.md
```diff
@@ -49,13 +49,15 @@ This metric is [recommended][MetricRecommended].
...
-| [`cicd.pipeline.run.state`](/docs/registry/attributes/cicd.md) | string | The pipeline run goes through these states during its lifecycle. | `pending`; `executing`; `finalizing` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`cicd.pipeline.result`](/docs/registry/attributes/cicd.md) | string | The result of a pipeline run. | `success`; `failure`; `timeout`; `skipped` | `Conditionally Required` [1] | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`error.type`](/docs/registry/attributes/error.md) | string | Describes a class of error the operation ended with. [2] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | `Conditionally Required` If and only if the pipeline run failed. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| Key | Type | Summary | Example Values | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability | Value Captured |
```
kamphaus: Does it make sense to have the `Value Captured` column when talking about metrics?

joaopgrassi:
> yes it is the most relevant of the existing columns but is better suited as a separate column to preserve that each column has a single purpose which in the case of requirement level it is describing the requirement of implementing it.

I'm not convinced. Now, if you look at your changes I have the following issues with it:

- "Value Captured": Maybe just me but this is definitely not clear what it even means. It's not just the value that is captured, it's the entire attribute. I think this just adds to more confusion. 
- The new column is at the very end, hidden because the page width does not fit the entire table. So it is hidden anyway now and out of sight :D 

Again, I like the intention of making very visible which attributes are sampling relevant, but this isn't it IMHO yet. I still believe that if we add a badge to any existing column is much better in making it "in your face" and "hard to miss" that certain attribute is relevant for sampling. I'm also not sure we ALSO need to add anything extra to the "not sampling relevant" ones, as we don't do that today and I don't think it's a problem.

AND: We don't need to explain what/when. If we just say "sampling relevant" it should be clear already that it means on span creation, as that is when sampling is done. I feel like encoding this as part of each table (via "SPan creation/Any time") just adds unnecessary repetition. If we lack a guidance/explanation on what sampling relevant means for attributes, then I feel this should be added in some other place, and not in each table.

docs/cicd/cicd-metrics.md
thompson-tomo: Title has been changed to capture scope which will enable it to be used to in the future also express scopes such as instrumentation, resource etc.

thompson-tomo:
Column Title has been changed to address feedback and decouple it from the attribute and instead have it focused on describing the attribute at a whole.

I agree the tables are not scaling well especially if the example value is a really long value. Those tables are one of the things which I was hoping to address with my prototype of documentation changes but that went nowhere.

> We don't need to explain what/when. If we just say "sampling relevant" it should be clear already that it means on span creation, as that is when sampling is done.

You are assuming alot which we should avoid doing. Don't forget tail sampling doesn't run on create but on completion. ;) There is a document explaining the values which I couldn't link as it is not possible to do absolute link and instead the doc first needs to be published.

This approach also support instrumentation scope attributes which ideally we were using more.

joaopgrassi:
> You are assuming alot which we should avoid doing. Don't forget tail sampling doesn't run on create but on completion. ;) There is a document explaining the values which I couldn't link as it is not possible to do absolute link and instead the doc first needs to be published.

I know, but I don't think we need to explain all here - it should be a concept that Semconv links to the API/SDK specs.

thompson-tomo:
@joaopgrassi i still see the benefit in the capture column scope as it would enable richer, more complete docs.

With that being said to progress this PR I have renamed the column to "role" with a value of "head sampling" when appropiate. This ensures that each column retains a clear single purpose.

joaopgrassi:
I'm not convinced we need a new column. I'd still in the opinion that a simple badge as mentioned before is easier/simpler. I don't think we today have a problem where "sampling relevant" is lacking explanation or people are confused. We never had problems or people coming to complain. So, I believe until we have that, this suggestions seems the best compromise: https://github.com/open-telemetry/semantic-conventions/pull/2678#issuecomment-3231609462

thompson-tomo:
I don't see this discussion being about seperate columns being any different to https://github.com/open-telemetry/semantic-conventions/pull/2970 where it was decided to go seperate. I have tweaked the jinja so only spans gain the additional column.

trask:
Closing as there does not appear to be interest from the necessary folks in order to move this forward at this time.

PR #2684, opened 2025-08-26
[chore] Clarify behaviour for db.query.parameter.<key>

### LINKED ISSUE #2667: Clarify expected behaviour for `db.query.parameter.<key>`

martincostello:
Fixes #2667

## Changes

Clarify that the casing of any named parameters should be left as-is.

Rather than introduce a new example, changes the casing of the existing example.

* [x] ~~[CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.~~
* [ ] ~~Links to the prototypes or existing instrumentations (when adding or changing conventions)~~

### DIFF
--- a/docs/database/cassandra.md
+++ b/docs/database/cassandra.md
@@ -131,6 +131,9 @@ then `<key>` SHOULD be the 0-based index.
 `db.query.parameter.<key>` SHOULD match
 up with the parameterized placeholders present in `db.query.text`.
 
+`db.query.parameter.<key>` SHOULD match the name of the parameterized
+placeholders verbatim, such as matching the exact casing of names.
+
 `db.query.parameter.<key>` SHOULD NOT be captured on batch operations.
 
 Examples:
@@ -138,8 +141,8 @@ Examples:
 - For a query `SELECT * FROM users where username =  %s` with the parameter `"jdoe"`,
   the attribute `db.query.parameter.0` SHOULD be set to `"jdoe"`.
 
-- For a query `"SELECT * FROM users WHERE username = %(username)s;` with parameter
-  `username = "jdoe"`, the attribute `db.query.parameter.username` SHOULD be set to `"jdoe"`.
+- For a query `"SELECT * FROM users WHERE username = %(userName)s;` with parameter
+  `userName = "jdoe"`, the attribute `db.query.parameter.userName` SHOULD be set to `"jdoe"`.
 
 The following attributes can be important for making sampling decisions
 and SHOULD be provided **at span creation time** (if provided at all):
--- a/docs/database/cassandra.md
+++ b/docs/database/cassandra.md
@@ -131,8 +131,8 @@ then `<key>` SHOULD be the 0-based index.
 `db.query.parameter.<key>` SHOULD match
 up with the parameterized placeholders present in `db.query.text`.
 
-`db.query.parameter.<key>` SHOULD match the name of the parameterized
-placeholders verbatim, such as matching the exact casing of names.
+It is RECOMMENDED to capture the value as provided by the application
+without attempting to do any case normalization.
 
 `db.query.parameter.<key>` SHOULD NOT be captured on batch operations.
 
--- a/model/database/registry.yaml
+++ b/model/database/registry.yaml
@@ -91,6 +91,9 @@ groups:
           `db.query.parameter.<key>` SHOULD match
           up with the parameterized placeholders present in `db.query.text`.
 
+          `db.query.parameter.<key>` SHOULD match the name of the parameterized
+          placeholders verbatim, such as matching the exact casing of names.
+
           `db.query.parameter.<key>` SHOULD NOT be captured on batch operations.
 
           Examples:
@@ -98,8 +101,8 @@ groups:
           - For a query `SELECT * FROM users where username =  %s` with the parameter `"jdoe"`,
             the attribute `db.query.parameter.0` SHOULD be set to `"jdoe"`.
 
-          - For a query `"SELECT * FROM users WHERE username = %(username)s;` with parameter
-            `username = "jdoe"`, the attribute `db.query.parameter.username` SHOULD be set to `"jdoe"`.
+          - For a query `"SELECT * FROM users WHERE username = %(userName)s;` with parameter
+            `userName = "jdoe"`, the attribute `db.query.parameter.userName` SHOULD be set to `"jdoe"`.
 
         examples: ["someval", "55"]
       - id: db.query.summary
--- a/model/database/registry.yaml
+++ b/model/database/registry.yaml
@@ -91,8 +91,8 @@ groups:
           `db.query.parameter.<key>` SHOULD match
           up with the parameterized placeholders present in `db.query.text`.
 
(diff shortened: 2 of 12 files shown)


martincostello:
> Changelog can be added by https://github.com/open-telemetry/semantic-conventions/tree/56669607fccf65338cc2002935dc06c8a3e27e5b/.chloggen

I removed it as this seemed to come under "editorial" rather than actual changes, so marked as `[chore]`.

model/database/registry.yaml
```diff
@@ -91,15 +91,18 @@ groups:
           `db.query.parameter.<key>` SHOULD match
           up with the parameterized placeholders present in `db.query.text`.
 
+          `db.query.parameter.<key>` SHOULD match the name of the parameterized
+          placeholders verbatim, such as matching the exact casing of names.
```
KalleOlaviNiemitalo: This is still not entirely clear on whether `@` should be included for SQL Server.

Is it possible to have an example with `@` included and render that on the SQL Server page only?

martincostello: > This is still not entirely clear on whether @ should be included for SQL Server.

I thought that was covered by `verbatim`.

I can look to see if it's easy add a SQL Server-specific example. Depends on how easy it is to interleave with the text being rendered from the registry.

martincostello: I can't see an obvious way to do this without it being added to every document generated from the registry (or having to duplicate the whole entry). If there is an easy/maintainable way to do this within the custom templating, then if you can point me to it I can make the relevant change.

KalleOlaviNiemitalo: The "verbatim" is somewhat countered by the example `SELECT * FROM users WHERE username = %(userName)s;`, which looks like some SQL dialect unfamiliar to me, being paired with `db.query.parameter.username` rather than `db.query.parameter.%(userName)s`.

martincostello: I don't know what the dialect is either, but the existing example implied to me it wasn't the name, but some escaping syntax.

matt-hensley: [That matches the Python string formatting syntax](https://docs.python.org/2/library/stdtypes.html#string-formatting-operations). Some database drivers like [psycopg](https://www.psycopg.org/psycopg3/docs/basic/params.html) use the same style for referencing named parameters in queries

model/database/registry.yaml
```diff
@@ -91,15 +91,18 @@ groups:
...
 
           Examples:
 
           - For a query `SELECT * FROM users where username =  %s` with the parameter `"jdoe"`,
             the attribute `db.query.parameter.0` SHOULD be set to `"jdoe"`.
 
-          - For a query `"SELECT * FROM users WHERE username = %(username)s;` with parameter
-            `username = "jdoe"`, the attribute `db.query.parameter.username` SHOULD be set to `"jdoe"`.
+          - For a query `"SELECT * FROM users WHERE username = %(userName)s;` with parameter
+            `userName = "jdoe"`, the attribute `db.query.parameter.userName` SHOULD be set to `"jdoe"`.
```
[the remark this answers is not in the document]
jsuereth: I'm not sure why this was marked as resolved.

I still see `username` and `userName` as a change here, and I'm not sure I udnerstand why...

docs/database/cassandra.md
```diff
@@ -131,15 +131,18 @@ then `<key>` SHOULD be the 0-based index.
...
 `db.query.parameter.<key>` SHOULD NOT be captured on batch operations.
 
 Examples:
 
 - For a query `SELECT * FROM users where username =  %s` with the parameter `"jdoe"`,
   the attribute `db.query.parameter.0` SHOULD be set to `"jdoe"`.
 
-- For a query `"SELECT * FROM users WHERE username = %(username)s;` with parameter
-  `username = "jdoe"`, the attribute `db.query.parameter.username` SHOULD be set to `"jdoe"`.
+- For a query `"SELECT * FROM users WHERE username = %(userName)s;` with parameter
```
jsuereth, line 144: This makes sense to me as it's `userName` in both places, I think the comment below from copilot was accurate, I think you missed migrating the query to show `userName` in both places.

model/database/registry.yaml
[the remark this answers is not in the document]
martincostello: I resolved it because one is the name of an item in the imaginary database table, and the other is a user supplied parameter - they don't _need_ to match.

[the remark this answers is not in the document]
jsuereth: Ah, ignore me, looking at this again, I see both are updated.

docs/database/cassandra.md
```diff
@@ -131,15 +131,18 @@ then `<key>` SHOULD be the 0-based index.
 `db.query.parameter.<key>` SHOULD match
 up with the parameterized placeholders present in `db.query.text`.
 
+`db.query.parameter.<key>` SHOULD match the name of the parameterized
+placeholders verbatim, such as matching the exact casing of names.
+
```
matt-hensley: A number of the db attributes already have this language:

> It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.

Consistency would be nice if the footnotes are equivalent, just worded differently.

martincostello: Nice - I didn't see that.

In that case I'll just copy that.

model/database/registry.yaml
lmolkova: it's not possible to have one of the examples (that are part of the note) specific to oracle/mssql/etc unless you want to copy-paste the whole note which is quite big.

https://github.com/open-telemetry/semantic-conventions/blob/33f76c399efc320eefb1b4e83d8d7c4d37e3d282/model/database/registry.yaml#L87-L105

we can change the example (or add another one) to use `:userName` of `@userName` for all databases - it does  not work for all of them regardless even today.

joaopgrassi (approved):

lmolkova:
@martincostello @matt-hensley @KalleOlaviNiemitalo I *think* discussions can be resolved now, could you please check?

It seems it's ready to go otherwise.

martincostello:
AFAIC it's ready for merge.

matt-hensley (approved):

matt-hensley:
LGTM 👍

model/database/registry.yaml
lmolkova: Let's resolve this discussion and if behavior wrt @ is not clear, we can follow up and add more examples

PR #2685, opened 2025-08-26
feat: add role attribute to agent spans

91pavan:
Fixes #

## Changes

Add a role attribute to GenAI agent spans to help categorize the type of "agent" for various use cases.
Note: The agent's assigned role may differ from its name.


model/gen-ai/spans.yaml
```diff
@@ -240,6 +240,9 @@ groups:
       - ref: gen_ai.agent.description
         requirement_level:
           conditionally_required: If provided by the application.
+      - ref: gen_ai.agent.role
```
gyliu513, line 243: Do you want to add it to `span.gen_ai.invoke_agent.client`?

Also please add a changelog as well, thanks

alexmojaki:
I would like to see a real life example where id, name, role, and description all exist and are different.

91pavan:
Here's a concrete example: https://docs.google.com/document/d/1CvKybmKGAZmDl8kLJdAngajMJgRcDEYjdORUAtZft9Y/edit?usp=sharing

Quick summary: 

`id: "asst_5j66UpCpwteGg4YSxUnt7lPY"` (Generated agent ID)
`name: "Travel Booking Helper"` (Human-readable name)
`role: "Customer Support"` (agent's functional classification)
`description: "Flight bookings..."` (Detailed capabilities)

The role attribute serves a different purpose than name/description - it's used for functional grouping (routing queries to "Customer Support" vs "Sales Assistant" agents), metrics aggregation, and access controls. This follows the same pattern as existing semantic conventions like `service.name` vs `service.instance.id`.

alexmojaki:
Your example has:

```python
# Creating agents via OpenAI Assistants API
assistant = client.beta.assistants.create(
    name="Travel Booking Helper",
    description="Assists customers with flight bookings, hotel reservations, and travel planning. Has access to booking APIs and can process payment information.",
    instructions="You are a helpful travel agent...",
    model="gpt-4-turbo"
)
```

`role: "Customer Support"` doesn't appear in there, so it would be impossible for any OpenAI instrumentation to capture it. You even say:

> Source: Our application's agent classification system

So in this example, `role` is something application-specific, I don't see why it needs a semantic convention.

91pavan:
Sorry, it might have been such a good example :) apologies.

IMO, agent attributes like `name` and `description` are also application-supplied in many cases, not something instrumentation can always auto-capture. The `role` attribute follows the same pattern: it’s optional enrichment that an app can add if it defines such a concept. The convention just provides a standard place to record it, rather than leaving each library to invent its own key.

With regards to how the instrumentation libraries can capture agentic telemetry, I don't know if there's a standard way to reliably capture them (since agents can be defined as methods/classes within the app) and can be implemented in a variety of different frameworks (or underlying SDKs). We're looking to add better support for this that all instrumentations can make use of as part of GenAI utils in contrib SDK: https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/util/opentelemetry-util-genai (PRs will be added soon).

alexmojaki:
That doesn't seem like a convincing argument, -1 from me.

91pavan:
@alexmojaki I appreciate feedback, but could you please provide specific concerns about the proposed attribute rather than just a -1?
If you have technical objections to the `role` attribute design, I'm happy to address them. 
The core question is whether there's value in standardizing an optional `role` field that applications can populate when they have this concept, rather than each instrumentation library inventing their own keys. I believe there is, following the same pattern as other optional agent enrichment attributes like `name` and `description` in the spec.

91pavan:
You're absolutely right that role wouldn't be auto-captured by OpenAI instrumentation. The role attribute is designed as optional application enrichment, similar to how `service.name` or `user.id` work in other semantic conventions namespaces.

I believe agent frameworks like LangChain, CrewAI, or AutoGen often DO have role concepts that instrumentation could capture. Adding this to the spec provides a standard place to put it rather than letting this get fragmented.

alexmojaki:
> I believe agent frameworks like LangChain, CrewAI, or AutoGen often DO have role concepts that instrumentation could capture.

That sounds like a source of helpful examples.

91pavan:
Few Examples - 

https://docs.agno.com/examples/getting-started/agent-team - Agno

```
web_agent = Agent(..., role="Handle web search requests and general research")
news_agent = Agent(..., role="Handle news requests and current events analysis")
```

https://docs.crewai.com/en/concepts/agents - CrewAI

```
agent = Agent(
  role='Research Analyst',
  goal='Provide up-to-date market analysis',
  backstory='An expert analyst with a keen eye for market trends.',
```

lmolkova:
so we have 

- ADK which has `name` and `description` - https://google.github.io/adk-docs/get-started/quickstart/#agentpy
- Langgraph that only has `name` (?) - https://langchain-ai.github.io/langgraph/reference/agents/?h=create_react_agent#langgraph.prebuilt.chat_agent_executor.AgentState
- crewai which has `role` and `goal`
- agno that has `name` and `role` 
- openai assistants use `name` and `description` - https://platform.openai.com/docs/api-reference/assistants/createAssistant
- AWS bedrock agents have `name` and `description` https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_CreateAgent.html#bedrock-agent_CreateAgent-request-agentCollaboration, also and some [CollaboratorType enum](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_bedrock_alpha/AgentCollaboration.html)

It seems `role`, `goal` and `description` are the same concept - extended info about the agent as a free form string. And we already have `gen_ai.agent.description` for it.

I think description or goal make more sense - `role` sounds like an enum rather than free-form text.

alexmojaki:
> Few Examples -
> 
> https://docs.agno.com/examples/getting-started/agent-team - Agno
> 
> ```
> web_agent = Agent(..., role="Handle web search requests and general research")
> news_agent = Agent(..., role="Handle news requests and current events analysis")
> ```

If `role` goes in `gen_ai.agent.role`, then what's used for `gen_ai.agent.description`? I'd rather populate `description` than `role`.

> https://docs.crewai.com/en/concepts/agents - CrewAI
> 
> ```
> agent = Agent(
>   role='Research Analyst',
>   goal='Provide up-to-date market analysis',
>   backstory='An expert analyst with a keen eye for market trends.',
> ```

Similarly, what's the `gen_ai.agent.name` here? `role` is actually the best fit for that.

91pavan:
@alexmojaki agreed. Frameworks like CrewAI and Agno don’t stick to standard fields such as name or description. 

IMO, a good approach would be to use `name` as the agent label, `description` for the detailed explanation of what the agent does, and reserve `role` only when there’s a consistent function taxonomy available.

alexmojaki:
> reserve role only when there’s a consistent function taxonomy available.

it's not clear to me what that means or when it applies. Again, do you have a concrete example?

brunobat:
Please consider that `role` is different from `capability`. An agent can have multiple capabilities, however, on a particular context it might only have a few `roles`. This is similar to `role` in the context of Java security.

Issue #2627, opened 2025-08-13
Deprecated: rename rego policy should check if the attribute type is the same


lmolkova:
We can and should have an automated check that prevents changes like this: https://github.com/open-telemetry/semantic-conventions/pull/2605

lmolkova:
fixed in #2688

PR #2688, opened 2025-08-27
Check deprecated.rename_to type compatibility and more

### LINKED ISSUE #2630: Deprecated: renamed policy should check if member exists
_Originally posted by @lmolkova in https://github.com/open-telemetry/semantic-conventions/pull/2614#pullrequestreview-3120763108_

lmolkova:
Fixes #2630, #2627

Add policies that check:
 - `renamed_to` type is the same:
     - OK: same standard type to same standard type,  enum <-> enum of the same type, string enum <-> string, int enum <-> int, 
     - Not OK: string <-> int and such, string <-> int enum, string enum <-> int enum
 - `renamed_to` enum member exists on the same enum and is not deprecated 

Fixes existing violations, brings back some dropped deprecations - context https://cloud-native.slack.com/archives/C041APFBYQP/p1756196369228549?thread_ts=1756165118.156879&cid=C041APFBYQP

The reason to keep type the same when renaming is the following: the goal is to apply rename and other transformations in the collector (and make [schema transformations](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/schemas#collector-assisted-schema-transformation) more real). `Rename` transformation would literally rename attribute. We could introduce more complex transformations that could change type, etc, but they are not supported yet. So `rename` deprecation can only be used with pure renames at least for the time being.


xrmx (approved):

maryliag (approved):

maryliag:
missing changelog, otherwise LGTM

model/gen-ai/deprecated/registry-deprecated.yaml
```diff
@@ -129,8 +135,8 @@ groups:
               value: "xai"
               brief: 'xAI'
               deprecated:
-                reason: renamed
-                renamed_to: "x_ai"
+                reason: uncategorized
+                note: Use `x_ai` on the `gen_ai.provider.name` attribute
```
trask: this one feels a little odd, is there no replacement in this (deprecated) enum? would it make sense to "undeprecate" this value (within the still deprecated enum)?

Issue #2690, opened 2025-08-28
[editorial] Problems publishing v1.37.0+ to the OTel.io website


chalin:
### Area(s)

docs

### What's the problem?

The OTel.io **website build is failing** for semconv v1.37.0. For details, see:

- https://github.com/open-telemetry/opentelemetry.io/pull/7638#pullrequestreview-3166568854

### Detailed list of problems

- [ ] New JSON schema files are now contained in various spec folders. This is a new feature that the OTel.io website will need to adapt to. See comments further below for a discussion about how we want to resolve this.
- [ ] New `non-normative` **subfolders** were introduced, e.g., https://github.com/open-telemetry/semantic-conventions/tree/main/docs/gen-ai/non-normative. In the past, all non-normative pages were put under https://github.com/open-telemetry/semantic-conventions/tree/main/docs/non-normative. We'll need to discuss how to recover from this.
- [x] The hash target `/docs/general/metrics/#metric-requirement-levels`, used in many places, doesn't exist. @trask any idea why the link check didn't pick up on this? For example, visit https://github.com/open-telemetry/semantic-conventions/blob/main/docs/hardware/gpu.md#metric-hwgpuio, and click on the [recommended](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#metric-requirement-levels) link. It visits the page, but the hash target doesn't exist.
  - [x] I _think_ that the use of that invalid link should be replaced by https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attribute-requirement-level.md#recommended.
[Update]: fix - https://github.com/open-telemetry/semantic-conventions/pull/2691 
- [ ] ~Inline images are not being rendered correctly (they would technically need to be marked `.img-initial`). For context, see https://github.com/open-telemetry/semantic-conventions/pull/2970#issuecomment-3467668552, or just see the badge images in https://opentelemetry.io/docs/specs/semconv/registry/entities/browser/, for example.~ See https://github.com/open-telemetry/opentelemetry.io/issues/8288

### Describe the solution you'd like

TBC

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

[...]

PR #2691, opened 2025-08-29
Fix metrics requirement level link

lmolkova:
One of the problems in https://github.com/open-telemetry/semantic-conventions/issues/2690#issuecomment-3235276173

why link-check didn't detect it - https://github.com/open-telemetry/semantic-conventions/issues/2690#issuecomment-3235455543


chalin (approved):
👍🏻

Issue #2426, opened 2025-06-24
Add missing docs & cleanup


thompson-tomo:
### Area(s)

docs

### What's the problem?

Inconsistent navigation experience

### Describe the solution you'd like

Readme's are generated for

- App
- Browser
- Code
- Dns
- Graphql
- OTEL
- Url

Following pages are moved
- cli-spans to be under app
- Events to be under browser
- Mobile events to be under app

Following pages are removed

- Code
- Url

PR #2694, opened 2025-08-29
Consolidate App docs into App domain

thompson-tomo:
Progresses #2426 

## Changes

Consolidates cli and mobile apps to be grouped with the other app documentation to form the app domain.

This follows what is done with messaging etc.


thompson-tomo:
@lmolkova the focus is on Grouping the app domain docs into a single location just like we have with other domains ie messaging, db. As per the pr this has been split out from, I ideally would want a mobile app, cli & web app doc but have been advised to do it gradually hence the 2 step approach. Would you rather I do it as 1 step and introduce the app type docs as part of this PR?

lmolkova:
I would suggest to focus ion substantial work that moves semantic conventions forward. Having dozens of tiny PRs without clear benefit recognized by other members of the community is not helpful.

thompson-tomo:
there you are i have reworked the changes to increase the benefit. As a client side app developer you now have a single page to go to where the events related to core app handling are defined and as such you no longer need to use multiple pages and increases exposure to the lifecycle events.

lmolkova:
while I see some analogy between mobile applications and CLI programs, it's not particularly helpful to have a page that list them both - there is little to no intersection today. It also moves mobile events into generic app page. If you look carefully, some pieces of markdown were also lost - the whole context that device things are specific to mobile was lost. So, while doing some minor benefit, this PR does some harm. 

Some of the ways to make substantial changes would be to 
- Consider if CLI instrumentation should make use of some app attributes. Should it use app or service? When and what would be useful?
- Go and prototype changes in python
- Come back with some analysis on what's missing today and how to fix it
- Get consensus with mobile / client app folks

Writing semconv is a lot of work, please help us do the work, instead of going through the motions

thompson-tomo:
> there is little to no intersection today.

They should be sharing a span so have added the definition of such span.

> If you look carefully, some pieces of markdown were also lost - the whole context that device things are specific to mobile was lost. 

I had left it out as I didn't see it any different to the app jank events but have added it in.

> Consider if CLI instrumentation should make use of some app attributes. Should it use app or service? When and what would be useful?

Both are useful and should be used.

> Writing semconv is a lot of work, please help us do the work, instead of going through the motions

I am trying to help and I am also aware that alot of people don't pay attention to finer points especially when it is formatting hence willing to do those quick repeative housekeeping.

PR #2695, opened 2025-08-29
[chore] Relocate Misplaced session documents from General

### LINKED ISSUE #2426: Add missing docs & cleanup

thompson-tomo:
Closes #2426

## Changes

Relocate the session document to be located in its own group rather than buried in general. This is because the page is describing the events for this concept of sessions.

### DIFF
--- a/docs/session/session-events.md
+++ b/docs/session/session-events.md
@@ -2,7 +2,7 @@
 linkTitle: Events
 --->
 
-# Semantic conventions for session
+# Semantic conventions for session in events
 
 **Status**: [Development][DocumentStatus]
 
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -41,7 +42,6 @@ Semantic Conventions by signals:
 * [Events](general/events.md): Semantic Conventions for event data.
 * [Logs](general/logs.md): Semantic Conventions for logs data.
 * [Metrics](general/metrics.md): Semantic Conventions for metrics.
-* [Profiles](general/profiles.md): Semantic Conventions for profiles.
```
florianl, line 43: Profiles is a [recognized](https://github.com/open-telemetry/opentelemetry-proto/tree/main/opentelemetry/proto) [signal](https://opentelemetry.io/docs/concepts/signals/), just like Logs, Metrics and Traces. Therefore, I think, it should stay at this point.

florianl: FYI: @open-telemetry/profiling-approvers

thompson-tomo: yes it may be a signal but the document being linked to is an attribute table. With this change it follows the precedence set by events. Profiling doesn't have a document of a similar nature to the others in that list. Note also Resources are also missing from that list.

florianl: If the concern is, that profiles link directly to its used attributes, rather than a general guideline for the signal, then the general guideline should be introduced and not the profiles signal moved to a general purpose section.

Looking at the logs signal - it also directly links to the attribute table.

thompson-tomo: Good catch with the logs doc will look at it.

You are more than welcome to introduce a profile guideline doc via a separate PR.

florianl (dismissed):
The [Profiling signal](https://opentelemetry.io/docs/concepts/signals/) should not get moved to the general section and stay with the other OTel signals.

docs/README.md
aalexand: +1 to @florianl - even though newer, profiles is the same top-level signal as logs and metrics, so it should be listed here at the same level / place as the other signals.

thompson-tomo: it is about providing a consistent experience for user and being able to meet expectations. 

What is interesting is that there isn't even a definition of the profiling concept in the general section of the website. Also there is no ability to define a profile signal in semconv which already makes it different to metrics, traces, events. Have put it back to avoid blocking this pr.

docs/session/session-events.md
```diff
@@ -19,25 +19,6 @@ When a session reaches end of life, typically due to user inactivity or session
 will be assigned. The previous session identifier may be provided by the instrumentation so that telemetry
 backends can link the two sessions (see [Session Start Event](#event-sessionstart) below).
 
-## Attributes
-
```
joaopgrassi, line 23: Shouldn't we move some of this text above into the new README.md? It seems general enough for all signals on session, not just for events.

thompson-tomo: Sure, have moved a large chunk of it & added some to the scents.

model/session/events.yaml
```diff
@@ -8,9 +8,13 @@ groups:
     note: >
       For instrumentation that tracks user behavior during user sessions, a `session.start` event MUST be emitted
       every time a session is created. When a new session is created as a continuation of a prior session,
-      the `session.previous_id` SHOULD be included in the event. The values of `session.id` and `session.previous_id`
+      the `session.previous_id` SHOULD be included in the event so that telemetry backends may link the two sessions.
```
joaopgrassi:
```suggestion
      the `session.previous_id` SHOULD be included in the event so that they can be linked together.
```

Maybe? I'd refer from mentioning backends in conventions in general?

thompson-tomo: Don't have an opinion either way. Note I did copy that phrase from the old page.

Howabout "the new session can be associated with the old session." This way we avoid talking about the tool and people won't think we are talking about links which are an OTEL concept?

thompson-tomo: Change has been made.

docs/session/session-events.md
```diff
@@ -1,42 +1,12 @@
 <!--- Hugo front matter used to generate the website version of this page:
-linkTitle: Session
+linkTitle: Events
 --->
 
 # Semantic conventions for session
```
joaopgrassi:
```suggestion
# Semantic conventions for session in events
```

docs/session/session-events.md
```diff
@@ -55,7 +25,10 @@ The event name MUST be `session.start`.
 
 Indicates that a new session has been started, optionally linking to the prior session.
 
-For instrumentation that tracks user behavior during user sessions, a `session.start` event MUST be emitted every time a session is created. When a new session is created as a continuation of a prior session, the `session.previous_id` SHOULD be included in the event. The values of `session.id` and `session.previous_id` MUST be different.
+For instrumentation that tracks user behavior during user sessions, a `session.start` event MUST be emitted every time a session is created. When a new session is created as a continuation of a prior session, the `session.previous_id` SHOULD be included in the event so that the new session can be associated with the old session.
+
```
joaopgrassi: I'm not sure we also need this part here 

> When a new session is created as a continuation of a prior session, the `session.previous_id` SHOULD be included in the event so that the new session can be associated with the old session.

I think it kinda made sense having there before was due to the way the page was structured. But now that it is inside also the "Event: session.start" it's really just duplicated. So it'd be ok in removing it and keeping just the note.

joaopgrassi:
All: The initial changes made to profiling are reverted. This is now simply fixing the session page which was in /docs/general and now are under docs/session.

docs/session/session-events.md
thompson-tomo: Changes have been made to decrease duplication of the content within notes/briefs.

thompson-tomo:
Keep open

trask:

Issue #1501, opened 2024-10-21
`process`: change attributes named as `*_type` to `*.type`


braydonk:
### Area(s)

area:process

### Is your change request related to a problem? Please describe.

Based on discussion from the Semantic Conventions WG Meeting Oct 21, 2024, there may be new guidance coming related to when to use `_` or `.` in naming. The following attributes use `_` in a way that is not recommended:

* `process.context_switch_type`
* `process.paging.fault_type`

### Describe the solution you'd like

In advance of the guidance going out, we can rename these attributes to:

* `process.context_switch.type`
* `process.paging.fault.type`

braydonk:
Currently blocked on #1462

ChrsMark:
@braydonk from what I can see there is no need to rename `process.paging.fault_type` as part of this issue since it will be changed by https://github.com/open-telemetry/semantic-conventions/issues/1803, right? 

I've sent a PR for doing the change on `process.context_switch_type`: https://github.com/open-telemetry/semantic-conventions/pull/2855

PR #2704, opened 2025-08-30
Rename attribute context switch type to follow conventions #1501

thompson-tomo:
Progresses #1501

## Changes

The `process.context_switch_type` attribute has now become `process.context_switch.type` to follow conventions.


model/process/registry.yaml
```diff
@@ -217,15 +217,17 @@ groups:
...
         brief: "Specifies whether the context switches for this data point were voluntary or involuntary."
         type:
           members:
             - id: voluntary
               value: 'voluntary'
+              brief: Voluntary
               stability: development
             - id: involuntary
               value: 'involuntary'
+              brief: Involuntary
```
mx-psi: From discussion on the 2025-09-11 System semconv working group, we should not add briefs to the enum values based on https://github.com/open-telemetry/semantic-conventions/issues/2555#issuecomment-3201697007. Could you remove that change from this PR so we can move forward?

model/process/metrics.yaml
```diff
@@ -135,6 +135,7 @@ groups:
     unit: "{context_switch}"
     attributes:
       - ref: process.context_switch_type
```
ChrsMark: I wonder if this attribute should be removed.

model/process/registry.yaml
thompson-tomo: Have brought across more informative briefs.

model/process/metrics.yaml
jsuereth: Yeah - we should remove this, as you have the "renamed" deprecation on the ref already.

thompson-tomo: done

thompson-tomo:
@braydonk i attempted to use the definition from man page https://man7.org/linux/man-pages/man1/time.1.html but I have just gone and tweaked the wording further to improve how it reads.

braydonk:
1. `time(1)` is not the correct place to source this information. In Linux, the context switch counts are sourced from `/proc/[pid]/status`, which contains the voluntary and involuntary context switch counts for the process. Both of those can happen for many reasons other than what is laid out in the briefs, and there is enough nuance in both that it is too much information to put in a brief. Hence my review feedback.

2. My review feedback was not addressed and I still don't approve the PR.

Issue #1803, opened 2025-01-23
`system` and `process` duplicated attributes


braydonk:
### Area(s)

area:process, area:system

### What's missing?

2 attributes are similar enough between `system` and `process` that they can likely be consolidated while keeping their semantic value intact.

### Describe the solution you'd like

* `system.paging.type` and `process.paging.fault_type` could be merged into `memory.paging.fault_type`
* `system.process.status` could use `process.state`, which is not finished yet but is a GA blocker for that namespace

mx-psi:
From the 2025-02-13 meeting we agree on doing this but we think the first one should be `system.memory.paging.fault_type`

mx-psi:
> From the 2025-02-13 meeting we agree on doing this but we think the first one should be `system.memory.paging.fault_type`

Since we have `system.paging` instead of `system.memory.paging` for the rest of attributes we feel like #1803 naming is fine

ChrsMark:
@open-telemetry/semconv-system-approvers could you confirm that incorporating what was suggested at https://github.com/open-telemetry/semantic-conventions/issues/1501 the attributes should be changed like this?

1. `system.paging.type` and `process.paging.fault_type` should be renamed into `system.paging.fault.type`
2. `system.process.status` become `process.state`

braydonk:
> 1. system.paging.type and process.paging.fault_type should be renamed into system.paging.fault.type

I think `system.paging.fault.type` is fine, initially I pitched `fault_type`, but assuming the metric will also be renamed `system.paging.fault.count` then `system.paging.fault.type` makes sense.

> 2. system.process.status become process.state

Yes.

PR #2705, opened 2025-08-30
Rename process state attribute #1803

thompson-tomo:
Progresses #1803

## Changes

This renames `system.process.status` to `process.state` to follow semconv guidance.

### DIFF
--- a/model/process/registry.yaml
+++ b/model/process/registry.yaml
@@ -258,6 +258,29 @@ groups:
         type: template[string]
         stability: development
         examples: ['ubuntu', '/usr/local/bin:/usr/bin']
+      - id: process.state
+        type:
+          members:
+            - id: running
+              value: 'running'
+              brief: Running
+              stability: development
+            - id: sleeping
+              value: 'sleeping'
+              brief: Sleeping
+              stability: development
+            - id: stopped
+              value: 'stopped'
+              brief: Stopped
+              stability: development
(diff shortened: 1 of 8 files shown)


model/process/registry.yaml
```diff
@@ -258,6 +258,29 @@ groups:
         type: template[string]
         stability: development
         examples: ['ubuntu', '/usr/local/bin:/usr/bin']
+      - id: process.state
+        type:
+          members:
+            - id: running
+              value: 'running'
+              brief: Running
```
ChrsMark: Should `brief` added in enums? Based on https://github.com/open-telemetry/semantic-conventions/issues/2555#issuecomment-3201697007, adding this for enums is blocked at the moment, right?

model/system/metrics.yaml
```diff
@@ -505,6 +505,7 @@ groups:
     unit: "{process}"
     attributes:
       - ref: system.process.status
+      - ref: process.state
```
ChrsMark, line 507: Should the above `system.process.status` be removed completely since it's renamed?

thompson-tomo:
schema.next is now auto generated hence an entry is not needed

model/process/registry.yaml
thompson-tomo: No, what is blocked is simply casing change. Which this was changed from a while ago.

PR #2706, opened 2025-08-30
Consolidate the paging fault attributes #1803

thompson-tomo:
Progresses #1803
Progresses #1501

## Changes

The `system.paging.type` and `process.paging.fault_type` attribute is consolidated into the single attribute `system.paging.fault_type` as per the linked issues

Briefs of the paging metrics have been set which were previously missing with text taken from host collector.

### DIFF
--- a/.chloggen/consolidate_paging-faults.yaml
+++ b/.chloggen/consolidate_paging-faults.yaml
@@ -7,7 +7,7 @@
 change_type: breaking
 
 # The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
-component: system, process
+component: system, process, container, k8s
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: The `system.paging.type` and `process.paging.fault_type` attribute is consolidated into the single attribute `system.paging.fault_type`
--- a/docs/system/container-metrics.md
+++ b/docs/system/container-metrics.md
@@ -255,16 +255,16 @@ In K8s, this metric is derived from the [MemoryStats.PageFaults](https://pkg.go.
 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
-| [`system.paging.type`](/docs/registry/attributes/system.md) | string | The memory paging type | `minor` | `Recommended` | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `system.paging.fault_type`. |
+| [`system.paging.fault_type`](/docs/registry/attributes/system.md) | string | The type of page fault for this data point. | `minor` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 
 ---
 
-`system.paging.type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
+`system.paging.fault_type` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.
 
 | Value  | Description | Stability |
 |---|---|---|
-| `major` | Major | ![Development](https://img.shields.io/badge/-development-blue) |
-| `minor` | Minor | ![Development](https://img.shields.io/badge/-development-blue) |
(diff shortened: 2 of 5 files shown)


model/system/registry.yaml
```diff
@@ -65,14 +65,16 @@ groups:
...
+      - id: system.paging.fault_type
         type:
           members:
             - id: major
               value: 'major'
+              brief: Major
               stability: development
             - id: minor
               value: 'minor'
+              brief: Minor
```
mx-psi: From discussion on the 2025-09-11 System semconv working group, we should not add briefs to the enum values based on https://github.com/open-telemetry/semantic-conventions/issues/2555#issuecomment-3201697007. Could you remove that change from this PR so we can move forward?

thompson-tomo: The verbose briefs have been added which uses info previously in the attribute brief.

mx-psi: Can we leave this PR with only the fault page attribute name change and separate the brief change into a different PR? That part seems more contentious given the comment I linked above so to make progress I think splitting makes sense

thompson-tomo: If you look at the change here, the member brief has actually been extracted from the attribute brief hence is providing additional value and ensuring that the attribute brief is focussed.

model/system/registry.yaml
```diff
@@ -65,17 +65,19 @@ groups:
         stability: development
         brief: "The memory paging state"
         examples: ["free"]
-      - id: system.paging.type
+      - id: system.paging.fault_type
         type:
           members:
             - id: major
               value: 'major'
+              brief: Major/hard page faults.
```
braydonk: The main value of this brief is to clarify the relation between the "major" vs "hard" verbiage so I'd like to suggest a reword to make that more explicit.
```suggestion
              brief: Major page fault. Sometimes referred to as hard page fault.
```

model/system/registry.yaml
```diff
@@ -65,17 +65,19 @@ groups:
...
               stability: development
             - id: minor
               value: 'minor'
+              brief: Minor/soft page faults.
```
braydonk:
```suggestion
              brief: Minor page fault. Sometimes referred to as soft page fault.
```

model/system/metrics.yaml
```diff
@@ -221,6 +222,7 @@ groups:
     unit: "{operation}"
     attributes:
       - ref: system.paging.type
```
braydonk: The deprecated attribute name can be removed.

model/system/deprecated/registry-deprecated.yaml
```diff
@@ -105,3 +105,20 @@ groups:
...
+            - id: major
+              value: 'major'
+              brief: Major/hard page faults.
+              stability: development
+            - id: minor
+              value: 'minor'
+              brief: Minor/soft page faults.
+              stability: development
+        stability: development
+        brief: The type of page fault for this data point.
```
braydonk, line 126: The deprecated entry should maintain the old values for briefs.

model/process/metrics.yaml
```diff
@@ -149,6 +149,7 @@ groups:
     instrument: counter
     unit: "{fault}"
     attributes:
+      - ref: system.paging.fault_type
       - ref: process.paging.fault_type
```
braydonk: The deprecated attribute can be removed.

braydonk (approved):

dashpole (dismissed):
LGTM for k8s semconv

ChrsMark (commented):
I don't see https://github.com/open-telemetry/semantic-conventions/pull/2706#discussion_r2348403153 being sufficiently resolved. And in general, as a reviewer reviewing PRs that involve different types of changes is not preferable. 

@thompson-tomo could you apply the change requested at https://github.com/open-telemetry/semantic-conventions/pull/2706#discussion_r2348403153 and clean up this PR to only address https://github.com/open-telemetry/semantic-conventions/issues/1803? You can add any extra changes for follow-up PRs and have them reviewed separately. Otherwise I don't think this PR will make it as-is.

thompson-tomo:
@ChrsMark actually that feedback has been resolved by implementing the feedback & requested change I recieved here https://github.com/open-telemetry/semantic-conventions/pull/2706#discussion_r2359774548

PR #2708, opened 2025-08-30
[chore] Consolidate thread details into yaml/registry

thompson-tomo:
## Changes

Consolidate the details about thread source into the yaml and by extension the registry.


model/thread/registry.yaml
```diff
@@ -10,10 +10,48 @@ groups:
         stability: development
         brief: >
           Current "managed" thread ID (as opposed to OS thread ID).
+        note: >
+
+          Examples of where the value can be extracted from:
+
+
+          | Language or platform  | Source |
+
```
joaopgrassi, line 19: Do we need the extra empty lines here for it to work?

joaopgrassi (approved):

model/thread/registry.yaml
thompson-tomo: Correct the empty line forces the line break.

PR #2710, opened 2025-08-30
[chore] Increase consistency of general signal docs

thompson-tomo:
## Changes

Link to attribute registry rather than duplicate attribute tables into the general docs.  This enables the general docs to be informative about the signal rather than also conveying attributes which might be used.


lmolkova (changes requested):
This PR seems to randomly remove some parts of the markdown and move other parts of markdown around. I don't understand the benefit or a reason for this change.

I understand there are attribute groups that are duplicated between registry and are also rendered as separate tables and it would be great to improve it. But improving it should not come as a set of semi-random changes. Ideally we need to understand if there is a difference between attribute groups that are rendered intentionally and in the registry. E.g. code attributes (removed here) are special  - they can appear as a group on logs and spans. We need to find a formal way to express groups like this. 

I would discourage attempts to do small random changes without attempt to understand the bigger picture and suggest general solution.

thompson-tomo:
I have gone and additional links to the logs page linking to other namespaces which could be useful to add to logs, hence further improving the experience of reading the page.

> E.g. code attributes (removed here from general) are special - they can appear as a group on logs and spans. 

The code attributes were not removed, what was removed which was a general description of code attributes, but it was lacking any actual attributes. The brief and note for the attributes have since been improved to incorporate more of that information.

With the changes now, under both logs & span pages there is a section for useful additional attribute namespaces for which code was a listed namespace. In effect with this change a user reading about spans, can see the general information about spans and are then provide a simple list of general attribute namespaces which can be used to add additional context to their span. If the reader wants to read more about a namespace, they simply click the link and are taken to a complete list of the attributes which includes both current and past attributes.

> I would discourage attempts to do small random changes without understanding the bigger picture and suggesting general solution.

A general solution was suggested and implemented within the initial commit which was then extended in the subsequent commit to also be implemented for events. The solution is that a section is provided under the general information about a signal to identify the namespaces which can be used to provide additional context for the signal. These identified namespaces are then linked to the corresponding attribute registries which shows both the current and past attributes. 

Note I have been advised repeatedly to implement solution/changes in small pieces especially when multiple code owners are involved.

thompson-tomo:
Following the feedback in semconv meeting of not wanting to point users to registry, the tables of attributes are now being rendered directly on to the pages rather than linking.

Note I would have preferred to put them in the registry hence #2470 but that didn't progress after semconv meeting etc.

trask:
Discussed in SIG meeting, let's hold for now to see how attribute groups proceeds in semconv v2 work

PR #2712, opened 2025-08-31
Introduction of Auto-scaler group #2711

### LINKED ISSUE #2711: Describe auto-scaler which resource belongs to
### Area(s)

area:app, area:container, area:host, area:azure, area:aws, area:gcp

### What's missing?

The ability to describe what auto-scaling group the resource belongs to which can then be used by resource detectors.

The introduction of the autoscaler namespace which can be used by the different cloud providers and the resource types.

I considered adding it to the host namespace which is where the intial need came from however it currently became clear that, auto-scaling could also be used for apps and containers hence seperate namespace.

### Tip

thompson-tomo:
Fixes #2711

## Changes

This introduces the auto-scaler namespace to capture the auto-scaling group which the resource belongs. The resource could be a host (vm), container or even an app hence the generic approach to solving this.

The azure vm resource detector already reports the scaleset via an un-defined attribute and would be switched across to this attribute following acceptance in atleast the dotnet implementation. This attribute is already available in other implementations of the resource detectors as well.

## Implementations
- https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/azure-resources
- https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Resources.Azure

Note attribute to be renamed.


thompson-tomo:
@ChrsMark see description there is 2 sample implementations which would use it. They are currently using `azure.vm.scaleset.name` to capture this exact information but that is not a defined semconv attribute and concept applies more broadly then just Azure.

braydonk:
Seems fine to me at a glance.  
Truthfully I don't have much opinion on this from a System Semconv point of view. This is more of a cloud-level concern than a system-level one. I'm not sure what SIG this area is intended to belong to, but System is probably not the one.

thompson-tomo:
In some cases yes it would be done by a cloud provider but in other cases it would not. I see auto scaling as a part of the system domain given it is describing the software concept used to provide the environment where the service is running.

braydonk:
"Cloud" is perhaps an overloaded term; I mean that this is about some orchestration level over VMs or applications. This is still higher level than the System group is concerned with. I don't really have much opinion on this at all from the perspective of the System group.

ChrsMark:
Both implementations shared are from Azure detectors so probably that's a cloud's domain need?

I would also scope it more as part of the specific context of cloud rather than generalising. K8s domain has its own autoscaler entity: https://opentelemetry.io/docs/specs/semconv/registry/entities/k8s/#k8s-hpa

In general, I don't think this is a priority right now for the System SemConv SIG to deal with it specially when GA-ing system SemConv is/should-be top priority: https://github.com/orgs/open-telemetry/projects/55/views/4

thompson-tomo:
That's only an example but it could also by vmware which is not cloud.

I am aware of that board and that is where I picked a couple of issues to work on from with pr's submitted.

lmolkova (changes requested):
Listed components set one azure-specific attribute that's provided by the environment. They do not implement anything close to the changes being proposed.

Adding new conventions of this scope would need a new project - https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#suggesting-conventions-for-a-new-area

They would need to take k8s and other hosting environments into account.

thompson-tomo:
it is not azure specific, just the first implementation would be azure as it is already using a comparable attribue. For aws the value can be obtained from the instance tag availed from imds aka http://169.254.169.254/latest/meta-data/tags/instance/aws:autoscaling:groupName. I expect for gcp the imds would also provide this group. 
I would understand the need for a project if we were also implementing metrics, spans, events etc as such was a full project. Whereas what I am trying to do is improve the information captured via a resource detector to describe a host.

jsuereth (changes requested):
Defining an Auto-Scaler group deserves a semocnv SIG / project.  I'm with @lmolkova here.  We want SMEs looking at this, and that means forming such a group of individuals and evaluating the state of auto-scaling o11y in OpenTelemetry, and what the right level of abstraction is.

I can add comments directly in this PR, but I think I'd rather see folks from auto-scaling communities work with us here, and propose a direction first before a change.

Adding an additional blocking comment.

Issue #2223, opened 2025-05-04
Should we design `process.command_args` as template attr?


lmolkova:
`process.command_args` is somewhat similar to `db.query.parameter.<key>` where we capture each params as an individual attribute also allowing to opt-in into them independently.

How it could work for `process.command_args`:

- the basic instrumentation just splits args and reports them as positional `process.command.args.{index}`. E.g. `C:\cmd\otecol --config="my directory\config.yaml"` would result in `process.command.args.0 = "C:\cmd\otecol"`, `process.command.args.1 = "--config="my directory\config.yaml"`, 
- advanced generic instrumentation or process-specific one may do smarter things based on cmd parser being used or common patterns, it can do `process.command.args.config = "my directory\config.yaml"`

While implementing anything beyond positional args is not straightforward, we should consider adopting the params pattern used across semconv to make it possible to provide great experience.

trisch-me:
I think for named attributes it would be a nice feature, but for the positional it’s already there assuming the order of strings in array is determined
From other side named attributes will loose then their order and this is not great. For example there are security rules that analysing the order of process parameters to catch command injection, path manipulations etc.

mx-psi:
Live discussion from System SemConv in-person meeting: there is no standard for how this should work and it would be hard to come up with a generic way of representing things. For example: how would one represent `ls *.jpg *.png`, `weaver registry infer` or `otelcol validate --config`? How does one represent the same flag being passed to the parent command and the subcommand with different values (this seems like an edge case but the world is full of edge cases)? What about if you have repeated arguments and the ordering matters (also an edge case, also probably something that exists out there)?

We think we should keep it as is unless there is a way to come up with a way to represent these concepts in a robust way.

PR #2713, opened 2025-09-01
Semantic Conventions for Generative AI Tasks (gen_ai.task.*) - Task Identity and Purpose 

divyapathak24:
Fixes 

- open-telemetry/semantic-conventions-genai#35 
- open-telemetry/semantic-conventions-genai#37 

***
## Changes

This PR introduces a foundational set of semantic attributes under a new domain prefix: `gen_ai.task.*`, specifically for representing **task-level telemetry** in agentic and generative AI systems. **Tasks** are fundamental **units** of work in AI Agentic systems, capturing structured goals like planning, reasoning, and synthesis. The ability to trace and observe these tasks is essential for production readiness, particularly from an **SRE or observability** standpoint. 

Benefits: 
(1) Adding GenAI tasks will provide **unified tracing** by standardizing spans and traces across different agent middleware frameworks (such as CrewAI, LangGraph, AutoGen). 
(2) Manifestation of raw spans from the task perspective, i.e., **n spans may abstract to m** (where, m <= n)  tasks, by merging non-informative child spans into their parent spans recursively. 

This initial proposal on tasks includes attributes to support task identity and purpose.

 **New Semantic Conventions Added (gen_ai.task.*):**

- gen_ai.task.id – Unique identifier for a task instance.
- gen_ai.task.parent.id – ID of the parent task (if applicable).
- gen_ai.task.name – Human-readable task label.
- gen_ai.task.code.id – Reference to the source code defining the task.
- gen_ai.task.code.vendor – Vendor/framework used (e.g., LangGraph, CrewAI).
- gen_ai.task.kind – Describes the task's purpose (e.g., retrieval, reasoning).
- gen_ai.task.tags – Custom tags for filtering, analytics, and domain grouping.

**Motivation**

These conventions standardize how agent task metadata is captured across spans. This supports:
- Trace correlation across distributed workflows
- Root cause analysis via parent-child task relationships
- Filtering by task type and tags for debugging, performance, and SLO monitoring

**For more details and a demo example, please refer to this document:** [Google Document Link](https://docs.google.com/document/d/15s1VPlzIaIx4XI7JnmUET4eZZROsdShp8NTH8VSKCO4/edit?tab=t.0)

***


divyapathak24:
cc: @PRATIBHA-Moogi @gyliu513 @hk-bmi

model/gen-ai/registry.yaml
```diff
@@ -504,3 +504,66 @@ groups:
...
+      - id: gen_ai.task.code.id
+        stability: development
+        type: string
+        brief: Fully qualified reference to the source code defining the task logic or structure.
+        examples: ["https://github.com/crewAIInc/crewAI/blob/main/src/crewai/task.py:56:Task"]
+      - id: gen_ai.task.code.vendor
+        stability: development
+        type: string
+        brief: The name of the vendor or framework used to develop the agent application.
+        examples: ["langgraph", "crewAI"]
```
thompson-tomo, line 540: Is the code sub-namespace necessary?

model/gen-ai/registry.yaml
```diff
@@ -504,3 +504,66 @@ groups:
...
+        brief: A user-defined display name for the task.
+        examples: ["Summarize Customer Review"]
+        note: |
+          If not explicitly set, it could be inferred from the code, metadata, agent prompt, or system instrumentation.
+          Improves trace readability and simplifies debugging by making traces easier to scan and interpret.
+      - id: gen_ai.task.code.id
+        stability: development
+        type: string
+        brief: Fully qualified reference to the source code defining the task logic or structure.
+        examples: ["https://github.com/crewAIInc/crewAI/blob/main/src/crewai/task.py:56:Task"]
```
thompson-tomo, line 535: Would it be possible to use the code namespace?

model/gen-ai/spans.yaml
```diff
@@ -327,3 +327,27 @@ groups:
         requirement_level: required
       - ref: aws.bedrock.knowledge_base.id
         requirement_level: recommended
+
+  - id: span.gen_ai.task.client
+    span_kind: client
```
thompson-tomo: Would internal be more accurate given client implies a client/server relationship? Or add server attributes.

thompson-tomo (commented):
Does gen_ai.task.id uniquely identify a definition of a task which is reused everytime the task is used or is it identifying the usage of the definition aka the instance and will change?

model/gen-ai/registry.yaml
```diff
@@ -504,3 +504,66 @@ groups:
...
+            - action
+            - evaluation
+            - delegation
+            - synthesis
+            - coordination
+            - refinement
+            - other
+
+          Enables filtering and grouping of traces by task type.
+          Supports task-specific SLOs, such as tighter latency goals for "retrieval" than for "planning".
```
thompson-tomo: This should also be an enum

ShipraJain01:
@divyapathak24 : thanks for putting up the PR. We have a proposal created on similar lines to define Multi-Agents tracing and introducing a new Span type - 'execute-task' and its relevant attributes earlier in July'25. Please find more details here - https://docs.google.com/document/d/1fcPe3SB_koRNeOoioq28RbA1BC7MNA7-/edit?usp=drive_link&ouid=109914447143824537379&rtpof=true&sd=true


we can discuss further on slack

model/gen-ai/registry.yaml
```diff
@@ -504,3 +504,66 @@ groups:
...
+          Allows constructing task trees such as "Summarize → Retrieve → Generate → Evaluate", with each step represented as a task/subtask.
+      - id: gen_ai.task.name
+        stability: development
+        type: string
+        brief: A user-defined display name for the task.
+        examples: ["Summarize Customer Review"]
+        note: |
+          If not explicitly set, it could be inferred from the code, metadata, agent prompt, or system instrumentation.
+          Improves trace readability and simplifies debugging by making traces easier to scan and interpret.
+      - id: gen_ai.task.code.id
```
gyliu513, line 531: What is the benefit of specifying the exact code here? Why not use documentation instead? For the end user customer of GenAI observability, using documentation maybe better?

model/gen-ai/spans.yaml
```diff
@@ -327,3 +327,27 @@ groups:
         requirement_level: required
       - ref: aws.bedrock.knowledge_base.id
         requirement_level: recommended
+
+  - id: span.gen_ai.task.client
```
gyliu513: this needs to be put to https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md, you can refer to https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#2-update-the-markdown-files for how to add a new group.

ShipraJain01: @divyapathak24 : could you help explain where in the Agent's hierarchy this span would be initiated? Is this proposed to be a parent or child to agent's span?

model/gen-ai/registry.yaml
ShipraJain01: @divyapathak24 : same question

model/gen-ai/registry.yaml
```diff
@@ -504,3 +504,66 @@ groups:
...
+        stability: development
+        type: string
+        brief: Fully qualified reference to the source code defining the task logic or structure.
+        examples: ["https://github.com/crewAIInc/crewAI/blob/main/src/crewai/task.py:56:Task"]
+      - id: gen_ai.task.code.vendor
+        stability: development
+        type: string
+        brief: The name of the vendor or framework used to develop the agent application.
+        examples: ["langgraph", "crewAI"]
+      - id: gen_ai.task.kind
```
ShipraJain01, line 541: @divyapathak24 : do you think `kind` depends on the multi-agent pattern and not all of the enum stated below are generically applicable?

model/gen-ai/registry.yaml
```diff
@@ -504,3 +504,66 @@ groups:
...
+            - evaluation
+            - delegation
+            - synthesis
+            - coordination
+            - refinement
+            - other
+
+          Enables filtering and grouping of traces by task type.
+          Supports task-specific SLOs, such as tighter latency goals for "retrieval" than for "planning".
+      - id: gen_ai.task.tags
```
ShipraJain01, line 593: @divyapathak24 : Every additional string array attribute increases payload size. An unbounded attribute like `tags` without strict schema can become inconsistent, reducing reliability for analytics, and would lead to telemetry overhead. In what cases could we really need `tags` in telemtry?

model/gen-ai/registry.yaml
ShipraJain01, line 531: @divyapathak24 : How do you see reference to source code of task logic in telemetry helps?

model/gen-ai/registry.yaml
dany-moshkovich: @ShipraJain01 @thompson-tomo @divyapathak24 
The gen_ai.task.id reflects the specific instance of a task (not a meta-class) and is used to identify all updates to the task throughout its lifecycle. 
The gen_ai.task.code.* attributes are used to represent the meta-class associated with the instance. I hope this answers your second question.

model/gen-ai/spans.yaml
dany-moshkovich: @thompson-tomo @divyapathak24 I also think that internal would be more appropriate here

model/gen-ai/registry.yaml
thompson-tomo: Based on that, the name/kind attributes should also be in the same sub namespace as they correspond to id to remove confusion.

thompson-tomo:
So on revisiting this and clarification via comment, I have the feeling that we should look to use a common definition as proposed in open-telemetry/semantic-conventions#2223 for the core concept and enrich it with gen-ai attributes.

Hence my suggestion would be to pick what is required from open-telemetry/semantic-conventions#2223 the Workflow.task & Workflow.execution sub namespaces. Then only add the gen-ai attributes which is specific to tasks within gen-ai scope.

model/gen-ai/registry.yaml
divyapathak24: @thompson-tomo @ShipraJain01 
1. `gen_ai.task.id` uniquely identifies a task instance within a trace
2. `task.name` and `task.kind` reflect the task’s runtime name and role and are not static, so they belong at the top level namespace, while .code.* is vendor-specific and static.

model/gen-ai/registry.yaml
divyapathak24: @thompson-tomo Just to clarify—are you suggesting keeping `gen_ai.code.*`? If so, we placed `code.*` under task because it describes the implementation of a specific task, not a general code entity, so `gen_ai.task.code.*` preserves that scoped relationship.

model/gen-ai/registry.yaml
divyapathak24: @gyliu513 @ShipraJain01 Linking the exact code can prove critical for debugging — for example, if gen_ai.task.code.id points to `task.py:56:Task`, an SRE/engineers can jump straight from a failing span to the exact function implementation, instead of searching through docs that may be outdated or ambiguous. For end users, docs can aid understanding, but linking exact code ensures engineers resolve issues quickly.

model/gen-ai/registry.yaml
divyapathak24: @ShipraJain01 We have made the `gen_ai.task.kind` values as generic as possible. For example, action—“Executing a concrete command (e.g., tool, API, function)”—can represent any actionable step. Depending on the multi-agent pattern, you can choose the most appropriate value, but the enum values are designed to apply broadly across workflows.

model/gen-ai/registry.yaml
divyapathak24: @ShipraJain01 `tags` are useful and **optional** in telemetry when you need flexible, filterable metadata—like "customer-support", "marketing", "compliance".

model/gen-ai/registry.yaml
thompson-tomo: I understand but *.id/*.name is usually used & viewed as a pair of attributes but here you are breaking that established pattern which will lead to confusion like it has.

Hence I would rename task.id to be task.instance.id similar to other namespaces. Also get rid of the code sub-namespace as there is no reason to split based on if it is static given that they are all properties of the same object.

docs/gen-ai/gen-ai-spans.md
```diff
@@ -415,6 +415,108 @@ Datastore: A tool used by the agent to access and query structured or unstructur
...
+| [`gen_ai.task.code.id`](/docs/registry/attributes/gen-ai.md) | string | Fully qualified reference to the source code defining the task logic or structure. | `https://github.com/crewAIInc/crewAI/blob/main/src/crewai/task.py:56:Task` | `Conditionally Required` If provided by the application. | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`server.port`](/docs/registry/attributes/server.md) | int | GenAI server port. [8] | `80`; `8080`; `443` | `Conditionally Required` If `server.address` is set. | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`server.address`](/docs/registry/attributes/server.md) | string | GenAI server address. [9] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`gen_ai.task.code.vendor`](/docs/registry/attributes/gen-ai.md) | string | The name of the vendor or framework used to develop the agent application. | `langgraph`; `crewAI` | `Opt-In` | ![Development](https://img.shields.io/badge/-development-blue) |
```
pradystar, line 445: I remember there was a discussion around this attribute and the plan was to add a `gen_ai.framework` to inference spans. Based on the description both attributes mean the same. Can you consider switching to `gen_ai.framework`?

PR #2715, opened 2025-09-03
Revive `event.name` attribute

### LINKED ISSUE #2597: Revive `event.name` attribute

pellared:
Fixes #2597

The `event.name` attribute is now undeprecated and can be used by:
  - Non-OTLP exporters to emit the `EventName`
  - Applications using existing logging libraries to add event name information that can be used to set the `EventName` field by Collector or SDK components

Prior-art:
- https://github.com/open-telemetry/opentelemetry-go/pull/7128

### DIFF
--- a/model/event/registry.yaml
+++ b/model/event/registry.yaml
@@ -12,7 +12,7 @@ groups:
           Identifies the class / type of event.
         note: >
           This attribute SHOULD be used by non-OTLP exporters
-          when destination does not support `EventName` or equivalent property.
+          when destination does not support `EventName` or equivalent field.
           This attribute MAY be used by applications using existing logging
           libraries so that it can be used to set the `EventName` field by
           Collector or SDK components.
--- a/docs/registry/attributes/event.md
+++ b/docs/registry/attributes/event.md
@@ -11,4 +11,4 @@ Attributes for Events represented using Log Records.
 |---|---|---|---|---|
 | <a id="event-name" href="#event-name">`event.name`</a> | string | Identifies the class / type of event. [1] | `browser.mouse.click`; `device.app.lifecycle` | ![Development](https://img.shields.io/badge/-development-blue) |
 
-**[1] `event.name`:** This attribute SHOULD be used by non-OTLP exporters when destination does not support `EventName` or equivalent property. This attribute MAY be used by applications using existing logging libraries so that it can be used to set the `EventName` field by Collector or SDK components.
+**[1] `event.name`:** This attribute SHOULD be used by non-OTLP exporters when destination does not support `EventName` or equivalent field. This attribute MAY be used by applications using existing logging libraries so that it can be used to set the `EventName` field by Collector or SDK components.


.chloggen/2597.yaml
```diff
@@ -0,0 +1,8 @@
+change_type: enhancement
+component: event
+note: "Revive `event.name` attribute for use by non-OTLP exporters and logging libraries."
+issues: [2597]
+subtext: |
+  The `event.name` attribute is now undeprecated and can be used by:
+  - Non-OTLP exporters (e.g. console exporter) to emit the EventName
```
lmolkova: I'd imagine OTel console exporter to print EventName as a top-level property like severity or timestamp

```suggestion
  - Non-OTLP exporters to emit the EventName
```

.chloggen/2597.yaml
```diff
@@ -0,0 +1,8 @@
...
+  - Applications using existing logging libraries to add event name information that can be processed by SDK or Collector components
```
lmolkova: my understanding that we want bridges to map external `event.name` to EventName:

```suggestion
  - Applications using existing non-OTel logging libraries to add event name information that would be converted by logging bridges to top-level EventName property
```

does it matches yours?

model/event/registry.yaml
```diff
@@ -8,10 +8,11 @@ groups:
...
-          reason: uncategorized
-          note: >
-            Replaced by EventName top-level field on the LogRecord.
         brief: >
           Identifies the class / type of event.
+        note: >
+          This attribute is used by non-OTLP exporters (e.g. console exporter)
+          to emit the EventName and by applications using existing logging
+          libraries to add event name information that can be processed by SDK
+          or Collector components to set the EventName on log records.
```
lmolkova: same suggestions as the changelog
```suggestion
        note: >
          This attribute SHOULD be used by exporters when destination does not 
          support top-level `EventName` property.
          
          It MAY also be used by applications using 3rd party logging libraries
          that don't have top level `EventName` property (or a similar one that uniquely identifies 
          event structure.        
          [Log bridges](https://opentelemetry.io/docs/specs/otel/glossary/#log-appender--bridge) 
          SHOULD map `event.name` attribute into top-level `EventName` property.
```

SHOULD everywhere since it's a new requirement and some might not be able to support it at least right away. 

Also, not blocking this PR, wonder if we should have something general for bridge/appenders on how to map - related to #1550 and we'd reference it in the https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model-appendix.md

.chloggen/2597.yaml
pellared: you are right

.chloggen/2597.yaml
pellared:
```suggestion
  - Applications using existing logging libraries to add event name information that can be used to set the EventName field by SDK or Collector components
```

model/event/registry.yaml
pellared: > SHOULD map `event.name` attribute into top-level `EventName` property

> Also, not blocking this PR, wonder if we should have something general for bridge/appenders on how to map - related to https://github.com/open-telemetry/semantic-conventions/issues/1550 and we'd reference it in the https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model-appendix.md

I am against doing it in the logging bridges.
There is no need that each bridge would have to implement this.
The Collector can do it and if anyone does not use a Collector then they cam use something like: https://github.com/open-telemetry/opentelemetry-go/pull/7128

pellared:
```suggestion
        note: >
          This attribute SHOULD be used by non-OTLP exporters
          when destination does not support `EventName` or equivalent property.
          This attribute MAY be used by applications using existing logging
          libraries that does not support `EventName` or equivalent property
          so that it can be used to set the `EventName` field by Collector or SDK components.
```

arminru (approved):

PR #2717, opened 2025-09-03
Don't use placeholder values for the controller and action in the http.route example

### LINKED ISSUE #2616: Improve guidelines to HTTP span name generation

Some instrumentation libraries confuse how to generate correct http server span name for template based routing. Add more examples and fix opentelemetry.io interpretation.

**Additional context.**

Contradictions in opentelemetry.io may suggest that http span name can be named as (some examples) and cause very low cardinality:

`GET {controller=Home}/{action=Index}/{id?}` (actual req path: `url.path /Home/About`)
`GET [controller}/{action}/{id}` (actual req path: `url.path /Home/About`)

Latest spec
https://github.com/open-telemetry/opentelemetry-specification/blob/v1.47.0/specification/trace/api.md#span

opentelemetry.io
https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name
https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/#http-route

Guidelines to improve suggestions.

Template based routing:

* Leave static route tokens the same
* Replace framework variables in route template with actual values and leave variable name
* Leave dynamic variables as template style tokens

Request: (HTTP 200) http://example.org/weather/today/london
`GET api/{controller=Weather}/{action=Today}/{location-id}`

static assests:

* Leave static asset path since they are mostly static
* If template based routing is used, follow template based routing guidelines

Request: (HTTP 200) http://example.org/js/framework.js?v=abcde
`GET /js/framework.js`

invalid requests (HTTP 400-500):

* different back-end providers may want to have different logic
* different clients may want to have different logic (bots may cause high cardinality attacks)
* use configuration to switch between verbose and basic name.

Request: HTTP 404 http://example.org/invalid-path
`GET` vs `GET /invalid-path`

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

lmolkova:
Fixes #2616

Using `{controller}/{action}/...` in http.route examples turns out to be confusing - according to #2616 it creates the impression that it should be used verbatim, while the intention is to report the route information available from HTTP server  framework which usually contains full-text controller/actions.

The #2616 suggests to fix span name description, but I still struggle to understand what's missing there - it just tells to use HTTP route:

https://github.com/open-telemetry/semantic-conventions/blob/b321d3ae8bc1e4025197c148afc4cc9af5c0f945/docs/http/http-spans.md?plain=1#L74-L80


RassK:
Suggestion for `http.route` standardization example `my-controller/my-action/{id?}` to remove all route transformers and constraints also. (`?` is a constraint, like `:string` or `:int`)

`?` for sure causes another issue. Can execute just `my-controller/my-action/` and hit the same route as `{controller}/{action}/{id?}` . So in this case `{id}` should not be used.

request: `http://example.org/my-controller/my-action`
route template: `{controller}/{action/{id?}`
http.route: `my-controller/my-action`

request: `http://example.org/my-controller/my-action/5`
route template: `{controller}/{action/{id?}`
http.route: `my-controller/my-action/{id}`

lmolkova:
@RassK HTTP conventions don't define any semantics for HTTP route - the value is coming from the HTTP frameworks. Do you think there a better way to describe it than 

https://github.com/open-telemetry/semantic-conventions/blob/b321d3ae8bc1e4025197c148afc4cc9af5c0f945/model/http/registry.yaml#L172-L178

? 

Given it's stable, we won't be able to standardize the format.

Kielek (approved):
Thank you!

IMO it will reduce friction related to the recommendation what should be real outcome for `http.route` and the span name.

RassK:
> @RassK HTTP conventions don't define any semantics for HTTP route - the value is coming from the HTTP frameworks. Do you think there a better way to describe it than

Yes, I see no direct conventions. I'm interpreting it like that: There is a spec for http server span name, and if the `http.route` must be included (if exists) in the span name, then span naming spec is a constraint to `http.route`.

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.47.0/specification/trace/api.md#span
> The span name concisely identifies the work represented by the Span, for example, an RPC method name, a function name, or the name of a subtask or stage within a larger computation. The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances while still being human-readable. That is, "get_user" is a reasonable name, while "get_user/314159", where "314159" is a user ID, is not a good name due to its high cardinality. Generality SHOULD be prioritized over human-readability.

This could be seen constraining `http.route` not to include framework specifics (like `?`, `:int`, `:string`, `:transformer`). Since framework specifics does not follow "The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans".

### Suggestions

Some help from chatgpt:

| Attribute | Type | Description |
| ---------- | ----- | ------------- |
http.route | string | The matched route template for the request, with dynamic path segments replaced by {param} placeholders. MUST be low-cardinality and independent of framework syntax. Example: /users/{userId}, /articles/{articleId}/comments.

Rules:

* Replace dynamic path segments (IDs, UUIDs, slugs, etc.) with {param}.
* Normalize all placeholders to {param} format, regardless of framework (:id, <id>, {id}).
* Omit optional markers (?) — always record the general case.
* Include all parent segments that are part of the matched route.
* MUST NOT include actual values from requests (e.g., /users/314159).

Examples:

* Express /users/:id → /users/{id}
* Django /articles/<int:id>/edit/ → /articles/{id}/edit
* ASP.NET {controller}/{action}/{id?} (resolved to Users/Get/42) → /users/get/{id}
* Spring Boot /orders/{orderId} → /orders/{orderId}

matt-hensley:
In the case of a route template like `{controller}/{action}/{id?}`, I'd personally categorize all 3 parameters as dynamic. I can see how others may consider `controller` and `action` static, as there are a finite set of values for most applications.

Some additional detail might leave less room for interpretation. Here is an attempt:

> The matched route template for the request. This MUST be low-cardinality and include all static path segments, with dynamic path segments represented with placeholders.

Along with a footnote defining static vs dynamic and examples for common frameworks like @RassK commented:

> A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that are constrained to a finite, predefined set of values `{controller}` or `{action}`.
>
> A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
> `<table of examples>`

lmolkova:
@matt-hensley @RassK makes sense, would you be willing to send a PR? I can close this one.

matt-hensley:
@lmolkova opened #2734 for consideration

lmolkova:
closing in favor of #2734

PR #2734, opened 2025-09-04
Clarify routes may contain static and dynamic segments

### LINKED ISSUE #2616: Improve guidelines to HTTP span name generation

matt-hensley:
Fixes #2616

## Changes

Alternative proposal to #2717 with additional language to explain intent behind "static" and "dynamic" route segments and how their value should be handled.

### DIFF
--- a/docs/http/http-spans.md
+++ b/docs/http/http-spans.md
@@ -516,9 +516,9 @@ If the request has completed successfully, instrumentations SHOULD NOT set `erro
 SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
 
 A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that
-are constrained to a finite, predefined set of values `{controller}` or `{action}`.
+are constrained to a finite, predefined set of values, e.g. `{controller}` or `{action}`.
 
-A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
+A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like static path segments.
 
 Instrumentations SHOULD use routing information provided by the corresponding web framework. They SHOULD pick the most precise source of routing information and MAY
 support custom route formatting. Instrumentations SHOULD document the format and the API used to obtain the route string.
--- a/docs/registry/attributes/http.md
+++ b/docs/registry/attributes/http.md
@@ -80,9 +80,9 @@ Examples:
 SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
 
 A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that
-are constrained to a finite, predefined set of values `{controller}` or `{action}`.
+are constrained to a finite, predefined set of values, e.g. `{controller}` or `{action}`.
 
-A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
+A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like static path segments.
 
 Instrumentations SHOULD use routing information provided by the corresponding web framework. They SHOULD pick the most precise source of routing information and MAY
 support custom route formatting. Instrumentations SHOULD document the format and the API used to obtain the route string.
--- a/docs/dotnet/dotnet-aspnetcore-metrics.md
+++ b/docs/dotnet/dotnet-aspnetcore-metrics.md
@@ -81,9 +81,9 @@ All routing metrics are reported by the `Microsoft.AspNetCore.Routing` meter.
 SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
 
 A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that
-are constrained to a finite, predefined set of values `{controller}` or `{action}`.
+are constrained to a finite, predefined set of values, e.g. `{controller}` or `{action}`.
 
-A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
+A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like static path segments.
 
 Instrumentations SHOULD use routing information provided by the corresponding web framework. They SHOULD pick the most precise source of routing information and MAY
 support custom route formatting. Instrumentations SHOULD document the format and the API used to obtain the route string.
--- a/docs/http/http-metrics.md
+++ b/docs/http/http-metrics.md
@@ -130,9 +130,9 @@ If the request has completed successfully, instrumentations SHOULD NOT set `erro
 SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
 
 A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that
-are constrained to a finite, predefined set of values `{controller}` or `{action}`.
+are constrained to a finite, predefined set of values, e.g. `{controller}` or `{action}`.
 
-A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
+A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like static path segments.
 
 Instrumentations SHOULD use routing information provided by the corresponding web framework. They SHOULD pick the most precise source of routing information and MAY
 support custom route formatting. Instrumentations SHOULD document the format and the API used to obtain the route string.
@@ -335,9 +335,9 @@ If the request has completed successfully, instrumentations SHOULD NOT set `erro
 SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
 
 A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that
-are constrained to a finite, predefined set of values `{controller}` or `{action}`.
+are constrained to a finite, predefined set of values, e.g. `{controller}` or `{action}`.
 
-A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
+A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like static path segments.
 
 Instrumentations SHOULD use routing information provided by the corresponding web framework. They SHOULD pick the most precise source of routing information and MAY
 support custom route formatting. Instrumentations SHOULD document the format and the API used to obtain the route string.
@@ -469,9 +469,9 @@ If the request has completed successfully, instrumentations SHOULD NOT set `erro
 SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
 
 A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that
-are constrained to a finite, predefined set of values `{controller}` or `{action}`.
+are constrained to a finite, predefined set of values, e.g. `{controller}` or `{action}`.
 
-A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
+A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like static path segments.
 
 Instrumentations SHOULD use routing information provided by the corresponding web framework. They SHOULD pick the most precise source of routing information and MAY
 support custom route formatting. Instrumentations SHOULD document the format and the API used to obtain the route string.


docs/registry/attributes/http.md
```diff
@@ -79,6 +79,17 @@ Examples:
 **[5] `http.route`:** MUST NOT be populated when this is not supported by the HTTP server framework as the route attribute should have low-cardinality and the URI path can NOT substitute it.
 SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
 
+A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that are constrained to a finite, predefined set of values `{controller}` or `{action}`.
+
+A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
+
+Examples of converting route templates to `http.route` values:
+
+* ASP.NET: `{controller}/{action}/{id?}` becomes `/users/get/{id}`
```
lmolkova: I like this section and having examples, but otel instrumentations don't do any conversion. All they do is pick fx-specific API that captures low-cardinality route.

E.g. 
https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.routing.patterns.routepattern.rawtext?view=aspnetcore-9.0#microsoft-aspnetcore-routing-patterns-routepattern-rawtext

or 

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/14a536051a9cdcaea3fffff3b2fc5832b4e2c5f5/instrumentation/spring/spring-webmvc/spring-webmvc-6.0/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/v6_0/HttpRouteSupport.java#L112

or 

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fd2156395753823b82acd87347e2f050b5235e91/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py#L520

I'll leave a few suggestions on how to capture this aspect

docs/registry/attributes/http.md
```diff
@@ -79,6 +79,17 @@ Examples:
...
+* Django: `/articles/int:id/edit/` becomes `/articles/:id/edit`
```
lmolkova: what would be the benefit of this conversion for the end users? who would do it?

keep in mind, this is stable spec and implemented by stable instrumentation. We don't guarantee that attribute values remain stable, but there should still be a strong reason to change something that's stable.

docs/registry/attributes/http.md
```diff
@@ -79,6 +79,17 @@ Examples:
 **[5] `http.route`:** MUST NOT be populated when this is not supported by the HTTP server framework as the route attribute should have low-cardinality and the URI path can NOT substitute it.
 SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
 
+A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that are constrained to a finite, predefined set of values `{controller}` or `{action}`.
+
+A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
+
```
lmolkova, line 86: [updated]  based on https://github.com/open-telemetry/semantic-conventions/pull/2734#issuecomment-3255646641

Maybe we can document the current state:

```suggestion
Instrumentations SHOULD use routing information provided by corresponding web-framework. They SHOULD pick the most precise source of routing information and MAY support custom route formatting. Instrumentations SHOULD document the format and the API used to obtain route string.
```

lmolkova (commented):
Thanks! 

I like the clarifications and examples of how routes look like in different frameworks, but I still don't get why we'd change stable instrumentation behavior and change routing info provided by the framework.

docs/http/http-spans.md
```diff
@@ -515,6 +515,17 @@ If the request has completed successfully, instrumentations SHOULD NOT set `erro
 **[6] `http.route`:** MUST NOT be populated when this is not supported by the HTTP server framework as the route attribute should have low-cardinality and the URI path can NOT substitute it.
 SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
 
+A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that are constrained to a finite, predefined set of values `{controller}` or `{action}`.
+
+A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
+
+Examples of converting route templates to `http.route` values:
+
+* ASP.NET: `{controller}/{action}/{id?}` becomes `/users/get/{id}`
```
lmolkova: where `{controller}/{action}/{id?}` come from? can I repro it?

RassK: `{controller}/{action}/{id?}` is a single default route template that asp.net uses for many years.

matt-hensley: This is the default route for ASP.NET Core MVC (and the older ASP.NET MVC).

https://learn.microsoft.com/en-us/aspnet/core/mvc/controllers/routing?view=aspnetcore-9.0#crd6

Run `dotnet new mvc` in an empty directory and examine the `Program.cs`. There will be some lines like:

```
app.MapControllerRoute(
    name: "default",
    pattern: "{controller=Home}/{action=Index}/{id?}")
    .WithStaticAssets();
```

With `Home` and `Index` being the default values for their respective route parameters.

RassK:
> I like the clarifications and examples of how routes look like in different frameworks, but I still don't get why we'd change stable instrumentation behavior and change routing info provided by the framework.

Because customers don't like how there is no cardinality (only 1 grouping) in asp.net which is `{controller}/{action}/{id?}`. There is absolutely 0 value, especially if there are no other route templates used.

docs/registry/attributes/http.md
RassK: this v1.47.0 span name spec which I always bring up.

>The span name concisely identifies the work represented by the Span, for example, an RPC method name, a function name, or the name of a subtask or stage within a larger computation. The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances while still being human-readable.

These examples follow the spec more than what's currently there. And Span name is defined as `{http-method} {http.route}`.

lmolkova: my comment is about this example in particular. what does `/articles/int:id/edit/`-> `/articles/:id/edit` (removal of `int`) achieve? 

I don't understand the principle or benefit of this.

docs/http/http-spans.md
lmolkova: If I do this:
```csharp
app.MapControllerRoute(
    name: "default",
    pattern: "{controller=Home}/{action=Index}/{id?}");
```

I get  `http.route: {controller=Home}/{action=Index}/{id?}` which is meaningful. right?

If I do 

```csharp
app.MapControllerRoute(
    name: "weather",
    pattern: "{controller}/{action}/{id?}",
    defaults: new { controller = "WeatherForecast", action = "Get" });
```

I get `http.route: weatherforecast`

When do I get `{controller}/{action}/{id?}` ?

lmolkova:
> > I like the clarifications and examples of how routes look like in different frameworks, but I still don't get why we'd change stable instrumentation behavior and change routing info provided by the framework.
> 
> Because customers don't like how there is no cardinality (only 1 grouping) in asp.net which is `{controller}/{action}/{id?}`. There is absolutely 0 value, especially if there are no other route templates used.

I'm with you - what I don't get is where do we produce absolutely useless `{controller}/{action}/{id?}` - do you have a repro?

RassK:
> > > I like the clarifications and examples of how routes look like in different frameworks, but I still don't get why we'd change stable instrumentation behavior and change routing info provided by the framework.
> > 
> > 
> > Because customers don't like how there is no cardinality (only 1 grouping) in asp.net which is `{controller}/{action}/{id?}`. There is absolutely 0 value, especially if there are no other route templates used.
> 
> I'm with you - what I don't get is where do we produce absolutely useless `{controller}/{action}/{id?}` - do you have a repro?

I'm not sure I follow 100%. 
https://opentelemetry.io/docs/languages/dotnet/traces/getting-started-aspnetcore/
The default asp.net core template + opentelemetry with asp.net core tracing. 

(See more faulting spans at https://github.com/open-telemetry/semantic-conventions/issues/2616#issuecomment-3178926572)

produces:
Activity.DisplayName:        GET {controller=Home}/{action=Index}/{id?}
http.route: {controller=Home}/{action=Index}/{id?}

```
Activity.TraceId:            0afccf0559cddc29dfaf5c808d44d2f9
Activity.SpanId:             ff3deeb995003d11
Activity.TraceFlags:         Recorded
Activity.DisplayName:        GET {controller=Home}/{action=Index}/{id?}
Activity.Kind:               Server
Activity.StartTime:          2025-09-04T20:19:03.3173784Z
Activity.Duration:           00:00:00.1346969
Activity.Tags:
    server.address: localhost
    server.port: 5218
    http.request.method: GET
    url.scheme: http
    url.path: /
    network.protocol.version: 1.1
    user_agent.original: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36 Edg/139.0.0.0
    http.route: {controller=Home}/{action=Index}/{id?}
    http.response.status_code: 200
Instrumentation scope (ActivitySource):
    Name: Microsoft.AspNetCore
Resource associated with Activity:
    service.name: WebApplication2
    service.instance.id: 00052429-df01-4204-aaff-de18e8e6f3a7
    telemetry.sdk.name: opentelemetry
    telemetry.sdk.language: dotnet
    telemetry.sdk.version: 1.12.0
```

lmolkova:
what's the problem with `http.route: {controller=Home}/{action=Index}/{id?}`? it shows which controller and action were hit (`home` and `index`). What would you capture if not that?

RassK:
> what's the problem with `http.route: {controller=Home}/{action=Index}/{id?}`? it shows which controller and action were hit (`home` and `index`). What would you capture if not that?

No, in terms of ASP.NET it's the syntax to say which controller and action is default if you hit the root. (see `url.path: /`)

See this example (there is no {id} even, because ? allows it to miss):
url.path: /Home/Privacy
http.route: {controller=Home}/{action=Index}/{id?}
Activity.DisplayName:        GET {controller=Home}/{action=Index}/{id?}
```
Activity.TraceId:            aa64df047db6b9bb0cf5a1c01765b0a5
Activity.SpanId:             a732ab6df880c40b
Activity.TraceFlags:         Recorded
Activity.DisplayName:        GET {controller=Home}/{action=Index}/{id?}
Activity.Kind:               Server
Activity.StartTime:          2025-09-04T20:29:37.2783252Z
Activity.Duration:           00:00:00.0096519
Activity.Tags:
    server.address: localhost
    server.port: 5218
    http.request.method: GET
    url.scheme: http
    url.path: /Home/Privacy
    network.protocol.version: 1.1
    user_agent.original: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36 Edg/139.0.0.0
    http.route: {controller=Home}/{action=Index}/{id?}
    http.response.status_code: 200
Instrumentation scope (ActivitySource):
    Name: Microsoft.AspNetCore
Resource associated with Activity:
    service.name: WebApplication2
    service.instance.id: 2fe20641-f8df-4f9c-9604-6eaa3555c702
    telemetry.sdk.name: opentelemetry
    telemetry.sdk.language: dotnet
    telemetry.sdk.version: 1.12.0
```

docs/http/http-spans.md
RassK: I think the syntax changes some time. I cannot remember if the border was from ASP.NET MVC to ASP.NET Core.

`{controller}/{action}/{id?}` was the default route template for asp.net mvc (pre .net core versions).

That's why I also like stripping all framework specific tokens out of the route.

docs/registry/attributes/http.md
RassK: `The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans`

:int as a constraint is not the most general string. Although one can argue that if you have routes:
/articles/int:id/edit/
/articles/string:id/edit/

... then those should be treated as separate '(statistically) interesting class of Spans'. Which then breaks the rule about 'be the most general string'. Constraints can add / change some context.

RassK:
More about ASP.NET route tests here, you can see it's not simple, there are different methods and outcomes:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests

lmolkova:
I think I was able to repro. So the problem statement as I understand this: 

1. ASP.NET Core instrumentation does not use the best API to get actual routing data. E.g. when I use MVC app with default routing, current instrumentation populates default template when it shouldn't. See more on this below
2. Semconv is not precise enough to tell where info should come from. Suggested something in https://github.com/open-telemetry/semantic-conventions/pull/2734/files#r2323288932 to hopefully help with it.

More on p1: let's say I create 

```csharp
...
app.MapControllerRoute(
    name: "default",
    pattern: "{controller=Home}/{action=Index}/{id?}")
    .WithStaticAssets();
...

public class HomeController : Controller
{
    private readonly ILogger<HomeController> _logger;
    public HomeController(ILogger<HomeController> logger)  { _logger = logger; }

    public IActionResult Index() { return View();  }

    public IActionResult Privacy()
    {
        _logger.LogInformation("Privacy page requested, RoutePattern {}, routedata: controller {}, action {}",
            (HttpContext.GetEndpoint() as RouteEndpoint)?.RoutePattern.RawText,
            RouteData.Values.GetValueOrDefault("controller"),
            RouteData.Values.GetValueOrDefault("action"));
        return View();
    }
}
```

What I see in the log is 

```
info: MvcMovie.Controllers.HomeController[0]
      Privacy page requested, RoutePattern {controller=Home}/{action=Index}/{id?}, 
      routedata controller Home, action Privacy
```

OTel instrumentation uses `RoutePattern.RawText` which is `{controller=Home}/{action=Index}/{id?}` which is indeed wrong and useless, while `RouteData` is more precise.

To me it sounds like we need to figure out what the right behavior for ASP.NET Core instrumentation is - do we know it already?

docs/registry/attributes/http.md
lmolkova: So you'd like otel instrumentations to have a freedom to format routes as they see fit to optimize user experience while staying within the spec boundaries? That's fine, we can put this language in to semconv.  I don't believe nothing in HTTP semconv contradicts it though.

Specifying how exactly it should be done would be hard and would need extensive research across web frameworks and design. I don't see a huge benefit given that it very fx and instrumentation-specific.

RassK:
> To me it sounds like we need to figure out what's the right behavior for ASP.NET Core instrumentation is - do we know it already?

My answer would be general standardization to offer unified experience. `/users/get/{id}` or `/users/get/:id` would only benefit the developer, as they have the knowledge of the framework. Monitoring teams (who mostly purchases such products imo)  can monitor multiple frameworks, so they do not see it as a unified experience.

Why this little string even matters that much, is because of the overall metrics of `(statistically) interesting class of Spans`. In context of web frameworks, those metrics are similar like database query performance metrics (so you always want to group similar queries).

<ins>The other option is to detach span name from http.route and leave http.route as is provided by the framework.</ins>

docs/registry/attributes/http.md
lmolkova: explicitly added custom formatting in this suggestion https://github.com/open-telemetry/semantic-conventions/pull/2734/files#r2323288932

lmolkova:
@RassK could you folks please create an issue for ASP.NET Core instrumentation, suggest the fix that would make span names AND routes useful for ASP.NET Core and once you have consensus, we'd make sure semconv can accomodate it?

update: it's within ASP.NET Core instrumentation freedom on what to put into http.route. What it puts today does not seem to work well with MVC - let's fix it.

lmolkova:
@matt-hensley I think you need to re-generate markdown from yaml to make checks pass, could you please also undraft the PR? it looks great otherwise

matt-hensley:
> @matt-hensley I think you need to re-generate markdown from yaml to make checks pass, could you please also undraft the PR? it looks great otherwise

Saw those checks fail, resolving some issues with my Windows env and the docs tooling

docs/http/http-spans.md
lmolkova: resolving this, since examples are gone for now and we recommend instrumentations to document the format

model/http/registry.yaml
```diff
@@ -170,12 +170,19 @@ groups:
...
-        note: >
+            The matched route template for the request. This MUST be low-cardinality and include all static path segments, with dynamic path segments represented with placeholders.
+        examples: ['/users/:userID?', 'my-controller/my-action/{id?}']
+        note: |
           MUST NOT be populated when this is not supported by the HTTP server framework as the route attribute should have low-cardinality and the URI path can NOT substitute it.
-
           SHOULD include the [application root](/docs/http/http-spans.md#http-server-definitions) if there is one.
+
+          A static path segment is a part of the route template with a fixed, low-cardinality value. This includes literal strings like `/users/` and placeholders that
+          are constrained to a finite, predefined set of values `{controller}` or `{action}`.
```
trask:
```suggestion
          are constrained to a finite, predefined set of values, e.g. `{controller}` or `{action}`.
```

model/http/registry.yaml
```diff
@@ -170,12 +170,19 @@ groups:
...
+
+          A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like `{id}`.
```
trask:
```suggestion
          A dynamic path segment is a placeholder for a value that can have high cardinality and is not constrained to a predefined list like static path segments.
```

trask (approved):

Issue #2719, opened 2025-09-04
RPC: add network.protocol.name|version to spans and metrics


lmolkova:
subj

thompson-tomo:
@lmolkova  Should the following metrics also have the attributes:

 - rpc.server.requests_per_rpc
 - rpc.server.responses_per_rpc
 - rpc.client.requests_per_rpc
 - rpc.client.responses_per_rpc

Reason is the closest equivalent in http "http.client.active_requests"/"http.server.active_requests" does not have it.

thompson-tomo:
Note #2784 has been raised based on discussion.

PR #2736, opened 2025-09-05
Add attributes to each rpc metric via groups #2311

### LINKED ISSUE #2311: RPC metrics do not list any attributes
### Area(s)

area:rpc

### What's the problem?

When I generate RPC metrics using weaver all the metrics do not have any attributes associated with them.

https://cloud-native.slack.com/archives/C0697EXNTL3/p1748380674319339

[The attributes related to the RPC metrics](https://github.com/open-telemetry/semantic-conventions/blob/ab88f2a4edc2ddfb995d950b9b1c086ab554c0fe/model/rpc/metrics.yaml#L2-L16) to be included when rendering with weaver.

thompson-tomo:
Fixes #2311

## Changes

This links the attribute group directly to each metric they apply to.

Changes to explicit markdown is:
- Free-text has been added as notes to the applicable attributes and removed from free-text

Model changes:
- `network.type` has been removed from all RPC metrics
- `error.type` has been added to all RPC metrics


docs/rpc/rpc-metrics.md
```diff
@@ -512,13 +512,63 @@ to end-of-batch, it's hard to interpret in practice.
...
+| [`network.type`](/docs/registry/attributes/network.md) | string | [OSI network layer](https://wikipedia.org/wiki/Network_layer) or non-OSI equivalent. [4] | `ipv4`; `ipv6` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`rpc.method`](/docs/registry/attributes/rpc.md) | string | The name of the (logical) method being called, must be equal to the $method part in the span name. [5] | `exampleMethod` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`rpc.service`](/docs/registry/attributes/rpc.md) | string | The full (logical) name of the service being called, including its package name, if applicable. | `myservice.EchoService` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`server.address`](/docs/registry/attributes/server.md) | string | Server domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name. [6] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
lmolkova: we should also have  `error.type` - goes on pretty much any metric

docs/rpc/rpc-metrics.md
```diff
@@ -512,13 +512,63 @@ to end-of-batch, it's hard to interpret in practice.
...
 |---|---|---|---|---|---|
+| [`rpc.system`](/docs/registry/attributes/rpc.md) | string | A string identifying the remoting system. See below for a list of well-known identifiers. | `grpc`; `java_rmi`; `dotnet_wcf` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`server.port`](/docs/registry/attributes/server.md) | int | Server port number. [1] | `80`; `8080`; `443` | `Conditionally Required` [2] | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.transport`](/docs/registry/attributes/network.md) | string | [OSI transport layer](https://wikipedia.org/wiki/Transport_layer) or [inter-process communication method](https://wikipedia.org/wiki/Inter-process_communication). [3] | `tcp`; `udp` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`network.type`](/docs/registry/attributes/network.md) | string | [OSI network layer](https://wikipedia.org/wiki/Network_layer) or non-OSI equivalent. [4] | `ipv4`; `ipv6` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
```
lmolkova: do we actually need `network.type` on RPC metrics? I get they were here on attributes, but does it actually make sense? Would someone be really interested to see difference in RPC duration between ipv4 and ipv6? 

Errors / network io should be captured on the system level, plus there are means to see problems over spans where we'd capture actual IPs.

Transport makes at least some sense because you can run gRPC and maybe others on transports like unix domain sockets or named pipes and it might be significant. MCP (which is even higher abstraction) can run on stdio.

Type - I find it hard to find a good scenario where it's a meaningful dimension. RPC is also one layer on top of HTTP and even there address family is too low-level and not always visible / meaningful on the level instrumentation runs. 

Also, we have `network.protocol.name | version` that are consistently used on db and most importantly http. Given that rpc primarily runs on top of HTTP these two could replace `network.transport` in most cases (it's either TCP for http/2 or udp for http/3).

So the proposal is to at least remove `network.type` and (could be a follow up) add `network.protocol.name | version`.

docs/rpc/rpc-metrics.md
thompson-tomo: Sure have added it.

docs/rpc/rpc-metrics.md
thompson-tomo: hmm, i don't see a particular need and it appears the only location outside of network where it is used other than rpc is in kestrel. So have gone and deprecated it.

In relation to network.protocol.name/version, yes my plan was to tackle that via separate pr which is mostly done but have a question i was going to post on the issue. #2719

model/rpc/common.yaml
```diff
@@ -0,0 +1,34 @@
...
+          conditionally_required: If and only if the operation failed.
+  - id: attributes.metrics.rpc.client
+    type: attribute_group
+    brief: "Describes RPC metric attributes."
+    extends: attributes.metrics.rpc.server
+    attributes:
+      - ref: rpc.service
+        note: ""
+      - ref: server.port
+        note: Describes the server port they are connecting to.
```
lmolkova: who are `they`? is there a need to override note?

model/rpc/common.yaml
```diff
@@ -0,0 +1,34 @@
...
+        requirement_level:
+          conditionally_required: If the connection is IP-based and the port is available.
```
lmolkova: is there a need to update this note? why should it be different than in generic (or HTTP case)

model/rpc/common.yaml
```diff
@@ -0,0 +1,34 @@
...
+          the service name provided in `rpc.service` does not necessarily have to match the `service.name` resource attribute.
+
+          One process can expose multiple RPC endpoints and thus have multiple RPC service names.
+          From a deployment perspective, as expressed by the `service.*` resource attributes, it will be treated as one deployed service with one `service.name`.
+      - ref: rpc.method
+      - ref: network.transport
+      - ref: server.address
+      - ref: server.port
+        note: Describes the port the client is connecting from.
+        requirement_level: opt_in
```
lmolkova: seems address is recommended and port is opt-in - why?

model/rpc/common.yaml
```diff
@@ -0,0 +1,34 @@
...
+    attributes:
+      - ref: rpc.system
+        requirement_level: required
+      - ref: rpc.service
+        note: >
+          On the server process receiving and handling the remote procedure call,
+          the service name provided in `rpc.service` does not necessarily have to match the `service.name` resource attribute.
+
+          One process can expose multiple RPC endpoints and thus have multiple RPC service names.
+          From a deployment perspective, as expressed by the `service.*` resource attributes, it will be treated as one deployed service with one `service.name`.
```
lmolkova: is this note necessary? IS there a real case when rpc.service can be confused with service.name that cant be resolved by providing clear definition of what rpc.service is?


I get is was already there, but we don't need to carry over unnecessary parts

model/rpc/common.yaml
thompson-tomo: Have updated the note handling. Key thing is the note only applies to the client side.

model/rpc/common.yaml
thompson-tomo: Have updated address to be opt-in as per the http metrics.

model/rpc/common.yaml
thompson-tomo: Have gone and made changes to the notes/briefs to make it clearer this relates to the rpc interface perspective.

model/rpc/common.yaml
```diff
@@ -0,0 +1,33 @@
...
+    brief: "Describes RPC metric attributes."
+    extends: attributes.metrics.rpc.server
+    attributes:
+      - ref: server.address
+        note: When communicating through an intermediary, `server.address` SHOULD represent the server address behind any intermediaries, for example proxies, if it's available.
+        requirement_level: required
+      - ref: server.port
+        note: When communicating through an intermediary, `server.port` SHOULD represent the server port behind any intermediaries, for example proxies, if it's available.
+        requirement_level:
+          conditionally_required: If the connection is IP-based and the port is available.
```
lmolkova: nit, for brevity

```suggestion
          conditionally_required: If applicable.
```

model/rpc/common.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - id: attributes.metrics.rpc.server
+    type: attribute_group
+    brief: "Describes RPC metric attributes."
+    attributes:
+      - ref: rpc.system
+        requirement_level: required
+      - ref: rpc.service
+        note: ""
+      - ref: rpc.method
+        note: ""
```
lmolkova, line 11: why do we need to remove attribute notes on all method and service? if the notes are not right, let's create an issue to correct them, but overriding them everywhere does not seem necessary or right

model/rpc/common.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - ref: error.type
+        requirement_level:
+          conditionally_required: If and only if the operation failed.
+  - id: attributes.metrics.rpc.client
+    type: attribute_group
+    brief: "Describes RPC metric attributes."
+    extends: attributes.metrics.rpc.server
+    attributes:
+      - ref: server.address
+        note: When communicating through an intermediary, `server.address` SHOULD represent the server address behind any intermediaries, for example proxies, if it's available.
```
lmolkova: this is essentially a copy-paste of note in the definition, do we need to rewrite it? I see the reason we want to have custom note for server-side, can we write it in the way that we don't erase the original note and then have to bring it back?

model/rpc/common.yaml
thompson-tomo: I had been focussed on brevity ie not saying when observed from the client when we are looking at a client span but have taken this brevity away.

lmolkova: "observed from the client side" is on `server.*`, but here you're removing notes from the `rpc.*` - wh remove them at all and why specifically on metrics?

model/rpc/common.yaml
```diff
@@ -0,0 +1,31 @@
+groups:
+  - id: attributes.metrics.rpc.client
```
lmolkova, line 2: nit, if it's specific to metrics, it should probably be in the metrics file.

model/aws/sdk-spans.yaml
```diff
@@ -23,11 +23,13 @@ groups:
           - aws-api
       - ref: rpc.service
         brief: "The name of the service to which a request is made, as returned by the AWS SDK."
+        note: ""
```
lmolkova: is this change necessary, especially in the context of this PR?

thompson-tomo: Without it we would still be changing the notes in the markdown due to changing the base note. That review would highlight it as being inaccurate.

model/rpc/common.yaml
thompson-tomo: Ah, I miss read your comment.

The reason for removing them from metrics is that the notes talk about using an additional attribute which is fine for spans but is not for metrics.

Take the method note for example:

> The method can be different from the name of any implementing method/function. The `code.function.name` attribute may be used to store the latter (e.g., method actually executing the call on the server side, RPC client stub method on the client side).

The only part relevant for metrics is

> The method can be different from the name of any implementing method/function.

But that doesn't provide additional information especially considering the brief is:

> This is the logical name of the method from the RPC interface perspective.

model/aws/sdk-spans.yaml
lmolkova: the base note used to be there in markdown and should still be there with minor updates to the note.

image

The whole idea of using rpc in AWS conventions is questionable, we should tackle it separately and it's triaged to be post-stability - https://github.com/open-telemetry/semantic-conventions/issues/1545.

model/rpc/common.yaml
lmolkova: since you're already changing the note:

>> The method can be different from the name of any implementing method/function.

> But that doesn't provide additional information especially considering the brief is 

Any reason to keep this is in the base note then? 

> The code.function.name attribute may be used to store the latter (e.g., method actually executing the call on the server side, RPC client stub method on the client side).

Seems this part of the note should be taken from the base note and only applied on spans?

thompson-tomo: Hmm, I like that suggestion of moving that note to be on the rpc span only so will add it to the span via #2503 and allow this to focus on metrics.

lmolkova (approved):

thompson-tomo:
@trask could you please have a review as this is a Blocker for network.protocol.name & the deprecation PR.

trask (approved):

Issue #2720, opened 2025-09-04
RPC: rewrite system-specific pages to cover all conventions, not just tech-specific attributes


lmolkova:
See https://github.com/open-telemetry/semantic-conventions/blob/9ce82b8ced480416f677dfdff3b532d11b3fb05a/docs/rpc/json-rpc.md as an example

lmolkova:
Related: #2228, #2503

thompson-tomo:
Documents have been tackled via a pr per rpc system.

- #2503
- #2737
- #2738

PR #2737, opened 2025-09-05
Adds in an explicit Connect rpc spans #2720

thompson-tomo:
Progresses #2720

## Changes

Explicitly define the connect rpc spans based on base definition. The status message is now explicitly on the span


jsuereth:
Has this been discussed in RPC Semconv SIG?  cc @trask / @lmolkova ?

From my recollection on gRPC - it's unclear how we want to model this or if this portion of the spec was ever implemented.  I'd prefer to hold on RPC changes until the RPC SIG plan emerges.

thompson-tomo:
@jsuereth yes it was discussed and during the triage the issue was moved into the todo status. It is also why the jsonrpc PR was redone in a similar fashion of becoming span definitions rather than just attributes.

thompson-tomo:
Waiting for #2503 to be merged and then this can be rebased to be ready for review.

model/rpc/spans.yaml
```diff
@@ -35,6 +35,60 @@ groups:
...
+
+  - id: span.rpc.connect_rpc.server
+    type: span
+    stability: development
+    extends: rpc_service.server
+    span_kind: server
+    brief: This span represents an incoming Remote Procedure Call (RPC).
+    note: |
+      Remote procedure calls can only be represented with these semantic conventions
+      when the names of the called service and method are known and available.
```
lmolkova: I think we removed this sentence on some other PR, can we remove it here as well?

model/rpc/spans.yaml
```diff
@@ -35,6 +35,60 @@ groups:
...
+
+      `rpc.system` MUST be set to `"connect_rpc"` and SHOULD be provided **at span creation time.**
+
+      **Span name:** refer to the [Span Name](./rpc-spans.md#span-name) section.
```
lmolkova: nit, could you please use absolute path (/docs/rpc/...) so that it does not matter where the snippet is rendered, we still get valid link?

model/rpc/spans.yaml
```diff
@@ -35,6 +35,60 @@ groups:
...
+
+      **Span kind** MUST be `SERVER`.
+
+      **Span status**  If `rpc.connect_rpc.error_code` is set,
+      [Span Status](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.48.0/specification/trace/api.md#set-status) MUST be set to `Error` and left unset in all other cases.
```
lmolkova: it's usually more complicated than that (e.g. exception will result in no status code, but status would be an error), or `cancelled` code does not have to be an error and MUST is too strong. 

Let's create an issue to add a section where we'd describe this and remove span status from here for the time being. I believe it would be populated to something generic like [Recording Errors](/docs/general/recording-errors.md) link.

model/rpc/spans.yaml
```diff
@@ -35,6 +35,60 @@ groups:
...
+    attributes:
+      - ref: rpc.connect_rpc.error_code
+        requirement_level:
+          conditionally_required: If response is not successful and if error code available.
```
lmolkova:
```suggestion
          conditionally_required: if available.
```

?

`canceled` or `not_found` are not  necessarily errors, but should still be populated.

model/rpc/spans.yaml
thompson-tomo: Change has been made and the default text/link is now shown.

model/rpc/spans.yaml
thompson-tomo: Correct & I have raised a seperate PR #2810 to update it on the base for consistency.

model/rpc/spans.yaml
thompson-tomo: Have also fixed it on the other rpc pages via the consistency PR #2810.

model/rpc/spans.yaml
thompson-tomo: Issue to improve is #2811

model/rpc/spans.yaml
```diff
@@ -35,6 +35,48 @@ groups:
...
+    note: |
+      `rpc.system` MUST be set to `"connect_rpc"` and SHOULD be provided **at span creation time.**
+
+      **Span name:** refer to the [Span Name](/docs/rpc/rpc-spans.md#span-name) section.
+
+      **Span kind** MUST be `SERVER`.
+    attributes:
+      - ref: rpc.connect_rpc.error_code
+        requirement_level:
+          conditionally_required: if available.
```
trask, line 74: any need for this change?

```suggestion
          conditionally_required: If response is not successful and if error code available.
```

model/rpc/spans.yaml
```diff
@@ -35,6 +35,48 @@ groups:
...
+
+      **Span name:** refer to the [Span Name](/docs/rpc/rpc-spans.md#span-name) section.
+
+      **Span kind** MUST be `CLIENT`.
+    extends: rpc_service
+    span_kind: client
+    attributes:
+      - ref: rpc.connect_rpc.error_code
+        requirement_level:
+          conditionally_required: if available.
```
trask, line 53: (same)

```suggestion
          conditionally_required: If response is not successful and if error code available.
```

trask (approved):

model/rpc/spans.yaml
thompson-tomo: Feedback from @lmolkova https://github.com/open-telemetry/semantic-conventions/pull/2737#discussion_r2380733600

model/rpc/spans.yaml
thompson-tomo: Feedback from @lmolkova https://github.com/open-telemetry/semantic-conventions/pull/2737#discussion_r2380733600

lmolkova (approved):

PR #2739, opened 2025-09-05
Allow constants to be indicated via annotations

thompson-tomo:
## Changes

Introduce meaningful annotations for indicating constants ie db.system.name

With this change we gain:
- Attribute is listed in the attribute table including link to definition
- Attribute is included in sampling relevant list if applicable
- The need to override the note with info about expected db system etc is removed
- Code generation could enforce the value to be set
- reduced copy paste as attribute can now be defined much closer to the base item being extended.

What this solution doesn't solve but final solution will need to
- Restricting enum to subset of values
- Multi usage of metrics which requires different constant values

This should not be seen as the final solution but a stepping stone which offers an improvement to the status quo.

### DIFF
--- a/docs/browser/events.md
+++ b/docs/browser/events.md
@@ -24,6 +24,8 @@ The event name MUST be `browser.web_vital`.
 
 This event describes the website performance metrics introduced by Google, See [web vitals](https://web.dev/vitals).
 
+
+
 **Body fields:**
 
 :warning: Body fields will be moved to complex attributes once the
--- a/docs/database/mariadb.md
+++ b/docs/database/mariadb.md
@@ -26,14 +26,13 @@ linkTitle: MariaDB
 
 Spans representing calls to MariaDB adhere to the general [Semantic Conventions for Database Client Spans](/docs/database/README.md).
 
-`db.system.name` MUST be set to `"mariadb"` and SHOULD be provided **at span creation time**.
-
 **Span kind** SHOULD be `CLIENT`.
 
 **Span status** SHOULD follow the [Recording Errors](/docs/general/recording-errors.md) document.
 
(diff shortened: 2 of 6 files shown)


lmolkova (changes requested):
Using `example` as a source of constant value and `annotations` as a source of enum-ref-type is hacky. This syntax is used by 10+ OTel SDK to generate code from semantic conventions. We can't merge hacks and then come up with proper solution - it would be breaking for everybody who uses codegen.

Before sending PRs, please suggest a longer-term schema design in the https://github.com/open-telemetry/weaver/issues/479 and get some consensus.

thompson-tomo:
I agree it is not perfect but it is a big improvement to the current solution where we don't even associate the attribute to the span and just put it as a note. That means code-gen doesn't know about it as well as other limitations.

If I change the annotation path out of code_generation and into documentation for example would that address your concerns about it potentially becoming an issue for codegen?

I have previously stated which proposal I like for the new schema for enums and thoughts on subset of values. I will also share thoughts on implentation specific constants.

lmolkova:
@thompson-tomo please come up with a proposal for weaver on how to do it properly and help implementing it.

thompson-tomo:
I have shared details on my proposal for schema v2 on how to resolve the constants situation.

In relation to this given the benefits it brings to be able to have that association present in the definitions, is changing the annotation path sufficient? Or Does the attribute need to also be excluded from code generation to preserve current behaviour until it can be included while ensuring the expected value is used?

lmolkova:
I'm going to close this PR. I like the idea of having specialized solution to constant problem and I left a starting proposal on how to do it in https://github.com/open-telemetry/weaver/issues/479#issuecomment-3258660938

I don't believe further discussion here is helpful.

PR #2744, opened 2025-09-07
feat: Define app.screen.name

### LINKED ISSUE #2743: Add app.screen.name
### Area(s)

area:app

### What's missing?

As discussed during Client SIG on [7/22/2025](https://docs.google.com/document/d/16Vsdh-DM72AfMg_FIt9yT9ExEWF4A_vRbQ3jRNBe09w/edit?tab=t.0#heading=h.cgwi8r81t23c), we want to define a new semantic convention to capture screen names for applications. We've proposed the following: `app.screen.name`. Creating this issue to attach in follow up PR.

_No response_

### Tip

limhjgrace:
Fixes #2743

## Changes

There is currently no semantic convention for capturing application screen names. This PR proposes `app.screen.name`, as discussed during Client SIG on [07/22/2025](https://docs.google.com/document/d/16Vsdh-DM72AfMg_FIt9yT9ExEWF4A_vRbQ3jRNBe09w/edit?tab=t.0#heading=h.cgwi8r81t23c). 

### DIFF
--- a/docs/registry/attributes/app.md
+++ b/docs/registry/attributes/app.md
@@ -16,8 +16,9 @@ Describes attributes related to client-side applications (e.g. web apps or mobil
 | <a id="app-jank-threshold" href="#app-jank-threshold">`app.jank.threshold`</a> | double | The minimum rendering threshold for this jank, in seconds. | `0.016`; `0.7`; `1.024` | ![Development](https://img.shields.io/badge/-development-blue) |
 | <a id="app-screen-coordinate-x" href="#app-screen-coordinate-x">`app.screen.coordinate.x`</a> | int | The x (horizontal) coordinate of a screen coordinate, in screen pixels. | `0`; `131` | ![Development](https://img.shields.io/badge/-development-blue) |
 | <a id="app-screen-coordinate-y" href="#app-screen-coordinate-y">`app.screen.coordinate.y`</a> | int | The y (vertical) component of a screen coordinate, in screen pixels. | `12`; `99` | ![Development](https://img.shields.io/badge/-development-blue) |
-| <a id="app-widget-id" href="#app-widget-id">`app.widget.id`</a> | string | An identifier that uniquely differentiates this widget from other widgets in the same application. [3] | `f9bc787d-ff05-48ad-90e1-fca1d46130b3`; `submit_order_1829` | ![Development](https://img.shields.io/badge/-development-blue) |
-| <a id="app-widget-name" href="#app-widget-name">`app.widget.name`</a> | string | The name of an application widget. [4] | `submit`; `attack`; `Clear Cart` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="app-screen-name" href="#app-screen-name">`app.screen.name`</a> | string | The name of an application screen. [3] | `Home`; `Cart`; `MainActivity` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="app-widget-id" href="#app-widget-id">`app.widget.id`</a> | string | An identifier that uniquely differentiates this widget from other widgets in the same application. [4] | `f9bc787d-ff05-48ad-90e1-fca1d46130b3`; `submit_order_1829` | ![Development](https://img.shields.io/badge/-development-blue) |
+| <a id="app-widget-name" href="#app-widget-name">`app.widget.name`</a> | string | The name of an application widget. [5] | `submit`; `attack`; `Clear Cart` | ![Development](https://img.shields.io/badge/-development-blue) |
 
 **[1] `app.installation.id`:** Its value SHOULD persist across launches of the same application installation, including through application upgrades.
 It SHOULD change if the application is uninstalled or if all applications of the vendor are uninstalled.
@@ -39,6 +40,8 @@ More information about Android identifier best practices can be found in the [An
 
 **[2] `app.jank.frame_count`:** Depending on platform limitations, the value provided MAY be approximation.
 
-**[3] `app.widget.id`:** A widget is an application component, typically an on-screen visual GUI element.
+**[3] `app.screen.name`:** A screen represents a distinct view or page within an application,  typically containing multiple widgets or UI components.
 
-**[4] `app.widget.name`:** A widget is an application component, typically an on-screen visual GUI element.
+**[4] `app.widget.id`:** A widget is an application component, typically an on-screen visual GUI element.
+
+**[5] `app.widget.name`:** A widget is an application component, typically an on-screen visual GUI element.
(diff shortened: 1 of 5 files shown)


model/app/registry.yaml
```diff
@@ -59,6 +59,14 @@ groups:
           The y (vertical) component of a screen coordinate, in screen pixels.
         stability: development
         examples: [ 12, 99 ]
+      - id: app.screen.name
+        type: string
+        stability: development
+        brief: The name of an application screen.
+        note: >
+          A screen represents a distinct view or page within an application,
```
LikeTheSalad: I'm not sure "distinct view" is clear enough. Would that "view" have to cover the whole device's screen to be categorized as a screen? Or is it ok that there might be multiple "distinct views" on a device's screen at once, each one being considered a "screen" (which might happen on tablet UIs, for example)? I'm fine either way tbh, I just think we should make it clear here.

model/app/events.yaml
```diff
@@ -19,6 +19,8 @@ groups:
         requirement_level: required
       - ref: app.screen.coordinate.y
         requirement_level: required
+      - ref: app.screen.name
+        requirement_level: opt_in
```
thompson-tomo, line 25: Should we also have an app.screen.id which is required similar to the widget counterpart?

model/app/registry.yaml
limhjgrace: The latter. I was thinking "view" in terms of the Swift SDK for UIKit/SwiftUI views. These don't have to cover the entire screen, but is a larger scope than a singular widget. Updating notes and examples in Rev2 ([diff](https://github.com/open-telemetry/semantic-conventions/pull/2744/commits/6531a1b62b658a04056b42e918d30e8cb13d9e42)).

model/app/events.yaml
limhjgrace: I think I understand why we would want `app.widget.id` since there may be multiple instances of the same widget in even a single screen. And technically, the same applies to screens since there may be multiple instances of unique screens in an application. 

I'm hesitant, however, on whether we want to make this required. This is mostly coming from (1) the fact that in terms of application monitoring, having `app.screen.name` is sufficient (we're interested in the performance of specific unique screens, not individual instances) and (2) currently, OTel Android SDK only uses `screen.name` (and the ADOT Swift SDK I'm working uses `view.name`). 

But this may be a valid edge/use case that we haven't considered. If it is required, in the worst case, we could consider just using the same value for `.name` and `.id` but at that point it feels redundant to me. If it's still strongly recommended we add the `.id` counterpart though, I don't mind adding it.

thompson-tomo: So I have just looked into the Android implementation and it appears that it is using the getSimpleName method which returns just the class name without namespace/package. hence I don't see it as being a unique identifier. I would set the id to be the result of getCanonicalName hence becomes a unique identifier within the scope of the app.

model/app/registry.yaml
LikeTheSalad: Sounds good. Thank you.

model/app/events.yaml
lmolkova: the rule of thumb in semconv is to avoid defining attributes until you know they are necessary - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/how-to-write-conventions/README.md#defining-attributes

it's easy to add new attribute later. The cost of having something that might be useful (or just available) is too high (telemetry volume, instrumentations, semconv alignment, etc).

limhjgrace: Thanks for the insight. It seems like we've aligned with adding just `app.screen.name` for now since that's what the Android SDK uses (and what we plan to do in the ADOT Swift SDK as well as contribute to upstream Swift SDK). Will resolve the failing checks below in the next revision and hopefully that'll be enough to merge this PR.

thompson-tomo: I understand that but I would put name as opt-in and id as recommended. that way you gain a unique identifier for the screen on the event without increasing the attribute count if you don't want to. Alternatively the description of the name should indicate that the value should be unique.

model/app/registry.yaml
```diff
@@ -59,6 +59,17 @@ groups:
           The y (vertical) component of a screen coordinate, in screen pixels.
         stability: development
         examples: [ 12, 99 ]
+      - id: app.screen.name
+        type: string
+        stability: development
+        brief: The name of an application screen.
+        note: >
```
bidetofevil, line 83: So an app not taking up an entire device display makes sense. But are you also saying that an app can have multiple `app.screen.name` associated with it at one time as well that maps to different parts of the device display? If so, then what stops someone from using it to use it for a widget? By the letter of the definition, it doesn't seem to forbid that, in which case "screen" seems like the wrong sub-domain to put it in.

That is to say, is there a stronger definition we can give `screen`, especially one that works well for both this new attribute as well as the existing ones like `app.screen.coordinate.x` (which seem to define screen more like the way you've defined "device display")

docs/registry/attributes/app.md
```diff
@@ -39,6 +40,8 @@ More information about Android identifier best practices can be found in the [An
 
 **[2] `app.jank.frame_count`:** Depending on platform limitations, the value provided MAY be approximation.
 
-**[3] `app.widget.id`:** A widget is an application component, typically an on-screen visual GUI element.
+**[3] `app.screen.name`:** A screen represents a distinct view or page within an application, typically containing multiple widgets or UI components. A screen is larger in scope than individual widgets but may not fill the entire device display - multiple screens can coexist on the same display simultaneously.
```
naftaly: Apologies for chiming in a bit late, and if this has been discussed before. `screen` seems a bit misleading for the intended purpose due to it having another meaning in the real world (ie: display, monitor, the full area you're looking at on a device). I think something along the line of `surface` might be more appropriate.

williazz: Surface can also be misleading for the same reason as screen. Should we consider just `view.name`?

limhjgrace: I also think `app.view.name` might be a better since screen may be more commonly understood to be the device display but `surface` is not as intuitive to understand in this context.

model/app/registry.yaml
limhjgrace: I think we can merge this conservation with the one below, but the TL;DR is that we're proposing `app.view.name` instead.

model/app/events.yaml
limhjgrace: ~~We intend to propose `app.view.name` instead since a screen can be overloaded/misunderstood to be the device display. With that being said, I don't think this new attribute belongs in this event. Will remove from here and just define the attribute for now. It'll be added to the upcoming spans I plan on proposing.~~

Update: Aligned on `app.screen.name` during SIG with some adjustments required for the attribute note.

docs/registry/attributes/app.md
bidetofevil: As discussed in the SIG, `view` is even more problematic as it has specific meaning in both iOS and Android. That `surface` is a bit more nebulous allows the meaning to be more malleable, especially given that the definition describes there being multiple *noun*s being present at one time.

While I think `surface` is an improvement over `screen` because it connotes something less specific, I'm OK with the latter if we can restrict the definition to:

1. It representing only the part of the device display drawn by the app
2. There always being one 

And really, 1 is the part I feel most strongly about and eliminates the confusing between device display and the area the app has rendering control over. 2 I feel relatively less strong about and probably can be talked down from as being unnecessary. I still feel having it more strongly separates it from the `widget` definition, which I'd like, but not dead set on.

limhjgrace: I agree with 1 but am hesitant on 2 as @LikeTheSalad mentioned in the first thread. I'd prefer we don't add the latter requirement so that we can accommodate the split views use case on tablets.

limhjgrace: Updated `app.screen.name` note in rev 4 ([ref](github.com/open-telemetry/semantic-conventions/pull/2744/files/633b2487b7605520bf126a5312f2c7c8d7c437be..834c78e2fade0c04eeefdb7a73378f28c2b9c88c))

model/app/events.yaml
limhjgrace: Added `app.screen.id` as a recommended attribute in rev 4 ([ref](github.com/open-telemetry/semantic-conventions/pull/2744/files/633b2487b7605520bf126a5312f2c7c8d7c437be..834c78e2fade0c04eeefdb7a73378f28c2b9c88c))

breedx-splk (approved):
LGTM thank you!

PR #2754, opened 2025-09-09
Extend gen_ai content JSON Schemas with multimodal blob and file_data

aabmass:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1556

## Changes

Added two new types to the `MessagePart` union for capturing multimodal prompt/response data:
- `BlobPart` which contains inline base64.
- `FileDataPart` which contains a URI referencing data.

Also updated the ipynb to directly write to the JSON Schemas for simpler updating.

## Prototypes
- [Prototype in `opentelemetry-instrumentation-google-genai`](https://github.com/open-telemetry/opentelemetry-python-contrib/compare/main...aabmass:opentelemetry-python-contrib:adk-sample#diff-bb66127b5e9aa7fb3e0000b3f65b8614a2210b2229fe71dfe566a9b25ff20ea9)


alexmojaki:
Noting the questions I asked in https://github.com/open-telemetry/semantic-conventions/issues/1556#issuecomment-3144699449, especially:

> - Do we distinguish between inline data and URLs? Note that e.g. OpenAI's `image_url` field doesn't, it just uses a `data:` URL for inline data. This seems pretty simple and convenient. Maybe an additional boolean to indicate which it is? Or people could just check if the URL starts with `data:`. Alternatively, we have two different types, e.g. `media_inline` and `media_url`, or 2 per media type e.g. `image_inline` etc.

You've chosen two types:

```python
class BlobPart(BaseModel):
    type: Literal["blob"] = Field(description="The type of the content captured in this part.")
    mime_type: str = Field(description="The IANA MIME type of the attached data.")
    data: bytes = Field(description="base64 encoded bytes of the attached data.")

    class Config:
        extra = "allow"

class FileDataPart(BaseModel):
    type: Literal["file_data"] = Field(description="The type of the content captured in this part.")
    mime_type: str = Field(description="The IANA MIME type of the attached data.")
    file_uri: str = Field(description="A URI referencing to reference attached data. Should be recorded without modification, as it was sent to the model.")

    class Config:
        extra = "allow"
```

The only difference between these part types is the `data` vs `file_uri`. What do you think is the advantage of having an extra type for this instead of using data URIs containing base64?

aabmass:
> What do you think is the advantage of having an extra type for this instead of using data URIs containing base64?

Hadn't considered this before, mainly because I'm more familiar with the [Gemini format](https://github.com/googleapis/googleapis/blob/b246fb02ef3664ca126151bad614baeb6fbf47ba/google/cloud/aiplatform/v1/content.proto#L112-L116) that inspired this. But a few thoughts

- Having the type as `data: bytes` would encode to OTLP protobuf as [AnyValue.bytes_value](https://github.com/open-telemetry/opentelemetry-proto/blob/d1315d7f2c1504c36577a82cd42a73e3977fdd88/opentelemetry/proto/common/v1/common.proto#L38), so you don't have to pay base64 size overhead. I think this would apply to many storage formats like parquet vs requiring base64 URI in a string.

  The way I wrote the description requires base64, so I would need to tweak it.
- Using data URLs for raw bytes does feel a little arbitrary because you could make the same argument for text: `data:text/plain,This%20is%20a%20plain%20text%20string`. Ofc, the escaping is ugly and takes extra space.

alexmojaki:
I see bytes in the protobuf, but not the spec, although byte arrays are allowed: https://opentelemetry.io/docs/specs/otel/logs/data-model/#type-any

aabmass:
Is `bytes` not a byte array? This applies to attributes too https://opentelemetry.io/docs/specs/otel/common/attribute-type-mapping/#byte-sequences.

alexmojaki:
I guess that must be right and I misinterpreted.

Do we have any precedent/guidance for bytes that can end up inside a complex attribute on a span? Right now it has to convert to JSON, which means base64 encoding, and then to the backend the field will just look like a string. The backend then has to know that it should decode that based on semconv if it wants to end up with the same thing as if it had received the attribute in a log body using protobuf instead of JSON.

alexmojaki:
cc @lmolkova for the above question

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -84,6 +84,22 @@
...
+    "    mime_type: str = Field(description=\"The IANA MIME type of the attached data.\")\n",
+    "    data: bytes = Field(description=\"base64 encoded bytes of the attached data.\")\n",
+    "\n",
+    "    class Config:\n",
+    "        extra = \"allow\"\n",
+    "\n",
+    "class FileDataPart(BaseModel):\n",
+    "    type: Literal[\"file_data\"] = Field(description=\"The type of the content captured in this part.\")\n",
+    "    mime_type: str = Field(description=\"The IANA MIME type of the attached data.\")\n",
+    "    file_uri: str = Field(description=\"A URI referencing to reference attached data. Should be recorded without modification, as it was sent to the model.\")\n",
```
lmolkova: nit:

```suggestion
    "    file_uri: str = Field(description=\"A URI referencing attached data. Should be recorded without modification, as it was sent to the model.\")\n",
```

?

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -146,7 +164,8 @@
     "    pass\n",
     "\n",
     "# Print the JSON schema for the InputMessages model\n",
-    "print(json.dumps(InputMessages.model_json_schema(), indent=4))"
+    "with open(\"../gen-ai-input-messages.json\", \"w\") as file:\n",
+    "    print(json.dumps(InputMessages.model_json_schema(), indent=4), file=file)"
```
lmolkova, line 210: 👍

docs/gen-ai/gen-ai-input-messages.json
```diff
@@ -44,6 +79,34 @@
...
+                "mime_type": {
+                    "description": "The IANA MIME type of the attached data.",
+                    "title": "Mime Type",
+                    "type": "string"
+                },
+                "file_uri": {
+                    "description": "A URI referencing to reference attached data. Should be recorded without modification, as it was sent to the model.",
+                    "title": "File Uri",
+                    "type": "string"
+                }
```
lmolkova: OpenAI chat completions doesn't have the URL, but it has file_id and filename https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages-user-message-content

image

Responses API has all - name, id, url, data

image

Claude seems to be using file-id - https://docs.anthropic.com/en/api/messages#response-content-file-id and does not have URL.

wonder if we should introduce file_id to accomodate OpenAI chat completions and Anthropic or, if we think those belong in system-specific extensions, we should probably remove file_url. WDYT?

docs/gen-ai/gen-ai-output-messages.json
```diff
@@ -1,5 +1,62 @@
...
+        "BlobPart": {
+            "additionalProperties": true,
+            "properties": {
+                "type": {
+                    "const": "blob",
+                    "description": "The type of the content captured in this part.",
+                    "title": "Type",
+                    "type": "string"
+                },
+                "mime_type": {
```
lmolkova, line 12: would you envision image, audio, and video captured in the blob part? 

also, would we be able to infer proper mime type from OpenAI's vague mp3/wav https://platform.openai.com/docs/api-reference/responses/create#responses-create-input-input-item-list-item-input-message-content-input-audio-input_audio-format

image

lmolkova:
> I guess that must be right and I misinterpreted.
> 
> Do we have any precedent/guidance for bytes that can end up inside a complex attribute on a span? Right now it has to convert to JSON, which means base64 encoding, and then to the backend the field will just look like a string. The backend then has to know that it should decode that based on semconv if it wants to end up with the same thing as if it had received the attribute in a log body using protobuf instead of JSON.

It's temporary that we serialize to json string, byte array is a good long-term solution and would work nicely with spec changes that are in-flight - https://github.com/open-telemetry/opentelemetry-specification/pull/4651, specifically

https://github.com/open-telemetry/opentelemetry-specification/blob/4f92d3dce6224797045da43e680a1a3dcebeac89/specification/common/README.md?plain=1#L38

docs/gen-ai/gen-ai-input-messages.json
aabmass: np that seems fine as an optional field 👍 

One thing I didn't understand from the OpenAI docs is, when using `file_data`, should the data be a data URI or just base64 string.

docs/gen-ai/gen-ai-output-messages.json
aabmass: Ya that was the intention. Do you see a benefit in the OpenAI way having separate variants per each? You still need to include the format so I don't see a big benefit.

> OpenAI's vague mp3/wav

Good question. Unless I'm missing something we can just do a 1:1 mapping following https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types?

docs/gen-ai/gen-ai-input-messages.json
alexmojaki: based on https://github.com/open-telemetry/semantic-conventions/issues/1556#issuecomment-3144699449 (but please double check) you would use just the base64 data for audio, but a data url for images and files.

aabmass:
One other thing to note, we have [this code for non-complex attributes](https://github.com/open-telemetry/opentelemetry-python/blob/22d1fd1ca685ea8f47e193ab4b2fb7e88cc79a5a/opentelemetry-api/src/opentelemetry/attributes/__init__.py#L219-L224) which attempts to interpret python `bytes` as unicode and drops the attribute if not. I think this [predates bytes attributes in OTLP](https://github.com/open-telemetry/opentelemetry-python/issues/623#issuecomment-638885642). I think we can remove it independently of https://github.com/open-telemetry/opentelemetry-specification/pull/4651

alexmojaki:
How should this be instrumented? https://platform.openai.com/docs/guides/images-vision?api-mode=responses&format=base64-encoded#analyze-images

```python
response = client.responses.create(
    model="gpt-4.1",
    input=[
        {
            "role": "user",
            "content": [
                { "type": "input_text", "text": "what's in this image?" },
                {
                    "type": "input_image",
                    "image_url": f"data:image/jpeg;base64,{base64_image}",
                },
            ],
        }
    ],
)
```

A `FileDataPart` with `image_url` kept as is, or a `BlobPart` where the data URL is decoded?

aabmass:
My preference would be to generate a `Blob` for the reason above about bytes being serialized more efficiently. I appreciate the ambiguity though.

@alexmojaki do you have a strong preference for using data URLs?

alexmojaki:
Not strongly, but it feels like something should be stated in the PR.

docs/gen-ai/gen-ai-output-messages.json
lmolkova: we might need extra properties to describe unique properties of the image - e.g.  https://platform.openai.com/docs/api-reference/images/object lists quite a few. Having said this, I don't see any of such properties being used in the chat history API and binary data is generally useful, so it makes sense to start with binary data for images and if we discover we need a special type, we'll need to revisit it.

docs/gen-ai/gen-ai-input-messages.json
```diff
@@ -44,6 +79,34 @@
...
+                    "description": "The type of the content captured in this part.",
+                    "title": "Type",
+                    "type": "string"
+                },
+                "mime_type": {
+                    "description": "The IANA MIME type of the attached data.",
+                    "title": "Mime Type",
+                    "type": "string"
+                },
+                "file_uri": {
```
Kludex: Can this be called `uri`? The `file` part is redundant.

alexmojaki: Most parts use `content`, I think BlobPart should too instead of `data`, I wonder if maaaaybe FilePart should too?

docs/gen-ai/gen-ai-input-messages.json
```diff
@@ -44,6 +79,34 @@
...
+                    "description": "A URI referencing attached data. Should be recorded without modification, as it was sent to the model.",
+                    "title": "File Uri",
+                    "type": "string"
+                }
+            },
+            "required": [
+                "type",
+                "mime_type",
```
Kludex: Can we make the `mime_type` not required? Anthropic's image via URL looks like this:

```ts
export interface BetaURLImageSource {
  type: 'url';

  url: string;
}
```

alexmojaki: seems like a good reason to separate media type (image, audio, document) from format (mp3, wav, jpeg, png, pdf)

docs/gen-ai/gen-ai-input-messages.json
Kludex: I agree with `"content"`.

docs/gen-ai/gen-ai-input-messages.json
Kludex: Yeah, I understand the idea of having a generic file data part, but I think the idea is to match the data that most of the LLM providers send - so I agree with having the media type separate would help here.

I can bring more data regarding other providers like OpenAI and Gemini, if needed.

alexmojaki: oh i didn't mean separate `type` values, i meant something like https://github.com/open-telemetry/semantic-conventions/issues/1556#issuecomment-3144766619

docs/gen-ai/gen-ai-input-messages.json
alexmojaki: https://github.com/open-telemetry/semantic-conventions/issues/1556#issuecomment-3294324508 shows using `content` for a URL

docs/gen-ai/gen-ai-input-messages.json
aabmass: To make sure I understand, you're proposing
- add a separate media type here `image | video | ...`
- make mime type optional

?

alexmojaki: Yes that works, although it might feel odd if mime type always starts with the media type.

Or maybe mime type could be allowed to just be e.g. `image` if nothing more is known.

docs/gen-ai/gen-ai-input-messages.json
aabmass: IMO the JSON is slightly more human readable if the keys are descriptive.

```json
[
  {
    "type": "text",
    "content": "https://... is the url",
  },
  {
    "type": "uri",
    "uri": "https://...",
  },
]
```

alexmojaki: yes, `content` for a URI is definitely questionable, happy to leave that. but for BlobPart I think it should be `content`.

aabmass: Changed

- `FileData.file_uri` -> `FileData.uri`
- `Blob.data` -> `Blob.content`

docs/gen-ai/gen-ai-input-messages.json
aabmass: > Yes that works, although it might feel odd if mime type always starts with the media type.

Looking at the [Anthropic API docs](https://docs.claude.com/en/api/messages), those don't match up with official IANA mime types at all. It can be `thinking`, `document`, `tool_use`, etc. However, if you send base64 image or document to Anthropic, you must include the mime type in `media_type`.

So it's more analogous to the `type: 'text' | 'tool_call' | 'blob' | 'file_data'` we already have in the OTel json schemas.

To summarize




<table>
  <thead>
    <tr>
      <th>Provider</th>
      <th>API Construct</th>
      <th>Input Variant(s)</th>
      <th>Media Type Required?</th>
      <th>Notes</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td rowspan="2">Gemini</td>
      <td><code>Blob</code></td>
      <td>Inline bytes</td>
      <td>Yes (<code>mime_type</code>)</td>
      <td>For inline bytes, any type.</td>
    </tr>
    <tr>
      <td><code>FileData</code></td>
      <td>URLs</td>
      <td>Yes (<code>mime_type</code>)</td>
      <td>For URLs, any type.</td>
    </tr>
    <tr>
      <td rowspan="3">OpenAI</td>
      <td><code>input_image</code></td>
      <td>URL, Data URL</td>
      <td>In data url only</td>
      <td>For images.</td>
    </tr>
    <tr>
      <td><code>input_file</code></td>
      <td>File Upload (implied)</td>
      <td>No</td>
      <td>Only PDFs (?)</td>
    </tr>
    <tr>
      <td><code>input_audio</code></td>
      <td>Base64</td>
      <td>No</td>
      <td>Requires <code>format</code> (e.g., <code>mp3</code>, <code>wav</code>), not a data URL.</td>
    </tr>
    <tr>
      <td rowspan="7">Anthropic</td>
      <td rowspan="3"><code>image</code></td>
      <td>Base64</td>
      <td>Yes (<code>media_type</code>)</td>
      <td>Requires media type (e.g., <code>image/jpeg</code>).</td>
    </tr>
    <tr>
      <td>URL</td>
      <td>No</td>
      <td></td>
    </tr>
    <tr>
      <td>File</td>
      <td>No</td>
      <td></td>
    </tr>
    <tr>
      <td rowspan="4"><code>document</code></td>
      <td>Base64</td>
      <td>Yes (<code>media_type</code>)</td>
      <td>Requires <code>application/pdf</code>.</td>
    </tr>
    <tr>
      <td>Text</td>
      <td>Yes (<code>media_type</code>)</td>
      <td>Requires <code>text/plain</code>.</td>
    </tr>
    <tr>
      <td>URL</td>
      <td>No</td>
      <td></td>
    </tr>
    <tr>
      <td>File</td>
      <td>No</td>
      <td></td>
    </tr>
  </tbody>
</table>

alexmojaki: what i was really proposing was:

```
  media_type: 'image',
  format: 'jpeg',  // optional because it may be unknown
```

aabmass: What would you put for media_type if it's a "file"? Say you're calling the OpenAI Responses with [file_id](https://platform.openai.com/docs/guides/pdf-files#uploading-files).

```shell
curl "https://api.openai.com/v1/responses" \
    -H "Content-Type: application/json" \
    -H "Authorization: Bearer $OPENAI_API_KEY" \
    -d '{
        "model": "gpt-5",
        "input": [
            {
                "role": "user",
                "content": [
                    {
                        "type": "input_file",
                        "file_id": "file-6F2ksmvXxt4VdoqmHRw6kL"
                    },
                    {
                        "type": "input_text",
                        "text": "What is the first dragon in the book?"
                    }
                ]
            }
        ]
    }'
```

docs/gen-ai/gen-ai-input-messages.json
aabmass: - Added `file_id` for referenced file uploads
- Added `filename` for inline blob uploads

PTAL

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -84,6 +84,24 @@
...
+    "    content: bytes = Field(description=\"base64 encoded bytes of the attached data.\")\n",
+    "\n",
+    "    class Config:\n",
+    "        extra = \"allow\"\n",
+    "\n",
+    "class FileDataPart(BaseModel):\n",
+    "    type: Literal[\"file_data\"] = Field(description=\"The type of the content captured in this part.\")\n",
+    "    mime_type: Optional[str] = Field(default=None, description=\"The IANA MIME type of the attached data.\")\n",
+    "    uri: Optional[str] = Field(default=None, description=\"A URI referencing attached data.\")\n",
+    "    file_id: Optional[str] = Field(default=None, description=\"An identifier referencing a file that was pre-uploaded to the provider.\")\n",
```
alexmojaki: Having a uri vs having a file_id feel like 2 different types of part, unless it's possible to have both.

docs/gen-ai/gen-ai-input-messages.json
alexmojaki: I think you wouldn't put a media/mime type in that case, which might be another argument that it should be a separate part type.

aabmass: I'm going off what you said before.

> seems like a good reason to separate media type (image, audio, document) from format (mp3, wav, jpeg, png, pdf)

My understanding of the Responses API is that this would be "document".

alexmojaki: Well that was for the example given with `BetaURLImageSource` where we know it's an image but maybe nothing else.

I'm assuming you can upload any type of file and get a file_id, including images, not just documents.

docs/gen-ai/non-normative/models.ipynb
aabmass: I think it would be one or the other. You can make them mutually exclusive in the JsonSchema but it's hard to represent that in python types

docs/gen-ai/gen-ai-input-messages.json
aabmass: I would really have liked to use the standard IANA mime types, but document is not really a thing in. Separate part types might be better given the inference APIs we are modelling. WDYT of this?

```py
class FileDataPart(BaseModel):
    type: Literal["image", "audio", "video", "document"] = Field(description="The type of the content captured in this part.")
    mime_type: Optional[str] = Field(default=None, description="The IANA MIME type of the attached data.")
    source: BlobSource | FileSource | UriSource
    
    class Config:
        extra = "allow"

class BlobSource(BaseModel):
    type: Literal["blob"]
    data: bytes

class FileSource(BaseModel):
    type: Literal["file"]
    file_id: str

class UriSource(BaseModel):
    type: Literal["uri"]
    uri: str
```

docs/gen-ai/non-normative/models.ipynb
Kludex: FYI, you can represent it like this:
```python
from pydantic import BaseModel

class _FileDataPartUri(BaseModel):
    type: Literal["file_data"]
    mime_type: str | None
    uri: str

class _FileDataPartFileId(BaseModel):
    type: Literal["file_data"]
    mime_type: str | None
    file_id: str
    
type FileDataPart = _FileDataPartUri | _FileDataPartFileId
```

docs/gen-ai/gen-ai-input-messages.json
aabmass: > I'm assuming you can upload any type of file and get a file_id, including images, not just documents.

Based on trying the API, I think no it only supports pdf

<details><summary>error</summary>
<p>

```json
{
  "error": {
    "message": "Invalid input: Expected context stuffing file type to be a supported format: .pdf but got .jpg.",
    "type": "invalid_request_error",
    "param": "input",
    "code": null
  }
```

</p>
</details>

JamesNK:
I think you should update https://opentelemetry.io/docs/specs/semconv/gen-ai/non-normative/examples-llm-calls/ with one or more examples of capturing/providing the new message parts in calls.

For example, a speech to text example (blob data in). And an image generation example (blob data out).

docs/gen-ai/gen-ai-input-messages.json
lmolkova: > Added filename for inline blob uploads

~~I can see [`file_id`](https://platform.openai.com/docs/api-reference/responses/create#responses-create-input-input-item-list-item-input-message-content-input-image-file_id) for images, but not the filename. where does the `filename` for blobs come from?~~

----
Really weird to see filename in the blob part and not see it in the file part. Maybe we can call it `title` like anthropic does? https://docs.claude.com/en/api/messages#document ?

docs/gen-ai/gen-ai-input-messages.json
lmolkova: I tried with openai responses ("type": "input_image" instead of `input_file`) and it works with images ( .jpeg, .jpg, .png, .gif, .webp)

Giving it something else returns
```
{'error': {'message': 'Invalid input: Expected image type to be a supported format: .jpeg, .jpg, .png, .gif, .webp but got .log.', 'type': 'invalid_request_error', 'param': 'input', 'code': None}}
```

But it doesn't change the fact  - neither type, nor mime type are reliably available with FileSource / Part when `input_file` is used.

----

I like https://github.com/open-telemetry/semantic-conventions/pull/2754#discussion_r2370089822, but wonder if we need things to be deeply nested. Could we have more flat structure with `BlobPart`, `FilePart`, `UriPart`? Each of them could have optional `mime_type` and `media_type`. Even TextPart may benefit from mime type (application/json for structured outputs if we have a reliable way of knowing it's json).

docs/gen-ai/gen-ai-input-messages.json
aabmass: Ah you're right, filename/title is not used for images in either API. Will update.

docs/gen-ai/gen-ai-input-messages.json
aabmass: > Each of them could have optional `mime_type` and `media_type`

What were you imagining `media_type` to be though? Maybe the vendor specific category for the modality since they are not very consistent

alexmojaki: I think an important part is that the frontend of a telemetry consumer should have some idea of how to render a URL without relying on a full mime type, a file extension, or peeking at the content. The HTML `` tags don't need a mime type, e.g. this works:

```html
<img src="https://avatars.githubusercontent.com/u/1510004"/>
```

Assuming it's better to use `` where possible than e.g. `<embed>` for everything, then a consistent OTel name for each of those types is useful. Whereas 'document' isn't super meaningful.

lmolkova: > What were you imagining media_type to be though? Maybe the vendor specific category for the modality since they are not very consistent

I mean `image` | `video` | `audio`, if we flatten down and include it on `*Part` then `type` is taken by part type, we need a different name for content category field.

aabmass: Yes that works great for well defined types, but what about document? I'm ok to leave PDFs out of scope for now and address in a follow up if that sounds good.

lmolkova: discussion: consider modality for media_type

aabmass: > I'm ok to leave PDFs out of scope for now and address in a follow up if that sounds good.

Given that PDF/document will be addressed later, I'm going to remove filename for now too

docs/gen-ai/non-normative/models.ipynb
aabmass: Hmm I gave it a try but this just expands the main parts union. I was hoping to generate a `oneOf` json schema constraint

aabmass: The problem was it got flattened into the main union. I tried `RootModel` too and that didn't work. The only thing that worked was using `Discriminator` and manually adding tags. I think `json_schema_extra` solution generates a nicer json schema end result after all.

docs/gen-ai/gen-ai-input-messages.json
aabmass: I removed filename for now since it's only for PDFs. Follow up in https://github.com/open-telemetry/semantic-conventions/issues/2840

docs/gen-ai/gen-ai-input-messages.json
aabmass: Updated with a new modality field and added some examples. Follow up for PDFs: https://github.com/open-telemetry/semantic-conventions/issues/2840

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -84,6 +92,46 @@
...
+    "    model_config = ConfigDict(\n",
+    "        extra=\"allow\",\n",
+    "        json_schema_extra={\n",
+    "            \"oneOf\": [\n",
+    "                {\"required\": [\"file_id\"]},\n",
+    "                {\"required\": [\"uri\"]},\n",
+    "            ]\n",
+    "        }\n",
+    "    )\n",
+    "\n",
```
alexmojaki: Why do this instead of having two part types?

aabmass: https://github.com/open-telemetry/semantic-conventions/pull/2754#discussion_r2393362760

Did you mean with a new `type` value? I could do that but this seems easier to comprehend for readers of the json schema.

alexmojaki: yes. i don't see how this differs from blob parts.

aabmass: Passing by id or uri seems semantically the same. i'm not sure what I would name a 3rd type passed file id, any suggestion?

alexmojaki: UriPart and FilePart?

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -84,6 +92,46 @@
...
+    "        extra = \"allow\"\n",
+    "\n",
+    "class FileDataPart(BaseModel):\n",
+    "    \"\"\"Represents an external referenced file sent to the model\"\"\"\n",
+    "    type: Literal[\"file_data\"] = Field(description=\"The type of the content captured in this part.\")\n",
+    "    mime_type: Optional[str] = Field(default=None, description=\"The IANA MIME type of the attached data.\")\n",
+    "    modality: Union[Modality, str] = Field(\n",
+    "        description=\"The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.\"\n",
+    "    )\n",
+    "    uri: Optional[str] = Field(default=None, description=\"A URI referencing attached data.\")\n",
```
aabmass: Should specify that data URIs should be converted to raw bytes and stored in blob part.

aabmass: Addressed in 78f1131

docs/gen-ai/non-normative/models.ipynb
aabmass: Addressed in 78f1131 PTAL. Tbh I think I still prefer keeping one Part type for these since they are semantically very similar, although I understand the argument for keeping the structure consistent.

docs/gen-ai/non-normative/examples-llm-calls.md
```diff
@@ -201,6 +204,98 @@ Event:
...
+  {
+    "role": "user",
+    "parts": [
+      {
+        "type": "text",
+        "content": "What is in the attached data?"
+      },
+      // A image with a URI
+      {
+        "type": "file_data",
```
alexmojaki:
```suggestion
        "type": "uri",
```

docs/gen-ai/non-normative/examples-llm-calls.md
```diff
@@ -201,6 +204,98 @@ Event:
...
+      // A video with a vendor specific URI
+      {
+        "type": "file_data",
+        "modality": "video",
+        "mime_type": "video/mp4",
+        "uri": "gs://my-bucket/my-video.mp4"
+      },
+      // An image with opaque file ID e.g. the OpenAI files api
+      {
+        "type": "file_data",
```
alexmojaki:
```suggestion
        "type": "file",
```

similar for other examples

docs/gen-ai/non-normative/examples-llm-calls.md
```diff
@@ -201,6 +204,98 @@ Event:
...
+## Multimodal chat completion
+
+Multimodal chat completions follow the same sequence and telemetry structure as [simple chat
+completion](#simple-chat-completion) above, but contain additional types of Parts in the
+`gen_ai.input.messages` and `gen_ai.output.messages` span/event attributes:
+
+- `Blob` parts, which represent data sent inline to or from the model.
+- `FileData` parts, which represent a reference to a remote file.
+
+Both of these contain an optional `modality` field to capture the general category of the
```
alexmojaki: Needs updating

.chloggen/genai-multimodal-1556.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: gen_ai
+note: Extend gen_ai content JSON Schemas with multimodal blob and file_data
```
alexmojaki: needs updating

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -94,6 +110,37 @@
...
+    "        extra = \"allow\"\n",
+    "\n",
+    "class UriPart(BaseModel):\n",
+    "    \"\"\"Represents an external referenced file sent to the model by URI\"\"\"\n",
+    "    type: Literal[\"uri\"] = Field(description=\"The type of the content captured in this part.\")\n",
+    "    mime_type: Optional[str] = Field(default=None, description=\"The IANA MIME type of the attached data.\")\n",
+    "    modality: Union[Modality, str] = Field(\n",
+    "        description=\"The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.\"\n",
+    "    )\n",
+    "    uri: Optional[str] = Field(\n",
```
Kludex: This is not nullable anymore.

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -94,6 +110,37 @@
...
     "        extra = \"allow\"\n",
     "\n",
+    "class FilePart(BaseModel):\n",
+    "    \"\"\"Represents an external referenced file sent to the model by file id\"\"\"\n",
+    "    type: Literal[\"file\"] = Field(description=\"The type of the content captured in this part.\")\n",
+    "    mime_type: Optional[str] = Field(default=None, description=\"The IANA MIME type of the attached data.\")\n",
+    "    modality: Union[Modality, str] = Field(\n",
+    "        description=\"The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.\"\n",
+    "    )\n",
+    "    file_id: Optional[str] = Field(default=None, description=\"An identifier referencing a file that was pre-uploaded to the provider.\")\n",
```
Kludex: This is also not nullable anymore.

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -94,6 +110,37 @@
     "    class Config:\n",
     "        extra = \"allow\"\n",
     "\n",
+    "class FilePart(BaseModel):\n",
+    "    \"\"\"Represents an external referenced file sent to the model by file id\"\"\"\n",
+    "    type: Literal[\"file\"] = Field(description=\"The type of the content captured in this part.\")\n",
+    "    mime_type: Optional[str] = Field(default=None, description=\"The IANA MIME type of the attached data.\")\n",
```
Kludex, line 116: Would it be possible to represent this as `NotRequired` (using `TypedDict` and `TypeAdapter`)/`MISSING` instead of `nullable`? 

https://pydantic.dev/articles/pydantic-v2-12-release#missing-sentinel

aabmass: What is the benefit of `NotRequired` over `MISSING`?

also, there are many existing nullable fields in this schema already (from previous PRs).

Kludex: > What is the benefit of `NotRequired` over `MISSING`?

`NotRequired` can only be used with `TypedDict`. I'm not sure if there's a benefit on `MISSING` over `NotRequired`, but `MISSING` can be used on `BaseModel`.

---

> also, there are many existing nullable fields in this schema already (from previous PRs).

Which ones? I think that's the most correct representation tho - but I guess it's fine if that's the pattern used?

aabmass: Sorry I misspoke. I meant what is the benefit over just allowing it to be nullable? I agree it makes the JSON nicer. We are using dataclasses in the instrumentation, is there a good way to make `dataclasses.asdict()` omit None values?

> Which ones? I think that's the most correct representation tho - but I guess it's fine if that's the pattern used?

I guess it's just the tool call and tool call response `id` field.

Kludex: > I guess it's just the tool call and tool call response id field.

But I think those fields being nullable matches the providers.

> We are using dataclasses in the instrumentation, is there a good way to make dataclasses.asdict() omit None values?

Can I see a usage of `asdict()`? You can pass `dict_factory`, but if it's internal use, why do you use dataclasses and `asdict()` instead of using `TypedDict`s?

Kludex: > I meant what is the benefit over just allowing it to be nullable?

I know most of the people don't make a distinction, but I think it's more correct semantically.

---

All that said... This is not a blocker... If you want to proceed with nullable, it's fine.

aabmass: > But I think those fields being nullable matches the providers.

I thought it was because only some providers give IDs. Gemini doesn't give tool call ids, you just match them by tool name and index.

> You can pass dict_factory, but if it's internal use, why do you use dataclasses and asdict() instead of using TypedDicts?

They're generally just nicer to use/read and support default values. The dataclasses are being used directly by instrumentations 

I appreciate the concern and think it would apply to tool call IDs too, I filed https://github.com/open-telemetry/semantic-conventions/issues/2906

docs/gen-ai/non-normative/examples-llm-calls.md
```diff
@@ -201,6 +204,98 @@ Event:
...
+        "type": "uri",
+        "modality": "video",
+        "mime_type": "video/mp4",
+        "uri": "gs://my-bucket/my-video.mp4"
+      },
+      // An image with opaque file ID e.g. the OpenAI files api
+      {
+        "type": "file",
+        "modality": "image",
+        "mime_type": "image/png",
```
lmolkova: nit: are these realistically known ? doesn't seem so https://platform.openai.com/docs/guides/pdf-files?api-mode=responses#uploading-files

maybe

```suggestion
```

?

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -94,6 +101,45 @@
...
     "        extra = \"allow\"\n",
     "\n",
+    "class BlobPart(BaseModel):\n",
+    "    \"\"\"Represents blob binary data sent inline to the model\"\"\"\n",
+    "    type: Literal[\"blob\"] = Field(description=\"The type of the content captured in this part.\")\n",
+    "    mime_type: Optional[str] = Field(default=None, description=\"The IANA MIME type of the attached data.\")\n",
+    "    modality: Union[Modality, str] = Field(\n",
+    "        description=\"The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.\"\n",
+    "    )\n",
+    "    content: bytes = Field(description=\"Raw bytes of the attached data. This field SHOULD be encoded as a base64 string when transmitted as JSON.\")\n",
```
lmolkova: not sure I understand this along with `bytes` type : 

> This field SHOULD be encoded as a base64 string when transmitted as JSON

Is this to support current limitation that does not allow to set complex attrs on spans?
Assuming it's solved - do we want to recommend base64 or byte array for all cases? 

Or is it about model API and how content is transmitted to the app? 

-----
not a strong opinon, but:

It seems most, if not all models (maybe Google gRPC-based APIs use bytes?) do base64 encoding and it'd be natural for instrumentation to pass the string instead of decoding it.

lmolkova (commented):
Look great, with some final concents on bytes / base64 string for content

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -94,6 +101,45 @@
     "    class Config:\n",
     "        extra = \"allow\"\n",
     "\n",
+    "class BlobPart(BaseModel):\n",
+    "    \"\"\"Represents blob binary data sent inline to the model\"\"\"\n",
+    "    type: Literal[\"blob\"] = Field(description=\"The type of the content captured in this part.\")\n",
+    "    mime_type: Optional[str] = Field(default=None, description=\"The IANA MIME type of the attached data.\")\n",
+    "    modality: Union[Modality, str] = Field(\n",
+    "        description=\"The general modality of the data if it is known. Instrumentations SHOULD also set the mimeType field if the specific type is known.\"\n",
+    "    )\n",
```
alexmojaki, line 110: Would be nice to put mime_type and modality in a common base class

docs/gen-ai/non-normative/models.ipynb
lmolkova: SIG discussion, let's converge on bytes and figure out JSON encoding offline

docs/gen-ai/non-normative/examples-llm-calls.md
aabmass: Thanks good catch 👍

docs/gen-ai/non-normative/models.ipynb
lmolkova: I assumed that most json serialization libs would encode byte array as array of ints, but it seems it's not the case. Two libs I checked - Java's Jackson and .NET System.Text.Json both default to base64 strings. 

In this context, the original description makes sense, I'd suggest a tiny change though:

```suggestion
    "    content: bytes = Field(description=\"Raw bytes of the attached data. This field SHOULD be encoded as a base64 string when serialized to JSON.\")\n",
```

When we declare the BlobPart in language-specific API, we have some control over how `content` property is serialized, but not how it's transmitted

Cirilla-zmh: Even though we declare content as a bytes type, due to potential encoding conflicts, it remains essentially a base64-encoded string. Under this premise, storing the raw bytes data directly within a structured JSON is often not very helpful for data consumption — evaluators or data processors would still need to decode it back into a `byte[]` during processing.

Perhaps we could allow users to define a `pre_upload_hook`, enabling these byte files to be uploaded directly to external storage. The instrumentation could then record this message as a `UriPart` (or add additional metadata indicating that it was uploaded via a hook). This would make those base64 strings more meaningful — they could be consumed as standalone files with clear references, and evaluators or data processors could retrieve them in a manner similar to how they handle `UriPart`.

If you're interested in the above proposal, we’d be happy to provide some prototype implementations.

gyliu513 (approved):

docs/gen-ai/non-normative/models.ipynb
lmolkova: let's follow-up on this https://github.com/open-telemetry/semantic-conventions/issues/3003

PR #2755, opened 2025-09-09
[chore] Clarify deprecation of http.flavor

lcian:
## Changes

It seems that `http.flavor` has been split into two attributes: `network.protocol.name` and `network.protocol.version`.
This PR changes the deprecation message so that this is more clearly reflected.

It would also be possible to change the `deprecated.reason` into `obsoleted`, like we do for other attributes that are split
https://github.com/open-telemetry/semantic-conventions/blob/f7bac20256a5da6969912965820be14be3186a1e/model/http/deprecated/registry-deprecated.yaml#L39
However I'm not sure if this would be considered a breaking change or could have other unexpected effects on the consumers of the conventions, so I haven't modified it for now.
	

model/http/deprecated/registry-deprecated.yaml
```diff
@@ -128,7 +128,7 @@ groups:
               value: 'QUIC'
               brief: 'QUIC protocol.'
               stability: development
-        brief: 'Deprecated, use `network.protocol.name` instead.'
+        brief: 'Deprecated, use `network.protocol.name` and `network.protocol.version` instead.'
         deprecated:
           reason: renamed
           renamed_to: network.protocol.name
```
lmolkova:
```suggestion
          reason: uncategorized
          note: "Split into `network.protocol.name` and `network.protocol.version`"
```

lmolkova (approved):
Great catch @lcian ! you're right we also need to change the reason - left a suggestion. This attribute is not stable along with schema-transformation processor in the collector that could make use of it. 

So I think we should fix it even though it might change behavior of something downstream.

lcian:
Thanks @lmolkova, applied the suggestion and regenerated!
Do we need a changelog?

lmolkova:
> Thanks @lmolkova, applied the suggestion and regenerated! Do we need a changelog?

thank you! I don't think we need a changelog for this, so all good

PR #2756, opened 2025-09-09
[cloud] Add Akamai Cloud (Linode) as cloud entity

paulojmdias:
Relates to: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42544

## Changes

This PR introduces new `cloud.provider` and `cloud.platform` values for Akamai Cloud (akamai_cloud) and Akamai Cloud Compute (akamai_cloud_compute), previously known as Linode.
These additions align with the new Linode resource detection processor in the Collector and ensure consistent attribute naming across all resources.


matt-hensley:
Forwarded this PR to friends at Linode/Akamai. They're going to have someone review and provide feedback/approval.

paulojmdias:
Thank you @matt-hensley 

I'm not sure if this shouldn't be named now as Akamai Cloud, so their input will be amazing

cmcgalliard:
Hey there,
Thanks for adding us to the list as a cloud entity! We'd prefer that we be considered Akamai first, so a slug of `akamai_cloud` and referenced as `Akamai Cloud (Linode)`

paulojmdias:
@cmcga1125 updated accordingly

model/cloud/registry.yaml
```diff
@@ -32,6 +32,10 @@ groups:
               value: 'ibm_cloud'
               brief: 'IBM Cloud'
               stability: development
+            - id: 'akamai_cloud'
```
```suggestion
            - id: 'akamai'
```

Same for this one. I'm not sure we need the "cloud" here. the vendor/provider is Akamai, unless it's a trademark name one has to use when talking about it I'd keep the short one.

model/cloud/registry.yaml
```diff
@@ -107,6 +111,10 @@ groups:
       - id: cloud.platform
         type:
           members:
+            - id: akamai_cloud_compute
```
```suggestion
            - id: akamai.cloud_compute
```

model/cloud/registry.yaml
cmcgalliard: It appears they're following the pattern from the entities above, e.g., `ibm_cloud`, which might confuse due to Akamai's different segments. So, leaving it `akamai_cloud` is probably a good idea?

model/cloud/registry.yaml
cmcgalliard: Maybe `akamai_cloud` to match the ones above?

model/cloud/registry.yaml
paulojmdias: @cmcga1125 I think we should keep `akamai` on the `cloud.provider` and `akamai.cloud_compute` on the `cloud.platform` due to what @joaopgrassi said above.

I updated accordingly with my previous sentence.

model/cloud/registry.yaml
joaopgrassi: No, we want to properly use namespaces to group cloud vendor "products". The older names are not right and should be used as example.

joaopgrassi:
@cmcga1125 just to clarify things: (apologies, I didn't get at first you are from Akamai) if you think Akamai as a "Cloud provider" should be named `akamai_cloud` because that's how the company wants to be called, then it is fine to go that route.

My initial review was based on the feeling @paulojmdias  used `akamai_cloud` because of the existing values, which we want to get rid of. We want to use namespaces (`.`) so all the existing ones will be changing at some point. Underscores are used on multi-word things, so if the provider name is "Akamai Cloud" then it is correct to use `akamai_cloud`. 

Let me know if it isn't clear.

cmcgalliard:
Yeah, my concern would be confusion between our different products that Akamai offers. I want to be specific; we're discussing [Akamai Cloud](https://www.akamai.com/cloud). Although I'm unsure if we would use `akamai_cdn` or `akamai_security` in this context, clearly labeling it as `akamai_cloud` would be beneficial.

joaopgrassi:
> Yeah, my concern would be confusion between our different products that Akamai offers. I want to be specific; we're discussing [Akamai Cloud](https://www.akamai.com/cloud). Although I'm unsure if we would use `akamai_cdn` or `akamai_security` in this context, clearly labeling it as `akamai_cloud` would be beneficial.

Then what do you all think on using `akamai_cloud` for the provider, and then `akamai_cloud.compute` for the platform?

cmcgalliard:
I'm good with that!

paulojmdias:
@joaopgrassi @cmcga1125 updated accordingly.

PR #2757, opened 2025-09-09
[cloud] Add Vultr as cloud entity

paulojmdias:
Relates to https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42570

## Changes

This PR introduces a new `cloud.provider` and `cloud.platform` values for Vultr Cloud (vultr_cloud) and Vultr Cloud Compute (vultr_cloud_compute).
These additions align with the new Vultr resource detection processor in the Collector and ensure consistent attribute naming across all resources.


model/cloud/registry.yaml
```diff
@@ -40,6 +40,10 @@ groups:
               value: 'tencent_cloud'
               brief: 'Tencent Cloud'
               stability: development
+            - id: 'vultr_cloud'
```
joaopgrassi: Is Vultr cloud the name of the cloud provider? From the website I can only see Vultr as the listed name.

model/cloud/registry.yaml
```diff
@@ -297,6 +287,10 @@ groups:
               value: 'tencent_cloud_scf'
               brief: Tencent Cloud Serverless Cloud Function (SCF)
               stability: development
+            - id: vultr_cloud_compute
```
```suggestion
            - id: vultr.cloud_compute
```

model/cloud/registry.yaml
```diff
@@ -297,6 +287,10 @@ groups:
...
+              value: 'vultr_cloud_compute'
```
```suggestion
              value: 'vultr.cloud_compute'
```

model/cloud/registry.yaml
paulojmdias: I was thinking this was a standard now. But I agree, and I updated to be only `vultr`.

model/cloud/registry.yaml
paulojmdias: Updated

model/cloud/registry.yaml
paulojmdias: Updated

lmolkova (approved):
Could you please also review other cloud attributes (specifically `cloud.resource_id`, `cloud.region`) and document any specifics on how they should be populated for Vultr?

PR #2758, opened 2025-09-09
[cloud] Add Hetzner as cloud entity

paulojmdias:
Relates to https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42477

## Changes

This PR introduces a new `cloud.provider` and `cloud.platform` values for Hetzner Cloud (hetzner_cloud) and Hetzner Cloud Compute (hetzner_cloud_compute).
These additions align with the new Hetzer resource detection processor in the Collector and ensure consistent attribute naming across all resources.


model/cloud/registry.yaml
```diff
@@ -28,6 +28,10 @@ groups:
               value: 'heroku'
               brief: 'Heroku Platform as a Service'
               stability: development
+            - id: 'hetzner_cloud'
```
```suggestion
            - id: 'hetzner.cloud'
```

We are moving away from the `_` as separator here and preferring a proper namespacing using a `.`, specially when the provider has multiple "products". So, I'd recommend doing it here.

paulojmdias: Updated as requested

model/cloud/registry.yaml
```diff
@@ -263,8 +263,8 @@ groups:
               value: 'gcp_openshift'
               brief: Red Hat OpenShift on Google Cloud
               stability: development
-            - id: hetzner_cloud_compute
-              value: 'hetzner_cloud_compute'
+            - id: hetzner.cloud.compute
```
```suggestion
            - id: hetzner.cloud_compute
```

Do they have an actual "cloud compute" product? Looking at their website, https://www.hetzner.com/cloud/ I only see "cloud server" as what seems the product name. I don't know much about them, so maybe it's fine but I'd like to be sure we did our research before having to rename it again later.

model/cloud/registry.yaml
```diff
@@ -28,8 +28,8 @@ groups:
               value: 'heroku'
               brief: 'Heroku Platform as a Service'
               stability: development
-            - id: 'hetzner_cloud'
-              value: 'hetzner_cloud'
+            - id: 'hetzner.cloud'
```
```suggestion
            - id: 'hetzner'
```

Why not just Hetzner?

model/cloud/registry.yaml
paulojmdias: image

I'm ok with renaming to `hetzner.cloud_server` or even to `hetzner.cloud`. WDYT?

model/cloud/registry.yaml
paulojmdias: Agree. Renamed

model/cloud/registry.yaml
joaopgrassi: Tricky yeah.. product names can come and go, so we have to pick something. Just cloud is too broad I think so.. let's go with `cloud_server`. At least it matches 1:1 with what they have on their product website page.

paulojmdias: Updated to `hetzner.cloud_server`

joaopgrassi (approved):

PR #2764, opened 2025-09-11
Document that *.system.name enums are not exhaustive and should not be added without conventions

### LINKED ISSUE #2751: System names: document open-enum approach for db.system.name, messaging.system.name, etc
SIG meeting notes:
- db.system.name is open
- experimental instrumentations can report arbitrary values not documented in semconv
- we recommend full docs, not just constant name
- undocumented system-support can be enabled in generic instrumentations (e.g. jdbc, entity framework) via feature flag.

lmolkova:
Fixes #2751

Adds more guidance on how to define attributes - in this case how to approach *.system.name enums:
- no need to be exhaustive
- constant along with convention is encouraged
- only stable instrumentations need to document all constants

### DIFF
--- a/docs/how-to-write-conventions/README.md
+++ b/docs/how-to-write-conventions/README.md
@@ -176,7 +176,7 @@ for naming guidance.
 > [!IMPORTANT]
 > System identifier enums don't need to list every possible system, component, or technology.
 >
-> OpenTelemetry instrumentations that use enum attributes SHOULD document their values before releasing **stable** artifacts.
+> OpenTelemetry instrumentations that use enum attributes SHOULD document their values before releasing **stable** artifacts and MAY support undocumented values in unstable artifacts or behind a feature flag.
 
 Only define new system identifiers when you also document how conventions apply
 to that system. For example, when adding a new `db.system.name` value, create
--- a/docs/how-to-write-conventions/README.md
+++ b/docs/how-to-write-conventions/README.md
@@ -151,7 +151,7 @@ Consider the scope of the attribute and how it may evolve in the future:
 > - If the attribute represents something useful in a narrow set of scenarios or
 >   is specific to certain system metrics, spans, or events, it likely does not need to be generic.
 
-##### Defining enum attribute members
+#### Defining enum attribute members
 
 Enum attributes generally fall into three main categories:
 
--- a/docs/how-to-write-conventions/README.md
+++ b/docs/how-to-write-conventions/README.md
@@ -12,7 +12,7 @@ linkTitle: How to define new semantic conventions
   - [Best practices](#best-practices)
     - [Prototyping](#prototyping)
     - [Defining attributes](#defining-attributes)
-      - [Defining enum attribute members](#defining-enum-attribute-members)
+    - [Defining enum attribute members](#defining-enum-attribute-members)
     - [Defining spans](#defining-spans)
     - [Defining metrics](#defining-metrics)
     - [Defining entities](#defining-entities)
--- a/docs/how-to-write-conventions/README.md
+++ b/docs/how-to-write-conventions/README.md
@@ -12,6 +12,7 @@ linkTitle: How to define new semantic conventions
   - [Best practices](#best-practices)
     - [Prototyping](#prototyping)
     - [Defining attributes](#defining-attributes)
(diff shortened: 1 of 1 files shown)


docs/how-to-write-conventions/README.md
```diff
@@ -150,6 +151,38 @@ Consider the scope of the attribute and how it may evolve in the future:
...
+both follow general database conventions, but populate attributes like `db.collection.name` differently and have system-specific attributes. Each system has its own span
+definitions and documentation.
+See [MongoDB](/docs/database/mongodb.md) and [MySQL](/docs/database/mysql.md) as examples
+and check out [system-specific naming](../general/naming.md#system-specific-attributes)
+for naming guidance.
+
+> [!IMPORTANT]
+> System identifier enums don't need to list every possible system, component, or technology.
+>
+> OpenTelemetry instrumentations that use enum attributes SHOULD document their values before releasing **stable** artifacts.
```
thompson-tomo: It would be helpful if we could cater for instrumentations that target multiple db systems ie jdbc, ef core with some not defined providers.

KalleOlaviNiemitalo:
In the PR title, I think "and not exhaustive" should be "are not exhaustive".

docs/how-to-write-conventions/README.md
lmolkova: cater how? someone reading it should know that they can release something unstable (artifact or behind a feature flag) for undocumented databases. Is there some specific wording that would help clarify it further?

thompson-tomo: Perhaps something like:
> OpenTelemetry instrumentations that use enum attributes SHOULD document atleast their highly used/most common values before releasing stable artifacts. Additional values may be enabled via feature flag or updated unstable artifact's.

lmolkova:
```suggestion
> OpenTelemetry instrumentations that use enum attributes SHOULD document their values before releasing **stable** artifacts and MAY support undocumented values in unstable artifacts or behind a feature flag.
```

lmolkova: updated to 

>OpenTelemetry instrumentations that use enum attributes SHOULD document their values before releasing **stable** artifacts and MAY support undocumented values in unstable artifacts or behind a feature flag.

docs/how-to-write-conventions/README.md
```diff
@@ -150,6 +151,38 @@ Consider the scope of the attribute and how it may evolve in the future:
 > - If the attribute represents something useful in a narrow set of scenarios or
 >   is specific to certain system metrics, spans, or events, it likely does not need to be generic.
 
+##### Defining enum attribute members
```
joaopgrassi:
```suggestion
#### Defining enum attribute members
```
Maybe like this? It is very small now. I think it can be its own section.

docs/how-to-write-conventions/README.md
```diff
@@ -150,6 +151,38 @@ Consider the scope of the attribute and how it may evolve in the future:
...
+both follow general database conventions, but populate attributes like `db.collection.name` differently and have system-specific attributes. Each system has its own span
+definitions and documentation.
+See [MongoDB](/docs/database/mongodb.md) and [MySQL](/docs/database/mysql.md) as examples
+and check out [system-specific naming](../general/naming.md#system-specific-attributes)
+for naming guidance.
+
+> [!IMPORTANT]
+> System identifier enums don't need to list every possible system, component, or technology.
+>
+> OpenTelemetry instrumentations that use enum attributes SHOULD document their values before releasing **stable** artifacts and MAY support undocumented values in unstable artifacts or behind a feature flag.
```
pellared, line 179: Should it not be part of https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/telemetry-stability.md?

lmolkova: I think it already is

> Stable instrumentations authored by OpenTelemetry SHOULD NOT produce telemetry that is
not described by OpenTelemetry semantic conventions.

https://github.com/open-telemetry/opentelemetry-specification/blob/1e04e1be0e17cae6d01c862049bbeb298e0ffa06/specification/telemetry-stability.md?plain=1#L62-L63

PR #2766, opened 2025-09-11
[chore] Use annotations for tables

thompson-tomo:
Fixes #

## Changes


docs/resource/os.md
```diff
@@ -49,6 +49,17 @@ In case of virtualized environments, this is the operating system as it is obser
...
 | `zos` | IBM z/OS | ![Development](https://img.shields.io/badge/-development-blue) |
+---
+
+**Sourcing Attribute Values:**
+
+SHOULD the value not be provided natively via the SDK the following sources can be used:
+
+| Attribute | Implementation | Location |
+| --- | --- | --- |
+| [`os.build_id`](/docs/registry/attributes/os.md) | Linux | `BUILD_ID` from `/etc/os-release` or `/usr/lib/os-release` |
```
kamphaus: The previous note listed `contents of '/proc/sys/kernel/osrelease'` as a third fallback location.
Was this meant to be removed as part of this PR?

kamphaus (commented):
In my opinion having a single Sourcing Attribute Values table is more readable than multiple notes
especially for several attributes like in [entities/os.md](https://github.com/open-telemetry/semantic-conventions/blob/d75ec9a8e01b3966c712bda081b77e12b882174b/docs/registry/entities/os.md).

Taking this information from structured yaml instead of note will also allow future improvements like filtering.

I also agree with what Josh said in Monday's SemConv SIG meeting that ideally it would be possible to see the note / sourcing attribute value for a given attribute in the attribute table by hovering over an icon or `[1]` note text or by expanding a collapsed section.

thompson-tomo:
I also agree that it would be nice solution however we are constrained to functionality which is supported in markdown hence unable to do those advanced layouts currently.

docs/resource/os.md
thompson-tomo: Fixed

lmolkova:
annotations are intended for extensibility. Here they are used for core functionality.  While the intent is nice, this would need more work on finding proper structure such as having windows/linux/etc refinement for the entity. this is what we currently do for spans such as [database](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/database) ones. If entities need a different approach, this needs to be discussed further.

Given lack of interest and support for this PR and that the implementation details don't align with schema language, I'm going to close it.

PR #2770, opened 2025-09-12
Add `start_time ` and `end_time` to session events

### LINKED ISSUE #2769: Add `start_time` and `end_time` to session events
### Area(s)

area:session

### Context

Sessions are defined as user sessions that expire after a period of inactivity. This is how it has been implemented in OTel Android and iOS.

1. Android - https://github.com/open-telemetry/opentelemetry-android/blob/7afdd119fa77adbaa2bf29c49a2404f4027b0b99/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdTimeoutHandler.kt#L28
2. iOS - https://github.com/open-telemetry/opentelemetry-swift/pull/899

This also fits the "time based expiration" strategy defined by Google - https://support.google.com/analytics/answer/2731565?hl=en#zippy=%2Cin-this-article

### Problem

The problem with this approach is that we don't know when these sessions truly expire, since semantic convention for `session.end` only includes the default observed timestamp provided out of the box. 

* Observed timestamp of `session.end` (aka `timeUnixNano`) is when the session end event was observed and logged. 
* However, true `session.end` should be calculated as the start of the final period of inactivity. This should be thought of as the last timestamp when the session was extended. (It can also be thought of as `expiryTime` minus `sessionTimeout`). 

### Solution

I propose to add `session.end_time` to session end events as a mandatory field (in unix nanoseconds), because the current observed timestamp does not accurately describe the actual session lifecycle. 

### Tip

williazz:
Fixes #2769

## Changes

Sessions start and end events are currently marked by their observed timestamps; however, these do not represent the true session lifecycle. Session are implement with time-based expirations that represent a user session. This means that session `start_time` and `end_times` occur earlier than when they are finally logged. 

1. `timeUnixNano` - timestamp of when the session start and end events are logged. 
2. `session.start_time` - timestamp of when the session was first created. This occurs either on the initial app start, or when a previous session was observed to be expired. 
3. `session.end_time` - timestamp of when the session ended. This can be thought of as the last timestamp during which the session was extended, or as session expire time minus session timeout interval. 

## Diff

`session.start` event
1. Add `session.start_time`

`session.end` event
1. Add `session.start_time`
2. Add `session.end_time`
3. Add `session.previous_id` (seemed strange that this was previously missing).

* [ x ] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ x ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
* [ x ] Links to the prototypes or existing instrumentations (when adding or changing conventions)

### DIFF
--- a/model/session/registry.yaml
+++ b/model/session/registry.yaml
@@ -13,6 +13,11 @@ groups:
       When a session reaches end of life, typically due to user inactivity or session timeout, a new session identifier
       will be assigned. The previous session identifier may be provided by the instrumentation so that telemetry
       backends can link the two sessions.
+
+      When a session expires, then the end time is calculated by the expiry timestamp minus the session timeout interval.
+      This is not necessarily the same as when session events are recorded, since they are only created after a period of inactivity.
+      To account for this difference, session `start_time` and `end_time` represent the true session lifecycle, whereas `timeUnixNano`
+      marks when session events are observed.
     attributes:
       - id: session.id
         type: string
@@ -24,3 +29,13 @@ groups:
         stability: development
         brief: "The previous `session.id` for this user, when known."
         examples: "00112233-4455-6677-8899-aabbccddeeff"
+      - id: session.start_time
+        type: int
+        stability: development
+        brief: "The timestamp when the session started (in unix nanoseconds)."
+        examples: 1757348655674899200
+      - id: session.end_time
+        type: int
+        stability: development
+        brief: "The timestamp when the session expired (in unix nanoseconds)."
+        examples: 1757348656658491100
--- a/.chloggen/feat_sessions.yaml
+++ b/.chloggen/feat_sessions.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: session
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `start_time` and `end_time` to session events to represent true lifecycle.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [2769]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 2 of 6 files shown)


model/session/registry.yaml
```diff
@@ -24,3 +29,13 @@ groups:
...
+      - id: session.start_time
+        type: int
+        stability: development
+        brief: "The timestamp when the session started (in unix nanoseconds)."
+        examples: 1757348655674899200
+      - id: session.end_time
+        type: int
+        stability: development
+        brief: "The timestamp when the session expired (in unix nanoseconds)."
+        examples: 1757348656658491100
```
thompson-tomo, line 41: Type should be string in ISO 8601 format as per https://github.com/open-telemetry/semantic-conventions/blob/main/docs/how-to-write-conventions/README.md.

thompson-tomo, line 41: Feel this should be session.end.time to potentially allow for other attributes in the future to describe the ending of the session.

model/session/registry.yaml
```diff
@@ -24,3 +29,13 @@ groups:
         stability: development
         brief: "The previous `session.id` for this user, when known."
         examples: "00112233-4455-6677-8899-aabbccddeeff"
+      - id: session.start_time
+        type: int
+        stability: development
+        brief: "The timestamp when the session started (in unix nanoseconds)."
+        examples: 1757348655674899200
```
thompson-tomo, line 36: When would this differ from the event time? Could we perhaps go with an event.occured attribute which would be reusable by other signals?

breedx-splk:
I don't agree with this -- the intention of the session events is to report when a session starts and when a session ends. The timestamp of the event is the time that the session start/end event happened. Why complicate that? If an implementation reports something different, then it will in accurate. 

I don't think that we should try and account for that with additional (and frankly confusing) timestamps.

williazz:
The problem can be solved by using `timestamp` vs `observedTimestamp`, so I am closing this issue. 

> I don't agree with this -- the intention of the session events is to report when a session starts and when a session ends. The timestamp of the event is the time that the session start/end event happened. Why complicate that? If an implementation reports something different, then it will in accurate.

The problem is already implied by otel convention for timestamp vs observedtimestamp, so I don't mean to over-complicate things. So for example, session end should probably be reported with timestamp=expiry and observedTimestamp=now. This is a good idea because by the time you realize your session had expired, it could be days later and cause your session data to look very strange.

PR #2771, opened 2025-09-12
Service instance is now standalone entity #2880

### LINKED ISSUE #2880: Decide if `service` and `service.instance` are different entities
Right now, we have two stabilized "resource"s in semantic conventions.

- `service` - includes `service.name`, `service.namespace`, etc.
- `telemetry.sdk`

However the `service.instance.id` field is NOT stable, but marked as required and an identifying field.  

I propose two options:

1. We stabilize `service.instance.id` and make it identifying on service.
2. We create a new entity `service.instance`.  There would be a "has-a" relationship between `service` and `service.instance`.  Initially, `service.instance` would only consist of its id.   `service` would be identified by `namespace` and `name`.

UPDATE:

We will take option (2) and open in semconv.

thompson-tomo:
Fixes #2880

## Changes

Splits the service entity to have seperate instance entity as described in https://github.com/orgs/open-telemetry/projects/85/views/1?pane=issue&itemId=86371458


docs/registry/entities/service.md
```diff
@@ -11,19 +11,40 @@
...
+**Identifying Attributes:**
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`service.instance.id`](/docs/registry/attributes/service.md) | string | The string ID of the service instance. [3] | `627cc493-f310-47de-96bd-71410b7dec09` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+**[3] `service.instance.id`:** MUST be unique for each instance of the same `service.namespace,service.name` pair (in other words
 `service.namespace,service.name,service.instance.id` triplet MUST be globally unique). The ID helps to
 distinguish instances of the same service that exist at the same time (e.g. instances of a horizontally scaled
 service).
```
KalleOlaviNiemitalo, line 50: `service.instance.id` is the only identifying attribute of the `service.instance` entity, but it alone does not suffice for identifying the entity; one also needs the `service.namespace` and `service.name` attributes of the `service` entity.  Are there already other entity types whose identifying attributes are incomplete in this manner?

thompson-tomo: Actually it does suffice and is complete as it's purpose is to identify an instance of the service and per the docs it is recommended to use a guid. 

Eventually a relationship between service & Service.instance entity would be defined.

KalleOlaviNiemitalo: If `service.instance.id` is required to suffice for identifying the service instance, even without `service.namespace` and `service.name`, then this text should be replaced:

> **`service.instance.id`:** MUST be unique for each instance of the same `service.namespace,service.name` pair (in other words `service.namespace,service.name,service.instance.id` triplet MUST be globally unique).

Are there currently any relations between any entities?  I tried searching in <https://github.com/open-telemetry/weaver> for how a relation would be represented in YAML but did not find.

thompson-tomo: I am not quite seeing what is wrong with that text, as that triplet still must be globally unique. If someone was to use the same instance id for multiple services, that would also be ok. 

> Are there currently any relations between any entities? I tried searching in https://github.com/open-telemetry/weaver for how a relation would be represented in YAML but did not find.

Relations are not currently definable.

PR #2783, opened 2025-09-17
Add triage document

### LINKED ISSUE #2699: [Project management] Introduce a well-defined triage process
Semconv has received a lot of [activity lately](https://opentelemetry.devstats.cncf.io/d/1/activity-repository-groups?orgId=1&var-period=m&var-repogroups=All&from=now-6M&to=now), and while it is awesome to see the community excited about it, it has been challenging on maintainers to keep up with the stream of work.

The semconv project lacks a well defined triage process which contributes to the overall fatigue and frustrations on both sides - contributors and maintainers.

We should have a documented and implemented triage process that defines from start to end what happens, when it happens, and who is responsible at each stage. This offers a more transparent and unbiased process for contributors, as they will be able to clearly understand, say, if their issue has been rejected. 

A clear triage process hopefully also lift some of the burden on maintainers, as they can rely on the process to guide their decisions and actions. 

The work here includes:

- Define a workflow that lays out the lifecycle of issues/PRs from start to finish
  - for example, something like it was proposed in the spec repo: https://github.com/open-telemetry/opentelemetry-specification/issues/3821 
- Define labels and what they mean and where they should be used
- ~~Make things as automated as possible (e.g., applying labels, moving things on project boards, assigning to SIGs based on label, etc)~~ To be done as a follow up, once the general triage is agreed.

joaopgrassi:
Fixes #2699

## Changes

This PR introduces the issue management document, where the issue triage workflow is defined, along with roles and labels involved in the process.

The goal is to get agreement on this first, and next start implementing automation to enforce it. 

### DIFF
--- a/issue-management.md
+++ b/issue-management.md
@@ -53,8 +53,9 @@ graph TD
     SIG_REJECTED --> REJECTED([Label: triage:rejected:declined])
     L_READY_SIG([Label: **triage:accepted:ready-with-sig**])
 
-    NEEDS_INFO -->|No, small in scope| L_READY([Label: **triage:accepted:ready**])
-    NEEDS_INFO -->|No, SIG issue | L_READY_SIG
+    NEEDS_INFO -->|No| L_CHECK_SCOPE{What's the scope of the issue?}
+    L_CHECK_SCOPE -->|Small or Trivial| L_READY([Label: **triage:accepted:ready**])
+    L_CHECK_SCOPE -->|Other| SIG_EXISTS{Does a SIG/project exist?}
 
     %% Define Classes
     classDef styleStart fill:#c6dcff,stroke:#498bf5,stroke-width:2px,color:#000,font-size:1.2em;


issue-management.md
```diff
@@ -0,0 +1,148 @@
...
+    FEEDBACK_PROVIDED --> SIG_EXISTS{Does a SIG/project exist?}
+    SIG_EXISTS -->|Yes| L_READY_SIG
+    SIG_EXISTS -->|No| L_NEEDS_SIG([Label: **triage:accepted:needs-sig**])
+    L_NEEDS_SIG --> SIG_FORMED[SIG/Project formed]
+    SIG_FORMED --> L_READY_SIG
+    L_NEEDS_SIG --> SIG_REJECTED[SIG rejected issue]
+    SIG_REJECTED --> REJECTED([Label: triage:rejected:declined])
+    L_READY_SIG([Label: **triage:accepted:ready-with-sig**])
+
+    NEEDS_INFO -->|No, small in scope| L_READY([Label: **triage:accepted:ready**])
```
jsuereth: I think you could make a new node for "NEEDS_INFO -->|No|".

E.g.

```
NEEDS_INFO -->|No| L_CHECK_SCOPE{What's the scope of the issue?}
L_CHECK_SCOPE -->|Small or Trivial| L_READY([Label: **triage:accepted:ready**])
L_CHECK_SCOPE -->|Other| SIG_EXISTS{Does a SIG/project exist?}
```

This just splits some of the decisions to keep them simple, but it's not necessary.

jsuereth (approved):

issue-management.md
joaopgrassi: Makes sense! Done in cc7c08a3

joaopgrassi:
> the diagram text is cut off in several places, but it's all in the markdown so not a big deal if it's hard to fix

Really? The docs says text should wrap automatically. For me it looks alright but maybe is one of my GH extensions 🤔 
image

trask:
ah, looks fine when I use Firefox, 🚢

issue-management.md
```diff
@@ -0,0 +1,149 @@
...
+
+Exceptions to this rule apply only to "trivial" pull requests, which include the
+following:
+
+- Pull requests originating from SIGs.
+- Pull requests generated by automation (e.g., Renovate).
+- Pull requests containing trivial changes (e.g., typos, styling-only updates,
+  tooling improvements).
+
+All pull requests that do not meet these requirements will be automatically
```
lmolkova, line 124: nit: is it automatic? can we actually detect above automatically?

joaopgrassi: I think so, but some parts may be complicated, like looking if the PR author is part of a SIG for ex. I'm thinking about the options at the moment and experimenting. Not fun with GH actions though :D

PR #2785, opened 2025-09-18
[chore] Move messaging notes to yaml

thompson-tomo:
## Changes

This moves a couple of attribute notes for kafka/rocketmq from being freetext markdown to be notes in the yaml and generated,

No changes in content have been done.


docs/messaging/kafka.md
```diff
@@ -66,14 +66,15 @@ For Apache Kafka, the following additional attributes are defined:
...
 | [`messaging.message.id`](/docs/registry/attributes/messaging.md) | string | A value used by the messaging system as an identifier for the message, represented as a string. | `452a7c7c7c7048c2f887f61572b18fc2` | `Recommended` If span describes operation on a single message. | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`server.port`](/docs/registry/attributes/server.md) | int | Server port number. [10] | `80`; `8080`; `443` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
-| [`messaging.message.body.size`](/docs/registry/attributes/messaging.md) | int | The size of the message body in bytes. Only applicable for spans describing single message operations. [11] | `1439` | `Opt-In` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`peer.service`](/docs/registry/attributes/peer.md) | string | The [`service.name`](/docs/resource/README.md#service) of the remote service. SHOULD be equal to the actual `service.name` resource attribute of the remote service if any. [11] | `AuthTokenCache` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
trask: it may be worth waiting until attribute groups gets worked out and we come up with a clean story for documenting these more generic attributes

otherwise it seems odd why we would add `peer.service` here and not to http, database, rpc, etc

thompson-tomo: so it seemed even more odd to me that we have a note describing how an attribute should be used but the attribute is not even part of the definition. Anyway have moved the note to be on the group to avoid adding the missing attribute and making it more obvious.

thompson-tomo:
@joaopgrassi with the new automation how can we do chores like this which is just moving free form text to be part of the model so that it can be consistently included/rendered as well as to improve readability?

trask:
Closing as the messaging SIG is not currently active to review it

PR #2805, opened 2025-09-25
Update RPC stability warning

### LINKED ISSUE #2718: Update stability warning for RPC semconv
The RPC stabilization effort is kicking off, and so we need to add/update the stability warning on RPC semantic conventions.

trask:
Fixes #2718

Copied the warning over from Database semantic conventions.

### DIFF
--- a/docs/rpc/rpc-metrics.md
+++ b/docs/rpc/rpc-metrics.md
@@ -35,29 +35,28 @@ metrics can be filtered for finer grain analysis.
 
 > **Warning**
 > Existing RPC instrumentations that are using
-> [v1.20.0 of this document](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/metrics/semantic_conventions/rpc-metrics.md)
+> [v1.37.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/rpc/rpc-metrics.md)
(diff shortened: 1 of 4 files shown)


lmolkova (approved):

docs/rpc/rpc-metrics.md
```diff
@@ -35,29 +35,28 @@ metrics can be filtered for finer grain analysis.
 
 > **Warning**
 > Existing RPC instrumentations that are using
-> [v1.20.0 of this document](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/metrics/semantic_conventions/rpc-metrics.md)
+> [v1.37.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/rpc/rpc-metrics.md)
```
Kielek, line 38: Version update here is intentional?

trask: yeah, it was intentional to try to keep the story simpler, instead of trying to document how to handle version bumps for both network attributes and RPC attributes, which could be bumped separately, e.g. Java already major version bumped and is emitting the stable network attributes as part of RPC semantic conventions

Kielek: Make sense to me. It will be strange to emit so old http/network attributes.

PR #2806, opened 2025-09-25
Add browser navigation event

martinkuba:
## Changes

This adds a browser event for capturing navigation actions.

Proposed [here](https://github.com/open-telemetry/opentelemetry-browser/issues/3#issuecomment-3288597619)


docs/browser/browser-events.md
```diff
@@ -51,4 +51,72 @@ semantic convention tooling supports complex attributes
...
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `browser.navigation`.
+
+This event represents a browser navigation action.
+
+| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|
+| [`browser.navigation.same_document`](/docs/registry/attributes/browser.md) | boolean | A boolean that is true if the navigation is within the same document [1] | `true`; `false` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
```
BenoitZugmeyer, line 71: **question:** Any reason to set this attribute as required? I'm worried about browser support here, since the Navigation API is not yet supported in Firefox and Safari. 

At the same time, I'm also wondering about the relevance of this attribute: are we expecting the SDK to collect navigation entries related to past documents? Shouldn't those entries already be collected when the past document was active, in which case this attribute would always be true?

martinkuba: This indicates whether the navigation resulted in loading a new document. I felt it was important to capture because it is the main indicator to specifically distinguish hard page loads.

For Firefox and Safari, we can instrument the History API. Any call to `history.pushState` or `history.replaceState` should be captured as same_document = true, while the event captured when the page is first loading should be same_document = false.

docs/browser/browser-events.md
```diff
@@ -51,4 +51,72 @@ semantic convention tooling supports complex attributes
...
+| [`url.full`](/docs/registry/attributes/url.md) | string | Absolute URL describing a network resource according to [RFC3986](https://www.rfc-editor.org/rfc/rfc3986) [2] | `https://www.foo.bar/search?q=OpenTelemetry#SemConv`; `//localhost` | `Required` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`browser.navigation.hash_change`](/docs/registry/attributes/browser.md) | boolean | A boolean that is true if the navigation is a fragment navigation. [3] | `true`; `false` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`browser.navigation.type`](/docs/registry/attributes/browser.md) | string | The type of navigation [4] | `push`; `replace`; `reload`; `traverse` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
vigneshshanmugam, line 74: Not sure how much of a benefit we would have in resembling the Navigation event from Browser spec which is basically a History change activity. For Spec, we would be better off capturing this as `hard-navigation`, `soft-navigation`, `route-change`, `click` etc that triggered the navigations. 

I don't think we need to be in 1-1 with Browser Spec. 

If we take this route, it would be easier to group all the navigations that are associated within a Session in the backend. 

We can also eliminate all the other hashChange, sameDocument fields are the SDK can basically trigger a pageView event dictating whether it was a navigation event that happened due to this type.

docs/browser/browser-events.md
BenoitZugmeyer: > This indicates whether the navigation resulted in loading a new document.

Forgive me if I'm wrong as I am new to the Navigation API, but from my tests it doesn't seem to be the case. `sameDocument` indicates that the `NavigationHistoryEntry` was on the current document (also the [doc](https://developer.mozilla.org/en-US/docs/Web/API/NavigationHistoryEntry/sameDocument) seems to confirm this).

Test:

* load https://example.com/
* in the devtools, look at `navigation.currentEntry.sameDocument`. It should be `true` (even if it's a whole new document)
* Make a soft navigation using `history.pushState(null, '', 'foo')` then look at `navigation.entries()`. A new navigation entry was added to the list, and both entries have `sameDocument: true`
* Make a hard navigation using `location.href = '/'` then look at `navigation.entries()` again. A new navigation entry with `sameDocument: true` was added to the list, and the first two entries `sameDocument` became `false`.

BenoitZugmeyer: Ooh I understand your point of view as you are referring to the `navigateEvent.destination.sameDocument` property. In this case yes it distinguishes between soft and hard navigations, but I don't think we should collect hard navigations this way? The hard navigation should be collected after the page is loaded, no?

PR #2819, opened 2025-09-29
[chore] Automatically split attribute tables based on deprecation status

thompson-tomo:
## Changes

This splits the attribute tables based on the deprecation status.

This provides the following benefits:
- deprecated attributes no longer needs to be moved to seperate groups simplifying process and improving history
- The attributes on signals are grouped the same as the registry
- The deprecated tables can have a different structure with links to new attributes

Outstanding tasks:
- consolidating deprecated attributes in to main file
- Putting informative briefs back on attributes rather than deprecation reason.


trask:
> consolidating deprecated attributes in to main file

There are pros and cons to having a single yaml for both deprecated and non-deprecated attributes

thompson-tomo:
@joaopgrassi can you reopen, when I reverted the changes I missed a carriage return.

Also shouldn't android be owned by mobile/client sig etc?

joaopgrassi:
I'm sorry but did we discussed this change? I'm not aware that we wanted to do any of these changes. Where did the need of this came from? I think such changes needs to be discussed and agreed first before we start sending PRs.

thompson-tomo:
Yes it was discussed when we discussed the changes to the attribute tables. The feedback was to avoid needing to update the yaml which is what I have done.

By having a different layout for deprecated we can hide irrelevant columns or even add additional. That is what has been done here.

Note when a deprecated is rendered on a signal it can now have a clickable link to the replacement. This logic could be added to the table in the attribute registry.

Anyway it won't be possible to re-open due to force pushing the typo fix. So have #3016

Issue #2301, opened 2025-05-26
Review updowncounter naming in k8s


lmolkova:
See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/k8s-metrics.md, plenty of  UpdownCounters are pluralized

PR #2822, opened 2025-09-29
[k8s] Fix updowncounters to follow the non-pluralization rule

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/2301

## Changes

This brings names in alignment with the pluralisation guidelines: https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/general/naming.md#do-not-pluralize-updowncounter-names

  Renamed metrics:
  - k8s.node.allocatable.pods -> k8s.node.pod.allocatable
  - k8s.daemonset.current_scheduled_nodes -> k8s.daemonset.node.current_scheduled
  - k8s.daemonset.desired_scheduled_nodes -> k8s.daemonset.node.desired_scheduled
  - k8s.daemonset.misscheduled_nodes -> k8s.daemonset.node.misscheduled
  - k8s.daemonset.ready_nodes -> k8s.daemonset.node.ready
  - k8s.job.active_pods -> k8s.job.pod.active
  - k8s.job.failed_pods -> k8s.job.pod.failed
  - k8s.job.successful_pods -> k8s.job.pod.successful
  - k8s.job.desired_successful_pods -> k8s.job.pod.desired_successful
  - k8s.job.max_parallel_pods -> k8s.job.pod.max_parallel
  - k8s.cronjob.active_jobs -> k8s.cronjob.job.active
  - k8s.deployment.desired_pods -> k8s.deployment.pod.desired
  - k8s.deployment.available_pods -> k8s.deployment.pod.available
  - k8s.replicaset.desired_pods -> k8s.replicaset.pod.desired
  - k8s.replicaset.available_pods -> k8s.replicaset.pod.available
  - k8s.replicationcontroller.desired_pods -> k8s.replicationcontroller.pod.desired
  - k8s.replicationcontroller.available_pods -> k8s.replicationcontroller.pod.available
  - k8s.statefulset.desired_pods -> k8s.statefulset.pod.desired
  - k8s.statefulset.ready_pods -> k8s.statefulset.pod.ready
  - k8s.statefulset.current_pods -> k8s.statefulset.pod.current
  - k8s.statefulset.updated_pods -> k8s.statefulset.pod.updated
  - k8s.hpa.desired_pods -> k8s.hpa.pod.desired
  - k8s.hpa.current_pods -> k8s.hpa.pod.current
  - k8s.hpa.max_pods -> k8s.hpa.pod.max
  - k8s.hpa.min_pods -> k8s.hpa.pod.min

### DIFF
--- a/model/k8s/deprecated/metrics-deprecated.yaml
+++ b/model/k8s/deprecated/metrics-deprecated.yaml
@@ -8,8 +8,8 @@ groups:
     stability: development
     deprecated:
       reason: renamed
-      renamed_to: k8s.replicationcontroller.desired_pods
-    brief: "Deprecated, use `k8s.replicationcontroller.desired_pods` instead."
+      renamed_to: k8s.replicationcontroller.pod.desired
+    brief: "Deprecated, use `k8s.replicationcontroller.pod.desired` instead."
     instrument: updowncounter
     unit: "{pod}"
   - id: metric.k8s.replication_controller.available_pods
@@ -21,8 +21,8 @@ groups:
     stability: development
     deprecated:
       reason: renamed
-      renamed_to: k8s.replicationcontroller.available_pods
-    brief: "Deprecated, use `k8s.replicationcontroller.available_pods` instead."
+      renamed_to: k8s.replicationcontroller.pod.available
+    brief: "Deprecated, use `k8s.replicationcontroller.pod.available` instead."
     instrument: updowncounter
     unit: "{pod}"
   - id: metric.k8s.replicationcontroller.desired_pods


docs/non-normative/k8s-migration.md
```diff
@@ -107,7 +107,7 @@ The changes between collector implementation and semantic conventions:
 | `k8s.node.allocatable_cpu`                 (type: `gauge`)                         | `k8s.node.allocatable.cpu`               (type: `updowncounter`) |
 | `k8s.node.allocatable_memory`              (type: `gauge`)                         | `k8s.node.allocatable.memory`            (type: `updowncounter`) |
 | `k8s.node.allocatable_ephemeral_storage`   (type: `gauge`)                         | `k8s.node.allocatable.ephemeral_storage` (type: `updowncounter`) |
-| `k8s.node.allocatable_pods`                (type: `gauge`)                         | `k8s.node.allocatable.pods`              (type: `updowncounter`) |
+| `k8s.node.allocatable_pods`                (type: `gauge`)                         | `k8s.node.allocatablepod.count`              (type: `updowncounter`) |
```
dashpole:
```suggestion
| `k8s.node.allocatable_pods`                (type: `gauge`)                         | `k8s.node.allocatable.pod.count`              (type: `updowncounter`) |
```

model/k8s/deprecated/metrics-deprecated.yaml
```diff
@@ -8,10 +8,192 @@ groups:
     stability: development
     deprecated:
       reason: renamed
-      renamed_to: k8s.replicationcontroller.desired_pods
```
AlexanderWert, line 11: Should we maybe add `k8s.replicationcontroller.desired_pods` as an additional deprecated metric?
(as this is the second time this metric is being renamed)

Otherwise `k8s.replicationcontroller.desired_pods` wouldn't be mentioned anywhere as a deprecated metric with only the original (from the first rename) name being documented here (i.e. `k8s.replication_controller.desired_pods`).

model/k8s/deprecated/metrics-deprecated.yaml
```diff
@@ -21,7 +203,239 @@ groups:
     stability: development
     deprecated:
       reason: renamed
-      renamed_to: k8s.replicationcontroller.available_pods
```
AlexanderWert, line 24: same as the comment above about adding additional deprecated metric

model/k8s/deprecated/metrics-deprecated.yaml
ChrsMark: I had both metrics originally, but the validation failed cause it was pointing to a non-existing metric (the first rename). Not sure if this is expected or an unhandled corner-case of our tooling 🤔

jinja2 (approved):

dmitryax:
Even after https://github.com/open-telemetry/semantic-conventions/issues/2306 is resolved, it still unclear why is this change required. Looking at [these guidelines](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/general/naming.md?plain=1#naming-rules-for-counters-and-updowncounters) it seems like existing naming still valid.
1. They cannot be taken as namespaces
2. "value being recorded represents discrete instances of a [countable quantity](https://wikipedia.org/wiki/Count_noun)"


```
### Naming rules for counters and UpDownCounters

#### Pluralization

Metric namespaces SHOULD NOT be pluralized.

Metric names SHOULD NOT be pluralized, unless the value being recorded
represents discrete instances of a
[countable quantity](https://wikipedia.org/wiki/Count_noun).
Generally, the name SHOULD be pluralized only if the unit of the metric in
question is a non-unit (like `{fault}` or `{operation}`).

Examples:

- `system.filesystem.utilization`, `http.server.request.duration`, and `system.cpu.time`
  should not be pluralized, even if many data points are recorded.
- `system.paging.faults`, `system.disk.operations`, and `system.network.packets`
  should be pluralized, even if only a single data point is recorded.

#### Do not pluralize UpDownCounter names

UpDownCounter names SHOULD NOT be pluralized.

For example if we have a namespace `system.process` which contains all metrics related
to the processes then to represent the count of the processes we can have a metric named
`system.process.count` instead of `system.processes`. Similarly, `cicd.pipeline.run.active` is preferred
over the `cicd.pipeline.active_runs`.
```

What am I missing?

dmitryax:
It introduces a significant number of breaking changes that don’t seem sufficiently justified

dmitryax:
The PR introduces a significant number of breaking changes, and I’m not sure the current justification is sufficient.

ChrsMark:
@dmitryax the guidance moves us towards consistency based on the discussions happened in the respective guidance PR and specifically https://github.com/open-telemetry/semantic-conventions/pull/2317#discussion_r2129164157. The part that applies here is that `UpDownCounter`s should not be pluralised, with the following example:

```
Similarly, `cicd.pipeline.run.active` is preferred
over the `cicd.pipeline.active_runs`.
```

Also note that those metrics are not in use yet in the Collector in the current format hence it's not really an impactful breaking change. Those metrics are already different to the Collector's ones as we can see in https://github.com/open-telemetry/semantic-conventions/pull/2822/files#diff-ba5fa96c108eca125c2c88f68cdfb0a3d0c95792912f5518eedd55666bc1c033.

I don't have a strong preference for this change either but it seems that we need it so as to be consistent with the rest of the project.

@lmolkova please chime in if we miss anything here.

dmitryax:
Ok, but the guidelines in the spec are still not strong enough for this change. All of the existing metrics seem to follow this section:

> Metric names SHOULD NOT be pluralized, **unless the value being recorded
> represents discrete instances of a
> [countable quantity](https://wikipedia.org/wiki/Count_noun).
> Generally, the name SHOULD be pluralized only if the unit of the metric in
> question is a non-unit (like `{fault}` or `{operation}`)**.

Do we need to remove that part from the spec?

In general, looking at the changed metrics names as a user, the existing ones seem to be more intuitive than to the new ones.

.chloggen/fix_k8s_updowncounters.yaml
```diff
@@ -0,0 +1,48 @@
...
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Renamed metrics:
+  - k8s.node.allocatable.pods -> k8s.node.allocatable.pod.count
+  - k8s.daemonset.current_scheduled_nodes -> k8s.daemonset.current_scheduled_node.count
+  - k8s.daemonset.desired_scheduled_nodes -> k8s.daemonset.desired_scheduled_node.count
+  - k8s.daemonset.misscheduled_nodes -> k8s.daemonset.misscheduled_node.count
+  - k8s.daemonset.ready_nodes -> k8s.daemonset.ready_node.count
```
dmitryax: Is it expected that we always just add `.count` suffix? Looking at the discussions behind the new guidelines, should we consider something like this?
```suggestion
  - k8s.daemonset.current_scheduled_nodes -> k8s.daemonset.node.current_scheduled
  - k8s.daemonset.desired_scheduled_nodes -> k8s.daemonset.node.desired_scheduled
  - k8s.daemonset.misscheduled_nodes -> k8s.daemonset.node.misscheduled
  - k8s.daemonset.ready_nodes -> k8s.daemonset.node.ready
```

Similar for other metrics. Just asking for sake of discussion.

ChrsMark: Yeap, adding `.count` is one option but what you suggest is also an option. I went for `.count` which looked less controversial to me (i.e. `k8s.daemonset.node.ready` might not be so clear?) but I would be happy to change it to what you suggest if it's preferable.

@lmolkova could you let us know what your suggestion would be here?
@open-telemetry/semconv-k8s-approvers please let us know what you think.

model/k8s/deprecated/metrics-deprecated.yaml
jsuereth: I believe the tooling expects all the previous metric names to remain around.

Given these are not "stable" I think there's less enforcement around keeping things around, although it is risky depending on whether the intermediate name was used.

If you have a caveat asking folks NOT to migrate to new metrics past version "X" then you can drop the intermediary if it's between "X" and now, as (theoretically) anyone taking a dependency there new this might be changing.

jsuereth (approved):

.chloggen/fix_k8s_updowncounters.yaml
jsuereth: I believe your PR is ready to be merged.  @lmolkova is OOO right now.  

Personally, I like the consistency of `.count` you have here, but I think both options are viable

jsuereth: I marked the PR as ready to be merged.  If the k8s SIG feels confident in this direction, please close this comment and this will be merged, otherwise please leave this open.

model/k8s/deprecated/metrics-deprecated.yaml
lmolkova: The automation should not allow to remove metric definitions regardless of stability - this helps minimize dependency conflicts in generated semconv artifacts.

The check was updated recently - https://github.com/open-telemetry/semantic-conventions/pull/2836, I imagine if this PR is rebased and checks re-run, it'd fail requiring metrics to be preserved (as deprecated).

.chloggen/fix_k8s_updowncounters.yaml
lmolkova: the @dmitryax 's  proposal looks great to me: 

e.g. in `k8s.daemonset.ready_nodes -> k8s.daemonset.node.ready`, `node` becomes a namespace we can group things around. We now default to `.` and namespacing whenever it makes sense.

It also looks consistent with other metrics we have, e.g. https://github.com/open-telemetry/semantic-conventions/blob/main/docs/otel/sdk-metrics.md

ChrsMark:
I will update this based on the outcome of https://github.com/open-telemetry/semantic-conventions/pull/2822#discussion_r2445508514.

.chloggen/fix_k8s_updowncounters.yaml
ChrsMark: Ok, I will update the PR accordingly since that's preferable and more consistent with the rest of the project.

updated: [41b1cc5](https://github.com/open-telemetry/semantic-conventions/pull/2822/commits/41b1cc554fea68f4cbdb8698f4f95dc4c5204cce)

model/k8s/deprecated/metrics-deprecated.yaml
ChrsMark: Added that but the validation is failing:

```
Diagnostic report:

Violation: Metric 'k8s.replication_controller.available_pods' was renamed to 'k8s.replicationcontroller.available_pods', but the new metric does not exist or is deprecated.
  - Category         : deprecation_violation
  - SemConv group    : metric.k8s.replication_controller.available_pods
  - SemConv attribute: 

Violation: Metric 'k8s.replication_controller.desired_pods' was renamed to 'k8s.replicationcontroller.desired_pods', but the new metric does not exist or is deprecated.
  - Category         : deprecation_violation
  - SemConv group    : metric.k8s.replication_controller.desired_pods
  - SemConv attribute: 
```

@lmolkova could you clarify what's the expectation here? Should we preserve the renames' "history" or just point to the latest existing metric from the deprecated ones?

ChrsMark:
@open-telemetry/semconv-k8s-approvers please take another look since this PR has been updated.

@open-telemetry/specs-semconv-maintainers could you help with the deprecation question from https://github.com/open-telemetry/semantic-conventions/pull/2822#discussion_r2390441581? Right now the checks fail because of the "double" deprecation chain, or do I miss anything else there?

docs/system/k8s-metrics.md
```diff
@@ -47,7 +47,7 @@ and therefore inherit its attributes, like `k8s.pod.name` and `k8s.pod.uid`.
   - [Metric: `k8s.node.allocatable.cpu`](#metric-k8snodeallocatablecpu)
   - [Metric: `k8s.node.allocatable.memory`](#metric-k8snodeallocatablememory)
   - [Metric: `k8s.node.allocatable.ephemeral_storage`](#metric-k8snodeallocatableephemeral_storage)
-  - [Metric: `k8s.node.allocatable.pods`](#metric-k8snodeallocatablepods)
+  - [Metric: `k8s.node.pod.allocatable](#metric-k8snodepodallocatable)
```
jinja2:
```suggestion
  - [Metric: `k8s.node.pod.allocatable`](#metric-k8snodepodallocatable)
```

docs/system/k8s-metrics.md
```diff
@@ -62,40 +62,40 @@ and therefore inherit its attributes, like `k8s.pod.name` and `k8s.pod.uid`.
   - [Metric: `k8s.node.filesystem.capacity`](#metric-k8snodefilesystemcapacity)
   - [Metric: `k8s.node.filesystem.usage`](#metric-k8snodefilesystemusage)
 - [Deployment metrics](#deployment-metrics)
-  - [Metric: `k8s.deployment.desired_pods`](#metric-k8sdeploymentdesired_pods)
-  - [Metric: `k8s.deployment.available_pods`](#metric-k8sdeploymentavailable_pods)
+  - [Metric: `k8s.deployment.pod.desired`](#metric-k8sdeploymentpoddesired)
+  - [Metric: `k8s.deployment.pod.available](#metric-k8sdeploymentpodavailable)
```
jinja2:
```suggestion
  - [Metric: `k8s.deployment.pod.available`](#metric-k8sdeploymentpodavailable)
```

docs/system/k8s-metrics.md
```diff
@@ -62,40 +62,40 @@ and therefore inherit its attributes, like `k8s.pod.name` and `k8s.pod.uid`.
...
 - [ReplicaSet metrics](#replicaset-metrics)
-  - [Metric: `k8s.replicaset.desired_pods`](#metric-k8sreplicasetdesired_pods)
-  - [Metric: `k8s.replicaset.available_pods`](#metric-k8sreplicasetavailable_pods)
+  - [Metric: `k8s.replicaset.pod.desired](#metric-k8sreplicasetpoddesired)
+  - [Metric: `k8s.replicaset.pod.available](#metric-k8sreplicasetpodavailable)
```
jinja2:
```suggestion
  - [Metric: `k8s.replicaset.pod.desired`](#metric-k8sreplicasetpoddesired)
  - [Metric: `k8s.replicaset.pod.available`](#metric-k8sreplicasetpodavailable)
```

docs/system/k8s-metrics.md
```diff
@@ -62,40 +62,40 @@ and therefore inherit its attributes, like `k8s.pod.name` and `k8s.pod.uid`.
...
 - [ReplicationController metrics](#replicationcontroller-metrics)
-  - [Metric: `k8s.replicationcontroller.desired_pods`](#metric-k8sreplicationcontrollerdesired_pods)
-  - [Metric: `k8s.replicationcontroller.available_pods`](#metric-k8sreplicationcontrolleravailable_pods)
+  - [Metric: `k8s.replicationcontroller.pod.desired](#metric-k8sreplicationcontrollerpoddesired)
+  - [Metric: `k8s.replicationcontroller.pod.available](#metric-k8sreplicationcontrollerpodavailable)
```
jinja2:
```suggestion
  - [Metric: `k8s.replicationcontroller.pod.desired`](#metric-k8sreplicationcontrollerpoddesired)
  - [Metric: `k8s.replicationcontroller.pod.available`](#metric-k8sreplicationcontrollerpodavailable)
```
Looks like backticks missing in a few places, commenting with fix for the ones i noticed

docs/system/k8s-metrics.md
```diff
@@ -62,40 +62,40 @@ and therefore inherit its attributes, like `k8s.pod.name` and `k8s.pod.uid`.
...
 - [StatefulSet metrics](#statefulset-metrics)
-  - [Metric: `k8s.statefulset.desired_pods`](#metric-k8sstatefulsetdesired_pods)
-  - [Metric: `k8s.statefulset.ready_pods`](#metric-k8sstatefulsetready_pods)
-  - [Metric: `k8s.statefulset.current_pods`](#metric-k8sstatefulsetcurrent_pods)
-  - [Metric: `k8s.statefulset.updated_pods`](#metric-k8sstatefulsetupdated_pods)
+  - [Metric: `k8s.statefulset.pod.desired](#metric-k8sstatefulsetpoddesired)
+  - [Metric: `k8s.statefulset.pod.ready](#metric-k8sstatefulsetpodready)
+  - [Metric: `k8s.statefulset.pod.current](#metric-k8sstatefulsetpodcurrent)
+  - [Metric: `k8s.statefulset.pod.updated](#metric-k8sstatefulsetpodupdated)
```
jinja2:
```suggestion
  - [Metric: `k8s.statefulset.pod.desired`](#metric-k8sstatefulsetpoddesired)
  - [Metric: `k8s.statefulset.pod.ready`](#metric-k8sstatefulsetpodready)
  - [Metric: `k8s.statefulset.pod.current`](#metric-k8sstatefulsetpodcurrent)
  - [Metric: `k8s.statefulset.pod.updated`](#metric-k8sstatefulsetpodupdated)
```

docs/system/k8s-metrics.md
```diff
@@ -62,40 +62,40 @@ and therefore inherit its attributes, like `k8s.pod.name` and `k8s.pod.uid`.
...
 - [HorizontalPodAutoscaler metrics](#horizontalpodautoscaler-metrics)
-  - [Metric: `k8s.hpa.desired_pods`](#metric-k8shpadesired_pods)
-  - [Metric: `k8s.hpa.current_pods`](#metric-k8shpacurrent_pods)
-  - [Metric: `k8s.hpa.max_pods`](#metric-k8shpamax_pods)
-  - [Metric: `k8s.hpa.min_pods`](#metric-k8shpamin_pods)
+  - [Metric: `k8s.hpa.pod.desired](#metric-k8shpapoddesired)
+  - [Metric: `k8s.hpa.pod.current](#metric-k8shpapodcurrent)
+  - [Metric: `k8s.hpa.pod.max](#metric-k8shpapodmax)
+  - [Metric: `k8s.hpa.pod.min](#metric-k8shpapodmin)
```
jinja2:
```suggestion
  - [Metric: `k8s.hpa.pod.desired`](#metric-k8shpapoddesired)
  - [Metric: `k8s.hpa.pod.current`](#metric-k8shpapodcurrent)
  - [Metric: `k8s.hpa.pod.max`](#metric-k8shpapodmax)
  - [Metric: `k8s.hpa.pod.min`](#metric-k8shpapodmin)
```

docs/system/k8s-metrics.md
```diff
@@ -62,40 +62,40 @@ and therefore inherit its attributes, like `k8s.pod.name` and `k8s.pod.uid`.
...
-  - [Metric: `k8s.job.active_pods`](#metric-k8sjobactive_pods)
-  - [Metric: `k8s.job.failed_pods`](#metric-k8sjobfailed_pods)
-  - [Metric: `k8s.job.successful_pods`](#metric-k8sjobsuccessful_pods)
-  - [Metric: `k8s.job.desired_successful_pods`](#metric-k8sjobdesired_successful_pods)
-  - [Metric: `k8s.job.max_parallel_pods`](#metric-k8sjobmax_parallel_pods)
+  - [Metric: `k8s.job.pod.active](#metric-k8sjobpodactive)
+  - [Metric: `k8s.job.pod.failed](#metric-k8sjobpodfailed)
+  - [Metric: `k8s.job.pod.successful](#metric-k8sjobpodsuccessful)
+  - [Metric: `k8s.job.pod.desired_successful](#metric-k8sjobpoddesired_successful)
+  - [Metric: `k8s.job.pod.max_parallel](#metric-k8sjobpodmax_parallel)
```
jinja2:
```suggestion
  - [Metric: `k8s.job.pod.active`](#metric-k8sjobpodactive)
  - [Metric: `k8s.job.pod.failed`](#metric-k8sjobpodfailed)
  - [Metric: `k8s.job.pod.successful`](#metric-k8sjobpodsuccessful)
  - [Metric: `k8s.job.pod.desired_successful`](#metric-k8sjobpoddesired_successful)
  - [Metric: `k8s.job.pod.max_parallel`](#metric-k8sjobpodmax_parallel)
```

docs/non-normative/k8s-migration.md
```diff
@@ -111,7 +111,7 @@ The changes between collector implementation and semantic conventions:
 | `k8s.node.allocatable_cpu`                 (type: `gauge`)                         | `k8s.node.allocatable.cpu`               (type: `updowncounter`) |
 | `k8s.node.allocatable_memory`              (type: `gauge`)                         | `k8s.node.allocatable.memory`            (type: `updowncounter`) |
 | `k8s.node.allocatable_ephemeral_storage`   (type: `gauge`)                         | `k8s.node.allocatable.ephemeral_storage` (type: `updowncounter`) |
-| `k8s.node.allocatable_pods`                (type: `gauge`)                         | `k8s.node.allocatable.pods`              (type: `updowncounter`) |
+| `k8s.node.allocatable_pods`                (type: `gauge`)                         | `k8s.node.pod.allocatable`              (type: `updowncounter`) |
```
jinja2, line 114: I am confused why allocatable property is in different positions here, for e.g., `k8s.node.allocatable.cpu` vs. `k8s.node.pod.allocatable`.  Is there a reason for this difference in namespacing for pod resource only?

dmitryax: Good point. `k8s.node.allocatable.pod.count` then, if we don't allow plurals? even though I still find `k8s.node.allocatable.pods` more intuitive.

ChrsMark: I had in mind to also change those in a follow-up PR (since they are not pluralized), i.e. `k8s.node.memory.allocatable`. We already have similar ones like [`k8s.node.memory.available`](https://github.com/open-telemetry/semantic-conventions/blob/1e7e3961abb3c9ee9dfe110d9424a96b5d1ceee9/docs/system/k8s-metrics.md?plain=1#L1140). Would that make sense?

docs/system/k8s-metrics.md
ChrsMark: Ouch, good catch. For some reason the renaming messed things up. I have fixed those.

docs/non-normative/k8s-migration.md
jinja2: sounds good to me

model/k8s/deprecated/metrics-deprecated.yaml
lmolkova: If I understand correctly:

- metric A was renamed to B
- then metric B was renamed to C

Should deprecation note on A now point to C? 
Yes!

If you migrate from old version with metric A, you should migrate directly to C, no need to even know that there was B in the middle. 
If someone is looking for migration path to older version of semconv (before C existed), they could find B in that version.

lmolkova (approved):

model/k8s/deprecated/metrics-deprecated.yaml
ChrsMark: Thank's for clarifying this, should be fixed now: [4675f3f](https://github.com/open-telemetry/semantic-conventions/pull/2822/commits/4675f3f63299dac8ce69ed6b953e2a7385fc58e5)

model/k8s/deprecated/metrics-deprecated.yaml
ChrsMark: Same as https://github.com/open-telemetry/semantic-conventions/pull/2822#discussion_r2463754932

docs/non-normative/k8s-migration.md
ChrsMark: Follow-up PR: https://github.com/open-telemetry/semantic-conventions/pull/2973

PR #2826, opened 2025-09-29
[chore] System Semconv Description Guidance

braydonk:
## Changes

This PR explains how the `brief` and `note` fields will be used in the namespaces covered by the System Semantic Conventions group.

The general concept is that we want to assume a baseline level of knowledge for readers of our conventions, as this frees us up to focus the descriptions of our instrumentation on instrumentation details rather than on explaining general concepts.

### DIFF
--- a/docs/non-normative/groups/system/design-philosophy.md
+++ b/docs/non-normative/groups/system/design-philosophy.md
@@ -168,21 +168,65 @@ applied in a cross-platform manner. See
 
 ### Value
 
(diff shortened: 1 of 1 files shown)


trask (approved):
I suspect these briefs and notes will ultimately show up in lots of end user dashboards.

I think links to authoritative sources with more detailed information is a good compromise.

ChrsMark (approved):

PR #2831, opened 2025-09-30
feat: Add time to first draw span for app screen

### LINKED ISSUE #2830: Add app spans
### Area(s)

area:app

### What's missing?

Spans for:
1. screen load
2. app launch
3. time to first appear

_No response_

### Tip

limhjgrace:
Fixes #2830

## Changes

This PR proposes the a span definition for an application screen's time to first draw following spans (spanName = `span.app.screen.time_to_first_draw.internal`)

## Revisions
- **Rev2**: [Rev1-Rev2 diff](https://github.com/open-telemetry/semantic-conventions/pull/2831/files/bf828ad10871f4a1e9c786462de81c4156dcc35e..395004f7198b87e3861c667ee464c641e311645f)
      1. App start span with only `cold` and `warm` values defined
      2. Time to first appear client span --> time to first draw internal span
      4. Screen load client span --> screen load internal span. Definition also changed: Time to main thread idle --> time to first stable frame
      5. Screen visible client span --> screen visible event 
- **Rev3**: [Rev2-Rev3 diff](https://github.com/open-telemetry/semantic-conventions/pull/2831/files/395004f7198b87e3861c667ee464c641e311645f..c8f9774bce7dde07fe0c891d689d4cd2fca316ea)
      1. App start client span --> app start internal span. Clarified end time to be the moment before UI rendering begins. 
      2. Removed screen load span. Will raise separate PR to get other proposals unblocked.
      3. Unit fix (ns -> s)
      4. Type fix (int -> double)
      6. Rename `app.screen.first_draw` --> `app.screen.time_to_first_draw`
      7. `app.visible` event -> `app.time_on_screen` internal span
- **Rev4**: [Rev3-Rev4 diff](https://github.com/open-telemetry/semantic-conventions/pull/2831/files/c8f9774bce7dde07fe0c891d689d4cd2fca316ea..af0e648a70eb0ee580dd4d0ebfa5e1b6306d9899)
      1. Removed `*.duration` attributes
      2. Removed app start span so that it is reviewed in a separate PR: https://github.com/open-telemetry/semantic-conventions/pull/2965
      3. Removed `span.app.time_on_screen.internal`

### DIFF
--- a/.chloggen/feat_app-spans.yaml
+++ b/.chloggen/feat_app-spans.yaml
@@ -10,7 +10,7 @@ change_type: enhancement
 component: app
 
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add span for application screen's time to first draw 
+note: Add span for application screen's time to first draw
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 # The values here must be integers.


model/app/registry.yaml
```diff
@@ -86,6 +86,137 @@ groups:
           is larger in scope than individual widgets. Multiple screens can
           coexist on the same display simultaneously (e.g., split view on tablets).
         examples: ["MainActivity", "ProductDetailFragment", "ProfileView", "ProfileViewController"]
+      - id: app.screen.depth
```
thompson-tomo: Not sure how a depth could be anything other than 0 based on description of app.screen.name

> screen represents only the part of the device display drawn by the app. .... Multiple screens can coexist on the same display simultaneously (e.g., split view on tablets).

App.widget is defined for ui components ie modals

model/app/registry.yaml
```diff
@@ -86,6 +86,137 @@ groups:
...
+          such as taps, swipes, or gestures.
+        examples: [1000]
+      - id: app.screen.main_thread_busy_time
+        type: int
+        stability: development
+        brief: >
+          The time, in nanoseconds, during which the main thread is occupied or
+          blocked from processing user input, such as during rendering, animations,
+          or network requests. This indicates the time the app is "busy" and unable to respond to user interactions.
+      - id: app.screen.nodes
```
thompson-tomo: Not sure of the usefulness or how easy this would be to calculate.

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
...
+        requirement_level: opt_in
+      - ref: app.screen.load.duration
+        requirement_level: required
+      - ref: app.screen.nodes
+        requirement_level: opt_in
+      - ref: app.screen.depth
+        requirement_level: opt_in
+      - ref: app.screen.main_thread_busy_time
+        requirement_level: opt_in
+  - id: span.app.screen.visible.client
```
thompson-tomo: Shouldn't this be an event or even a metric to avoid long running span.

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
+groups:
+  - id: span.app.screen.first_appear.client
```
thompson-tomo: Would recommend looking at what has been defined for web_vital and doing a similar approach.

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
...
+    type: span
+    span_kind: client
```
thompson-tomo: Should be internal

model/app/spans.yaml
limhjgrace: https://github.com/open-telemetry/semantic-conventions/blob/main/model/browser/events.yaml

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - ref: app.screen.name
+        requirement_level: required
+      - ref: app.screen.first_appear.duration
+        requirement_level: required
+      - ref: app.screen.type
+        requirement_level: opt_in
+      - ref: app.screen.nodes
+        requirement_level: opt_in
+      - ref: app.screen.depth
+        requirement_level: opt_in
```
LikeTheSalad: Do you have an existing use case for these 2 attributes? I'm asking because I can't find one atm, and I think it's often easy to add a lot of attrs to the spec upfront, for the sake of having them there (even as `opt_in`), that end up not being used, or worse, not useful enough for when a use case that needs them arises, which can cause having to deprecate them, which is often a tedious process.

Also, the current description for `depth` sounds a bit confusing to me:

> Depth represents the screen's position in the screen hierarchy tree.
> The root screen of the app is considered to have a depth of 0. Screens
> that are nested within other screens (such as in tab navigation, modals,
> or multi-screen flows) will have increasing depth values (1, 2, etc.).

The mention of the `root` screen at first seems to suggest that the depth is about _navigation depth_, therefore, the root will be 0. However, the following part seems to suggest that depth is actually about embedded screens, and also a "z-index" sort of mechanism. In either case, I think the spec isn't clear enough to become feasible to use. 

For example, considering that we're talking about a _navigation depth_ in the sense that there will be several screens on an app that take the whole device's screen, and we wanted to number them in terms of "how many screens I have to visit before I get to this one", with the root screen starting at `0`, from a user perspective the order of the screens isn't always linear, so if I visit the screens in this order: A -> B -> C, in several cases I can also do A -> C -> B, or even start the app by going straight into a "non-root" screen, which is the case for deeplinks for example, where something like this could happen: C -> A. Not to mention the cases where an otherwise "full screen" on a smaller device can become an "embedded screen" on a bigger one.

In the case of embedded screens and/or z-index measurements, it would be less tricky but still confusing at times, for example, a screen with a "bottom navigation" bar displays a different screen depending on the item on the bar that's selected at a given time. In this case, what "depth value" should we assign to these screens? Are they all `1` because they have to take turns to be the "current visible embedded screen" on the root one? Or they should be numbered based on the order in which they appear on the bar (which also raises a question of which side we start to count from, i.e. left-right or right-left)? Also, if a "popup/dialog" modal is shown when clicking on a button that's on the root screen's layout, where there is also an embedded screen nearby, what depth value should we assign said popup? `1`, because it came off of the root screen, so all the visible items on a screen at a time must share the same value? or `2` because it sits on top of embedded screens, which already have the number `1` assigned?

Regarding the `nodes` attribute, I'm a bit confused about why it would be needed, considering that it seem to provide the number of UI elements that are present on a screen, which, ignoring that it would be costly to gather this information at runtime, the outcome would be something that developers can tell by looking at their source code, so I'm confused about it.

Generally speaking, I think it's best to start simple and then let the spec organically evolve as needed.

model/app/spans.yaml
LikeTheSalad: It looks to me that this could better suit an event, rather than a span, considering that it's mostly about gathering the `duration` attribute and that it doesn't seem to matter much when the process started and ended.

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
+groups:
+  - id: span.app.screen.first_appear.client
+    type: span
+    span_kind: client
+    brief: This span describes the time until the first visual render of an application UI screen.
+    note: |
+      **Span name**: MUST be one of the following based on the measurement:
```
LikeTheSalad: I would prefer to avoid this ambiguity; we should probably pick one to ensure consistency across implementations, and if we need the other one later, it should probably become its own event.

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - ref: app.screen.type
+        requirement_level: opt_in
+      - ref: app.screen.nodes
+        requirement_level: opt_in
+      - ref: app.screen.depth
+        requirement_level: opt_in
+  - id: span.app.screen.load.client
+    type: span
+    span_kind: client
+    brief: This span describes an application screen load operation, from navigation trigger to when the main thread becomes ready for user input.
```
LikeTheSalad: I'd avoid technical wording here and probably go with something like:

"This span describes an application screen load operation, from when the screen launches to when it becomes ready for user input."

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
...
+      **Span name:** MUST be `ScreenLoad`.
+
+      **Span kind** MUST be `CLIENT`.
+
+      This span captures the time from user initiation of navigation (e.g., tapping a button) to the point when the main thread is
+      ready to handle user input (e.g., taps, scrolls, gestures).
+      **Main thread readiness** refers to when the UI thread is no longer blocked by rendering, animations, or background tasks,
+      and is free to accept user input.
+      It is important to note that while the main thread may be ready, some dynamic content, animations, or UI components may
+      still be loading, making the screen "appear" interactive even if it's not fully ready.
```
LikeTheSalad: I would think that this event should be triggered only after everything has finished loading, but this sentence suggests otherwise. Which is confusing because then it looks like this event would completely overlap with `app.screen.first_appear.client`?

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - ref: app.screen.name
+        requirement_level: required
+      - ref: app.screen.type
+        requirement_level: opt_in
+      - ref: app.screen.load.duration
+        requirement_level: required
+      - ref: app.screen.nodes
+        requirement_level: opt_in
+      - ref: app.screen.depth
+        requirement_level: opt_in
```
LikeTheSalad: Same comment as the one for `app.screen.first_appear.client`.

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - ref: app.screen.main_thread_busy_time
```
LikeTheSalad: This attribute is very confusing, to the point that I don't think it's worth having it available because it can do more harm than good. My understanding is that the main thread is always busy rendering the UI and oftentimes doing some work in between, though I wouldn't consider it to "become fully idle" technically speaking, even if the UI is done rendering. So it doesn't seem clear to me how to either implement this or read it later.

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
...
+        requirement_level: required
+      - ref: app.screen.first_appear.duration
+        requirement_level: required
+      - ref: app.screen.type
+        requirement_level: opt_in
+      - ref: app.screen.nodes
+        requirement_level: opt_in
+      - ref: app.screen.depth
+        requirement_level: opt_in
+  - id: span.app.screen.load.client
```
LikeTheSalad: Similar to "app.screen.first_appear.client", it looks to me that this suits better an event rather than a span.

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
...
+    type: span
+    span_kind: client
+    brief: This span represents an application start operation.
+    note: |
+      **Span name:** MUST be `AppStart`.
+
+      **Span kind** MUST be `CLIENT`.
+
+      This span captures the time from user initiation (e.g., tapping the app icon or opening a link)
+      to the moment the app is ready for interaction.
```
LikeTheSalad: I think this description somewhat links this event to the "screen load" or "screen first time", in the sense that the duration is gathered up to the point where the "root screen" is either finished rendering, or finished fully loading. Though it's not clear which one of those cases it refers to.

model/app/spans.yaml
```diff
@@ -0,0 +1,97 @@
...
+      It does not include times when the app is in the background or when other screens/modal views are covering it.
+    stability: development
+    attributes:
+      - ref: app.screen.name
+        requirement_level: required
+      - ref: app.screen.type
+        requirement_level: opt_in
+      - ref: app.screen.visible.duration
+        requirement_level: required
+  - id: span.app.start.client
```
LikeTheSalad: I believe this suits an event better than a span, which I've mentioned on a couple of other items, but I think for this one is critical to not be a span, as the duration from which an app is "launched" to the point it becomes visible to the user covers the process where OTel is initialized as well, and it wouldn't be feasible to start a span before the OTel client is initialized, however, it's possible to keep track of the time since the very beginning and use it to send it in the proposed "duration" attribute within an event later on.

model/app/registry.yaml
```diff
@@ -86,6 +86,137 @@ groups:
...
+      - id: app.start.duration
+        type: int
+        stability: development
+        brief: >
+          The total time, in nanoseconds, taken for the application to start, from user initiation to being ready for interaction.
+        note: >
+          This is the time, in nanoseconds, between the user's initiation of the app start (e.g., tapping the app icon or opening a link) and the point when the app is fully ready for interaction,
+          such as the main screen becoming visible or the app’s main functionality being accessible.
+        examples: [1000]
+      - id: app.start.type
```
LikeTheSalad: Do all these types of events map to all platforms? Generally speaking, I think it's best to start with basic enum values, such as `cold` and `hot`, and then extend them as needed.

model/app/registry.yaml
limhjgrace: This metric could be helpful in troubleshooting a high screen load time since it gives details about the complexity and amount of work done of/for the screen. I know in Android, we can get this by extending the logic of the FirstDrawListener and including an extension function View.getComplexity() that will calculate UI complexity statistics (we plan on making a contribution to the upstream Android SDK soon) and my understanding is that this is feasible for iOS applications too.

model/app/registry.yaml
limhjgrace: Sorry, this was supposed to be `app.screen.view.depth` and `app.screen.view.nodes` but it got changed around as we went between `app.screen.*` vs. `app.view.*` for the screen name attribute. Will update in next rev.

model/app/registry.yaml
Doohl: > Not sure of the usefulness or how easy this would be to calculate.

So count of screen nodes and screen depth is helpful for debugging long draw spans. You can make the assumption that more complex views will have longer draw times. These attributes can help users make those connections in their telemetry.

I have a PR I intend to put out for Android that will use this.

model/app/registry.yaml
thompson-tomo: I am not sure how that helps as you still have 1 screen. Maybe you mean `app.widget.zindex`?

model/app/registry.yaml
thompson-tomo: Sure it could be calculated based on recursively counting the items but I don't see that as an easy (efficient) solution. If the use is debugging, then the same could be achieved with guidance. Also how does it differ to app.jank https://opentelemetry.io/docs/specs/semconv/app/app/#event-appjank

model/app/registry.yaml
limhjgrace: Removing this attribute from this PR as we plan on reviewing in a separate PR with Android implementation.

model/app/registry.yaml
limhjgrace: Also removing this attribute from this PR as we plan on reviewing in a separate PR with Android implementation.

model/app/spans.yaml
limhjgrace: Will convert to an event

model/app/spans.yaml
limhjgrace: +1, will remove from this PR as we plan on adding these in a separate PR with it implemented in Android for better understanding. I think Leo's taken a look at this PR but will ask him to refer to comments from here as he works on the proposal.

model/app/registry.yaml
limhjgrace: Will remove to only contain `cold` and `warm` since I think those are the ones that apply to all platforms. But I'm not familiar with how/where I should extend them. For example, for Android we would have to add `hot` and for iOS we would have to add `pre-warm`. Could you point me to an example of where we've extended existing attributes?

model/app/spans.yaml
limhjgrace: Yes, will remove in next rev.

model/app/spans.yaml
limhjgrace: No, I think there are use cases where process start and end is still important. For example, if I wanted to create a visualization of all the telemetry in chronological order I would need these start and end times. 

Maybe this can be achieved by using the duration and the observedTimestamp but with that aside, and maybe I've oversimplified the definition, I thought we should be treating telemetry with a duration as a span and logs to be things that happened in an instance in time? Unless it's a long-running span that has a high risk of not gracefully being ended. Or is the correct definition that we should consider a span to be telemetry with start/end times of significance?

model/app/spans.yaml
limhjgrace: Agreed. Will convert so that it's just for TimeToFirstDraw since this is platform agnostic.

model/app/spans.yaml
limhjgrace: Will refactor in next rev

model/app/spans.yaml
limhjgrace: I disagree. Given a screen load is expected to take at most a couple seconds, I would prefer to keep it a span.

model/app/spans.yaml
limhjgrace: Agreed. I plan on changing the proposed screen load definition anyways but will make sure it doesn't contain technical wording.

model/app/spans.yaml
limhjgrace: Resolving as this definition is going to be obsolete in next rev with latest proposed definition.

model/app/spans.yaml
limhjgrace: Removed, as they'll be proposed in a separate PR.

model/app/spans.yaml
limhjgrace: That's a fair point since the SDK will technically start after the app has launched. However, would we not able to set the start time of a span to be the timestamp of when the app launched? Or is this automatically derived and would always be the start time of the SDK? To me, app start is more intuitive as a span since it has a duration.

model/app/spans.yaml
limhjgrace: No, the intention was not to link it to either as technically an application can be considered launched/started before the first screen has been loaded. My understanding is that for the Android SDK, the end of an app start is defined to be when the "Activity Resumed" life cycle event occurs for the first activity. Similarly, Apple provides [`didBecomeActiveNotification`](https://developer.apple.com/documentation/uikit/uiapplication/didbecomeactivenotification) which could be used but doesn't necessarily mean a screen has loaded.

model/app/spans.yaml
limhjgrace: Decided to just propose as time to first draw. Will create a separate time to first appear span proposal for iOS

model/app/registry.yaml
```diff
@@ -86,6 +86,95 @@ groups:
...
+        type: int
+        stability: development
+        brief: >
+          The time taken, in nanoseconds, for a screen to load and become stable, measured from when
+          the screen begins appearing until the first stable frame has been rendered.
+        note: >
+          This attribute represents the total duration, in nanoseconds, of a screen load as experienced by the user.
+          The screen load is considered complete when the first stable frame of the screen has been rendered.
+          A frame is considered stable when no major layout or frame changes occur for a brief stability interval (e.g. 150 ms).
+        examples: [1000]
```
thompson-tomo: Sem conv does time measurements in seconds and uses doubles.

model/app/registry.yaml
```diff
@@ -86,6 +86,95 @@ groups:
...
+        examples: [1200, 5500, 700]
+      - id: app.start.duration
+        type: int
+        stability: development
+        brief: >
+          The total time, in nanoseconds, taken for the application to start, from user initiation to being ready for interaction.
+        note: >
+          This is the time, in nanoseconds, between the user's initiation of the app start (e.g., tapping the app icon or opening a link) and the point when the app is fully ready for interaction,
+          such as the main screen becoming visible or the app’s main functionality being accessible.
+        examples: [1000]
```
thompson-tomo: Sem conv does time measurements in seconds and uses doubles.

model/app/registry.yaml
```diff
@@ -86,6 +86,95 @@ groups:
...
+              brief: App start from background state
+              stability: development
+        stability: development
+        brief: >
+          The type of application start, indicating the state of the app when it was initiated.
+        note: |
+          Start types include:
+
+          - **Cold**: The app is started from a terminated state, meaning no prior instance of the app is running.
+          - **Warm**: The app is started from the background, meaning an instance of the app is still in memory, but not active.
```
thompson-tomo: This should be brief &/or note on the members

model/app/events.yaml
```diff
@@ -62,3 +62,22 @@ groups:
...
+      This event captures the duration that a screen was visible to the user, indicating the time between when the screen becomes
+      visible and when it stops being visible.
+    note: >
+      This event measures the screen's visibility, in nanoseconds, meaning it includes only the time the screen is actively being shown to the user.
+      It does not include times when the app is in the background or when other screens/modal views are covering it. For example:
+      - **iOS**: From `viewDidAppear` to `viewWillDisappear`, indicating when the screen is first shown and when it's about to disappear.
+      - **Android**: From `onResume` to `onPause`, indicating when the screen comes to the foreground and when it goes to the background.
+    attributes:
+      - ref: app.screen.name
+        requirement_level: required
```
thompson-tomo: Id should be required & name opt-in as per other events.

model/app/spans.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - ref: app.screen.load.duration
+        requirement_level: required
+  - id: span.app.start.client
+    type: span
+    span_kind: client
+    brief: This span represents an application start operation.
+    note: |
+      **Span name:** MUST be `AppStart`.
+
+      **Span kind** MUST be `CLIENT`.
```
thompson-tomo: How is this a client? I see it as internal

model/app/spans.yaml
LikeTheSalad: I think the fact that we gather a `duration` attribute makes it more suitable for an event rather than a span. To me, a span is useful to track the start/end of a process, from which the duration can be computed. Also, to be able to keep track of children spans in a way that they can be displayed together in a waterfall UI. Unless I missed something, the use case we're discussing here doesn't need those features, though please let me know if that's not the case.

model/app/spans.yaml
LikeTheSalad: > technically an application can be considered launched/started before the first screen has been loaded

The description reads:

> "to the moment the app is ready for interaction"

How can it be possible that the app is ready for interaction without the first screen being loaded?

model/app/spans.yaml
LikeTheSalad: > would we not able to set the start time of a span to be the timestamp of when the app launched?

I haven't tried it. It's probably possible, though it feels hacky to me to go this way instead of using an event. I've added more details on it in [this other comment](https://github.com/open-telemetry/semantic-conventions/pull/2831#discussion_r2428600162).

model/app/registry.yaml
LikeTheSalad: > Could you point me to an example of where we've extended existing attributes?

By extending it I meant to add them later if needed. However, my understanding of this proposal is that it's meant to be platform-agnostic, so what you mentioned about adding `hot` for Android and `pre-warm` for iOS doesn't seem to match that goal. If they need to be separate, then I think it's better to define them on a platform-specific doc instead.

model/app/spans.yaml
LikeTheSalad: > Given a screen load is expected to take at most a couple seconds.

Isn't it something that can be tracked in the `duration` attribute? - Also, regarding the spans vs events discussion, I think it's better to have it in a single place. I've added more details in [this thread](https://github.com/open-telemetry/semantic-conventions/pull/2831#discussion_r2428600162).

model/app/spans.yaml
thompson-tomo: Agree with @LikeTheSalad description that spans are useful for tracking a process and seeing all the child processes triggered as a result. Wheras events are useful for knowing x happened at time Y.

I do however see the use of having one span to describe the navigation process which can act as the root but none currently do that for me.

model/app/spans.yaml
limhjgrace: That's fair, I'll update to remove this specification and make it more broad/applicable.

model/app/spans.yaml
limhjgrace: > To me, a span is useful to track the start/end of a process, from which the duration can be computed.

+1. But for the `duration` calculation, I wanted to do this client side instead of server-side or query-time as this adds scalability, performance, and technical concerns. Is there a concern in adding a `duration` attribute to spans even though it can be derived from the start/endtime? 

> Also, to be able to keep track of children spans in a way that they can be displayed together in a waterfall UI. Unless I missed something, the use case we're discussing here doesn't need those features, though please let me know if that's not the case.

Yes, this is a valid use case for services who want to provide a way to visualize an entire session. There was an on going discussion on how we could provide this view by having the specs support a session span but given the concerns of a long running span, the community didn't align on this. 

> spans are useful for tracking a process and seeing all the child processes triggered as a result

+1. To me, a screen load and time to first draw are both processes and is analogous to a web page load, not vitals to a screen. I know below that they've have been compared to the web vitals but I don't think that is an apples to apples comparison, especially when not all the web vitals have a timing related component to it e.g. CLS. It was also mentioned that we could have a span for the screen but I'm thinking this will have similar long running concerns as a session span and won't be viable. That and I'm assuming the suggestion was to add these as span events? But I thought those are being deprecated?

Maybe this is a question of what is a span and what is a log event given there is some overlap between the two where a log event should be used in place of creating long running spans. Technically, most spans could also just be defined as a log event and the discussion feels like it's heading towards that direction. If so, are we defining spans to be used only for tracking a process for its child processes? Or is the suggestion to combine both of these spans into one and have the timing defined as attributes (e.g. `attributes.time_to_first_draw` and `attributes.time_to_first_stable_draw`)?

model/app/registry.yaml
limhjgrace: Ack, updating in next rev.

model/app/registry.yaml
limhjgrace: Ack, updating in next rev.

model/app/spans.yaml
limhjgrace: Completely missed this one, sorry about that. Updating in next rev.

model/app/events.yaml
limhjgrace: Ack, updating in next rev

model/app/registry.yaml
limhjgrace: Ack, done in next rev

model/app/spans.yaml
limhjgrace: Actually, to completely decouple the app start with the screen load, I've updated the definition so that app start completes the moment before UI rendering begins. This works for both Android (AppStart span ends after the first Activity finishes creating which is before UI rendering)  and iOS according to their [app launch sequence doc](https://developer.apple.com/documentation/uikit/about-the-app-launch-sequence).

model/app/spans.yaml
thompson-tomo: > To me, a screen load and time to first draw are both processes and is analogous to a web page load, not vitals to a screen. 

The way I look at is the navigation is the process that was triggered and items like first draw are events of the navigation.

> It was also mentioned that we could have a span for the screen but I'm thinking this will have similar long running concerns as a session span and won't be viable.

Long running shouldn't be an issue as the span would only represent the navigation process.

> I'm assuming the suggestion was to add these as span events? But I thought those are being deprecated?

Yes and no, the span events api is being deprecated however events will still be able to be able to be logged via the logger api. These events can also still be associated with the span/trace.

> is the suggestion to combine both of these spans into one and have the timing defined as attributes

That is one option but additional attributes would be needed on that span. the focus has been on thinking about it each stage as an event.

model/app/spans.yaml
limhjgrace: Actually reopening this as I don't think it should be an event given it has a duration. 

My understanding with long running spans is that the concern comes from the risk of data loss since it can result in an all or none scenario, assuming the span is being used as the root span for many child spans. For example, the session span. Having a long running session span is risky assuming all the spans within the session with have this as its parent span. If this span isn't sent for some reason, all the child spans are lost. 

For this span, however, the idea is to create the span only once at the end of the screen's visibility, like how we would if we were to create a log. We can store the timestamps locally and avoid creating a long span by doing span.setStartTime(prev).end(now) once the screen is no longer visible.

model/app/spans.yaml
LikeTheSalad: Thank you for all the details you've shared, @limhjgrace, and for your patience. I've had issues in the past around deprecations of attributes, and renaming of events, and it's no fun, because even though they are deprecated in the spec, if a vendor has delivered them in their tools, chances are that they must continue supporting those for a while after the spec is deprecated. Which is why I think it makes sense to try and be sure about how much value a new spec provides upfront, which can be tricky, of course, because we can't predict the future, but I've found that we can follow a couple of guidelines that help reduce the need to add breaking changes later:

1. Starting small. That is, with the least amount of attributes/enums and events as possible, because it's always possible and easier to add stuff later rather than having to remove stuff.
2. (if possible) Having a real-world use case that needs these signals. It's probably the most helpful one, because it takes the burden off of ourselves of having to "guess" what real users need and how we can present the data in a way that's helpful to them.
3. (if possible) Having a working implementation that has been used for a while. This one might sound like doing things backwards, though it often helps with things such as knowing what's even possible to implement and how, such as what I mentioned earlier about not being practical to define a span for an app launch, as the launch also covers the OTel client initialization, so starting a span before the client is ready is not possible, leaving implementations to resort to probably hacky-ish solutions, which I'm not completely against doing, but I'd like to reduce them as much as possible.

Having said that, I think we're making good progress here, though I figured that the communication might be easier if I explained the background of my decision-making process around these issues, which I'm open to discussing too, in case some of the things I mentioned don't make sense. So, regarding your comments:

> But for the duration calculation, I wanted to do this client side instead of server-side or query-time as this adds scalability, performance, and technical concerns. Is there a concern in adding a duration attribute to spans even though it can be derived from the start/endtime?

I'm not against adding a `duration` attribute to events, for example, as there's no other way to get that info out of those. With spans, however, it can be calculated as I mentioned earlier, which makes it a bit redundant, and so I think it's not compatible with the "starting small" (point 1) mentioned either. It seems like the reason to add it is about performance, though if that's the reason alone, it sounds to me that we might be falling into a premature optimization situation, unless we have a use-case that both requires this to be a span and also to have its duration precalculated. Otherwise, what prevents us from asking the same for every other span as well? I.e. things like "if this span has a precalculated duration attribute, shouldn't every client span have it too to increase performance everywhere?".

> Yes, this is a valid use case for services who want to provide a way to visualize an entire session. There was an on going discussion on how we could provide this view by having the specs support a session span but given the concerns of a long running span, the community didn't align on this.

I think this is a key point, because it sheds some light on what could be a potential real-world use case for these signals. If the idea is to provide a session-wide visualization, I love it and I'm all up for being able to provide something like that, though given that the "standard" way of visualizing data covers spans only, it leaves the idea of using events as not too useful for these cases, as they won't typically show up alongside spans, which I think can make the idea of "making everything a span" interesting, as it will ensure visibility for these events in traditional UI solutions. However, as you mentioned, it isn't easy, or even possible in some cases, to successfully create long-running spans on clients, which is why the session-related talks in the client sig focus on events instead. I think ideally we should have UI solutions that show both spans and events in a single timeline, but for the time being, I guess the only way to fulfill a "session view" use case is by using spans, and I'm not even sure if sessions will stay as events or will evolve to something else in the future (@bidetofevil might have some insights on this), so I'd say that this would be a point in favor of choosing a span for this use-case.

> To me, a screen load and time to first draw are both processes and is analogous to a web page load, not vitals to a screen

I agree with what @thompson-tomo replied:

> The way I look at is the navigation is the process that was triggered and items like first draw are events of the navigation.

Now, it boils down to what's needed, because in the end, both points of view can be valid depending on what's the problem we want to solve. For example, if we want to know not only how long does it take for a screen to load but also what subprocesses happen within that span of time, then a span makes sense. But if we just want to know what screens are slow to load, then an event should do. Just by the name, it seems like it might make sense to make "screen load" a span and "time to first draw" an event with `duration` to make queries easier for the duration part alone. So I don't think it has to be either/or, it may be a combination of things, but we can have an endless discussion about it unless we know what we want to solve with these, so if you have a real-world use case that you need to address I think it would help us better understand what's the minimum amount of signals/attributes needed to start small but to solve at least a basic issue first.

> It was also mentioned that we could have a span for the screen but I'm thinking this will have similar long running concerns as a session span and won't be viable.

I'd say is less tricky than having a span per session, but it still can be challenging in case the screen is abruptly closed, causing the span not to end properly and therefore losing the data. Still, despite the challenges, if we wanted to capture not only a screen duration but also all the subprocesses that happen within, I can't think of a better solution than using a span. There will be details to polish, such as apps that can have a single screen open for a long time, such as for kiosks, and how much data can UIs show in a single waterfall view, that I think should be defined. Now, I'm not sure we're conflating different issues here, because my understanding so far of this PR is that it's trying to add signals to track times only (time for app start, for screen load, and for time on screen), which, at first glance, sounds like events should do.

> That and I'm assuming the suggestion was to add these as span events? But I thought those are being deprecated?

I haven't kept track of this in detail, but my understanding is that the concept of an event being attached to a span as the only way to create events is what's changing, and now it should be possible to have standalone events, as well as events that are somehow attached to spans, and my guess is that it'd make the span-specific APIs to create events become redundant, maybe causing those to probably get deprecated? I don't have too many details on it though, so don't take my word for it.

> Maybe this is a question of what is a span and what is a log event given there is some overlap between the two where a log event should be used in place of creating long running spans.

In my experience, these questions are easier to answer when there's a real-world use case to address, because it makes the end goal clear. For example, if the use case is to just track the duration of these events to query them later to do some calculation, such as the average load time, or things like that, to me for this case the answer is easily an event.

> Technically, most spans could also just be defined as a log event and the discussion feels like it's heading towards that direction.

If we're creating a pair of events, one that tracks the start of something, and another that tracks the end of that something, I think we'd be overlapping with spans quite a lot (though leaving outside key span features such as trace context propagation). What I've proposed a couple of times is to have only an event that carries the whole duration of a process, because based on the descriptions, it seems like that's the question we're trying to answer.

> If so, are we defining spans to be used only for tracking a process for its child processes? Or is the suggestion to combine both of these spans into one and have the timing defined as attributes (e.g. attributes.time_to_first_draw and attributes.time_to_first_stable_draw)?

What's the use case you're trying to address? Is it just having some sort of metrics that can be aggregated, or is it displaying a hierarchy of processes? And if the answer to that is "for now some metrics, but in the future we might need to display some of these processes as traces", then my questions would be: Is it needed to combine these use cases in single span/event? Also, is it needed to have everything defined upfront in this PR, or could further conventions be added in the future as needed?

limhjgrace: Thanks for the detailed response and all the thought you're putting into this! I definitely agree that having to support deprecated schemas becomes painful very quickly. 

I have two main use cases:
1. Session timeline, kind of like the network requests in the console network tab where you can see horizontal bars for the requests over time and what DataDog currently supports (https://docs.datadoghq.com/tracing/trace_explorer/visualize/). In addition, I want to include both spans and events so that I can see everything that happened in the session in one glance. If something occurred and it had a duration, I want to see the horizontal bar represent its duration. If it was just a moment in time, I'll see it as just a line to indicate something happened even though it had 0 duration. 
2. I want to be able to aggregate data across sessions. For example, I want to know the p99 time to first draw duration, the average time on screen, the max app launch time, etc. I need one singular duration value to do these aggregations so if the start and endtime is the only thing provided, I would have to calculate and store this duration server side so that I can support these aggregations. However, I'm okay with removing this from p0 if it will unblock PR. 

In terms of a working implementations, I could point to the current OTel Android SDK where: 
1. AppStart is a span ([ref](https://github.com/open-telemetry/opentelemetry-android/blob/672413652aa1970c47e0723c9d8acd796253ae8d/instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityTracer.java#L91-L95))
2. TimeToFirstDraw is being contributed by Leo as a span (https://github.com/open-telemetry/opentelemetry-android/pull/1281)

As for the visibility span, I see that Embrace is also recording this as a span: https://github.com/embrace-io/embrace-apple-sdk/blob/836f3f46d7cdb30cc3a3ef83dfc83a6b3803fd52/Sources/EmbraceCore/Capture/UX/View/UIViewControllerHandler.swift#L299-L309

LikeTheSalad: Hi @limhjgrace 

Thank you for your patience and for explaining your use cases. Based on what you mentioned, you definitely need a span to show it as a trace, and you also need some value to aggregate durations, and you'd like to provide this for individual screen loading as well as the app launch process, the latter I think makes reviewing this PR less straightforward because it implies that the same approach for screens can be extrapolated to the app launch, which is mostly where I have concerns. Also, the proposal seems to mix not only different use cases, but it also seems to make signals step on each other's toes, which I agree is not always easy to find the right one for the right job, but if, on top of that, we add different use cases into the mix, the conversation can become difficult to enable a way forward.

So to move forward, I'd suggest the following:

- Pick a single use case between the screen load and the app launch, preferably the screen load one, so that we can focus on it properly and then address the app launch one in a separate PR.
- Regarding the "working implementations" that you mentioned, @Doohl's [PR](https://github.com/open-telemetry/opentelemetry-android/pull/1281) might not entirely fit what I mentioned about _"a working implementation that has been used for a while"_, but still, I think there is some prior work that addresses similar issues, so it's not an entirely new use case, so I think we can try and make this work by checking on both PRs at the same time. I'll follow up with @Doohl's [here](https://github.com/open-telemetry/opentelemetry-android/pull/1281) to see if we can support this semconv PR from there.
- Regarding the other working implementation that you mentioned from OTel Android ([AppStart](https://github.com/open-telemetry/opentelemetry-android/blob/672413652aa1970c47e0723c9d8acd796253ae8d/instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityTracer.java#L91-L95)), it might not be what its name suggests, unfortunately. It seems to cover only the creation of the first Activity, which ignores the previous part of the app creation that covers an app's launch. This is one of the reasons why I think it'd be better to focus on the screen load use case first, because the app launch one might take a while to get sorted, so I think it deserves its own dedicated place to discuss it.
- Lastly, regarding the screen load scenario, I think what you need sounds quite similar to what's already done for HTTP calls, for example, in the sense that they also need to be shown as a trace, and there's also a duration attribute for them that serves the aggregation purposes. I think we should try to be as consistent as possible, so it makes sense to me to try and follow such scenarios as a baseline for similar new ones. For the case of HTTP, they use 2 signals, [spans](https://opentelemetry.io/docs/specs/semconv/http/http-spans/) and [metrics](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/). The duration attribute is captured as a [metric](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration), so that would be my initial thought to go about it. However, since metrics aren't something that can be tracked per session, I think the next best option would be an event. So I'd suggest defining both for the screen load use case.

limhjgrace: > preferably the screen load one

Done

> I'll follow up with @Doohl's https://github.com/open-telemetry/opentelemetry-android/pull/1281 to see if we can support this semconv PR from there.

Took a look at Leo's PR and also left comments re. semConv 

>  its own dedicated place to discuss it.

That's fair. Created separate PR for this: https://github.com/open-telemetry/semantic-conventions/pull/2965

>  However, since metrics aren't something that can be tracked per session, I think the next best option would be an event. So I'd suggest defining both for the screen load use case.

I disagree about adding it as both a span and event. But as I mentioned before, I'm fine with dropping duration as this can be calculated server side. Removed `*.duration` attributes in next revision to unblock PR.

limhjgrace:
It looks like the PR was auto closed because there's no dedicated SIG for `area:app`. @thompson-tomo could I know how to proceed forward in getting this PR reviewed? Similar scenario for the PR I created for app start: https://github.com/open-telemetry/semantic-conventions/pull/2965

lmolkova:
@limhjgrace I believe this should fix the automation: https://github.com/open-telemetry/semantic-conventions/pull/2968 (but until it's merged any push would close this PR again, sorry about this)

model/app/spans.yaml
thompson-tomo: Thanks for descoping the pr and focusing on a single  use case. As it stands the biggest question to be addresses is when a screen load span should end and based on that update the name. 

My feeling based on comments in linked pr is that it should encompass the time until fully drawn. Then in subsequent pr's associate details about time taken to get to first draw etc.

limhjgrace: Actually, my intention with this PR is to just define the span for time to first draw. I wanted to descope defining the end of a screen load span since getting alignment on this is looking to be a lot more significant than I expected.

model/app/spans.yaml
```diff
@@ -0,0 +1,20 @@
...
+    type: span
+    span_kind: internal
+    brief: This span represents an application screen first draw operation.
+    note: |
+      **Span name**: MUST be `app.screen.time_to_first_draw`.
+
+      **Span kind** MUST be `INTERNAL`.
+
+      This span captures the time from screen initialization to the first frame
+      of an application UI screen being drawn.
```
LikeTheSalad, line 12: My understanding from this description is that the span covers an initialization process done **before** drawing a screen's UI? Is that correct? I'm asking because at first I thought it was about covering the UI first draw, which may be the case and I'm not properly understanding the description, so I wanted to double-check.

breedx-splk:
Let's keep this conversation going please.

PR #2835, opened 2025-09-30
Add automation to flag PRs on areas without an active SIG/project

joaopgrassi:
Fixes ##2834

## Changes

TODO

### DIFF
--- a/.github/scripts/triage/check-changes-ownership.js
+++ b/.github/scripts/triage/check-changes-ownership.js
@@ -1,0 +1,37 @@
+const utils = require('./utils')
+
+async function changesInInactiveAreas(core, context, github, areas, changedAreas) {
+    let changesWithoutOwners = [];
+    const areaOwnersMap = utils.getActiveAreasWithCodeOwners(areas);
+
+    changedAreas.forEach(ca => {
+        if (!areaOwnersMap.has(ca)) {
+            changesWithoutOwners.push(ca);
+        }
+    });
+
+    if (changesWithoutOwners.length > 0) {
+        core.setFailed(`The pull request PR contains changes to areas that are not active:\n${changesWithoutOwners.join('\n')}`)
+        await github.rest.issues.addLabels({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              labels: ['triage:rejected:declined']
+            })
+        await github.rest.pulls.createReviewComment({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            pull_number: context.issue.number,
+            commit_id: context.sha,
+            body: 'PR contains changes to areas that are not active'
+        });
+    }
+
+    return changesWithoutOwners;
+}
+
+const findChangesInInactiveAreas = async (core, context, github, areas, changedAreas) => {
+    await changesInInactiveAreas(core, context, github, areas, JSON.parse(changedAreas));
+}
+
+module.exports = findChangesInInactiveAreas;
--- a/.github/scripts/triage/utils.js
+++ b/.github/scripts/triage/utils.js
@@ -1,0 +1,27 @@
+function isAreaActive(area) {
+    return area.status.includes('accepting_contributions') || area.status.includes('active');
+}
+
+function getActiveAreasWithCodeOwners(areas) {
+    const areaOwnersMap = new Map();
+    areas.filter(a => isAreaActive(a)).forEach((area) => {
+        area.labels.forEach((l) => {
+            const labelWithoutPrefix = l.slice('area:'.length);
+            // use all areas as map key
+            if (!areaOwnersMap.has(labelWithoutPrefix)) {
+                areaOwnersMap.set(labelWithoutPrefix, new Set());
+            }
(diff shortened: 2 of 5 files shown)


.github/scripts/triage/check-changes-ownership.ts
```diff
@@ -0,0 +1,88 @@
...
+import { Octokit } from "@octokit/action";
+
+const octokit = new Octokit();
+
+const [owner, repo] = process.env.GITHUB_REPOSITORY!.split("/");
+const prNumber: number = +process.env.PR_NUMBER!;
+const changes: string[] = process.env.CHANGED_FILES!.split(',');
+
+/**
+ * Checks if the PR already has the 'triage:accepted:ready' label, meaning the triage checks should be skipped.
```
jsuereth, line 11: Are there other labels we have (other than ready) that would make us want to skip the check?

joaopgrassi: Not today, but I can add if you prefer a more explicit one for that purpose. Let me know!

.github/workflows/check-changes-ownership.yml
```diff
@@ -0,0 +1,60 @@
...
+          files: model/**
+          separator: ","
+
+  validate-area-ownership:
+    runs-on: ubuntu-latest
+    needs: get-changed-files
+    permissions:
+      contents: read
+      pull-requests: write
+    if: ${{ needs.get-changed-files.outputs.any_changed == 'true' }} && ${{ github.repository_owner == 'open-telemetry' }}
```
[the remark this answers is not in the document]
arminru: Not sure if that really matters, but in their example, they do combine it within one expression:
https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#example

.github/workflows/check-changes-ownership.yml
```diff
@@ -0,0 +1,60 @@
+name: 'Check changes ownership'
+on:
+  pull_request_target:
+    types: [opened, synchronize]
+    branches: [ 'main*' ]
```
pichlermarc: Q: do you use other branches than `main` exactly? 🤔 
If the answer is no then using limiting it to `main` may be a good idea 🙂 
```suggestion
    branches: [ 'main' ]
```

.github/workflows/check-changes-ownership.yml
```diff
@@ -0,0 +1,60 @@
...
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+
+      # uses node 22 to be able to use Typescript
+      - uses: actions/setup-node@v5
+        with:
+          cache: 'npm'
+          cache-dependency-path: |
+            .github/scripts/triage/package-lock.json
```
pichlermarc, line 48: IIRC for this to work the file must exist - so it might be necessary to add it to the PR 🙂

.github/workflows/check-changes-ownership.yml
```diff
@@ -0,0 +1,60 @@
...
+          node-version: '22'
```
pichlermarc: I see you've been using `@types/node@24` in `devDependencies` but `22` here. That's not nessarily a problem, but the type-definition may be broader than what's available in `22`.

Aligning them is optimal 🙂 
```suggestion
          node-version: '24'
```

.github/workflows/check-changes-ownership.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Install script dependencies
+        working-directory: .github/scripts/triage
+        run: npm install
```
pichlermarc: If you decide to `package-lock.json` you can use `npm ci --ignore-scripts` here. That will use the exact lockfile you used for installing and can improve stability. (`--ignore-scripts` can also be used with `npm install` and disallows postinstall scripts that were used in the recent npm supply chain attacks)

pichlermarc (commented):
looks good, just a few nits while I'm here 🙂

.github/scripts/triage/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "type": "module",
+  "scripts": {},
+  "keywords": [],
+  "author": "Semantic Conventions",
```
lmolkova: nit:

```suggestion
  "author": "OpenTelemetry Authors",
```

like in https://github.com/open-telemetry/opentelemetry-js-contrib/blob/23c5f4e8884ba87ab2c697bf34b34e106ed3abc5/package.json#L44C14-L44C27

?

joaopgrassi: Sure! Changed.

.github/workflows/check-changes-ownership.yml
joaopgrassi: Nice! thank you for the hint. I will go with `npm ci`

.github/scripts/triage/check-changes-ownership.ts
joaopgrassi: Closing this. we can always change and add a different label if we prefer.

Issue #2703, opened 2025-08-29
RPC: adopt system-specific naming practices


lmolkova:
Implement https://github.com/open-telemetry/semantic-conventions/pull/1708 for RPC

See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#system-specific-naming  and https://github.com/open-telemetry/semantic-conventions/pull/1734 for the details

- [ ] rename `rpc.system` to `rpc.protocol|system|etc.name`
- [ ] review RPC system list and clean it up
- [ ] remove `rpc` prefix from system-specific conventions

steverao:
Following the discussion of RPC stability on September 25, 2025 (CST), I would like to add some comments regarding point one, based on my investigation of the Apache Dubbo framework’s protocols.
1. In Dubbo 3.x, a native protocol called Triple has been implemented based on gRPC, allowing Dubbo clients to directly invoke gRPC servers and vice versa. For more details, please refer to the [related documentation](https://dubbo.apache.org/en/overview/mannual/java-sdk/tasks/protocols/triple/grpc/). Additionally, this [demo](https://github.com/apache/dubbo-samples/tree/master/2-advanced/dubbo-samples-triple-grpc) illustrates how to use the Triple protocol, and the trace structure appears as follows: <img width="2022" height="866" alt="Image" src="https://github.com/user-attachments/assets/063cdbcf-9b98-4081-9a87-677c5256f2b2" />
2. Besides Triple, Dubbo also provides another native protocol named Dubbo, which is based on TCP and was introduced in version 2.x.

Based on above information, I suggest renaming `rpc.system` to `rpc.protocol.name`. It can tell the users the specific communication protocol of Dubbo application. We can use `otel.scope.name` to record framework-specific information just like mentioned by @trask

PR #2842, opened 2025-10-01
#2703 define guidance for rpc protocol vs framework

thompson-tomo:
Progresses #2703 

## Changes

This seeks to adopt system specific naming for the rpc namespace by first defining the attributes and what the requirements are to be considered a well known value for either the framework and/or protocol attribute.


docs/rpc/README.md
```diff
@@ -24,4 +24,49 @@ Specifications defined by maintainers of RPC systems:
...
+These implementation's will usually only expose a subset of functionality of the network protocol
+and may only be compatible with newer versions of the network protocol ie grpc will not work over http v1.
+
+Some pre-defined protocols are:
+
+* [gRPC](#)
+* [http](#)
+* [triple](#)
+* [soap](#)
+* [Connect RPC](#)
```
KalleOlaviNiemitalo: Why are these links to `#`?

docs/rpc/README.md
```diff
@@ -24,4 +24,49 @@ Specifications defined by maintainers of RPC systems:
...
+The framework in adddition to providing the api can define properties which are sent along with the message
+to provide additional context. For instance a framework might require a message id to be sent with the message.
+The framework doesn't care how it is transported across the network.
+
+Some pre-defined frameworks are:
+
+* [Apache Dubbo](#)
+* [Dapr](#)
+* [Connect RPC](#)
+* [JSON-RPC](#)
```
KalleOlaviNiemitalo: I presume WCF would also go here.  It is pretty much SOAP-focused but has been extended to HTTP without SOAP too.

KalleOlaviNiemitalo: I'm not sure I'd categorize JSON-RPC as an RPC Framework.  The JSON-RPC 1.0 and 2.0 specifications focus on message formats, rather than on APIs.

However, when JSON-RPC is used over a stream-based network protocol, there are different ways of framing the JSON-RPC messages, i.e. indicating where each message starts and ends.  Would you then call each framing method an RPC protocol?

docs/rpc/README.md
```diff
@@ -24,4 +24,49 @@ Specifications defined by maintainers of RPC systems:
...
+
+It is expected that these protocols will define protocol specific attributes, for instance gRPC would contain a status attribute.
+As such the usage of these attribute/s should be documented via a dedicated protocol page providing complete definitions.
+
+### What is a RPC Framework?
+
+A RPC Framework describes the api's made available to applications wanting to implement RPC Communications
+in a protocol agnostic manner. This is why protocol and framework is not a one-to-one relationship but rather one-to-many.
+
+The framework in adddition to providing the api can define properties which are sent along with the message
```
KalleOlaviNiemitalo: "addition"

docs/rpc/README.md
```diff
@@ -24,4 +24,49 @@ Specifications defined by maintainers of RPC systems:
...
 
+## RPC Service vs RPC Framework
+
+### What is a RPC Protocol?
+
+A RPC protocol describes the manner in which a message is transported from one service to another.
+This protocol may have the same value as the network protocol ie http or it may differ when,
+an implementation of the network protocol is used ie gRPC.
+
+These implementation's will usually only expose a subset of functionality of the network protocol
```
KalleOlaviNiemitalo: "implementations"

docs/rpc/README.md
```diff
@@ -24,4 +24,49 @@ Specifications defined by maintainers of RPC systems:
 
 * [gRPC](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md): Semantic Conventions for *gRPC*.
 
+## RPC Service vs RPC Framework
+
+### What is a RPC Protocol?
+
+A RPC protocol describes the manner in which a message is transported from one service to another.
+This protocol may have the same value as the network protocol ie http or it may differ when,
+an implementation of the network protocol is used ie gRPC.
```
KalleOlaviNiemitalo: I think you mean "e.g." (for example) rather than "i.e." (that is)

docs/rpc/README.md
```diff
@@ -24,4 +24,49 @@ Specifications defined by maintainers of RPC systems:
 
 * [gRPC](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md): Semantic Conventions for *gRPC*.
 
+## RPC Service vs RPC Framework
```
KalleOlaviNiemitalo: The text under this heading talks about "RPC protocol" and "RPC Framework" but not "RPC Service".

docs/rpc/README.md
```diff
@@ -24,4 +24,49 @@ Specifications defined by maintainers of RPC systems:
...
+
+These implementation's will usually only expose a subset of functionality of the network protocol
+and may only be compatible with newer versions of the network protocol ie grpc will not work over http v1.
+
+Some pre-defined protocols are:
+
+* [gRPC](#)
+* [http](#)
+* [triple](#)
+* [soap](#)
```
KalleOlaviNiemitalo: OK, so for SOAP as the RPC protocol, the network protocol might be HTTP or SMTP or FTP at least.

docs/rpc/README.md
thompson-tomo: Have added wcf to the list

docs/rpc/README.md
thompson-tomo: Correct.

docs/rpc/README.md
thompson-tomo: Thanks for identifying that.

docs/rpc/README.md
thompson-tomo: Hmm, interesting point. I do agree that json-rpc is difficult to place as it is not at the same level as the other frameworks in terms of functionality and in fact is most similiar to cloud events. As such I am starting to think that it should look at one of the following options:
- Introduce a separate attribute such as `rpc.message.specification.name`
- we adapt the definition of a framework.
- Use attributes in the transport protocol to capture it ie http.request.header.<key>

Let's decide how to handle it via #2871

KalleOlaviNiemitalo:
In some other semconv issues, streaming was mentioned.  That reminded me about Blocks Extensible Exchange Protocol (BEEP, [IETF RFC 3080](https://www.rfc-editor.org/rfc/rfc3080)), which has streaming responses: you send a `MSG` message, and then get back any number of `ANS` messages and finally a `NUL` message.  BEEP however does not define any protocol elements for method names or arguments — message payloads just consist of bytes — so in the terminology of this PR, it's a network protocol (like HTTP), rather than an RPC protocol or an RPC framework.  One can run SOAP over BEEP, as per [IETF RFC 4227](https://www.rfc-editor.org/rfc/rfc4227).

docs/rpc/README.md
```diff
@@ -103,4 +106,10 @@ with a general "if applicable for the rpc framework" condition placed on the req
 A framework may have its own page when the framework defines its own signals which require the inclusion of framework specific
 attributes for the signal to have value & meaning.
 
+## Supplementry Documents
```
KalleOlaviNiemitalo: "Supplementary"

docs/rpc/README.md
```diff
@@ -34,16 +34,78 @@ This document defines semantic conventions for remote procedure calls (RPC)
...
+
+These implementations will usually only expose a subset of functionality of the network protocol
+and may only be compatible with newer versions of the network protocol ie grpc will not work over http v1.
+
+Another aspect of how the `network.protocol.*` differs to `rpc.transport.protocol.*` is that,
+the transport protocol can implement additional client side functionality such as retry, caching, cancellation etc.
+
+Some pre-defined transport protocols are:
+
+* [gRPC](grpc.md)
```
steverao: How about replace the semconv url by corresponding introduction url here?

docs/rpc/README.md
```diff
@@ -34,16 +34,78 @@ This document defines semantic conventions for remote procedure calls (RPC)
...
+
+The framework in addition to providing the api can define properties which are sent along with the message
+to provide additional context. For instance a framework might require a message id to be sent with the message.
+The framework doesn't care how it is transported across the network.
+
+Some pre-defined frameworks are:
+
+* Apache Dubbo
+* Dapr
+* [Connect RPC](connect-rpc.md)
```
steverao: Similar question with above, how about using introduction url of related RPC framework, we could have another part to list all related semconv urls below like https://github.com/open-telemetry/semantic-conventions/tree/c8c3066e24b81d1663c0b84434b6da517f6c4210/docs/http

docs/rpc/README.md
thompson-tomo: This is similar to what is done on other pages and this way a reader has quick access to the corresponding definitions. It also ties into earlier paragraphs.

Details about the protocol would be linked via the attribute value.

docs/rpc/README.md
thompson-tomo: Note the page would automatically gain the list when published online.

docs/rpc/README.md
```diff
@@ -7,7 +7,7 @@ linkTitle: RPC
 **Status**: [Development][DocumentStatus]
 
 This document defines semantic conventions for remote procedure calls (RPC)
-, also called "remote method invocations" (RMI).
+, sometimes called "remote method invocations" (RMI).
```
trask: I'd support dropping RMI, I think it's only in Java and even then it's pretty old Java tech...

docs/rpc/README.md
```diff
@@ -34,16 +34,78 @@ This document defines semantic conventions for remote procedure calls (RPC)
...
+This remote process call will often travel across the network to reach it's destination.
+
+The details of this invocation can be captured using the following signals:
 
 * [RPC Spans](rpc-spans.md): Semantic Conventions for RPC client and server *spans*.
 * [RPC Metrics](rpc-metrics.md): Semantic Conventions for RPC *metrics*.
 
-Technology specific semantic conventions are defined for the following RPC systems:
+These generic conventions can be extended based on the transport protocol/framework being used.
+This is described in more detail in the [RPC Transport Protocol vs RPC Framework](#rpc-transport-protocol-vs-rpc-framework) section.
```
trask: I'd suggest moving list of semantic convention for rpc protocols here, and then below can explain protocol vs framework.

```suggestion
```

docs/rpc/README.md
```diff
@@ -6,47 +6,106 @@ linkTitle: RPC
...
+- [Database](/docs/database/README.md)
+- [Generative AI](/docs/gen-ai/README.md)
+- [Messaging](/docs/messaging/README.md)
+
+If your focus is on the network layer calls rather than the logical calls being made,
+these RPC documents are not for you but instead what you are after is described via the below conventions:
+
+- Http
+
+## RPC Protocol vs RPC Framework
```
lmolkova, line 73: do we need to describe this at all? The difference is very subjective and nothing(?) in the semconv depends on this terminology being established

thompson-tomo: The purpose was to provide clear documented guidance/explanation which serves as the definition of the corresponding attributes. This can then be used when new values are requested for additional instrumentation and it avoids the issues which we have with messaging.system.

lmolkova: I still don't see how it contributes to explaining what comes into `rpc.system.name` and what not or how it provides subjectivity in differentiation. I don't think this change is necessary

thompson-tomo:
For rpc if we are wanting to federate responsibility of conventions there needs to be a common definition which can be followed.

docs/rpc/README.md
thompson-tomo: This PR pre-dates decision to remain with rpc.system and is from when we were discussing having the 2 attributes.

The key thing is:
- protocol describes how the message is transported
- framework describes the apis used to send/recieve/process messages

Scenario is if you are using connectrpc client to send a grpc request, the protocol would be grpc with the framework on the client being connectrpc but on server could be Dubbo or something else.

PR #2843, opened 2025-10-01
Update Heroku Semantic Conventions to Match Official Documentation

anna-cross:
## Changes

The Heroku Telemetry Team is  looking to update the current Heroku semantic conventions to match our existing documentation for Fir generation applications. This PR aligns the semantic conventions with the official [Heroku OpenTelemetry Signals and Attributes Reference](https://devcenter.heroku.com/articles/heroku-opentelemetry-signals-and-attributes-reference).

#### **Updated Attributes to Match Official Documentation**
- **Removed non-documented attributes**: `heroku.dyno_index`, `heroku.dyno_type`, `heroku.release.creation_timestamp`, `heroku.release.commit`
- **Kept only officially documented attributes**:
  - `heroku.app.id` - unique identifier of app
  - `heroku.app.name` - name of application  
  - `heroku.release.id` - unique identifier of app release
  - `heroku.release.version` - version of app release when telemetry is generated
  - `heroku.workload.id` - workload identifier, as defined by the process type

#### **Additional Information**
- All attributes are automatically provided by the Heroku platform
- Attributes are specific to Heroku Fir applications (not traditional Cedar dynos)
- Service-level attributes can be configured via standard OTEL environment variables (`OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`)

(Note: I'm currently going through internal process of getting CLA signed) 

### DIFF
--- a/docs/resource/cloud-provider/heroku.md
+++ b/docs/resource/cloud-provider/heroku.md
@@ -12,13 +12,25 @@
 
 **type:** `heroku`
 
-**Description:** [Heroku dyno metadata](https://devcenter.heroku.com/articles/dyno-metadata)
+**Description:** Heroku Fir Application Attributes as defined in the  [Heroku OpenTelemetry Signals and Attributes Reference](https://devcenter.heroku.com/articles/heroku-opentelemetry-signals-and-attributes-reference)
 
 | Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
 |---|---|---|---|---|---|
-| [`heroku.app.id`](/docs/registry/attributes/heroku.md) | string | Unique identifier for the application | `2daa2797-e42b-4624-9322-ec3f968df4da` | `Opt-In` | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`heroku.release.commit`](/docs/registry/attributes/heroku.md) | string | Commit hash for the current release | `e6134959463efd8966b20e75b913cafe3f5ec` | `Opt-In` | ![Development](https://img.shields.io/badge/-development-blue) |
-| [`heroku.release.creation_timestamp`](/docs/registry/attributes/heroku.md) | string | Time and date the release was created | `2022-10-23T18:00:42Z` | `Opt-In` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`heroku.app.id`](/docs/registry/attributes/heroku.md) | string | The unique identifier of the Heroku application. This is a UUID that uniquely identifies the application across the Heroku platform. [1] | `9daa2797-e49b-4624-932f-ec3f9688e3da`; `c3d3df33-8afb-4323-ac49-a9bf41a50dd1` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`heroku.app.name`](/docs/registry/attributes/heroku.md) | string | The name of the Heroku application. This is the human-readable name of the application as it appears in the Heroku dashboard. [2] | `my-app`; `test-app` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`heroku.release.id`](/docs/registry/attributes/heroku.md) | string | The unique identifier of the app release. This is a UUID that uniquely identifies each deployment/release of the application. [3] | `release-afc4d88c-7d89-4bc6-b364-7658cd60ba57`; `release-3bd90b80-16a1-4f5b-9465-a61e1b7464d4` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`heroku.release.version`](/docs/registry/attributes/heroku.md) | string | The version of the app release when telemetry is generated. This is typically a version number or identifier for the specific release. [4] | `v19`; `v42`; `v1.2.3` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+| [`heroku.workload.id`](/docs/registry/attributes/heroku.md) | string | The workload identifier, as defined by the process type. This typically corresponds to the process type (web, worker, etc.) or specific workload identifier in the telemetry system. [5] | `web`; `worker`; `scheduler`; `release` | `Recommended` | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
+
+**[1] `heroku.app.id`:** This attribute is automatically provided by Heroku and cannot be customized.
+
+**[2] `heroku.app.name`:** This attribute can be customized by renaming the application in the Heroku dashboard. It is used as the default value for service.name if not explicitly set.
+
+**[3] `heroku.release.id`:** This attribute is automatically provided by Heroku.
+
+**[4] `heroku.release.version`:** This attribute is automatically provided by Heroku. It is commonly used as the default value for service.version.
+
+**[5] `heroku.workload.id`:** This attribute is automatically provided by Heroku and is used to identify the workload type. It is particularly useful for filtering telemetry data by process type.
 
 <!-- markdownlint-restore -->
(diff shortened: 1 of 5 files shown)


model/heroku/registry.yaml
```diff
@@ -3,23 +3,64 @@ groups:
...
+        note: >
+          This attribute is automatically provided by Heroku and cannot be customized.
+
+      - id: heroku.app.name
         type: string
-        stability: development
+        stability: stable
         brief: >
-          Commit hash for the current release
-        examples: [ 'e6134959463efd8966b20e75b913cafe3f5ec' ]
```
thompson-tomo, line 19: By convention we don't remove attributes in semantic  conventions but instead we deprecate them, so could you please add the deprecation object/properties to the old object rather than removing them.

model/heroku/entities.yaml
```diff
@@ -4,11 +4,11 @@ groups:
...
-      - ref: heroku.release.creation_timestamp
-        requirement_level: opt_in
-      - ref: heroku.release.commit
-        requirement_level: opt_in
       - ref: heroku.app.id
-        requirement_level: opt_in
+      - ref: heroku.app.name
+      - ref: heroku.workload.id
+      - ref: heroku.release.id
+      - ref: heroku.release.version
```
thompson-tomo, line 25: Could you have an attempt at setting the role of the attributes as per https://opentelemetry.io/docs/specs/semconv/how-to-write-conventions/resource-and-entities/#how-to-define-identifying-attributes.

My thought would be the entity should actually be heroku.app with heroku.release.id & heroku.app.id as identifying with others being descriptive.

model/heroku/entities.yaml
```diff
@@ -4,11 +4,11 @@ groups:
...
+      [Heroku OpenTelemetry Signals and Attributes Reference](https://devcenter.heroku.com/articles/heroku-opentelemetry-signals-and-attributes-reference)
     attributes:
-      - ref: heroku.release.creation_timestamp
-        requirement_level: opt_in
-      - ref: heroku.release.commit
-        requirement_level: opt_in
       - ref: heroku.app.id
-        requirement_level: opt_in
+      - ref: heroku.app.name
+      - ref: heroku.workload.id
```
thompson-tomo: So an entity represents an object of interest associated with produced telemetry: traces, metrics, logs, profiles etc.

Based on that a heroku workload appears that it should be defined as a seperate object/entity.

anna-cross: Workload id is more of a process type of the application emitting the signals. We have "web", "worker", "scheduler" as possible examples. 

We do have other services / entities emitting signals like api, router or add on services. We can add them to the documentation or keep it application specific if it makes sense.

model/heroku/registry.yaml
anna-cross: Added deprecated section!

model/heroku/entities.yaml
anna-cross: That makes sense! I've added roles to both attributes

model/heroku/entities.yaml
```diff
@@ -1,14 +1,16 @@
...
-        requirement_level: opt_in
-      - ref: heroku.release.commit
-        requirement_level: opt_in
       - ref: heroku.app.id
-        requirement_level: opt_in
+        role: identifying
+      - ref: heroku.release.id
+        role: identifying
+      - ref: heroku.app.name
+      - ref: heroku.workload.id
```
thompson-tomo: These should have the descriptive role.

model/heroku/entities.yaml
thompson-tomo: Thanks, upon looking closer at the description of the attributes I am now having second thoughts about this entity especially after reading https://devcenter.heroku.com/articles/releases

What I now see is that we should have a `heroku.app` & a `heroku.release` entity this way when changing the config only a new release entity is created, the identifying would be the corresponding *.id attribute being the identifying attributes.

model/heroku/entities.yaml
thompson-tomo: Ok based on that and https://devcenter.heroku.com/articles/procfile `heroku.process.process_type` seems more accurate but I assume renaming it is not an option?

I would recommend removing `heroku.workload.id` from the `heroku.app` entity and instead placing it on a new `heroku.process` entity.

Reason for this is, my understanding is that an app can have multiple process types with each process type running in a seperate process. By splitting it we can have a common long lived `heroku.app` entity for all signals and then a short lived `heroku.process` entity for the process. The `heroku.process` would be an extension to the `process` entity as described in https://opentelemetry.io/docs/specs/semconv/how-to-write-conventions/resource-and-entities/#extending-an-entity

docs/resource/cloud-provider/heroku.md
```diff
@@ -27,13 +39,13 @@
...
-| `HEROKU_RELEASE_VERSION`           | `service.version`                   |
-| `HEROKU_SLUG_COMMIT`               | `heroku.release.commit`             |
+For FIR applications, service-level attributes can be configured using standard OpenTelemetry environment variables:
+
+| OTEL Environment Variable | Resource Attribute | Description |
+|---------------------------|-------------------|-------------|
+| `OTEL_SERVICE_NAME` | `service.name` | Service name (replaces default application name) |
+| `OTEL_RESOURCE_ATTRIBUTES` | `service.namespace` | Namespace for service name |
+| `OTEL_RESOURCE_ATTRIBUTES` | `service.version` | Version of app release |
+| `OTEL_RESOURCE_ATTRIBUTES` | `service.instance.id` | Unique dyno identifier (Heroku-provided) |
```
thompson-tomo: Don't see the need for this given it should be handled via the service documentation.

anna-cross: What if we create an entity "heroku.workload" instead of "heroku.process". It would be hard to rename that attribute. I've added it in my change, lmk if that looks right! 

I've seperated them into three entities entity.heroku.release, entity.heroku.app and entity.heroku.workload

thompson-tomo: Sure that works for me. I would add "extends: process" to the entity definition and then make the workload attribute descriptive.

anna-cross: I was able to add entity.process to workload definition like so: 

```
  - id: entity.heroku.workload
    type: entity
    stability: development
    name: heroku.workload
    extends: entity.process
    brief: >
      Heroku Fir Workload Attributes representing individual dyno instances running within a Heroku application.
    attributes:
      - ref: heroku.workload.id
        role: descriptive

```

But this appends `process.*` attributes to the `docs/resource/cloud-provider/heroku.md `that we don't provide, does that seem correct?

thompson-tomo: It does to me as if you deploy 2 copies of a release with the same process type you need a way to distinguish each instance. That can now be done via the process.pid and It also follows what is spoken about in that article.

jsuereth:
Can you please accept the CLA?

jsuereth:
@anna-cross Would you be interested in forming a group to "own" Heroku semantic conventions within Semconv?  We lack ownership of that area right now, so it's possible our automation will auto-close this.  Given Heroku is filling out these conventions, we'd like to work with you on modelling the Heroku entities.

docs/resource/cloud-provider/heroku.md
anna-cross: Hello! Apologies for late update but right now we are actually looking at adding two new attributes to replace the "heroku.workload.id" , with something like: 

Renaming:
- heroku.workload.id -> heroku.dyno.type 

Deprecating:
- heroku.workload.id

Adding:
- heroku.dyno.name (identifying attribute)

anna-cross:
@jsuereth Hello! Yes, I would love to form a group to own Heroku attributes. Right now we are still reworking attributes on our end, so it will be good to form a group. How would you like to go about forming it? 

Also sorry for taking long on CLA, should be resolved soon!

PR #2848, opened 2025-10-01
Update lychee to latest version which really does verify link fragments

trask:


ChrsMark (approved):

CHANGELOG.md
```diff
@@ -414,7 +414,7 @@ the VCS metrics to include `vcs.repository.name` as a recommended attribute.
 ### 🧰 Bug fixes 🧰
 
 - `service`: Merge `resource` experimental and stable groups for service and telemetry.sdk (#1423)
-  Discovered when fixing [weaver#306](https://github.com/open-telemetry/weaver/issues/306#issue-2458430277)
```
kamphaus, line 417: This was a valid link fragment.
Why was it removed?

trask: because lychee can't validate it, and I didn't think that linking to the PR description was any better than linking to the PR

kamphaus: Linking to the PR description does not bring much benefit indeed.
I was wondering whether linking to a comment might be of use, but that is probably not an appropriate level of detail for the CHANGELOG.

arminru (approved):

kamphaus (approved):

PR #2850, opened 2025-10-01
Small CI improvement

trask:


.github/workflows/build-system-check.yml
```diff
@@ -23,6 +23,7 @@ jobs:
   make-default-target:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false # it's helpful to know if only one of these fails
```
trask, line 26: also because every time I see the canceled job in the PR results I get confused

lmolkova (approved):

arminru (approved):

PR #2856, opened 2025-10-02
Fix `metric.v8js.memory.heap.space.physical_size` and `metric.v8js.memory.heap.space.available_size` metric names

maryliag:
## Changes
Aligning the name of metrics related to memory heap. Currently some of them don't have the `memory` in its name. We did the implementation for it in the JS SDK, using all the metric names with `memory`. It was a mistake on the semantic convention, and it should also have the `memory` on the name here.
This PR marks the previous name as deprecated (there is no implementation using those names anyway), and update the names.

Reference: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3130#discussion_r2396002810


docs/runtime/v8js-metrics.md
joaopgrassi: You need to also rename this one here to the new name `metric.v8js.memory.heap.space.available_size`

trask:

PR #2861, opened 2025-10-03
pprof: add pprof.profile.comment

florianl:
## Changes

The Profiling SIG intends to convert a field in the OTel Profiles signal into an attribute.

Friendly ping: @open-telemetry/profiling-approvers 


jhalliday (approved):

docs/general/profiles.md
```diff
@@ -81,6 +81,7 @@ They may be used in any Profiles record they apply to.
 | [`pprof.mapping.has_functions`](/docs/registry/attributes/pprof.md) | boolean | Indicates that there are functions related to this mapping. |  | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`pprof.mapping.has_inline_frames`](/docs/registry/attributes/pprof.md) | boolean | Indicates that there are inline frames related to this mapping. |  | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
 | [`pprof.mapping.has_line_numbers`](/docs/registry/attributes/pprof.md) | boolean | Indicates that there are line numbers related to this mapping. |  | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| [`pprof.profile.comment`](/docs/registry/attributes/pprof.md) | string[] | Free-form text associated with the profile. This field should not be used to store any machine-readable information, it is only for human-friendly content. | `["hello world", "bazinga"]` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
aalexand, line 84: Do we want to explicitly mention that in terms of `AnyValue` the value should be an array of strings?

aalexand, line 84: As a thought: technically it's indeed from pprof, but it also seems generally applicable. We might consider treating it as a standard attribute, not just a pprof compatibility one. But I'm fine with the change as is for now and we can change this later if needed.

aalexand:
Thank you @florianl - left a couple of comments.

docs/general/profiles.md
florianl: As the `Type` of the new attribute `pprof.profile.comment` is set to `string[]` I think we don't need to care about `AnyValue` here. 
Looking at `browser.brands` as an example, it will always be handled as array of strings: https://github.com/open-telemetry/opentelemetry-go/blob/84e3f3ac8b25204f3a0f77a805437a5e08573b35/semconv/v1.37.0/attribute_group.go#L1883-L1888

aalexand: Ah, great - thanks!

christos68k: I checked to see if there's an existing OTel attribute that can serve this purpose but I don't see anything relevant. If we want to clearly distinguish provenance for this attribute it might still make sense to have a pprof-specific attribute and a non-pprof profile comment attribute: that way provenance is explicit and one can populate both with different values (e.g. the non-pprof profile comment can include additional text that wasn't in the original pprof data, maybe related to conversion specifics)

lmolkova (approved):

model/pprof/registry.yaml
```diff
@@ -35,3 +35,12 @@ groups:
           profile changes.
         stability: development
         type: boolean
+      - id: pprof.profile.comment
+        brief: >
+          Free-form text associated with the profile. This field should not be
+          used to store any machine-readable information, it is only for
+          human-friendly content.
+        type: string[]
```
lmolkova, line 43: any reason for it to be an array? Based on the description it looks like a simple string.

aalexand: It's a repeated string field in pprof, so we are capturing those semantics as array. It would be good to clarify the wording a bit maybe.

florianl: @aalexand do you have suggestions for the clarification?

lmolkova: Feel free to resolve and follow up later (if so, please create an issue to follow up).

PR #2867, opened 2025-10-04
[chore] Reorder columns #2824

### LINKED ISSUE #2824: Reorder columns in markdown tables: prioritize stability and req levels
Discussed during SemConv call on 9/29:

Currently the order of columns is Attribute Name -> Type -> Brief -> Examples -> Req Level -> Stability. E.g.

|---|---|---|---|---|---|
| [`http.request.method`](/docs/registry/attributes/http.md) | string | HTTP request method. [1] | `GET`; `POST`; `HEAD` | `Required` | Stable |

The end users, instrumentation authors and backends probably care more about req levels and stability more than other things.

What if we did something like

| Attribute  |  [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Type | Description  | Examples  | 
|---|---|---|---|---|
| [`http.request.method`](/docs/registry/attributes/http.md)<br>Stable Since 1.23.1 <br>Sampling_relevant| `Required` | string | HTTP request method. [1] | `GET`; `POST`; `HEAD` | 

(we can also have multiple badges or multi-part badges)

thompson-tomo:
Fixes #2824

## Changes

The attribute tables have been re-organised in the following manner:

- Attribute heading added above attribute tables to provide guidance as to what the table is
- Attribute column heading renamed to Key as it relates to the key of the attribute which is a row
- Examples column heading renamed to Examples Values as it relates to the value of the attribute rather than an attribute
- Type column heading renamed to Value Type as it relates to the type expected of the value rather than an attribute
- Description column heading renamed to Summary as it should be a brief summary of what it is.
- Requirement and stability moved to be after key to give increased focus


trask:
Examples of new layout, anywhere there's an attribute table.

Stability and Requirement level have been moved closed to the beginning, and a couple of column renames:

image

kamphaus:
I think we should have either Summary or Description for both:
image

thompson-tomo:
@kamphaus i agree that in the end they should match. I was planning to update the signal property headings/labels is in a follow up pr to not change too much here. That pr  would also standardise title casing & : for all labels. I will raise that later this week and link the 2 pr's. 

Unless @trask things it would be better to have 1 bigger pr to cover both.

trask:
I'd suggest leaving it as Description in this PR, and you can follow up later with a PR to propose changing them both at the same time

christos68k (approved):

thompson-tomo:
FYI @mmanciop this PR implements the reordering of the columns as you suggested.

mmanciop (approved):

thompson-tomo:
@joaopgrassi to avoid the continuous need to resolve conflicts, can you let me know once there is enough approvals and I will do it.

joaopgrassi:
Let's talk about it on the meeting today and then if no objections we can merge 👍

PR #2886, opened 2025-10-07
Add new AWS Attributes for Bedrock AgentCore services #1794

liustve:
## Changes

Add new AWS Attributes for all AWS Bedrock AgentCore services including Runtime, Memory, Identity, Code Interpreter, Gateway, and Browser.


model/aws/registry.yaml
```diff
@@ -581,3 +581,81 @@ groups:
...
+    display_name: Amazon Bedrock AgentCore Code Interpreter Attributes
+    brief: >
+        This document defines attributes for AWS Bedrock AgentCore Code Interpreter.
+    attributes:
+      - id: aws.bedrock.agentcore.code_interpreter.id
+        type: string
+        stability: development
+        brief: >
+          The unique identifier of the created code interpreter.
+        examples: ["myInterpreter-a1b2c3d4e5"]
```
thompson-tomo: Should this not be gen_ai.code_interpreter.id as code interpreter is not only an aws bedrock concept?

model/aws/registry.yaml
```diff
@@ -581,3 +581,81 @@ groups:
...
+    display_name: Amazon Bedrock AgentCore Browser Attributes
+    brief: >
+        This document defines attributes for AWS Bedrock AgentCore Browser.
+    attributes:
+      - id: aws.bedrock.agentcore.browser.id
+        type: string
+        stability: development
+        brief: >
+          The unique identifier of the created browser.
+        examples: ["myBrowser-a1b2c3d4e5"]
```
thompson-tomo: Should this not be gen_ai.browser.id as browser is not only an aws bedrock concept?

model/aws/registry.yaml
```diff
@@ -581,3 +581,81 @@ groups:
...
+    display_name: Amazon Bedrock AgentCore Gateway Attributes
+    brief: >
+        This document defines attributes for AWS Bedrock AgentCore Gateway.
+    attributes:
+      - id: aws.bedrock.agentcore.gateway.id
+        type: string
+        stability: development
+        brief: >
+          The unique identifier of the created gateway.
+        examples: ["mygateway-a1b2c3d4e5"]
```
thompson-tomo: Should this not be gen_ai.gateway.id as gateway is not only an aws bedrock concept?

model/aws/registry.yaml
```diff
@@ -581,3 +581,81 @@ groups:
...
+    display_name: Amazon Bedrock AgentCore Runtime Attributes
+    brief: >
+        This document defines attributes for AWS Bedrock AgentCore Runtime.
+    attributes:
+      - id: aws.bedrock.agentcore.runtime.id
+        type: string
+        stability: development
+        brief: >
+          The unique identifier of the AgentCore Runtime.
+        examples: ["myRuntime-a1b2c3d4e5"]
```
thompson-tomo: Should this not be gen_ai.runtime.id as runtime is not only an aws bedrock concept?

model/aws/registry.yaml
```diff
@@ -581,3 +581,81 @@ groups:
...
+    display_name: Amazon Bedrock AgentCore Memory Attributes
+    brief: >
+        This document defines attributes for AWS Bedrock AgentCore Memory.
+    attributes:
+      - id: aws.bedrock.agentcore.memory.id
+        type: string
+        stability: development
+        brief: >
+          The unique identifier of the memory.
+        examples: ["myMemory-a1b2c3d4e5"]
```
thompson-tomo: Should this not be gen_ai.memory.id as memory is not only an aws bedrock concept?

model/aws/registry.yaml
```diff
@@ -581,3 +581,81 @@ groups:
...
+    display_name: Amazon Bedrock AgentCore Identity Attributes
+    brief: >
+        This document defines attributes for AWS Bedrock AgentCore Identity.
+    attributes:
+      - id: aws.bedrock.agentcore.identity.credential_provider.arn
+        type: string
+        stability: development
+        brief: >
+          The Amazon Resource Name (ARN) of the OAuth2 credential provider.
+        examples: ["arn:aws:acps:us-east-1:123456789012:token-vault/my-vault/oauth2credentialprovider/my-provider"]
```
thompson-tomo: Should we not remove bedrock.agentcore from the id so it can be reused by other aws conventions?

thompson-tomo (commented):
It would be helpful to understand where these attributes would be used especially the identity attributes.

model/aws/registry.yaml
liustve: That is reasonable, I changed the name to `aws.auth.credential_provider.arn` since `identity` is also an AWS service, did not want to make it confusing: 
https://aws.amazon.com/identity/

model/aws/registry.yaml
liustve: Sure, that makes sense to me. Will make changes.

model/aws/registry.yaml
liustve: Sure, that makes sense to me. Will make changes.

model/aws/registry.yaml
liustve: Sure, that makes sense to me. Will make changes.

model/aws/registry.yaml
liustve: Sure, that makes sense to me. Will make changes.

model/aws/registry.yaml
liustve: Sure, that makes sense to me. Will make changes.

liustve:
> It would be helpful to understand where these attributes would be used especially the identity attributes.

These attributes would be set by calls to AWS Bedrock AgentCore in botocore instrumentation:

AgentCore component ARNs (code_interpreter, browser, memory, runtime, gateway, runtime_endpoint) would be set when agents interact with these services.

Example we could capture runtime_endpoint when calls are made to: https://docs.aws.amazon.com/bedrock-agentcore-control/latest/APIReference/API_CreateAgentRuntimeEndpoint.html

model/aws/registry.yaml
thompson-tomo: Actually the convention is to include the service if you take a look at https://opentelemetry.io/docs/specs/semconv/registry/attributes/aws/

thompson-tomo:
@liustve thanks for the changes, it would be good to see how the new attributes could be used in the current bedrock docs https://opentelemetry.io/docs/specs/semconv/gen-ai/aws-bedrock/ for instance is there new span definition for browser spans or are they added to the existing.

model/aws/registry.yaml
liustve: I think my original idea and interpretation was that `credential_provider` isn’t limited to just the identity service, other AWS services, like Bedrock AgentCore Identity or IAM, could also make use of this attribute. My intention was to keep it generic under `auth` so it could apply more broadly across services, since I think calling it Identity would imply that it should only be used for that service. Would it make sense to move this under more general AWS attributes instead?

liustve:
> @liustve thanks for the changes, it would be good to see how the new attributes could be used in the current bedrock docs https://opentelemetry.io/docs/specs/semconv/gen-ai/aws-bedrock/ for instance is there new span definition for browser spans or are they added to the existing.

Sure that makes sense, I added new documentation for `span.gen_ai.infrastructure` attributes and updated `aws-bedrock.md` with AgentCore attributes to extend them.

liustve:
Not stale.

liustve:
Not stale.

trask:
Closing as there does not appear to be engagement from the necessary folks in order to move this forward at this time. It may help to discuss this in the Semantic Convention GenAI SIG:

| SIG Name | Meeting Time | Meeting Notes | Slack Channel | Calendar |
|----------|--------------|---------------|---------------|----------|
| [Semantic Conventions and Instrumentation: GenAI](https://github.com/open-telemetry/community?tab=readme-ov-file#sig-genai-instrumentation) | Every Tuesday 9:00 PT and every other Tuesday 10:00 UTC+8 | [Google Doc](https://docs.google.com/document/d/1P6am7kDTcHzO-fs_LoUW0xQf_qAqent0QWg_JW1hjRc/edit?tab=t.0) | [#otel-genai-instrumentation](https://cloud-native.slack.com/archives/C07SBALBBBN) | [calendar-semconv](https://groups.google.com/a/opentelemetry.io/g/calendar-semconv) |

PR #2890, opened 2025-10-08
Replace system.paging.type and process.paging.fault_type with system.paging.fault.type

ChrsMark:
Part of https://github.com/open-telemetry/semantic-conventions/issues/1803

## Changes

This PR replaces `system.paging.type` and `process.paging.fault_type` with `system.paging.fault.type` based on the what was suggested at https://github.com/open-telemetry/semantic-conventions/issues/1803 and https://github.com/open-telemetry/semantic-conventions/issues/1501 (`_type` should be `.type`).

### DIFF
--- a/.chloggen/fix_paging_fault_type.yaml
+++ b/.chloggen/fix_paging_fault_type.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: system
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Replace `system.paging.type` and `process.paging.fault_type` with `system.paging.fault.type`
+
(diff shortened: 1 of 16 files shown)


model/system/registry.yaml
joaopgrassi: Is the brief here still accurate?

braydonk: The meaning of this attribute isn't changing, but the brief as it exists here isn't great. Right now it actually adds to confusion. It would be nice to be able to just leave it blank since the meaning is as obvious as it can be from the attribute name, but if we need one then `The paging fault type` would be what I'd change it to.

ChrsMark: Good catch, I changed that at [adac8d7](https://github.com/open-telemetry/semantic-conventions/pull/2890/commits/adac8d75ca8bd17201e34eac50a376b9cccb2109)

.chloggen/fix_paging_fault_type.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: system
```
joaopgrassi:
```suggestion
component: system, process
```

braydonk (approved):

jinja2 (approved):

PR #2911, opened 2025-10-14
[chore] Use generic/ficticious example for conditionaly required attributes

### LINKED ISSUE #1508: Revise the description of `Conditionally Required`
The new database conventions define `server.address` as `Recommended`. The following description of `Conditionally Required` should use a different example.

https://github.com/open-telemetry/semantic-conventions/blob/35664be5d835b2f2e8a2e414dea5cef3b04d045a/docs/general/attribute-requirement-level.md?plain=1#L85-L90

joaopgrassi:
Fixes #1508

## Changes

I tried finding an example, but wasn't very successful + I think it's better to keep it ficticious to avoid it "breaking" again. I think it still serves the purpose that it initially had. 

### DIFF
--- a/docs/general/attribute-requirement-level.md
+++ b/docs/general/attribute-requirement-level.md
@@ -82,13 +82,11 @@ the `Opt-In` requirement level on the attribute.
 
 <!-- TODO(jsuereth) - make examples not break on changes to semconv -->
 
-For example, `example.attribute1` is `Conditionally Required` by a convention
-when available.
-If only `example.attribute2` is available instead, which may require an expensive
-operation (e.g., a lookup or computation) to retrieve, instrumentation can
-use `example.attribute2` to populate `example.attribute1`. However, this
-should only be done if the user explicitly enables the instrumentation to
-perform such operations, considering the potential performance impact.
+For example, `server.address` is `Conditionally Required` by a convention. When
+server IP address is available instead, instrumentation can do a DNS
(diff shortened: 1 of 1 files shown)


docs/general/attribute-requirement-level.md
```diff
@@ -82,12 +82,13 @@ the `Opt-In` requirement level on the attribute.
...
-lookup, cache and populate `server.address`, but only if the user explicitly
-enables the instrumentation to do so, considering the performance issues that
-DNS lookups introduce.
+For example, `example.attribute1` is `Conditionally Required` by a convention
+when available.
+If only `example.attribute2` is available instead, which may require an expensive
+operation (e.g., a lookup or computation) to retrieve, instrumentation can
+use `example.attribute2` to populate `example.attribute1`. However, this
+should only be done if the user explicitly enables the instrumentation to
+perform such operations, considering the potential performance impact.
```
lmolkova: I think the point here was that IP address is available, attribute name is somewhat irrelevant. 

Wonder if we can use a fictitious convention instead to keep the context, e.g.

```suggestion
For example, `server.address` is `Conditionally Required` by a convention. When
server IP address is available instead, instrumentation can do a DNS
lookup, cache and populate `server.address`, but only if the user explicitly
enables the instrumentation to do so, considering the performance issues that
DNS lookups introduce.
```

WDYT?

lmolkova (approved):

PR #2779, opened 2025-09-16
Add metrics for openshift's clusterquota

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/2078.
Part of https://github.com/open-telemetry/semantic-conventions/issues/1032.

## Changes

This PR defines openshift's clusterquota metrics that are already in use by the [Opentelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.111.0/receiver/k8sclusterreceiver/documentation.md).

The metrics added are the following similarly to the respective [generic ResourceQuota PR](https://github.com/open-telemetry/semantic-conventions/pull/2113):

```
openshift.clusterquota.cpu.limit.hard
openshift.clusterquota.cpu.limit.used
openshift.clusterquota.cpu.request.hard
openshift.clusterquota.cpu.request.used
openshift.clusterquota.memory.limit.hard
openshift.clusterquota.memory.limit.used
openshift.clusterquota.memory.request.hard
openshift.clusterquota.memory.request.used
openshift.clusterquota.hugepage_count.request.hard
openshift.clusterquota.hugepage_count.request.used
openshift.clusterquota.storage.request.hard
openshift.clusterquota.storage.request.used
openshift.clusterquota.persistentvolumeclaim_count.hard
openshift.clusterquota.persistentvolumeclaim_count.used
openshift.clusterquota.ephemeral_storage.request.hard
openshift.clusterquota.ephemeral_storage.request.used
openshift.clusterquota.ephemeral_storage.limit.hard
openshift.clusterquota.ephemeral_storage.limit.used
openshift.clusterquota.object_count.hard
openshift.clusterquota.object_count.used
```

## Merge requirement checklist

* [ ] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [ ] Links to the prototypes or existing instrumentations (when adding or changing conventions)

PR #2917, opened 2025-10-15
[editorial] Rename resource/k8s/k8s.md to resource/k8s/README.md

### LINKED ISSUE #2915: [editorial] Rename of `docs/resource/k8s.md` might be wrong. It also needs an alias to avoid 404s
_Originally posted by @chalin in [#2779](https://github.com/open-telemetry/semantic-conventions/pull/2779/files#r2430096146)_

> Hmm, this is [breaking links on OTel.io](https://github.com/open-telemetry/opentelemetry.io/actions/runs/18505057657/job/52731806415?pr=7863). Should this rename:
> 
> - have instead been `docs/resource/k8s.md`‎ to `‎docs/resource/k8s/README.md‎`
> - rather than to `docs/resource/k8s/k8s.md`?
> 
> Also, as usual, when a file is moved, there should be an alias or redirect rule added.

chalin:
- Fixes #2915
- Followup to #2779
  - Renames `resource/k8s/k8s.md` to `resource/k8s/README.md`
  - Fixes code owner entry
- Because the rename is essentially to the index page of a section with the same path, we don't need a redirect rule


joaopgrassi:
What about we add a README.md (`docs/resource/k8s/README.md`) ? Then we can just use that as the entry point for both the k8s.md and openshift.md. With something like:

---

# Container orchestration

This document defines semantic conventions for resource in container orchestration systems.

* [K8s](k8s.md): Semantic Conventions for K8s.
* [OpenShift](openshift.md): Semantic Conventions for OpenShift.

chalin:
> What about we add a README.md (`docs/resource/k8s/README.md`) ? Then we can just use that as the entry point for both the k8s.md and openshift.md. With something like:

Sure, that works, if it is what you prefer. It's just that what you propose feels a bit less natural. E.g., you write "conventions for resource in K8s" twice: once for the landing page and then again for a subpage. WDYT?

joaopgrassi:
Yeah maybe we have the landing page with "Container Orchestration" instead :). Fixed the example above.

chalin:
> Yeah maybe we have the landing page with "Container Orchestration" instead :). Fixed the example above.

@joaopgrassi - hmm, but then the folder name should be changed too? So rename the `k8s` folder to, say, `container`? Otherwise, I can't create a redirect for the moved `k8s` page. If I try, it creates a loop for the index page.

If the section really is about k8s resources, then I think the change proposed in this PR is probably best. But I defer to you.

joaopgrassi:
Alright, let's keep your suggestion then, and we can revisit later if necessary.

joaopgrassi (approved):

trisch-me:
@joaopgrassi I like your idea of having page with links to k8s and openshift but let’s first merge as is to unblock

Issue #560, opened 2023-11-24
Convention for cancelled spans


charleskorn:
It would be great if there was a language- and protocol-agnostic convention for marking a span as representing a cancelled operation.

At present, different languages and protocols treat cancelled operations in different ways, which makes it difficult to do things like filter a trace to show only non-cancelled operations, or visually distinguish between successful, failed and cancelled operations.

For example, a span representing a gRPC call would have the `rpc.grpc.status_code` attribute set to 1 (for the gRPC `CANCELLED` [status code](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md)), while a span representing a Golang operation might be recorded with an event with a `context canceled` message.

Instead, it'd be great if there was a convention for an attribute that marked a span as cancelled, such as `cancelled`.

trask:
oh no, we'll need to make a decision on canceled vs cancelled 😅

in seriousness, a "canceled" (or "cancelled") attribute would be very welcome in opentelemetry-java-instrumentation repo where we currently have (at least):

* `guava.canceled`
* `rxjava.canceled`
* `jaxrs.canceled`
* `reactor.canceled`
* `grpc.canceled`

mateuszrzeszutek:
Didn't we want to use the `error.type` attribute for that? At least in the HTTP scenarios, e.g. `error.type = cancelled`

pyohannes:
> Didn't we want to use the error.type attribute for that? At least in the HTTP scenarios, e.g. error.type = cancelled

The problem with using `error.type` is, that a cancelled operation doesn't necessarily imply an error. However, the definition of [`error.type`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/error.md) states this:

> If the operation has completed successfully, instrumentations SHOULD NOT set `error.type`.

lmolkova:
I'd be in favor of defining something in common. 

I think we can polish `error.type` description to accommodate it - "If the operation has completed successfully, instrumentations SHOULD NOT set error.type." - cancelled operation did not complete successfully and it should be totally legit to use `error.type` for it. In the same was as we set `error.type` to `"404"` on the client side even when 404 does not represent an error.

Some problem with the `cancel(l)ed` convention:
- E.g. on Java the async operations (like reactor) can be cancelled, but sync versions of them throw something like `TimeoutException` or `InterruptedException` - I'm not sure how far we can go unifying and if by replacing exception type with `cancelled` is helpful.
[...]

PR #2920, opened 2025-10-15
Deprecate per-rpc status code attributes in favor of `rpc.response.status_code`

### LINKED ISSUE #1504: Rename `rpc.grpc.status_code` to `grpc.response.status_code`
### Area(s)

area:rpc

The current attribute `rpc.grpc.status_code` didn't follow the convention of having a `response` inside the name.

List of attributes that have `response` inside:

- [`http.response.status_code`](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/)
- [`db.response.status_code`](https://opentelemetry.io/docs/specs/semconv/attributes-registry/db/)

Rename `rpc.grpc.status_code` to `rpc.grpc.response.status_code` to keep consistent.

_No response_

### Additional context

_No response_

### LINKED ISSUE #2063: JSON RPC: deprecate duplicate attributes

lmolkova:
Fixes #1504, fixes #2063

Some of the RPC frameworks generate status code on client side when response cannot be received. This includes :

- gRPC:
   https://grpc.io/docs/what-is-grpc/core-concepts/#rpc-termination
   > In gRPC, both the client and server make independent and local determinations of the success of the call, and their conclusions may not match. This means that, for example, you could have an RPC that finishes successfully on the server side (“I have sent all my responses!”) but fails on the client side (“The responses arrived after my deadline!”)

  https://grpc.io/docs/guides/status-codes/
  > When an error situation occurs, the gRPC library may produce a corresponding status. The library may do this either on the client- or the server-side. 
  
- Connect RPC. Similar to gRPC - client returns error/throws exception when any error happens and may generate deadline_exceeded / cancelled / unknown / unavailable errors 

   https://github.com/connectrpc/connect-go/blob/5398e9930ae8e7c06b500f4a0b8ed46c82192d5e/error.go#L274-L278

  Even invalid URL is reported with code `unavailable` - https://github.com/connectrpc/connect-go/blob/5398e9930ae8e7c06b500f4a0b8ed46c82192d5e/client.go#L391
- Apache Dubbo. Similar - pretty much everything (network, timeouts, serialization errors) is wrapped into `RpcException` with code defaulting to `unknown` https://github.com/apache/dubbo/blob/0d8c15adea9061aac1296ce5077384ad8e7b947b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java#L280-L335

Even though status code could be generated by the client, we're using `rpc.response.status_code`  to stay consistent with existing conventions. Given we're focusing on logical layer, response essentially means response on the API level and includes response codes assigned on the client.

This PR 
- introduces common `rpc.response.status_code` attribute 
- deprecates `rpc.connect_rpc.error_code`, `rpc.grpc.status_code`, `rpc.jsonrpc.error_code`
- adds recommendation on reporting span status based on `rpc.response.status_code` value
- follow up: reporting error message https://github.com/open-telemetry/semantic-conventions/issues/2967 (sensitivity concerns)


trask:
> Some of the RPC frameworks generate status code on client side when response cannot be received. This includes gRPC, Connect RPC and Apache Dubbo. So not including response in the status_code

Can you include a couple examples of this in the PR description? Thanks

lmolkova:
> Can you include a couple examples of this in the PR description? Thanks

@trask sure, updated description with links

docs/rpc/rpc-spans.md
```diff
@@ -101,7 +101,10 @@ This span represents an outgoing Remote Procedure Call (RPC).
 
 **Span kind** MUST be `CLIENT`.
 
-**Span status** SHOULD follow the [Recording Errors](/docs/general/recording-errors.md) document.
+**Span status** Refer to the [Recording Errors](/docs/general/recording-errors.md)
+document for details on how to record span status. Semantic conventions for
+individual systems SHOULD specify which values of `rpc.status_code`
+classify as errors.
```
trask:
```suggestion
are classified as errors.
```

model/rpc/spans.yaml
```diff
@@ -39,12 +49,20 @@ groups:
       **Span name:** refer to the [Span Name](/docs/rpc/rpc-spans.md#span-name) section.
 
       **Span kind** MUST be `CLIENT`.
+
+      **Span status** Refer to the [Recording Errors](/docs/general/recording-errors.md)
+      document for details on how to record span status. When error response is returned and
+      it contains [`message`](https://connectrpc.com/docs/web/errors/#error-messages),
+      the raw error message SHOULD be recorded as the span status description.
```
trask: (related https://github.com/open-telemetry/opentelemetry-specification/issues/3496)

model/rpc/spans.yaml
```diff
@@ -129,6 +184,13 @@ groups:
       **Span name:** refer to the [Span Name](/docs/rpc/rpc-spans.md#span-name) section.
 
       **Span kind** MUST be `CLIENT`.
+
+      **Span status** Refer to the [Recording Errors](/docs/general/recording-errors.md)
+      document for details on how to record span status. Responses that include
+      [`error` object](https://www.jsonrpc.org/specification#error_object)
+      are considered errors. When error object is returned  with
```
trask:
```suggestion
      are considered errors. When error object is returned with
```

model/rpc/spans.yaml
```diff
@@ -157,6 +222,13 @@ groups:
       **Span name:** refer to the [Span Name](/docs/rpc/rpc-spans.md#span-name) section.
 
       **Span kind** MUST be `SERVER`.
+
+      **Span status** Refer to the [Recording Errors](/docs/general/recording-errors.md)
+      document for details on how to record span status. Responses that include
+      [`error` object](https://www.jsonrpc.org/specification#error_object)
+      are considered errors. When error object is returned  with
```
trask:
```suggestion
      are considered errors. When error object is returned with
```

trask (approved):

thompson-tomo:
For me I see grpc as warranting it's own attribute to increase interoperability with the grpc metrics & also allow the rpc.status_code to be set by the framework ie Dubbo. Deprecating the other 2 makes sense.

lmolkova:
> For me I see grpc as warranting it's own attribute to increase interoperability with the grpc metrics

could you elaborate? How would having `rpc.status_code` and `grpc.status_code` would help with grpc metrics interoperability?

> also allow the rpc.status_code to be set by the framework ie Dubbo. Deprecating the other 2 makes sense.

Looking into Dubbo, it [juggles with 3 different sets of error codes](https://github.com/apache/dubbo/blob/3.3/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/TriRpcStatus.java) (Dubbo, Triple, and HTTP). If it makes sense to report both protocol (grpc / HTTP) and framework (dubbo), I'd assume we'd have dubbo-specific attribute for it instead of making it gRPC responsibility.

model/rpc/spans.yaml
lmolkova: I removed the part about error message and created https://github.com/open-telemetry/semantic-conventions/issues/2967 which is a blocker for RPC stability

thompson-tomo:
The grpc.status attribute is an established attribute used by the grpc metrics hence it would decrease the gap between the specs.

Having network request status seperate provides clear indication that the network has been used and it would enable the framework to have a different status to the network. That is important when the framework supports/uses client side caching.

In relation to the examples in the description should we not be using error/exception to capture the details of the status and only setting the  rpc.status_code when it is being returned from the server.

Also note grpc doesn't define cancelled as an error but the note on the new attribute indicates everything but ok is an error.

lmolkova:
> The grpc.status attribute is an established attribute used by the grpc metrics hence it would decrease the gap between the specs.

Targeting compatibility with current [gRPC metrics](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md) would involve way more than `grpc.status` attribute. Having one attribute with the same name would not help with compatibility if metric names and everything else is different. 

> Having network request status seperate provides clear indication that the network has been used and it would enable the framework to have a different status to the network. That is important when the framework supports/uses client side caching.

Framework (when there is a framework and it needs separate status code) should then provide that extra status code attribute. Why have discrepancy on the protocol level?

> In relation to the examples in the description should we not be using error/exception to capture the details of the status and only setting the rpc.status_code when it is being returned from the server.

could you please add a suggestions on the specific lines so we can have focused discussion and I can understand the context of your comments?

> Also note grpc doesn't define cancelled as an error but the note on the new attribute indicates everything but ok is an error.

Same here, in the future, please leave comment on specific lines so it's easier to understand what you mean. 
Substantially: I broke down the table we already have  here https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/grpc.md#grpc-status into notes on `rpc.status_code` without changing the behavior. I agree cancelled status being error is the issue and we're going to address  it in https://github.com/open-telemetry/semantic-conventions/issues/560 which is blocking for RPC stability.

lmolkova:
SIG discussion:
- generic `error.type` mentions gRPC status code - let's generalize it
- let's  refine `error.type` on RPC conventions to explain what to do wrt status code

.chloggen/rpc-status-code.yaml
```diff
@@ -0,0 +1,6 @@
+change_type: breaking
+component: rpc
+note: >
+  Deprecate `rpc.grpc.status_code`, `rpc.connect_rpc.error_code` and `rpc.jsonrpc.error_code` attributes
+  in favor of the more general `rpc.status_code` attribute.
```
lmolkova: SGI discussion: let's add response - we instrumnet logical and response is something client returns.

trask:
Can you update the PR description based on the updated attribute name?

model/rpc/spans.yaml
arminru: Might be good to mention this in the PR description.

model/rpc/spans.yaml
```diff
@@ -129,6 +176,11 @@ groups:
       **Span name:** refer to the [Span Name](/docs/rpc/rpc-spans.md#span-name) section.
 
       **Span kind** MUST be `CLIENT`.
+
+      **Span status** Refer to the [Recording Errors](/docs/general/recording-errors.md)
+      document for details on how to record span status. Responses that include
+      [`error` object](https://www.jsonrpc.org/specification#error_object)
```
arminru:
```suggestion
      an [`error` object](https://www.jsonrpc.org/specification#error_object)
```

model/rpc/spans.yaml
```diff
@@ -157,6 +212,11 @@ groups:
       **Span name:** refer to the [Span Name](/docs/rpc/rpc-spans.md#span-name) section.
 
       **Span kind** MUST be `SERVER`.
+
+      **Span status** Refer to the [Recording Errors](/docs/general/recording-errors.md)
+      document for details on how to record span status. Responses that include
+      [`error` object](https://www.jsonrpc.org/specification#error_object)
```
arminru:
```suggestion
      an [`error` object](https://www.jsonrpc.org/specification#error_object)
```

model/rpc/spans.yaml
```diff
@@ -62,10 +74,23 @@ groups:
...
+      **Span status** Refer to the [Recording Errors](/docs/general/recording-errors.md)
+      document for details on how to record span status.
     attributes:
-      - ref: rpc.connect_rpc.error_code
+      - ref: rpc.response.status_code
         requirement_level:
           conditionally_required: if available.
+        brief: The [error code](https://connectrpc.com//docs/protocol/#error-codes) of the Connect response.
+        note: |
+          The following error codes SHOULD be considered errors:
```
arminru, line 86: Have we considered RESOURCE_EXHAUSTED, FAILED_PRECONDITION, ABORTED, and OUT_OF_RANGE as potential server-side errors for both Connect and gRPC?

model/rpc/spans.yaml
lmolkova: done

model/rpc/spans.yaml
lmolkova: This follows current guidance: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/grpc.md#grpc-status

All of those are errors on the client side, but not errors on the server (e.g. if client sent a invalid request that resulted in OUT_OF_RANGE, server correctly processed it - there is no error). This matches what we do on HTTP - client errors like 400 are not server errors. 

We could definitely update it based on the feedback. 
The current logic is implemented at least in some of the instrumentations: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/671b2e96d8f41a89748cbe87fa8aef471cd6ddaa/instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/GrpcSpanStatusExtractor.java#L56

I briefly searched for related issues in java instrumentation repo and don't see anything mentioning the attribute name or any of the values you mentioned.

So I'm inclined to leave things as they are, unless there is a strong signal from users to change status quo

arminru: Alright! Thanks for the research in the Java instrumentation repo, that's good backing data for keeping it as-is until someone approaches us with it on their own.

Issue #1869, opened 2025-02-04
Should search engines follow database semantic conventions?


lmolkova:
Services like [Azure Search](https://learn.microsoft.com/en-us/azure/search/search-what-is-azure-search), [AWS Kendra](https://docs.aws.amazon.com/kendra/latest/dg/what-is-kendra.html), or [Google Cloud Search](https://developers.google.com/cloud-search/docs/guides) provide an interface similar to databases (that allows to query data), but they are not usually the ones that store this data.

For example, Azure search indexes data from external storage. It can also act as a proxy when writing data.

Arguably, it looks like a duck, quacks like a duck, but it's not a database. It's a search engine. 

1. Option 1:  pretend that search engines are databases. Populate `db.system.name` and other DB attributes on them
   - Pros: common DB conventions would apply reasonably well on the client side
   - Cons: 
     - they won't apply on the server-side
     - some concepts are much more applicable to search engines than databases in general (ranking scores, big free-form documents, OCR), the terminology is different. Evolving search inside `db` namespace would be limiting to search and not beneficial to `db`
2. Option 2: have separate conventions for search engines
    - Pros: we'll be able to record semantics more precisely and in more idiomatic way on the client and server side
    - Cons: 
      - more conventions and some confusion between database (data source) and search engine being separate
      - it's not clear what elasticsearch or opensearch are (they ca both - data store AND search engine with external store)

----

Proposal (Option 2):
1. Define what DB conventions cover: data stores, not search engines
2. Search engines should have separate conventions that are not in `db` namespace and are focused on search capabilities. 
3. Each search engine system can decide what it is (or if it's both) and can leverage attributes from both namespaces. I.e. it may end up conforming to more than one 'type' of conventions

lmolkova:
curious what elasticsearch and opensearch folks think about having search vs DB conventions. @gregkalapos @Aneurysm9

PR #2924, opened 2025-10-16
Retrieval Span Support

### LINKED ISSUE #2907: Add retrieval /search span support to Semantic Conventions
### Area(s)

area:gen-ai

### Propose new conventions

We should add the retrieval type to the existing list of genai client spans so they can be utilized in future genai instrumentations: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md#inference

### Tip

JWinermaSplunk:
Fixes #2907

## Changes

Add retrieval span support to db spans

> [!IMPORTANT]

Implementations:

- https://github.com/traceloop/openllmetry/blob/main/packages/opentelemetry-instrumentation-llamaindex/opentelemetry/instrumentation/llamaindex/base_retriever_instrumentor.py

### DIFF
--- a/.chloggen/retrieval-span-support.yaml
+++ b/.chloggen/retrieval-span-support.yaml
@@ -4,7 +4,7 @@
 # your pull request title with [chore] or use the "Skip Changelog" label.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: 'new_component'
+change_type: 'enhancement'
 
 # The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
 component: gen-ai


lmolkova (changes requested):
Retrieval is in many cases based on the database retrieval. e.g. postgreds or MongoDB instrumentation has no knowledge that it's used in the context of GenAI application.

So retrieval in a general case is just a DB call with semantics described in https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/README.md 

There is a question of whether search engines are databases or should be covered by a separate or additional set of the conventions - https://github.com/open-telemetry/semantic-conventions/issues/1869 - this is where OpenAI retrieval API should probably belong.

lmolkova:
Adding more context from GenAI SIG call: 

- langchain, llamaindex, haystack offer retriever abstraction. 
- the question is: should corresponding implementations be instrumented and if so, which conventions they should follow

My take:

Reading through lanchain, llamaindex, haystack docs, retriever is in most cases is an thin layer on top of a database or a search client which may be instrumented using [database conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/README.md) and/or hypothetical [search conventions](https://github.com/open-telemetry/semantic-conventions/issues/1869). 

Retrievers could be more complicated and combine multiple source of data or perform additional logic, in these cases, the spans they emit might be significantly different than underlying DB calls and then multiple layers may be instrumented at the same time. In the case of thin layer abstraction / wrapper, having two spans does not improve observability, but increases costs and noise.

When it comes to abstractions, there is a classic problem of instrumentation layers (ORM vs database spans, lanchain LLM vs underlying model-client spans, etc): both layers could be instrumented and there are pros and cons (DB layer has more low-level info, framework layer represents caller perspective better). The duplication is a common problem. Solutions may include:

- users decide to enable/disable corresponding instrumentation (by installing instr library or by enabling/disabling retrival spans on lanchain / llamaindex / etc instrumentation)
- suppressing internal DB spans in scope of higher level retrieval span (similar to [http suppression](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/34db73ebfd97761ba65523c3f07315ce2974dc78/opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py#L224) or more broadly https://github.com/open-telemetry/oteps/pull/172)

I think path forward for this PR:

- separate retrieval from GenAI domain. Most of the attributes defined here can be generic `db` or `search` attributes. Please follow the discussion in #1869
- lanchain / llamaindex / etc instrumentation would cover retrieval following that `db` or `search` conventions and would provide an option to disable their retrieval instrumentation (e.g. when user prefers to use underlying existing DB client instrumentation)

lmolkova:
Also related https://github.com/open-telemetry/semantic-conventions/pull/1231

JWinermaSplunk:
Hi @lmolkova,

Here are a few examples from trace loop and our instrumentation of proposed retrieval spans, also updated in the google doc linked to the issue 
https://docs.google.com/document/d/1DVE3Ht686nuxww-Z1JNyy2VInn9LeCJQYw8NffbjGtg/edit?tab=t.0.

I believe, we also discussed that we are fine moving retrieval spans from the genai to db spaces, but would the optionality to have or enable/disable genai attributes be possible as well, similar to enabling/disabling retrievals as a whole?  So that retrievals could parent a db + embedding operation.
unnamed
unnamed-2

JWinermaSplunk:
Retrieval Semconv Proposal doc (sample telemetry, attributes), adding from the open issue to the PR for more exposure:
https://docs.google.com/document/d/1DVE3Ht686nuxww-Z1JNyy2VInn9LeCJQYw8NffbjGtg/edit?tab=t.0

model/gen-ai/registry.yaml
```diff
@@ -370,6 +370,10 @@ groups:
               value: "embeddings"
               brief: 'Embeddings operation such as [OpenAI Create embeddings API](https://platform.openai.com/docs/api-reference/embeddings/create)'
               stability: development
+            - id: retrieval
+              value: "retrieval"
+              brief: 'Retrieval operation such as [OpenAI Create retrievals API](https://platform.openai.com/docs/guides/retrieval)'
```
lmolkova: I think we don't want create vector store to have the same type as query/search. Retrieval operation should represent search API and we should give a link to that specific operation like https://platform.openai.com/docs/api-reference/vector-stores/search

model/gen-ai/registry.yaml
```diff
@@ -426,6 +430,57 @@ groups:
...
+        stability: development
+        type: string
+        brief: The type of retrieval operation being performed.
+        examples: ["similarity", "hybrid", "keyword"]
+      - id: gen_ai.retrieval.top_k
+        stability: development
+        type: int
+        brief: The maximum number of results requested.
+        examples: [5, 10, 20]
+      - id: gen_ai.retrieval.documents_retrieved
```
lmolkova: do we really need it? is having this attribute helpful without knowing document contents? And/or could we record a skeleton instead of document contents that would capture score + id/name of the doc, so count would be derivable anyway?

model/gen-ai/registry.yaml
```diff
@@ -426,6 +430,57 @@ groups:
...
+          do the best effort to deserialize it to an array. When recorded on spans, it MAY be recorded
+          as a JSON string if structured format is not supported and SHOULD be recorded in structured form otherwise.
+
+          > [!Warning]
+          > This attribute may contain sensitive information.
+        examples:
+          - |
+              [
+                {
+                  "id": "doc_123",
```
lmolkova, line 449: I think we need to formally define schema with at least two properties: `string` id (or name) and `double` score and any additional properties.

model/gen-ai/spans.yaml
```diff
@@ -217,6 +217,32 @@ groups:
...
+
+      **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.request.model}`.
+    extends: attributes.gen_ai.common.client
+    attributes:
+      - ref: gen_ai.operation.name
+        requirement_level: required
+      - ref: gen_ai.retrieval.type
+        requirement_level: recommended
+      - ref: gen_ai.retrieval.query.text
+        requirement_level: recommended
```
lmolkova: this may contain sensitive information and should be opt-in, we should also have a sensitive data warning in the definition

model/gen-ai/spans.yaml
```diff
@@ -217,6 +217,32 @@ groups:
       - ref: gen_ai.embeddings.dimension.count
         requirement_level: recommended
 
+  - id: span.gen_ai.retrieval.client
+    type: span
+    span_kind: client
+    stability: development
+    brief: >
+      Describes GenAI retrieval span - a request to a Generative AI model or service
```
```suggestion
      Describes GenAI retrieval span - a request to a Generative AI service or framework
```

we wanted it to be applicable to lanchain / llamaindex, right?

model/gen-ai/spans.yaml
```diff
@@ -217,6 +217,32 @@ groups:
...
+      that retrieves relevant information or context from a vector database or search system.
+
+      The `gen_ai.operation.name` SHOULD be `retrieval`.
+
+      **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.request.model}`.
```
lmolkova: model attribute is not present on this span and is not applicable here. I guess the best (low cardinality) span name we can come up with is just `gen_ai.operation.name`

model/gen-ai/spans.yaml
```diff
@@ -217,6 +217,32 @@ groups:
...
+      - ref: gen_ai.retrieval.type
+        requirement_level: recommended
+      - ref: gen_ai.retrieval.query.text
+        requirement_level: recommended
+      - ref: gen_ai.retrieval.top_k
+        requirement_level: recommended
+      - ref: gen_ai.retrieval.documents_retrieved
+        requirement_level: recommended
+      - ref: gen_ai.retrieval.documents
+        requirement_level: opt_in
```
lmolkova, line 269: do we want to ref `gen_ai.provider.name` here (with requirement level `recommended: when applicable`)? We'd set it for OpenAI

model/gen-ai/spans.yaml
```diff
@@ -217,6 +217,32 @@ groups:
...
+
```
lmolkova, line 279: please reference `error.type` similarly to all other gen_ai spans, if operation fails, we need to record the error in generic/aggegatable way.

lmolkova, line 279: please reference `server.address | port` attributes - when call is against a service (like OpenAI), we want to record destination endpoint

model/gen-ai/registry.yaml
```diff
@@ -426,6 +430,57 @@ groups:
         type: int
         brief: The number of dimensions the resulting output embeddings should have.
         examples: [512, 1024]
+      - id: gen_ai.retrieval.type
+        stability: development
+        type: string
+        brief: The type of retrieval operation being performed.
+        examples: ["similarity", "hybrid", "keyword"]
+      - id: gen_ai.retrieval.top_k
```
lmolkova: not a strong option, but wdyt about reusing [gen_ai.request.top_k](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) instead of defining a new attribute? It's possible to override brief/note when you reference the attribute on the retrieval span.

I'd  be against that if we used something like `search` or `retrieval` instead  of `gen_ai.retrieval`.

model/gen-ai/registry.yaml
```diff
@@ -426,6 +430,57 @@ groups:
         type: int
         brief: The number of dimensions the resulting output embeddings should have.
         examples: [512, 1024]
+      - id: gen_ai.retrieval.type
+        stability: development
+        type: string
+        brief: The type of retrieval operation being performed.
```
lmolkova: please link relevant docs - are there frameworks that provide this option? do they define an enum like this?

.chloggen/retrieval-span-support.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'new_component'
```
lmolkova:
```suggestion
change_type: 'enhancement'
```

.chloggen/retrieval-span-support.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: gen-ai
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add retrieval span support to the gen-ai semantic conventions.
```
lmolkova, line 13: I believe there are some prototypes for this work, could you please add links to prototypes to the PR description (there is a check box there for it)

model/gen-ai/spans.yaml
JWinermaSplunk: is `error.type` included via `extends: attributes.gen_ai.common.client`, as I see it in the gen-ai-spans.md? same with `server.address` and `server.port`

JWinermaSplunk: mentioned above

model/gen-ai/registry.yaml
JWinermaSplunk: not specifically an enum, but how do these look?
https://docs.haystack.deepset.ai/docs/retrievers
https://developers.llamaindex.ai/python/framework/module_guides/querying/retriever/retriever_modes/
https://docs.langchain.com/oss/python/integrations/vectorstores

JWinermaSplunk: per the SIG: will remove gen_ai.retrieval.type and potentially add at a later date

JWinermaSplunk:
Commenting to keep PR open

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -15,17 +15,37 @@
...
+     "end_time": "2025-12-17T17:50:51.425667Z",
+     "start_time": "2025-12-17T17:50:50.822844Z"
     }
    },
-   "outputs": [],
    "source": [
     "pip install pydantic~=2.0"
-   ]
+   ],
+   "outputs": [
```
lmolkova: could you please clear outputs ?

model/gen-ai/registry.yaml
```diff
@@ -426,6 +430,50 @@ groups:
...
+        note: |
+          Instrumentations MUST follow [Retrieval documents JSON schema](/docs/gen-ai/gen-ai-retrieval-documents.json).
+          When the attribute is recorded on events, it MUST be recorded in structured
+          form. When recorded on spans, it MAY be recorded as a JSON string if structured
+          format is not supported and SHOULD be recorded in structured form otherwise.
+
+          Each document object SHOULD contain at least the following properties:
+          `id` (string): A unique identifier for the document, `score` (double): The relevance score of the document
+
+          Additional properties MAY be included (e.g., `content`, `metadata`, `title`, `uri`).
```
lmolkova: It should be part of the formal schema, is this necessary as a free form text?
Also,  `content`, `metadata`, `title`, `uri` are not mentioned in JSON schema, so instrumentations by default would not populate them. Do you want to mention `content` explicitly in the schema?

model/gen-ai/spans.yaml
```diff
@@ -217,6 +217,34 @@ groups:
...
+        requirement_level: recommended
+      - ref: gen_ai.retrieval.documents
+        requirement_level: opt_in
+      - ref: gen_ai.provider.name
+        requirement_level:
+          conditionally_required: when applicable
+      - ref: error.type
+        requirement_level:
+          conditionally_required: "if the operation ended in an error"
+
```
lmolkova, line 279: I think we should reference [`gen_ai.data_source.id`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md#gen-ai-data-source-id)

It'd match [vector_store_id](https://platform.openai.com/docs/api-reference/vector-stores/search#vector_stores_search-vector_store_id) for OpenAI search or [index id](https://docs.cloud.google.com/vertex-ai/docs/vector-search/quickstart) for Vertex AI

model/gen-ai/spans.yaml
```diff
@@ -217,6 +217,34 @@ groups:
...
+    type: span
+    span_kind: client
+    stability: development
+    brief: >
+      Describes GenAI retrieval span - a request to a Generative AI service or framework
+      that retrieves relevant information or context from a vector database or search system.
+
+      The `gen_ai.operation.name` SHOULD be `retrieval`.
+
+      **Span name** SHOULD be `{gen_ai.operation.name}`.
```
lmolkova: See my other comment on `gen_ai.data_source.id` - it's much more informative to see `retrieval books_index_prod`  than just `retrieval`

Also I imagine it's somewhat provider and retriever-specific, so the suggestion is

```suggestion
      **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.data_source.id}`.  Semantic conventions for individual GenAI providers and retrievers MAY specify different span name format.
```

model/gen-ai/registry.yaml
JWinermaSplunk: No, it isn't necessary. Will delete, but let's stick to just id and score in the schema.

model/gen-ai/spans.yaml
JWinermaSplunk: Yes, sounds good to me!

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -18,6 +18,10 @@
    "execution_count": null,
    "id": "3c4ec73b",
    "metadata": {
+    "ExecuteTime": {
```
lmolkova: nit: could you please undo this, I think it should happen when you clear outputs

model/gen-ai/spans.yaml
```diff
@@ -217,6 +217,38 @@ groups:
...
+    type: span
+    span_kind: client
+    stability: development
+    brief: >
+      Describes GenAI retrieval span - a request to a Generative AI service or framework
+      that retrieves relevant information or context from a vector database or search system.
+
+      The `gen_ai.operation.name` SHOULD be `retrieval`.
+
+      **Span name** SHOULD be `{gen_ai.operation.name} `{gen_ai.data_source.id}`. Semantic conventions
```
```suggestion
      **Span name** SHOULD be `{gen_ai.operation.name} {gen_ai.data_source.id}`. Semantic conventions
```

model/gen-ai/registry.yaml
```diff
@@ -426,6 +430,45 @@ groups:
...
+          Instrumentations MUST follow [Retrieval documents JSON schema](/docs/gen-ai/gen-ai-retrieval-documents.json).
+          When the attribute is recorded on events, it MUST be recorded in structured
+          form. When recorded on spans, it MAY be recorded as a JSON string if structured
+          format is not supported and SHOULD be recorded in structured form otherwise.
+
+          Each document object SHOULD contain at least the following properties:
+          `id` (string): A unique identifier for the document, `score` (double): The relevance score of the document
+
+          > [!Warning]
+          > This attribute may contain sensitive information.
```
lmolkova: with just id and score, there is no sensitive information, is there?

PR #2957, opened 2025-10-22
Add `gcp.agent_engine` as a value for `cloud.platform`

aabmass:
## Changes

Add new value in `cloud.platform` for [Vertex AI Agent Engine](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/agent-engine/overview). 

### DIFF
--- a/.chloggen/2957_gcp_agent_engine.yaml
+++ b/.chloggen/2957_gcp_agent_engine.yaml
@@ -1,4 +1,4 @@
 change_type: enhancement
-component: entities
+component: cloud
 note: Add `gcp.agent_engine` as a value for `cloud.platform`
 issues: [2957]


aabmass:
Will keep getting auto closed until https://github.com/open-telemetry/semantic-conventions/pull/2958

joaopgrassi:
Overriding automation.

lmolkova:
I think a new flow for this would be to start with GH issue, get it triaged and accepted and then send a PR, is this right, @joaopgrassi ?

model/cloud/registry.yaml
```diff
@@ -245,6 +245,10 @@ groups:
               annotations:
                 code_generation:
                   exclude: true
+            - id: gcp_agent_engine
+              value: 'gcp_agent_engine'
```
aabmass: @jsuereth you mentioned offline this should be `gcp.agent_engine`, or should change them all at once later on?

joaopgrassi:
> I think a new flow for this would be to start with GH issue, get it triaged and accepted and then send a PR, is this right, @joaopgrassi ?

Yes, that's the ideal flow we want to follow. We accept direct PRs only that come from a SIG.

model/cloud/registry.yaml
jsuereth: I believe we want to try to change them now, cc @lmolkova for confirmation

lmolkova: we want consistent naming between attribute/metric names and cloud.platform | gen_ai.system.name - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#system-specific-naming

E.g. `clould.platform = gcp.agent_engine` might produce platform-specific metrics that look like `gcp.agent_engine.foo.bar`.

We have gone through the  round of such renames for `azure`, but not for other cloud platforms. 

Here we have a choice to be consistent with the 
1. rest of `cloud.provider = gcp_*` (and later on will do a breaking change to rename them all at once)
2. naming guidance (and then be inconsistent with other gcp cloud platforms)

My preference would be on 2 (use `gcp.agent_engine`). It also aligns better with `gen_ai.provider.name` which uses new naming guidance - https://github.com/open-telemetry/semantic-conventions/blob/ef973b361860408cc3bd4ad621037d9763f9d464/model/gen-ai/registry.yaml#L28

jsuereth (approved):

model/cloud/registry.yaml
aabmass: OK, let's go ahead with `gcp.agent_engine`, I'll update the PR.

Thank you both!

.chloggen/2957_gcp_agent_engine.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: entities
```
lmolkova: nit: 

```suggestion
component: cloud
```

lmolkova (approved):

PR #2959, opened 2025-10-22
[chore] Additional test case cases for SQL sanitisation and summary

stevejgordon:
## Changes

Adds extra test cases for SQL query text sanitisation and summary parsing to cover more scenarios.

We have used these test cases in the [.NET implementation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3116) while refining and optimising it. 

An open area for discussion is whether it should be optional to skip numeric literal sanitisation when the value is used as part of a TOP (e.g. `SELECT TOP (10) FROM MyTable`) statement or type declaration (e.g. `ALTER TABLE Orders ADD COLUMN OrderStatus NVARCHAR(50)`). The tests currently allow this as an optional expectation. 

cc @alanwest 

* [ ] ~Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).~
* [ ] ~Links to the prototypes or existing instrumentations (when adding or changing conventions)~

### DIFF
--- a/docs/non-normative/database-test-cases/db-sql-test-cases.json
+++ b/docs/non-normative/database-test-cases/db-sql-test-cases.json
@@ -294,7 +294,8 @@
         },
         "expected": {
             "db.query.text": [
-                "ALTER  TABLE MyTable ADD Name varchar(?)"
+                "ALTER  TABLE MyTable ADD Name varchar(?)",
+                "ALTER  TABLE MyTable ADD Name varchar(255)"
             ],
             "db.query.summary": "ALTER TABLE MyTable"
         }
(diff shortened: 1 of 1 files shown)


maryliag:
>whether it should be optional to skip numeric literal sanitization

I think we can give the option, as you mentioned, because it has the advantage of performance (since it's one less action to do), but just to confirm I still rather have sanitization as default, since having the value can increase cardinality.

docs/non-normative/database-test-cases/db-sql-test-cases.json
```diff
@@ -311,5 +312,864 @@
...
+        "SELECT * FROM SelectedData"
+      ],
+      "db.query.summary": "SELECT SelectedData"
+    }
+  },
+  {
+    "name": "summary_truncated",
+    "input": {
+      "db.system.name": "other_sql",
+      "query": "SELECT * FROM Vecnzotjejucwitzgrfifscuevittljlnrlpbruvkezeptqciyvbjhsytmbucbhwttidayecnthaztxbyppbyztcqccedeirgkxzrfezjxfwbtuqxeusroqgbvulgmsvnelovkxqsaqmlogomkhtjuirzhaocxlrmerihnmwaelullionarkmxwdamhduwrbooknqsnilurutgyerxphokeqnnoumbpcfjtmqrbpukjllofiwaltyoawkp o, OrderDetails od"
```
maryliag, line 333: this one just makes me think of
```
When performing sanitization, instrumentation MAY truncate the sanitized value for performance considerations (since sanitizing has a performance cost).
```

do have any an implementation that truncates the value? curious because we didn't provide a guide on the size we should truncate, but wondering if is worth a test

docs/non-normative/database-test-cases/db-sql-test-cases.json
```diff
@@ -311,5 +312,864 @@
...
+      "db.system.name": "other_sql",
+      "query": "SELECT * FROM table WHERE value IN ('abc', 0xAB, .456,"
+    },
+    "expected": {
+      "db.query.text": [
+        "SELECT * FROM table WHERE value IN (?, ?, ?,"
+      ],
+      "db.query.summary": "SELECT table"
+    }
+  },
```
alanwest, line 1146: I want to hear others' opinions on whether they think we should maintain "malformed" test cases in this shared test suite.

In .NET we test these cases to ensure we're not leaking any sensitive information because it is not easy for us to identify that the query is invalid. If other languages have the same challenge then it might make sense to keep these test cases. On the other hand, if other languages can more readily identify invalid queries, then I'd imagine they would simply refrain from setting `db.query.text`/`db.query.summary` in the first place.

docs/non-normative/database-test-cases/db-sql-test-cases.json
```diff
@@ -311,5 +312,864 @@
...
+        "CREATE ROLE app_admin;"
+      ],
+      "db.query.summary": "CREATE ROLE app_admin"
+    }
+  },
+  {
+    "name": "alter_role",
+    "input": {
+      "db.system.name": "other_sql",
+      "query": "ALTER ROLE app_admin ADD MEMBER johndoe;"
```
alanwest, line 845: I think this syntax may be unique to `microsoft.sql_server`. For example, postgres has an `ALTER ROLE` operation but it has a separate `GRANT` operation for what `ADD MEMBER` is doing here.

My original intent was for `db.system.name` to identify the dialect test case was meant to target. In this way, an implementation that only targets a specific dialect could be validated against only the relevant test cases. We were using `other_sql` to indicate that the test case is relevant for all dialects.

Since we've really only focused on MSSQL Server so far, there may be other test cases in this PR that should just be marked `microsoft.sql_server`. It's hard to tell though until we study other dialects a bit more. At this point, maybe it just makes sense to do our best to mark things `microsoft.sql_server` when we know for sure it's just SQL Server. We can come back and refine things as we learn more.

docs/non-normative/database-test-cases/db-sql-test-cases.json
maryliag: I think it make sense to keep it, I don't know if all languages can easily identify it was malformed, so I see a few possible scenarios:

**we can identify is malformed**
- replace sensitive information and send what we have sanitize
- don't return any query text, this way we save some resources and skip sanitization. In this case we can return some message about being malformed, but I assume it would cause an error to be created and we would knew it was malformed

**we can't identify is malformed**
- we need to sanitize, because we might think is a valid and send the query text

So I do think we should have the tests, but I'm not clear on what the results should be

docs/non-normative/database-test-cases/db-sql-test-cases.json
alanwest: This test case is about the truncation required for `db.query.summary` not the sanitized `db.query.text`.

[The spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#generating-a-summary-of-the-query) for generating the query summary states:

> Instrumentations that parse the query to set db.query.summary SHOULD truncate the summary to 255 characters (ensuring truncation does not occur within an operation name or target).

So in this case `db.query.summary` is equal to `SELECT` rather than

```
SELECT Vecnzotjejucwitzgrfifscuevittljlnrlpbruvkezeptqciyvbjhsytmbucbhwttidayecnthaztxbyppbyztcqccedeirgkxzrfezjxfwbtuqxeusroqgbvulgmsvnelovkxqsaqmlogomkhtjuirzhaocxlrmerihnmwaelullionarkmxwdamhduwrbooknqsnilurutgyerxphokeqnnoumbpcfjtmqrbpukjllofiwaltyoawkp OrderDetails
```

docs/non-normative/database-test-cases/db-sql-test-cases.json
alanwest: Would it make sense to mark test cases representing invalid SQL with a boolean of some sort?

That way if there are languages that would otherwise not set `db.query.summary` and `db.query.text` when the query is invalid can skip the test case or validate that these attributes are not set.

maryliag: make sense to me! This way the languages can decide how to handle the results

stevejgordon:
@joaopgrassi / @alanwest  - Is it worth re-opening this and trying to get final agreement on what can/should go into the example test cases? We're maintaining these tests in the .NET contrib project for now. If these are useful for other language implementations, we can dust them off here.

trask:
hey @stevejgordon! I've re-opened it. I'm interested in this as well. Can you look over the Java instrumentation tests and see what makes sense or not to unify?

- [`SqlQueryAnalyzerTest.java`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api-incubator/src/test/java/io/opentelemetry/instrumentation/api/incubator/semconv/db/SqlQueryAnalyzerTest.java)
- [`SqlQuerySummaryEdgeCasesTest.java`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api-incubator/src/test/java/io/opentelemetry/instrumentation/api/incubator/semconv/db/SqlQuerySummaryEdgeCasesTest.java)

stevejgordon:
@trask - Sure, I'll take a look and get this updated so we can have fresh conversations around what to include.

stevejgordon:
@trask I've been through both Java test files (with some help from Claude Code!) and compared them against the cases here. A good chunk is straightforwardly unifiable, but there are two areas where the two implementations disagree and the spec doesn't say either way, so I think those are worth settling before I add anything.

The first is comments and whitespace in `db.query.text`. Java preserves comments and collapses runs of whitespace to a single space, whereas these cases strip comments and preserve the original formatting. That isn't something this PR introduces: by my reading of Java's expectations, eight of the twenty-four cases already on main would fail against it on the query text, even though both agree on the summary. That feels like a question for the "Sanitization of db.query.text" section.

The second is a set of questions for "Generating a summary of the query": whether CTE names count as targets, how multiple statements are separated in a summary, whether `GRANT`/`REVOKE`/`SHOW` drop their targets, whether a prefix such as `EXPLAIN` is preserved as an operation, and whether table functions like `UNNEST(...)` are targets. Java has taken a position on all of these and it seems reasonable to me. We could bring the tests across and let that stand as the agreement, but implementations that don't run these tests would have no way of knowing. Also, .NET may fail on some of these new rules and we'd need to determine an appropriate path to update our implementation there. So the one thing I'd like to settle is whether we encode the agreed rules in the spec, or add a note pointing implementors at these tests as the "source of truth" for sanitisation and summary generation.

On scale, for context: `SqlQuerySummaryEdgeCasesTest` only asserts the summary, but of its 134 cases, 96 contain nothing that needs sanitising, so those can come across with `db.query.text` set to the input text and both implementations should pass as-is. Another 33 do contain literals, and they include the same `TOP`/type-declaration question from the PR description, plus `LIMIT`, `OFFSET` and `FETCH FIRST`, which fall in the same category. The remaining five are quoting edge cases I can deal with individually. For the 33 we can either keep accepting multiple valid outputs or take a position.

Happy to keep this here rather than opening an issue, unless you'd prefer to split the spec question out separately.

stevejgordon:
@trask - Just pinging to see if you had any thoughts before I proceed here. Also cc @alanwest.

lmolkova:
@stevejgordon @trask should we eventually move existing and new scenarios to https://github.com/open-telemetry/semantic-conventions-conformance ? we don't have DB helpers or tests there yet but we can build them in a way that uses this set as a contract (similar to [HTTP one](https://github.com/open-telemetry/semantic-conventions-conformance/blob/main/tools/http/test-client/contract.json))

trask:
oh good idea!

stevejgordon:
@lmolkova Looks like a good home. Shall we try to get the cases agreed here first, then look at moving them? In that new repo, are you proposing to keep the JSON file as a source that SDKs can pull and test against, as well as eventually including the contract testing also?

trask:
where there's ambiguity, it probably makes sense to add recommendations in this repo, e.g. comments and whitespace, but we could go ahead and start working on the contract testing in the new conformance repo

lmolkova:
@stevejgordon  no concerns adding them here in parts or as a whole, the worry I have is that here they are hard to validate / review and using them in actual tests involves copy-paste. The eyeball review process is something that stopped me from approving it in the first place.

In the conformance repo they'd be something we can plug into actual tests and see which instrumentations follow / not follow them and how far away the reality is. 

I'm fine starting here and moving over as long as we implemented these in instrumentations and see these test cases are implementable / valid.

martincostello:
My only concern regarding the conformance repo is that in the .NET SqlClient instrumentation the redaction isn't a public API - it's an internal implementation detail.

Actually validating that if you do a _specific_ query X you get reaction Y is probably going to be messy as the queries will all fail without a real backend with the exact schema/data in it. I'm sure it's _doable_ to some degree with mocks and error handling, but I think it'll be messy.

It also seems a lot more in-depth validation that just _"are these attributes present and do the values meet a schema"_.

trask:
> without a real backend with the exact schema/data in it

in general, I'm expecting the conformance tests to need a real backend when it comes to database and messaging instrumentations, i'll give it a try and see what it might look like

lmolkova:
> My only concern regarding the conformance repo is that in the .NET SqlClient instrumentation the redaction isn't a public API - it's an internal implementation detail.
> 
> Actually validating that if you do a _specific_ query X you get reaction Y is probably going to be messy as the queries will all fail without a real backend with the exact schema/data in it. I'm sure it's _doable_ to some degree with mocks and error handling, but I think it'll be messy.
> 
> It also seems a lot more in-depth validation that just _"are these attributes present and do the values meet a schema"_.

It should be possible to extend conformance to validate values too - they are in the raw weaver report and can be validated in any possible way once telemetry is collected. It should also be possible to have testcontainer with a database.

As long as there is a way to configure instrumentation to (un)redact things, .NET instrumentation should be testable.

I might have missed your concern @martincostello, lmk

martincostello:
What I mean is how we test the redaction right now is essentially passing the strings in the test cases corpus into a function that more or less accepts a string and returns a string, with no actual database querying happing (i.e. we unit test it).

There's no public API surface to do that in the instrumentation, so to do conformance tests outside our own codebase you'd need to actually do the queries.

lmolkova:
> What I mean is how we test the redaction right now is essentially passing the strings in the test cases corpus into a function that more or less accepts a string and returns a string, with no actual database querying happing (i.e. we unit test it).
> 
> There's no public API surface to do that in the instrumentation, so to do conformance tests outside our own codebase you'd need to actually do the queries.

I see, thanks! I think in conformance we should always do integration-level testing, so running queries would be the way to go.

PR #2964, opened 2025-10-23
[chore] [editoral] Add info on selecting signal type

thompson-tomo:
## Changes

This adds in some guidance on selecting a signal type based on questions regarding what you want to do.

A slight re-ordering has been done to improve grouping/readability.

> [!IMPORTANT]

### DIFF
--- a/docs/how-to-write-conventions/README.md
+++ b/docs/how-to-write-conventions/README.md
@@ -192,7 +192,7 @@ database.
 The first step in defining a telemetry signal is to decide on what signal type (
 Event, Entities, [Metric](https://opentelemetry.io/docs/concepts/signals/metrics/),
 [Span/Trace](https://opentelemetry.io/docs/concepts/signals/traces/)),
-to use. To help with the decision, below you will find some guidance based on bes practice:
+to use. To help with the decision, below you will find some guidance based on best practice:
 
 - I want to debug/investigate the operations which are being performed including how long. Use a span.
 - I want to be able to see what operation invoked my operation (parent) or what child operation was invoked. Use a span.


docs/how-to-write-conventions/README.md
```diff
@@ -191,14 +200,24 @@ TBD
...
-
 #### Defining events
 
 TBD
 
+## Recomendations
+
+### Prototyping
+
+It is strongly recommended to prototype proposed conventions in one or more instrumentations and:
```
jsuereth: Nit: I think we want this advise near the top -

Folks need to understand that we need real world implementations and use cases for semantic conventions.

docs/how-to-write-conventions/README.md
```diff
@@ -183,6 +170,28 @@ to that system. For example, when adding a new `db.system.name` value, create
 documentation and span definitions that show how generic attributes work for that
 database.
 
+### Defining entities
+
+Follow the [Entity Modeling Guide](resource-and-entities.md).
+
+### Defining Telemetry Signals
+
+The first step in defining a telemetry signal is to decide on what type of signal to use based on their strengths and weaknesses.
```
jsuereth: This is just a pro/con lists between signals, but still seems to lack guidance on when to use one vs another.  I think we should be more principled on the *goal* of a signal and tie it to the thing you're trying to accomplish to begin with.

E.g. a  strawman:

- I want to record how long an operations takes - When do I use metric vs. span? Should we define both?
- I need to debug an operation that occurs within a set of microservices - Define a span
- I want to record that something happened - Use an Event
- ... more examples here...

I'd use this as inspiration: [metric instrument selection guidance](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md#guidelines-for-instrumentation-library-authors)

docs/how-to-write-conventions/README.md
```diff
@@ -183,6 +170,28 @@ to that system. For example, when adding a new `db.system.name` value, create
...
+- Event:
+  - Pro: Can Record the exact time when something happened with ability to include a large amount of detail.
+  - Con: Vendor support is not at the same level as other signals and can only be associated with the active span no other signal.
+- Metric:
+  - Pro: Good for capturing measurements which are aggregated based on attributes.
+  - Con: Usually lacks contextual information due to aggregation.
+- Trace:
+  - Pro: Allows for the visualization of when an operation/process is running including start/stop as well as,
+    what children both local and remote are performed as a part of this operation.
+  - Con: Only emitted once trace is completed and lacks granularity to see what was happening when during the operation.
```
joaopgrassi: > lacks granularity to see what was happening when during the operation.

Hum not sure about this. What you mean by not enough granularity? Often people complain of too many spans, not too few I feel. 

If logs have span ids on them, then you can also see the logs that happened during the trace. And all the other operations that also happened during the trace. 

So I'm not sure this point has any merit. I feel the whole point of a trace is to already achieve that.

joaopgrassi (commented):
I think the idea is good, but we need more work on the actual guidance. I feel if I don't know and read this I still don't get which one should I pick out of the available signals. 

Maybe we should give examples or focus on problems. Like, I want to know how much memory my system is using -> use a metric.

docs/how-to-write-conventions/README.md
thompson-tomo: I have repositioned it to be higher in the doc.

docs/how-to-write-conventions/README.md
thompson-tomo: Have reworked it as suggested.

docs/how-to-write-conventions/README.md
thompson-tomo: Have reworked it.

The reason they complain of too many is that the implementation is using spans to see what was happening when to overcome the spans not having that granularity.

Anyway it has been reworked, hence closing.

thompson-tomo:
Have reduced the amount of lines/sections which have  been moved but some is necessary to create the section to put the new content in and have it flow naturally into the sections describing the signal types.

docs/how-to-write-conventions/README.md
```diff
@@ -183,6 +187,27 @@ to that system. For example, when adding a new `db.system.name` value, create
...
+
+### Defining Telemetry Signals
+
+The first step in defining a telemetry signal is to decide on what type of signal to use based on their strengths and weaknesses,
+as well as what you want to achieve.
+
+Here are some rules to guide you in choosing the correct signal type:
+
+- I want to debug/investigate the operations which are being performed including hoe long. Use a span.
+- I want to see what operations are triggered from another operation. Use a span.
```
jsuereth: I think it's slightly more nuanced than this.

In OpenTelemetry, since we propagate TraceID around, you can actually see operations via Event.

I think this is more about *parenting* or *grouping* operations.

docs/how-to-write-conventions/README.md
```diff
@@ -183,6 +187,27 @@ to that system. For example, when adding a new `db.system.name` value, create
...
+- I want to analyze how long operations take to complete. Use a metric.
+- I want to see how much/many of a resource is used/available. Use a metric.
```
jsuereth, line 200: Apologies if I plug the "SRE" mindset here, but this would be called "saturation" golden signals.
It's possible we could just have a rule of:

```
- I want to track golden signals (saturation, errors, latency, volume).  Use a metric.
```

docs/how-to-write-conventions/README.md
```diff
@@ -183,6 +187,27 @@ to that system. For example, when adding a new `db.system.name` value, create
...
+- I want to see when something happened ie state changed. Use an event.
```
jsuereth: Let's hold on this one.  I think this is where the Entity SIG needs to figure out the plan for the entity signal, and state / configuration tracking.

Additionally, today, we're using metrics for this (e.g. kube state metrics).

docs/how-to-write-conventions/README.md
```diff
@@ -183,6 +187,27 @@ to that system. For example, when adding a new `db.system.name` value, create
...
+### Defining entities
+
+Follow the [Entity Modeling Guide](resource-and-entities.md).
+
+### Defining Telemetry Signals
+
+The first step in defining a telemetry signal is to decide on what type of signal to use based on their strengths and weaknesses,
+as well as what you want to achieve.
+
+Here are some rules to guide you in choosing the correct signal type:
```
jsuereth: This is already a much better list, and kudos for lifting it up to this level with an initial list.

I think defining this list is pretty hard, and we'll have to take a few iterations.  That said, let's figure out some *key* questions folks have in observability.

A suggested first cut from what you have and my own experience:

- I want to track an operation in my system, how long it takes and what interactions it has with other parts of my system: Use a span
- I want to track "golden signals" (latency, traffic, errors, and saturation): Use a metric
- I'm unable to debug my current system on failures: Add Spans and Events
- [we can cut this, just my current thinking for today] I want to track state changes
  - I need to overlay these changes on metrics or alert on them - Use a metric
  - I need high fidelity - Use an event

I think if we combine this with a pros/cons list (somewhat similar to what you had, but targeted at ambiguous cases), we have a good start.

docs/how-to-write-conventions/README.md
```diff
@@ -183,6 +187,27 @@ to that system. For example, when adding a new `db.system.name` value, create
 documentation and span definitions that show how generic attributes work for that
 database.
 
+### Defining entities
```
jsuereth: Why are Entities no longer signals?

Let's not refactor the layout beyond adding which signal to use.

See my comment about state/config + entity.

jsuereth (commented):
This is looking much improved, but still a hard task to accomplish.

Some suggestions and thoughts.

docs/how-to-write-conventions/README.md
thompson-tomo: Previously we were not Grouping based on signal type. I have not put it under signal types and added an additional rule for when to define an entity.

docs/how-to-write-conventions/README.md
thompson-tomo: The key thing is "when", have re-worked the wording to make it more explicit is it includes the time.

docs/how-to-write-conventions/README.md
thompson-tomo: Wording has been tweaked to increase focus on parent/child relationships.

docs/how-to-write-conventions/README.md
thompson-tomo: Have encorporated a rule similar to as suggested.

docs/how-to-write-conventions/README.md
thompson-tomo: I have tweaked and extended the rules incorporating your feedback. I am not sure about your third point as it feels too vague and could lead to span overload.

In relation to pros and cons if we have the rules, I don't see a need to have them here. Perhaps they would be better suited in the concepts section of otel.io that way a reader of the concept can see what the strengths and weaknesses are.

docs/how-to-write-conventions/README.md
```diff
@@ -183,6 +187,25 @@ to that system. For example, when adding a new `db.system.name` value, create
 documentation and span definitions that show how generic attributes work for that
 database.
 
+### Defining Telemetry Signals
+
+The first step in defining a telemetry signal is to decide on what signal type (
+Event, Entities, [Metric](https://opentelemetry.io/docs/concepts/signals/metrics/),
```
kamphaus, line 193: Since we link to metrics and traces, can we also link to event and entities?

thompson-tomo: I would've liked to but there is no concept page for them.

thompson-tomo:
Still looking for more inputs

docs/how-to-write-conventions/README.md
thompson-tomo: @jsuereth what is needed to progress this further as a first release?

trask:

PR #2963, opened 2025-10-23
Initial cut at modelling service entities based on SIG discussions.

jsuereth:
Fixes #2880

## Changes

- Updates the `service` entity to consist of three components and describes each
- Adds a service specific markdown file for resource modelling
- Migrates existing attributes to appropriate entities in service.

> [!IMPORTANT]
> Pull requests acceptance are subject to the triage process as described in [Issue and PR Triage Management](https://github.com/open-telemetry/semantic-conventions/blob/main/issue-management.md).
> PRs that do not follow the guidance above, may be automatically rejected and closed.

## Merge requirement checklist

* [ ] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [ ] Links to the prototypes or existing instrumentations (when adding or changing conventions)

PR #2970, opened 2025-10-24
[chore] Clean up Entity snippets - make sure they include identifying vs. descriptive roles

jsuereth:
- Re-use same macros both in registry and entity snippets.
- Make sure identifying/descriptive attributes show up in snippets.
- Ensure warning about identifying/descriptive attributes and stability shows up in snippets.
- Move all attributes for an entity into a single table.


lmolkova:
WDYT about something like 

| Attribute | Type | Description | Examples | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
|---|---|---|---|---|---|
| [`service.name`](/docs/registry/attributes/service.md) Identifying | string | Logical name of the service. [1] | `shoppingcart` | `Required` | Stable |
| [`service.instance.id`](/docs/registry/attributes/service.md) Identifying | string | The string ID of the service instance. [2] | `627cc493-f310-47de-96bd-71410b7dec09` | `Recommended` | Development |
| [`service.namespace`](/docs/registry/attributes/service.md) Identifying | string | A namespace for `service.name`. [3] | `Shop` | `Recommended` | Development |
| [`service.version`](/docs/registry/attributes/service.md) | string | The version string of the service API or implementation. The format is not defined by these conventions. | `2.0.0`; `a01dbef8a` | `Recommended` | Stable |


vs (in the current version)

| | Attribute | Type | Description | Examples | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
|---|---|---|---|---|---|---|
| **identity** | | | | | |
| | [`service.name`](/docs/registry/attributes/service.md) | string | Logical name of the service. [1] | `shoppingcart` | `Required` | Stable |
| | [`service.instance.id`](/docs/registry/attributes/service.md) | string | The string ID of the service instance. [2] | `627cc493-f310-47de-96bd-71410b7dec09` | `Recommended` | Development |
| | [`service.namespace`](/docs/registry/attributes/service.md) | string | A namespace for `service.name`. [3] | `Shop` | `Recommended` | Development |
| **description** | | | | | |
| | [`service.version`](/docs/registry/attributes/service.md) | string | The version string of the service API or implementation. The format is not defined by these conventions. | `2.0.0`; `a01dbef8a` | `Recommended` | Stable |

it might work similar to @joaopgrassi 's proposal for sampling-relevant attrs here 
https://github.com/open-telemetry/semantic-conventions/pull/2678#issuecomment-3228485610

jsuereth:
@lmolkova I'd prefer to have it (a) sorted so identity is first and (b) a clear separation between identity + description.  I think, going forward, it's going to be important to delineate what these mean to folks using entities.

Anyone in @open-telemetry/entities-maintainers have thoughts?

joaopgrassi:
Also I guess we want to align the columns with https://github.com/open-telemetry/semantic-conventions/pull/2867?

docs/registry/entities/service.md
```diff
@@ -7,19 +7,20 @@
 
 ## Service
 
-**Status:** ![Stable](https://img.shields.io/badge/-stable-lightgreen)
+**Status:** ![Mixed](https://img.shields.io/badge/-mixed-yellow)
```
lmolkova: from the Semconv SIG call discussion, could we use the same status as in yaml? 


There've been some exceptions made for service entity allowing us to have stable status and unstable identity attributes, and we should stay consistent between yaml and markdown.

It seems yaml will be corrected in the https://github.com/open-telemetry/semantic-conventions/pull/2963 - 👍

jsuereth: done

jsuereth:
@lmolkova PTAL at latest incarnation of the table - it's a blend of what you did and what I did

docs/registry/entities/android.md
```diff
@@ -13,14 +13,15 @@
 
 **Description:** The Android platform on which the Android application is running.
 
-**Other Attributes:**
 
 > :warning: This entity definition contains attributes without a role.
 > Stable Entities MUST NOT have attributes without a defined role.
```
thompson-tomo, line 17: Is there a way we could enforce this in rego and remove it from docs?

templates/registry/markdown/entity_macros.j2
```diff
@@ -1,6 +1,57 @@
...
+{%- endmacro %}
+{% macro all_attribute_table(e) -%}
+{%- set id_attrs = e.attributes | selectattr("role", "equalto", "identifying") -%}
+{%- set desc_attrs = e.attributes | selectattr("role", "equalto", "descriptive") -%}
+{%- set misc_attrs = e.attributes | rejectattr("role", "defined") -%}
+{%- if misc_attrs | length > 0 %}
+
+> :warning: This entity definition contains attributes without a role.
+> Stable Entities MUST NOT have attributes without a defined role.
+
```
thompson-tomo:
```suggestion

```

To reduce un-necessary line breaks in the docs

templates/registry/markdown/entity_macros.j2
```diff
@@ -1,6 +1,57 @@
...
+{% endif %}
+| Role | Attribute | Type | Description | Examples | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|---|{%- if id_attrs | length > 0 %}
+{{ attribute_table(id_attrs, "/docs/registry/attributes", e.lineage.attributes, notes, "identity") | trim }}{% endif %}{%- if desc_attrs | length > 0 %}
+{{ attribute_table(desc_attrs, "/docs/registry/attributes", e.lineage.attributes, notes, "description") | trim }}{% endif %}{%- if misc_attrs | length > 0 %}
+{{ attribute_table(misc_attrs, "/docs/registry/attributes", e.lineage.attributes, notes, "other") | trim }}
+{% endif %}
+{{ notes.render() }}
```
thompson-tomo:
```suggestion
{{ notes.render() | trim }}
```

Also reducing un-necessary line breaks

templates/registry/markdown/entity_namespace.md.j2
```diff
@@ -2,46 +2,13 @@
...
-
-**Other Attributes:**
-
-> :warning: This entity definition contains attributes without a role.
-> Stable Entities MUST NOT have attributes without a defined role.
-
-{{ at.generate(misc_attrs, "", "/docs/registry/attributes", e.lineage.attributes) | trim}}
-{%- endif %}
-
+{{ entity.header(e) }}{{ entity.all_attribute_table(e) }}
```
thompson-tomo:
```suggestion
{{ entity.header(e) | trim }}

{{ entity.all_attribute_table(e) | trim }}
```

To tackle multiple blank lines

templates/registry/markdown/entity_macros.j2
```diff
@@ -1,6 +1,57 @@
...
+{% else %}![Other](https://img.shields.io/badge/-other-red)
+{% endif %}
+{%- endmacro %}
+{#- Header for snippets. -#}
+{% macro header(entity) -%}
+**Status:** {{ stability.badge(entity.stability, entity.deprecated, entity.brief) | trim }}
+
+**type:** `{{ entity.name }}`
+
+**Description:** {{ entity.brief }}
```
thompson-tomo:
```suggestion
**Description:** {{ entity.brief | trim }}
```

thompson-tomo (commented):
Some general comments to tackle repeated empty lines.

Also is it still the plan to remove the resource pages as per #2327

docs/registry/entities/container.md
```diff
@@ -69,18 +73,11 @@ The ID is assigned by the container runtime and can vary in different environmen
...
-
-**Descriptive Attributes:**
-
-| Attribute  | Type | Description  | Examples  | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
-|---|---|---|---|---|---|
-| [`container.runtime.description`](/docs/registry/attributes/container.md) | string | A description about the runtime which could include, for example details about the CRI/API version being used or other customisations. | `docker://19.3.1 - CRI: 1.22.0` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
+| Role | Attribute | Type | Description | Examples | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Stability |
+|---|---|---|---|---|---|---|
+| ![Identity](https://img.shields.io/badge/-identity-purple) | [`container.runtime.name`](/docs/registry/attributes/container.md) | string | The container runtime managing this container. | `docker`; `containerd`; `rkt` | `Recommended` | ![Development](https://img.shields.io/badge/-development-blue) |
```
joaopgrassi: not sure if it's just me, but the `identity` and `descriptive` badges are super small and a bit hard to read.

Image

Maybe we have to play with the size? Like in here https://github.com/badges/shields/discussions/5672

docs/registry/entities/android.md
jsuereth: It is, this is a notice to SIGs that they need to update their definitions.  Consider this advertising so you're not surprised on build failure.

jsuereth:
> Also is it still the plan to remove the resource pages as per https://github.com/open-telemetry/semantic-conventions/issues/2327

Yes, with the caveat - there's a lot more urgent things to sort out first.

docs/registry/entities/container.md
jsuereth: So we can't modify size with PURE markdown, we'd need to use an HTML tag for that.

Also - these files should be the same size as the stability tags, if they use the same mechanism.

Do you remember why we needed to avoid HTML tags in our markdown for opentelemetry.io?

jsuereth: I'm also not seeing the same size as you:

image

joaopgrassi:
I liked the badges 😞. Hopefully we can find a long-term solution with the html limitation with the OTel website. CC @chalin we cannot use html inside markdown tables, right?

chalin:
@joaopgrassi - OTel.io isn't the limitation. Here's my take on all this. IMHO:

- The **tables are complicated enough already**. Adding raw HTML is just going to make editing them more of a challenge.

- I see this repo as the raw source of the semconv -- that's why pages are written in Markdown rather than a markup language or Rich Text Format. We should avoid stylistic embellishments and focus on semantics/content.

- Previewing markdown pages via GH is a convenience. As you know GH preview has limitations that are by design: you can't use custom styling, nor redirects, etc. In order to use features such as styles, etc., GH recommends using GH Pages or, like we do, other tools such as Hugo.

We can fix the badge styling in OTel.io using CSS. (We could even implement an auto-rendering of badges for certain (solo) keywords like "Stable" and "Development" that might appear in tables (maybe via a special marker `!Stable`))

Thoughts?

As I looked at the semconv pages on OTel.io, I just noticed that badges aren't rendering properly. I'll create an issue for that.

/cc @open-telemetry/docs-maintainers

chalin:
@jsuereth wrote:

> Do you remember why we needed to avoid HTML tags in our markdown for opentelemetry.io?

You _can_ have HTML inside a Markdown table cell. 

Maybe you are remembering https://github.com/open-telemetry/semantic-conventions/issues/2536? The problem there was that links inside HTML-encoded tables weren't getting processed correctly.

chalin:
Let me know if you want me to look into the badge-size issue. FYI, I'll be fixing the current bug with badge display via:

- https://github.com/open-telemetry/opentelemetry.io/issues/8288

jsuereth:
@chalin AH yes, I think you're right.  I couldn't emember the details and wanted to avoid breaking opentelemetry.io.  Thanks for the links, so images are ok, but that gets into your other comment.

I absolutely agree with what you're saying - github pages are a convenience and we need rich markup going forward.  If you think this is fixable on opentelemetry.io I'm happy to defer to that.  I agree these tables are already too complicated.

docs/registry/entities/container.md
jsuereth: alright, removed the badge

PR #2971, opened 2025-10-25
[chore] [editoral] switch to md frontmatter

thompson-tomo:
Progresses #1821 

## Changes

This switches the manual docs to use github front matter syntax rather than embedding the front matter in comments. Currently these comments need to be processed before they can be used on the website version.

This change will also help in reviews as the link title etc are front and center.

The docs now all have an explicit title front matter property which MUST match the h1 and is enforced by textlint.

Textlint also enforces that the title is always the first property followed by linkTitle when it is defined.

Question's:
- ~~Do we need to keep `# Hugo front matter used to generate the website version of this page`? Or can it be removed?~~ Has been removed based on feedback

> [!IMPORTANT]


thompson-tomo:
Build issue due to current toc generator having known issues with front matter syntax & isn't mantained. Will need to be addressed by switching to https://www.npmjs.com/package/doctoc

thompson-tomo:
@chalin finally found a way to keep the build process happy. The one thing which i am not sure on is if the aliases in [resource-and-entities.md](https://github.com/open-telemetry/semantic-conventions/pull/2971/files#diff-45c6c3aee5d1ce4f912148cf9efc0f39984b2c4c8f67f806bd99d6ce263dac21) are still valid. If not we could switch to a redirect

chalin:
Nice! Thanks @thompson-tomo. I'll get back to you as soon as I can (it's a very busy EOM).

thompson-tomo:
No worries, I will fix the formatting which is causing the build to fail very soon.

trask:
> By switching to front matter directly it enables turning on the first line h1 check.
>
> Note the awk/sed command are necessary so that the document still passes the lint rules and we don't have a formatter to format it.

can these be done in a follow-up PR? if so, please revert those changes to keep this large mechanical PR easy to review

docs/README.md
```diff
@@ -1,4 +1,5 @@
-<!--- Hugo front matter used to generate the website version of this page:
+---
+# Hugo front matter used to generate the website version of this page:
```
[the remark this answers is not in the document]
chalin: I haven't had any issues in the past with such comments inside YAML-formatted front matter.

chalin (commented):
A first few questions and comments:

- Spec and semconv folks expressedly asked me to include the comment "Hugo front matter used to generate the website version of this page" so that all contributors know why the front matter is there (and hopefully realize that it's important to keep it up to date). Maybe this preference has changed. I'll let @open-telemetry/specs-maintainers @open-telemetry/specs-semconv-maintainers comment about this. I'm ok either way.
- Why not keep the (inline) array syntax for aliases?

chalin:
> The one thing which i am not sure on is if the aliases in [resource-and-entities.md](https://github.com/open-telemetry/semantic-conventions/pull/2971/files#diff-45c6c3aee5d1ce4f912148cf9efc0f39984b2c4c8f67f806bd99d6ce263dac21) are still valid. If not we could switch to a redirect

Yes the alias is valid, but I'd rather keep the inline array syntax, esp. in these cases that have a single entry.

chalin:
FYI, I haven't had the time to look into this more, but in support of #1821 and https://github.com/open-telemetry/opentelemetry.io/issues/6101. We probably will need to have a `title` field in the front matter too.

I'm wrapping up and will be OOO so I won't be able to get back to this until the new year (FYI).

thompson-tomo:
> Spec and semconv folks expressedly asked me to include the comment "Hugo front matter used to generate the website version of this page" so that all contributors know why the front matter is there (and hopefully realize that it's important to keep it up to date). Maybe this preference has changed. I'll let @open-telemetry/specs-maintainers @open-telemetry/specs-semconv-maintainers comment about this. I'm ok either way.

The issue I ran into was that this causing issues with the toc plugin. If necessary I can look to put it back and hope the current awk/sed commands keep it happy or possibly extend it.

> Why not keep the (inline) array syntax for aliases?

This is being caused by the necessity to process the docs with awk/sed after toc generation to keep linting happy. I would've preferred to keep them unchanged as well.

> We probably will need to have a title field in the front matter too.

Yeap am aware and that would be tackled by seperate PR.

chalin:
> Build issue due to current toc generator having known issues with front matter syntax & isn't mantained. Will need to be addressed by switching to https://www.npmjs.com/package/doctoc

OMG, the last update was 8 years ago in NPM. doctoc seems like an interesting choice; https://npm-compare.com/doctoc,markdown-toc

Maybe it would be worth switching toc tools before working further on this PR?

/cc @vitorvasc

thompson-tomo:
I certainly am open to switching tools. If I remember correctly there was issues using do toc. What I found was:
- https://github.com/thlorenz/doctoc/issues/278
- https://github.com/thlorenz/doctoc/issues/279

So doctoc would only work if we removed comments from front matter then & also remove the h1.

Note doc toc also has not had a release in years.

chalin:
Hmm. Ok, maybe another tool then.
(Creating a markdownlint rule to handle tocs might be a fun project ;))

thompson-tomo:
Let's see if I can a feature into doctoc to ignore level 1 headings as i wasn't having much luck finding an alternative.

chalin:
Btw, if you still need to massage the files, rathr than awk & sed, I'd suggest using Perl because it is more portable across OSs and container environments in my experience (esp. if you avoid Perl modules that don't ship with Perl).

thompson-tomo:
So just an update, I have been able to merge in feature/fixes to doctoc which enables it to be a viable option for us. Hopefully a release will occur soon. At which point a seperate PR will be raised to switch this repo over & this PR will be rebased.

thompson-tomo:
@chalin can you start looking into this. Note I am still working towards a release of doctoc which will eliminate the grep etc but that will be seperate PR.

chalin:
> Note I am still working towards a release of doctoc which will eliminate the grep etc but that will be seperate PR.

Great that you've been able to get some doctoc changes merged. 🙌🏻 

While I'm eager to make progress here, our current setup works. I'd rather not move forward until we have a full solution that includes:

- `title` field in the front matter. Both the in-page h1 and title text should match (could be enforced in this repo)
- I think that if we're going to change the toc tool, that should be done first
- IMHO, we should reach a point where the only change introduced in the .md files by this PR would be to replace the HTML comment delimiters (used to wrap the front matter) by YAML `---` section delimiters.

WDYT?

thompson-tomo:
> - title field in the front matter. Both the in-page h1 and title text should match (could be enforced in this repo)

There doesn't appear to be any usage of `title` front matter property but rather just `linkTitle` hence I don't think anything needs to be done here

> - I think that if we're going to change the toc tool, that should be done first

The options are:
- change toc tool to doctoc and be able to retain hugo comment
- remain with current tool but remove hugo comment and use grep.
- remain with current tool but remove hugo comment and use Prettier. Prettier has gotten push back hence grep option.

> - IMHO, we should reach a point where the only change introduced in the .md files by this PR would be to replace the HTML comment delimiters (used to wrap the front matter) by YAML --- section delimiters.

I assume you don't mean to create a seperate PR to split the hugo comment to be on a separate line if you want to keep the hugo comment?

thompson-tomo:
Still being worked on and am awaiting clarification.

chalin:
> There doesn't appear to be any usage of `title` front matter

- The OTel.io layouts expect a `title`, so it needs to be there (at least for now).

---

- Thinking about the Hugo comment: at this point I think that there's enough maintainer / community experience that we could probably drop it.

- I'd vote to switch tools first.

---

> I assume you don't mean to create a seperate PR to split the hugo comment to be on a separate line if you want to keep the hugo comment?

Yeah, assuming the Hugo comment goes it mostly looks like what is done in this PR (execpt for the array format changes that should be undone):

image

docs/general/attributes.md
```diff
@@ -1,7 +1,8 @@
-<!--- Hugo front matter used to generate the website version of this page:
+---
 linkTitle: Attributes
-aliases: [general-attributes]
---->
+aliases:
+  - general-attributes
```
thompson-tomo: Undo

thompson-tomo:
@chalin i have now gone and added title front matter to all the pages, there is a custom textlint rule which ensures that they match if not report an error, arrays have been switched back. Please review and let me know if anything other resolving the TOC issue is needed.

thompson-tomo:
@chalin this repo can now switch to doctoc as update is now available. PR to switch is #3463 and as such can you review this PR knowing that the ci issues will be resolved by merging the linked pr.

chalin (changes requested):
> @chalin i have now gone and added title front matter to all the pages, 

Great!

Let's follow convention and ensure that the `title` is the first front-matter field.

> there is a custom textlint rule which ensures that they match if not report an error, 

Nice, though I'd rather not introduce textlint into the toolchain if we can avoid it. I'd prefer that we use a custom Markdownlint rule instead (I've been writing custom mdl2 rules lately, so I can look into it if that can help).

That being said, if we don't want to delay the front-matter conversion of this PR, we can move the tooling bits to a followup PR. 

> @chalin this repo can now switch to doctoc as update is now available. PR to switch is #3463 ...

Well done! I left a review comment with suggested changes. Looking forward to seeing that merged once suggestions are addressed.

thompson-tomo:
> Let's follow convention and ensure that the title is the first front-matter field.

Will also action this when on a computer later today.

> I'd prefer that we use a custom Markdownlint rule instead (I've been writing custom mdl2 rules lately, so I can look into it if that can help).
>
> That being said, if we don't want to delay the front-matter conversion of this PR, we can move the tooling bits to a followup PR.

Let's do tooling seperate, would you rather I remove textlint from this PR or keep it and then remove the custom rule later?

chalin:
> Let's do tooling seperate, would you rather I remove textlint from this PR or keep it and then remove the custom rule later?

Factor it out into another PR (and remove it from this one).

chalin:
To followup on the textlint rule: I'm not against it, esp. if textlint does become a part of the toolchain (as it has in the main spec repo), but we can discuss that separately now that there will be another PR for the tool.

thompson-tomo:
No worries @chalin textlint has been removed and front matter properties reordered. Let me know if there is anything else. 🙂

thompson-tomo:
Still awaiting @chalin feedback

chalin:
@thompson-tomo - let's get #3463 in first, I just submitted a review.

Btw, I was OOO in March. Been juggling other priorities since then, but this is still on my radar. Note that the follow will be supportive of the work done in this PR:

- https://github.com/open-telemetry/opentelemetry-specification/issues/5049

thompson-tomo:
@chalin when you get a change could you circle back and review this. Note I have split the auto generated docs out into https://github.com/open-telemetry/semantic-conventions/pull/3671. This should make it easier to review as the pages are auto-generated by the templates.

thompson-tomo:
@open-telemetry/specs-semconv-maintainers can we re-open as we await @chalin or someone else to review. Key thing is this is needed to improve website publishing process & is a pre-requiste for introducing specs.opentelemetry.io.

chalin:
@thompson-tomo - can you ensure that this is up to date, and I'll start final tests on the otel.io side to ensure that this can land smoothly.

CONTRIBUTING.md
```diff
@@ -215,9 +214,10 @@ At the top of all Markdown files under the `docs/` directory, you will see
 headers like the following:
 
 ```md
-<!--- Hugo front matter used to generate the website version of this page:
+---
+title: HTTP
```
chalin: This isn't a realistic example. We need a full `title` here, not an echo of the abridged `linkTitle`:

```suggestion
title: Semantic conventions for HTTP
```

.markdownlint.yaml
```diff
@@ -16,3 +16,5 @@ no-space-in-code: false
 first-line-h1: false
 no-duplicate-heading:
   siblings_only: true
+single-title:
+  front_matter_title: ""
```
chalin, line 20: @thompson-tomo - what does this do again?

chalin (commented):
@thompson-tomo - FYI, I'll be making some progress on this this month. See inline comments for a tweak and question.

Btw, what's in place to ensure that the Hugo fm title and page title match. I thought that we had a checker in place, but I'm not seeing that in the scope of this PR.

.markdownlint.yaml
thompson-tomo: This disables mdlint from flagging that there is a front matter title & a h1 heading.

docs/general/attribute-requirement-level.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Attribute requirement levels
+linkTitle: Attribute requirement levels
```
chalin: As clarified via https://github.com/open-telemetry/semantic-conventions/pull/3843#discussion_r3521236521, there should be no `linkTitle` entry when it matches the `title`. This applies to many changes in this commit.

.markdownlint.yaml
chalin: Ah, right,... and I see from the docs (https://github.com/DavidAnson/markdownlint/blob/main/doc/md025.md), that `""` is the convention. 👍

thompson-tomo:
I have just updated the branch and completed an audit so it is up to date. Note #3671 will update the files under the registry directory as they are 100% auto generated.

So currently there is nothing in place to keep them in sync. I did previously have a custom textlint rule in-place.

Note I already have a pr open to perform the textlint corrections for casing etc #3747 & a draft pr to enforce textlint in ci. #3578.

docs/README.md
```diff
@@ -1,4 +1,6 @@
...
```
chalin: For sake of consistency in this PR (and moving forward), we might as well drop this comment (I seem to recall we went back and forth on this, but I'm ok dropping it):

```suggestion
```

chalin:
> I have just updated the branch and completed an audit so it is up to date.

How did you "complete the audit"? Do you have a script? Is it a part of this repo already?

> So currently there is nothing in place to keep them in sync ...

I'm ok landing this earlier and brining in the sync check later.

> ... I did previously have a custom textlint rule in-place.

~Right, that rings a bell. Where is that work? Is textlint already a part of this repo's toolchain? I vaguely recall commenting on this (in this repo or the main spec repo?); possibly even mentioning that I'd more likely favor use of markdownlint instead.~ (Nm, found the context) In any case, we can land the changes first so that this PR isn't delayed.

thompson-tomo:
> How did you "complete the audit"? Do you have a script? Is it a part of this repo already?

I manually went through checking that each page under docs except the registry section. had both a title & linkTitle. The titles were previously all set via a script which I doubt I have anymore. Link titles were mostly already there.

joaopgrassi (changes requested):
Before even reviewing this, I'd like us to have a mechanism to:

1. Check that each page has the frontmatter
2. That it matches the header

Without it this will become stale, so I think it needs to be done together and not as a follow up.

thompson-tomo:
@joaopgrassi based on your request for linting of content to ensure that frontmatter is present & matches h1. Can you review https://github.com/open-telemetry/semantic-conventions/pull/3578 which adds textlint (used in otel.io & spec) to do the linting and then in this pr we would configure the additional linting rules/plugins.

thompson-tomo:
@chalin i have committed the textlint rules to https://github.com/wemicroit/textlint-rules and hope to have 1 if not both published within the next few days. By publishing it we can use the same rule in the other spec repos.

docs/general/attribute-requirement-level.md
thompson-tomo: My mistake. Have reverted that.

thompson-tomo:
@chalin the scenarios of linkTitle being the same as title has been addressed.

Note https://github.com/open-telemetry/semantic-conventions/pull/3671 should be a quick pr to review as those markdown files are all system generated from 1 of 4 templates.

thompson-tomo:
@joaopgrassi & @chalin we now have textlint validating that there is a title frontmatter property and that it matches the h1. When it doesn’t match, it will fail the ci. It will also fail if frontmatter is not present in the doc.

overbalance (approved):

thompson-tomo:
Awaiting review before fixing conflicts.

Issue #1661, opened 2024-12-09
Rename linux.memory.* metrics and attributes to memory.linux.* 


rogercoll:
### Area(s)

area:system

### Is your change request related to a problem? Please describe.

The root namespace should contain the overall category the name belongs to. The important category in these schemas is `memory`, the `linux` namespace just gives additional contextual information.

Reference https://github.com/open-telemetry/semantic-conventions/pull/1364#discussion_r1852465994

### Describe the solution you'd like

**Attributes**
- `linux.memory.slab.state` -> `memory.linux.slab.state`

**Metrics**
- `system.linux.memory.available` -> `system.memory.linux.available`
- `system.linux.memory.slab.usage` -> `system.memory.linux.slab.usage`

mx-psi:
This was discussed on the general semconv meeting from 2025-02-03 but no decision was made. PR #1708 would give general guidance, if we deviate from it we would have to explicitly state it

dmitryax:
This now can be renamed given that we have the guidance https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/groups/system/design-philosophy.md#operating-system-in-names

PR #2984, opened 2025-10-27
Rename *.linux.memory to *.memory.linux

rogercoll:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1661

## Changes

> [!IMPORTANT]

### DIFF
--- a/docs/non-normative/groups/system/design-philosophy.md
+++ b/docs/non-normative/groups/system/design-philosophy.md
@@ -213,7 +213,7 @@ The `brief` field should explain what the metric/attribute is, and if the
 explanation of a value is simple (i.e. simply surfacing a value from a common
 source like `procfs`) then the explanation of what the value should be can go in
 the brief. If the value needs some calculation explanation and justification,
-the information should be moved to the `note` field.  
+the information should be moved to the `note` field.
 For enum values, it is often the case that the intention of these values is
 obvious given whatever `brief` was provided for the attribute as a whole. A
 brief can be included in a scenario where we have had to make some choice on the
@@ -277,6 +277,7 @@ down with more specificity.
 Relevant discussions:
 [\#1255](https://github.com/open-telemetry/semantic-conventions/issues/1255),
 [\#1364](https://github.com/open-telemetry/semantic-conventions/pull/1364#discussion_r1852465994)
+[\#2984](https://github.com/open-telemetry/semantic-conventions/pull/2984#discussion_r2466369361)
 
 Monitoring operating systems is an old practice, and there are numerous heavily
 differing approaches within different platforms. There are lots of metrics, even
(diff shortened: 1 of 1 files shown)


rogercoll:
Reopening because of this PR changes removing any metric/attribute in the Linux area

.chloggen/rename_system_linux.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: system, linux
```
lmolkova, line 10: mind updating 

https://github.com/open-telemetry/semantic-conventions/blob/ef973b361860408cc3bd4ad621037d9763f9d464/areas.yaml

and https://github.com/open-telemetry/semantic-conventions/blob/main/.github/CODEOWNERS to put linux area under @open-telemetry/semconv-system-approvers ? 

this will make automation happier.

Thanks!

model/system/metrics.yaml
```diff
@@ -545,21 +545,21 @@ groups:
     entity_associations:
       - host
 
-  - id: metric.system.linux.memory.slab.usage
+  - id: metric.system.memory.linux.slab.usage
     type: metric
-    metric_name: system.linux.memory.slab.usage
+    metric_name: system.memory.linux.slab.usage
```
lmolkova, line 550: I think https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/groups/system/design-philosophy.md#operating-system-in-names talks about root namespace, is there a need or benefit in renaming `system.linux.memory.slab.usage` to `system.memory.linux.slab.usage` ?

In the new version, linux looks like a property of `memory` and separates actual `slab.usage` property. In the old version it feels in the right place - it's a property of a `system`.

rogercoll: We might need to rephrase it, but I think instead of "root namespace" what we meant is {area}. The way we think a user interacts with system metrics is by first looking into the area of interest (e.g. memory consumption alert) and then (if needed), looking to the OS more fine-grained metrics for that area.

If we just apply the rule to root namespaces, we could end up having a mix of unordered metrics like `memory.linux.*` and `system.linux.memory.*`.

cc @braydonk

.chloggen/rename_system_linux.yaml
rogercoll: Definitely, added in https://github.com/open-telemetry/semantic-conventions/pull/2984/commits/b45cd4730b16a6cfaf761320a7b5a57763e40799

joaopgrassi (approved):

model/system/metrics.yaml
lmolkova: I'm not going to block on this, but I'm still raising concern. It seems random where OS name is included and it affects name readability

lmolkova: Semconv SIG discussion: could you please update the guidance based on this discussion? thanks!

rogercoll: Added a note in the guidance docs: https://github.com/open-telemetry/semantic-conventions/pull/2984/commits/14a4c234f97eb9022a146a9a198b9014e9b5e924

Wdyt? Thanks for bringing this to the SIG

lmolkova (approved):

PR #2992, opened 2025-10-28
feat: Add Browser click event

joaquin-diaz:
## Changes

Added a new event for browsers named user_action.click. This event captures the "click" action in browsers

Prototype instrumentation:
https://github.com/open-telemetry/opentelemetry-browser/pull/35

> [!IMPORTANT]

### DIFF
--- a/docs/browser/browser-events.md
+++ b/docs/browser/browser-events.md
@@ -70,14 +70,16 @@ This event describes a user click action in the browser.
 
 | Key | Stability | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Value Type | Description | Example Values |
 |---|---|---|---|---|---|
-| [`browser.css_selector`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | Target element CSS Selector | `#main > div:nth-child(2) > button.submit` |
+| [`browser.css_selector`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | Target element CSS Selector [1] | `#main > div:nth-child(2) > button.submit` |
 | [`browser.page.x`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | int | Click x (horizontal) coordinates (in pixels) relative to the entire document. | `10` |
 | [`browser.page.y`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | int | Click y (vertical) coordinates (in pixels) relative to the entire document. | `10` |
 | [`browser.tag_name`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | Target element tag name obtained via event.target.tagName. | `BUTTON` |
-| [`browser.element.attributes.<key>`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | Map of data-otel-* attributes from the target element. [1] | `testBtn`; `submit` |
+| [`browser.element.attributes.<key>`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | Map of data-otel-* attributes from the target element. [2] | `testBtn`; `submit` |
 | [`hw.mouse.button`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | User friendly name of the mouse button pressed. See [MouseEvent.buttons](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons#value). | `left` |
 
-**[1] `browser.element.attributes.<key>`:** Key is the attribute name after "data-otel-" and value is the attribute value. Example: `<button data-otel-id="testBtn" data-otel-role="submit">` SHOULD be recorded as `browser.element.attributes.id = "testBtn"` and `browser.element.attributes.role = "submit"`.
+**[1] `browser.css_selector`:** Value SHOULD be a CSS selector that uniquely identifies the target element within the document. Value SHOULD be escaped using CSS.escape() to avoid issues with special characters.
+
+**[2] `browser.element.attributes.<key>`:** Key is the attribute name after "data-otel-" and value is the attribute value. Example: `<button data-otel-id="testBtn" data-otel-role="submit">` SHOULD be recorded as `browser.element.attributes.id = "testBtn"` and `browser.element.attributes.role = "submit"`.
 
 ---
 
--- a/model/browser/registry.yaml
+++ b/model/browser/registry.yaml
@@ -80,6 +80,9 @@ groups:
         stability: development
         brief: 'Target element CSS Selector'
         examples: ["#main > div:nth-child(2) > button.submit"]
+        note: >
+          Value SHOULD be a CSS selector that uniquely identifies the target element within the document.
+          Value SHOULD be escaped using CSS.escape() to avoid issues with special characters.
       - id: hw.mouse.button
         type:
           members:
--- a/docs/registry/attributes/browser.md
+++ b/docs/registry/attributes/browser.md
@@ -12,25 +12,27 @@ The web browser attributes
 | Key | Stability | Value Type | Description | Example Values |
 |---|---|---|---|---|
 | <a id="browser-brands" href="#browser-brands">`browser.brands`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string[] | Array of brand name and version separated by a space [1] | `[" Not A;Brand 99", "Chromium 99", "Chrome 99"]` |
-| <a id="browser-css-selector" href="#browser-css-selector">`browser.css_selector`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | Target element CSS Selector | `#main > div:nth-child(2) > button.submit` |
-| <a id="browser-element-attributes" href="#browser-element-attributes">`browser.element.attributes.<key>`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | Map of data-otel-* attributes from the target element. [2] | `testBtn`; `submit` |
-| <a id="browser-language" href="#browser-language">`browser.language`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | Preferred language of the user using the browser [3] | `en`; `en-US`; `fr`; `fr-FR` |
-| <a id="browser-mobile" href="#browser-mobile">`browser.mobile`</a> | ![Development](https://img.shields.io/badge/-development-blue) | boolean | A boolean that is true if the browser is running on a mobile device [4] |  |
+| <a id="browser-css-selector" href="#browser-css-selector">`browser.css_selector`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | Target element CSS Selector [2] | `#main > div:nth-child(2) > button.submit` |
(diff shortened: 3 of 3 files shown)


model/browser/events.yaml
```diff
@@ -52,3 +52,47 @@ groups:
...
+        brief: 'Type of user interaction.'
+        note: >
+          Possible values inspired by
+          [ApplicationInsights Click Analytics Enums](https://github.com/microsoft/ApplicationInsights-JS/blob/main/extensions/applicationinsights-clickanalytics-js/src/Enums.ts).
+        examples: ['click.right']
+      - id: browser.xpath
+        type: string
+        stability: development
+        brief: 'Target element XPath using the XPathExpression API.'
+        examples: ["//*[@id='testBtn']"]
```
thompson-tomo: These should be ref's to the registry.

model/browser/registry.yaml
```diff
@@ -50,3 +50,40 @@ groups:
...
+      - id: browser.page.x
+        type: int
+        stability: development
+        brief: 'Click x (horizontal) coordinates (in pixels) relative to the entire document.'
+        examples: [10]
+      - id: browser.page.y
+        type: int
+        stability: development
+        brief: 'Click y (vertical) coordinates (in pixels) relative to the entire document.'
+        examples: [10]
```
thompson-tomo, line 62: Have we considered using https://opentelemetry.io/docs/specs/semconv/registry/attributes/app/#app-screen-coordinate-x etc

model/browser/registry.yaml
```diff
@@ -50,3 +50,40 @@ groups:
...
+          Arbitrary maps are not supported as attribute types in OTel, so this is stored as string.
+        examples: ['{"id": "123", "name": "Name"}']
+      - id: browser.user_action.type
+        type: string
+        stability: development
+        brief: 'Type of user interaction.'
+        note: >
+          Possible values inspired by
+          [ApplicationInsights Click Analytics Enums](https://github.com/microsoft/ApplicationInsights-JS/blob/main/extensions/applicationinsights-clickanalytics-js/src/Enums.ts).
+        examples: ['click.right']
```
thompson-tomo: Should be defined here in the yaml what the values should be

model/browser/registry.yaml
```diff
@@ -50,3 +50,40 @@ groups:
...
+        examples: ['BUTTON']
+      - id: browser.element.attributes
+        type: string
+        stability: development
+        brief: 'JSON-serialized map of data-otel-* attributes from the target element.'
+        note: >
+          Represented as a JSON object string. Example:
+          {"id": "123", "name": "Name"}.
+          Arbitrary maps are not supported as attribute types in OTel, so this is stored as string.
+        examples: ['{"id": "123", "name": "Name"}']
```
thompson-tomo: Why is this not a templated attribute to avoid the json.

model/browser/registry.yaml
```diff
@@ -50,3 +50,40 @@ groups:
...
+      - id: browser.page.y
+        type: int
+        stability: development
+        brief: 'Click y (vertical) coordinates (in pixels) relative to the entire document.'
+        examples: [10]
+      - id: browser.tag_name
+        type: string
+        stability: development
+        brief: 'Target element tag name obtained via event.target.tagName.'
+        examples: ['BUTTON']
```
thompson-tomo, line 67: Could we use https://opentelemetry.io/docs/specs/semconv/registry/attributes/app/#app-widget-name?

model/browser/registry.yaml
joaquin-diaz: I was trying to find something like that but I couldn't. Could you point me to an example? I am not sure what a templated attribute is. What I would use in code is a `map[string]string` since we actually don't know which attributes could be attached to the element

joaquin-diaz:
> This seems too generic and instead I feel we should aim for tailored click, key etc events. Ideally at a mimum following https://www.w3schools.com/jsref/obj_events.asp but also potentially splitting a couple ie mouse & click. That way we can easily tailor attributes based on the trigger.

Yeah.. I think having multiple events makes sense. It would make it easier to know which attributes to expect vs a bunch of mixed possible attributes depending on type 

> Also for click events Why not extend https://opentelemetry.io/docs/specs/semconv/app/app/#event-appwidgetclick but keep the same name?

I wasn't part of the original [discussion](https://github.com/open-telemetry/semantic-conventions/pull/1941#discussion_r2138377069) but it seems they preferred to use browser terminology instead of translating e.g. `event.pageX` -> `app.screen.coordinate.x`. 

Some of these `app` attributes as easy to translate from browser APIs but some are not as straight forward. `app.widget.id` and `app.widget.name` are hard to understand in a browsers perspective. Based on their definitions:

* app.widget.id: An identifier that uniquely differentiates this widget from other widgets in the same application. Examples: f9bc787d-ff05-48ad-90e1-fca1d46130b3; submit_order_1829
* app.widget.name: The name of an application widget. Examples submit; attack; Clear Cart

> A widget is an application component, typically an on-screen visual GUI element.

Component is a loaded term in web, it can be confused for example with a React component that can be composed of multiple dom elements. 

`app.widget.id` is marked as required in `app.widget.click` but there's no equivalent in the browser that is always present, you may have an id, but not always. You also have xpath but that can change if dom elements are added or removed. `app.widget.name` could be the element tag name

IMO I would like to have shared events between client SDKs (web and mobile) but they work for both cases, `app.widget.click` seems to be too tailored for mobile apps and it feels like a stretch to apply it here. I'll bring this up to the browser sig on Thursday and get their input

model/browser/registry.yaml
thompson-tomo: That is fine, see https://github.com/open-telemetry/semantic-conventions/blob/main/model/http/registry.yaml#L135 for an example of a templated attribute.

thompson-tomo:
> I wasn't part of the original discussion but it seems they preferred to use browser terminology instead of translating e.g. event.pageX -> app.screen.coordinate.x.

So I have seen at https://www.w3schools.com/jsref/obj_mouseevent.asp and page/app are in effect different measurements so perhaps we go with app.page.coordinate.x.

> Some of these app attributes as easy to translate from browser APIs but some are not as straight forward. app.widget.id and app.widget.name are hard to understand in a browsers perspective. 

I agree but I think we can improve the description to resolve those issues and use the attribute notes to define how to source it. For widget id = html element id and fall back to xpath when not defined.

docs/browser/browser-events.md
```diff
@@ -51,4 +51,45 @@ semantic convention tooling supports complex attributes
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
+<!-- semconv event.browser.user_action.click -->
```
thompson-tomo, line 56:
```suggestion
## Event: `browser.user_action.click`

<!-- semconv event.browser.user_action.click -->
```

model/browser/registry.yaml
```diff
@@ -50,3 +50,50 @@ groups:
...
+              value: 'middle'
+              brief: Middle button
+              stability: development
+            - id: right
+              value: 'right'
+              brief: Right button
+              stability: development
+        stability: development
+        brief: 'User friendly name of the mouse button pressed. See [MouseEvent.buttons](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons#value).'
+        examples: ["left"]
```
thompson-tomo: Perhaps we rename this to hw.mouse.button given it relates to the hardware?

joaquin-diaz: Makes sense! Good call

model/browser/registry.yaml
joaquin-diaz: Nice, updated. Thanks!

thompson-tomo:
Could someone from the browser sig or a maintainer add the triage accepted label so that the hw.mouse.button can be added to hardware namespace and not have this PR auto closed?

model/browser/registry.yaml
```diff
@@ -50,3 +50,51 @@ groups:
...
```
thompson-tomo, line 103: These should be in the registry file in the hardware folder. But only make the change once the triage:accepted label has been added.

docs/browser/browser-events.md
```diff
@@ -51,4 +51,47 @@ semantic convention tooling supports complex attributes
...
+This event describes a user click action in the browser.
+
+**Attributes:**
+
+| Key | Stability | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Value Type | Description | Example Values |
+|---|---|---|---|---|---|
+| [`browser.page.x`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | int | Click x (horizontal) coordinates (in pixels) relative to the entire document. | `10` |
+| [`browser.page.y`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | int | Click y (vertical) coordinates (in pixels) relative to the entire document. | `10` |
+| [`browser.tag_name`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | Target element tag name obtained via event.target.tagName. | `BUTTON` |
+| [`browser.xpath`](/docs/registry/attributes/browser.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | Target element XPath using the XPathExpression API. | `//*[@id='testBtn']` |
```
BenoitZugmeyer: **💬 suggestion:**  Have we considered using a CSS selector instead of xpath?

* CSS selectors are way more common than xpath. It would be easier to interpret for the user.
* CSS selectors are more performant than xpath when evaluating them in the UI
* CSS selectors allow to use [`CSS.escape`](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape_static) to easily escape attributes/ids/weird tag names. I'm not aware of a similar function for xpath, and implementing it in the SDK might be tedious.

joaquin-diaz: Updated, thanks for the suggestion

thompson-tomo:
@joaquin-diaz could you merge in main/rebase this branch followed by a regeneration of the docs?

scheler:
@joaquin-diaz @martinkuba the current span based instrumentation for click allows for linking the click to any xhr/fetch calls made from the click handler, which is helpful to know. The approach used there is to propagate trace context and use a common trace across click and all the xhr/fetch calls. This approach poses a problem that the same API calls are modeled different with and without RUM instrumentation, as these API calls get their own traceId when RUM is not involved. I prefer modeling them using independent traces. However, I do find it helpful to have the xhr/fetch calls linked to the click event. 

a) is there a span.links equivalent for events where we can add related spans, events, etc?
b) is it possible to identify the xhr/fetch related to a click event without using trace context?

thompson-tomo:
> a) is there a span.links equivalent for events where we can add related spans, events, etc?

An event can only be linked to 1 span but a span can have many events.

> b) is it possible to identify the xhr/fetch related to a click event without using trace context?

My suggestion is that we should have an app interaction span which acts as the root span. This root span could then contain the navigation, http etc child spans.

scheler:
@joaquin-diaz 
> My suggestion is that we should have an app interaction span which acts as the root span. This root span could then contain the navigation, http etc child spans.

This is what the current [user-interaction](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-user-interaction/src/instrumentation.ts) instrumentation does today; however, it has the issue I described above. The trace context ends up propagating to the backend instrumentation, whose behavior then varies depending on whether the user interaction instrumentation is set up or not, and this is not ideal.

thompson-tomo:
Are we not solving that by having the attributes on the root span for the span initiator hence removing the need for inference based on the parent span?

sandy2008:
Thanks, this PR is very useful context for my proposal in open-telemetry/semantic-conventions#3592.

I agree the right direction is not to introduce a generic new “user action” family, but to build from existing app events where possible.

For native mobile, my intent is:
- reuse existing click/tap-related app events as the basis where they already fit
- only propose additions for clear gaps such as scroll / gesture, lifecycle composition, and session correlation
- keep everything log-based and privacy-safe

So I’m treating open-telemetry/opentelemetry-specification#2992 as prior art rather than trying to redefine click from scratch.

PR #2994, opened 2025-10-28
GenAI: mark sampling-relevant attributes

lmolkova:
Mark the following attributes sampling-relevant on all GenAI spans they appear

* `gen_ai.operation.name`
* `gen_ai.provider.name`
* `gen_ai.request.model`
* `server.address`
* `server.port`

(also add `gen_ai.provider.name` to embeddings spans - it seems it has been forgotten)

### DIFF
--- a/docs/gen-ai/gen-ai-agent-spans.md
+++ b/docs/gen-ai/gen-ai-agent-spans.md
@@ -364,6 +364,7 @@ and SHOULD be provided **at span creation time** (if provided at all):
 
 * [`gen_ai.operation.name`](/docs/registry/attributes/gen-ai.md)
 * [`gen_ai.provider.name`](/docs/registry/attributes/gen-ai.md)
+* [`gen_ai.request.model`](/docs/registry/attributes/gen-ai.md)
 * [`server.address`](/docs/registry/attributes/server.md)
 * [`server.port`](/docs/registry/attributes/server.md)
 
--- a/model/gen-ai/spans.yaml
+++ b/model/gen-ai/spans.yaml
@@ -314,6 +314,8 @@ groups:
         sampling_relevant: true
       - ref: gen_ai.operation.name
         sampling_relevant: true
+      - ref: gen_ai.request.model
+        sampling_relevant: true
       - ref: gen_ai.agent.id
         requirement_level:
           conditionally_required: if applicable.


model/gen-ai/spans.yaml
```diff
@@ -119,6 +119,15 @@ groups:
       - ref: gen_ai.provider.name
         # TODO: Not adding to common attributes because of https://github.com/open-telemetry/weaver/issues/479
         requirement_level: required
+        sampling_relevant: true
+      - ref: gen_ai.operation.name
```
gyliu513, line 122: @lmolkova I’d also suggest marking the following LLM attributes as sampling_relevant:

- gen_ai.usage.input_tokens: this is a direct proxy for request cost and prompt size. Backends often want to preferentially sample:
  - very large prompts (potentially problematic or high-value requests),
  - or enforce higher sampling rates for expensive calls (billing / cost analysis, SLOs, abuse detection). Marking it as sampling_relevant makes it clear that collectors/backends can safely use it for tail-based / cost-aware sampling.

- gen_ai.usage.output_tokens: similarly, this is a proxy for response cost and response size. It’s especially useful for:
  - sampling unusually large responses (which may indicate misuse, prompt injection, or runaway generation),
  - and for reconstructing usage distributions from sampled data.
Including it as sampling_relevant helps backends implement “sample more when cost is high or output is anomalous” strategies.

- gen_ai.response.finish_reasons: finish reasons are very important for quality and error-focused sampling. Backends typically want to:
  - always sample traces where the model stopped due to error, max_tokens (truncation), or content_filtered,
  - use lower sampling rates for “normal” stop cases.
Marking this field as sampling_relevant makes it explicit that it’s safe and expected to drive policies like “always keep truncated or filtered generations” while down-sampling ordinary ones.

Together these three fields cover cost-aware sampling (usage.*tokens) and quality/error-aware sampling (finish_reasons), which are the two main dimensions GenAI backends care about when deciding which LLM traces to keep. Thoughts? Thanks!

lmolkova: non of these could be sampling-relevant - they are not available at start time

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -354,6 +363,14 @@ Since this attribute could be large, it's NOT RECOMMENDED to populate
 it by default. Instrumentations MAY provide a way to enable
 populating this attribute.
 
+The following attributes can be important for making sampling decisions
```
wrisa, line 362: Why `gen_ai.request.model` is missing ?

model/gen-ai/spans.yaml
```diff
@@ -287,6 +327,9 @@ groups:
       - ref: server.address
         requirement_level:
           recommended: when span kind is `CLIENT`.
+        sampling_relevant: true
+      - ref: server.port
+        sampling_relevant: true
```
wrisa, line 336: `gen_ai.request.model` missing ?

wrisa (commented):
Added a couple of comments. Rest LGTM.

docs/gen-ai/gen-ai-agent-spans.md
lmolkova: great catch, updated!

PR #2996, opened 2025-10-28
Add Pressure Stall Information (PSI) metrics

### LINKED ISSUE #2995: Add Pressure Stall Information (PSI) metrics
### Area(s)

area:system

### Propose new conventions

Linux Pressure Stall Information (PSI) metrics provide valuable insights into resource pressure on CPU, memory, and I/O subsystems, available on Linux systems with kernel 4.20+.

PSI includes 20 new metric instruments:
- **CPU**: `system.psi.cpu.some.*` (avg10, avg60, avg300, total)
- **Memory**: `system.psi.memory.some.*` and `system.psi.memory.full.*` (avg10, avg60, avg300, total for each)
- **I/O**: `system.psi.io.some.*` and `system.psi.io.full.*` (avg10, avg60, avg300, total for each)

Where "some" indicates some tasks are stalled and "full" indicates all tasks are stalled. The avg* metrics represent pressure averages over 10, 60, and 300 second windows, while total metrics track cumulative stall time in microseconds.

PSI metrics can be automatically collected alongside existing host metrics when running on Linux systems with PSI support. On non-Linux systems or when PSI is unavailable, the implementation gracefully degrades with no impact.

### Additional Context

- Linux PSI documentation: https://docs.kernel.org/accounting/psi.html
- PSI has been backported to some older LTS kernels
- Prometheus node_exporter already exposes these metrics: https://github.com/prometheus/node_exporter

### Open Telemetry integration details

There are issues on this matter in:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42779
- https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8082

And 2 PRs that I am proposing to address these issues:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43823
- https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8083

### Tip

alpineQ:
Closes #2995

## Changes

This PR adds support for Linux Pressure Stall Information (PSI) metrics to the system semantic conventions.

PSI is a Linux kernel feature (available since kernel 4.20) that identifies and quantifies resource contention by measuring the time impact that CPU, memory, and I/O resource crunches have on workloads.

### New Metrics

- **`system.linux.psi.pressure`** (Gauge): Measures resource pressure as a percentage of time that tasks were stalled over a time window (10s, 60s, or 300s)
- **`system.linux.psi.total_time`** (Counter): Tracks the total cumulative stall time in microseconds since system boot

### New Attributes

- **`system.psi.resource`**: The resource type (`cpu`, `memory`, `io`)
- **`system.psi.stall_type`**: The stall severity (`some` for partial stalls, `full` for complete stalls where all non-idle tasks are blocked)
- **`system.psi.window`**: The time window for pressure calculation (`10s`, `60s`, `300s`)

### Use Cases

PSI metrics enable:
- Sizing workloads to hardware or provisioning hardware according to workload demand
- Detecting productivity losses caused by resource scarcity
- Dynamic system management (load shedding, job migration, strategic pausing)
- Maximizing hardware utilization without sacrificing workload health

### References

- [Linux kernel PSI documentation](https://docs.kernel.org/accounting/psi.html)
- [/proc/pressure/* interface](https://man7.org/linux/man-pages/man5/proc.5.html)

### Relevant issues and PRs
There are issues on this matter in:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42779
- https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8082

And 2 PRs that I am proposing to address these issues:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43823
- https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8083

> [!IMPORTANT]

  * [Prometheus node exporter](https://github.com/prometheus/node_exporter?tab=readme-ov-file#enabled-by-default) has PSI metrics enabled by default

### DIFF
--- a/docs/registry/attributes/system.md
+++ b/docs/registry/attributes/system.md
@@ -138,7 +138,7 @@ Describes Linux Pressure Stall Information attributes
 |---|---|---|---|---|
 | <a id="system-psi-resource" href="#system-psi-resource">`system.psi.resource`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The resource experiencing pressure [2] | `cpu`; `memory`; `io` |
 | <a id="system-psi-stall-type" href="#system-psi-stall-type">`system.psi.stall_type`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The PSI stall type | `some`; `full` |
-| <a id="system-psi-window" href="#system-psi-window">`system.psi.window`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The time window over which pressure is calculated [3] | `10s`; `60s`; `300s` |
+| <a id="system-psi-window" href="#system-psi-window">`system.psi.window`</a> | ![Development](https://img.shields.io/badge/-development-blue) | int | The time window over which pressure is calculated in seconds. [3] | `10`; `60`; `300` |
 
 **[2] `system.psi.resource`:** Linux PSI (Pressure Stall Information) measures resource pressure for CPU, memory, and I/O. See [Linux kernel PSI documentation](https://docs.kernel.org/accounting/psi.html).
 
--- a/docs/system/system-metrics.md
+++ b/docs/system/system-metrics.md
@@ -1345,7 +1345,7 @@ See [Linux kernel PSI documentation](https://docs.kernel.org/accounting/psi.html
 |---|---|---|---|---|---|
 | [`system.psi.resource`](/docs/registry/attributes/system.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | The resource experiencing pressure [1] | `cpu`; `memory`; `io` |
 | [`system.psi.stall_type`](/docs/registry/attributes/system.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | The PSI stall type | `some`; `full` |
-| [`system.psi.window`](/docs/registry/attributes/system.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | The time window over which pressure is calculated [2] | `10s`; `60s`; `300s` |
+| [`system.psi.window`](/docs/registry/attributes/system.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | int | The time window over which pressure is calculated in seconds. [2] | `10`; `60`; `300` |
 
 **[1] `system.psi.resource`:** Linux PSI (Pressure Stall Information) measures resource pressure for CPU, memory, and I/O. See [Linux kernel PSI documentation](https://docs.kernel.org/accounting/psi.html).
 


model/system/metrics.yaml
```diff
@@ -563,3 +563,81 @@ groups:
...
+    brief: "Linux Pressure Stall Information (PSI) metric measuring resource contention as percentage of time."
+    note: |
+      PSI (Pressure Stall Information) identifies and quantifies resource contention.
+      The metric represents the percentage of time that tasks were stalled on a given resource
+      over the specified time window.
+
+      The "some" stall type indicates at least some tasks are stalled on the resource.
+      The "full" stall type indicates all non-idle tasks are stalled simultaneously, representing
+      a more severe state where the system is thrashing and CPU cycles are wasted.
+
```
thompson-tomo:
```suggestion
```

model/system/metrics.yaml
```diff
@@ -563,3 +563,81 @@ groups:
...
+    brief: "Linux Pressure Stall Information (PSI) total cumulative stall time."
+    note: |
+      This metric tracks the total absolute stall time in microseconds since system boot.
+      Unlike the percentage-based `system.linux.psi.pressure` metric, this allows detection
+      of latency spikes that wouldn't necessarily make a noticeable impact on time averages.
+      It also enables calculating average trends over custom time frames.
+
+      The "some" stall type indicates at least some tasks are stalled on the resource.
+      The "full" stall type indicates all non-idle tasks are stalled simultaneously.
+
```
thompson-tomo:
```suggestion
```

model/system/metrics.yaml
```diff
@@ -563,3 +563,81 @@ groups:
...
+      over the specified time window.
+
+      The "some" stall type indicates at least some tasks are stalled on the resource.
+      The "full" stall type indicates all non-idle tasks are stalled simultaneously, representing
+      a more severe state where the system is thrashing and CPU cycles are wasted.
+
+      PSI is available on Linux systems with kernel 4.20 or later and requires CONFIG_PSI=y.
+      CPU "full" stall is reported as zero at the system level for backward compatibility (available since 5.13).
+
+      Values are percentages in the range [0, 100]. The ratios are tracked over 10-second, 60-second,
```
thompson-tomo:
```suggestion
      The ratios are tracked over 10-second, 60-second,
```

model/system/metrics.yaml
```diff
@@ -563,3 +563,81 @@ groups:
...
+
+      PSI is available on Linux systems with kernel 4.20 or later and requires CONFIG_PSI=y.
+      CPU "full" stall is reported as zero at the system level for backward compatibility (available since 5.13).
+
+      This is a monotonically increasing counter that resets on system reboot.
+
+      See [Linux kernel PSI documentation](https://docs.kernel.org/accounting/psi.html) and
+      [/proc/pressure/*](https://man7.org/linux/man-pages/man5/proc.5.html) files.
+    instrument: counter
+    unit: "us"
```
thompson-tomo:
```suggestion
    unit: "s"
```

model/system/metrics.yaml
```diff
@@ -563,3 +563,81 @@ groups:
...
+        note: "Typically one of: 10s, 60s, or 300s"
+    entity_associations:
+      - host
+
+  - id: metric.system.linux.psi.total_time
+    type: metric
+    metric_name: system.linux.psi.total_time
+    annotations:
+      code_generation:
+        metric_value_type: int
```
thompson-tomo:
```suggestion
        metric_value_type: double
```

model/system/metrics.yaml
```diff
@@ -563,3 +563,81 @@ groups:
...
+    stability: development
+    brief: "Linux Pressure Stall Information (PSI) total cumulative stall time."
+    note: |
+      This metric tracks the total absolute stall time in microseconds since system boot.
```
thompson-tomo:
```suggestion
      This metric tracks the total absolute stall time in seconds since system boot.
```

model/system/metrics.yaml
```diff
@@ -563,3 +563,81 @@ groups:
...
+      The "full" stall type indicates all non-idle tasks are stalled simultaneously.
+
+      PSI is available on Linux systems with kernel 4.20 or later and requires CONFIG_PSI=y.
+      CPU "full" stall is reported as zero at the system level for backward compatibility (available since 5.13).
+
+      This is a monotonically increasing counter that resets on system reboot.
+
+      See [Linux kernel PSI documentation](https://docs.kernel.org/accounting/psi.html) and
+      [/proc/pressure/*](https://man7.org/linux/man-pages/man5/proc.5.html) files.
+    instrument: counter
```
thompson-tomo, line 622:
```suggestion
    instrument: updowncounter
```

braydonk: Documentation says it's cumulative/monotonic, so counter seems right.

model/system/metrics.yaml
braydonk: Documentation says it's in nanoseconds.

model/system/metrics.yaml
alpineQ: Agreed on seconds + double, that aligns with the OTel duration conventions.
Not sure I follow on the UpDownCounter change though. Total stall time should be monotonically increasing (it's cumulative), which is what Counter is for. UpDownCounter is meant for things that can go up and down. Am I missing something?

thompson-tomo: The counter will go down after a restart hence to me fits an up down counter as per https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments

braydonk: The counter going down after a restart is a counter reset. That doesn't make it an `updowncounter`. This should be reverted.

model/system/registry.yaml
```diff
@@ -143,3 +143,64 @@ groups:
...
+              note: >
+                The "full" line indicates the share of time in which all non-idle
+                tasks are stalled on a given resource simultaneously. This represents
+                a state where actual CPU cycles are going to waste and the workload
+                is thrashing. CPU full is undefined at the system level and is set to
+                zero for backward compatibility (available since Linux 5.13).
+        stability: development
+        brief: "The PSI stall type"
+        examples: ["some", "full"]
+        note: >
```
braydonk: This note might be a bit redundant now that the definition of `some` and `full` are in the enum member briefs, and the note doesn't contain any additional information.

model/system/metrics.yaml
alpineQ: Can you two sync up on this and let me know what the consensus is? Happy to make the change once we're aligned on the approach

model/system/metrics.yaml
```diff
@@ -563,3 +563,74 @@ groups:
...
+      This is a monotonically increasing counter that resets on system reboot.
+
+      See [Linux kernel PSI documentation](https://docs.kernel.org/accounting/psi.html) and
+      [/proc/pressure/*](https://man7.org/linux/man-pages/man5/proc.5.html) files.
+    instrument: updowncounter
+    unit: "s"
+    attributes:
+      - ref: system.psi.resource
+        requirement_level: required
+        brief: "The resource experiencing pressure (cpu, memory, or io)"
```
braydonk: This same overwritten brief is used in both spots that reference the attribute. That indicates to me that the brief of the attribute definition should just be this and it shouldn't be overwritten here.

In addition, the possible attribute values being replicated in the brief are probably unnecessary here because in the rendered markdown they'll be right next to this brief anyway.

model/system/metrics.yaml
```diff
@@ -563,3 +563,74 @@ groups:
...
+      of latency spikes that wouldn't necessarily make a noticeable impact on time averages.
+      It also enables calculating average trends over custom time frames.
+
+      PSI is available on Linux systems with kernel 4.20 or later and requires CONFIG_PSI=y.
+      CPU "full" stall is reported as zero at the system level for backward compatibility (available since 5.13).
+
+      This is a monotonically increasing counter that resets on system reboot.
+
+      See [Linux kernel PSI documentation](https://docs.kernel.org/accounting/psi.html) and
+      [/proc/pressure/*](https://man7.org/linux/man-pages/man5/proc.5.html) files.
```
braydonk: `/proc/pressure` isn't listed in this man page so it's kind of confusing to have the man page linked. The linked Kernel docs talk about the format of the `/proc/pressure` file already so I would suggest just removing the mention and link here.

model/system/metrics.yaml
```diff
@@ -563,3 +563,74 @@ groups:
...
+    instrument: updowncounter
+    unit: "s"
```
braydonk, line 623: If this is going to be instrumented as seconds instead, the note should clarify that it is reported by the kernel in microseconds and needs to be converted to a seconds and be reported as a double.

FWIW, I thought it was fine as microseconds. I think either is okay, you can decide. But if the decision is seconds, please add the clarification.

model/system/metrics.yaml
jsuereth: > The counter will go down after a restart hence to me fits an up down counter as per

Resets are allowed in counters, that's fundamental in their design.  If we need to clarify that in concepts and signals we can.

You should use `counter` and not `updowncounter` here.

model/system/metrics.yaml
alpineQ: I am all for leaving it in microseconds as given by kernel - less work on processing and converting. I was under the impression that open telemetry has some tendency of using seconds more often than not and it would be a good practice to convert it

braydonk: It is true that often these types of metrics are reported as `s` and `double`, but to my knowledge that's not because of a written rule. The reason we did it for some other metrics such as `system.cpu.time` is to allow for cross-platform unification (various platforms may report their times in different denominations so it makes sense to pick one).

In this case, this is a Linux exclusive metric and is on specific opt-in metrics, and generally when designing metrics like that in this namespace we like to match what a Linux user who knows what these metrics mean already would expect. If they were monitoring PSI directly, they'd see microseconds for these values, so our conventions for them might as well match that.

I don't think `s` and `double` is necessarily wrong either, but if it was purely up to me I would be team `us` and `int`.

model/system/metrics.yaml
braydonk: It's worth mentioning that [`system.cpu.time`](https://github.com/open-telemetry/semantic-conventions/blob/b2a742e1f8d6d001f180d12302d3ff3e4a4bbe95/model/system/metrics.yaml#L50) is a semantically similar metric also instrumented via cumulative monotonic data from `procfs`, and it is instrumented as a `counter` for the same reason as @jsuereth mentioned above.

As a member of the group who owns this area of the conventions, I would not approve this PR with this instrumented as an `updowncounter`.

model/system/metrics.yaml
trask: does this apply here?

https://github.com/open-telemetry/semantic-conventions/blob/b2a742e1f8d6d001f180d12302d3ff3e4a4bbe95/docs/general/metrics.md?plain=1#L116

braydonk: Ah thanks @trask, I didn't know this actually was written down anywhere. 

I'm unsure about the terminology `duration` here. This metric isn't counting a particular contiguous "duration of time" so to speak, rather it's a counter of microseconds spent in a pressured state over a longer period of time. I would consider `http.server.request.duration` a `duration` (regardless of that being in the name) because it's measuring a contiguous span of time of one object.

So I'd wonder if there could be some clarification on that rule. If it's to be expanded to mean "any metric that is measuring time should be `s`" I'm okay with that and we can follow it here (and anywhere else we may be inadvertently violating the rule because of my unawareness).

trask: I sort of suspect it applies to more than contiguous duration of time, but I'm not sure, here's the background on where the recommendation originally came from: https://github.com/open-telemetry/opentelemetry-specification/issues/2977

model/system/metrics.yaml
alpineQ: Counter it is :ok_hand:

model/system/metrics.yaml
braydonk: Thanks trask, we discussed in the System Semconv SIG Call Oct 30 and we came to the decision as a group to abide by this rule.

@alpineQ This can be changed back to `s` and `double`. Sorry for all the back and forth on it :bow:

alpineQ:
Is there anything else I can or should do to get this PR merged?

model/system/registry.yaml
```diff
@@ -155,3 +155,61 @@ groups:
...
+                is thrashing. CPU full is undefined at the system level and is set to
+                zero for backward compatibility (available since Linux 5.13).
+        stability: development
+        brief: "The PSI stall type"
+        examples: ["some", "full"]
+      - id: system.psi.window
+        type: string
+        stability: development
+        brief: "The time window over which pressure is calculated"
+        examples: ["10s", "60s", "300s"]
```
thompson-tomo:
```suggestion
        type: int
        stability: development
        brief: "The time window over which pressure is calculated in seconds."
        examples: [10, 60, 300]
```

alpineQ:
> It should also be confirmed if the metrics should be renamed to system.psi.linux.* to follow the changes made in #2984

Any updates on whether this should be changed?

braydonk:
> It should also be confirmed if the metrics should be renamed to system.psi.linux.* to follow the changes made in https://github.com/open-telemetry/semantic-conventions/pull/2984

This won't be necessary for `system.psi.*`. OS name in metric names is for subsystems where some metrics are shared across platforms (`system.memory.*` has some metrics that are shared and some that are Linux-exclusive). PSI is a Linux Kernel subsystem, so there will be no ambiguity.

model/system/registry.yaml
braydonk: I think this should stay a string duration specifier. People using query systems to specify rates will be used to windows specified in this string time duration format.

thompson-tomo:
> This won't be necessary for system.psi.*. OS name in metric names is for subsystems where some metrics are shared across platforms

@braydonk currently this is not being followed as the metrics are currently named `system.linux.psi.*` if os name is not needed then shouldn't these metrics be `system.psi.*` just like the  attributes.

alpineQ:
Wrong rebase order :see_no_evil: 
Fixed it in my fork. Please reopen

joaopgrassi:
There are still issues with the rebase, as I see changes on the `otel` namespace.

alpineQ:
"Files changed" tab in this PR no longer updates changes as PR was closed. Actual changes can be seen here: https://github.com/open-telemetry/semantic-conventions/compare/main...alpineQ:semantic-conventions:main

I can open a new PR if this will help with the issue

braydonk:
> OS name in metric names is for subsystems where some metrics are shared across platforms (system.memory.* has some metrics that are shared and some that are Linux-exclusive). PSI is a Linux Kernel subsystem, so there will be no ambiguity.

I was mixed up when I said this. `linux` should definitely not be after `psi`, but I forgot that according to our guidance the OS name should still be there. `system.linux` is the right order though; the point is that the OS name goes after the "area of concern" where different platforms have divergent offerings. In this case `system` is the divergence point, whereas all metrics withing `psi` are automatically all Linux.

> If os name is not needed then shouldn't these metrics be system.psi.* just like the attributes.

We should do the inverse of this; the attributes should be changed to `system.linux`. You can find our written guidance on this here. https://github.com/open-telemetry/semantic-conventions/blob/65f5c612261508f55800ab622a989349f29ba5ef/docs/non-normative/groups/system/design-philosophy.md?plain=1#L275

thompson-tomo:
Ah ok, then attributes should be renamed.

I am aware of that guidance which is what prompted my suggestion to rename as I was seeing psi as the area of concern just like memory. Might be worthwhile to update that doc to define what the area of concern and encorporate terminology about diverging.

braydonk:
I think that's what this last line is intending to say, perhaps could be worded better: https://github.com/open-telemetry/semantic-conventions/blob/65f5c612261508f55800ab622a989349f29ba5ef/docs/non-normative/groups/system/design-philosophy.md?plain=1#L304

thompson-tomo:
Have raised https://github.com/open-telemetry/semantic-conventions/issues/3067 to capture the improvement.

And yes the last line is trying to do that but a reader doesn't know what the area should be.

@alpineQ i would recommend opening a new as it is unlikely this PR can be re-opened due to removed/changed commits. Note the attributes should be renamed.

PR #2623, opened 2025-08-13
[chore] Show deprecated enum members in docs #2496

thompson-tomo:
Fixes #2496

## Changes

Have the documentation tables to now include deprecated members to assist with transition etc.

Note: if the PR is touching an area that is not listed in the [existing areas](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/README.md), or the area does not have sufficient [domain experts coverage](https://github.com/open-telemetry/semantic-conventions/blob/main/.github/CODEOWNERS), the PR might be tagged as [experts needed](https://github.com/open-telemetry/semantic-conventions/labels/experts%20needed) and move slowly until experts are identified.

## Merge requirement checklist

* [ ] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [ ] Links to the prototypes or existing instrumentations (when adding or changing conventions)

PR #3016, opened 2025-10-30
[chore] Automatically split attribute tables based on deprecation status

thompson-tomo:
## Changes

This splits the attribute tables based on the deprecation status.

This provides the following benefits:
- deprecated attributes no longer needs to be moved to seperate groups simplifying process and improving history
- The attributes on signals are grouped the same as the registry
- The deprecated tables can have a different structure with links to new attributes
- Reduction of items in table of contents.

Out of scope tasks:
- consolidating deprecated attributes in to main file
- Putting informative briefs back on attributes rather than duplicating deprecation reason.

> [!IMPORTANT]

### DIFF
--- a/docs/registry/attributes/android.md
+++ b/docs/registry/attributes/android.md
@@ -31,9 +31,7 @@ The Android platform on which the Android application is running.
 
 | Key | Value Type | Summary | Example Values | Deprecation Explanation |
 |---|---|---|---|---|
-| <a id="android-state" href="#android-state">`android.state`</a> | string | This attribute represents the state of the application. [1] | `created`; `background`; `foreground` |  Use `android.app.state` instead.  |
-
-**[1] `android.state`:** The Android lifecycle states are defined in [Activity lifecycle callbacks](https://developer.android.com/guide/components/activities/activity-lifecycle#lc), and from which the `OS identifiers` are derived.
+| <a id="android-state" href="#android-state">`android.state`</a> | string | Deprecated. Use `android.app.state` attribute instead. | `created`; `background`; `foreground` |  Use `android.app.state` instead.  |
 
 ---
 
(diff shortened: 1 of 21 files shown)


lmolkova (changes requested):
The previous PR (https://github.com/open-telemetry/semantic-conventions/pull/2819) was closed - there is no support for this change among approvers and you've got feedback that this is a controversial change. 

See https://github.com/open-telemetry/semantic-conventions/pull/2623 and https://github.com/open-telemetry/semantic-conventions/pull/2607#discussion_r2294199591 for the feedback points.

Lack of support from approvers along with the feedback usually means that there is no point in reopening the PR without some discussion. I'm going to close it.

thompson-tomo:
@lmolkova i had gone and addressed the controversial points which entered around polluting signals pages with deprecated attributes. This pr makes no changes to which attributes are rendered hence I saw that as closed. It also doesn't introduce a new concept. After which I did raise it at a semconv meeting the other week and the feedback I got was to avoid needing to consolidate yaml files which was also implemented.

As it stands the changes introduced here are just:
- render deprecated attributes under the same heading as the non-deprecated siblings but in a different section. This is mostly used in the registry.
- use a reduced/tailored table for deprecated attributes to remove un-necessary content and be able to link to new attribute where appropriate.

lmolkova:
thanks for the explanation. Not needing to separate deprecated from active will indeed be useful with schema v2 where we won't have groups anymore. Most of jinja work will need to be redone at that time anyway. 

I don't believe changes in markdown are beneficial though. And I'd postpone jinja work until we have a need (schema v2).

thompson-tomo:
Would it not be beneficial to extend this PR to also flatten the groups now hence simplifying the jinja changes and get users use to no Grouping. That way when switching to v2 the generated markdown would hopefully not need to change.

By having the different layout we can reduce the amount of content being produced and we could even remove the example values if desired to further simplify it.

PR #3030, opened 2025-11-04
[events] refine general event semantics documentation

pellared:
Towards stabilization of https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/events.md

Fixes https://github.com/open-telemetry/semantic-conventions/issues/1651

## Changes

This PR refines the documentation for general event semantics to improve clarity and structure.

### Documentation improvements
- Restructure event semantics section for better readability
- Consolidate event name definition with its uniquely identifying role
- Simplify cardinality and naming guidelines into single concise statement

### Requirement level clarifications
- **Attributes**: Strengthen from MAY to SHOULD to emphasize their importance for providing event context
- **Severity number**: Strengthen from MAY to SHOULD to encourage consistent severity tracking
- Remove explicit "body MAY be present" statement (still implied in first bullet) given unclear use cases. Body SHOULD NOT be used.

### Technical changes
- Consistent hyperlinking to all specification fields (event name, attributes, body, severity number)
- Align with RFC 2119 keyword usage patterns (use SHOULD instead of RECOMMENDED)
- Improve consistency with naming guidelines reference
- Enhanced external event compatibility guidance with explicit instrumentation SHOULD

## Rationale

The changes improve developer experience by:
1. Making requirements clearer and easier to scan
2. Emphasizing best practices (using attributes and severity numbers)
3. Reducing redundancy while maintaining completeness
4. Better aligning with the structure used in other semantic convention documents
5. Providing more concise and actionable guidance for convention authors


pellared:
I do not think this needs a changelog. Feel free to remove the label if you disagree.

docs/general/events.md
```diff
@@ -17,29 +17,26 @@ network connection severed, etc.
...
+  that uniquely identifies the event structure: its [attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-attributes)
+  and the type of the [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-body).
 
-It's RECOMMENDED to use the collection of attributes to represent the details of
-the event.
+* Event names SHOULD be of low cardinality and follow the [Naming guidelines](/docs/general/naming.md).
+
+* Attributes SHOULD be used to represent details and provide additional context about the event.
+
+* Events SHOULD specify a [severity number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-severitynumber).
```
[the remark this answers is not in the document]
pellared: Should we add something like the following


> * Events SHOULD NOT use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-body) and SHOULD use attributes to provide all event details.


so that we address https://github.com/open-telemetry/semantic-conventions/issues/1651?

Even if so, then I propose to do it in a separate PR to avoid bikeshedding.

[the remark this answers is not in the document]
lmolkova: I'd prefer to just keep this line in this PR

> It's RECOMMENDED to use the collection of attributes to represent the details of
the event.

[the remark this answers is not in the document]
pellared: Addressed https://github.com/open-telemetry/semantic-conventions/pull/3030/commits/45991864fef8c79c7d2955357238753867ae74a0

docs/general/events.md
```diff
@@ -17,29 +17,27 @@ network connection severed, etc.
...
 
-It's RECOMMENDED to use the collection of attributes to represent the details of
-the event.
+* Event names SHOULD be of low cardinality and follow the [Naming guidelines](/docs/general/naming.md).
 
-<!-- Body use cases are not clear - see https://github.com/open-telemetry/semantic-conventions/issues/1651 for the context.-->
+* Attributes SHOULD be used to represent details and provide additional context about the event.
+
+* Events SHOULD NOT use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-body)
+  and SHOULD use attributes to provide all event details.
```
lmolkova:
```suggestion
```

It's a repetition of the sentence above, no?

> Attributes SHOULD be used to represent details and provide additional context about the event.

docs/general/events.md
```diff
@@ -17,29 +17,26 @@ network connection severed, etc.
 
 In OpenTelemetry, Events are implemented as a specific type of [`LogRecord`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/api.md) that conforms to the conventions included here.
 
-OpenTelemetry Semantic Conventions that define events SHOULD document the event name along
-with attributes and the type of the body if any.
+Semantic conventions that define events SHOULD document the event name, its attributes, and the type of the body (if any).
```
cijothomas: Why is it just `SHOULD` and not `MUST`? If event name and schema is not documented, its not really a useful OTel event right?

docs/general/events.md
```diff
@@ -17,29 +17,26 @@ network connection severed, etc.
...
+  and the type of the [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-body).
 
-It's RECOMMENDED to use the collection of attributes to represent the details of
-the event.
+* Event names SHOULD be of low cardinality and follow the [Naming guidelines](/docs/general/naming.md).
 
-<!-- Body use cases are not clear - see https://github.com/open-telemetry/semantic-conventions/issues/1651 for the context.-->
+* Attributes SHOULD be used to represent details and provide additional context about the event.
+
+* Events SHOULD NOT use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-body).
```
cijothomas: mm..not sure why we need to discourage body? Sorry I might be missing some context around this...

pellared: The current suggestion is to use attribute(s). Even an attribute with the same name as the event name is nice becasue we can dogfood the semantic conventions "toolchain"+registry+weaver could be use to define those events. This has been discussed during a few (dozen?) of various SIG meetings. @lmolkova has probably more insights into this.

Related https://github.com/open-telemetry/semantic-conventions/pull/3030#discussion_r2491098538

docs/general/events.md
pellared: I agree

pellared:
```suggestion
Semantic conventions that define events MUST document the event name, its attributes, and the type of the body (if any).
```

docs/general/events.md
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/batch_log_processor.rs#L216-L219
This is an example of OTel Rust's internal log/event. It's an event with no attributes, but have a body. I am not sure if there is anything wrong with that. Is the recommendation to put what we are currently putting in body to an attribute instead?

pellared: Yup, `BatchLogProcessor.Emit.AfterShutdown` attribute instead of using body.
_However, I think the semconv would rather propose a name/attribute key like `log.processor.batch.emit_after_shutdown`._

@lmolkova, am I correct?

lmolkova: it's interesting! 

It's both a log (with human-readable text) and an event (has name). Unless you use a message (body) you don't get a good experience when writing it as a traditional log intended for humans. 

I think what we wanted is to discourage using bodies in semantic conventions, there is nothing wrong with using body for log message in user apps or libraries.

We avoided being too restrictive with prev line 

>  Attributes SHOULD be used to represent details and provide additional context about the event.

If we want more clarity we can say that semconv authors SHOULD NOT use bodies. It  would still allow log->event and similar exceptions to exist in semconv if they come.

docs/general/events.md
```diff
@@ -17,29 +17,26 @@ network connection severed, etc.
...
-* The event name uniquely identifies event structure: the set of attributes and
-  the type of the body.
-* Events MAY have [severity number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-severitynumber).
+* An event MUST have an [event name](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-eventname)
+  that uniquely identifies the event structure: its [attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-attributes)
+  and the type of the [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-body).
 
-It's RECOMMENDED to use the collection of attributes to represent the details of
-the event.
+* Event names SHOULD be of low cardinality and follow the [Naming guidelines](/docs/general/naming.md).
```
trask: I'm not sure what low cardinality means in terms of event name

docs/general/events.md
cijothomas: > BatchLogProcessor.Emit.AfterShutdown attribute instead of using body.

mm..That is just duplicating the EventName into an attribute! Not sure why we recommend that.

docs/general/events.md
pellared: Mme too. It comes from the original document. I suggest changing

```suggestion
* Event names SHOULD follow the [Naming guidelines](/docs/general/naming.md).
```

Please leave a :+1: if you agree with the suggestion.

docs/general/events.md
cijothomas: > I think what we wanted is to discourage using bodies in semantic conventions, there is nothing wrong with using body for log message in user apps or libraries.

If we take the example from OTel Rust internal log/event below, assuming some time in the future, we'll have semantic convention for these things, what is wrong with using body there?

```
otel_warn!(
                    name: "BatchLogProcessor.Emit.AfterShutdown",
                    message = "Logs are being emitted even after Shutdown. This indicates incorrect lifecycle management of OTelLoggerProvider in application. Logs will not be exported."
                );
```

pellared: > If we want more clarity we can say that semconv authors SHOULD NOT use bodies.

and

> An event MUST have an [event name](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-eventname)
  that uniquely identifies the event structure: its [attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-attributes)
  **and the type of the [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-body).**

Given the type of the body should be specified is not already a kind of [semantic] convention? According to this rule we do not want different instrumentation libraries to create different body structure for the same event name.

Also I am not sure if the events need to have always "exactly" same attributes (some may be optional, other could be recommended and finally some are going to be required).

Maybe we just need

> An event MUST have an [event name](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/logs/data-model.md#field-eventname)
  that uniquely identifies the event type.

And change 

> Semantic conventions that define events MUST document the event name, its attributes, and the type of the body (if any).

into

> Semantic conventions that define events MUST document the event name and its attributes, and SHOULD document severity numbers.

Then Body is just some additional metadata and the event.name + attributes resembles the semantic conventions for spans (span name + attributes) and metrics (metric/instrument name + attributes).

lmolkova: discussed in the Logs SIG
- let's remove body from this doc (scoped to semconv only). it's ok for non-semconv events to have body and there might be exceptions  inside semconv
- semconv are opinionated and we don't recommend using body  
- spec is not opinionated - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-eventname, body type is still identified by the event name and is allowed

The example above is an interesting mix of log and event: the `name` and `message` are somewhat duplicative. They are the same info intended for two different consumers (machines and humans). In theory, the message could be localized and/or populated by a very smart processor/backend that knows about human-friendly descriptions based on the event name.

pellared: Addressed comments brought up during the SIG meeting: https://github.com/open-telemetry/semantic-conventions/pull/3030/commits/c52b3a86f4163e12b5b4c39b6aa6a7f835cc2a2c

@lmolkova, @trask, PTAL

PR #3034, opened 2025-11-05
[otel] add otel.event.name attribute

### LINKED ISSUE #2913: Stabilize `otel.event.name` attribute
Used by appenders and processors for mapping from bridges to OTel and nothing else (does not go over the wire).

Prototypes needed for stabilization:
- [x] Log4j - https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/16220
- [x] Logback - https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/16220
- [x] Go (log processor) - https://pkg.go.dev/go.opentelemetry.io/otel/sdk/log#example-Processor-EventName

Behavior:
- bridge sets it as `EventName` top-level and does not set the attribute
- OR a log processor sets the top-level `EventName` 

This allows to decouple new thing vs `event.name` which has different meaning (goes over the wire, has prior instrumentations and behavior).

We should also re-deprecate `event.name`.
EDIT (pellared): Done in https://github.com/open-telemetry/semantic-conventions/pull/2914

pellared:
Fixes #2913

This PR introduces the `otel.event.name` attribute for use by non-OTLP exporters and logging libraries.

The `otel.event.name` attribute can be used by:
- Non-OTLP exporters to emit the `EventName` when the destination does not support `EventName` or an equivalent field
- Applications using existing logging libraries to add event name information that can be used to set the `EventName` field by Collector or SDK components

This follows a similar pattern to #2715, but uses the `otel.*` namespace instead of `event.*` to properly represent OpenTelemetry-specific concepts used for bridging between formats.

Example from OTel Go:
- https://github.com/open-telemetry/opentelemetry-go/pull/7568

### DIFF
--- a/docs/registry/attributes/otel.md
+++ b/docs/registry/attributes/otel.md
@@ -5,6 +5,7 @@
 
 - [OTel Attributes](#otel-attributes)
 - [OTel Component Attributes](#otel-component-attributes)
+- [OTel Event Attributes](#otel-event-attributes)
 - [OTel Scope Attributes](#otel-scope-attributes)
 - [Deprecated OTel Library Attributes](#deprecated-otel-library-attributes)
 
@@ -100,6 +101,18 @@ E.g. for Java the fully qualified classname SHOULD be used in this case.
 | `simple_log_processor` | The builtin SDK simple log record processor | ![Development](https://img.shields.io/badge/-development-blue) |
 | `simple_span_processor` | The builtin SDK simple span processor | ![Development](https://img.shields.io/badge/-development-blue) |
 | `zipkin_http_span_exporter` | Zipkin span exporter over HTTP | ![Development](https://img.shields.io/badge/-development-blue) |
+
+## OTel Event Attributes
+
+Attributes used by non-OTLP exporters to represent OpenTelemetry Event's concepts.
+
+**Attributes:**
+
+| Key | Stability | Value Type | Description | Example Values |
+|---|---|---|---|---|
+| <a id="otel-event-name" href="#otel-event-name">`otel.event.name`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | Identifies the class / type of event. [3] | `browser.mouse.click`; `device.app.lifecycle` |
+
+**[3] `otel.event.name`:** This attribute SHOULD be used by non-OTLP exporters when destination does not support `EventName` or equivalent field. This attribute MAY be used by applications using existing logging libraries so that it can be used to set the `EventName` field by Collector or SDK components.
 
 ## OTel Scope Attributes
 
--- a/model/otel/registry.yaml
+++ b/model/otel/registry.yaml
@@ -76,6 +76,23 @@ groups:
         brief: The schema URL of the instrumentation scope.
         examples: ['https://opentelemetry.io/schemas/1.31.0']
         stability: development
+  - id: registry.otel.event
+    type: attribute_group
+    display_name: OTel Event Attributes
+    brief: Attributes used by non-OTLP exporters to represent OpenTelemetry Event's concepts.
+    attributes:
+      - id: otel.event.name
+        type: string
+        stability: development
+        brief: >
+          Identifies the class / type of event.
+        note: >
+          This attribute SHOULD be used by non-OTLP exporters
+          when destination does not support `EventName` or equivalent field.
+          This attribute MAY be used by applications using existing logging
+          libraries so that it can be used to set the `EventName` field by
+          Collector or SDK components.
+        examples: ["browser.mouse.click", "device.app.lifecycle"]
   - id: registry.otel.component
     type: attribute_group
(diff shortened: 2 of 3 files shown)


model/otel/registry.yaml
```diff
@@ -76,6 +76,23 @@ groups:
         brief: The schema URL of the instrumentation scope.
         examples: ['https://opentelemetry.io/schemas/1.31.0']
         stability: development
+  - id: registry.otel.event
+    type: attribute_group
+    display_name: OTel Event Attributes
+    brief: Attributes used by non-OTLP exporters to represent OpenTelemetry Event's concepts.
+    attributes:
+      - id: otel.event.name
```
pellared, line 84: I started to think about other "log record" fields that may not fit some non-OTLP exporters. Maybe this should be `otel.log.event_name` so that we can later add e.g. `otel.log.severity_number`?

@open-telemetry/specs-logs-approvers, thoughts?

docs/registry/attributes/otel.md
```diff
@@ -101,6 +102,18 @@ E.g. for Java the fully qualified classname SHOULD be used in this case.
...
+
+Attributes used by non-OTLP exporters to represent OpenTelemetry Event's concepts.
+
+**Attributes:**
+
+| Key | Stability | Value Type | Description | Example Values |
+|---|---|---|---|---|
+| <a id="otel-event-name" href="#otel-event-name">`otel.event.name`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | Identifies the class / type of event. [3] | `browser.mouse.click`; `device.app.lifecycle` |
+
+**[3] `otel.event.name`:** This attribute SHOULD be used by non-OTLP exporters when destination does not support `EventName` or equivalent field. This attribute MAY be used by applications using existing logging libraries so that it can be used to set the `EventName` field by Collector or SDK components.
```
cijothomas, line 115: Question on the second part: If a user does logging with a 3rd party logging library, and sets an attribute with `otel.event.name`, will that get stored in OTel Sdk LogRecord as the top-level "EventName" automatically? Or it requires additional LogRecordProcessor/similar to do that promotion?

model/otel/registry.yaml
cijothomas: Do we have such conventions for Spans? i.e is there any recommended attribute name to store top level things in Span, when using non-otlp?

pellared: Yes, we do https://github.com/open-telemetry/semantic-conventions/blob/1540bf89cf53ad6c13ed5e1f4008ee548eff0560/model/otel/registry.yaml#L2-L41

Maybe we should put the attribute into this registry instead of creating a new one?
@open-telemetry/specs-semconv-maintainers, do you have recommendations??

cijothomas: interesting. Wondering why only a subset of span's top level field have such attributes defined? For eg:, we don't seem to have anything for Span.Kind?

pellared: > Wondering why only a subset of span's top level field have such attributes defined? For eg:, we don't seem to have anything for Span.Kind?

Probably things are just being added ad-hoc depending on need (open-source) :wink:

cijothomas: Do we have a need for otel.event.name attribute for non-otlp? Any example scenario(s) which would benefit from this?

(I am supportive of adding `otel.event.name` , but the only use-case I had in my mind was to let 3rd party logging library emit OTel Events, even when they don't have a top-level concept called EventName)

docs/registry/attributes/otel.md
pellared: > will that get stored in OTel Sdk LogRecord as the top-level "EventName" automatically? Or it requires additional LogRecordProcessor/similar to do that promotion?

I would leave it to the language to decide what is better for them. Personally, I think that doing it via `LogRecordProcessor` is "better code reuse" (no need to apply the same logic in each logging bridge) and also gives more flexibility (you can decide for which loggers/libraries you apply this logic).

model/otel/registry.yaml
pellared: E.g. https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/splunkhecexporter would benefit from it.

lmolkova: I'd rather let `otel.event.name` coexist with `otel.log.severity.number` than use `otel.log.event.name` - `otel.event.name` is shorter but is as descriptive. 

Also if we consider log and event to be synonyms, this works from pedantic perspective as well.

lmolkova (approved):

docs/registry/attributes/otel.md
trask: Unless the log record processor can be added automatically somehow, I'd prefer the mapping to be done in the logging library instrumentation so users don't need to do anything extra. But I don't think the spec is opinionated one way or the other today.

Issue #2585, opened 2025-08-01
Builtin (server side) tool calls


alexmojaki:
### Area(s)

area:gen-ai

### What's missing?

As discussed in https://github.com/open-telemetry/semantic-conventions/pull/2179#discussion_r2243144052, we need to decide how to represent tools that are part of a model API instead of functions called by the client, e.g. the OpenAI code interpreter.

### Describe the solution you'd like

In general I think these should be represented similarly to function tool calls for simplicity and consistency. For example, OpenAI returns an output part like this for the code interpreter:

```python
    {
        'id': 'ci_687fc324daac819ea0e2974e28da08e90031b95cdb2aa2fb',
        'code': 'import random\n'
        '\n'
        '# Generate a random number\n'
        'first_random_number = random.randint(1, 100)\n'
        'first_random_number',
        'container_id': 'cntr_687fc324201c8191a8332730996f1f8301bf4203e99b0b23',
        'outputs': [{'logs': '21', 'type': 'logs'}],
        'status': 'completed',
        'type': 'code_interpreter_call',
    },
```

which I think should be represented by these parts:

```python
    {
        'type': 'tool_call',
        'name': 'code_interpreter_call',
        'id': 'ci_687fc324daac819ea0e2974e28da08e90031b95cdb2aa2fb',
        'arguments': {
            'code': 'import random\n\n# Generate a random number\nfirst_random_number = random.randint(1, 100)\nfirst_random_number',
            'container_id': 'cntr_687fc324201c8191a8332730996f1f8301bf4203e99b0b23',
        },
    },
    {
        'type': 'tool_call_response',
        'id': 'ci_687fc324daac819ea0e2974e28da08e90031b95cdb2aa2fb',
        'response': {
            'outputs': [{'logs': '21', 'type': 'logs'}],
            'status': 'completed',
        },
    },
```

Optionally there could be an extra boolean to indicate that this is a server-side call.

In Gemini, there's already two parts in the output:

```json
          {
            "executableCode": {
              "language": "PYTHON",
              "code": "import random\n\nrandom_number = random.randint(1, 100)\nprint(f\"The random number generated is: {random_number}\")\n"
            }
          },
          {
            "codeExecutionResult": {
              "outcome": "OUTCOME_OK",
              "output": "The random number generated is: 6\n"
            }
          },
```

Note that there's nothing to use here as a tool call ID, so sticking with that format may mean making that field optional, or generating a random ID.

[...]

PR #3038, opened 2025-11-05
Built-in Tools Support

singankit:
Fixes # [2585](https://github.com/open-telemetry/semantic-conventions/issues/2585)

Document for Responses API with built-in tools proposal : [Comparison for built in tools](https://github.com/singankit/semantic_convention_proposal/blob/main/responses-builtin-tools/comparison.md)

[Prototype](https://github.com/singankit/opentelemetry-python-contrib/blob/e9d5cec2682f6f17d56562a16a9cbc60c451ff5a/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py#L1107) 
[OpenAI Agents Sample](https://github.com/singankit/opentelemetry-python-contrib/blob/users/singankit/responses-api-instrumentation/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/examples/zero-code/responses_api_with_agent.py) 

## Changes

> [!IMPORTANT]


singankit:
[https://github.com/open-telemetry/semantic-conventions/issues/2585](url)

docs/gen-ai/non-normative/examples-llm-calls.md
```diff
@@ -715,11 +725,64 @@ sequenceDiagram
...
-TODO
+```json
+[
+  {
+    "role": "assistant",
+    "parts": [
+      {
+        "type": "tool_call",
+        "id": "call_VSPygqKTWdrhaFErNvMV18Yl",
+        "name": "code_interpreter",
```
lmolkova, line 758: I think we need to document the schema, not just the example, so that we know how to record different types of tools. 

In this proposal shape tool call part would have to accommodate file search, code_interpreter and everything else without clear idea on what the format of each of them and which property is used as a discriminator.

singankit: @lmolkova Are you suggesting having a discriminator that can distinguish which tool is called and based on the tool call schema can be inferred. For example, in below example , tool is code_interpreter and "code_interpreter" property is present only when `tool_type` is code_interpreter.
```json
# Code Interpreter tool call part
{
    "type": "tool_call"
    "id": "id",
    "name": "code_interpreter",
    "tool_type": "code_interpreter", # discriminator
    "code_interpreter": {
        "arguments": {
          "code": "import random\n\n# Generate a random number\nrandom_number = random.randint(1, 100)\n\n# Execute some operation with the random number (e.g., squaring it)\nresult = random_number ** 2\n\nrandom_number, result",
          "container_id": "cntr_690bdbfed8688190884efd4c7ae6435b0db1f006442e8941",
        }
     }
}
```

For Function Tool Call

```json
{
    "role": "assistant",
    "parts": [
      {
        "type": "tool_call",
        "id": "call_VSPygqKTWdrhaFErNvMV18Yl",
        "tool_type": "function",
       "function": {
             "name": "get_weather",
             "arguments": {
                  "location": "Paris"
             }
        },
       
      }
    ],
    "finish_reason": "tool_call"
  }
```

lmolkova: Yeah, I'm suggesting to add a formal json schema that specifies how to record built-in tools and use some discriminator to separate them.

singankit: Have updated the PR including schema for code interpreter and making tool call request/response polymorphic to allow any tool to be supported (just need tool call/response schema to be added)

DylanRussell: Is it worthwhile to maintain a formal schema for each built in tool ? 

Gemini has 6 or 7 built in tools. This would require us to maintain 6 or 7 schemas here for the response.. Each model has it's own set of built-in tools that probably overlap a little but not completely, and I think it'll be kind of a pain to converge on a shared schema for each of them..

Or can we just use a generic catch-all `Any` type  instead of doing that?

Also for Gemini there is no ToolCallRequest for built-in tools. Instead built-in tools are enabled, and you can only tell whether the model used them by checking what the model returned in the response.

singankit: Thanks @DylanRussell for the feedback. Having `Any` for any unknown tools makes sense to me. This proposal supports including tool schemas for different tools even provider specific tools if needed. Having ToolCallRequest Part provided information about tool being called. However, if it captures the actual arguments tool is called can probably be optional.

docs/gen-ai/gen-ai-input-messages.json
```diff
@@ -118,6 +118,118 @@
...
+                    "default": null,
+                    "description": "Container identifier for code execution environment.",
+                    "title": "Container Id"
+                }
+            },
+            "required": ["type", "code"],
+            "title": "CodeInterpreterToolCall",
+            "type": "object"
+        },
+        "CodeInterpreterToolCallResponse": {
```
lmolkova: let's also update the python code

docs/gen-ai/non-normative/examples-llm-calls.md
```diff
@@ -715,11 +736,68 @@ sequenceDiagram
...
+        "tool_call": {
+          "type": "code_interpreter",
```
lmolkova, line 760: should this be provider-specific or generic? are code_interpreter tools are similar ?

docs/gen-ai/gen-ai-input-messages.json
```diff
@@ -276,15 +432,23 @@
                     "title": "Name",
                     "type": "string"
                 },
-                "arguments": {
-                    "default": null,
-                    "description": "Arguments for the tool call.",
-                    "title": "Arguments"
+                "tool_call": {
```
aabmass: What do you think of flattening the changes into a new part type instead of extending the `type: "tool_call"` Part here?

singankit:
**Feedback from Dec 16th Meeting : **
Way to make progress:
- Let's compare openai, google, anthropic 
- Option1: one part (request + response)
- Option2: two parts
- Option 2.5: two parts and extend Function calls

--------------------------------------------------------------------------------------------------------------------------------------------
Based on feedback I have added details and a comparison table in this [doc](https://github.com/singankit/semantic_convention_proposal/blob/main/responses-builtin-tools/comparison.md) . It compares OpenAI, Gemini and Claude built-in tools for tool call and tool response representation. It has detailed information about some of the popular built in tools.

**Based on information in this doc, I am recommending representing tool call and tool response for built-in tools as separate parts like it is done for client tools (already with `tool_call_part` and `tool_response_part`). **

@lmolkova @alexmojaki  @aabmass Please review and share feedback.

docs/gen-ai/gen-ai-input-messages.json
singankit: @aabmass I have put down different options [here](https://github.com/singankit/semantic_convention_proposal/blob/main/responses-builtin-tools/server_tool_representation.md) on how to represent built-in tools. Option 2 seems like a good way if to separate built-in tools from client tools. Please share your feedback.

docs/gen-ai/gen-ai-output-messages.json
```diff
@@ -49,6 +49,39 @@
             "title": "BlobPart",
             "type": "object"
         },
+        "CodeInterpreterToolCall": {
+            "additionalProperties": false,
```
lmolkova: why false?

docs/gen-ai/gen-ai-output-messages.json
```diff
@@ -49,6 +49,39 @@
...
+            "description": "Code interpreter tool call.",
+            "properties": {
+                "type": {
+                    "const": "code_interpreter",
+                    "description": "Type identifier for code interpreter calls.",
+                    "title": "Type",
+                    "type": "string"
+                },
+                "code": {
+                    "description": "The Python code to execute.",
```
lmolkova:
```suggestion
                    "description": "The code to execute.",
```

docs/gen-ai/gen-ai-output-messages.json
```diff
@@ -110,6 +143,31 @@
             "title": "FinishReason",
             "type": "string"
         },
+        "FunctionToolCall": {
+            "additionalProperties": false,
```
lmolkova: why false?

docs/gen-ai/gen-ai-output-messages.json
```diff
@@ -110,6 +143,31 @@
...
+            "description": "Function tool call.",
+            "properties": {
+                "type": {
+                    "const": "function",
+                    "description": "Type identifier for function calls.",
+                    "title": "Type",
+                    "type": "string"
+                },
+                "name": {
```
lmolkova: name is already included as a require property on ToolCallRequestPart. I think it's generic enough to be kept as a top level property, just not required

docs/gen-ai/gen-ai-output-messages.json
```diff
@@ -302,15 +360,23 @@
...
+                "tool_call": {
+                    "description": "Polymorphic tool call details with type discriminator.",
+                    "oneOf": [
+                        {
+                            "$ref": "#/$defs/FunctionToolCall"
+                        },
+                        {
+                            "$ref": "#/$defs/CodeInterpreterToolCall"
+                        }
+                    ],
```
lmolkova: we should introduce generic tool call as well

docs/gen-ai/non-normative/examples-llm-calls.md
```diff
@@ -673,9 +684,19 @@ Span:
 > The format of `gen_ai.output.messages` is not yet specified for built-in tool calls (check [#2585](https://github.com/open-telemetry/semantic-conventions/issues/2585)
 > for the details).
 
-This is an example of telemetry generated for a chat completion call with `code_interpreter` tool that results in
+This is an example of telemetry generated for a responses call with `code_interpreter` tool that results in
```
lmolkova:
```suggestion
This is an example of telemetry generated for OpenAI Responses API call with `code_interpreter` tool that results in
```

docs/gen-ai/gen-ai-input-messages.json
lmolkova: why is it defined in input messages only? we should have it on both

docs/gen-ai/gen-ai-input-messages.json
```diff
@@ -118,6 +118,118 @@
...
+                    "description": "Container identifier for code execution environment.",
+                    "title": "Container Id"
+                }
+            },
+            "required": ["type", "code"],
+            "title": "CodeInterpreterToolCall",
+            "type": "object"
+        },
+        "CodeInterpreterToolCallResponse": {
+            "additionalProperties": false,
```
lmolkova: why false?

DylanRussell:
Regarding the comparison: https://github.com/singankit/semantic_convention_proposal/blob/main/responses-builtin-tools/comparison.md

I would not describe Function Calling for gemini as a "built in tool", because the model is not the one that calls the functions, either the user's application calls it, or the SDK calls it (on the client side).. Same with MCP..

Otherwise I agree with your comparison table for gemini, there is no Tool Call object, only a Tool Call response object that we can glean from the request/response.. We can probably also determine whether a server-side tool was Enabled -- is there a way we can capture that.. ?

docs/gen-ai/non-normative/examples-llm-calls.md
singankit: @DylanRussell Have updated the proposal to include a generic way to capture tool call/response as a catch all. Please review.

docs/gen-ai/gen-ai-input-messages.json
singankit: Updated to include the new schemas that were added as part of this proposal. Please review.

docs/gen-ai/non-normative/examples-llm-calls.md
singankit: As discussed in SIG call on 1/20/2026 this is out of scope for this PR. However, in future can have a PR in an attempt to generically represent common server tools available across providers.

docs/gen-ai/gen-ai-input-messages.json
singankit: As discussed in SIG call on 1/20/2026 we agreed on new `server_tool_call` and `server_tool_call_response` part to capture built in tools. Making it polymorphic to capture different built-in tools from different providers.

docs/gen-ai/gen-ai-output-messages.json
singankit: This part has is now out of scope for this PR hence resolving this comment.

docs/gen-ai/gen-ai-output-messages.json
singankit: This part has is now out of scope for this PR hence resolving this comment.

docs/gen-ai/gen-ai-output-messages.json
singankit: This part has is now out of scope for this PR hence resolving this comment.

docs/gen-ai/gen-ai-output-messages.json
singankit: As discussed in SIG call on 1/20/2026 this is out of scope for this PR. I have update the PR to remove this part. Hence resolving the comment.

docs/gen-ai/non-normative/examples-llm-calls.md
singankit: Thanks for the suggestion have incorporated it.

docs/gen-ai/gen-ai-input-messages.json
singankit: Thanks for catching it. Have included it in input and output messages.

docs/gen-ai/gen-ai-input-messages.json
singankit: As discussed in SIG call on 1/20/2026 this is out of scope for this PR. Have updated the PR to remove it. Hence resolving it.

singankit:
> Regarding the comparison: https://github.com/singankit/semantic_convention_proposal/blob/main/responses-builtin-tools/comparison.md
> 
> I would not describe Function Calling for gemini as a "built in tool", because the model is not the one that calls the functions, either the user's application calls it, or the SDK calls it (on the client side).. Same with MCP..
> 
> Otherwise I agree with your comparison table for gemini, there is no Tool Call object, only a Tool Call response object that we can glean from the request/response.. We can probably also determine whether a server-side tool was Enabled -- is there a way we can capture that.. ?

Agree function calls are not built in tools. I have added that to all the other providers too for a fair comparison (In case you see a discrepancy please do flag it to me). 

For Server-side tool being Enabled, would `tool_definitions` under `invoke agent` be a better place for it? Since if a tool is disabled, we will not see any tool call/response messages anyway.

docs/gen-ai/gen-ai-output-messages.json
singankit: Yes I did add generic tool call thanks to @DylanRussell  comment too. Please review

DylanRussell:
> For Server-side tool being Enabled, would tool_definitions under invoke agent be a better place for it? Since if a tool is disabled, we will not see any tool call/response messages anyway.

Where is `tool_definitions` ?  I don't see that sem conv anywhere but maybe i'm looking in the wrong place.. Yes if a tool is disabled we will not see it in the response, but the reverse is not true (if a tool is enabled we may also not see it, but it still could be interesting to see that it was enabled)..

singankit:
> > For Server-side tool being Enabled, would tool_definitions under invoke agent be a better place for it? Since if a tool is disabled, we will not see any tool call/response messages anyway.
> 
> Where is `tool_definitions` ? I don't see that sem conv anywhere but maybe i'm looking in the wrong place.. Yes if a tool is disabled we will not see it in the response, but the reverse is not true (if a tool is enabled we may also not see it, but it still could be interesting to see that it was enabled)..

@DylanRussell It can be found [here](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-agent-spans/#invoke-agent-span), property is [gen_ai.tool.definitions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/)

DylanRussell:
I see, yeah that makes sense. It's not part of the input / output messages,  so it makes sense to be in a different attribute. This LGTM

docs/gen-ai/non-normative/models.ipynb
```diff
@@ -45,141 +45,7 @@
...
... and properties.\n    This allows for extensibility with custom message part types.\n    \"\"\"\n    type: str = Field(description=\"The type of the content captured in this part.\")\n\n    class Config:\n        extra = \"allow\"\n\n# This Union without discriminator will generate anyOf in JSON schema\nMessagePart = Union[\n    TextPart,\n    ToolCallRequestPart,\n    ToolCallResponsePart,\n    ServerToolCallPart,\n    ServerToolCallResponsePart,\n    BlobPart,\n    FilePart,\n    UriPart,\n    ReasoningPart,\n    GenericPart,  # Catch-all for any other type\n    # Add other message part types here as needed,\n    # e.g. structured output, hosted tool call, etc.\n]\n\nclass Role(StrEnum):\n    SYSTEM = \"system\"\n    USER = \"user\"\n    ASSISTANT = \"assistant\"\n    TOOL = \"tool\"\n\nclass ChatMessage(BaseModel):\n    role: Union[Role, str] = Field(\n        description=\"Role of the entity that created the message.\")\n    parts: List[MessagePart] = Field(\n        description=\"List of message parts that make up the message content.\")\n    name: Optional[str] = Field(default=None, description=\"The name of the participant.\")\n\n    class Config:\n        extra = \"allow\""
```
DylanRussell: Should we remove " or a built-in tool call outcome and details." from the description of `ToolCallResponsePart` -- since we now have `ServerToolCallResponsePart`

DylanRussell (approved):

docs/gen-ai/non-normative/examples-llm-calls.md
singankit: Marking this conversation as resolved, since the feedback provided on this thread has been incorporated.

docs/gen-ai/gen-ai-output-messages.json
singankit: Marking this conversation as resolved since the feedback has been incorporated in the PR

docs/gen-ai/non-normative/models.ipynb
singankit: Thanks @DylanRussell for the feedback, have updated the description please review.

docs/gen-ai/gen-ai-input-messages.json
singankit: @aabmass Please let me know if your feedback is addressed? If yes, I would like to mark this conversation as resolved.

docs/gen-ai/non-normative/examples-llm-calls.md
```diff
@@ -688,38 +695,95 @@ sequenceDiagram
...
-TODO
+```json
+[
+  {
+    "role": "assistant",
+    "parts": [
+      {
+        "type": "server_tool_call",
+        "id": "call_VSPygqKTWdrhaFErNvMV18Yl",
+        "name": "code_interpreter",
```
aabmass, line 758: Do we plan to normalize these names across providers? It almost feels like the `name` here should match with `server_tool_call.type`

aabmass (approved):
Apologies for slowness getting back to this PR. It LGTM.

Do you plan to follow up on schematizing some common server side tools, like code interpreter? Is there any issue for it?

docs/gen-ai/non-normative/examples-llm-calls.md
singankit: It would be great to normalize built-on tools names across providers. However, it is out of scope for this PR. 

I agree right now most of built-on tools name matches type. However, there are cases where type and name are different for example in [Claude Code Execution Tool](https://platform.claude.com/docs/en/agents-and-tools/tool-use/code-execution-tool#response-format) it supports bash, text editor etc. `name` will be `code_execution` but `server_tool_call.type` can be `bash_execution` , `text_editor_execution` etc.

Please share your feedback.

singankit:
> Apologies for slowness getting back to this PR. It LGTM.
> 
> Do you plan to follow up on schematizing some common server side tools, like code interpreter? Is there any issue for it?

Thanks @aabmass for the feedback. Created this [issue ](https://github.com/open-telemetry/semantic-conventions-genai/issues/77)to capture this feedback.

PR #3043, opened 2025-11-07
Add in code attributes to exception event #1981

### LINKED ISSUE #1981: Add opt-in code attributes to exception event
See - https://github.com/open-telemetry/semantic-conventions/pull/1942, we can generalize code attributes as opt-in on all exception events.

thompson-tomo:
Fixes #1981

## Changes

Adds in the code attributes to the exception event as opt-in

> [!IMPORTANT]


thompson-tomo:
@joaopgrassi can this be reopened. The issue has been accepted but pr still got closed.

model/exceptions/events.yaml
```diff
@@ -14,3 +14,11 @@ groups:
...
       - ref: exception.stacktrace
       - ref: exception.escaped
+      - ref: code.function.name
+        requirement_level: opt_in
+      - ref: code.file.path
+        requirement_level: opt_in
+      - ref: code.line.number
+        requirement_level: opt_in
+      - ref: code.column.number
+        requirement_level: opt_in
```
trask, line 24: will this be the code where the exception was originally thrown? or the user code where it was captured?

cc @open-telemetry/browser-approvers since related to #1942

thompson-tomo: My preference would be the code which threw the exception if the language provides it.

trask:
are there any prototypes / usages of this in instrumentation under the open-telemetry org?

model/exceptions/events.yaml
thompson-tomo: Note my thought matches what would be the case for javascript https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent

thompson-tomo:
I had thought the javascript were using them based on the closed pr but it was taken out. 

As such I have extended the existing [otel contrib lib](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-web-exception/src/instrumentation.ts) to map additional properties from [ErrorEvent](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent) to otel attributes as seen in this [commit](https://github.com/thompson-tomo/opentelemetry-js-contrib/commit/7a0a567437549704723f31f14c52c8c388cc99dc)

overbalance:
@thompson-tomo Is anything else needed here? cc @bidetofevil

pellared:
@thompson-tomo, this most likely will be superseded by https://github.com/open-telemetry/semantic-conventions/pull/3256

thompson-tomo:
@pellared I am aware span events are being deprecated but wouldn't this also be the definition used by the log record event? And as such shouldn't #3256 leave the definition in place and just change the span document to point to the logs page.

PR #3048, opened 2025-11-08
[chore] switch to uaparser.dev for example parsing

### LINKED ISSUE #2909: [editorial] whatsmyua.info has expired certificate, consider whatismybrowser.com alternative
What do you think about this alternative? The website seems more actively maintained:

- https://www.whatismybrowser.com/detect/what-is-my-user-agent/

The certificate of https://www.whatsmyua.info expired over 100 days ago:

```console
$ curl -sS https://www.whatsmyua.info
curl: (60) SSL certificate problem: certificate has expired
```

It's used as an example of getting UA info:

```console
find . -type f -exec grep --color=auto -nH --null -e 'whatsmyua\.info' \{\} +
./docs/registry/attributes/user-agent.md:19:**[1] `user_agent.name`:** [Example](https://www.whatsmyua.info) of extracting browser's name from\
 original string. In the case of using a user-agent for non-browser products, such as microservices with multiple names/versions inside the `u\
ser_agent.original`, the most significant name SHOULD be selected. In such a scenario it shoul[…]
./docs/registry/attributes/user-agent.md:21:**[2] `user_agent.version`:** [Example](https://www.whatsmyua.info) of extracting browser's versio\
n from original string. In the case of using a user-agent for non-browser products, such as microservices with multiple names/versions inside \
the `user_agent.original`, the most significant version SHOULD be selected. In such a scenario[…]
./model/user-agent/registry.yaml:23:          [Example](https://www.whatsmyua.info) of extracting browser's name from original string. In the \
case of using
./model/user-agent/registry.yaml:34:          [Example](https://www.whatsmyua.info) of extracting browser's version from original string. In t\
he case of
```

thompson-tomo:
Fixes #2909
Progresses #3040

## Changes

The link for an example of decoding ua string now points to the js library site.

> [!IMPORTANT]


thompson-tomo:
@joaopgrassi could you also re-open this one?

Should user-agent also be a part of the browser sig?

Note one thing I have noticed is that the re-opened pr's when they were closed via the automation they are not appearing on the triage board.

joaopgrassi:
What about not having a link at all for this? Why do we really need it?

thompson-tomo:
I did think about that but I feel that the link brings additional information as it can also be used to validate the parsing. Ie put in the original ua string and the site will show you the values of OS, browser etc which are also seperate semconv attributes.

arminru (approved):

arminru:
@chalin FYI - this addresses your https://github.com/open-telemetry/semantic-conventions/issues/2909

joaopgrassi:
GOing to merge because this is blocking other PRs.

chalin:
Nice, thanks!

PR #3057, opened 2025-11-10
[chore] align areas with codeowners

trisch-me:
I have noticed that codeowners file has additional areas that are not mentioned in the areas file. 
Fixed it

@joaopgrassi FYI


areas.yaml
```diff
@@ -143,6 +148,10 @@ areas:
       - area:client
       - area:server
       - area:network
+      - area:error
+      - area:aspnetcore
+      - area:dotnet
+      - area:signalr
```
trask: I think I'd prefer to reduce coverage for HTTP in the CODEOWNERS file instead of adding these areas here

trisch-me: @trask could you please update it and propose changes there? I will reflect them here

trask: #3099

arminru:
```suggestion
```
as per https://github.com/open-telemetry/semantic-conventions/pull/3099

trisch-me: updated!

Issue #2827, opened 2025-09-29
Add HTTP span events for network phases breakdown


surbhiia:
### Area(s)

area:http

### What's missing?

Currently, OpenTelemetry **Client HTTP** instrumentations (such as [OkHttp3](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/okhttp/okhttp-3.0/library)) do not capture granular network phase timing, missing critical timed events like **DNS resolution** start/end, **TLS handshake** start/end, and **TCP connection** establishment start/end timestamps. When captured as [span events]( https://github.com/open-telemetry/opentelemetry-specification/blob/0e1094949a0cf57dd82c48fdb2b8e9e80cf3220e/specification/trace/api.md#add-events), this detailed timing data would enable observability backends to identify specific performance bottlenecks (DNS latency, TLS negotiation delays, connection issues) within individual HTTP requests and implement their own metric aggregation strategies (P95 DNS time, TLS success rates etc) tailored to their analytical needs. OpenTelemetry's [existing metric signals ](https://github.com/open-telemetry/semantic-conventions/blob/5c7d07a147a0116be50000119c19bbd8ccaef9ed/docs/http/http-metrics.md#http-client)cannot address these requirements due to their aggregated nature and inability to preserve request-level timing correlation.

### Describe the solution you'd like

Capture the relevant events as span events. 

Proposing event names (as per [naming guidelines](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md)) to be captured as span events for client HTTP spans (Span name and timestamp matters the most, no additional attributes needed as span attributes are enough)

| Event Name | Description | Timing Phase | Required |
|------------|-------------|--------------|----------|
| http.dns.start | DNS resolution begins | Network Setup | Optional |
| http.dns.end | DNS resolution completes | Network Setup | Optional |
| http.connection.start | TCP connection establishment begins | Connection Phase | Optional |
| http.connection.end | TCP connection establishment completes | Connection Phase | Optional |
| http.tls.start | TLS handshake begins (HTTPS only) | Security Phase | Optional |
| http.tls.end | TLS handshake completes (HTTPS only) | Security Phase | Optional |
| http.request.header.start | Request headers transmission begins | Request Phase | Optional |
| http.request.header.end | Request headers transmission completes | Request Phase | Optional |
| http.request.body.start | Request body upload begins | Request Phase | Optional |
[...]

PR #3069, opened 2025-11-11
Add Browser Resource Timing Event

hectorhdzg:
This is to continue discussion in original PR #1943, thanks @Karlie-777 for your efforts, I will try to add comments for open questions/issues here.

## Changes

Adds Browser Resource Timing events to the Semantic Conventions


hectorhdzg:
@martinkuba updated url.full to be part of body field and updated description, related to
https://github.com/open-telemetry/semantic-conventions/pull/1943/files#r2108097641

hectorhdzg:
Open questions are related to having query string separately, let me know your thoughts @joaquin-diaz @overbalance @trask 
https://github.com/open-telemetry/semantic-conventions/pull/1943/files#r2416951311

docs/browser/events.md
```diff
@@ -0,0 +1,55 @@
...
+| `responseStart` | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | double | The [time](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) immediately after the browser receives the first byte of the response from the server, cache, or local resource, in milliseconds. See [responseStart](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/responseStart). | `508.6000000014901` |
+| `secureConnectionStart` | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | double | The [time](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) immediately before the browser starts the handshake process to secure the current connection, in milliseconds. See [secureConnectionStart](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/secureConnectionStart). | `476.6000000014903` |
+| `url` | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | The full URL of the resource being fetched. [1] | `https://www.example.com/path`; `//localhost` |
```
thompson-tomo: Why don't we make this `file.path` and have it as an attribute? If we don't create a dedicated span

thompson-tomo (commented):
We need to consider the outcome of https://github.com/open-telemetry/semantic-conventions/issues/2827#issuecomment-3412305237 and ensure that what is done for resources is consistent with http etc.

Ie do we just use the http span or do we need a new file.download.client span? I suspect we might need to stick to the http span if it is too difficult for instrumentation to determine which span definition needs to be used.

model/browser/events.yaml
```diff
@@ -52,3 +52,114 @@ groups:
...
+      It is recommended to be captured by using the PerformanceResourceTiming API.
+      If the page starts unloading before the page loads,
+      then the partial data can be captured by calling the [performance.getEntriesByType](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType) method.
+    body:
+      id: browser.resource_timing
+      requirement_level: required
+      stability: development
+      type: map
+      fields:
+        - id: url
```
overbalance: In Karlie's previous PR we discussed dropping querystring parameters (from what would be `url.full` in HTTP) to reduce cardinality. I would like to see a new name for `origin` + `pathname`, which is how this value will likely be constructed in javascript. This combo does not exist in the [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL) or current [OTel semconv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/).

I propose **`url.canonical`** or **`url.resource`**, in that order. My first choice is [Canonical](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel#canonical) because it's already a well understood term on the web.

`url.query` could be added as an additional, optional attribute for those that want it.

docs/browser/events.md
overbalance: Not all resources are files. ResourceTiming also applies to fetch, xhr, and beacon.

thompson-tomo: For fetch/xhr/Beacon why aren't we using the standard http client span? We can use the pr linked in my other comment to handle capturing the timing details.

model/browser/events.yaml
```diff
@@ -52,3 +52,114 @@ groups:
...
+          type: string
+          stability: development
+          brief: The full URL of the resource being fetched.
+          requirement_level: required
+          examples: ["https://www.example.com/path", "//localhost"]
+          note: >
+            This is the URL of the resource being measured, not the page URL.
+            Generally the url doesn't include any query strings or any leading credentials,
+            and it's formatted as scheme://full domain/path only, for example https://user:pwd@www.example.com/path?a=b&c=d
```
thompson-tomo: Examples should not be showing the capturing of username &/or passwords as attribute values.

docs/browser/events.md
```diff
@@ -0,0 +1,55 @@
...
+
+The event name MUST be `browser.resource_timing`.
+
+This event describes the timing metrics as provided by PerformanceResourceTiming Performance API. These metrics are related to fetching a resource, such as XMLHttpRequest, Fetch, sendBeacon APIs, SVG, image or script.
+
+This event captures data from the [ResourceTiming](https://www.w3.org/TR/resource-timing/). It is recommended to be captured by using the PerformanceResourceTiming API. If the page starts unloading before the page loads, then the partial data can be captured by calling the [performance.getEntriesByType](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType) method.
+
+**Body fields:**
+
+:warning: Body fields will be moved to complex attributes once the
```
joaquin-diaz: Aren't we ready to move to complex attributes now? Not saying that these shouldn't be body attributes but if that's the case we should remove this warning, or move them to attributes

hectorhdzg: Updated to complex attributes

model/browser/events.yaml
hectorhdzg: Added both canonical URL and url query

model/browser/events.yaml
```diff
@@ -52,3 +52,50 @@ groups:
...
+          The canonical URL includes the scheme (if present), full domain, and path only.
+          Query strings and credentials MUST be excluded to reduce cardinality and avoid capturing PII.
+          For example, `https://www.example.com/path?a=b&c=d` would be captured as `https://www.example.com/path`.
+          In JavaScript, this typically corresponds to `${url.origin}${url.pathname}` or `${location.protocol}//${location.host}${location.pathname}`.
+      - ref: browser.resource_timing.url.query
+        requirement_level: opt_in
+        note: >
+          This attribute is opt-in due to potential high cardinality and PII concerns.
+          Sensitive content provided in the query string SHOULD be scrubbed when instrumentations can identify it.
+          Use this attribute when query parameters are relevant to the resource being fetched (e.g., image dimensions, format specifications).
```
overbalance: I pitched `url.canonical` to the Browser SIG today and agreed it needs more discussion. For now, please revert the name to `url` and drop `url.query` altogether.

overbalance (changes requested):

overbalance (approved):

model/browser/events.yaml
hectorhdzg: Removed canonical and query and added back full URL, still under discussion and could be addressed at a later time

hectorhdzg:
@joaquin-diaz @thompson-tomo let me know if there is any more feedback for this PR

overbalance:
Not stale 🤖

hectorhdzg:
Not seeing any option to re open this :(

lmolkova:
@open-telemetry/browser-approvers if this PR is ready and there is a consensus in the SIG, could you please approve it? It'd be a good signal for @open-telemetry/specs-semconv-approvers to also take a look. thanks!

surbhiia:
I have a similar PR for one of the http client instrumentations [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/15664). I used `http.<phase>.start_time` and `http.<phase>.end_time` as attribute names. For example - `http.connect.start_time`. (refer `NetworkTimingEventListener.java` file in the PR). Should we try to come up with common naming for browser and mobile? 

I plan to propose the same in my exiting [issue](https://github.com/open-telemetry/semantic-conventions/issues/2827) and would bring up the same in semantic convention WG on Mon - Feb2 - 8-9am. 

I would also be proposing replication of the original HTTP span attributes on this log record as an interim solution so backends which can't correlate two different signals for metric generation (if metric generation happens at ingestion time) can use this standalone log record until they're equipped to correlate two signals.

overbalance:
Not stale 🤖 cc @martinkuba

hectorhdzg:
Not stale

hectorhdzg:
@open-telemetry/browser-approvers can you take a look? as far as I know there is not any outstanding question or concern

model/browser/registry.yaml
```diff
@@ -50,3 +50,94 @@ groups:
...
           `navigator.language`.
         examples: ["en", "en-US", "fr", "fr-FR"]
+
+  - id: registry.browser.resource_timing
+    type: attribute_group
+    display_name: Browser Resource Timing Attributes
+    brief: >
+      Attributes for browser resource timing events
+    attributes:
+      - id: browser.resource_timing.url
```
overbalance: This is already covered by `url.full` and can be removed

model/browser/registry.yaml
```diff
@@ -50,3 +50,94 @@ groups:
...
+          Measured in milliseconds as a DOMHighResTimeStamp relative to the time origin.
+          See the [Performance Resource Timing API](https://www.w3.org/TR/resource-timing-2/#dom-performanceresourcetiming-connectstart).
+        examples: [131.0, 261.3]
+      - id: browser.resource_timing.secure_connection_start
+        type: double
+        stability: development
+        brief: 'Time immediately before the browser starts the handshake process to secure the connection'
+        note: >
+          Measured in milliseconds as a DOMHighResTimeStamp relative to the time origin.
+          If the scheme is not HTTPS, this value is 0.
```
overbalance: I see the spec says the default value is 0 but I prefer to leave this key out if the page is insecure. What do you think?

model/browser/events.yaml
```diff
@@ -52,3 +52,37 @@ groups:
           brief: A unique ID representing this particular metric instance.
           requirement_level: required
           examples: ["v3-1677874579383-6381583661209"]
+  - id: event.browser.resource_timing
+    stability: development
+    type: event
+    name: browser.resource_timing
+    brief: >
+      This event describes the timing metrics as provided by PerformanceResourceTiming Performance API.
```
overbalance:
```suggestion
      This event describes the timing metrics provided by the PerformanceResourceTiming interface of the Resource Timing API.
```

model/browser/registry.yaml
```diff
@@ -50,3 +50,94 @@ groups:
...
+      - id: browser.resource_timing.url
+        type: string
+        stability: development
+        brief: 'Absolute URL describing a network resource according to [RFC3986](https://www.rfc-editor.org/rfc/rfc3986)'
+        note: >
+          Generally the url doesn't include any query strings or any leading credentials,
+          and it's formatted as scheme://full domain/path only, for example
+          `https://www.example.com/path?a=b&c=d` would be sent as
+          `https://www.example.com/path` by both removing any potential PII and
+          reducing the cardinality of the name.
```
overbalance:
```suggestion
          reducing the cardinality of the url.
```

model/browser/registry.yaml
```diff
@@ -50,3 +50,94 @@ groups:
...
+    display_name: Browser Resource Timing Attributes
+    brief: >
+      Attributes for browser resource timing events
+    attributes:
+      - id: browser.resource_timing.url
+        type: string
+        stability: development
+        brief: 'Absolute URL describing a network resource according to [RFC3986](https://www.rfc-editor.org/rfc/rfc3986)'
+        note: >
+          Generally the url doesn't include any query strings or any leading credentials,
```
overbalance: nit: the specs use the term `userinfo`
```suggestion
          Generally the url doesn't include any query strings or any userinfo,
```

model/browser/events.yaml
```diff
@@ -52,3 +52,37 @@ groups:
...
+      These metrics are related to fetching a resource, such as
+      XMLHttpRequest, Fetch, sendBeacon APIs, SVG, image or script.
```
overbalance:
```suggestion
      XMLHttpRequest, Fetch, and sendBeacon, as well as SVG, image, and script elements.
```

model/browser/events.yaml
```diff
@@ -52,3 +52,37 @@ groups:
...
+    note: >
+      This event captures data from the [ResourceTiming](https://www.w3.org/TR/resource-timing/).
+      It is recommended to be captured by using the PerformanceResourceTiming API.
+      If the page starts unloading before the page loads,
+      then the partial data can be captured by calling the [performance.getEntriesByType](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType) method.
```
overbalance:
```suggestion
      Resources not fully loaded at collection time will have incomplete timing data.
   Use PerformanceObserver or [performance.getEntriesByType](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType) to collect entries before page unload.
```

model/browser/events.yaml
```diff
@@ -52,3 +52,37 @@ groups:
...
+    stability: development
+    type: event
+    name: browser.resource_timing
+    brief: >
+      This event describes the timing metrics as provided by PerformanceResourceTiming Performance API.
+      These metrics are related to fetching a resource, such as
+      XMLHttpRequest, Fetch, sendBeacon APIs, SVG, image or script.
+    note: >
+      This event captures data from the [ResourceTiming](https://www.w3.org/TR/resource-timing/).
+      It is recommended to be captured by using the PerformanceResourceTiming API.
```
overbalance: This looks similar enough to the surrounding lines that it could be dropped entirely
```suggestion
```

overbalance (changes requested):
I noticed the w3c urls are now forwarding from `https://www.w3.org/TR/resource-timing-2/` to `https://www.w3.org/TR/resource-timing/` and can be updated. I'll post this up to the SIG slack so others can review again too since this has been in the queue so long.

model/browser/registry.yaml
hectorhdzg: It make sense to me, this not required field so updating the comment

model/browser/registry.yaml
hectorhdzg: Removed url from here

overbalance (approved):
Awesome, thank you!

hectorhdzg:
Sorry about the mess, don't leave AI to do your git commands :) trying to fix now, I may need to create another PR

hectorhdzg:
thank you very much @trask  for the help

joaquin-diaz (approved):

david-luna (approved):

martinkuba:
There is an alternative proposal to use unified HTTP semantics [here](https://github.com/open-telemetry/semantic-conventions/issues/3385), but this is still under discussion. Let's leave this open until the decision is finalized.

lmolkova:
@martinkuba @hectorhdzg should we close this for the time being or is there an active work going on to consolidate with https://github.com/open-telemetry/semantic-conventions/issues/3385 ? thanks!

martinkuba:
@lmolkova We have actually released an instrumentation with these conventions. But the work on the unified HTTP semantics is still ongoing, and will likely replace them. I think we can close this for now and reopen it if something changes.

lmolkova:
@martinkuba @hectorhdzg if this captures existing conventions in instrumentations, it makes sense to document them here. The #3385 doesn't look like something that would materialize in the next few months, so unless you folks are actively working on the next version or #3385, it would make sense to finish and merge this PR.

In any case, feel free to close the PR if you're not going to pursue it.

Issue #3073, opened 2025-11-13
K8's metric tables are not rendering as expected with content split over multiple lines


thompson-tomo:
### Area(s)

area:k8s

### What's the problem?

When i look at k8s metrics the content is not being rendered correctly, An example of this is  https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/k8s-metrics.md#metric-k8sresourcequotaobject_countused & https://opentelemetry.io/docs/specs/semconv/system/k8s-metrics/#metric-k8sresourcequotaobject_countused 

When I turned off markdownlint suppression in the sample the below errors were produced:

> docs/system/k8s-metrics.md:2219:14 MD034/no-bare-urls Bare URL used [Context: "https://kubernetes.io/docs/ref..."]
docs/system/k8s-metrics.md:2241:14 MD034/no-bare-urls Bare URL used [Context: "https://kubernetes.io/docs/ref..."]
docs/system/k8s-metrics.md:2263:14 MD034/no-bare-urls Bare URL used [Context: "https://kubernetes.io/docs/ref..."]
docs/system/k8s-metrics.md:2285:14 MD034/no-bare-urls Bare URL used [Context: "https://kubernetes.io/docs/ref..."]
docs/system/k8s-metrics.md:2307:14 MD034/no-bare-urls Bare URL used [Context: "https://kubernetes.io/docs/ref..."]
docs/system/k8s-metrics.md:2329:14 MD034/no-bare-urls Bare URL used [Context: "https://kubernetes.io/docs/ref..."]
docs/system/k8s-metrics.md:2351:14 MD034/no-bare-urls Bare URL used [Context: "https://kubernetes.io/docs/ref..."]
docs/system/k8s-metrics.md:2373:14 MD034/no-bare-urls Bare URL used [Context: "https://kubernetes.io/docs/ref..."]
docs/system/k8s-metrics.md:2446:104 MD055/table-pipe-style Table pipe style [Expected: leading_and_trailing; Actual: leading_only; Missing trailing pipe]
docs/system/k8s-metrics.md:2446:104 MD056/table-column-count Table column count [Expected: 6; Actual: 4; Too few cells, row will be missing data]
docs/system/k8s-metrics.md:2447:1 MD055/table-pipe-style Table pipe style [Expected: leading_and_trailing; Actual: trailing_only; Missing leading pipe]
[...]

PR #3013, opened 2025-10-30
[chore] #189 standardise list symbol

thompson-tomo:
Progresses #189

## Changes

This ensures that lists use - as opposed to * as per prettier recommendations.

Note all pages which have a toc also require - to be list symbol.

> [!IMPORTANT]
> Pull requests acceptance are subject to the triage process as described in [Issue and PR Triage Management](https://github.com/open-telemetry/semantic-conventions/blob/main/issue-management.md).
> PRs that do not follow the guidance above, may be automatically rejected and closed.

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [ ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [ ] Links to the prototypes or existing instrumentations (when adding or changing conventions)

Issue #189, opened 2023-07-12
[editorial] Format doc pages using Prettier?


chalin:
Might maintainers consider using Prettier to format docs pages? We're happily using Prettier for the OTel website. Advantages, among others:

- Uniformity in markdown syntax layout
- We save time and energy by avoiding debates over markdown layout
- **Easier to diff changes to doc pages (because Prettier line-wraps paragraphs --- vs. a single text line per paragraph)**

The last point is one of the _main_ advantages IMHO. I was prompted to create this issue following #188.

- [x] #192

joaopgrassi:
We should consider if it's possible to enforce prettier formatting as a PR check. Otherwise we will keep running into these issues.

thompson-tomo:
agree and that is the intention once the rules have been applied to the docs as otherwise the pr would be quite big. this way it is split in to smaller chunks.

thompson-tomo:
Pr to add it as a check has been done as #3015

lmolkova:
Is there any practical need in prettier? some specific issues with inconsistent lists or italic syntax? We have yamllint and markdown lint that validate syntax. Could we configure policies that are important on markdown lint?

Adding pretties would add yet another tool and yet another complication and failed check. 

We don't have any debates over  proper markdown syntax and I have never seen any practical concern related to different list or italic syntax.

thompson-tomo:
So the biggest benefit would be being able to run it on the generated output. As well as ensure consistent syntax for formatting which linting does not do. This would then mean that we can improve the readability of our jinja templates due to the templates being able to prioritise readability/maintainability instead of focussing on what the raw output looks like.

Prettier & linter are complementary tools.

lmolkova:
I don't understand it. We do run linting on generated output. Our linting is not opinionated and doesn't impose arbitrary limitations like _ for italic. There is no maintainability or readability problem wrt italic choices or list styles. 

Yet another linting tool would be a source of failures. 

Notice how you can't even see the failures in CI log, you have to run it (markdownlint on the contrary shows what errors are in the logs):

prettier log https://github.com/open-telemetry/semantic-conventions/actions/runs/18942144616/job/54083463665?pr=3015

```
Checking formatting...
[warn] docs/non-normative/groups/system/cpu-metrics-guidelines.md
[warn] docs/non-normative/groups/system/design-philosophy.md
[warn] Code style issues found in 2 files. Run Prettier with --write to fix.
```

markdownlint log
```
[...]

PR #3090, opened 2025-11-15
[chore] Switch aspnetcore list to - rather than * to enable linting

thompson-tomo:
## Changes

This switches the list nested in an aspnetcore note to - so that the doc can be linted.

> [!IMPORTANT]


lmolkova (changes requested):
See discussion in #189

trask:
@thompson-tomo is this the only place `*` is used other than `-`? we don't really want to enforce at this time, and we also don't want a lot of PRs that only do these changes that don't have any clear benefit

thompson-tomo:
@lmolkova i can see why you would draw a bow to prettier however that is not the correct bow to draw. This PR is part of getting **linting** running on all the generated content which is currently not the case and different to adding a formatter. That is why it is linked to #3074 and without this change the entire signal definition needs a manual suppression as seen in #3074

@trask this in the only case needed to eliminate linting suppression from general docs (manually created excluding resource docs). If you want to see the benefit of having linting run on the docs, just look at #3073 which would've been caught by linting.

trask:
why isn't this part of #3013?

thompson-tomo:
@trask because #3013 focused on standardizing the manually defined symbol's and didn't touch any defined via models.

trask:
ok, let's close this for now and you can come back to it after the bigger decision in #3013 is made

PR #3097, opened 2025-11-17
Introduce service.peer.name and service.peer.namespace; deprecate peer.service

### LINKED ISSUE #2945: Revisit peer.* namespace to match service.*
### Area(s)

area:peer

### What's missing?

The `peer.service` attribute does not match the way we define the identity of services in the `service.*` namespace via `service.name` and `service.namespace`.

Part of this issue was also brought up as #2338

One option would be to introduce `peer.service.name` and `peer.service.namespace` with the intention of supporting match the usage of `service.namespace` + `service.name` as identity for services.

I could also envision "special-casing" this to avoid breaking OpenTracing users by allowing `peer.service` to have a special syntax to represent combinations of `service.namespace` + `service.name`. However, I know from experience that some organizations already use `/` in their `peer.service` to represent the operation used on the invoked service.

### Tip

mmanciop:
Fixes #2945

## Changes

The goals of this PR are twofold.

### Goal 1: Stabilization

Renaming `peer.service` is necessary for stabilization: the same way the SIG renamed `deployment.environment` to `deployment.environment.name`, the lack of extensibility of `peer.service` prevents its stabilization.

The extensibility of `service.peer.*` is itself important because of the next goal.

### Goal 2: Provide means of fully describing service identity

Today's `peer.service` is meant to contain the same value of the `service.name` attribute of the service on the other end of communication. However, `service.name` by itself does not actually suffice to identify every service, because of the existence and adoption of `service.namespace`:

> **`service.namespace:`** A string value having a meaning that helps to distinguish a group of services, for example the team name that owns a group of services. service.name is expected to be unique within the same namespace. If service.namespace is not specified in the Resource then service.name is expected to be unique for all services that have no explicit namespace defined (so the empty/unspecified namespace is simply one more valid namespace). Zero-length namespace string is assumed equal to unspecified namespace.

(From https://opentelemetry.io/docs/specs/semconv/resource/#service)

That is, in order to allow "references" to a service, we need to be able to also add `service.peer.namespace`.

The importance of `service.peer.namespace` is directly related with the adoption of `service.namespace`. I pulled statistics across the entirety of the use-base of Dash0, and almost exactly 1/3 of organizations (which equate to a "user site") set `service.namespace` to one or more of their services. (I admit this came to me as a surprise, I had guess-stimated the adoption to be around 10%). This high adoption, in my eyes, justifies making an exception to the rule about existing implementations.

For the record, about implementations: we intend in Dash0 to automatically add `service.peer.name` and `service.peer.namespace` to metrics generated for span pairs connecting services. Which means 1/3 of Dash0 users will automatically adopt these attributes.


model/service/registry.yaml
```diff
@@ -69,3 +69,22 @@ groups:
...
+        brief: >
+          Logical name of the service on the other side of the connection.
+          SHOULD be equal to the actual [`service.name`](/docs/resource/README.md#service) resource attribute of the remote service if any.
+        examples: ["shoppingcart"]
+      - id: service.peer.namespace
+        type: string
+        stability: development
+        brief: >
+          A namespace for `service.peer.name`.
+        examples: ["Shop"]
```
thompson-tomo, line 91:
```suggestion
      - id: server.service.name
        type: string
        stability: development
        brief: >
          Logical name of the service running on the server side of the connection.
          SHOULD be equal to the actual [`service.name`](/docs/resource/README.md#service) resource attribute of the remote service if any.
        examples: ["shoppingcart"]
      - id: server.service.namespace
        type: string
        stability: development
        brief: >
          A namespace for `server.service.name`.
        examples: ["Shop"]
```

model/peer/common.yaml
```diff
@@ -4,4 +4,6 @@ groups:
     brief: "Operations that access some remote service."
     attributes:
       - ref: peer.service
-        requirement_level: recommended
+        deprecated:
+          reason: renamed
+          renamed_to: service.peer.name
```
thompson-tomo: Should be specified on the attribute definition/registry and not the usage of it.

```suggestion
```

thompson-tomo (commented):
PR should be Blocked due to old attributes not actually being deprecated but rather only a single usage of the attribute being deprecated. Also there is 2 change logs when there should only be 1.

If we think about the use of the attributes for me it makes more sense to put these new attributes under the server namespace given that they are being set based on the server address/port.
It would also enable the server namespace to be the sole namespace containing attributes describing the server side of a connection and the service namespace to be describing the service being used.

mmanciop:
> PR should be Blocked due to old attributes not actually being deprecated but rather only a single usage of the attribute being deprecated. Also there is 2 change logs when there should only be 1.

I am trying to figure out how to make this work the right way. Pointers appreciated :-)

> If we think about the use of the attributes for me it makes more sense to put these new attributes under the server namespace given that they are being set based on the server address/port.

IMO, it is really not a given that these are set on server address/port. It is difficult to generalize on something like this, and it is going to be very implementation-dependent, but more than one implementation of peer.service I have seen happened at a pretty high level, e.g., embedded in client libraries automatically generated from OpenAPI specs.

thompson-tomo:
> I am trying to figure out how to make this work the right way. Pointers appreciated 

Move the deprecated block from common.yaml into the registry.yaml file and then regenerate the docs. The generation process will propagate the deprecation info onto the attributes on the signals.

> IMO, it is really not a given that these are set on server address/port.

This is functionality which is defined via declarative config.

> I have seen happened at a pretty high level, e.g., embedded in client libraries automatically generated from OpenAPI specs.

That would also be fine and still fit having it under server as server would be describing the remote side of the connection.

trask:
hi @mmanciop!

thanks for taking this on.

we discussed this in yesterday's Service and Deployment Semconv SIG meeting (https://zoom.us/rec/share/W7L9lhkmwN6GGzh1ykMxcF-dr2B0frjOAaO52BSe-8VP-Zag6n15gQObv_sFtqM.OCG1dIwmHpD2D2rP starting around 34:00 minute marker).

I believe we're onboard with renaming `peer.service` to `service.peer.name`.

We'd like a bit more understanding of whether `service.peer.namespace` is needed, since we haven't had that previously, and so it would need to go through a bit more prototyping / discussion.

For `service.peer.name`, we want to make sure we have a clear path towards stabilizing, as we have strong preference for "defacto stable" things like `peer.service` to only break once when going to (true) stable. Similar to what we did for HTTP and database semconv.

We still need to decide whether we want to recommend `OTEL_SEMCONV_STABILITY_OPT_IN` as part of this migration.

Can you investigate the existing usages of `peer.service` in OpenTelemetry repositories? We'd like to understand how widespread the usage is (outside of the Java agent where we know it has decent usage).

Once we decide on the migration and stabilization plan, we want to communicate it via a short blog post, to hopefully avoid the surprises that folks had with the `deployment.environment` -> `deployment.environment.name` change.

mmanciop:
> hi @mmanciop!
> 
> thanks for taking this on.
> 
> we discussed this in yesterday's Service and Deployment Semconv SIG meeting (https://zoom.us/rec/share/W7L9lhkmwN6GGzh1ykMxcF-dr2B0frjOAaO52BSe-8VP-Zag6n15gQObv_sFtqM.OCG1dIwmHpD2D2rP starting around 34:00 minute marker).
> 
> I believe we're onboard with renaming `peer.service` to `service.peer.name`.

Very nice!

> We'd like a bit more understanding of whether `service.peer.namespace` is needed, since we haven't had that previously, and so it would need to go through a bit more prototyping / discussion.

The identity of a service is the combination of its `service.name` and, if set, its  `service.namespace`.  From https://opentelemetry.io/docs/specs/semconv/resource/:

>> service.namespace: A string value having a meaning that helps to distinguish a group of services, for example the team name that owns a group of services. service.name is expected to be unique within the same namespace. If service.namespace is not specified in the Resource then service.name is expected to be unique for all services that have no explicit namespace defined (so the empty/unspecified namespace is simply one more valid namespace).

The canonical example in my head is the `frontend` service: most complex systems have one, and this is where the `service.namespace` kicks in. If the service on the other side cannot say which application is talking to, but just that it is "some frontend", that nuance is lost.

> For `service.peer.name`, we want to make sure we have a clear path towards stabilizing, as we have strong preference for "defacto stable" things like `peer.service` to only break once when going to (true) stable. Similar to what we did for HTTP and database semconv.
> 
> We still need to decide whether we want to recommend `OTEL_SEMCONV_STABILITY_OPT_IN` as part of this migration.
>
> Can you investigate the existing usages of `peer.service` in OpenTelemetry repositories? We'd like to understand how widespread the usage is (outside of the Java agent where we know it has decent usage).

Sure, I had already done the research before coming up with the proposal, and I just spent some more time with GitHuib search to make sure I did not miss anything. To my knowledge, the list is as follow, going through SDKs and contribs in alphabetical order:

* **C++:** Found nothing relevant
* **Erlang:** Found nothing relevant
* **Go:** There's some [examples in contrib](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-go-contrib%20PeerService&type=code), but I do not see any automation built into the SDK or instrumentations 
* **Java:** Various Java instrumentations can [automatically add `peer.service`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/a8622adb471921b43f70a6161e41de23b51d14d3/instrumentation-api-incubator/src/main/java/io/opentelemetry/instrumentation/api/incubator/semconv/net/PeerServiceAttributesExtractor.java#L24) based on [configurable mappings](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/a8622adb471921b43f70a6161e41de23b51d14d3/instrumentation-api-incubator/src/main/java/io/opentelemetry/instrumentation/api/incubator/config/internal/CommonConfig.java#L51) of peer services (see e.g. [PeerServiceResolverImpl](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/a8622adb471921b43f70a6161e41de23b51d14d3/instrumentation-api-incubator/src/main/java/io/opentelemetry/instrumentation/api/incubator/semconv/net/PeerServiceResolverImpl.java#L20)). These functionalities are in incubation.
* **JS:** Found nothing relevant
* **.NET:** Found nothing relevant, besides the Zipkin exporter mentioned below
* **PHP:** Found nothing relevant, besides the Zipkin exporter mentioned below
* **Python:** Found nothing relevant
* **Ruby:** Some instrumentations [allow to configure `peer.service`](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-ruby-contrib%20peer.service&type=code)
* **Rust:** Found nothing relevant

* **Collector:**
  * [Datadog exporter in the collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6968a70a32d3f0e0d02b8d53c2c974dea4432224/pkg/datadog/config/traces.go#L67), which is used to generate metrics about how services interact, and `peer.service` is one of the dimensions. It is opt-in.
  * [Service Graph Connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/6968a70a32d3f0e0d02b8d53c2c974dea4432224/connector/servicegraphconnector) (in alpha) uses `peer.service` as a way to look up connections. From the code, it seems to me it is the default way it uses, but I am not quite sure.
  * The [`AWS XRay Exporter`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6968a70a32d3f0e0d02b8d53c2c974dea4432224/exporter/awsxrayexporter/internal/translator/segment.go#L384) uses `peer.service` as a fallback for naming segments.

Others:

* Zipkin exporters, e.g., the [.NET one](https://github.com/open-telemetry/opentelemetry-dotnet/blob/2a288c30bd525ce00ae7d5e8727d54430f1d7b2f/src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs#L225) of the [PHP one] (https://github.com/open-telemetry/opentelemetry-php/blob/569eac08ce735a945fee5719c010580f1c90d304/src/Contrib/Zipkin/SpanConverter.php#L31), use `peer.service` (and other fallbacks based on `server.*` and other semconvs) to determine the remote endpoint, which in turn, AFAICT, is used by Zipkin's dependency graph. However, Zipkin exporters [are deprecated](https://opentelemetry.io/blog/2025/deprecating-zipkin-exporters/).
 
> Once we decide on the migration and stabilization plan, we want to communicate it via a short blog post, to hopefully avoid the surprises that folks had with the `deployment.environment` -> `deployment.environment.name` change.

Sounds good to me.

I also think that adding one more fallback case for `service.peer.name` to most of the projects above is not difficult. There is, however, the fact that their semantics of service identity is going to be incomplete anyhow if the logic is not extended to also support `service.peer.namespace`. But I suspect most stakeholders will answer with "`service.namespace`, what's that?".

model/service/registry.yaml
jsuereth: The entire rationale behind this PR is that we want the `service` related names to be under the same "authority" / maintainership.

moving into `server`, again, splits the ownership in ways we do not want.  

The goal is when we change `service.*` attributes, we can more easily see the impact on `service.peer` attributes and maintain them cohesively.

model/service/registry.yaml
```diff
@@ -69,3 +69,22 @@ groups:
...
+    display_name: Service Attributes for Peer Services
+    brief: >
+      How to describe the service on the other side of a request.
+    attributes:
+      - id: service.peer.name
+        type: string
+        stability: development
+        brief: >
+          Logical name of the service on the other side of the connection.
+          SHOULD be equal to the actual [`service.name`](/docs/resource/README.md#service) resource attribute of the remote service if any.
```
jsuereth, line 83: What has me concerned here is we don't have a specified way for *instrumentation* (i.e the OTel API) to access the resource attribute and fill out a `service.name` here, e.g. if using baggage.

@mmanciop I do think we need an overall "Improve the state of service-peer in OTEL Specification" tracking bug, with known examples/demos you've collected.

However, there's enough usage in OTEL now, that I think we this is ok for now. I just believe this requirement is "too hard" to accomplish in practice in OTEL.

model/service/common.yaml
```diff
@@ -0,0 +1,9 @@
+groups:
+  - id: service.peer
+    type: attribute_group
+    brief: "Operations that access some remote service."
```
jsuereth: Nit:  "Operations that access some remote service MAY fill out these attributes"

mmanciop: Addressed in 27565a6

model/service/registry.yaml
mmanciop: @jsuereth https://github.com/open-telemetry/semantic-conventions/issues/3190

.chloggen/peer-service-deprecation.yaml
```diff
@@ -0,0 +1,22 @@
...
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: peer
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: The `peer.service` attribute has been deprecated in favor of `service.peer.name`.
```
lmolkova, line 13: SemConv SIG discussion:

Please update PR description to make a case for service.peer.namespace and some data points on usage (outside of otel org). It does not have to be the same attribute, just the scenario.

For stabilization, we might want to see more real-life usage (of service.peer.namespace in otel or outside of otel org).

trask:
> For the record, about implementations: we intend in Dash0 to automatically add `service.peer.name` and `service.peer.namespace` to metrics generated for span pairs connecting services.

oh, span-to-metrics is a great use case for capturing these attributes 👍

> [Service Graph Connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/6968a70a32d3f0e0d02b8d53c2c974dea4432224/connector/servicegraphconnector) (in alpha) uses `peer.service` as a way to look up connections. From the code, it seems to me it is the default way it uses, but I am not quite sure.

@mapno, @JaredTan95, @open-telemetry/collector-contrib-approvers we're trying to make a case for `service.peer.name` (renamed from `peer.service`) and `service.peer.namespace`, and it would be great to understand if the Service Graph Connector is using `peer.service` already and if it would benefit from the rename and addition of `service.peer.namespace`. Thanks!

axw:
> @mapno, @JaredTan95, @open-telemetry/collector-contrib-approvers we're trying to make a case for service.peer.name (renamed from peer.service) and service.peer.namespace, and it would be great to understand if the Service Graph Connector is using peer.service already and if it would benefit from the rename and addition of service.peer.namespace. 

It would be good to get confirmation from @mapno and/or @JaredTan95, but from my read:

By default, the connector will create a virtual node in the service graph representing a peer service, based on the presence of one or more of these attributes in a client or server span: `peer.service`, `db.name`, `db.system`. I'm not particularly familiar with this component, but based on my experience working on similar things, this would be for representing an uninstrumented source/destination service.

It should be easy enough to add `service.peer.name` and `service.peer.namespace` to the default list, and phase out `peer.service`.

(FWIW I like the name change. Besides making room for `service.peer.namespace`, it fits well with `network.peer.*`)

mapno:
>It should be easy enough to add service.peer.name and service.peer.namespace to the default list, and phase out peer.service.

Sorry for the delay (🎄). Agree, that should be the way forward for the servicegraph processor. No concerns from my side 👍

docs/general/attributes.md
```diff
@@ -349,7 +351,32 @@ For `Unix` and `pipe`, since the connection goes over the file system instead of
 
 ## General remote service attributes
 
-This attribute may be used for any operation that accesses some remote service.
+### Service Peer
+
+Attributes of the `service.peer.*` namespace may also be used for any operation that accesses some remote service.
```
trask:
```suggestion
Attributes of the `service.peer.*` namespace may be used for any operation that accesses some remote service.
```

docs/general/attributes.md
```diff
@@ -349,7 +351,32 @@ For `Unix` and `pipe`, since the connection goes over the file system instead of
...
+Users can define what the name of a service is based on their particular semantics in their distributed system.
+Instrumentations SHOULD provide a way for users to configure this name.
```
trask: my understanding is that the main use case is populating these in span to metrics pipelines?
```suggestion
```

docs/general/attributes.md
```diff
@@ -349,7 +351,32 @@ For `Unix` and `pipe`, since the connection goes over the file system instead of
...
+| Key | Stability | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Value Type | Description | Example Values |
+| --- | --- | --- | --- | --- | --- |
+| [`service.peer.name`](/docs/registry/attributes/service.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | Logical name of the service on the other side of the connection. SHOULD be equal to the actual [`service.name`](/docs/resource/README.md#service) resource attribute of the remote service if any. | `shoppingcart` |
+| [`service.peer.namespace`](/docs/registry/attributes/service.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | A namespace for `service.peer.name`. | `Shop` |
+
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+### Peer namespace
```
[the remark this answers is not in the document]
trask: I think I'd just delete this whole section, the deprecated attributes are still in the registry, but don't necessarily need to be in this document

model/service/registry.yaml
```diff
@@ -69,3 +69,22 @@ groups:
...
+        examples: ["shoppingcart"]
+      - id: service.peer.namespace
+        type: string
+        stability: development
+        brief: >
+          A namespace for `service.peer.name`.
```
trask: SHOULD be equal to the actual [`service.namespace`](/docs/resource/README.md#service) resource attribute of the remote service if any.

docs/general/attributes.md
mmanciop: Yes but like `peer.service`, this could also appear in spans. However, the current mechanics to add attributes to spans seem perfectly sufficient to me, so I am OK with dropping this.

model/service/registry.yaml
mmanciop: Done in 42991e5b

docs/general/attributes.md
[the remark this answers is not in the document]
mmanciop: Done in 42991e5b

model/peer/common.yaml
trask, line 1: I think this file can be deleted now, I think it was only used by `docs/general/attributes.md`

model/peer/deprecated.yaml
trask, line 1: rename to `deprecated.yaml` (just convention)

model/service/registry.yaml
```diff
@@ -69,3 +69,23 @@ groups:
...
```
trask: (just to align with `service.peer.name` definition above)
```suggestion
          Logical namespace of the service on the other side of the connection.
```

model/service/common.yaml
```diff
@@ -0,0 +1,9 @@
+groups:
+  - id: service.peer
+    type: attribute_group
+    brief: "Operations that access some remote service MAY fill out these attributes to describe that remote service."
+    attributes:
+      - ref: service.peer.name
+        requirement_level: recommended
+      - ref: service.peer.namespace
+        requirement_level: recommended
```
trask: I'm not sure it makes a difference here, but I think this would be more accurate (@lmolkova?)

```suggestion
      - ref: service.peer.name
        requirement_level: opt_in
      - ref: service.peer.namespace
        requirement_level: opt_in
```

jsuereth: +1 from me.

model/peer/common.yaml
mmanciop: Done in eec09d0ad4fef181cfad15245c802220a41a7739

model/peer/deprecated.yaml
mmanciop: Done in eec09d0ad4fef181cfad15245c802220a41a7739

PR #3105, opened 2025-11-21
Add containers to areas

lmolkova:
We have containers in the codeowners, but not in the areas.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -90,9 +90,9 @@
 /docs/non-normative/naming-known-exceptions.md  @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-k8s-approvers
 
 # Container semantic conventions
-/docs/resource/container.md       @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-k8s-approvers
-/model/container/                 @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-k8s-approvers
-/model/oci/                       @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-k8s-approvers
+/docs/resource/container.md       @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-k8s-approvers @open-telemetry/semconv-container-approvers
+/model/container/                 @open-telemetry/specs-semconv-approvers @open-telemetry/semconv-k8s-approvers @open-telemetry/semconv-container-approvers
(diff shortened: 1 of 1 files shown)


areas.yaml
```diff
@@ -257,3 +257,13 @@ areas:
...
       - active
+
+  - name: "Semantic Conventions: Container"
+    project: "N/A"
+    board: "N/A"
+    labels:
+      - area:container
+      - area:oci
+    status:
+      - accepting_contributions
```
lmolkova: @open-telemetry/semconv-container-approvers is the group still active and interested in contributions? Please let me know and I'll update the PR to reflect actual status. 

Also if you have a project board or any info about the project, could you please share links?

areas.yaml
```diff
@@ -257,3 +257,13 @@ areas:
...
+      - active
```
trask: another option is to add these areas under the "Semantic Conventions: K8s" and remove @open-telemetry/semconv-container-approvers in favor of @open-telemetry/semconv-k8s-approvers

ChrsMark: The truth is that the groups have a significant overlap and for now only K8s area is a priority. I think the reality is that new contributions that are also related to k8s make sense right now, otherwise for container specific stuff (i.e. podman or docker specifics) I'm not sure if the group will the have the capacity to handle them right now.

areas.yaml
ChrsMark: See https://github.com/open-telemetry/semantic-conventions/pull/3105#discussion_r2556041482

areas.yaml
lmolkova: discussed on semconv SIG:

We have similar situation with HTTP where approvers own adjusted areas like `server`, so merging `container` into `k8s` would fit into existing practices and we should do it.

If someone sends a large PR against `container` that is not aligned with k8s group goals, the group can reject changes.

So, let's start with merging containers into k8s.

@ChrsMark and @open-telemetry/semconv-k8s-approvers what should we do with @open-telemetry/semconv-container-approvers  ? Can/should we delete this group?

dmitryax (approved):

ChrsMark (approved):

areas.yaml
ChrsMark: k8s-approvers group is a super-set of the container-approvers so I guess it's fine if we remove the latter.

PR #3107, opened 2025-11-21
[process] Add requirement levels to attribute refs

braydonk:
Part of #864 

## Changes

This PR adds the requirement levels for all attribute refs in the `process` namespace.

An attribute being marked `required` is our way of saying that we do not see a reason for the metric to be instrumented without that attribute present; perhaps there might be uses for aggregating over a particular attribute at query time, but at collection time we expect the attributes to always be produced.

> [!IMPORTANT]

### DIFF
--- a/.chloggen/process_req_levels.yaml
+++ b/.chloggen/process_req_levels.yaml
@@ -1,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: process
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Added specific requirement levels to all attribute refs within process metrics.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [864]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:


docs/system/process-metrics.md
```diff
@@ -63,7 +63,7 @@ This metric is [recommended][MetricRecommended].
 
 | Key | Stability | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Value Type | Description | Example Values |
 |---|---|---|---|---|---|
-| [`cpu.mode`](/docs/registry/attributes/cpu.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | A process SHOULD be characterized _either_ by data points with no `mode` labels, _or only_ data points with `mode` labels. [1] | `user`; `system` |
+| [`cpu.mode`](/docs/registry/attributes/cpu.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | A process SHOULD be characterized _either_ by data points with no `mode` labels, _or only_ data points with `mode` labels. [1] | `user`; `system` |
```
mx-psi, line 66: Given the description (*data points with no `mode` labels, or only data points with `mode` labels.*), should this be conditionally required?

rogercoll: Does it make sense to aggregate `process.cpu.time` over `cpu.mode` (data points with no mode labels)? I think that should be equal to the `process.uptime` metric, making the aggregation not recommended (attribute required)?

braydonk: So I actually don't know if I agree with the brief here; I'm not sure at what point we would ever recommend recording CPU times without a cpu mode. I don't think we have any unambiguous way to report that. A `process.cpu.time` metric without the mode in my eyes could be one of two things:
 
1. Sum of all states, which is equivalent to all elapsed wall clock seconds (aka equivalent to `process.uptime` as @rogercoll mentioned)
2. Something like `user` + `system` with `wait` state ignored

1 is generally not useful and I think we would recommend against doing it as it could be easily misinterpreted; a dashboard interpreting a "cpu time" metric that is aggregated in this way could make it look like a process is literally operating on a CPU at all times.

I could see 2 being useful somehow.

The problem then is the ambiguity of these two scenarios; a `process.cpu.time` metric reported without a `cpu.mode` has no way to encode which of these 2 versions of the metric being reported are actually being used. The way to disambiguate a scenario that allows the second scenario to exist would just be to report a `cpu.mode` attribute, defeating the purpose of ever reporting without the attribute. We could try and write some kind of guidance that says if you report without an attribute then what you mean is all non-`wait` time of the process, but that also has usability concerns because if a process is truly using all wall clock seconds to operate on the CPU, it is much easier to know unambiguously that this is the case if you also report a `wait` state with a value of 0.  
Also in spite of all of this, scenario 2 is better served by a query-time aggregation anyway; if instrumentation aggregated all non-`wait` states at collection time you'd never be able to split that out again. (this is a similar scenario to utilization metrics erasing important time information because of collection-time aggregation).

Hopefully my rant is clear enough. All of this is to say that I believe the attribute should remain `required`, and we should adjust the brief to not suggest that we ever want this metric reported without a `cpu.mode`. How does that sound @open-telemetry/semconv-system-approvers ?

ChrsMark: > Hopefully my rant is clear enough. All of this is to say that I believe the attribute should remain required, and we should adjust the brief to not suggest that we ever want this metric reported without a cpu.mode. How does that sound @open-telemetry/semconv-system-approvers ?

I would support making this required and fixing the brief accordingly.

Historically this is coming back from the merge of cpu's mode/state attributes: https://github.com/open-telemetry/semantic-conventions/pull/1026/changes#diff-55cde26310e8ffbb851757ff445f7bee875b31612bf2e97f9cdab636212221ccL12-L13. I'm just curious if we know about any specific reason for having this `SHOULD be characterized _either_ by data points with no mode labels, _or only_ data points with mode labels` statement 🤔 . If can't find any reference for this we can freely change it now I guess.

braydonk:
@open-telemetry/specs-semconv-maintainers could we please re-open this PR? I've updated the branch and would prefer to re-open this rather than opening a new PR so that the discussion stays attached.

mx-psi:
It says "The process_req_levels_branch was force-pushed or recreated." if I hover over the reopen button, I suspect it won't be possible to reopen the PR

braydonk:
Okay, I will recreate the PR

PR #3108, opened 2025-11-21
[chore] Rename */database to */db to match area names

lmolkova:
Area name `db` (and root namespace) is inconsistent with folder name (`database`) resulting in complications when determining area like in https://github.com/open-telemetry/semantic-conventions/pull/2924#issuecomment-3564275144

It was also a problem in experiments around schema v2 - https://github.com/open-telemetry/semantic-conventions/compare/main...lmolkova:semantic-conventions:schema-v2-http-db

So renaming to `db` to simplify things.

---

TL;DR: area name must be the same as folder name


docs/db/cassandra.md
```diff
@@ -1,5 +1,6 @@
 <!--- Hugo front matter used to generate the website version of this page:
 linkTitle: Cassandra
+aliases: [/docs/specs/semconv/database/cassandra.md]
```
chalin, line 3: FYI, aliases are website paths, and so should not end with `.md`. /cc @vitorvasc

docs/db/README.md
```diff
@@ -1,5 +1,6 @@
 <!--- Hugo front matter used to generate the website version of this page:
 linkTitle: Database
+aliases: [/docs/specs/semconv/database]
```
chalin, line 3: I suggest that we drop this and every other `aliases` entry that was _added_ via this PR (#3108), and replace it by a single wildcard redirect rule in this README.md file. I'll create a separate issue to track this.

PR #3110, opened 2025-11-22
[chore] use npm ci instead of npm install

thompson-tomo:
## Changes

This switches from using npm install to npm ci which has required the package-lock.json to be checked im. This lock file will be managed via renovate.

By making this switch it should increase our ossf score card by ensuring all dependencies are pinned as currently there is 6 which is not and that appears to be connected to the 6 instances of npm install. https://scorecard.dev/viewer/?uri=github.com/open-telemetry/semantic-conventions

> [!IMPORTANT]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,7 +220,7 @@ fix: table-generation registry-generation misspell-correction markdown-toc
 
 .PHONY: install-tools
 install-tools: $(MISSPELL)
-	npm ci
+	npm ci --ignore-scripts
 	@echo "All tools installed"
 
 $(CHLOGGEN):


.github/workflows/build-system-check.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
 
     - name: install dependencies
-      run: npm install
+      run: npm ci
```
trask: let's use this everywhere instead of adding the `.npmrc` file

appears to be the way other otel repos are handling it

```suggestion
      run: npm ci --ignore-scripts
```

Makefile
```diff
@@ -220,7 +220,7 @@ fix: table-generation registry-generation misspell-correction markdown-toc
 
 .PHONY: install-tools
 install-tools: $(MISSPELL)
-	npm install
+	npm ci
```
arminru:
```suggestion
	npm ci --ignore-scripts
```
Or is this left our here on purpose?

thompson-tomo: Had missed it by mistake.

PR #3118, opened 2025-11-24
Update GraphQL Recommended vs Opt-In requirements

### LINKED ISSUE #2985: GraphQL "Recommended" for GraphQL documents is better as "Opt-In"
### Area(s)

area:graphql

### What's the problem?

The `graphql.document` is user-inputted and unspecified in length (potentially megabytes+ long), and due to the use of string and numeric literals (failure to use `variables)` it has more high-cardinality than the existing `graphql.operation.name` (which is already warned against) but also the risk of containing sensitive information which is difficult to fully sanitize.

For this reason, `graphql.document` is a liability to have listed as `Recommended` without serious infrastructure considerations and needs, often that are beyond the requirements of most users, making `Opt-In` a better option with a lower risk profile.  In our customer adoption of OpenTelemetry at Apollo (in the Apollo Router), we've found our GraphQL customers unwittingly following this currently published configuration/instruction from these conventions while not understanding the full implications.  We can mitigate against some of this in our own documentation and implementation, but a more opinionated "upstream" stance is worth considering.  (For example, in our own implementation we can defend against operation name length to reasonable defaults (e.g., max operation name of 255 characters), but `graphql.document` is harder to be opinionated about without compromising the executability of the operation).

In terms of debugability, in most cases, the lesser-liability of `graphql.operation.name` (the actual operation name) is sufficient to track usage since in many GraphQL deployments there is generally correlation between the two (in client code-bases) which can be fetched out of band.  This also spares the transmission of a large payload of the operation body to metric providers.

That said, the operation name (`graphql.operation.name`) isn't entirely without its risk, but it's more likely to be dozens of *bytes* of a limited character set rather than dozens, hundreds or potentially thousands of *kilobytes*.  In that regard, it's reasonable that `graphq.operation.name` be left as `Recommended`, though the argument could easily be made that it should also be `Opt-In`.

At this time, I recommend:

* `graphql.document` *become* `Opt-In` on account of it's more serious liabilities (including sensitive information redaction) which are not straightforward to defend against or mitigate in most deployments
* `graphql.operation.name` remain `Recommended`, though should still remain disclaimed in regards to its high-cardinality/user-input concerns (which are already noted).

I am happy to open a PR.

### Tip

abernix:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/2985

## Changes

Changes the firmness for `graphql.document` from `Recommended` to `Opt-In`.

## Context

The `graphql.document` is user-inputted, often contains sensitive information, is potentially unbounded in length and is also high-cardinality in the same way as the existing `graphql.operation.name` (which is already warned about).

Put another way, `graphql.document` is a general liability to have listed as `Recommended` without serious infrastructure considerations and needs.  This makes it a better candidate for being an `Opt-In` attribute.  In our customer adoption of OpenTelemetry and GraphQL, we've found GraphQL customers following this configuration/instruction purely based on its naming in this (still, experimental) specification and not understanding the implications (or just not thinking them through).

In most all cases, the lesser liability of `graphql.operation.name` (The actual operation name) is sufficient, as in many GraphQL deployments there is a link between the two which can be correlated out of band (e.g., client code-bases, operation manifests, etc.).  While the operation name isn't without its risk, but it's more likely to be dozens of *bytes* of a limited character set rather than dozens, hundreds or potentially thousands of *kilobytes*.  In that regard, it's reasonable that `graphql.operation.name` be left as `Recommended` for user experience (though the argument could easily be made that it should also be `Opt-In`).

### DIFF
--- a/.chloggen/abernix_graphql-spec-requirements.yaml
+++ b/.chloggen/abernix_graphql-spec-requirements.yaml
@@ -4,7 +4,7 @@
 # your pull request title with [chore] or use the "Skip Changelog" label.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: enhancement
+change_type: breaking
 
 # The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
 component: graphql
--- a/.chloggen/abernix_graphql-spec-requirements.yaml
+++ b/.chloggen/abernix_graphql-spec-requirements.yaml
@@ -1,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: graphql
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Change `graphql.document` attribute requirement level from Recommended to Opt-In due to sensitive data, cardinality, and size concerns"
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [2985]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The `graphql.document` attribute contains user-inputted data that is often sensitive,
+  potentially unbounded in length, and high-cardinality. Moving from Recommended to Opt-In
+  ensures operators make an informed decision before enabling this attribute.
+  Also updated the note to use SHOULD for stronger normative language regarding sanitization.
--- a/model/graphql/registry.yaml
+++ b/model/graphql/registry.yaml
@@ -31,5 +31,5 @@ groups:
         brief: "The GraphQL document being executed."
         type: string
         stability: development
-        note: The value may be sanitized to exclude sensitive information.
+        note: The value SHOULD be sanitized to exclude sensitive information.
         examples: 'query findBookById { bookById(id: ?) { name } }'
--- a/model/graphql/registry.yaml
+++ b/model/graphql/registry.yaml
@@ -31,5 +31,5 @@ groups:
         brief: "The GraphQL document being executed."
         type: string
         stability: development
-        note: The value SHOULD be sanitized to exclude sensitive information.
+        note: If instrumentation can reliably identify and redact sensitive information it SHOULD do it.
         examples: 'query findBookById { bookById(id: ?) { name } }'
--- a/docs/graphql/graphql-spans.md
+++ b/docs/graphql/graphql-spans.md
@@ -36,11 +36,11 @@ the span SHOULD be named `GraphQL Operation`.
 
 | Key | Stability | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Value Type | Description | Example Values |
 | --- | --- | --- | --- | --- | --- |
-| [`graphql.document`](/docs/registry/attributes/graphql.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | The GraphQL document being executed. [1] | `query findBookById { bookById(id: ?) { name } }` |
 | [`graphql.operation.name`](/docs/registry/attributes/graphql.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | The name of the operation being executed. | `findBookById` |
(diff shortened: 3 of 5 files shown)


joaopgrassi:
Note you will have to change the [model file](https://github.com/open-telemetry/semantic-conventions/blob/main/model/graphql/spans.yml#L27), and then generate the [update commands](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#2-update-the-markdown-files). The markdown is not to be modified manually.

Also, consider adding a note to the opt_in explaining the rationaly. See an example here: https://github.com/open-telemetry/semantic-conventions/blob/2e687561e5678d3c47012cc041c770493bfee854/model/http/spans.yaml#L130

abernix:
> Note you will have to change the [model file](https://github.com/open-telemetry/semantic-conventions/blob/main/model/graphql/spans.yml?rgh-link-date=2025-11-24T13%3A44%3A31Z#L27), and then generate the [update commands](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md?rgh-link-date=2025-11-24T13%3A44%3A31Z#2-update-the-markdown-files). The markdown is not to be modified manually.

Thanks! And sorry, I realized that just after I opened it and shortly after I signed the CLA, which was the first thing I wanted to tackle.

I'm not sure why this PR was just auto-closed though — do you have suggestions on that @joaopgrassi ?

abernix:
@joaopgrassi It looks like some relatively new automation auto-closed this, so ping again on the above if you get a chance.   I guess this needs `triage:accepted:ready` if you're willing to do that for this change.  I recognize that would be outside of a SIG group's coverage if i'm understanding the state of `graphql` right now.  I can see what it would look like to spin up that group, but would also prefer to decouple it from this change if at all possible! :)

.chloggen/abernix_graphql-spec-requirements.yaml
```diff
@@ -0,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
joaopgrassi:
```suggestion
change_type: breaking
```
Right? because now it won't be recorded by default anymore.

abernix: Sounds right to me.  Fixed!

model/graphql/registry.yaml
```diff
@@ -31,5 +31,5 @@ groups:
         brief: "The GraphQL document being executed."
         type: string
         stability: development
-        note: The value may be sanitized to exclude sensitive information.
+        note: The value SHOULD be sanitized to exclude sensitive information.
```
lmolkova: I agree with the intention, but I don't think it's implementable in the current shape.

some suggestions (please review my English):


```suggestion
        note: If instrumentation can reliably identify and redact sensitive information is SHOULD do it.
```

or 

```suggestion
        note: Application developers should sanitize this attribute in telemetry processing pipeline to exclude sensitive information.
```

(we don't use normative language when talking about app developers - we don't have authority over them)

abernix:
Unfortunately, this fell stale over the holidays (I went on actual holiday! 😆 ).  I'd like to revisit it and get it in shape if that's still possible.

abernix:
Sorry auto-close bot — I'll be back soon, I promise!

model/graphql/registry.yaml
abernix: Good catch -- went with Option A, directing the normative language at instrumentation authors where the spec actually has authority.  Pushed in 41305119.

abernix:
Ok, appreciate the patience. I believe this is is ready for review again.

abernix:
@joaopgrassi Is it possible to get this reviewed again?  @lmolkova I think I addressed your feedback. Thanks!

lmolkova (approved):

abernix:
Is there anything else I need to do to help here?

trask (approved):

joaopgrassi (approved):

joaopgrassi:
@abernix not sure what is wrong with the action, maybe you can try rebasing/squashing the commits to see if it gets happy again? It seems your PR is not allowed for maintainers to change it, so I can't push to it.

jsuereth (approved):

abernix:
@joaopgrassi Hmm, I rebased and pushed and it still is failing.

I don't think it's related to my branch at all — it looks like an issue with the workflow that would affect any fork
PR?

I think it's something like this: To figure out what files changed, `tj-actions/changed-files` needs two things: the PR branch commits, and the current tip of upstream main to compare against.  The workflow checks out the fork repo (mine, to get the PR branch), but the fork **doesn't** have the latest upstream main commits — those only exist in this upstream.  So when the action tries to find the merge base, it can't locate the upstream commit and fails.

Perhaps one fix would be adding `use_rest_api: true` to the `tj-actions/changed-files` step, which uses the GitHub API instead of local git history and sidesteps the problem?

abernix:
I don't meant to claim understand what it means, but I will note that the failing CI check is not a "Required" check.

PascalSenn:
@abernix  were you aware of https://github.com/graphql/otel-wg and the resulting PR https://github.com/open-telemetry/semantic-conventions/pull/3515

Issue #3121, opened 2025-11-24
Add `oracle_cloud.realm` vendor attribute to Semantic Conventions


geekdave:
### Area(s)

area:cloud

## Background

Back in https://github.com/open-telemetry/semantic-conventions/issues/1658 we added Oracle Cloud Infrastructure (OCI) awareness to `cloud.provider` and `cloud.platform` but we didn't yet introduce any vendor-prefixed attributes.

We would like to introduce the first OCI vendor-specific attribute called `realm`. 

## What is a realm?

> A realm is a logical collection of cloud regions that are isolated from each other and do not allow customer content to traverse realm boundaries to a region outside that realm. Each realm is accessed separately. OCI has multiple realms, including a commercial public cloud realm, multiple government cloud realms for the US and UK, and the Dedicated Regions from each customer are contained within their own separate realm.

-- From the [Oracle Cloud Infrastructure Blog](https://blogs.oracle.com/cloud-infrastructure/sovereign-cloud-realms-enhanced-isolation)

## Why is this needed?

Operators running across commercial and government environments need this context to route telemetry to realm-specific endpoints and enforce data residency/compliance.  

More about realms: 
1. https://docs.oracle.com/en-us/iaas/Content/General/Concepts/regions.htm
2. https://blogs.oracle.com/cloud-infrastructure/sovereign-cloud-realms-enhanced-isolation

### Describe the solution you'd like

Add `oracle_cloud.realm` (type: `string`, stability: `development`) under a new vendor namespace (`registry.oracle_cloud`), documented with short examples (e.g., `oc1`, `oc2`) and a link to OCI realm docs. Keep it a free-form string to accommodate future realms.  

Will be used by the oracle cloud resource detection processor to populate the realm attribute.  See https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44408

cc @dashpole @tonychoe

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #3124, opened 2025-11-24
Add `oracle_cloud.realm` vendor attribute

geekdave:
Fixes open-telemetry/semantic-conventions#3121

## Changes

Add `oracle_cloud.realm` vendor-specific attribute for Oracle Cloud Infrastructure (OCI) `realm` identifier.

## Note

This will be used in the already-created Oracle Cloud resource detection processor, part of the collector-contrib project.  See: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44408

I went back and forth about whether this `realm` attribute should be placed at:

a) the root of the vendor-prefixed hierarchy such as `oracle_cloud.realm` or
b) in service-specific properties like `oracle_cloud.oracle_cloud_oke.realm` and `oracle_cloud.oracle_cloud_compute.realm`

I asked about this in `#otel-semantic-conventions` and got this feedback from @thompson-tomo :

> In the case here I lean towards having a single  attribute based on the assumption that a realm is more like cloud.region hence a signal would only be describing a single realm

I am open to more feedback on this.  But realm is a concept that applies to nearly all OCI services, so I think it makes sense to place it at the root.


geekdave:
cc @dashpole .  Not sure what to do about the above error.  Do we need to start a new SIG/project?

dashpole:
@jsuereth I Don't see a SIG/project associated with most cloud providers.  How can we add oracle-cloud?  Do we need a new SIG?

geekdave:
Paused, pending creation of `oracle_cloud` approvers group.  Proposal: https://github.com/open-telemetry/community/issues/3185

thompson-tomo:
@geekdave thoughts on having the attribute named `oracle_cloud.realm.id/key` that way we could at a later date add additional `oracle_cloud.realm.*` attributes if the need arises.

Also it might be a good idea to associate the new attribute with an entity. Based on https://opentelemetry.io/docs/specs/semconv/how-to-write-conventions/resource-and-entities/#when-to-define-an-is-a-relationship-vs-extending-descriptive-attributes my suggestion would be to extend the existing cloud entity with the additional attribute as i don't see us generating metrics specific to an Oracle cloud location which doesn't apply to all cloud locations.

For this `cloud` entity I would see the identifying attributes being `cloud.availability_zone` + `cloud.provider` but that would likely be separate pr to define those roles.

trask:
Closing and re-opening to see if it picks up new CODEOWNERS

geekdave:
@thompson-tomo thanks for the suggestion!  I attended the semconf SIG meeting via Zoom this morning and we discussed this topic.  The consensus was that it's OK to leave `realm` as a simple attribute that represents the entire identity, similar to how `availability_zone` works now.

lmolkova (approved):

PR #3130, opened 2025-11-25
Fix stale workflow config

trask:


.github/workflows/stale-pr.yml
```diff
@@ -2,13 +2,15 @@ name: "Close stale pull requests"
...
     - cron: "12 3 * * *"  # arbitrary time not to DDOS GitHub
+  workflow_dispatch:
 
 permissions:
   contents: read
 
 jobs:
   stale:
     permissions:
+      actions: write # for actions/stale to delete its old cache before saving new one
```
trask, line 13: hoping this (or `operations-per-run` below)

will fix it as it should be closing https://github.com/open-telemetry/semantic-conventions/pull/2504

.github/workflows/stale-pr.yml
```diff
@@ -17,11 +19,12 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           stale-pr-message: 'This PR was marked stale due to lack of activity. It will be closed in 7 days.'
           close-pr-message: 'Closed as inactive. Feel free to reopen if this PR is still being worked on.'
-          exempt-pr-labels: 'bug,work in progress,experts needed,never stale'
+          exempt-pr-labels: 'never stale'
```
trask, line 22: I think this change is inline with the new workflow

lmolkova (approved):

Issue #3120, opened 2025-11-24
[k8s] Stabilize k8s resource/entity attributes


ChrsMark:
Part of https://github.com/open-telemetry/semantic-conventions/issues/3119.

This issue suggests that we stabilise all resource/entity attributes that are used by the following Collector components:

- [k8sattributes processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) (from [metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/metadata.yaml#L14C1-L135C19) and [extract::metadata](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/config.go#L133-L147))
- [resourcedetection processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor)
- [filelog receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver#file-log-receiver)

Related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44130.

### Suggested Semantic Conventions

| Attribute| k8sattributes | resourcedetection | filelog | completion status | semconv link | PR |
| -------- | -------- | -------- |  ------- | ----- | ----- | ----- |
| k8s.pod.name     |  ✅     |  🚫     | ✅  | | [link](https://github.com/open-telemetry/semantic-conventions/blob/2e687561e5678d3c47012cc041c770493bfee854/model/k8s/entities.yaml#L55) | |
| container.id | ✅ |  |  | |
| container.image.name | ✅ |  |  | |
| container.image.repo_digests | ✅ |  |  | |
| container.image.tag | ✅ |  |  | |
| k8s.cluster.uid | ✅ |  |  |  | [link](https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/entities.yaml#L11) | |
| k8s.container.name | ✅ |  | ✅ | | [link](https://github.com/open-telemetry/semantic-conventions/blob/main/model/k8s/entities.yaml#L72) | |
| k8s.cronjob.name | ✅ |  |  | |
| k8s.cronjob.uid | ✅ |  |  | |
| k8s.daemonset.name | ✅ |  |  | |
| k8s.daemonset.uid | ✅ |  |  | |
[...]

PR #3135, opened 2025-11-26
[k8s] Define missing roles for entity attributes

ChrsMark:
Related to https://github.com/open-telemetry/semantic-conventions/issues/3120

## Changes

This PR adds roles for entity attributes that were missing this.

 ⚠️ Note that for `k8s.namespace.name` can be an identifying attribute within the scope of a K8s a cluster, hence `k8s.namespace.uid` is not defined at all nor used in [`k8sattributes` processor of the Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor). I'd love to hear what people think about this and if we should consider introducing the `k8s.namespace.uid` as an identifying attribute for consistency across k8s resources/entities.


model/k8s/entities.yaml
```diff
@@ -67,11 +69,13 @@ groups:
     name: k8s.container
     brief: >
       A container in a [PodTemplate](https://kubernetes.io/docs/concepts/workloads/pods/#pod-templates).
-    # TODO: define attributes roles
     attributes:
       - ref: k8s.container.name
+        role: identifying
       - ref: k8s.container.restart_count
+        role: identifying
```
jsuereth: I'm still  not certain we need to have this be identifying.

When grouping based on a k8s.container do I *want* to have restarts be considred separate containers?

It seems to me like you want to leverage a diffrent entity (like service.instance.id) to distinguish containers by restart.  Or you need something like "Container instance" entity that would track this.

ChrsMark: > When grouping based on a k8s.container do I want to have restarts be considred separate containers?

Hmm, I think you are right.

I remember the reason for having container's restart count both as an attribute and as a metric was to be able to use the attribute for identifying from which container instance a log was collected from. Specifically this is used for container's logs at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/container.md#add-metadata-from-file-path (not sure if used in other places too).

I'll defer to @dmitryax here since he should remember more details.

> It seems to me like you want to leverage a diffrent entity (like service.instance.id) to distinguish containers by restart. Or you need something like "Container instance" entity that would track this.

Probably for that purpose we could just use the `container.id`  which `k8sattributes` processor can add if `k8s.container.name` is present? I can try and come-up with a configuration that verifies this.

ChrsMark: @jsuereth I have changed the `k8s.container.restart_count` to descriptive attribute. Thinking of it again, a `k8s.container` within the Pod's scope can be uniquely identified by its name: [37887f0](https://github.com/open-telemetry/semantic-conventions/pull/3135/commits/37887f07cda0cc9fca1758139c5fd71c8c64fce2)

@dmitryax please confirm

ChrsMark: Chatted a bit about this with @dmitryax during last K8s SIG meeting.
Probably what we will need for "capturing" the k8s container's restarts/instances is sth like a `k8s.container.instance` entity similar to https://github.com/open-telemetry/semantic-conventions/blob/fd1fb75a0e697d6f685588b14001936a94d37189/model/service/entities.yaml#L18-L31. However for `k8s.container` entity the `k8s.container.restart_count` is just a descriptive attribute.

jsuereth: Thanks for following up on this! I think I agree, and I'm hoping as entity work lands in the collector, can start getting a practical feel for these decisions with OTEL-Operator / log collection architecture for k8s.

ChrsMark:
@open-telemetry/semconv-k8s-approvers please take a look

model/k8s/entities.yaml
lmolkova: @ChrsMark do you think this discussion can be resolved?

ChrsMark: I filed a follow-up [issue](https://github.com/open-telemetry/semantic-conventions/issues/3219) for the instance part so we don't forget. We can resolve this thread now :).

PR #3137, opened 2025-11-26
Apply stale workflow to draft PRs

trask:
And small improvement to stale comment

### DIFF
--- a/.github/workflows/stale-pr.yml
+++ b/.github/workflows/stale-pr.yml
@@ -17,14 +17,14 @@ jobs:
       - uses: actions/stale@5f858e3efba33a5ca4407a664cc011ad407f2008 # v10.1.0
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-          stale-pr-message: 'This PR was marked stale due to lack of activity. It will be closed in 7 days.'
-          close-pr-message: 'Closed as inactive. Feel free to reopen if this PR is still being worked on.'
+          stale-pr-message: >
+            This PR has been labeled as stale due to lack of activity.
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-pr.yml
```diff
@@ -17,14 +17,14 @@ jobs:
...
+            This PR has been labeled as stale due to lack of activity.
+            It will be automatically closed if there is no further activity over the next 14 days.
           exempt-pr-labels: 'never stale'
-          exempt-draft-pr: true
           # opt out of defaults to avoid marking issues as stale
           days-before-stale: -1
           days-before-close: -1
           # overrides the above only for pull requests
-          days-before-pr-stale: 15
+          days-before-pr-stale: 14
```
trask, line 28: aligning with weeks :)

.github/workflows/stale-pr.yml
```diff
@@ -17,14 +17,14 @@ jobs:
       - uses: actions/stale@5f858e3efba33a5ca4407a664cc011ad407f2008 # v10.1.0
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-          stale-pr-message: 'This PR was marked stale due to lack of activity. It will be closed in 7 days.'
-          close-pr-message: 'Closed as inactive. Feel free to reopen if this PR is still being worked on.'
```
arminru, line 21: Close message removed on purpose?

trask: yeah, it's not really correct (people aren't generally able to re-open their PR if it's closed by someone else), and doesn't seem needed (since the other comment is right before

PR #3140, opened 2025-11-27
[editorial] Update database redirect rules

### LINKED ISSUE #3138: [editorial] Drop aliases from 3108 and add a wildcard redirect rule
_Originally posted by @chalin in https://github.com/open-telemetry/semantic-conventions/pull/3108#discussion_r2566550006_

> I suggest that we drop this and every other `aliases` entry that was _added_ via this PR (#3108), and replace it by a single wildcard redirect rule in this README.md file. I'll create a separate issue to track this.

/cc @vitorvasc

vitorvasc:
Fixes #3138. Also referenced by https://github.com/open-telemetry/opentelemetry.io/pull/8311.

## Changes

- Drops the database alias entries that were added in #3108.
- Adds a new redirect rule to `docs/README.md`.

You can test the updates and the redirect rules here: https://deploy-preview-8311--opentelemetry.netlify.app.

When navigating to: https://deploy-preview-8311--opentelemetry.netlify.app/docs/specs/semconv/database/, it now correctly redirects to the new path: https://deploy-preview-8311--opentelemetry.netlify.app/docs/specs/semconv/db/.

### DIFF
--- a/docs/db/README.md
+++ b/docs/db/README.md
@@ -1,6 +1,5 @@
 <!--- Hugo front matter used to generate the website version of this page:
 linkTitle: Database
-redirects: [{ from: '../database/*', to: ':splat' }]
 --->
 
 # Semantic conventions for database calls and systems
--- a/docs/README.md
+++ b/docs/README.md
@@ -5,7 +5,9 @@ cascade:
   github_subdir: docs
   path_base_for_github_subdir: tmp/semconv/docs/
   github_project_repo: *repo
-redirects: [{ from: 'attributes-registry/*', to: 'registry/attributes/:splat' }]
+redirects:
+  - { from: 'attributes-registry/*', to: 'registry/attributes/:splat' }
+  - { from: 'database/*', to: 'db/:splat' }
 cSpell:ignore: semconv CICD
 auto_gen: below
 linkTitle: Semantic conventions


chalin (approved):
LGTM / is functional as is. I'd prefer having the db redirects more tightly scoped if that'll work. (Approving now in case I can't get back to it later.)

vitorvasc:
@chalin I did try at first, but I didn't really test it integrating with https://github.com/open-telemetry/opentelemetry.io/pull/8311. I was trying to test everything locally, but none of the redirect rules were working correctly (I'll follow-up about that in the #otel-comms-approvers thread we've been discussing it). 

My thinking was that if we added the rule to `docs/db/README.md`, the redirect would end up being `/docs/specs/semconv/db/database/ -> /docs/specs/semconv/db/`. In any case, I'll try pushing a few more commits to validate that behavior, I'm not totally sure that's how it actually works.

chalin:
@vitorvasc 

> @chalin I did try at first, but I didn't really test it integrating with [open-telemetry/opentelemetry.io#8311](https://github.com/open-telemetry/opentelemetry.io/pull/8311). I was trying to test everything locally, but none of the redirect rules were working correctly (I'll follow-up about that in the #otel-comms-approvers thread we've been discussing it).

👍🏻 Yeah, Netlify CLI doesn't always play nice when run locally.

> My thinking was that if we added the rule to `docs/db/README.md`, the redirect would end up being `/docs/specs/semconv/db/database/ -> /docs/specs/semconv/db/`. In any case, I'll try pushing a few more commits to validate that behavior, I'm not totally sure that's how it actually works.

The rules have special-case handling to avoid that. Either you'll need to use `from: database/*` or `from: ./database/*` I can't recall which :)

chalin:
@vitorvasc - even if you can't test using Netlify CLI, you can always inspect the redirects file that gets generated.

vitorvasc:
> @vitorvasc - even if you can't test using Netlify CLI, you can always inspect the redirects file that gets generated.

💡 💡 !

Okay, just tested it by checking the generated `_redirect` file, and looks well. I've pushed another commit to have the rule more scoped. Thanks! 😄

docs/db/README.md
```diff
@@ -1,6 +1,6 @@
 <!--- Hugo front matter used to generate the website version of this page:
 linkTitle: Database
-aliases: [/docs/specs/semconv/database]
+redirects: [{ from: 'database/*', to: 'db/:splat' }]
```
chalin:
```suggestion
redirects: [{ from: 'database/*', to: ':splat' }]
```

docs/db/README.md
```diff
@@ -1,6 +1,6 @@
 <!--- Hugo front matter used to generate the website version of this page:
 linkTitle: Database
-aliases: [/docs/specs/semconv/database]
+redirects: [{ from: 'database/*', to: ':splat' }]
```
chalin: @vitorvasc - let's give this one last try:

```suggestion
redirects: [{ from: '../database/*', to: ':splat' }]
```

The redirects might not yet have full relative-path support yet.

lmolkova (approved):

PR #3141, opened 2025-11-27
[chore]fix formating of gen ai docs

thompson-tomo:
## Changes

This ensures that the embed warning alert starts on a new line so that it is rendered correctly.

This issue is already present in the published docs https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#execute-tool-span within the foot notes section.

As an aside I have raised https://github.com/open-telemetry/semantic-conventions/issues/3142 to look at alternative options to handle the PII notes.

> [!IMPORTANT]

### DIFF
--- a/templates/registry/markdown/notes.j2
+++ b/templates/registry/markdown/notes.j2
@@ -4,7 +4,7 @@
 {% macro render() %}{% if ns.notes | length > 0 %}
 {%- for note in ns.notes %}
 {% if note.name %}**[{{ns.index+loop.index}}] `{{note.name}}`:**{% else -%}**[{{ns.index+loop.index}}]:**{%- endif -%}
-{% if note.note.startswith('>') or note.note.startswith('`') or note.note.startswith('|')%}{{"\n\n"}}{% else %} {% endif %}{{ note.note | trim }}{# If the note text starts with a '>', '|' or '```', insert 2 line breaks before the note content to ensure it starts on a new line after an empty line, otherwise just add a space before the note #}
+{% if note.note.startswith('>') or note.note.startswith('```') or note.note.startswith('|')%}{{"\n\n"}}{% else %} {% endif %}{{ note.note | trim }}{# If the note text starts with a '>', '|' or '```', insert 2 line breaks before the note content to ensure it starts on a new line after an empty line, otherwise just add a space before the note #}
 {%- if not loop.last -%}{{"\n"}}{%- endif -%}
 {% endfor %}{% set ns.index = ns.notes | length + ns.index %}{% set ns.notes = [] %}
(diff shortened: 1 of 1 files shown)


model/gen-ai/registry.yaml
```diff
@@ -261,6 +261,8 @@ groups:
         type: any
         brief: Parameters passed to the tool call.
         note: |
+          
+          
           > [!WARNING]
```
trask, line 264: This seems fragile, any other options?

thompson-tomo: Agree and it didn't end up working due to trimming. If we could add text before the note we would be fine. Hence have asked the gen-ai folks on slack for input.

lmolkova: from @open-telemetry/semconv-genai-approvers - let's not ~hack~ add text only to make tooling happy. It's ok with this approach (but as @trask mentioned will likely break at some point), but much better if could make tool not to trim it

thompson-tomo: Agree nicest solution would be to reduce the need for trimming which is often done so linting will pass as we don't have any formatter running on generated text.

I have tested a few things and the best solution i could get to work is string comparison looking for the "> [!" text which works for what we need.

templates/registry/markdown/notes.j2
```diff
@@ -3,7 +3,8 @@
 {%- macro add_with_limit(note) %}{% if note.note | length > 50 %}{% set ns.notes = [ns.notes, [note]] | flatten %} [{{ ns.notes | length + ns.index }}]{% elif note.note %} {{ note.note | trim }}{% endif %}{% endmacro %}
 {% macro render() %}{% if ns.notes | length > 0 %}
 {%- for note in ns.notes %}
-{% if note.name %}**[{{ns.index+loop.index}}] `{{note.name}}`:** {{ note.note | trim }}{% else -%}**[{{ns.index+loop.index}}]:** {{ note.note | trim }} {%- endif -%}
+{% if note.name %}**[{{ns.index+loop.index}}] `{{note.name}}`:**{% else -%}**[{{ns.index+loop.index}}]:**{%- endif -%}
+{% if note.note.startswith('> [!') %}{{"\n\n"}}{% else %} {% endif %}{{ note.note | trim }}
```
trask: can you add a brief comment here explaining? jinja already makes no sense but this is next level 😅 

> '> [!') %}{{"\n\n"}}

thompson-tomo: The key thing was if note starts with '> [!' Add 2 line breaks to force them and avoid trimming. This has been simplified to support all block quotes (not just alerts) and also code-blocks/tables. Does that note help?

PR #3163, opened 2025-12-02
genai: define cached tokens attributes

### LINKED ISSUE #23: More detailed token usage span attributes and metrics
### Area(s)

area:gen-ai

### What's missing?

Models tend to have more than just 'input' and 'output' token types. The exact type matters because they have different prices.

In [Anthropic](https://docs.anthropic.com/en/api/messages#response-usage-cache-creation-input-tokens), there's `cache_creation_input_tokens` and `cache_read_input_tokens`.

[Google](https://ai.google.dev/api/generate-content#UsageMetadata) has this:

```
{
  "promptTokenCount": integer,
  "cachedContentTokenCount": integer,
  "candidatesTokenCount": integer,
  "toolUsePromptTokenCount": integer,
  "totalTokenCount": integer,
  "promptTokensDetails": [
    {
      object (ModalityTokenCount)
    }
  ],
  "cacheTokensDetails": [
    {
      object (ModalityTokenCount)
    }
  ],
  "candidatesTokensDetails": [
    {
      object (ModalityTokenCount)
    }
  ],
  "toolUsePromptTokensDetails": [
    {
      object (ModalityTokenCount)
    }
  ]
}
```

where `ModalityTokenCount` is

```
{
  "modality": enum (Modality),
  "tokenCount": integer
}
```

and `Modality` is `TEXT`, `IMAGE`, etc. Some of the token counts in the main object are just the standard input/output, and some are totals of others.

There should be optional attributes in both spans and metrics with these details. Consistency between AI providers is probably not very important, but conventions for each provider are important for backends to do things like calculate costs.

Possible span attribute names for e.g. Anthropic's `cache_creation_input_tokens`:

```
gen_ai.usage.cache_creation_input_tokens
gen_ai.usage.detailed.cache_creation_input_tokens
gen_ai.usage.anthropic.cache_creation_input_tokens
```

Google's additional dimensions make it more complicated:

```
gen_ai.usage.cachedContentTokenCount
gen_ai.usage.cached_content_token_count
gen_ai.usage.cached_content_token_count.text
gen_ai.usage.text.cached_content_token_count
gen_ai.usage.text
gen_ai.usage.modality.text
```

(I'm assuming that the `chat` operation can still involve things like image tokens and thus modality matters)

For metrics, the `gen_ai.token.type` attribute might be enough to describe the difference between 'normal' tokens and other types like 'cached'. But a separate metric name for other types (not one metric name per type) would help prevent people from adding up token types that mustn't be added, since e.g. cached tokens are a subset of input tokens. For modalities, I suggest an attribute like `gen_ai.token.modality`.

verdie-g:
Fixes open-telemetry/semantic-conventions-genai#23

## Changes

This PR adds two new GenAI attributes to represent provider-level prompt caching:
- `gen_ai.usage.cache_read_input_tokens`
- `gen_ai.usage.cache_creation_input_tokens`

It also updates the description of gen_ai.usage.input_tokens to state that it must include cached tokens. OpenAI and Vertex AI already count cached tokens in input_tokens, while Anthropic excludes them.


model/gen-ai/spans.yaml
```diff
@@ -69,6 +69,19 @@ groups:
...
+          - **OpenAI & Vertex AI**: Their `input_tokens` fields already include cached tokens.
+            Set `gen_ai.usage.input_tokens` directly from the provider and populate
+            `gen_ai.usage.cache_read_input_tokens` from their respective cached-token fields.
+
+          - **Anthropic**: Their `input_tokens` excludes cached tokens. Compute:
+            `gen_ai.usage.input_tokens = input_tokens + cache_read_input_tokens + cache_creation_input_tokens`
+      - ref: gen_ai.usage.cache_read_input_tokens
+        requirement_level: recommended
+      - ref: gen_ai.usage.cache_creation_input_tokens
+        requirement_level: opt_in
```
alexmojaki: I think opt_in refers to the user opting in to instrumenting this attribute, which shouldn't be necessary. I don't think it's about whether the user has enabled the relevant feature.

eddyerburgh:
@lmolkova are there any concerns here or do you think this can be merged? Would be great to get these added to the spec

alexmojaki:
@eddyerburgh see my review comment, that's a required change

model/gen-ai/spans.yaml
verdie-g: Ah makes sense. I swiched to recommended.

alexmojaki (approved):

verdie-g:
@lmolkova could you have a look at this one before it gets automatically closed 🙏

alexmojaki:
There's merge conflicts now

model/gen-ai/registry.yaml
```diff
@@ -181,7 +181,23 @@ groups:
         stability: development
         type: int
         brief: The number of tokens used in the GenAI input (prompt).
+        note: |
+          This value MUST include cached tokens (cache reads and cache writes) when the provider exposes them.
+          Instrumentations SHOULD normalize provider responses so that `input_tokens` always reflects the
+          total effective input tokens the model processed.
         examples: [100]
+      - id: gen_ai.usage.cache_read_input_tokens
```
lmolkova: we recommend using dots whenever possible and namespacing things in semantic conventions.

Also suggesting a different pattern:

gen_ai.usage.{subtype}.{input | output}_tokens

```suggestion
      - id: gen_ai.usage.cache_read.input_tokens
```

model/gen-ai/registry.yaml
```diff
@@ -181,7 +181,23 @@ groups:
...
+        stability: development
+        type: int
+        brief: The number of input tokens served from a provider-managed cache.
+        note: Subset of `gen_ai.usage.input_tokens`.
```
lmolkova: Nit, just a bit more specific

```suggestion
        note: The value is typically included in `gen_ai.usage.input_tokens`. Semantic conventions for individual GenAI providers that do not include cache read tokens in `gen_ai.usage.input_tokens` SHOULD document this.
```

model/gen-ai/registry.yaml
```diff
@@ -181,7 +181,23 @@ groups:
...
+        examples: [50]
+      - id: gen_ai.usage.cache_creation_input_tokens
```
lmolkova: see my other comment
```suggestion
      - id: gen_ai.usage.cache_creation.input_tokens
```

model/gen-ai/registry.yaml
```diff
@@ -181,7 +181,23 @@ groups:
...
+        stability: development
+        type: int
+        brief: The number of input tokens written to a provider-managed cache.
+        note: Subset of `gen_ai.usage.input_tokens`.
```
lmolkova: please align with the outcome  of https://github.com/open-telemetry/semantic-conventions/pull/3163#discussion_r2705633086

model/gen-ai/spans.yaml
```diff
@@ -69,6 +69,19 @@ groups:
         requirement_level: recommended
       - ref: gen_ai.usage.input_tokens
         requirement_level: recommended
+        note: |
+          When normalizing provider responses:
+
+          - **OpenAI & Vertex AI**: Their `input_tokens` fields already include cached tokens.
```
lmolkova: could you please mention this in the individual OpenAI and Vertex AI conventions - there are separate dcs and yaml definitions for them. 

We don't have dedicated Anthropic file, please either 
- create one following openai example or 
- add a note on anthropic to all individual attributes  (in addition to keeping it here

model/gen-ai/registry.yaml
lmolkova: Even better story proposed here: https://github.com/open-telemetry/semantic-conventions/pull/3222/changes#r2650072575

let's normalize that input_tokens would include all types of input tokens for all providers and then instrumentations would do the math when necessary, e.g. something around these lines

```suggestion
        note: Regardless of the GenAI provider API, instrumentations SHOULD include all types of input tokens into `gen_ai.usage.input_tokens`.
```

model/gen-ai/registry.yaml
```diff
@@ -181,7 +181,29 @@ groups:
         stability: development
         type: int
         brief: The number of tokens used in the GenAI input (prompt).
+        note: |
+          This value MUST include cached tokens (cache reads and cache writes) when the provider exposes them.
```
lmolkova: I'm not sure we can always guarantee it

```suggestion
          This value SHOULD include cached tokens (cache reads and cache writes) when the provider exposes them.
```

model/gen-ai/registry.yaml
```diff
@@ -181,7 +181,29 @@ groups:
...
+          total effective input tokens the model processed.
         examples: [100]
+      - id: gen_ai.usage.cache_read.input_tokens
+        stability: development
+        type: int
+        brief: The number of input tokens served from a provider-managed cache.
+        note: >
+          The value is typically included in `gen_ai.usage.input_tokens`.
+          Semantic conventions for individual GenAI providers that do not include
+          cache read tokens in `gen_ai.usage.input_tokens` SHOULD document this.
```
```suggestion
        note: >
          The value SHOULD be included in `gen_ai.usage.input_tokens`.
```

to align with changes on `gen_ai.usage.input_tokens`

model/gen-ai/registry.yaml
```diff
@@ -181,7 +181,29 @@ groups:
...
+        examples: [50]
+      - id: gen_ai.usage.cache_creation.input_tokens
+        stability: development
+        type: int
+        brief: The number of input tokens written to a provider-managed cache.
+        note: >
+          The value is typically included in `gen_ai.usage.input_tokens`.
+          Semantic conventions for individual GenAI providers that do not include
+          cache creation tokens in `gen_ai.usage.input_tokens` SHOULD document this.
```
```suggestion
          The value SHOULD be included in `gen_ai.usage.input_tokens`.
```

model/gen-ai/registry.yaml
```diff
@@ -181,7 +181,29 @@ groups:
         stability: development
         type: int
         brief: The number of tokens used in the GenAI input (prompt).
+        note: |
+          This value MUST include cached tokens (cache reads and cache writes) when the provider exposes them.
+          Instrumentations SHOULD normalize provider responses so that `input_tokens` always reflects the
+          total effective input tokens the model processed.
```
lmolkova: I'm not sure we can always guarantee it


```suggestion
          This value SHOULD include all types of input tokens, including cached tokens.
          Instrumentations SHOULD make a best effort to populate this value, using a total
          provided by the   provider when available or, depending on the provider API, 
          by summing different token types parsed from the provider output.
```

model/gen-ai/spans.yaml
```diff
@@ -176,6 +180,14 @@ groups:
     attributes:
       - ref: gen_ai.request.model
         requirement_level: required
+      - ref: gen_ai.usage.input_tokens
+        note: >
+          OpenAI `input_tokens` already includes cached tokens.
+          Set `gen_ai.usage.input_tokens` directly from the provider response.
```
```suggestion
          The total input token count is returned by `usage.input_tokens` or a similar property in the model response.
```

model/gen-ai/spans.yaml
```diff
@@ -176,6 +180,14 @@ groups:
...
+      - ref: gen_ai.usage.cache_read.input_tokens
+        note: >
+          Subset of `gen_ai.usage.input_tokens`.
+          Populate from OpenAI `cached_tokens` field.
```
```suggestion
          Corresponds to `usage.input_tokens_details.cached_tokens` or a similar property in the model response.
```

model/gen-ai/spans.yaml
```diff
@@ -400,3 +412,29 @@ groups:
         requirement_level: required
       - ref: aws.bedrock.knowledge_base.id
         requirement_level: recommended
+
+  - id: span.anthropic.inference.client
```
lmolkova, line 446: we also need a markdown for it (it's not fully auto-generated). 

It can be done as a follow-up. If you want to add it here, please copy over something like https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/azure-ai-inference.md?plain=1 and replace `<!-- semconv span.azure.ai.inference.client -->` with the new group id (this is the only auto-generated part), plus make sure to update any handwritten text to talk about anthropic. Thanks!

lmolkova (approved):

lmolkova:
@verdie-g thanks for the contribution, everything looks great. I think you need to regenerate markdown tables and it'd be good to go!

.chloggen/genai_cache_tokens.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: enhancement
+component: gen-ai
+note: Add cache token attributes and provider-specific normalization guidance for GenAI usage metrics
+issues: [1959]
+subtext: |
+  - Add `gen_ai.usage.cache_read.input_tokens` attribute for tokens served from provider cache
```
stephentoub, line 6: Should these be recorded in metrics as well? Assuming yes, what should the gen_ai.token.type be?

lmolkova: @stephentoub not yet, there is no metric definition for it (no type defined). There are probably some design flaws we'll need to fix to make it happen. Created https://github.com/open-telemetry/semantic-conventions/issues/3341 so we don't forget

PR #3167, opened 2025-12-03
update param in `table-generation` Makefile command

marleypowell:
## Changes

Update `table-generation` Makefile command after spotting that there was no space after the `-D` param.

```
  -D, --param <PARAM>
          Parameters key=value, defined in the command line, to pass to the templates. The value must be a valid YAML value
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,7 +160,7 @@ table-generation:
 		--mount 'type=bind,source=$(PWD)/docs,target=/home/weaver/target' \
 		$(WEAVER_CONTAINER) registry update-markdown \
 		--registry=/home/weaver/source \
-		-D registry_base_url=/docs/registry/ \
+		--param registry_base_url=/docs/registry/ \
 		--templates=/home/weaver/templates \
 		--target=markdown \
 		--future \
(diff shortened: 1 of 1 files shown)


joaopgrassi:
Actually.. is this really a bug? 🤔 cc @lmolkova @jsuereth you folks know more about weaver features than I do.

Makefile
```diff
@@ -160,7 +160,7 @@ table-generation:
 		--mount 'type=bind,source=$(PWD)/docs,target=/home/weaver/target' \
 		$(WEAVER_CONTAINER) registry update-markdown \
 		--registry=/home/weaver/source \
-		-Dregistry_base_url=/docs/registry/ \
+		-D registry_base_url=/docs/registry/ \
```
joaopgrassi: There's another occurrence at line 198

marleypowell: Updated that one too

arminru (approved):

lmolkova:
-Dsomething is a typical way you'd pass Java system property, it works just fine in weaver (TIL `-D something` works as well since all checks are passing!), but indeed it looks a bit weird when used outside of Java ecosystem. 

I don't see a reason to fix it this way (with space) - it already works, but to make it less weird, I'd support changing to `--param registry_base_url=/docs/registry/`.

trask:
> I'd support changing to `--param registry_base_url=/docs/registry/`.

that looks a lot nicer to me (as a Java persion 😅)

marleypowell:
> -Dsomething is a typical way you'd pass Java system property, it works just fine in weaver (TIL `-D something` works as well since all checks are passing!), but indeed it looks a bit weird when used outside of Java ecosystem. 
> 
> 
> 
> I don't see a reason to fix it this way (with space) - it already works, but to make it less weird, I'd support changing to `--param registry_base_url=/docs/registry/`.  

TIL -Dsomething can also be valid!

I've changed it to `--param registry_base_url=/docs/registry/` if people think that is cleaner.

PR #3173, opened 2025-12-04
messaging: add messaging_latest_experimental opt-in value

### LINKED ISSUE #2928: Consider adding `messaging_latest_experimental` as a possible stabilitly opt-in value
### Area(s)

area:messaging

### What's missing?

In the OTel JS Contrib repo, we have a bunch of instrumentations that are using very outdated versions of semantic conventions. Every once in a while, motivated contributors come around that are interested in updating these instrumentation to the latest experimental semantic conventions (example: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2976). 

Unfortunately, we don't have a "safe" way forward for such changes at the moment as the messaging semconv only specifies [`messaging` and `messaging/dup`](https://github.com/open-telemetry/semantic-conventions/blob/6f59b508cfa6c2d50bb1a0eebf1152dae9e48164/docs/messaging/README.md?plain=1#L24-L32) as possible values to be used once the messaging semantic conventions become stable.

In the Gen AI Semantic Conventions' section about stability opt-in [there's a `gen_ai_latest_experimental` value](https://github.com/open-telemetry/semantic-conventions/blob/6f59b508cfa6c2d50bb1a0eebf1152dae9e48164/docs/gen-ai/README.md?plain=1#L21-L23) that can be set in `OTEL_SEMCONV_STABILITY_OPT_IN`. Having something like this for messaging (say `messaging_latest_experimental`) would 
- give us a way forward to update outdated instrumentations to the latest Semantic Conventions without immediately breaking users that rely on the old behavior
- help clearly communicate that breaking changes are expected when this value is set, helping with future updates until the messaging semantic conventions are stable.

### Tip

pichlermarc:
Fixes #2928 

## Changes

Adds `messaging_latest_experimental` as a possible value to `SEMCONV_STABILITY_OPT_IN` for instrumentations that want to update to the latest semconv without breaking end-users while messaging semconv is still in-development. 

The updated spec is also written in a way that once the messaging spec becomes stable, there's an opt-in flag for enabling the latest in-development semconv (attributes for instance) without having it on by default.

Note: I think the name of the flag could be aligned to be `messaging/latest-experimental` to follow a similar format to `messaging/dup`, but choose to align with the GenAI semconv for now where `genai_latest_experimental` is already used.

### DIFF
--- a/docs/messaging/README.md
+++ b/docs/messaging/README.md
@@ -24,9 +24,9 @@ This document defines semantic conventions for messaging systems spans, metrics 
 >   * `messaging` - emit the new, stable messaging conventions,
 >     and stop emitting the old experimental messaging conventions
 >     that the instrumentation emitted previously.
->   * `messaging_latest_experimental` - emit the latest experimental version of
-      Messaging semantic conventions (supported by the instrumentation) and do not emit the
-      old one (v1.24.0 or prior).
+>   * `messaging_latest_experimental` - emit the latest in-development messaging conventions
+>     and stop emitting the old experimental conventions (v1.24.0 or prior). If stable
+>     messaging conventions are available, emit both stable and latest in-development conventions.
 >   * `messaging/dup` - emit both the old and the stable messaging conventions,
 >     allowing for a seamless transition.
 >   * The default behavior (in the absence of one of these values) is to continue
--- a/docs/messaging/azure-messaging.md
+++ b/docs/messaging/azure-messaging.md
@@ -24,12 +24,17 @@ The Semantic Conventions for [Azure Service Bus](https://learn.microsoft.com/azu
 >   * `messaging` - emit the new, stable messaging conventions,
 >     and stop emitting the old experimental messaging conventions
 >     that the instrumentation emitted previously.
+>   * `messaging_latest_experimental` - emit the latest in-development messaging conventions
+>     and stop emitting the old experimental conventions (v1.24.0 or prior). If stable
+>     messaging conventions are available, emit both stable and latest in-development conventions.
 >   * `messaging/dup` - emit both the old and the stable messaging conventions,
 >     allowing for a seamless transition.
 >   * The default behavior (in the absence of one of these values) is to continue
 >     emitting whatever version of the old experimental messaging conventions
 >     the instrumentation was emitting previously.
->   * Note: `messaging/dup` has higher precedence than `messaging` in case both values are present
+>   * Notes:
+>     * `messaging/dup` has higher precedence than `messaging` in case both values are present
+>     * `messaging_latest_experimental` has higher precedence than both `messaging/dup` and `messaging`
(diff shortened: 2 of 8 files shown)


docs/messaging/README.md
```diff
@@ -24,12 +24,17 @@ This document defines semantic conventions for messaging systems spans, metrics
...
+      old one (v1.24.0 or prior).
 >   * `messaging/dup` - emit both the old and the stable messaging conventions,
 >     allowing for a seamless transition.
 >   * The default behavior (in the absence of one of these values) is to continue
 >     emitting whatever version of the old experimental messaging conventions
 >     the instrumentation was emitting previously.
->   * Note: `messaging/dup` has higher precedence than `messaging` in case both values are present
+>   * Notes:
+>     * `messaging/dup` has higher precedence than `messaging` in case both values are present
+>     * `messaging_latest_experimental` has higher precedence than both `messaging/dup` and `messaging`
```
JamieDanielson, line 37: It hasn't occurred yet, but when some messaging conventions do stabilize, we will be in a situation where someone wants both stable conventions where available and latest experimental where not yet stable. I'm not sure if this is just being pedantic, but we may want to consider specifying the behavior in that scenario. Would I use `messaging_latest_experimental,messaging/dup`? Do I assume `messaging_latest_experimental` includes stable conventions?

pichlermarc: It's a good point - not pedantic at all. I wanted to leave it out and possibly expand on it later, but it's likely best to do it now 👍 

Added additional wording in https://github.com/open-telemetry/semantic-conventions/pull/3173/commits/09166670a8ccc40d71dec4388d6245a3f5fc54bd

JamieDanielson (approved):

docs/messaging/azure-messaging.md
```diff
@@ -24,12 +24,17 @@ The Semantic Conventions for [Azure Service Bus](https://learn.microsoft.com/azu
 >   * `messaging` - emit the new, stable messaging conventions,
```
joaopgrassi, line 24: Should we be crystal clear here and say that this only emits stable conventions? Because it says "stop emitting the old, experimental. I kinda don't like "old" here because there might be "new" unstable conventions as well, and are those included or not? :D 

Maybe we just say: emit only stable messaging conventions. and keep it short. 

Could also be done as a follow up. I just wanted to bring here now for discussions. CC @open-telemetry/specs-semconv-maintainers

pichlermarc: > I kinda don't like "old" here because there might be "new" unstable conventions as well, and are those included or not? :D

Heh, good point! I'm happy to follow-up on it if we decide to go that route. 🙂 

> Maybe we just say: emit only stable messaging conventions. and keep it short.

I like that. Maybe that small change could be the resolution for #3200 🙂

Issue #2921, opened 2025-10-16
Decide on which `rpc.protocol.name` values to mark stable as part of initial RPC semconv stability


trask:
Tracking / blocking issue for initial RPC semconv stability.

trask:
Candidates:

## gRPC

Both a framework and a protocol

## Dubbo

A framework that supports multiple protocols

- gRPC
- [Dubbo2](https://dubbo.apache.org/en/overview/reference/protocols/tcp/)
- [Triple](https://dubbo.apache.org/en/overview/reference/protocols/triple/), part of Dubbo 3 framework, replacing Dubbo2 protocol

## ConnectRPC

- gRPC
- gRPC-web
- [Connect Protocol](https://connectrpc.com/docs/protocol/)

## JSON RPC

https://www.jsonrpc.org/specification

Not a framework

RPC protocol that supports multiple wire protocols

AlbumenJ:
Dubbo (as an SDK) supports both the Dubbo protocol and the Triple protocol. The Dubbo protocol has been supported since the Dubbo 2.0 (SDK) version, while the Triple protocol was introduced in the Dubbo 3.0 (SDK) version.

The Dubbo protocol is a binary protocol that runs directly on top of TCP. Its structure can be found at [https://dubbo.apache.org/en/overview/reference/protocols/tcp/](https://dubbo.apache.org/en/overview/reference/protocols/tcp/).

We consider the Triple protocol to be more complex. Simply put, the Triple protocol is a "large" protocol that runs on top of HTTP and supports multiple transport types.

When the Triple protocol acts as a server, it supports exposing services via HTTP/1 (including SSE, WebSocket, etc.), HTTP/2, and HTTP/3. Regarding the `content-type` support specific to HTTP, we support formats including, but not limited to, `application/json`, `text/xml`, `text/plain`, `application/octet-stream`, and `application/grpc`. For `application/grpc`, in addition to the native gRPC protocol, we also support scenarios where users do not have Protobuf IDL definitions.

When the Triple protocol acts as a client, we only support requests via HTTP/2 or HTTP/3 using the `application/grpc` content type.

The reason for this design is that when Dubbo is exposed as a server for **north-south** traffic, it can be highly compatible with all clients. When Dubbo acts as a node for **east-west** traffic, the high-performance communication it provides by default can meet the requirements of most requests within a data center.

[...]

PR #3176, opened 2025-12-04
RPC: align with naming best practices

### LINKED ISSUE #2703: RPC: adopt system-specific naming practices

lmolkova:
Fixes #2703

See also: #2921

This PR:
- renames `rpc.system` to `rpc.system.name`. Note: it also updates enum members:
  - `connect_rpc` is renamed to `connectrpc`
  - `apache_dubbo` is renamed to `dubbo`
  - `java_rmi`, `dotnet_wcf`, and `onc_rpc` are not included in the new enum - we don't have proper conventions for them and can add them to the new enum later on.
- renames `rpc.jsonrpc.request_id` to `jsonrpc.request.id` and `rpc.jsonrpc.version` to `jsonrpc.protocol.version`

AWS conventions are intentionally left with deprecated `rpc.system` - https://github.com/open-telemetry/semantic-conventions/issues/1545, AWS is not an RPC system and the assumption is that AWS conventions will eventually be fixed.


docs/registry/attributes/README.md
```diff
@@ -73,6 +73,7 @@ Currently, the following namespaces exist:
 - [Host](host.md)
 - [HTTP](http.md)
 - [iOS](ios.md)
+- [JSONRPC](jsonrpc.md)
```
lmolkova, line 76: can't easily do JSON-RPC because of the tooling limitations

docs/object-stores/s3.md
```diff
@@ -23,7 +23,7 @@ Semantic Conventions for AWS S3 client spans extend the general [AWS SDK Semanti
 
 | Key | Stability | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Value Type | Description | Example Values |
 | --- | --- | --- | --- | --- | --- |
-| [`rpc.system`](/docs/registry/attributes/rpc.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | The value `aws-api`. | `aws-api` |
+| [`rpc.system`](/docs/registry/attributes/rpc.md) | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `rpc.system.name`. | `Required` | string | The value `aws-api`. | `aws-api` |
```
trask, line 26: this reads a little confusing, I think because it's not really deprecated in this particular context, I don't have a better idea though

docs/registry/attributes/rpc.md
```diff
@@ -58,6 +56,8 @@ the `rpc.grpc.response.metadata.my-custom-key` attribute with value `["attribute
 **[6] `rpc.response.status_code`:** Usually it represents an error code, but may also represent partial success, warning, or differentiate between various types of successful outcomes.
 Semantic conventions for individual RPC frameworks SHOULD document what `rpc.response.status_code` means in the context of that system and which values are considered to represent errors.
 
+**[7] `rpc.system.name`:** The client and server RPC systems may differ for the same RPC interaction. For example, a client may use Apache Dubbo or Connect RPC to communicate with a server that uses gRPC since both protocols provide compatibility with gRPC.
```
trask: 👍

docs/registry/attributes/rpc.md
```diff
@@ -69,17 +69,14 @@ Semantic conventions for individual RPC frameworks SHOULD document what `rpc.res
...
 | --- | --- | --- |
-| `apache_dubbo` | Apache Dubbo | ![Development](https://img.shields.io/badge/-development-blue) |
-| `connect_rpc` | Connect RPC | ![Development](https://img.shields.io/badge/-development-blue) |
-| `dotnet_wcf` | .NET WCF | ![Development](https://img.shields.io/badge/-development-blue) |
-| `grpc` | gRPC | ![Development](https://img.shields.io/badge/-development-blue) |
-| `java_rmi` | Java RMI | ![Development](https://img.shields.io/badge/-development-blue) |
-| `jsonrpc` | JSON-RPC | ![Development](https://img.shields.io/badge/-development-blue) |
-| `onc_rpc` | [ONC RPC (Sun RPC)](https://datatracker.ietf.org/doc/html/rfc5531) | ![Development](https://img.shields.io/badge/-development-blue) |
+| `connectrpc` | [Connect RPC](https://connectrpc.com/) | ![Development](https://img.shields.io/badge/-development-blue) |
+| `dubbo` | [Apache Dubbo](https://dubbo.apache.org/) | ![Development](https://img.shields.io/badge/-development-blue) |
```
steverao, line 63: Don't prefer to add organization prefix here? According to my knowledge, we may want to distinguish with `alibaba_dubbo` here. From 2.7+, the project was donated to apache foundation and renamed to `apache_dubbo`. Due to historical reason, previous versions still use alibaba related package name called `alibaba_dubbo`. For otel-java-instrumentation, we only supported `apache_dubbo`. 
Do you have any comment here? @AlbumenJ

AlbumenJ: `Dubbo` is the overarching project name, while `alibaba_dubbo` and `apache_dubbo` represent different stages of its evolution. The term `Dubbo` encompasses both versions.

AlbumenJ (approved):

model/jsonrpc/registry.yaml
```diff
@@ -0,0 +1,28 @@
...
+          corresponding response.
+        examples: ['10', 'request-7']
+        note: >
+          The `id` property may be a string, number, null, or omitted entirely.
+          When omitted, the request is treated as a notification.
+          Using `null` is not equivalent to omitting the `id`, but it is
+          [discouraged](https://www.jsonrpc.org/specification).
+
+          Instrumentations SHOULD NOT capture this attribute when the `id` is
+          `null` or omitted.
```
matt-hensley: I think explicitly calling out the difference in handling the value would improve clarity. Something like:

```suggestion
          Under the
          [JSON-RPC specification](https://www.jsonrpc.org/specification),
          the `id` property may be a string, number, null, or omitted entirely.
          When omitted, the request is treated as a notification.
          Using `null` is not equivalent to omitting the `id`, but it is
          discouraged.

          Instrumentations SHOULD NOT capture this attribute when the `id` is
          `null` or omitted.
```

docs/registry/attributes/rpc.md
lmolkova: Given that `alibaba_dubbo` is a legacy, I'd prefer to give priority to apache `dubbo` and, if necessary, still use `alibaba_dubbo` for the old version (or have some other means to distinguish).

Issue #1798, opened 2025-01-23
Change `process.open_file_descriptors`


braydonk:
### Area(s)

area:process

### What's missing?

Note: first discussed in https://github.com/open-telemetry/semantic-conventions/pull/1797

`process.open_file_descriptors` has 2 problems:

1. The `open_` is arguably redundant, since a "file descriptor" is itself a representation of an open file (i.e. there's no such thing as counting `closed_file_descriptors`)
2. It can be confusing to Windows users, since on Windows you'd be counting `Handles`, a distinct windows exclusive concept, rather than `file_descriptors`.

### Describe the solution you'd like

I suggest renaming the metric to `process.file_descriptors`, or `process.unix.file_descriptors`. At first glance I like `unix` being in the name, but this would be the first time we use that as a platform name so we'll need to think a bit about that before making that jump.

There will also be a metric called `process.windows.handles` in the future. Hopefully the presence of this metric will alleviate the platform confusion.

braydonk:
Marking as GA Blocker since `open_file_descriptors` is an existing metric in use by instrumentation today.

lmolkova:
nit, since it's an up-down counter, we have a naming rule to avoid pluralization 

https://github.com/open-telemetry/semantic-conventions/blob/21c230d03faed5d2928b66892b59ae368e952baa/docs/general/naming.md?plain=1#L252-L260

So probably `process.file_descriptor.count`?

I'm a Windows user, and my subjective and not very-well though out reaction is to use the same metric if there are no substantial differences except name. 

I also wonder if we could be more approachable and just call it `process.open.file.count`

trask:
> `process.open_file.count`

I think you can have multiple open file descriptors for the same file, so could be misleading?

`process.open_file_handle.count` doesn't seem like the worst idea to me

from https://en.wikipedia.org/wiki/File_descriptor (my emphasis):

> In [Unix](https://en.wikipedia.org/wiki/Unix) and [Unix-like](https://en.wikipedia.org/wiki/Unix-like) computer operating systems, a file descriptor is a process-unique identifier ([**handle**](https://en.wikipedia.org/wiki/Handle_(computing))) for a [file](https://en.wikipedia.org/wiki/File_(computing))

I think I'd also be ok with keeping them as separate metrics in order to stay closer to the domain-specific terminology, e.g.

* `unix.process.open_file_descriptor.count`
* `windows.process.open_file_handle.count`

[...]

PR #3188, opened 2025-12-08
Introduce process.unix.file_descriptor.count and process.windows.handle.count metrics

ChrsMark:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1798

## Changes

This PR adds `process.unix.file_descriptor.count` and `process.windows.handle.count` metrics based on the decision from   https://github.com/open-telemetry/semantic-conventions/issues/1798.
In addition, the `process.open_file_descriptor.count` is deprecated accordingly.

Implementations:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/7c1b21623e1fce40e4989caf06eb67b6b5225481/receiver/hostmetricsreceiver/internal/scraper/processscraper/metadata.yaml#L180
- https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/7c1b21623e1fce40e4989caf06eb67b6b5225481/receiver/hostmetricsreceiver/internal/scraper/processscraper/process_scraper.go#L440-L453


docs/system/process-metrics.md
```diff
@@ -251,7 +253,41 @@ This metric is [recommended][MetricRecommended].
 
 | Name | Instrument Type | Unit (UCUM) | Description | Stability | Entity Associations |
 | -------- | --------------- | ----------- | -------------- | --------- | ------ |
-| `process.open_file_descriptor.count` | UpDownCounter | `{file_descriptor}` | Number of file descriptors in use by the process. | ![Development](https://img.shields.io/badge/-development-blue) | [`process`](/docs/registry/entities/process.md#process) |
+| `process.open_file_descriptor.count` | UpDownCounter | `{file_descriptor}` | Deprecated, use `process.unix.file_descriptor.count` instead. | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `process.unix.file_descriptor.count`. | [`process`](/docs/registry/entities/process.md#process) |
```
thompson-tomo: Should the deprecated metric be removed likely we usually do? Alternatively replaced the recommended requirement level with deprecated text.

ChrsMark: Right, I missed removing that.

Is there a place where it should be moved instead? I couldn't find any similar example 🤔

thompson-tomo: Currently deprecated signals are just removed from markdown which is what we did when we deprecated metrics for rpc.

docs/system/process-metrics.md
```diff
@@ -240,18 +241,35 @@ This metric is [recommended][MetricRecommended].
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
-### Metric: `process.open_file_descriptor.count`
+### Metric: `process.unix.file_descriptor.count`
```
lmolkova, line 244: following [`system.memory.linux.available`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemmemorylinuxavailable) and

https://github.com/open-telemetry/semantic-conventions/blob/e2e30b54b9a8af6a81d51f5ecc3ccbcbe63cd2be/docs/non-normative/groups/system/design-philosophy.md?plain=1#L304

 should this be `process.file_descriptor.unix.count` ?

braydonk: I think this is the right place for the OS name. I explained the distinction for why it moved in the `system.memory` case here: https://github.com/open-telemetry/semantic-conventions/pull/2996#issuecomment-3517001572

We could probably write it a bit more clearly in our guidance, but in this case `process` is the root level of concern not `process.file_descriptor`.

PR #3191, opened 2025-12-09
[chore] infra: switch to mdlint cli2

thompson-tomo:
## Changes

This switches to markdowndlint-cli2 which enables us to:
- use the github action in workflows
- receive annotations on files in pr's for violations which caused the pipeline to fail previously they were only in the logs

All existing configuration is natively supported hence is almost a direct drop-in.

This also ensures that the vscode extension uses the same configuration. If extension is not installed it is recommended.

> [!IMPORTANT]


Makefile
```diff
@@ -138,8 +138,8 @@ markdown-toc:
 
 .PHONY: markdownlint
 markdownlint:
-	@if ! npm ls markdownlint-cli; then npm ci --ignore-scripts; fi
-	npx --no -- markdownlint-cli -c .markdownlint.yaml "**/*.md" --ignore ./.github/ --ignore ./node_modules/ --ignore ./.git/
+	@if ! npm ls markdownlint-cli2; then npm ci --ignore-scripts; fi
+	npx --no -- markdownlint-cli2 '**/*.md'
```
thompson-tomo: Replace with npm script call

And ensure that the below is set in vscode settings.

```
"npm.enableScriptExplorer": true
```

.vscode/settings.json
```diff
@@ -9,5 +9,6 @@
         ]
     },
     "json.schemaDownload.enable": true,
-    "markdown.extension.toc.levels": "2..6"
+    "markdown.extension.toc.levels": "2..6",
+    "markdownlint.configFile": ".markdownlint.yaml"
```
thompson-tomo:
```suggestion
    "markdownlint.configFile": ".markdownlint.yaml",
    "npm.enableScriptExplorer": true
```

package.json
thompson-tomo: Need to add a check:lint:markdown script and also a fix:lint:markdown script.

.gitignore
```diff
@@ -29,9 +29,6 @@ internal/tools/bin
 # Node.js files for tools (e.g. markdown-toc)
 node_modules/
 
-# Visual Studio Code
-.vscode
```
jsuereth, line 33: This could suddenly muck up with folks vscode settings.

For @open-telemetry/specs-semconv-maintainers - Should we be proposing project-level vscode settings for this project via direct check-ins?

I've had very poor experiences with churn and subtle per-computer differences causing issues here on other IDEs. I don't have experience with that on vscode *specifically* because I've always seen it in .gitignore.

Personally, I'm "softly" against including IDE-specific extensions *by default*. If we wanted to provide a bootstrap for vscode, I think forcing folks to go through the extra hoop of manually including the file via "force" commands in git is working as designed, as IDE setup can be a personal holy war I'd rather not have debated in our PRs, and more-so, at least for me, there are work-related restrictions on what extensions can be used.

Also to be crystal clear - I'm *only* soft-blocking the removal of `.vscode` in `.gitignore`.  I think the updates to the `.vscode` directory you're providing here are fine.

jsuereth (commented):
Absolutely love getting markdown lint comments in GH pull requests, so I'd be a fan of switching.

One comment around the `.gitignore` changes and my rationale.

.gitignore
thompson-tomo: Have no issue adding it back if desired but keeping the changes.

> there are work-related restrictions on what extensions can be used

Note This does not install them automatically it just notifies you that it is suggested/recommended to install these extensions.

PR #3193, opened 2025-12-09
v2: Remove common attributes snippets in hw metrics

lmolkova:
In preparation for v2 schema, we're trying to see where public attribute groups are truly necessary.
With schema v2, we'll force ourselves to be more expressive and strict in yaml and won't rely on free-form text in markdown.

In HW metrics, we render common metrics attributes in addition to per-metric attributes. 
Based on the current description ("All * metrics may include the below attributes") it's not clear whether all or any of the particular attributes should be included in any of the metrics below. Some of the metrics extend the group and others don't.

Given this, I'm proposing to remove rendering of attribute group snippets.


lmolkova:
@trisch-me could you please help me understand if some of these removed attributes should be reported on metrics define under them in MD? Thanks!

trisch-me:
hey @lmolkova 
common attributes are a part of particular metric for specific hw type, so common metrics for battery are applicable for all battery metrics. 
I’m not sure how it should be provided in v2 schema, but those shouldn’t be just deleted I suppose
When you say some of them are extending the group and some are not - could you give examples? All metrics should extend  (include) common metrics for that particular type and they also should include basic attributes

lmolkova:
@trisch-me check out individual metrics definitions, e.g. https://github.com/open-telemetry/semantic-conventions/blob/main/docs/hardware/battery.md#metric-hwbatterycharge - this removed attribute table is repeated there because the corresponding metric definition  in yaml is extending that group:

https://github.com/open-telemetry/semantic-conventions/blob/18db3e4428b6a162838dbe10f129e992a2e121de/model/hardware/battery-metrics.yaml#L28 

The only definition that has a different set of attributes (for battery) is [`hw.status` (Battery)](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/hardware/battery.md#metric-hwstatus-battery). Based on the current language it's not clear if this metric should have all or some of the common attributes.

trisch-me:
So there is a set of [common attributes](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/hardware/common.md) that are used for every hw metrics, some of them are required such as `hw.id` or `hw.type`, others are recommended.
Every component has also a set of attributes applicable to every metric
i.e.

```
  # BATTERY METRICS
  - id: metric_attributes.hw.battery
    ...
    extends: hardware.attributes.common <-- common attributes for every metric

- id: metric.hw.battery.charge
    ...
    extends: metric_attributes.hw.battery <-- common attributes for battery metrics
```

`hw.status` is special case, every component has it’s own values for `hw.type` and `hw.state` and it’s not encoded in yaml but only in md files for now, because we had no means to redefine it in yaml
examples: [temperature](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/hardware/temperature.md#metric-hwstatus-temperature), [battery](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/hardware/battery.md#metric-hwstatus-battery)

PR #3197, opened 2025-12-09
Use consistent set of attributes between RPC spans and metrics

### LINKED ISSUE #2922: We should tailor generic note on `error.type` to RPC case. consistently across spans and metrics.
We should tailor generic note on `error.type` to RPC case. consistently across spans and metrics.

_Originally posted by @lmolkova in https://github.com/open-telemetry/semantic-conventions/pull/2852#discussion_r2433545798_

lmolkova:
Fixes #2922

Use consistent set of attributes between RPC spans and metrics
  - Use `rpc.response.status_code` on common metrics and spans.
  - Make `error.type` note consistent between metrics and spans. 
  - Promote notes on `rpc.method` and `rpc.service` to attribute definition.

### DIFF
--- a/.chloggen/rpc-consistent-attributes-metrics-spans.yaml
+++ b/.chloggen/rpc-consistent-attributes-metrics-spans.yaml
@@ -1,7 +1,7 @@
 change_type: enhancement
 component: rpc
 note: Use consistent set of attributes between RPC spans and metrics
-issues: [2922, TODO]
+issues: [2922, 3197]
 subtext: |
   - Use `rpc.response.status_code` on common metrics and spans.
   - Make `error.type` note consistent between metrics and spans.


model/rpc/metrics.yaml
```diff
@@ -1,4 +1,30 @@
 groups:
+  - id: attributes.metrics.rpc.client
```
joaopgrassi, line 2: are these new groups for future use?

joaopgrassi (approved):

model/rpc/metrics.yaml
lmolkova: they are already used in metrics below, will optimize them (and minimize nesting) when transitioning to v2

PR #3220, opened 2025-12-16
[chore] Align stale behaviour and stale comment

thompson-tomo:
## Changes 

With https://github.com/open-telemetry/semantic-conventions/commit/b0ee8f8d7c8ffdb42640bac43f178b779fc68804 the message for stale pr's was changed to 14 days however that change in duration was not intended. This brings the message back in alignment with the behaviour.

> [!IMPORTANT]

### DIFF
--- a/.github/workflows/stale-pr.yml
+++ b/.github/workflows/stale-pr.yml
@@ -19,12 +19,12 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           stale-pr-message: >
             This PR has been labeled as stale due to lack of activity.
-            It will be automatically closed if there is no further activity over the next 14 days.
+            It will be automatically closed if there is no further activity over the next 7 days.
           exempt-pr-labels: 'never stale'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-pr.yml
```diff
@@ -26,5 +26,5 @@ jobs:
           days-before-close: -1
           # overrides the above only for pull requests
           days-before-pr-stale: 14
-          days-before-pr-close: 7
+          days-before-pr-close: 14
```
arminru: I assume this was more of an oversight and not the intention of #3137.

@trask Did you intend to change the overall duration from last activity to closing the PR to 28 days without any interaction, or should it stay at 21 days (14 for the message + 7 more to close if there is no engagement)? In that case, we'd have to fix the message instead.

trask: oh yeah, we should just fix the message, thanks

arminru (approved):

lmolkova (approved):

PR #3222, opened 2025-12-16
GenAI: added reasoning tokens usage attribute

### LINKED ISSUE #3194: Add `usage.reasoning_tokens` to span attributes
### Area(s)

area:gen-ai

### Propose new conventions

Since `reasoning` tokens are a internal part of a chat run, they are neither input nor really output, it would help to clarify what the model did when the spec also adds reasoning tokens, as a separate category, next to input and output.

Alternatives could be `thinking`, which is what Google uses, for Gemini models, but probably `reasoning` makes more sense, since there are also reasoning message as input/output content conventions.

So concretely:
add:
`gen_ai.usage.reasoning_tokens`
as a optional span attributes for `chat`

### Tip

eavanvalkenburg:
Fixes #3194 

## Changes

Adds reasoning tokens as a attribute to the `gen_ai.usage` field.


model/gen-ai/registry.yaml
```diff
@@ -187,6 +187,11 @@ groups:
         type: int
         brief: The number of tokens used in the GenAI response (completion).
         examples: [180]
+      - id: gen_ai.usage.reasoning_tokens
+        stability: development
+        type: int
+        brief: The number of tokens used in the GenAI reasoning or thinking.
```
verdie-g: We should probably explicit if the reasoning tokens are included in `gen_ai.usage.output_tokens` or not.

eavanvalkenburg: That's a great point, updated to reflect this.

docs/gen-ai/aws-bedrock.md
```diff
@@ -63,6 +63,7 @@ Describes an AWS Bedrock operation span.
...
 | [`gen_ai.request.model`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` If available. | string | The name of the GenAI model a request is being made to. [7] | `gpt-4` |
 | [`gen_ai.request.seed`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` if applicable and if the request includes a seed | int | Requests with same seed value more likely to return same result. | `100` |
+| [`gen_ai.usage.reasoning_tokens`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` when available | int | The number of tokens used in the GenAI reasoning or thinking, only set this if the GenAI model provides a separate count for reasoning tokens and they are not also included in input or output tokens. | `180` |
```
stephentoub, line 66: Why would reasoning tokens only be set if not included in output count? I think it'd be better to just state whether they're counted as part of output or not (ideally yes, so that output is consistent regardless of whether reasoning tokens are separated out by a provider), and then an implementation can do the math to ensure the values match the convention based on whether the provider includes reasoning in the output it reports.

alexmojaki: Agreed, they should be a subset of output tokens.

lmolkova: related:  https://github.com/open-telemetry/semantic-conventions/pull/3163

model/gen-ai/registry.yaml
```diff
@@ -187,6 +187,12 @@ groups:
         type: int
         brief: The number of tokens used in the GenAI response (completion).
         examples: [180]
+      - id: gen_ai.usage.reasoning_tokens
```
lmolkova, line 190: Suggesting this pattern in https://github.com/open-telemetry/semantic-conventions/pull/3163/changes#r2705621129


```suggestion
      - id: gen_ai.usage.reasoning.output_tokens
```

verdie-g:
I recreated the PR there https://github.com/open-telemetry/semantic-conventions/pull/3383

PR #3223, opened 2025-12-16
Merge `rpc.method` and `rpc.service` into fully-qualified `rpc.method`, clarify span name and possible high cardinality 

### LINKED ISSUE #2863: Update rpc span name guidance
### Area(s)

area:rpc

### What's missing?

A span name convention which caters for no service being known ie json-rpc and also enables streaming messages to be identified

A span naming convention similiar to that of the messaging namespace where the operation and target are combined to define the name.

### Tip

### LINKED ISSUE #3196: RPC: should we merge `rpc.service` and `rpc.method` into one fully qualified `rpc.method.name`
Some data points:

## Dashboards/alerts 

Is there a point in breaking things down by service name only? 

I don't think it's the main use-case. Method within the same service could have quite different perf characteristics. 

If we merge service/method, it should still be possible to have startwith / regex filters on attribute value for corner cases where it makes sense.

## Consistency with other semantic conventions

RPC service + method is analogous to `http.route` which is fully qualified.

We deprecated `function.namespace` and `function.name` in favor of fully-qualified [`code.function.name`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/code.md#code-function-name)

----------------
## gRPC:

OTel Java gRPC instrumentation splits full method name into parts: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/fcf9a9ad2b577b9bf7d6ad2c0cf8ec463684f4c8/instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/GrpcRpcAttributesGetter.java#L27

 Official gRPC tracing plugin doesn't set either of those attributes, but populates span name
Image

There are no formal docs that would consider service and method to be independent and interesting without each other

## Connect RPC 
no otel instrumentations(?), it's based on and is compatible with gRPC

## Dubbo: 

method name and interface name are different https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/fcf9a9ad2b577b9bf7d6ad2c0cf8ec463684f4c8/instrumentation/apache-dubbo-2.7/library-autoconfigure/src/main/java/io/opentelemetry/instrumentation/apachedubbo/v2_7/DubboRpcAttributesGetter.java#L20

## JSON-RPC

method is fully qualified, no notion of service https://www.jsonrpc.org/specification#request_object

lmolkova:
Fixes #3196, fixes #2863

- Deprecate `rpc.method` and `rpc.service` in favor of new `rpc.method.name`
- Add note on restricting that new attribute to recognized methods (similarly to `http.request.method`)
- Introduce `rpc.method.name_original`  similarly to `http.request.method_original`
- Update span name section to use new attribute

### DIFF
--- a/docs/rpc/rpc-spans.md
+++ b/docs/rpc/rpc-spans.md
@@ -51,13 +51,13 @@ This document defines how to describe remote procedure calls
 
 ### Span name
 
-RPC spans MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.51.0/specification/trace/api.md#span).
+RPC spans MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.52.0/specification/trace/api.md#span).
 
-The *span name* SHOULD be `{rpc.method.name}` if it is
-available and not set to `_OTHER`.
+The *span name* SHOULD be `{rpc.method}` if it is available and not set to
+`_OTHER`.
 
-If `rpc.method.name` is unavailable or set to `_OTHER`,
-the span name SHOULD be `{rpc.system.name}`.
+If `rpc.method` is unavailable or set to `_OTHER`, the span name SHOULD be
+`{rpc.system.name}`.
 
 Semantic conventions for individual RPC systems MAY specify different span name
 format.


model/rpc/registry.yaml
```diff
@@ -40,28 +40,48 @@ groups:
           For example, a property `my-custom-key` with value `["attribute_value"]` SHOULD be recorded as
           the `rpc.response.metadata.my-custom-key` attribute with value `["attribute_value"]`
         examples: [["attribute_value"]]
-      - id: rpc.method
+      - id: rpc.method.name
```
lmolkova: alternative (closer to HTTP): `rpc.request.method` (and `rpc.request.method_original`)

docs/rpc/rpc-spans.md
```diff
@@ -52,37 +51,15 @@ This document defines how to describe remote procedure calls
...
-
-Examples of span names:
-
-- `grpc.test.EchoService/Echo`
-- `com.example.ExampleRmiService/exampleMethod`
-- `MyCalcService.Calculator/Add` reported by the server and
-  `MyServiceReference.ICalculator/Add` reported by the client for .NET WCF calls
-- `MyServiceWithNoPackage/theMethod`
-
-### Service name
```
lmolkova, line 73: I don't think this section is helpful here now since we're removing `rpc.service` along with source of possible confusion with `service.name`.

model/rpc/registry.yaml
lmolkova: SIG discussion:
- http method is just a verb
- rpc method is more rich and might have more properties (e.g. streaming type)

docs/db/dynamodb.md
```diff
@@ -111,8 +111,8 @@ This span represents a `DynamoDB.BatchWriteItem` call.
...
-| [`rpc.method`](/docs/registry/attributes/rpc.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | The name of the operation corresponding to the request, as returned by the AWS SDK | `GetItem`; `PutItem` |
-| [`rpc.service`](/docs/registry/attributes/rpc.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | The name of the service to which a request is made, as returned by the AWS SDK. | `DynamoDB`; `S3` |
+| [`rpc.method`](/docs/registry/attributes/rpc.md) | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Value should be included in `rpc.method.name` which is expected to be a fully-qualified name. | `Recommended` | string | The name of the operation corresponding to the request, as returned by the AWS SDK | `GetItem`; `PutItem` |
+| [`rpc.service`](/docs/registry/attributes/rpc.md) | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Value should be included in `rpc.method.name` which is expected to be a fully-qualified name. | `Recommended` | string | The name of the service to which a request is made, as returned by the AWS SDK. | `DynamoDB`; `S3` |
```
maryliag: It feels like is missing a change here, there is a lot of updates to deprecated and to use the new `rpc.method.name`, but I don't see a new entry for `rpc.method.name` being added to these tables. Is that correct?

docs/rpc/rpc-spans.md
```diff
@@ -52,37 +51,15 @@ This document defines how to describe remote procedure calls
...
-The *span name* MUST be the full RPC method name formatted as:
+RPC spans MUST follow the overall [guidelines for span names](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.51.0/specification/trace/api.md#span).
 
-```
-$package.$service/$method
-```
+RPC span names SHOULD be `{rpc.method.name}` if there is `rpc.method.name`
+available and it's not set to `_OTHER`.
+In other cases (when `{rpc.method.name}` is set to `_OTHER` or is not available),
+span name SHOULD be `{rpc.system.name}`.
```
matt-hensley:
```suggestion
The *span name* SHOULD be `{rpc.method.name}` if it is
available and not set to `_OTHER`.

If `rpc.method.name` is unavailable or set to `_OTHER`,
the span name SHOULD be `{rpc.system.name}`.
```

model/rpc/registry.yaml
lmolkova: Another alternative: `rpc.method` and `rpc.method_original`

----
Some discussion points:
- no other request properties on RPC, we also operate on call level here (so `request` doesn't make much sense)
- method might have other properties (type, idempotency, but that could be framework specific - e.g. `grpc.method.idempotency` and it's a theoretical concern)
- downside of `rpc.method` is reusing the same attribute for new purpose , harder when migrating
- `rpc.method` is shorter but is as clear as `rpc.method.name`

lmolkova: back to rpc.method - https://github.com/open-telemetry/semantic-conventions/pull/3223/commits/b8d10e49f70cd20d0500306c5ec6324246f07cbc

docs/db/dynamodb.md
lmolkova: thanks for checking! AWS REST API is not really an RPC system and it should not include RPC semconv - https://github.com/open-telemetry/semantic-conventions/issues/1545, so the long-term story for them is to switch to some other attributes. Switching to new RPC semconv does not make sense here. 

So, it's intentional that AWS conventions depend on deprecated attributes (and that corresponding instrumentations keep using them) - this would be a blocker for AWS stability though.

maryliag: Thanks for the explanation!

matt-hensley (approved):

PR #3226, opened 2025-12-17
Separate RPC events from spans

lmolkova:
Following [OTEP-265](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0265-event-vision.md) and [4430](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4430-span-event-api-deprecation-plan.md) span events are deprecated.

Events in semantic conventions are intended for [log-based events](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/events.md) and there is effectively no means to record span events.

This PR separates events from spans in model and markdown. When stabilizing RPC we also plan to stabilize spans, but not events/

### DIFF
--- a/docs/rpc/rpc-events.md
+++ b/docs/rpc/rpc-events.md
@@ -8,6 +8,31 @@ linkTitle: Events
 
 This document defines events applicable in the context of remote procedure calls.
 
+> [!IMPORTANT]
+> Existing RPC instrumentations that are using
+> [v1.37.0 of this document](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/rpc/rpc-spans.md#events)
+> (or prior):
+>
+> * SHOULD NOT change the version of the RPC conventions that they emit by
+>   default in their existing major version. Conventions include (but are not
+>   limited to) attributes, metric and span names, and unit of measure.
+> * SHOULD introduce an environment variable `OTEL_SEMCONV_STABILITY_OPT_IN`
+>   in their existing major version as a comma-separated list of category-specific values
+>   (e.g., http, databases, rpc). The list of values includes:
+>   * `rpc` - emit the stable RPC conventions, and stop emitting
+>     the experimental RPC conventions that the instrumentation emitted
+>     previously.
+>   * `rpc/dup` - emit both the experimental and stable RPC conventions,
+>     allowing for a phased rollout of the stable semantic conventions.
+>   * The default behavior (in the absence of one of these values) is to continue
+>     emitting whatever version of the old experimental RPC conventions
+>     the instrumentation was emitting previously.
+>   * Note: `rpc/dup` has higher precedence than `rpc` in case both values are present
+> * SHOULD maintain (security patching at a minimum) their existing major version
+>   for at least six months after it starts emitting both sets of conventions.
+> * MAY drop the environment variable in their next major version and emit only
+>   the stable RPC conventions.
+
 ## Message event
 
 <!-- semconv event.rpc.message -->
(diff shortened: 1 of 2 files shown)


docs/rpc/rpc-spans.md
```diff
@@ -343,59 +340,10 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
-<!-- END AUTOGENERATED TEXT -->
-<!-- endsemconv -->
-
 ### Distinction from HTTP spans
 
 HTTP calls can generally be represented using just [HTTP spans](/docs/http/http-spans.md).
 If they address a particular remote service and method known to the caller, i.e., when it is a remote procedure call transported over HTTP, the `rpc.*` attributes might be added additionally on that span, or in a separate RPC span that is a parent of the transporting HTTP call.
 Note that *method* in this context is about the called remote procedure and *not* the HTTP verb (GET, POST, etc.).
 
-## Semantic conventions for specific RPC technologies
```
lmolkova, line 393: this is covered in RPC readme 

https://github.com/open-telemetry/semantic-conventions/blob/cefd73a824531597a1ebd77e09c5eebe21665241/docs/rpc/README.md?plain=1#L42-L46

docs/rpc/rpc-events.md
```diff
@@ -0,0 +1,50 @@
...
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `rpc.message`.
+
+Describes a message sent or received within the context of an RPC call.
+
+In the lifetime of an RPC stream, an event for each message sent/received on client and server spans SHOULD be created. In case of unary calls only one sent and one received message will be recorded for both client and server spans.
```
lmolkova:
```suggestion
In the lifetime of an RPC stream, an event for each message sent/received on client and server spans SHOULD be created. In case of unary calls message events SHOULD NOT be recorded.
```

docs/rpc/rpc-events.md
```diff
@@ -0,0 +1,50 @@
+<!--- Hugo front matter used to generate the website version of this page:
+linkTitle: Events
+--->
+
+# Semantic conventions for RPC events
+
+**Status**: [Development][DocumentStatus]
+
+This document defines events applicable in the context of remote procedure calls.
```
lmolkova, line 9: TODO: add version warning blurb here

docs/rpc/rpc-events.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Message event
```
lmolkova, line 36: TODO: 

do we want  to recommend (experimental) instrumentations to keep emitting those as span events or switch to log events? they currently report them as span events, but if they switch to log-based events, they'd need to break people again at some point.

I'd rather recommend instrumentations to keep emitting these as span events if they already do (when experimental features are enabled).

trask: In HTTP and Database migration docs, we documented updates to things that still weren't stable yet. In Java instrumentation, we took these (not yet stable) changes in at the same time as the stable changes.

I think it's ok to make a different decision with respect to span events v events though given that's part of a separate migration effort.

lmolkova: Let's consider
- deprecating span events
- reaching out to otel-maintainers to see if there is a good reason to keep them / existing usage
- alternatively we can keep existing span-events version around (in development)

lmolkova: Created https://github.com/open-telemetry/semantic-conventions/issues/3266 to get feedback from the community. Let's get this PR in and follow up with deprecation.

Issue #2967, opened 2025-10-23
Consider sensitivity of `error|exception.message` and span status description


lmolkova:
See https://github.com/open-telemetry/opentelemetry-specification/issues/3496 for the context. 

TL;DR: exception messages may contain emails, user info, passwords / secrets and other sensitive information.

Our current (in development) [error guidance](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md) recommends recording exception / error messages on spans status description or `exception | error.message` attribute on other signals.

What can we do:

- Option 1: status quo. 
   - Keep recording them by default
   - It's user responsibility to sanitize them all on spans and events
   - We'll document that all of them (span status description, `exception.message` and `error.message`) may contain sensitive information 
- Option 2: don't populate error | exception messages by default unless non-sensitivity is somehow guaranteed. Allow users to opt-in.
   -  Note: global opt-in is probably not helpful. The exception | error messages are very useful for observability and it's rather a bug and an edge case when they contain sensitive details.  
- Option 3: limit where sensitive info appears to `error.message` attribute
   - Document it could have sensitive details
   - Don't recommend to populate span status description in semconv ever, eventually deprecate `exception.message` 
   - Users would need to manually (via custom processor) sanitize known occurrences of problematic error messages using Span/Log processor
- ?

lmolkova:
I think it's the question of picking the least bad solution. 

My preference is on Option 3:
- error | exception message is super-useful for observability. OTel already recommends recording it in `exception.message` (stable)
- it's a bug when error message contains sensitive info
- Having 3 places to put error message info is not great, we can try converging it to a single place with #2296 and https://github.com/open-telemetry/opentelemetry-specification/pull/4333. Span-event deprecation gives us opportunity to put `exception.message` to rest. This would somewhat simplify sanitization and would have positive effect on correlation things across signals

lmolkova:
RPC SIG discussion:
- exception.message is too useful to drop it by default
- we should document that it might contain sensitive (PII / secrets) data if someone adds that info by accident
- it could be sanitized in post-processing pipeline

chrikrah:
The `exception.message` half of this shipped in #3310, but the span status description half is still open, and `docs/general/recording-errors.md` now contradicts itself:

1. Line 51 says to set the span status description when the additional information "is not expected to contain sensitive details".
2. Line 57 says that when the operation fails with an exception, the description SHOULD be set to the exception message.
[...]

PR #3228, opened 2025-12-19
Refine recording errors documentation to include logs and avoid span events

pellared:
Towards OTEP: [Span Event API deprecation plan](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4430-span-event-api-deprecation-plan.md)

Towards https://github.com/open-telemetry/opentelemetry-specification/issues/4429

Towards https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7254#issuecomment-2975444593
Towards https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7470

Supersedes https://github.com/open-telemetry/semantic-conventions/pull/2296

"Prototype" (do this across all instrumentation libraries in Go): https://github.com/open-telemetry/opentelemetry-go-contrib/issues/7254#issuecomment-3153698098

Note that this is needed toward stabilization of [`otelhttp`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp) which is the OTel Go instrumentation library for HTTP server and client as we want to have a clear and stable way on how to record errors.

## Changes

- Adds comprehensive guidance on recording errors via logs (event records) instead of span events
- Clarifies the distinction between "error" and "failed operation" concepts
- Deprecates the recommendation to use `Span.RecordException` for error recording
- Establishes consistency requirements for `error.type` across all signals (spans, metrics, logs)

## Other

I was also considering deprecating [Semantic conventions for exceptions on spans](https://github.com/open-telemetry/semantic-conventions/blob/cefd73a824531597a1ebd77e09c5eebe21665241/docs/exceptions/exceptions-spans.md), but decided to hold off and better scope it to a separate PR.

An next steps (probably after stabilizing parts of this document) we could deprecate:
- section: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception
- document: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/exceptions.md

Changing "error" into "exception" terminology (which better aligns with the specification e.g. `Span.RecordException` and current semantic conventions) can be done in a separate PR

I was thinking about adding `error.stacktrace` per [2025-12-09 Logs SIG meeting](https://docs.google.com/document/d/1BKjQWP32FXL9g1cGbyj7DMXV1Uq_RL8_78rWaMBhN0A/edit?tab=t.0#heading=h.m4iywp2d6zok), but I felt that it would be better to scope it to a separate PR as this is already big enough. But probably we would prefer to use `exception.*` attributes anyway.

### DIFF
--- a/docs/general/recording-errors.md
+++ b/docs/general/recording-errors.md
@@ -64,9 +64,6 @@ Semantic conventions for operations usually define an operation duration histogr
 metric. This metric SHOULD include the [`error.type`][ErrorType] attribute.
 This enables users to derive throughput and error rates.
 
-Errors that were retried or handled (allowing an operation to complete gracefully)
-SHOULD NOT be recorded on metrics that describe this operation.
-
 Operations that complete successfully SHOULD NOT include the `error.type` attribute,
 allowing users to filter out errors.
 


docs/general/recording-errors.md
```diff
@@ -33,38 +33,49 @@ An operation SHOULD be considered as failed if any of the following is true:
...
+An operation SHOULD be considered as failed when it ends with an error.
+
+Errors that were retried or handled (allowing an operation to complete gracefully)
+SHOULD NOT be recorded on spans or metrics that describe this operation.
+
+## Recording errors
+
+Instrumentation SHOULD ensure that the same [`error.type`][ErrorType] attribute
+values are applied consistently across all signals. This means that all signals
+produce the same `error.type` value for a given error.
```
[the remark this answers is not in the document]
pellared: https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/b3522359f569c53cc0e116183b26f9ddd6dd7274

docs/general/recording-errors.md
```diff
@@ -76,50 +81,40 @@ messaging operation may involve sending multiple messages) and includes `error.t
...
-    span.setStatus(StatusCode.ERROR, e.getMessage());
-    throw e;
-  }
-}
-```
+## Recording errors on logs
+
+When recording an error using logs:
+
+- MUST set [`EventName`][EventName] with a value that would normally be
```
cijothomas: The exist SpanEvent guideline uses just "exception" as SpanEvent's name.
https://github.com/open-telemetry/opentelemetry-specification/blob/v1.52.0/specification/trace/exceptions.md

Any value in continuing that approach? For example, if a user currently reports exception as SpanEvents, but uses a processor to convert all SpanEvents to LogRecords, continuing with same event-name is backward compatible.

Not a common scenario I guess. I think error.type makes a great EventName. Just want to see if there are any other considerations.

docs/general/recording-errors.md
```diff
@@ -76,50 +81,40 @@ messaging operation may involve sending multiple messages) and includes `error.t
...
+
+When an error occurs outside the context of any span
+and it causes an operation to fail,
+the instrumentation SHOULD record it as an event record.
+In such scenario, [`SeverityNumber`][SeverityNumber] MUST be greater than
+or equal to 17 (ERROR).
+
+When an error occurs inside the context of a span
+and it causes an operation to fail,
+the instrumentation SHOULD NOT additionally record it as an event record.
```
cijothomas: "event record"  - is an unclear term.

docs/general/recording-errors.md
```diff
@@ -76,50 +81,40 @@ messaging operation may involve sending multiple messages) and includes `error.t
...
+
+> [!NOTE]
+>
+> Applications that also want error event records corresponding to spans
```
cijothomas: this note is not very clear.. consider rephrasing..
Did you mean to recommend users to use a SpanProcessor to convert SpanEvents to LogRecords?

docs/general/recording-errors.md
```diff
@@ -33,38 +33,43 @@ An operation SHOULD be considered as failed if any of the following is true:
...
+  definition, for example, an exception message.
 
-  When the operation fails with an exception, the span status description SHOULD be set to
-  the exception message.
+Note that [Span Status Code][SpanStatus] MUST be left unset if the instrumented
+operation has ended without any errors.
 
-Refer to the [recording exceptions](#recording-exceptions) on capturing exception
-details.
+It is NOT RECOMMENDED to record the error via a span event,
```
cijothomas, line 58: should we do some pros vs cons of doing so?
One easy thing is: if we report exception as SpanEvent, it naturally inherits Span's sampling decision. If we report as Event/Log, then extra effort is required to inherit sampling decision.

docs/general/recording-errors.md
pellared: Should I link to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#log-and-event-record-definition? Any other proposal?

docs/general/recording-errors.md
pellared: > Did you mean to recommend users to use a SpanProcessor to convert SpanEvents to LogRecords?

This is a separate thing also worth considering as a seperate PR towards [Span Event API deprecation plan](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4430-span-event-api-deprecation-plan.md).

Here it is about that error spans can be also emitted as event records if the users desires to have them. From https://docs.google.com/document/d/1BKjQWP32FXL9g1cGbyj7DMXV1Uq_RL8_78rWaMBhN0A/edit?tab=t.0#heading=h.m4iywp2d6zok: 

> record terminating errors on spans, some people still want their errors in logs, there could be a processor to record them as logs

How about https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/b8ca8281cc6fc2805fa037869a30736b69fffcd5?

I welcome other rephrasing proposals.

docs/general/recording-errors.md
pellared: For head sampling we could use a log processor or via [`LoggerConfig.trace_based`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#loggerconfig) to filter out logs that are inside a span that should not be sampled. Not sure if we should care about tail sampling as I think it would be probably done via a Collector (?).

docs/general/recording-errors.md
pellared: Not all languages have exceptions and we want to avoid from using this term. It also does not really provides any valuable information.

At some point of time we were considering setting `error.type` attribute and not setting any `EventName`. However, we feel that the error type is identifying and we want to promote using event records over log records.

docs/general/recording-errors.md
cijothomas: yes.. "event record" and "span event" might be confused.. Linking to EventRecord spec would avoid any confusion.

docs/general/recording-errors.md
cijothomas: makes sense!

docs/general/recording-errors.md
pellared: https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/00d876922003b6c9e9404659c22fa84e452d3735

docs/general/recording-errors.md
```diff
@@ -33,38 +33,43 @@ An operation SHOULD be considered as failed if any of the following is true:
...
 > application is simply checking whether the resource exists.
 >
-> Instrumentations that have additional context about a specific request MAY use
-> this context to set the span status more precisely.
+> Instrumentations that have additional context about a specific request SHOULD
+> use this context to classify whether the status code is an error.
 
-Errors that were retried or handled (allowing an operation to complete gracefully) SHOULD NOT
-be recorded on spans or metrics that describe this operation.
+## What constitutes a failed operation
```
lmolkova: nit: does it need a separate section?

docs/general/recording-errors.md
```diff
@@ -33,38 +33,43 @@ An operation SHOULD be considered as failed if any of the following is true:
...
+## Recording errors
 
-- SHOULD set the span status code to `Error`
-- SHOULD set the [`error.type`](/docs/registry/attributes/error.md#error-type) attribute
-- SHOULD set the span status description when it has additional information
-  about the error which is not expected to contain sensitive details and aligns
-  with [Span Status Description][SpanStatus] definition.
+Instrumentation SHOULD ensure that, for a given error, the same value is
+used as the [`error.type`][ErrorType] attribute on spans and metrics, and as
+[`EventName`][EventName] on logs.
```
lmolkova: I don't think we can or should require error.type to become event name.

E.g. `socket.connection_failed` event could have `error.type` attribute with different values like  `reset_by_peer` or `timeout`. 

So I'd rather ask `error.type` to be the same  on all signals.

docs/general/recording-errors.md
```diff
@@ -76,50 +81,41 @@ messaging operation may involve sending multiple messages) and includes `error.t
...
-  }
-}
-```
+## Recording errors on logs
+
+When recording an error using logs ([event records][EventRecord]):
+
+- MUST set [`EventName`][EventName] with a value that would normally be
+  used for an [`error.type`][ErrorType] attribute.
+- SHOULD set [`error.message`][ErrorMessage] attribute to add additional
```
lmolkova, line 90: this is not blocking (does not change status quo), but  FYI exception message turns out to be potentially sensitive https://github.com/open-telemetry/semantic-conventions/issues/2967

docs/general/recording-errors.md
```diff
@@ -76,50 +81,41 @@ messaging operation may involve sending multiple messages) and includes `error.t
...
+
+When an error occurs outside the context of any span
+and it causes an operation to fail,
+the instrumentation SHOULD record it as an event record.
+In such scenario, [`SeverityNumber`][SeverityNumber] MUST be greater than
+or equal to 17 (ERROR).
+
+When an error occurs inside the context of a span
+and it causes an operation to fail,
+the instrumentation SHOULD NOT additionally record it as an event record.
```
lmolkova: I don't understand this distinction between inside/outside of span context. Why logs instrumentation should care about spans at all?

docs/general/recording-errors.md
```diff
@@ -76,50 +81,41 @@ messaging operation may involve sending multiple messages) and includes `error.t
...
-    span.setStatus(StatusCode.ERROR, e.getMessage());
-    throw e;
-  }
-}
-```
+## Recording errors on logs
+
+When recording an error using logs ([event records][EventRecord]):
+
+- MUST set [`EventName`][EventName] with a value that would normally be
```
lmolkova: see my other comment on this

docs/general/recording-errors.md
```diff
@@ -76,50 +81,41 @@ messaging operation may involve sending multiple messages) and includes `error.t
...
+  used for an [`error.type`][ErrorType] attribute.
+- SHOULD set [`error.message`][ErrorMessage] attribute to add additional
+  information about the error, for example, an exception message.
+
+When an error is retried or handled, even when the overall operation completes successfully,
+the instrumentation SHOULD record it as an event record for diagnostic purposes.
```
lmolkova: Is there a generic OTel  instrumentation where this guidance would be useful? do we need to be that specific?

docs/general/recording-errors.md
pellared: The more I think about it the more I like your proposal 

Similar thread https://github.com/open-telemetry/semantic-conventions/pull/3228#discussion_r2635832271 for transparency.

docs/general/recording-errors.md
pellared: This is one of the reasons I decided to make it as SHOULD and I think it follows option 3 from the referenced issue. Resolving.

docs/general/recording-errors.md
pellared: There is no need to additionaly record an error using logs/events when it can be already captured via a span. Also see https://github.com/open-telemetry/semantic-conventions/pull/3228#discussion_r2635907016

E.g. when doing our HTTP instrumentation library in Go we can only record the errors via span attributes and status. We do not need to additionally add logs.

Is there anything I should clarify/rephrase?

docs/general/recording-errors.md
pellared: We have a similar section for defining "error". I think that the "failed operations" is a very important term used in this document that deserves also being defined. I can refactor it to a single section like "Definitions used in this document". However, I would prefer doing such structural changes in a separate PR.

docs/general/recording-errors.md
pellared: Addressed: 
- https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/02744ca955b82f3a819997c4ccf7b6fb5d511152
- https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/aa386707a313e7d61c42bab0787ebd9ff96d786f

docs/general/recording-errors.md
pellared: This would be useful for people asking how to observe "retries". I think it is better to call it out rather than leaving it unspecified. 

Changing to "MAY" and moving to the bottom (as this is not that important).

pellared: Addressed https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/1be560d24343902450ec830dab9620fc556ea921

docs/general/recording-errors.md
```diff
@@ -76,50 +80,43 @@ messaging operation may involve sending multiple messages) and includes `error.t
...
+- SHOULD set [`EventName`][EventName] with a value that help indicating
+  what operation failed, e.g. `socket.connection_failed`,
+- SHOULD set the [`error.type`][ErrorType] attribute,
+- SHOULD set [`error.message`][ErrorMessage] attribute to add additional
+  information about the error, for example, an exception message.
+
+When an error occurs outside the context of any span
+and it causes an operation to fail,
+the instrumentation SHOULD record it as an event record.
+In such scenario, [`SeverityNumber`][SeverityNumber] MUST be greater than
```
dashpole: I'm not convinced every "operation that fails" needs to be an error-level log. Even if otelhttp did not have trace instrumentation, it would not be helpful to log all failed requests at error level.

docs/general/recording-errors.md
dashpole: What if the user is not using tracing? Should it then emit an error log in that case? Instrumentation should be able to check if the tracer is enabled.

What if a library adds log instrumentation before trace instrumentation? Then, presumably, they would want to remove their logs to have a correct experience?

I don't like trace and log instrumentation guidance being linked in this way. It seems like it just complicates both the instrumentation experience and the consumption experience.

pellared: @dashpole, do you have any proposals? Stick to one (e.g. trace instrumentation when operation fails) or allow both? Should we specify some preference?

dashpole: I would honestly prefer both. But I think our guidance should be that errors SHOULD be logged at debug level unless we know it should definitely be error-level. See my other comment here: https://github.com/open-telemetry/semantic-conventions/pull/3228#discussion_r2662119559.

Unrelated, but I would love to see a logger "mode" that collects more detailed logs during sampled spans to recover some of the "span event" experience we are losing here.  Maybe a "sampled span verbosity" level or something similar.

pellared: PTAL https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/36405a0d4112682d80f35785cb7a90f4cb536754

It is the most simple, yet flexible, solution that I was able to come up which is (hopefully) in the spirit of your suggestion.

docs/general/recording-errors.md
pellared: _Resolving as duplicate of the other thread https://github.com/open-telemetry/semantic-conventions/pull/3228#discussion_r2662256992_

docs/general/recording-errors.md
```diff
@@ -33,38 +33,42 @@ An operation SHOULD be considered as failed if any of the following is true:
...
+  definition, for example, an exception message.
 
-  When the operation fails with an exception, the span status description SHOULD be set to
-  the exception message.
+Note that [Span Status Code][SpanStatus] MUST be left unset if the instrumented
+operation has ended without any errors.
 
-Refer to the [recording exceptions](#recording-exceptions) on capturing exception
-details.
+It is NOT RECOMMENDED to record the error via a span event,
```
alexmojaki, line 58: Without this, I don't see anywhere that a stacktrace gets recorded.

alexmojaki: I see at the end of the PR description you wrote:

> I was thinking about adding `error.stacktrace` per [2025-12-09 Logs SIG meeting](https://docs.google.com/document/d/1BKjQWP32FXL9g1cGbyj7DMXV1Uq_RL8_78rWaMBhN0A/edit?tab=t.0#heading=h.m4iywp2d6zok), but I felt that it would be better to scope it to a separate PR as this is already big enough.

Currently, this PR is not that big (+83, -73). Was it bigger before?

> But probably we would prefer to use `exception.*` attributes anyway.

This is a big thing to leave hanging. It's also the essence of https://github.com/open-telemetry/opentelemetry-specification/issues/4429. Without that, I don't see how this PR can be marked as closing that issue.

docs/general/recording-errors.md
```diff
@@ -33,38 +33,42 @@ An operation SHOULD be considered as failed if any of the following is true:
...
-  with [Span Status Description][SpanStatus] definition.
+Instrumentation SHOULD ensure that, for a given error,
+the same [`error.type`][ErrorType] attribute value is used across all signals.
 
-  It's NOT RECOMMENDED to duplicate status code or `error.type` in span status description.
+## Recording errors on spans
+
+When the instrumented operation failed, the instrumentation:
+
+- SHOULD set the span status code to `Error`,
```
alexmojaki: By the definitions above, this would happen any time the span ends with an exception thrown by the instrumented operation. But the instrumentation may know that the exception doesn't actually represent an error, e.g. that the exception is expected to be handled outside the span (see https://github.com/pydantic/logfire/issues/1361 for example).

We still want to be able to record info about these kinds of exceptions, including a traceback. But they shouldn't be marked as errors. That means that there's also no place here to store the exception message, since the span status description can't be set.

docs/general/recording-errors.md
```diff
@@ -76,50 +80,34 @@ messaging operation may involve sending multiple messages) and includes `error.t
...
+
+- SHOULD set [`EventName`][EventName] with a value that help indicating
+  what operation failed, e.g. `socket.connection_failed`,
+- SHOULD set the [`error.type`][ErrorType] attribute,
+- SHOULD set [`error.message`][ErrorMessage] attribute to add additional
+  information about the error, for example, an exception message,
+- SHOULD set [`SeverityNumber`][SeverityNumber].
+
+When the error happens during an operation,
+it is RECOMMENDED to set [`SeverityNumber`][SeverityNumber] below 9 (INFO).
```
alexmojaki: I think what you're saying here is "when the error happens before the operation has ended and is handled, so the operation is still expected to complete successfully". That should be clarified.

In that situation, I find it surprising to consider this as severity below INFO. I would usually log this as a warning, it's concerning enough to warrant some attention. A user is likely to configure logging to filter out everything below INFO, meaning they wouldn't see any sign of transient errors at all. If there was internal retrying to overcome the error, you'd likely see a span taking mysteriously long and think there was a performance problem.

docs/general/recording-errors.md
pellared: > This is a big thing to leave hanging. It's also the essence of https://github.com/open-telemetry/opentelemetry-specification/issues/4429. Without that, I don't see how this PR can be marked as closing that issue.

I agree. Changing to "towards" from "fixes".


Just more context: There is no `error.stacktrace` attribute, yet and I do not want to a new attribute in scope of this PR. Especially that it is likely that we will decide to use `exception.stacktrace`.

docs/general/recording-errors.md
pellared: > I think what you're saying here is "when the error happens before the operation has ended and is handled, so the operation is still expected to complete successfully". That should be clarified.

It is also when it has ended and when it is not handled (operation fails).
This tries to address https://github.com/open-telemetry/semantic-conventions/pull/3228#discussion_r2662124477 

>  I would usually log this as a warning, it's concerning enough to warrant some attention. 

I disagree. Transient errors are happening all the time. I would consider it as logs overuse. 

> A user is likely to configure logging to filter out everything below INFO, meaning they wouldn't see any sign of transient errors at all. If there was internal retrying to overcome the error, you'd likely see a span taking mysteriously long and think there was a performance problem.

This is not true. Note that the user would get the information about the transient errors via metrics. Then if the users want diagnose it further they can always set even "TRACE" severity for given `error.type`.

docs/general/recording-errors.md
pellared: Decided to address it via https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/6e232d4c6297ff5a971cb39c7ee12a412ca73a59

docs/general/recording-errors.md
pellared: PTAL https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/6e232d4c6297ff5a971cb39c7ee12a412ca73a59

docs/general/recording-errors.md
```diff
@@ -33,38 +33,42 @@ An operation SHOULD be considered as failed if any of the following is true:
...
-- SHOULD set the [`error.type`](/docs/registry/attributes/error.md#error-type) attribute
-- SHOULD set the span status description when it has additional information
-  about the error which is not expected to contain sensitive details and aligns
-  with [Span Status Description][SpanStatus] definition.
+## Recording errors on spans
+
+When the instrumented operation failed, the instrumentation:
 
-  It's NOT RECOMMENDED to duplicate status code or `error.type` in span status description.
+- SHOULD set the span status code to `Error` if this is a semantical error,
```
alexmojaki:
```suggestion
- SHOULD set the span status code to `Error` if this is a semantic error,
```

i had to actually look up 'semantical', is it more common in some contexts, or does it mean something subtly different from 'semantic' here?

docs/general/recording-errors.md
```diff
@@ -33,38 +33,42 @@ An operation SHOULD be considered as failed if any of the following is true:
...
+- SHOULD set the [`error.type`][ErrorType] attribute,
+- SHOULD set [`error.message`][ErrorMessage] attribute to add additional
+  information about the error, for example, an exception message,
+- SHOULD set the `error.stacktrace` attribute.
```
alexmojaki, line 53: i thought you didn't want to introduce new attributes. why not `exception.stacktrace`?

docs/general/recording-errors.md
alexmojaki: i think the condition needs to be elaborated

docs/general/recording-errors.md
pellared: > Changing "error" into "exception" terminology (which better aligns with the specification e.g. Span.RecordException and current semantic conventions) can be done in a separate PR.

There are too many open discussions around this and it is not clear yet which way we want to go.

docs/general/recording-errors.md
pellared: Any proposal how? Here is an example what constitutes a semantic error for HTTP spans: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#status. Maybe a hyperlink as this is a good example?

docs/general/recording-errors.md
alexmojaki: You'd only set the stacktrace in the case of an exception, right?

docs/general/recording-errors.md
alexmojaki: SHOULD set the span status code to `Error`, unless this is an exception that doesn't actually represent an error, e.g. it is being used for control flow, or is expected to be handled gracefully outside the span. For example, a 400 HTTP status code is not an error on an HTTP server span because it indicates a problem with the user's request, not the application. But a 500 HTTP status code is an error.

docs/general/recording-errors.md
alexmojaki: > I disagree. Transient errors are happening all the time. I would consider it as logs overuse.

Based on what? We record transient errors in our application as warnings and I'm glad we do.

> Note that the user would get the information about the transient errors via metrics.

Unless you're lucky, there would be no way to determine from metrics what happened for a specific given span. And the user might not even think to look at metrics at all, or know how.

> Then if the users want diagnose it further

But they can't diagnose past events.

> they can always set even "TRACE" severity for given error.type.

Is this particularly easy to do? In any case I think it would be better if the user had the option to set the severity of the log to something higher.

docs/general/recording-errors.md
```diff
@@ -33,38 +33,42 @@ An operation SHOULD be considered as failed if any of the following is true:
...
-  with [Span Status Description][SpanStatus] definition.
+## Recording errors on spans
+
+When the instrumented operation failed, the instrumentation:
 
-  It's NOT RECOMMENDED to duplicate status code or `error.type` in span status description.
+- SHOULD set the span status code to `Error` if this is a semantic error,
+- SHOULD set the [`error.type`][ErrorType] attribute,
+- SHOULD set [`error.message`][ErrorMessage] attribute to add additional
+  information about the error, for example, an exception message,
```
alexmojaki, line 52: This is incompatible with:

https://github.com/open-telemetry/semantic-conventions/blob/502202ec77ff9f0aed48e4a107d5b0b1c9fe5af6/docs/registry/attributes/error.md#L19-L21

one of them needs to change

docs/general/recording-errors.md
dashpole: Reading it back, I think we need to either be more nuanced, or specify less. Saying "above info" or "below info" without enough context will seem confusing. I also worry that "error" is overloaded and confusing right now (e.g. given other discussions on the PR).  From https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-severitynumber:

>  ERROR | An error event. Something went wrong.

I would suggest something like:

Errors SHOULD be logged, but not necessarily at the Error log level. For example, an error on an operation may be a Debug event if the operation will is retried, and if the failure is not generally relevant to the application owner.

Do you think we need additional clarification that this "operation error" is not the same as a real (logging) "error"?

This does have the potential downside of duplicated telemetry.  But for real "error" events, which should ideally be rare, this seems acceptable to me. But it seems important for log or trace telemetry to be useable independently, and for instrumentation guidance for one signal to not depend on whether a different signal is used.

docs/general/recording-errors.md
pellared: Side note. Problably when we change exception to error terminology it would be more clear.

docs/general/recording-errors.md
pellared: Going to address it together with https://github.com/open-telemetry/semantic-conventions/pull/3228#discussion_r2668761120

> Saying "above info" or "below info" without enough context will seem confusing.

docs/general/recording-errors.md
pellared: This is implementation/language specific. E.g. usually in Go errors do not have stacktraces. However, they may have it  e.g. when using [`github.com/pkg/errros`](https://pkg.go.dev/github.com/pkg/errors#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper).

docs/general/recording-errors.md
pellared: Similar feedback in https://github.com/open-telemetry/semantic-conventions/pull/3228#discussion_r2668761120

docs/general/recording-errors.md
alexmojaki: so will you add a new attribute to the registry?

docs/general/recording-errors.md
pellared: PTAL https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/14f21c66274281c60f85bdca23c07566f9c41dde

docs/general/recording-errors.md
pellared: PTAL https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/14f21c66274281c60f85bdca23c07566f9c41dde

docs/general/recording-errors.md
pellared: PTAL https://github.com/open-telemetry/semantic-conventions/pull/3228/commits/14f21c66274281c60f85bdca23c07566f9c41dde

docs/general/recording-errors.md
```diff
@@ -33,38 +33,44 @@ An operation SHOULD be considered as failed if any of the following is true:
...
-be recorded on spans or metrics that describe this operation.
+## What constitutes a failed operation
 
-## Recording errors on spans
+An operation SHOULD be considered as failed when it ends with an error.
 
-[Span Status Code][SpanStatus] MUST be left unset if the instrumented operation has
-ended without any errors.
+Errors that were retried or handled (allowing an operation to complete gracefully)
+SHOULD NOT be recorded on spans or metrics that describe this operation.
```
dashpole: Optional, but might be worth repeating here that in this case logs SHOULD be logged at a low severity.  Or you could say "SHOULD NOT be logged at Error level" if you want.

docs/general/recording-errors.md
pellared: Not in the scope of this PR or at least not until there is a clear agreement.

> There are too many open discussions around this and it is not clear yet which way we want to go.

Note that this document is in Development status.

docs/general/recording-errors.md
pellared: Removing the section after changes.

I removed the whole section so that we can stabilize each section independently. Also as noted https://github.com/open-telemetry/semantic-conventions/pull/3228#discussion_r2668294478 there are cases when we want to add attribute to spans even if the operation has not (semantically) failed. The metrics already has 

> Operations that complete successfully SHOULD NOT include the `error.type` attribute,
allowing users to filter out errors.

docs/general/recording-errors.md
pellared: PTAL

I removed the whole section so that we can stabilize each section independently. Also as noted https://github.com/open-telemetry/semantic-conventions/pull/3228#discussion_r2668294478 there are cases when we want to add attribute to spans even if the operation has not (semantically) failed. The metrics already has 

> Operations that complete successfully SHOULD NOT include the `error.type` attribute,
allowing users to filter out errors.

docs/general/recording-errors.md
dashpole: looks great

docs/general/recording-errors.md
dashpole: i'm happy with that

pellared:
SIG meeting notes:
Some changes done here are seen as too big and too controversial.
Going to make smaller steps on things that we agreed on.


Personal opinion: I think it was a good as this PR initiated some important discussions and helps us planning next steps.

pellared:
SIG meeting notes:
We also agreed that in [otelhttp](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp) we should not use `Record Exception` API nor "recording exception" guidelines. We should just follow "recording errors on spans" and "recording errors on metrics" guidelines. 
CC @open-telemetry/go-maintainers

Issue #2912, opened 2025-10-14
Add Workflows/Agents/Tasks to gen_ai


keith-decker:
### Area(s)

area:gen-ai

## Summary

This proposal extends the existing [GenAI Agent semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-agent-spans.md) with **Workflow** and **Task** span types to support complex multi-agent systems. It also adds agent context to existing LLM and embedding metrics for better attribution.

### Current State
The existing GenAI semantic conventions (as of v1.28.0) include:
- ✅ **Agent spans**: `create_agent` and `invoke_agent` operations ([docs](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-agent-spans.md))
- ✅ **LLM spans**: `gen_ai.client.chat` and `gen_ai.client.completion`
- ✅ **Embedding spans**: `gen_ai.client.embeddings`
- ✅ **Metrics**: Token usage, duration, etc.

### Gap
Modern agentic AI systems involve **orchestration** that isn't captured:
- **Workflows**: Top-level orchestration coordinating multiple agents (e.g., LangGraph graphs, CrewAI crews)
- **Tasks**: Discrete units of work assigned to or decomposed by agents
- **Hierarchical relationships**: Workflow → Agents → Tasks → LLM calls
- **Attribution**: Which agent/task generated which LLM calls and costs

Without workflow and task conventions, we can observe individual agents but not their orchestration or task decomposition.

Please see [complete proposal](https://github.com/keith-decker/semantic-conventions/blob/55be7efae3251ff4b9746f4a91fd1c2e6aeb98a4/docs/gen-ai/gen-ai-agent-proposal.md)

#### `gen_ai.workflow` Span
Represents a workflow orchestrating multiple agents and tasks.

**Span Name Format:**
- `gen_ai.workflow {workflow_name}`
- Examples: `gen_ai.workflow multi_agent_rag`, `gen_ai.workflow customer_support_pipeline`

**Required Attributes:**
- `gen_ai.workflow.name` (string): Name/identifier of the workflow
- `gen_ai.operation.name` (string): `"workflow"`

**Optional Attributes:**
- `gen_ai.workflow.type` (string): Orchestration type (e.g., "sequential", "parallel", "graph", "dynamic")
- `gen_ai.workflow.description` (string): Human-readable description of workflow's purpose
[...]

PR #3233, opened 2025-12-22
gen-ai: add security guardian (apply_guardrail) span + finding event

nagkumar91:
## Summary

Add semantic conventions for GenAI security guardrail operations — spans, attributes, events, and metrics for security evaluation of GenAI inputs/outputs.

### New Span

| Span (`gen_ai.operation.name`) | Kind | Description |
|---|---|---|
| `apply_guardrail` | internal | Security guardrail evaluation of content or actions |

Span name: `apply_guardrail {gen_ai.guardian.name} {gen_ai.security.target.type}` (e.g., `apply_guardrail "PII Filter" llm_input`). When `gen_ai.guardian.name` is unavailable, the fallback is `apply_guardrail {gen_ai.security.target.type}`.

Guardrail spans are children of the operation they protect (e.g., `chat`, `execute_tool`). Multiple guardrail spans MAY exist under a single parent if guardrails are chained.

### New Attributes

**Guardian (evaluating service):**

| Attribute | Type | Req Level | Description |
|---|---|---|---|
| `gen_ai.guardian.id` | string | conditionally_required | Machine-readable identifier of the guardrail service |
| `gen_ai.guardian.name` | string | recommended | Human-readable name of the guardrail service |
| `gen_ai.guardian.version` | string | recommended | Version of the guardrail service |
| `gen_ai.guardian.provider.name` | enum | recommended | Provider/vendor of the guardrail service. Open enum with well-known values: `azure.ai.content_safety`, `aws.bedrock`, `gcp.model_armor` |

**Security decision:**

| Attribute | Type | Req Level | Description |
|---|---|---|---|
| `gen_ai.security.decision.type` | enum | required | Decision made: `allow`, `deny`, `modify`, `warn`, `audit` |
| `gen_ai.security.decision.reason` | string | conditionally_required | Human-readable reason (when not `allow`) |
| `gen_ai.security.decision.code` | int | recommended | Numeric code for the security decision |

**Security target:**

| Attribute | Type | Req Level | Description |
|---|---|---|---|
| `gen_ai.security.target.type` | string | required | Free-form string identifying what was evaluated. Suggested values: `llm_input`, `llm_output`, `tool_call`, `tool_definition`, `memory_store`, `memory_retrieve`, `knowledge_query`, `knowledge_result`, `message` |
| `gen_ai.security.target.id` | string | conditionally_required | Identifier of the specific target |

**Policy:**

| Attribute | Type | Req Level | Description |
|---|---|---|---|
| `gen_ai.security.policy.id` | string | conditionally_required | Policy that triggered the decision |
| `gen_ai.security.policy.name` | string | recommended | Human-readable policy name |
| `gen_ai.security.policy.version` | string | recommended | Policy version |

**Content inspection (opt-in, PII-sensitive):**

| Attribute | Type | Req Level | Description |
|---|---|---|---|
| `gen_ai.security.content.input.value` | any | opt_in | Input content before guardrail processing |
| `gen_ai.security.content.output.value` | any | opt_in | Output content after guardrail processing |
| `gen_ai.security.content.input.hash` | string | conditionally_required | Hash of input for forensic correlation |
| `gen_ai.security.content.modified` | boolean | conditionally_required | Whether content was modified by the guardrail |

**External correlation:**

| Attribute | Type | Req Level | Description |
|---|---|---|---|
| `gen_ai.security.external_event_id` | string | conditionally_required | External correlation identifier for security events |

### New Event

| Event | Description |
|---|---|
| `gen_ai.security.finding` | Emitted when a guardrail produces a finding (deny, modify, warn) |

**Event attributes (risk assessment):**

| Attribute | Type | Req Level | Description |
|---|---|---|---|
| `gen_ai.security.risk.category` | string | required | Category of security risk (e.g., `prompt_injection`, `pii`, `toxicity`) |
| `gen_ai.security.risk.severity` | enum | required | Severity: `none`, `low`, `medium`, `high`, `critical` |
| `gen_ai.security.risk.score` | double | recommended | Numeric risk/confidence score (0.0–1.0) |
| `gen_ai.security.risk.metadata` | string[] | recommended | Non-content metadata about the risk |
| `gen_ai.security.policy.id` | string | conditionally_required | Policy that triggered the finding |
| `gen_ai.security.policy.name` | string | recommended | Human-readable policy name |
| `gen_ai.security.policy.version` | string | recommended | Policy version |

### Supporting Materials

| Resource | Description |
|---|---|
| 📄 [Non-Normative Implementation Spec](https://gist.github.com/nagkumar91/95efa05449c72b4c95958190f26f13ba) | Detailed spec with framework mappings and trace examples |
| 🔧 [Prototype Framework Code](https://gist.github.com/nagkumar91/7328b4be7f80d01db087929382ab6cab) | OTel bootstrap, guardian utils, and LangChain adapter |
| 📖 [Story Scripts & Trace Coverage](https://gist.github.com/nagkumar91/2be5a18ce1fc221f917b8d8230376f03) | 5 end-to-end scenarios with trace coverage matrix |

---

## Terminology Note

The semantic conventions use **guardian** (`gen_ai.guardian.*`) to describe the evaluating service/component and **guardrail** for the operation name (`apply_guardrail`) and policy configuration. This distinction was deliberate:

- A single **guardian** (e.g., Azure Content Safety) can enforce multiple **policies/guardrails**
- The operation is called `apply_guardrail` to align with industry terminology (AWS Bedrock Guardrails, NVIDIA NeMo Guardrails)

---

### Single guardrail — input filtering

```
chat "gpt-4"                                                         (1200ms)
├── apply_guardrail "Azure Content Safety" llm_input                 (45ms)
│   ├── gen_ai.operation.name: "apply_guardrail"
│   ├── gen_ai.guardian.id: "content-filter-v2"
│   ├── gen_ai.guardian.name: "Azure Content Safety"
│   ├── gen_ai.guardian.provider.name: "azure.ai.content_safety"
│   ├── gen_ai.security.decision.type: "allow"
│   ├── gen_ai.security.target.type: "llm_input"
│   └── gen_ai.security.policy.id: "default-policy"
│
└── (LLM inference)                                                  (1100ms)
```

### Chained guardrails — input + output pipeline

```
chat "gpt-4"                                                         (1500ms)
│
├── apply_guardrail "PII Filter" llm_input                           (30ms)
│   ├── gen_ai.guardian.id: "pii-filter-v3"
│   ├── gen_ai.guardian.name: "Custom PII Filter"
│   ├── gen_ai.guardian.provider.name: "custom"
│   ├── gen_ai.security.decision.type: "modify"
│   ├── gen_ai.security.content.modified: true
│   └── gen_ai.security.target.type: "llm_input"
│
├── (LLM inference)                                                  (1100ms)
│
├── apply_guardrail "Toxicity Filter" llm_output                     (40ms)
│   ├── gen_ai.guardian.id: "toxicity-v2"
│   ├── gen_ai.guardian.name: "Azure Content Safety"
│   ├── gen_ai.guardian.provider.name: "azure.ai.content_safety"
│   ├── gen_ai.security.decision.type: "deny"
│   ├── gen_ai.security.decision.reason: "toxicity_detected"
│   └── gen_ai.security.target.type: "llm_output"
│
└── apply_guardrail "Prompt Shield" llm_input                        (25ms)
    ├── gen_ai.guardian.id: "prompt-shield-v1"
    ├── gen_ai.guardian.name: "Prompt Shield"
    ├── gen_ai.guardian.provider.name: "azure.ai.content_safety"
    ├── gen_ai.security.decision.type: "allow"
    └── gen_ai.security.target.type: "llm_input"
```

### Tool call guardrail

```
invoke_agent "ResearchBot"                                           (3000ms)
├── chat "gpt-4"                                                     (800ms)
│
├── apply_guardrail "Tool Policy" tool_call                          (15ms)
│   ├── gen_ai.guardian.id: "tool-policy-v1"
│   ├── gen_ai.security.decision.type: "deny"
│   ├── gen_ai.security.decision.reason: "unauthorized_tool"
│   ├── gen_ai.security.target.type: "tool_call"
│   └── gen_ai.security.target.id: "delete_database"
│
└── execute_tool "web_search"                                        (500ms)
```


adityamehra:
@nagkumar91 We have a similar use-case and when a security incident happens for a chat span we create a new span as a chat span and add an attributed called `gen_ai.security.event_id`. The value for this attribute can either be in the response body of the inspection call, which is separate from the actual LLM call, or can be in the response header when inspection happens along with the LLM call. Is it possible to add support for this attribute in here? Thanks!

Here's the sample of how we add it as of now - https://github.com/signalfx/splunk-otel-python-contrib/tree/main/instrumentation-genai/opentelemetry-instrumentation-aidefense#trace-integration

adityamehra:
Also, it will be great if we can have another entry in the `otel-genai-util` for this new span type like we have for chat span - https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py#L96. Or probably this new type can extend the `LLMInvocation` type

model/gen-ai/spans.yaml
```diff
@@ -337,6 +337,76 @@ groups:
...
+      `gen_ai.operation.name` SHOULD be `apply_guardrail`.
+
+      **Span name** SHOULD be `apply_guardrail {gen_ai.guardian.name}`.
+      When `gen_ai.guardian.name` is not available, it SHOULD be `apply_guardrail {gen_ai.security.target.type}`.
+
+      Guardian spans SHOULD be children of the operation span they are protecting (for example,
+      `span.gen_ai.inference.client` or `span.gen_ai.execute_tool.internal`).
+
+      Multiple guardian spans MAY exist under a single operation span if multiple guardians are chained.
+    attributes:
```
adityamehra, line 458: Can we add an attribute called `gen_ai.security.event_id`?

nagkumar91:
> @nagkumar91 We have a similar use-case and when a security incident happens for a chat span we create a new span as a chat span and add an attributed called `gen_ai.security.event_id`. The value for this attribute can either be in the response body of the inspection call, which is separate from the actual LLM call, or can be in the response header when inspection happens along with the LLM call. Is it possible to add support for this attribute in here? Thanks!
> 
> Here's the sample of how we add it as of now - https://github.com/signalfx/splunk-otel-python-contrib/tree/main/instrumentation-genai/opentelemetry-instrumentation-aidefense#trace-integration

Would this be better as an event (security finding event as proposed in this span)? Wondering why its a chat span?

model/gen-ai/spans.yaml
nagkumar91: The event being proposed would be a `gen_ai.security.finding`

Apply guardrail span will have these for IDs:
- gen_ai.guardian.id
- gen_ai.security.target.id
- gen_ai.security.policy.id

Would any of those fit your need?

adityamehra: > Would this be better as an event (security finding event as proposed in this span)? Wondering why its a chat span?

Currently, the domain team is using `event_id` and it's as per their requirement. Also, in our case the event is generated elsewhere and not instrumentation side. We used chat span to use existing types in the `genai-utils` for now and manage span life cycle using it.

docs/registry/attributes/gen-ai.md
```diff
@@ -25,14 +25,18 @@ This document defines the attributes used to describe telemetry in the context o
...
-| <a id="gen-ai-output-type" href="#gen-ai-output-type">`gen_ai.output.type`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | Represents the content type requested by the client. [6] | `text`; `json`; `image` |
+| <a id="gen-ai-guardian-id" href="#gen-ai-guardian-id">`gen_ai.guardian.id`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The unique identifier of the security guardian or guardrail. [3] | `guard_abc123`; `sgi5gkybzqak`; `content-filter-v2` |
```
habibam: One nit, the draft currently mixes guardian (evaluator/service) vs guardrail (policy/config) identity (e.g., mapping provider guardrail IDs into gen_ai.guardian.id). I suggest keeping gen_ai.guardian.* for the evaluating component, and mapping guardrail/config identifiers (like aws.bedrock.guardrail.id) to gen_ai.security.policy.id (and/or on gen_ai.security.finding when policy-triggered). Otherwise it becomes hard/impossible to differentiate a guardian from a guardrail in traces, and we end up with confusing cardinality/semantics. This keeps cross provider correlation clean.

prototype/frameworks/mcp/guardian_adapter.py
```diff
@@ -0,0 +1,473 @@
...
+    def agent_id(self) -> Optional[str]:
+        """Get agent ID from server context."""
+        return self.server_name
+
+
+# ============================================================================
+# MCP Guardian Adapter
+# ============================================================================
+
+class MCPGuardianAdapter(BaseGuardianAdapter[MCPContext]):
```
habibam: We should add in an example of generating a span from the response of elicitation

nagkumar91:
Thanks for the feedback @adityamehra! 

Regarding the `otel-genai-util` entry - this is tracked separately since it lives in [opentelemetry-python-contrib](https://github.com/open-telemetry/opentelemetry-python-contrib). Once this semconv PR is approved and merged, I'll open a follow-up PR to add the corresponding types for `apply_guardrail` spans to the util package.

In the meantime, I've addressed your other feedback:
- Added `gen_ai.security.external_event_id` attribute for correlation with external security systems (SIEM, incident management). This should help with the use case you described where the event ID comes from an external inspection service response.

Let me know if the new attribute meets your needs or if you'd like any adjustments!

docs/registry/attributes/gen-ai.md
nagkumar91: Thanks @habibam! I've addressed this in the commit (88c269b7):

- Updated `gen_ai.guardian.*` attribute documentation to clarify these identify the **evaluating service/component**, not the policy
- Added explicit mapping guidance:
  - Guardian = service doing evaluation (e.g., "Azure Content Safety", "Bedrock Guardrails")  
  - Policy = configuration being applied (use `gen_ai.security.policy.id` for ARNs, blocklist IDs, etc.)
- Updated examples to be clearer

This should make cross-provider correlation cleaner by keeping guardian and guardrail/policy semantics distinct. Let me know if the updated documentation captures your concern!

prototype/frameworks/mcp/guardian_adapter.py
nagkumar91: Added in commit 88c269b7! 

New elicitation guard methods:
- `guard_elicitation_request()` - Guards outbound requests to user (prevents info leakage, excessive elicitation)
- `guard_elicitation_response()` - Guards user's response (detects PII, injection attempts)

Both map to `target_type=message` since they're user-facing interactions.

See the updated module docstring and __main__ examples (tests 6-9) for usage patterns.

nagkumar91:
Non-normative implementation guide (PR branch):
https://github.com/nagkumar91/semantic-conventions/blob/gen-ai-security-guardian/docs/gen-ai/non-normative/security_implementation_gen_ai_spec.md

Note on scope:
- All files under `prototype/` are included for reference and to help understand/simulate the spans (stories, adapters, viewer).
- These `prototype/` files are intended to be removed before merge.

alexgarden-mnemom (commented):
Great work on this, Nagkumar. We've been running shipping OTel exporters for AI agent safety checks (@mnemom/aip-otel-exporter on npm, aip-otel-exporter on PyPI) with production traces in Grafana Cloud Tempo — 50+ span attributes across six span types.

A few suggestions based on what we've learned from production:

**1. Dedicated span type:** Making safety checks a dedicated safety_check span operation (child of invoke_agent or chat) gives you independent observability — you can measure safety check latency, build verdict distribution dashboards, and alert without parsing parent spans.

**2. Three-value verdict:** Boolean triggered misses the middle ground. In production, we've found every major provider has intermediate states — Azure severity 2-5, Google LOW/MEDIUM/HIGH, OpenAI varying confidence. A pass/review/fail enum maps cleanly 
to all of them.

**3. Structured concern events:** When a guardrail fires, operators need structured detail. We emit gen_ai.safety.concern events (one per finding) with category, severity, and description. Gives you queryable data vs a boolean.

**4. Policy model:** Beyond guardrail.id — a full policy model (id/name/version/mode) covers enterprise deployments where teams roll out governance gradually. policy.mode (monitor/warn/enforce/off) is critical for staged rollouts.

**5. Continuous score:** A safety.score (0.0-1.0) alongside the verdict gives you a continuous signal for trending and thresholding. Every provider has a numeric score underneath their categorical output.

**6. Metrics:** A duration histogram and verdict counter would let teams build standard dashboards for safety check performance without custom queries.

We have a full provider mapping showing how these conventions apply across the ecosystem — AWS Bedrock Guardrails, Azure Content Safety, OpenAI Moderation, Google Safety Ratings, and our own AIP/AAP protocols. The proposed attributes map cleanly to all of them.

I think there's a path to combining your guardrail-specific work with a broader safety_check framework that covers guardrails, content filtering, integrity analysis, alignment verification, and policy evaluation as check types under a unified convention.

model/gen-ai/registry.yaml
```diff
@@ -359,6 +359,270 @@ groups:
...
+      - id: gen_ai.security.content.input.hash
+        stability: development
+        type: string
+        brief: Hash of the input content for forensic correlation.
+        note: >
+          Use when full content capture is not desired but correlation is needed.
+          Instrumentations SHOULD document the algorithm and any salting/HMAC
+          approach used.
+        examples: ['sha256:a3f2b8c9...']
+      - id: gen_ai.security.content.redacted
```
aabmass: `gen_ai.security.decision.type` enum has value `modified`. Can we align them one way or the other?


```suggestion
      - id: gen_ai.security.content.modified
```

model/gen-ai/spans.yaml
```diff
@@ -430,6 +430,79 @@ groups:
           the canonical name of exception that occurred, or another low-cardinality error identifier.
           Instrumentations SHOULD document the list of errors they report.
 
+  - id: span.gen_ai.apply_guardrail.internal
+    type: span
+    stability: development
+    span_kind: internal
+    brief: Describes a security guardian evaluation.
```
aabmass, line 447: What do you think of aligning on **`guardrail`** throughout this proposal? The GCP Model Armor team found "guardian" a bit esoteric and I agree. At least we should be consistent throughout.

Lmk if this has already been discussed

model/gen-ai/registry.yaml
```diff
@@ -359,6 +359,270 @@ groups:
...
+
+          Example mapping:
+
+          - Guardian: The evaluation service (e.g., "bedrock_guardrail_service", "azure_content_safety")
+          - Policy: The specific configuration (e.g., "arn:aws:bedrock:us-east-1:123456789:guardrail/abc123")
+        examples: ['guard_abc123', 'content-filter-v2', 'custom-guardian-001']
+      - id: gen_ai.guardian.name
+        stability: development
+        type: string
+        brief: Human-readable name of the guardian (the evaluating service or component).
```
aabmass, line 387: What is the difference between `guardian.name` and `guardian.id`?

The examples imply it's a static string, so I wonder if it belongs in instrumentation scope object or resource.

model/gen-ai/registry.yaml
```diff
@@ -359,6 +359,270 @@ groups:
...
+        note: >
+          This identifies the security evaluation service or component, not the
+          policy name. For policy names, use `gen_ai.security.policy.name`.
+        examples: ['Azure Content Safety', 'Bedrock Guardrails', 'Custom PII Filter']
+      - id: gen_ai.guardian.version
```
aabmass, line 392: Same comment as above, this probably belongs on the scope. @lmolkova is there a way to specify that in YAML?

model/gen-ai/spans.yaml
```diff
@@ -430,6 +430,79 @@ groups:
...
+    attributes:
+      - ref: gen_ai.operation.name
+        requirement_level: required
+      - ref: gen_ai.guardian.id
+        requirement_level:
+          conditionally_required: if available
+      - ref: gen_ai.guardian.name
+        requirement_level: recommended
+      - ref: gen_ai.guardian.version
+        requirement_level: recommended
```
aabmass, line 467: Would we expect this on a cloud / SaaS offering?

aabmass (commented):
I got some feedback from the GCP Model Armor team, which I'm sharing here

model/gen-ai/spans.yaml
nagkumar91: Yes, we'd still expect it on cloud/SaaS — it's useful in hybrid pipelines where you mix providers (e.g., Azure Content Safety + a custom local filter + a third-party DLP). It's already `recommended` (not required), so single-provider setups can skip it without any penalty.

model/gen-ai/registry.yaml
nagkumar91: Same reasoning as above — in a multi-guardrail pipeline, different guardrails may have different versions. Keeping it per-span allows correlating specific version behavior with specific decisions (e.g., "v1.0 allowed this but v2.0 would have blocked it").

It's `recommended`, so instrumentations can skip it when it's static/obvious.

model/gen-ai/spans.yaml
nagkumar91: Good point — we're discussing this internally with the team. "Guardrail" is clearly the more established industry term (AWS Bedrock Guardrails, NVIDIA NeMo Guardrails, Guardrails AI). The operation name `apply_guardrail` already uses it, so renaming `gen_ai.guardian.*` → `gen_ai.guardrail.*` would make everything consistent.

Will follow up once we have internal alignment — likely agreeing with your suggestion.

model/gen-ai/registry.yaml
nagkumar91: Agreed — renamed `gen_ai.security.content.redacted` → `gen_ai.security.content.modified` in commit 5e4bbe93. The broader term "modified" correctly covers redaction, sanitization, and rewriting, and aligns with the `modify` value in the `gen_ai.security.decision.type` enum.

model/gen-ai/registry.yaml
nagkumar91: `guardian.id` is the machine-readable identifier (e.g., `content-filter-v2`), while `guardian.name` is the human-readable display name (e.g., "Azure Content Safety"). Here's a trace showing both:

```
chat "gpt-4"                                                         (1200ms)
│
├── apply_guardrail "Custom PII Filter"                              (30ms)
│   ├── gen_ai.guardian.id: "pii-filter-v3"              ← machine-readable
│   ├── gen_ai.guardian.name: "Custom PII Filter"        ← human-readable
│   ├── gen_ai.guardian.version: "3.1.0"
│   ├── gen_ai.guardian.provider.name: "custom"
│   ├── gen_ai.security.decision.type: "modify"
│   ├── gen_ai.security.content.modified: true
│   └── gen_ai.security.target.type: "input"
│
├── apply_guardrail "Azure Content Safety"                           (45ms)
│   ├── gen_ai.guardian.id: "content-filter-v2"          ← different id
│   ├── gen_ai.guardian.name: "Azure Content Safety"     ← different name
│   ├── gen_ai.guardian.version: "2024-05-01"
│   ├── gen_ai.guardian.provider.name: "azure.ai.content_safety"
│   ├── gen_ai.security.decision.type: "deny"
│   ├── gen_ai.security.decision.reason: "toxicity_detected"
│   └── gen_ai.security.target.type: "output"
│
└── apply_guardrail "Prompt Shield"                                  (25ms)
    ├── gen_ai.guardian.id: "prompt-shield-v1"            ← different id
    ├── gen_ai.guardian.name: "Prompt Shield"              ← different name
    ├── gen_ai.guardian.version: "1.0.0"
    ├── gen_ai.guardian.provider.name: "azure.ai.content_safety"
    ├── gen_ai.security.decision.type: "allow"
    └── gen_ai.security.target.type: "input"
```

**On scope vs span**: We kept these per-span because real-world pipelines chain multiple guardrails under a single LLM call (PII filter on input + toxicity filter on output + prompt injection detector). If these were on the instrumentation scope, you couldn't distinguish which guardrail made which decision. The spec notes: *"Multiple guardian spans MAY exist under a single operation span if multiple guardians are chained."*

The `.id` is what you'd use in queries/filters/alerts (stable, machine-readable), while `.name` shows up in dashboards and trace UIs (human-readable). Since `.name` is `recommended` (not required), instrumentations can skip it if it's redundant with the id.

nagkumar91:
## Supporting Materials

The non-normative implementation spec, prototype code, and story scripts have been moved to gists to keep this PR focused on the YAML model + generated docs.

| Gist | Description |
|---|---|
| 📄 [Non-Normative Implementation Spec](https://gist.github.com/nagkumar91/95efa05449c72b4c95958190f26f13ba) | Detailed spec with framework mappings, attribute definitions, and trace examples |
| 🔧 [Prototype Framework Code](https://gist.github.com/nagkumar91/7328b4be7f80d01db087929382ab6cab) | OTel bootstrap, guardian utils, base adapter, and LangChain guardian adapter |
| 📖 [Story Scripts & Trace Coverage](https://gist.github.com/nagkumar91/2be5a18ce1fc221f917b8d8230376f03) | 5 end-to-end scenarios (enterprise RAG, multi-tenant, multi-agent, progressive jailbreak, error handling) with trace coverage matrix |

Previously these were in draft PR open-telemetry/semantic-conventions#3434, which can now be closed.

docs/gen-ai/gen-ai-security.md
```diff
@@ -0,0 +1,279 @@
...
+<!-- prettier-ignore-start -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+Describes a security guardian evaluation.
+
+`gen_ai.operation.name` SHOULD be `apply_guardrail`.
+
+**Span name** SHOULD be `apply_guardrail {gen_ai.guardian.name}`.
+When `gen_ai.guardian.name` is not available, it SHOULD be `apply_guardrail {gen_ai.security.target.type}`.
```
Cirilla-zmh, line 46: It's better to add : "Semantic conventions for individual guardians and frameworks MAY specify different span name format."

docs/gen-ai/gen-ai-security.md
```diff
@@ -0,0 +1,279 @@
...
+
+Guardian spans SHOULD be children of the operation span they are protecting (for example,
+`span.gen_ai.inference.client` or `span.gen_ai.execute_tool.internal`).
```
Cirilla-zmh, line 49: I believe it's better to leave a reference to the markdown sections here rather than the name of the registry.

```suggestion
[inference](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md#inference) or [execute tool span](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md#execute-tool-span)).
```

docs/gen-ai/gen-ai-security.md
```diff
@@ -0,0 +1,279 @@
...
+
+Multiple guardian spans MAY exist under a single operation span if multiple guardians are chained.
+
+**Span kind** SHOULD be `INTERNAL`.
```
Cirilla-zmh, line 53: The span kind is suitable to be `CLIENT` in case of applying the guardian running in another process.

docs/gen-ai/gen-ai-security.md
```diff
@@ -0,0 +1,279 @@
...
+| [`gen_ai.security.decision.type`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | The decision made by the security guardian. | `allow`; `deny`; `modify` |
+| [`gen_ai.security.target.type`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | The type of content or action the guardrail is applied to. [2] | `tool_call`; `llm_input`; `llm_output`; `tool_definition` |
+| [`error.type`](/docs/registry/attributes/error.md) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | `Conditionally Required` if the operation ended in an error | string | Describes a class of error the operation ended with. [3] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` |
+| [`gen_ai.agent.id`](/docs/registry/attributes/gen-ai.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` when applicable | string | The unique identifier of the GenAI agent. | `asst_5j66UpCpwteGg4YSxUnt7lPY` |
```
Cirilla-zmh, line 65: I think `gen_ai.agent.id` should be a traffic-related attribute and should be propagated along with the context. This is because the lifecycle of agents does not always align with that of services, and even their scopes may differ. An agent may consist of a group of services, and a single service may also contain multiple agents.

Given this, `gen_ai.agent.id` does not seem appropriate here. Perhaps we could clarify this in agent-related spans, especially since this attribute is not present in inference spans or execute tool spans either.

Cirilla-zmh: Let's discuss it if you be interested in this: https://github.com/open-telemetry/semantic-conventions/issues/2912#issuecomment-4116296207

model/gen-ai/registry.yaml
```diff
@@ -364,6 +364,270 @@ groups:
...
+          - 'Action exceeds agent permission scope'
+      - id: gen_ai.security.decision.code
+        stability: development
+        type: int
+        brief: Numeric code for the security decision (provider-specific).
+        examples: [112, 403, 5001]
+
+      - id: gen_ai.security.target.type
+        stability: development
+        type: string
```
Cirilla-zmh, line 468: Why wouldn't `member` work here?

model/gen-ai/registry.yaml
aabmass: Makes sense. Can you just clarify it and clean up the examples? Feel free to resolve.

aabmass: Actually, what do you think of using `gen_ai.provider.name` or changing it to `gen_ai.guardian.provider.name` (as an enum)?. That would align with what we have for inference providers https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md#gen-ai-provider-name

model/gen-ai/spans.yaml
```diff
@@ -440,6 +440,79 @@ groups:
...
+  - id: span.gen_ai.apply_guardrail.internal
+    type: span
+    stability: development
+    span_kind: internal
+    brief: Describes a security guardian evaluation.
+    note: |
+      `gen_ai.operation.name` SHOULD be `apply_guardrail`.
+
+      **Span name** SHOULD be `apply_guardrail {gen_ai.guardian.name}`.
+      When `gen_ai.guardian.name` is not available, it SHOULD be `apply_guardrail {gen_ai.security.target.type}`.
```
aabmass: I found this a little confusing since the fallback is pretty different from the primary. IIUC it would be 
- Primary: `apply_guardrail azure_ai_foo`
- Fallback: `apply_guardrail tool_call`

And if you have multiple chained guardians from the same `gen_ai.guardian.name`, all the spans would have the same name?

docs/gen-ai/gen-ai-security.md
habibam: I agree that a single service can host multiple agents, but I'd push back on the inverse (that a single agent spans multiple services). The agent is the reasoning loop: the component that holds the system prompt, uses an LLM as its reasoning engine, interprets the response, and decides what to do next. The LLM isn't a tool the agent calls (it's the agent's decision-maker) _but_ the LLM service hosting that inference (e.g., an Azure OpenAI endpoint) is shared infra, not a component of the agent, same way an application isn't "composed of" a database just because it queries one. Every major framework models it this way (AutoGen, Semantic Kernel, LangGraph, CrewAI) where an agent has a reasoning loop and a tool manifest. The tools can live anywhere (local functions, remote APIs/MCP servers, even other agents) but they don't carry the agent's identity. They're what the agent uses, not what it is. This matters for how we think about `gen_ai.agent.id` in the semantic conventions. The attribute should describe which agent's reasoning loop produced a span, not redefine the agent's boundary to include every service in its call graph. An agent using an LLM endpoint doesn't make that endpoint part of the agent, just as an agent calling a retrieval service doesn't make retrieval part of its identity.

model/gen-ai/registry.yaml
habibam: I'd lean toward gen_ai.guardrail.provider.name (as an enum). Reusing gen_ai.provider.name would be ambiguous. For example, an agent could use AOAI for inference `(client.chat.completions.create())` and Model Armor for content filtering `(client.sanitize_user_prompt())` in the same trace. Keeping them as separate attributes makes that distinction clear.

model/gen-ai/spans.yaml
habibam: Yeah I agree, good catch, the fallback also shifts the semantics of the span name from _who_ is evaluating to _what_ is being evaluated, which is confusing. And yeah, chaining the same guardrail on input and output would produce identical span names. Something like `apply_guardrail {guardian.name} {target.type}` would keep it consistent and unique across chained pipelines.

docs/gen-ai/gen-ai-security.md
nagkumar91: @Cirilla-zmh    Valid point, but agent identity propagation is a broader gen-ai spec question - it affects agent spans, inference spans, and tool spans equally, not just security guardrails. I'll mark this as resolved as we can take it up here - https://github.com/open-telemetry/semantic-conventions/issues/2912#issuecomment-4116296207 (as you suggested).

model/gen-ai/spans.yaml
nagkumar91: Addressed in commit 1fc46dc8 — span name is now `apply_guardrail {gen_ai.guardian.name} {gen_ai.security.target.type}`, making chained spans unique (e.g., "apply_guardrail PII Filter input" vs "apply_guardrail PII Filter output"). The fallback when guardian.name is unavailable is `apply_guardrail {gen_ai.security.target.type}`.

model/gen-ai/registry.yaml
nagkumar91: Changed `gen_ai.guardian.provider.name` to an open enum with well-known values: `azure.ai.content_safety`, `aws.bedrock`, `gcp.model_armor` (commit 1fc46dc8). This keeps it distinct from `gen_ai.provider.name` (inference provider), as @habibam suggested.

model/gen-ai/spans.yaml
nagkumar91: Done — adopted your suggestion in commit 1fc46dc8. Span name is now `apply_guardrail {gen_ai.guardian.name} {gen_ai.security.target.type}`.

model/gen-ai/registry.yaml
nagkumar91: Cleaned up in commit 1fc46dc8.

nagkumar91: Agreed — implemented as a separate `gen_ai.guardian.provider.name` enum in commit 1fc46dc8. Keeps inference vs guardrail providers distinct.

Nick-heo-eg:
This looks great, modeling guardrail enforcement as a separate concern makes a lot of sense.

One thing I noticed while reading through this: there seems to be a gap between evaluation and guardrail in terms of what actually gets captured in telemetry.

Right now:
- evaluation handles measurement signals (score, label)
- guardrail handles enforcement actions (allow, deny, modify)

But there's no clear way to represent the discrete result an evaluator produces when it isn't tied to any enforcement decision.

For example:
- an evaluator classifies content as "safe/unsafe" or "pass/fail"
- that result exists on its own, without triggering a guardrail
- and there's no numeric score involved

In that situation, the output isn't a continuous score, and it isn't an enforcement action either, it's a categorical evaluation result that isn't directly captured today.

There may be room for a small, vendor-neutral way to capture this kind of discrete evaluation outcome, sitting between score and guardrail decision.

model/gen-ai/registry.yaml
```diff
@@ -364,6 +364,287 @@ groups:
...
+              brief: Request permitted with modifications (redaction, sanitization).
+              stability: development
+            - id: warn
+              value: "warn"
+              brief: Request permitted but flagged for review.
+              stability: development
+            - id: audit
+              value: "audit"
+              brief: Request logged for audit purposes only, no enforcement.
+              stability: development
```
MohammadHaroonAbuomar, line 446: Guardrail Services describe a verdict that should happen but don't enforce it. We should tighten the language here that these are the verdicts, which is not the same as the enforcements.

E.g. AACS can return a warn verdict, but the caller may choose to handle it as a block (a typical use case in production). In this case the verdict is warn, but the enforcement is block.

model/gen-ai/registry.yaml
```diff
@@ -364,6 +364,287 @@ groups:
...
+        note: >
+          This identifies the guardrail provider, which may differ from the inference
+          provider (`gen_ai.provider.name`). For example, an agent may use OpenAI for
+          inference and Azure AI Content Safety for content filtering in the same trace.
+        examples: ['azure.ai.content_safety', 'aws.bedrock', 'gcp.model_armor']
+
+      - id: gen_ai.security.decision.type
+        stability: development
+        type:
+          members:
```
MohammadHaroonAbuomar, line 426: Missing an "escalate" verdict for human in the loop/approval workflows.

model/gen-ai/registry.yaml
```diff
@@ -364,6 +364,287 @@ groups:
...
+        type: string
+        brief: Version of the guardian service or component.
+        note: >
+          This is the version of the evaluating service/component, not the policy
+          version. For policy versions, use `gen_ai.security.policy.version`.
+        examples: ['1.0.0', '2024-05-01', 'v2']
+      - id: gen_ai.guardian.provider.name
+        stability: development
+        type:
+          members:
```
MohammadHaroonAbuomar, line 403: Should this be a string instead? We should support cases where someone is running custom guardrails implementation outside of the major providers.

model/gen-ai/registry.yaml
```diff
@@ -364,6 +364,287 @@ groups:
...
+          Suggested values include:
+
+          - `llm_input`: Prompt or messages being sent to the model.
+          - `llm_output`: Response or messages returned from the model.
+          - `tool_call`: Tool execution request (tool call invocation).
+          - `tool_definition`: Tool definition or schema (tool declaration).
+          - `memory_store`: Data being written to memory.
+          - `memory_retrieve`: Data being read from memory.
+          - `knowledge_query`: Knowledge/RAG query being sent.
+          - `knowledge_result`: Knowledge/RAG results being returned.
```
MohammadHaroonAbuomar, line 483: Aren't these tool calls also? tool call should also be split into input and output.

model/gen-ai/registry.yaml
```diff
@@ -364,6 +364,287 @@ groups:
...
+          - 'prompt_injection'
+          - 'sensitive_info_disclosure'
+          - 'jailbreak'
+          - 'custom:financial_advice_violation'
+          - 'azure:hate_speech'
+      - id: gen_ai.security.risk.severity
+        stability: development
+        type:
+          members:
+            - id: none
```
MohammadHaroonAbuomar, line 528: No risk is a none-existent state. Suggest to change to unspecified/unknown or something along those lines.

model/gen-ai/registry.yaml
```diff
@@ -364,6 +364,287 @@ groups:
...
+
+          Instrumentations MAY use additional values when appropriate, for
+          example: `jailbreak`, `toxicity`, `pii`, `custom:*`, `azure:*`.
+        examples:
+          - 'prompt_injection'
+          - 'sensitive_info_disclosure'
+          - 'jailbreak'
+          - 'custom:financial_advice_violation'
+          - 'azure:hate_speech'
+      - id: gen_ai.security.risk.severity
```
MohammadHaroonAbuomar, line 524: I would remove severity and keep score. Severity means different things for different applications/use cases, while a score is more independent/factual.

model/gen-ai/registry.yaml
```diff
@@ -364,6 +364,287 @@ groups:
...
+              brief: Critical severity risk requiring immediate action.
+              stability: development
+        brief: The severity level of the detected risk.
+        examples: ['low', 'medium', 'high', 'critical']
+      - id: gen_ai.security.risk.score
+        stability: development
+        type: double
+        brief: Numeric risk/confidence score (0.0 to 1.0).
+        examples: [0.85, 0.95, 0.42]
+      - id: gen_ai.security.risk.metadata
```
MohammadHaroonAbuomar, line 555: What's the intended use for this?

model/gen-ai/registry.yaml
```diff
@@ -364,6 +364,287 @@ groups:
...
+
+          - Field identifiers: `field:bcc`, `parameter:query`
+          - Pattern types: `pattern:ssn`, `pattern:credit_card`
+          - Counts: `count:3`
+          - Positions: `position:input[0].content`
+        examples:
+          - ['field:bcc', 'pattern:email']
+          - ['count:2', 'position:output.content']
+
+      - id: gen_ai.security.policy.id
```
MohammadHaroonAbuomar, line 574: In the case of a detailed policy under the same identifier, this policy id might be underspecified since a section/part of the larger policy triggered the verdict. I think it's useful to have this metadata, not sure how to make it "tight enough".

model/gen-ai/spans.yaml
MohammadHaroonAbuomar: Agree that guardian is interesting :) We're moving towards "controls" as the canonical name instead of guardrails/content filters back in the day.

wolfgangcodes:
Hello @nagkumar91! 

With the merging of [Move GenAI semantic conventions to its own dedicated repository](semantic-conventions#3696), the [`area:gen-ai`](https://github.com/open-telemetry/semantic-conventions/labels?q=area%3Agen-ai) semantic conventions have a new home. We've made this split in an to be better able to support the rapid pace of change that we are seeing with Gen AI related semantic conventions.

Our ask is that you move this PR to the new repo here: https://github.com/open-telemetry/semantic-conventions-genai

The contribution guide for the new repo can be found here: https://github.com/open-telemetry/semantic-conventions-genai/blob/main/CONTRIBUTING.md. A major difference is making sure there's a representative scenario and reference outlined in:

>  **4. Update reference scenarioss**
> Changes under model/ or docs/ typically require updating the reference scenarios under reference/ to demonstrate that the proposed updates are capturable.  
> 

You may also find that you need to additional updates to you proposal as the new repo use the [V2 Schema](https://github.com/open-telemetry/weaver/blob/main/schemas/semconv-schemas.md).

If you would like to discuss or have questions you can reply here, post in slack [here](https://cloud-native.slack.com/archives/C06KR7ARS3X), or add an agenda item for the [Gen AI SIG](https://github.com/open-telemetry/community#sig-genai-instrumentation). 

Thank you so much!

lmolkova:
Closing this PR here, but please consider reintroducing it in the new repo!

PR #3249, opened 2026-01-06
Add invoke_workflow operation name span

wrisa:
Fixes #
Issue: [2912](https://github.com/open-telemetry/semantic-conventions-genai/issues/55) partly

| Agentic framework | Mapping to Workflow | References |
| :--- | :---: | ---: |
| Langchain/LangGraph | First chain can indicate start of the workflow | [workflow agents](https://docs.langchain.com/oss/python/langgraph/workflows-agents) |
| CrewAI| Crew kickoff can start workflow | [Crew kickoff](https://docs.crewai.com/en/concepts/agents#direct-agent-interaction-with-kickoff) |
| OpenAI agents| Higher level Trace can be started as workflow | [Higher trace](https://openai.github.io/openai-agents-python/tracing/) |
| Custom/In-house| Multiple LLMs or agents can be grouped under a workflow | [In-house](https://github.com/signalfx/splunk-otel-python-contrib/blob/main/instrumentation-genai/opentelemetry-instrumentation-langchain/examples/multi_agent_travel_planner/in_house_version/README.md) |

## Changes

Workflow proposal https://github.com/open-telemetry/semantic-conventions/blob/b6bfc44631e28e87a0252011a8898e8098cb0ff9/docs/gen-ai/Session-and-Workflow-for-GenAI-Observability.md discussion led to suggestion of invoke workflow operation name in agent span.

> [!IMPORTANT]


model/gen-ai/registry.yaml
```diff
@@ -591,3 +595,13 @@ groups:
         type: string
         brief: A free-form explanation for the assigned score provided by the evaluator.
         examples: ["The response is factually accurate but lacks sufficient detail to fully address the question."]
+      - id: gen_ai.workflow.name
```
JWinermaSplunk, line 667: It would be good to link an existing instrumentation for workflow that has these attributes, or just a general doc that contains values for these attributes as an enum, etc, under prototypes in the PR description.

model/gen-ai/spans.yaml
```diff
@@ -349,3 +349,26 @@ groups:
         requirement_level: required
       - ref: aws.bedrock.knowledge_base.id
         requirement_level: recommended
+
+  - id: span.gen_ai.invoke_workflow.client
```
JWinermaSplunk: should this `extends: attributes.gen_ai.common.client`? this will add server common attributes relating to genai spans, like `error.type`, to the span

model/gen-ai/registry.yaml
wrisa: There is a table of attributes with sample values in docs/gen-ai/gen-ai-agent-spans.md for `Invoke workflow`

model/gen-ai/registry.yaml
```diff
@@ -596,3 +600,13 @@ groups:
...
+        stability: development
+        type: string
+        brief: Human-readable name of the GenAI workflow provided by the application.
+        examples: [ "multi_agent_rag", "customer_support_pipeline" ]
+      - id: gen_ai.framework
+        stability: development
+        type: string
+        brief: Framework implementing the workflow.
+        examples: [ "LangGraph", "CrewAI", "OpenAI agents" ]
```
91pavan: Should we also look at adding a new attribute to describe or capture the unique identifier of the GenAI workflow like `gen_ai.workflow.id`?

91pavan (commented):
Looks good.

model/gen-ai/registry.yaml
```diff
@@ -591,8 +595,13 @@ groups:
...
         brief: A free-form explanation for the assigned score provided by the evaluator.
         examples: ["The response is factually accurate but lacks sufficient detail to fully address the question."]
-      - id: gen_ai.prompt.name
+      - id: gen_ai.workflow.name
+        stability: development
         type: string
-        brief: The name of the prompt that uniquely identifies it.
+        brief: Human-readable name of the GenAI workflow provided by the application.
+        examples: [ "multi_agent_rag", "customer_support_pipeline" ]
+      - id: gen_ai.framework
```
zhirafovod: OK to have it in the PR, but may be moved to a separate PR to not conflate with the workflow

model/gen-ai/spans.yaml
```diff
@@ -400,3 +349,26 @@ groups:
...
+    type: span
+    span_kind: client
+    stability: development
+    brief: >
+      Represents a workflow orchestrating multiple agents with predefined code paths.
+    note: |
+      The `gen_ai.operation.name` SHOULD be `invoke_agent`.
```
zhirafovod: should be `invoke_workflow`

model/gen-ai/spans.yaml
```diff
@@ -400,3 +349,26 @@ groups:
...
+      **Span name** SHOULD be `invoke_workflow {gen_ai.workflow.name}`.
+    attributes:
+      - ref: gen_ai.workflow.name
+        requirement_level:
+          conditionally_required: when available
+      - ref: gen_ai.framework
```
zhirafovod: OK to leave it, but can move to a separate PR

zhirafovod (approved):
LGTM with some conditional notes left on the PR.

model/gen-ai/registry.yaml
```diff
@@ -596,3 +600,8 @@ groups:
         brief: The name of the prompt that uniquely identifies it.
         stability: development
         examples: ["analyze-code"]
+      - id: gen_ai.workflow.name
+        stability: development
+        type: string
+        brief: Human-readable name of the GenAI workflow provided by the application.
```
lmolkova, line 670: is it possible to add a note that would explain how to populate it for a couple of frameworks - crioai, langchain

model/gen-ai/spans.yaml
```diff
@@ -400,3 +400,24 @@ groups:
         requirement_level: required
       - ref: aws.bedrock.knowledge_base.id
         requirement_level: recommended
+
+  - id: span.gen_ai.invoke_workflow.client
+    type: span
+    span_kind: client
```
lmolkova, line 457: Consider 
```suggestion
    span_kind: internal
```

workflow always has underlying spans that call external services

model/gen-ai/registry.yaml
wrisa: @91pavan Can we add this in a follow-up PR ?

model/gen-ai/registry.yaml
```diff
@@ -596,3 +600,10 @@ groups:
...
+        examples: [ "multi_agent_rag", "customer_support_pipeline" ]
+        note: |
+          Tbis attribute can be populated in different frameworks eg: name of the trace in CrewAI, name of first the first chain in LangChain, etc.
```
kamphaus: Typo + duplicate word
```suggestion
          This attribute can be populated in different frameworks eg: name of the trace in CrewAI, name of the first chain in LangChain, etc.
```

model/gen-ai/spans.yaml
```diff
@@ -400,3 +400,24 @@ groups:
         requirement_level: required
       - ref: aws.bedrock.knowledge_base.id
         requirement_level: recommended
+
+  - id: span.gen_ai.invoke_workflow.internal
```
lmolkova, line 480: GenAI SIG discussion:

- we want to have a separate metric for workflows (maybe some other operations, but that's not related to this PR) because it has a specific set of attributes and also duration that's significantly different than a single llm call
- it's ok to do as a follow up, but please create an issue for this so we don't forget

.chloggen/genai-invoke-workflow-operation-name.yaml
```diff
@@ -0,0 +1,22 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'new_component'
```
kamphaus: Shouldn't this be an enhancement?
```suggestion
change_type: 'enhancement'
```

.chloggen/genai-invoke-workflow-operation-name.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: gen-ai
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add invoke workflow operation name support to the gen-ai agent span semantic conventions.
```
singankit:
```suggestion
note: Add invoke workflow operation name support to the gen-ai semantic conventions.
```

singankit (approved):

singankit:
@wrisa Is there any active discussions or doc about conditional nodes and how they can be represented, especially with gen ai workflows. For example [conditional edges in langgraph](https://docs.langchain.com/oss/python/langgraph/graph-api#conditional-edges).

wrisa:
> https://github.com/open-telemetry/semantic-conventions/blob/b6bfc44631e28e87a0252011a8898e8098cb0ff9/docs/gen-ai/Session-and-Workflow-for-GenAI-Observability.md

Not exactly, we briefly presented `step` for [intermediate steps](https://github.com/keith-decker/semantic-conventions/blob/7c295bedcc918b00ca37f3124211bd5adee838d9/docs/gen-ai/gen-ai-agent-proposal.md#step-span) like langchain's chains or CrewAi's tasks. But according to SIG discussions, it seemed expensive as all the info could be captured on agent span as well and also it seemed more framework specific. If we can have more use cases like you suggested we can start the discussion maybe again.
P.S: We added assigned agents attribute in step for the chain invoking agent, crewAI's task executed by an agent and can be used for nodes in conditional edge. May be conditional edge can be a 'step' or an extension to it.

wrisa:
> Is there any active discussions or doc about conditional nodes and how they can be represented, especially with gen ai workflows

@singankit  Not exactly, but we briefly presented step for [intermediate steps](https://github.com/keith-decker/semantic-conventions/blob/7c295bedcc918b00ca37f3124211bd5adee838d9/docs/gen-ai/gen-ai-agent-proposal.md#step-span) like langchain's chains or CrewAi's tasks. According to SIG discussions, it seemed expensive as all the info could be captured on agent span as well and also it seemed more framework specific. If we can have more use cases like you suggested we can start the discussion maybe again.
P.S: We added assigned agents attribute in step for the chain invoking agent, crewAI's task executed by an agent and can be used for nodes in conditional edge. May be conditional edge can be a 'step' or an extension to it.

.chloggen/genai-invoke-workflow-operation-name.yaml
wrisa: fixed. Thank you.

.chloggen/genai-invoke-workflow-operation-name.yaml
wrisa: fixed. Thank you.

model/gen-ai/spans.yaml
wrisa: @lmolkova added issue for workflow metrics https://github.com/open-telemetry/semantic-conventions/issues/3318

model/gen-ai/spans.yaml
```diff
@@ -432,3 +432,24 @@ groups:
...
+    type: span
+    span_kind: internal
+    stability: development
+    brief: >
+      Represents a workflow orchestrating multiple agents with predefined code paths.
```
lmolkova: I think it's usually multiple agents, but I don't believe there are any means to guarantee it.
"Also not sure if predefined code paths" is substantial, I think it does not matter?

```suggestion
      Represents a workflow orchestrating one or more agents.
```

model/gen-ai/spans.yaml
```diff
@@ -432,3 +432,24 @@ groups:
...
+    note: |
+      The `gen_ai.operation.name` SHOULD be `invoke_workflow`.
+      **Span name** SHOULD be `invoke_workflow {gen_ai.workflow.name}`.
```
lmolkova, line 489: I think we need to provide a bit more guidance on when this span should (not) be reported


```suggestion
      **Span name** SHOULD be `invoke_workflow {gen_ai.workflow.name}`.
      
      This span SHOULD be reported by the instrumentations when they can
      reliably determine that invocation is a workflow (i.e. groups several agent
      invocations) and SHOULD NOT be reported by instrumentations that 
      can't distinguish it `invoke_workflow` from `invoke_agent`.
```

(please correct if you see any issues)

model/gen-ai/spans.yaml
```diff
@@ -432,3 +432,24 @@ groups:
...
+    extends: attributes.gen_ai.common.client
+    attributes:
+      - ref: gen_ai.workflow.name
+        requirement_level:
+          conditionally_required: when available
+      - ref: gen_ai.input.messages
+        requirement_level: opt_in
+      - ref: gen_ai.output.messages
+        requirement_level: opt_in
+      - ref: gen_ai.system_instructions
```
lmolkova: is this common for workflows to have system instructions along with input messages?

wrisa: Workflow will not have its own but it could be combined from agents->llm's system messages.

model/gen-ai/spans.yaml
```diff
@@ -432,3 +432,29 @@ groups:
...
+    brief: >
+      Represents a workflow orchestrating one or more agents.
+    note: |
+      The `gen_ai.operation.name` SHOULD be `invoke_workflow`.
+      **Span name** SHOULD be `invoke_workflow {gen_ai.workflow.name}`.
+
+      This span SHOULD be reported by the instrumentations when they can
+      reliably determine that invocation is a workflow (i.e. groups several agent
+      invocations) and SHOULD NOT be reported by instrumentations that
+      can't distinguish it `invoke_workflow` from `invoke_agent`.
```
aabmass, line 494: How would you recommend instrumenting ADK?

model/gen-ai/spans.yaml
aabmass: I have the same concern. Does a workflow ever actually invoke a model or is it **staticly defined ** control flow?

wrisa: It is the later. Remove it.

docs/gen-ai/azure-ai-inference.md
```diff
@@ -1,4 +1,4 @@
-<!--- Hugo front matter used to generate the website version of this page:
+  <!--- Hugo front matter used to generate the website version of this page:
```
lmolkova: intended?

model/gen-ai/spans.yaml
lmolkova: it still seems to be client

model/gen-ai/spans.yaml
lmolkova: @wrisa could you please comment this? 

My understanding that ADK cannot reliably tell if it's used as an orchestrator or a lower-level agent and should report regular invoke_agent span. @aabmass do you think we could express it more clearly?

wrisa: @lmolkova @aabmass  ADK has [workflow agents](https://google.github.io/adk-docs/agents/workflow-agents/) which can be a differentiator to invoke_agent.

docs/gen-ai/azure-ai-inference.md
wrisa: no, will remove

model/gen-ai/spans.yaml
aabmass: I think we can actually tell if it's a workflow agent and do this, but it seems duplicative to have both.

Have you tried out an ADK agent with this @wrisa? It actually has native instrumentation already and I believe it would output invoke_agent with the workflow name.

aabmass: So I tried it out in ADK to see what happens. With this agent

```py
from google.adk.agents import LlmAgent, SequentialAgent

hello_agent = LlmAgent(
    model="gemini-2.5-flash",
    name="hello_agent",
    description="A helpful assistant for user questions.",
    instruction="Your job is to creatively say hello",
)

bye_agent = LlmAgent(
    model="gemini-2.5-flash",
    name="bye_agent",
    description="A helpful assistant to say bye.",
    instruction="Your job is to creatively reply to the greeting and reply with a quip",
)

root_agent = SequentialAgent(
    name="root_agent",
    description="An agent that says hello and then says goodbye.",
    sub_agents=[hello_agent, bye_agent],
)
```

You get this span tree, which is noticeably missing any info about the kind of workflow (parallel, sequential, loop).

image

I'm trying to get someone from ADK to look at this PR, but maybe we can do one of two things:
1. Create an extra workflow span under the workflow agent. A little unnecessary in terms of timings
2. Make the workflow span attributes available on the invoke_agent, or say they are "compatible" similar to [what we did for MCP tool calls](https://github.com/open-telemetry/semantic-conventions/blob/2011b917ba9bf28e062d3c9656cc7368f0079f17/docs/gen-ai/mcp.md?plain=1#L145).

wrisa: @aabmass
1) would it be fair to ask, ADK instrumentation to adapt to this new operation `invoke_workflow` and not produce span with `invoke_agent` for workflow agents like `SequentialAgent/ParallelAgent/LoopAgent`. Maybe because `invoke_workflow` operation name did not exist earlier and so adk is using `invoke_agent` ?
Also if you see the tree above, the root agent is not capturing input messages from the user and output messages after finishing the workflow which can be captured for span with operation name 'invoke_workflow'
2) `gen_ai.workflow.type` attribute can be added to record sequential/parallel/loop.

jawoszek: My 2 cents from ADK perspective:

>would it be fair to ask, ADK instrumentation to adapt to this new operation invoke_workflow and not produce span with invoke_agent for workflow agents like SequentialAgent/ParallelAgent/LoopAgent. Maybe because invoke_workflow operation name did not exist earlier and so adk is using invoke_agent ?

I'd be a little hesitant to do that because in ADK those workflow agents are still modeled as agents. So having different span types for different types of agents could be confusing.

>gen_ai.workflow.type attribute can be added to record sequential/parallel/loop.

This makes sense as an optional attribute to be used only in case of workflow-like agents.

wrisa: > in ADK those workflow agents are still modeled as agents

@jawoszek thanks for looking into this. Do workflow agents use LLM internally ? If they are then it makes sense. But if they are not using LLM would there still be any issue to have workflow span for them ?

aabmass:
@wrisa is this fixing https://github.com/open-telemetry/semantic-conventions-genai/issues/55?

wrisa:
> open-telemetry/semantic-conventions-genai#55

@aabmass yes, partly. I will update description and changelog. Thank you!

lmolkova:
Summarizing the story:

##  LangGraph

StateGraph operations would be 'workflows`. E.g.

```python 
parallel_builder  = StateGraph(State)
...
parallel_workflow = parallel_builder.compile()
state = parallel_workflow.invoke({"topic": "cats"})   <--- this would be workflow span, no workflow name available?
```
alternatively it could be

```python
@entrypoint()
def parallel_workflow(topic: str):  <--- this would be workflow span, workflow name would be the function name - `parallel_workflow`
    joke_fut = call_llm_1(topic)
   ...
    return aggregator(
        topic, joke_fut.result(), ...    ).result()
```

[examples](https://docs.langchain.com/oss/python/langgraph/workflows-agents)

## OpenAI 

Terminology like higher level traces OpenAI uses is not correct or applicable to OTel world, but what's important they use `workflow` in their docs.

```python
    agent = Agent(name="Joke generator", instructions="Tell funny jokes.")

    with trace("Joke workflow"): #    <----- this would be workflow span, with "Joke workflow" as a name
        first_result = await Runner.run(agent, "Tell me a joke")
        second_result = await Runner.run(agent, f"Rate this joke: {first_result.final_output}")
````

> [!WARNING]
>
> This [`trace`](https://openai.github.io/openai-agents-python/ref/tracing/#agents.tracing.trace) thing could mean absolutely anything, users might use it to track calls to their database (by mistake)
> I would not rely on it meaning workflow - an operation that wraps calls to (presumably) multiple agents

[examples](https://openai.github.io/openai-agents-python/tracing/#higher-level-traces)

### Agent wrapping agents

```python
orchestrator_agent = Agent(
    name="orchestrator_agent",
    instructions=....,
    tools=[
        spanish_agent.as_tool(tool_name="translate_to_spanish", ...),
        french_agent.as_tool(tool_name="translate_to_french", ...),
        italian_agent.as_tool(tool_name="translate_to_italian", ...),
    ],
)
...
await Runner.run(orchestrator_agent, msg)  # <---- this one might or might not know whether it's doing orchestration
```

## ADK

```python
from google.adk.agents import LlmAgent, SequentialAgent

hello_agent = LlmAgent(model="gemini-2.5-flash",  name="hello_agent",  instruction="Your job is to creatively say hello")
bye_agent = LlmAgent(model="gemini-2.5-flash", name="bye_agent",  instruction="Your job is to creatively reply to the greeting and reply with a quip")
root_agent = SequentialAgent(name="root_agent", sub_agents=[hello_agent, bye_agent])  # <--- this could be a workflow span, it knows it wraps other agents, workflow name could be `root_agent` or `SequentialAgent` if name is optional
```
[examples](https://google.github.io/adk-docs/agents/workflow-agents/sequential-agents)

## CreaAI

also use2 workflow terminology in the docs

```python
 crew = Crew(
            agents=self.agents,  # Automatically collected by the @agent decorator
            tasks=self.tasks,    # Automatically collected by the @task decorator.
            process=Process.sequential,
            verbose=True,
        )
crew.kickoff(inputs={"any": "input here"}) #    <----- this would be workflow span, no workflow name
````
[examples](https://docs.crewai.com/en/concepts/crews)

## MS agent framework

```python
    workflow = GroupChatBuilder(
        participants=[researcher, writer],
        intermediate_outputs=True,
        orchestrator_agent=AzureOpenAIResponsesClient(...).as_agent(
            name="Orchestrator",
            instructions="You coordinate a team conversation to solve the user's task.",
        ),
    ).build()
    workflow_agent = workflow.as_agent(name="GroupChatWorkflowAgent")
    agent_result = await workflow_agent.run(task)  #  <----- this probably should be the workflow span? Or the orchestrator? Or both?
```

[examples](https://github.com/microsoft/agent-framework/blob/main/python/samples/03-workflows)

## Agno

```python
article_workflow = Workflow(
    name="Enhanced Article Creation Workflow",
    description="Automated article creation with conditional parallel research",
    steps=[article_creation_sequence],
)
```

[full example](https://github.com/agno-agi/agno/blob/main/cookbook/04_workflows/01_basic_workflows/01_sequence_of_steps/workflow_using_steps_nested.py)

Let me write down more thoughts in a separate comment

lmolkova:
It seems we have two types of design: 

1. Workflow as a separate API (CrewAI, LangGraph, Agno, MS agent framework, OpenAI with @\trace annotation). Note: I didn't check if they also support calling agents from agents via tools)
2. Workflow as a specialized agent that is a composition of one or more agents (ADK, OpenAI Agents)

------

And there are several ways to address is:

1. Workflow is a separate operation name (not an invoke_agent)
    - Pros: we can have different metrics for workflows and individual agents
    - Cons: 
      - not clear what to do with workflows-that-are-also-agents. 
      - We can't really demand one operation name for everyone (crew ai would probably prefer `kickoff` - it'd be more user friendly)
2. Workflow is the same span as invoke_agent, `gen_ai.workflow.name` attribute means it's a workflow. We'll probably need a more reliable criteria.
    - Pros: can model agents and workflows
    - Cons: same workflow and agent metrics, might be problematic if latency is tremendously different
      We could even report 

It seems there is no perfect solution. I'm going to sleep on this and let's discuss it on the call tomorrow.

model/gen-ai/spans.yaml
aabmass: Discussed this offline, I'm going to close this out as I think we will keep it as agent invocation in the case of ADK.



> Do workflow agents use LLM internally ? If they are then it makes sense. But if they are not using LLM would there still be any issue to have workflow span for them ?

I think the answer is no

model/gen-ai/spans.yaml
```diff
@@ -468,3 +468,40 @@ groups:
...
+      eg: Some frameworks like ADK have workflow agents that orchestrate other agents
```
aabmass, line 495: Discussed this offline. I think we agreed that "invoke agent" is like a specialization of the proposed "workflow" span here.
- Workflow encapsulates some operation that takes in a user input and returns output to user, and probably applies some control flow.
- Invoke agent combines that with tools, a model, system instructions or some kind of persona, etc.

aabmass: Proposal: we can add update invoke agent span definition like "invoke agent spans are compatible/subtype of GenAI workflow spans, which add ..." and update the YAML to reflect that relationship. Allow instrumentations to choose between them based on how much information they have or how they model it.

---

The only thing I'm still unsure of is why we need separate `gen_ai.agent.name` and `gen_ai.workflow.name`.

WDYT @lmolkova @wrisa?

aabmass (approved):
LGTM but still need to figure out relationship to agent spans imo.

lmolkova (approved):

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -448,6 +451,119 @@ and SHOULD be provided **at span creation time** (if provided at all):
...
+### Invoke workflow span
+
+<!-- semconv span.gen_ai.invoke_workflow.internal -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+Represents a workflow that orchestrates one or more agents. Also a workflow can be predetermined static sequence of operations eg: LLM calls, tool calls, and retrievals without a specific agent orchestrating them. A workflow groups multiple operations together, accepting an input and producing a final output.
```
TaoChenOSU: The definition here sounds like an agent with LLM calls, tool calls, etc. Should the definition be strictly "more than one agents"?

model/gen-ai/spans.yaml
```diff
@@ -476,3 +476,43 @@ groups:
...
           Anthropic reports this separately from `input_tokens`.
           This value MUST be added to the Anthropic `input_tokens` to compute `gen_ai.usage.input_tokens`.
+
+  - id: span.gen_ai.invoke_workflow.internal
+    type: span
+    span_kind: internal
+    stability: development
+    brief: >
+      Represents a workflow that orchestrates one or more agents.
+      Also a workflow can be predetermined static sequence of operations eg: LLM calls, tool calls, and retrievals
```
TaoChenOSU: Here too

model/gen-ai/spans.yaml
```diff
@@ -476,3 +476,43 @@ groups:
         note: >
           Anthropic reports this separately from `input_tokens`.
           This value MUST be added to the Anthropic `input_tokens` to compute `gen_ai.usage.input_tokens`.
+
+  - id: span.gen_ai.invoke_workflow.internal
+    type: span
+    span_kind: internal
+    stability: development
+    brief: >
+      Represents a workflow that orchestrates one or more agents.
```
lmolkova: SIG discussion:

Cosmetic suggestion to make it more clear

```suggestion
      Represents a workflow that orchestrates agents.
```

It seems we need to explain what invoke_workflow is (rather than what it's not  - invoke_agent)

@TaoChenOSU volunteered to suggest something, thanks!

TaoChenOSU: How about: "Represents a workflow that orchestrates GenAI agents or defines a sequence of GenAI operations, including LLM calls, tool invocations, and retrieval steps. An operation represented by `invoke_agent` is a specialized form of a workflow. Use `invoke_agent` whenever possible, including when the underlying operations are encapsulated and exposed as a single agent. Use `invoke_workflow` when the encapsulated operations involve one or more agents and are exposed as a separate API."?

P.s. We probably shouldn't constraint workflows to be single output. There are use cases where a workflow of multiple agents will produce more than a single output.

wrisa: Here is another one -
Represents a workflow that could be predetermined static sequence of operations eg: Agent, LLM, tool, and retrieval invocations. A workflow groups multiple operations together, accepting input(s) and producing final output(s). It is useful for monitoring a specific execution path, which is explicitly supported by some AI agentic frameworks e.g. Langchain, CrewAI or explicitly defined by an application developer. `invoke_agent` should be used to group around a particular functional domain grouping and `invoke_workflow` may be used for particular execution path grouping.
cc @zhirafovod @adityamehra

lmolkova: WDYT about 

```suggestion
      Represents an operation that executes a coordinated process composed of multiple agents or other operations involving generative AI.
```

model/gen-ai/spans.yaml
```diff
@@ -476,3 +476,43 @@ groups:
...
+      Also a workflow can be predetermined static sequence of operations eg: LLM calls, tool calls, and retrievals
+      without a specific agent orchestrating them.
+      A workflow groups multiple operations together, accepting an input and producing a final output.
```
lmolkova: The brief is intended to be short. I think the suggestion I left in the comment above

> Represents an operation that executes a coordinated process composed of multiple agents or other operations involving generative AI.

covers all of these points.  If there is some aspect it does not cover, let's move it to the note below

model/gen-ai/spans.yaml
lmolkova: I left a suggestion in the https://github.com/open-telemetry/semantic-conventions/pull/3249#discussion_r2907479751, lmk if it works

docs/gen-ai/gen-ai-agent-spans.md
lmolkova: addressed in https://github.com/open-telemetry/semantic-conventions/pull/3249/commits/04f5384ce207b5dde9f9eaf594d14eef9915ec14

PR #3256, opened 2026-01-07
Update exception recording guidelines to not use span events

pellared:
Together with https://github.com/open-telemetry/opentelemetry-specification/pull/4825

Towards OTEP: [Span Event API deprecation plan](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4430-span-event-api-deprecation-plan.md)

Towards https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8386 (merged "prototype")

## Changes

- Remove the recommendation to use `Span.RecordException` for error recording
- Update the recording exception example to record logs and metrics.
- Deprecate "Semantic conventions for exceptions on spans" document in favor of logs for recording exceptions

### DIFF
--- a/docs/general/recording-errors.md
+++ b/docs/general/recording-errors.md
@@ -122,7 +122,8 @@ public boolean createIfNotExists(String resourceId) throws IOException {
 
     return false;
   } catch (IOException e) {
-    // this exception is expected to be handled by the caller and could be a transient error
+    // this exception is expected to be handled by the caller
+    // and could be a transient error
     logger.withEventName("acme.resource.create.error")
       .withException(e)
       .warn();


docs/general/recording-errors.md
```diff
@@ -83,14 +83,16 @@ include it if the operation succeeded.
 
 ## Recording exceptions
 
-When an instrumented operation fails with an exception, instrumentation SHOULD record
-this exception as a [span event](/docs/exceptions/exceptions-spans.md) or a [log record](/docs/exceptions/exceptions-logs.md).
+When an instrumented operation throws an exception, instrumentation SHOULD
+record this exception as a [log record](/docs/exceptions/exceptions-logs.md).
```
pellared:
```suggestion
record this exception as an [event](/docs/exceptions/exceptions-logs.md).
```

.github/copilot-instructions.md
```diff
@@ -96,7 +96,7 @@ these requirements:
 
 - **Spans**: Set status to Error, populate `error.type`, set description when helpful
 - **Metrics**: Include `error.type` attribute for filtering and analysis
-- **Exceptions**: Record as span events or log records using SDK APIs
+- **Exceptions**: Record as log records
```
pellared, line 99:
```suggestion
- **Exceptions**: Record as event records
```

model/exceptions/deprecated/events-deprecated.yaml
```diff
@@ -2,6 +2,9 @@ groups:
   - id: event.exception
     name: exception
     stability: stable
+    deprecated:
+      reason: obsoleted
+      note: Use logs to record exceptions instead.
```
lmolkova: It's complicated (not your PR) :)

events in semconv are log-based events. We don't have means to define span events.
So I think we don't need to obsolete the event in yaml (at least not until we decide its name should be different).

We can still deprecate the markdown doc

docs/exceptions/exceptions-spans.md
```diff
@@ -23,7 +23,7 @@ exceptions associated with spans.
 <!-- see templates/registry/markdown/snippet.md.j2 -->
 <!-- prettier-ignore-start -->
 
-**Status:** ![Stable](https://img.shields.io/badge/-stable-lightgreen)
+**Status:** ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Use logs to record exceptions instead.
```
lmolkova: check out my comment on the yaml. Maybe we can do this:
- deprecate this doc as a whole saying use log-based events
- move stacktrace representation that remains useful to the log-based events doc

docs/general/recording-errors.md
```diff
@@ -83,14 +83,16 @@ include it if the operation succeeded.
 
 ## Recording exceptions
 
-When an instrumented operation fails with an exception, instrumentation SHOULD record
-this exception as a [span event](/docs/exceptions/exceptions-spans.md) or a [log record](/docs/exceptions/exceptions-logs.md).
+When an instrumented operation throws an exception, instrumentation SHOULD
```
lmolkova: 👍

docs/general/recording-errors.md
```diff
@@ -83,14 +83,16 @@ include it if the operation succeeded.
...
+record this exception as an [event](/docs/exceptions/exceptions-logs.md).
```
lmolkova: so this is where things get blurry - the doc there does not define an event, it defines a bunch of attributes without defining event name or severity. 

I think we should clarify event-name and severity considerations there so we're providing comparable replacement.

docs/general/recording-errors.md
lmolkova: nit: "instrumented operation" makes sense for spans and partially to metrics, when it comes to logs it's less applicable

Maybe

```suggestion
When instrumented code throws an exception, instrumentation SHOULD
```

?

docs/general/recording-errors.md
```diff
@@ -100,22 +102,25 @@ to the caller should be recorded (or logged) once.
...
+    recordMetric("acme.resource.create.duration", System.nanoTime() - startTime);
     return true;
   } catch (ResourceAlreadyExistsException e) {
-    // not recording exception and not setting span status to error - exception is handled
-    // but we can set attributes that capture additional details
+    // not setting span status to error - as the exception is not an error
+    // but we still log and set attributes that capture additional details
+    logger.debug(e);
     span.setAttribute(AttributeKey.stringKey("acme.resource.create.status"), "already_exists");
+    recordMetric("acme.resource.create.duration", System.nanoTime() - startTime);
```
lmolkova, line 121: metric should include error type attribute

docs/general/recording-errors.md
```diff
@@ -100,22 +102,25 @@ to the caller should be recorded (or logged) once.
...
   try {
     create(resourceId);
+    recordMetric("acme.resource.create.duration", System.nanoTime() - startTime);
     return true;
   } catch (ResourceAlreadyExistsException e) {
-    // not recording exception and not setting span status to error - exception is handled
-    // but we can set attributes that capture additional details
+    // not setting span status to error - as the exception is not an error
+    // but we still log and set attributes that capture additional details
+    logger.debug(e);
```
lmolkova: since it's not a comment now, should we make it more realistic? with https://github.com/open-telemetry/opentelemetry-java/pull/7907 it might look like

```java
logger.logBuilder(DEBUG, "acme.resource.create.error")
  .setAttribute("acme.resource.create.status", "already_exists")
  .setException(e)
  .emit()
``` 

this also gives us chance to show and discuss what we feel about event name

docs/general/recording-errors.md
```diff
@@ -100,22 +102,25 @@ to the caller should be recorded (or logged) once.
 ```java
 public boolean createIfNotExists(String resourceId) throws IOException {
```
lmolkova, line 101: offtopic: this whole example is a great pitch for https://github.com/open-telemetry/oteps/pull/165

docs/general/recording-errors.md
pellared: > I think we should clarify event-name and severity considerations there so we're providing comparable replacement.

I agree. From the SIG meeting I understood that you could try making a PR to address it.

I changed back to `log record` https://github.com/open-telemetry/semantic-conventions/pull/3256/commits/f71de5ce4a629afc4467fded723a4bcc667855a6

docs/general/recording-errors.md
pellared: According to current metric semantics only if the operation failed. This exception is not seen as as an error (failure). Thus span status is also not set to error.

model/exceptions/deprecated/events-deprecated.yaml
pellared: > So I think we don't need to obsolete the event in yaml (at least not until we decide its name should be different).

I wanted to have it to change this line in the generated document https://github.com/open-telemetry/semantic-conventions/blob/514f732094fe5d792242d2d50581c07b6228333c/docs/exceptions/exceptions-spans.md?plain=1#L26

> We can still deprecate the markdown doc

PTAL a5272a5241ce19497f31935b1bcdc66039fa06ed

docs/general/recording-errors.md
```diff
@@ -100,22 +102,25 @@ to the caller should be recorded (or logged) once.
...
+    recordMetric("acme.resource.create.duration", System.nanoTime() - startTime);
     return false;
   } catch (IOException e) {
-    // recording exception here (assuming it was not recorded inside `create` method)
-    span.recordException(e);
-    // or
-    // logger.warn(e);
-
-    span.setAttribute(AttributeKey.stringKey("error.type"), e.getClass().getCanonicalName())
+    logger.error(e);
```
pellared: Given that the exception is rethrown I am not sure if this is not against

> It's NOT RECOMMENDED to record the same exception more than once.

Does it mean that the instrumentation library should record the exception at most once?

Or does it mean that the application developer has to record the exception?

I am thinking if maybe we should not redesign this example so that it does not rethrow the exception?

However, if we do not capture anything we do not have any guarantees that it would be eventually recorded.

Any thoughts? Do we need some clarification? Possibly better to be handled in a separate PR.

docs/general/recording-errors.md
pellared: https://github.com/open-telemetry/semantic-conventions/pull/3256/commits/e8b8f4fda63866b6d2c0015c1f06a6fa6406d8b0

docs/exceptions/exceptions-spans.md
pellared: Stacktrace is already there https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-logs.md

.github/copilot-instructions.md
```diff
@@ -96,7 +96,7 @@ these requirements:
 
 - **Spans**: Set status to Error, populate `error.type`, set description when helpful
 - **Metrics**: Include `error.type` attribute for filtering and analysis
-- **Exceptions**: Record as span events or log records using SDK APIs
+- **Exceptions**: Record as event records
```
pellared:
```suggestion
- **Exceptions**: Record as log records
```

docs/exceptions/exceptions-spans.md
```diff
@@ -4,7 +4,8 @@ linkTitle: Spans
 
 # Semantic conventions for exceptions on spans
 
-**Status**: [Stable][DocumentStatus]
+**Status**: [Deprecated](https://img.shields.io/badge/-deprecated-red)<br>
+See [Recording exceptions](../general/recording-errors.md#recording-exceptions)
```
trask: let's ref the stable log exception doc

pellared: Done https://github.com/open-telemetry/semantic-conventions/pull/3256/commits/f3bec71e6a291dcd969559fd1da5fd50b00aa789

docs/general/recording-errors.md
pellared: Discussed during the SIG.

model/exceptions/deprecated/events-deprecated.yaml
thompson-tomo: Not sure about this as shouldn't https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/ be using the definition especially as the ability to render attribute groups is removed in semconv v2.

Thinking about it, shouldn't we remove the auto generated block from the spans page, add the auto generated block to the logs page and on the spans page have text mentioning to follow the logs page.

docs/general/recording-errors.md
```diff
@@ -100,22 +102,38 @@ to the caller should be recorded (or logged) once.
...
+
     return false;
   } catch (IOException e) {
-    // recording exception here (assuming it was not recorded inside `create` method)
-    span.recordException(e);
-    // or
-    // logger.warn(e);
+    logger.withEventName("acme.resource.create.error")
+      .withException(e)
+      .error()
```
lmolkova:
```suggestion
      .warn()  // this exception is expected to be handled by the caller and could be a transient error
```

this is what #3311 is about

docs/general/recording-errors.md
```diff
@@ -100,22 +102,38 @@ to the caller should be recorded (or logged) once.
...
 
-    span.setAttribute(AttributeKey.stringKey("error.type"), e.getClass().getCanonicalName())
+    String errorType = e.getClass().getCanonicalName();
```
lmolkova, line 131: should we also record it on the log attributes? (we can address it in the follow up, not strictly part of this PR)

pellared: I agree with both. That we can do it but better as a separate PR (maybe even in https://github.com/open-telemetry/semantic-conventions/pull/3311).

model/exceptions/deprecated/events-deprecated.yaml
lmolkova: Let's tackle it in the https://github.com/open-telemetry/semantic-conventions/pull/3311 where we'll decide on event naming and this event has a future in semconv

docs/general/recording-errors.md
```diff
@@ -83,14 +83,16 @@ include it if the operation succeeded.
 
 ## Recording exceptions
 
-When an instrumented operation fails with an exception, instrumentation SHOULD record
-this exception as a [span event](/docs/exceptions/exceptions-spans.md) or a [log record](/docs/exceptions/exceptions-logs.md).
+When instrumented code throws an exception, instrumentation SHOULD
+record this exception as a [log record](/docs/exceptions/exceptions-logs.md).
```
trask: I think I like the previous "When an instrumented operation fails"

"When instrumented code throws an exception" sounds like a much larger scope

docs/general/recording-errors.md
lmolkova: Logs SIG discussion:

1.
    > When an instrumented operation throws an exception
    
    is a consensus

2. let's use SHOULD on spans - if span ends with an exception, instrumentation should also log exception
3. let's use MAY for other cases

docs/general/recording-errors.md
pellared: ef4f629565713a81300b909aafa45270963b0b5c

model/exceptions/deprecated/events-deprecated.yaml
pellared: https://github.com/open-telemetry/semantic-conventions/pull/3256/changes/f0d31ea12cd2e22e84efffd4571278ab28dd80a7

pellared: https://github.com/open-telemetry/semantic-conventions/pull/3256/changes/f0d31ea12cd2e22e84efffd4571278ab28dd80a7

pellared:
@lmolkova, @trask, PTAL
I tried to make it as minimal as possible to avoid more bike-shedding. I still think that there are things that has to be improved like when an exception is not an error. But it deserved separate PRs.

lmolkova (approved):

PR #3068, opened 2025-11-11
Add Pressure Stall Information (PSI) metrics (reopened #2996)

alpineQ:
Closes #2995

## Changes

This PR adds support for Linux Pressure Stall Information (PSI) metrics to the system semantic conventions.

PSI is a Linux kernel feature (available since kernel 4.20) that identifies and quantifies resource contention by measuring the time impact that CPU, memory, and I/O resource crunches have on workloads.

### New Metrics

- **`system.linux.psi.pressure`** (Gauge): Measures resource pressure as a percentage of time that tasks were stalled over a time window (10s, 60s, or 300s)
- **`system.linux.psi.total_time`** (Counter): Tracks the total cumulative stall time in microseconds since system boot

### New Attributes

- **`system.psi.resource`**: The resource type (`cpu`, `memory`, `io`)
- **`system.psi.stall_type`**: The stall severity (`some` for partial stalls, `full` for complete stalls where all non-idle tasks are blocked)
- **`system.psi.window`**: The time window for pressure calculation (`10s`, `60s`, `300s`)

### Use Cases

PSI metrics enable:
- Sizing workloads to hardware or provisioning hardware according to workload demand
- Detecting productivity losses caused by resource scarcity
- Dynamic system management (load shedding, job migration, strategic pausing)
- Maximizing hardware utilization without sacrificing workload health

### References

- [Linux kernel PSI documentation](https://docs.kernel.org/accounting/psi.html)
- [/proc/pressure/* interface](https://man7.org/linux/man-pages/man5/proc.5.html)

### Relevant issues and PRs
There are issues on this matter in:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42779
- https://github.com/open-telemetry/opentelemetry-go-contrib/issues/8082

And 2 PRs that I am proposing to address these issues:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43823
- https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8083

> [!IMPORTANT]
> Pull requests acceptance are subject to the triage process as described in [Issue and PR Triage Management](https://github.com/open-telemetry/semantic-conventions/blob/main/issue-management.md).
> PRs that do not follow the guidance above, may be automatically rejected and closed.

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
[...]

PR #3261, opened 2026-01-08
[chore] System Semconv OS-exclusive instrumentation philosophy

braydonk:
## Changes

This PR updates our OS-exclusive instrumentation naming philosophy/policy to reflect the group's current standing. This ruling hasn't been fully applied across our instrumentation yet but we are working on it as part of initial stabilization efforts.

### DIFF
--- a/docs/non-normative/groups/system/design-philosophy.md
+++ b/docs/non-normative/groups/system/design-philosophy.md
@@ -201,11 +201,11 @@ baseline level of knowledge of the concepts being instrumented**. The `brief`
 and `note` fields of metrics and attributes should be used to convey information
 that is crucial to understanding the instrumentation intention, i.e.:
 
-* Differences in the same piece of data when it is instrumented on different
+- Differences in the same piece of data when it is instrumented on different
   platforms
-* When we recommend calculations be done on particular data rather than
+- When we recommend calculations be done on particular data rather than
   surfacing direct values from existing tools
-* When particular names or enum values were chosen when there are common
+- When particular names or enum values were chosen when there are common
   alternate terms for the same concept
 
 For root metrics and attributes, we will strive to always have a `brief` field.
@@ -213,13 +213,13 @@ The `brief` field should explain what the metric/attribute is, and if the
 explanation of a value is simple (i.e. simply surfacing a value from a common
 source like `procfs`) then the explanation of what the value should be can go in
 the brief. If the value needs some calculation explanation and justification,
-the information should be moved to the `note` field.
-For enum values, it is often the case that the intention of these values is
-obvious given whatever `brief` was provided for the attribute as a whole. A
-brief can be included in a scenario where we have had to make some choice on the
-value that isn't immediately obvious; the most common scenario is when some
-terminology differs across platforms and we had to choose one term to represent
-all scenarios. In this case, the `brief` can be used to clarify our intention.
+the information should be moved to the `note` field. For enum values, it is
+often the case that the intention of these values is obvious given whatever
+`brief` was provided for the attribute as a whole. A brief can be included in a
+scenario where we have had to make some choice on the value that isn't
+immediately obvious; the most common scenario is when some terminology differs
+across platforms and we had to choose one term to represent all scenarios. In
+this case, the `brief` can be used to clarify our intention.
 
 In cases where information about a concept is required to describe our intention
 for instrumentation, all information must come with citations to authoritative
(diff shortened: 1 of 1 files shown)


docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -201,106 +201,96 @@ baseline level of knowledge of the concepts being instrumented**. The `brief`
...
+
+This is for the sake of hierarchical organization of names. We consider the root
+namespace of metrics to be crucial to the typical usage patterns we expect. A
+user who wants to discover all of their `memory` metrics should only need to
+search for `memory.*` to get all metrics related to `memory` in general. If the
+OS was the root namespace, they would need to search for `linux.memory.*`,
+`windows.memory.*`, etc. This is bad partially because of complicating queries,
+but it would also require a user to go out of their way to gather all of the
+platforms they need to search for just to get the whole picture on their
+`memory` metrics.
```
thompson-tomo:
```suggestion
`system memory` metrics.
```

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -201,106 +201,96 @@ baseline level of knowledge of the concepts being instrumented**. The `brief`
...
+metrics, even considering common stats like patterns of memory usage, where
+there are platform-exclusive pieces of information that are only valuable to
+those who specialize in that platform.
+
+With this in mind, we have made the following decision about the OS name in
+OS-exclusive instrumentation:
+
+### The OS name should be included in the metric in most cases
+
+The primary reason for this is for discoverability and disambiguation. Sometimes
```
thompson-tomo, line 248:
```suggestion
### When to include OS name

The name of a platform specific metric should by default include the OS name, 
with the primary reason for this is to enable discoverability and disambiguation. Sometimes
```

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -201,106 +201,96 @@ baseline level of knowledge of the concepts being instrumented**. The `brief`
...
+concern within their system that they are monitoring (processes, memory,
+filesystem, cpu, network, etc) and for a user with a specific concern to have a
+clear way to discover the instrumentation supporting their area of concern.
+Sometimes that area of concern is not just the absolute root of the namespace,
+but also a number of sub-components in the root. The `system` namespace in
+particular has many examples of this, `system.memory`, `system.network`, etc. If
+we named the instrumentation as `system.linux.memory`, then just as in the
+scenario above it becomes hard to discover all the available "system memory"
+metrics. As a result, the OS name goes after the area of concern, i.e.
+`system.memory.linux`.
```
thompson-tomo, line 294:
```suggestion
### Where to place OS name in metric name

Our expectation is that the OS name to appear after the area of
concern which is being monitored (processes, memory,
filesystem, cpu, network, etc).
This means that the OS name should never appear as the root namespace ie `linux.memory.*` and
often the area of concern will not even be the root namespace,
but rather a sub-namespace. The `system` root namespace in
particular has many examples of this, `system.memory.*`, `system.network.*`, etc.
As a result, the OS name goes after the area of concern, i.e.
`system.memory.linux.*` with non platform specific metrics being `system.memory.*`.

### Why we choose this

To provide a user with a specific concern, a
clear way to discover the instrumentation supporting their area of concern. If
we had named the instrumentation as `system.linux.memory.*`, then a user would need to search for `linux.memory.*`,
`windows.memory.*`, etc. This is bad partially because of complicating queries,
but it would also require a user to go out of their way to gather all of the
platforms they need to search for just to get the whole picture on their
`system memory` metrics.
```

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -201,106 +201,96 @@ baseline level of knowledge of the concepts being instrumented**. The `brief`
...
+This is for the sake of hierarchical organization of names. We consider the root
+namespace of metrics to be crucial to the typical usage patterns we expect. A
+user who wants to discover all of their `memory` metrics should only need to
+search for `memory.*` to get all metrics related to `memory` in general. If the
+OS was the root namespace, they would need to search for `linux.memory.*`,
+`windows.memory.*`, etc. This is bad partially because of complicating queries,
+but it would also require a user to go out of their way to gather all of the
+platforms they need to search for just to get the whole picture on their
+`memory` metrics.
+
```
thompson-tomo, line 281:
```suggestion
```
Has been consolidated into the below section.

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -201,106 +201,96 @@ baseline level of knowledge of the concepts being instrumented**. The `brief`
...
+### The OS name should be after the area of concern
```
thompson-tomo, line 282: Would be good to add details of what is an area of concern. For instance there is a discussion about `memory pressure`, in that case would the area of concern be `memory` or `memory.pressure`?

rogercoll: That's a great point. In this context, the 'area of concern' refers to the high-level resource or term shared across operating systems (like memory).

Since pressure (PSI) is a concept specific to Linux, we place the OS keyword after the shared resource to differentiate between common cross-platform metrics and OS-specific implementations. So, the hierarchy remains system.memory (shared) → linux (specific) → pressure (os unique metrics).

thompson-tomo: What happens when we consider that pressure applies to macos as well. Would the area of concern then be pressure?

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -201,106 +201,96 @@ baseline level of knowledge of the concepts being instrumented**. The `brief`
...
-from the OS, i.e. the direct value from `cat /proc/<pid>/cgroup`. With cgroups
-in particular, there is high likelihood that more specialized semconv
-instrumentation could be developed, particularly in support of more specialized
-container runtime or systemd instrumentation. It's more useful for a working
-group developing special instrumentation that leverages cgroups to be more
-prescriptive about how the cgroup information should be interpreted and broken
-down with more specificity.
-
-## Operating System in names
+## Naming Platform-Exclusive Metrics
```
thompson-tomo, line 230: Based on https://github.com/open-telemetry/semantic-conventions/pull/3068 it would be useful to have a section naming platform-exclusive attributes.

This section should cover when the attribute is added with the area ie system root namespace and when it is added with a root namespace = area of concern.

An example would be we currently have [`system.memory.state`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/system/#system-memory-state) but also have [`cpu.mode`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/cpu/#cpu-mode) both of which relate to area of concerns which applies to multiple to multiple areas.

My initial thought would belong to the area of concern unless they are describing the entity ie system, process producing the telemetry.

docs/non-normative/groups/system/design-philosophy.md
```diff
@@ -201,106 +201,96 @@ baseline level of knowledge of the concepts being instrumented**. The `brief`
...
+Monitoring operating systems is a long-standing practice, and there are numerous
+heavily differing approaches within different platforms. There are lots of
+metrics, even considering common stats like patterns of memory usage, where
+there are platform-exclusive pieces of information that are only valuable to
+those who specialize in that platform.
+
+With this in mind, we have made the following decision about the OS name in
+OS-exclusive instrumentation:
+
+### The OS name should be included in the metric in most cases
```
mx-psi, line 246:
```suggestion
### The OS name should be included in the metric name in most cases
```

We could even rephrase this as "Users should be able to tell if a metric is OS-specific based on the name", which I think is a fair summary of the text?

docs/non-normative/groups/system/design-philosophy.md
mx-psi: I don't think we are going to be able to create a 'hard' rule for this when it comes to what is the area of concern, if anything we can clarify that the area of concern should be made clear on each semconv area

PR #3229, opened 2025-12-20
Document mapping between gRPC (native) and OTel conventions

lmolkova:
Fixes #1773

Document mapping between gRPC native metrics and spans and OTel gRPC metrics and spans.

Depends on:
- [x] decide on message mapping - see https://github.com/open-telemetry/semantic-conventions/pull/3226#discussion_r2636675990
- [x] deprecate size metrics https://github.com/open-telemetry/semantic-conventions/pull/3267
- [x] deprecate rpc.message event #3266
- [x] add url.scheme to grpc #3317

PR #3267, opened 2026-01-10
RPC: deprecate `rpc.server|client.request|response.size` metrics

### LINKED ISSUE #3259: RPC: consider deprecating rpc.*.request.size
- need to be per attempt to show throughput
- need to record over-the-wire size

See https://github.com/open-telemetry/semantic-conventions/pull/3229 for the context

lmolkova:
Fixes #3259

Deprecate `rpc.server.request.size`, `rpc.server.response.size`, `rpc.client.request.size`,
and `rpc.client.response.size` metrics due to confusing definition and low adoption.

Came up in https://github.com/open-telemetry/semantic-conventions/pull/3229 when analyzing gRPC native metrics.

- these metrics have confusing names: they use `request|response` terminology, but apply to individual messages within a stream. E.g. Java instrumentation interprets them as per-call (and [records metrics with last-message-size](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/be30e0c903ded55d555ea2efaf591cd3eb62d130/instrumentation-api-incubator/src/main/java/io/opentelemetry/instrumentation/api/incubator/semconv/rpc/RpcClientMetrics.java#L103) while [Go instrumentation reports metric per-message](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/a75e7ef676fe59c0d63c7d179d0477bc15c7d6ea/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L310))
- it's not clear if these metrics should be recorded [on retries, when messages are resent](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) - if not, they no longer sum up to throughput and become even more confusing.
-  I could not find implementations of these metrics beyond Go and Java in OTel ecosystem.
- gRPC native metrics define `grpc.client.attempt.sent_total_compressed_message_size`,  `grpc.client.attempt.rcvd_total_compressed_message_size`, `grpc.server.call.sent_total_compressed_message_size`,  `grpc.server.call.rcvd_total_compressed_message_size` which are more specific. 

Retries and compression are framework/protocol specific and following T-Shaped signals philosophy, each framework can define its own metrics for message sizes when necessary. 

If there is enough commonality, new metrics can be added to OTel in the future, but we're not targeting them for initial RPC conventions stability.

### DIFF
--- a/.chloggen/rpc-deprecate-size-metrics.yaml
+++ b/.chloggen/rpc-deprecate-size-metrics.yaml
@@ -1,6 +1,6 @@
 change_type: deprecation
 component: rpc
 note: >
-  Deprecate `rpc.server.request.size`, `rpc.server.response.size`, `rpc.client.request.size`,
-  and `rpc.client.response.size` metrics due to confusing definition and low adoption.
+  Deprecate experimental `rpc.server.request.size`, `rpc.server.response.size`, `rpc.client.request.size`,
+  and `rpc.client.response.size` metrics due to ambiguous definitions and inconsistent implementation.
 issues: [3267]


docs/rpc/rpc-metrics.md
```diff
@@ -185,240 +181,6 @@ different processes could be listening on TCP port 12345 and UDP port 12345.
 <!-- END AUTOGENERATED TEXT -->
 <!-- endsemconv -->
 
-#### Metric: `rpc.server.request.size`
```
kamphaus, line 188: I noticed that there is no list of deprecated metrics.

Only in the registry do we list deprecated attributes.
However no such list exists for metrics/spans.

What is the reason for this?

.chloggen/rpc-deprecate-size-metrics.yaml
```diff
@@ -0,0 +1,6 @@
+change_type: deprecation
+component: rpc
+note: >
+  Deprecate `rpc.server.request.size`, `rpc.server.response.size`, `rpc.client.request.size`,
+  and `rpc.client.response.size` metrics due to confusing definition and low adoption.
```
matt-hensley: Could be helpful to include some light context from the PR description. Something like:

```suggestion
  Deprecate experimental `rpc.server.request.size`, `rpc.server.response.size`, `rpc.client.request.size`,
  and `rpc.client.response.size` metrics due to ambiguous definitions and inconsistent implementation.
```

docs/rpc/rpc-metrics.md
lmolkova: No explicit reason - it's just an existing practice (which can be improved).  We don't render them on the semconv docs to reduce noise and not show things that are not relevant anymore. People can still find metric definitions in the history (or yaml).

I think it's tracked here https://github.com/open-telemetry/semantic-conventions/issues/592 - we can and should generate metric registry similarly to attributes and entities and list/document deprecated metrics there.

PR #3268, opened 2026-01-11
Add service.owner entity

### LINKED ISSUE #3101: Proposal: Add service.owner entity
## Proposal: Formalize `service.owner` Entity

We have observed a common pattern where users utilize resource attributes to annotate services with ownership information, such as source code location and team communication channels. While `service.namespace` exists for logical grouping, it does not explicitly convey organizational ownership or contact details.

I propose formalizing this concept by introducing a `service.owner` entity. This would standardize how we define service ownership, enabling better automation for notifications, incident response, and diagnostics.

### Proposed Attributes

I suggest the following attributes for the `service.owner` entity:

*   `service.owner.name`: The name of the team or individual responsible for the service (e.g., "Checkout Team").
*   `service.owner.url`: The source control management URL or identifier (e.g., link to the GitHub repository).
*   `service.owner.contact` (or `service.owner.chat`?): The chat channel or contact point for the team (e.g., Slack channel URL or name).

This aligns with the original vision from the service and deployment semconv SIG. I'd love to hear thoughts on this structure and if there are other common ownership attributes we should consider.

jpkrohling:
## Changes

Implements the `service.owner` entity as proposed in #3101, adding semantic conventions for service ownership information.

This enables users to annotate services with organizational ownership details, supporting automation for notifications, incident response, and diagnostics.

### New Attributes

| Attribute | Type | Description |
|-----------|------|-------------|
| `service.owner.name` | string | Name of the team or individual responsible for the service |
| `service.owner.url` | string | URL to the service's source code repository or documentation |
| `service.owner.contact` | string | Contact point or communication channel (e.g., Slack, email) |

### New Entity

- **`service.owner`** - Identifies the team or individual responsible for a service
  - `service.owner.name` as identifying attribute (required)
  - `service.owner.url` and `service.owner.contact` as descriptive attributes

### Stability

All new attributes and the entity are introduced at `development` stability.

### Relationship to Existing Conventions

While `service.namespace` groups services logically, `service.owner` explicitly conveys organizational ownership and contact details - addressing a common pattern observed in production environments.

Closes #3101

### DIFF
--- a/model/service/entities.yaml
+++ b/model/service/entities.yaml
@@ -41,3 +41,21 @@ groups:
       - ref: service.namespace
         role: identifying
         requirement_level: required
+  - id: entity.service.owner
+    type: entity
+    name: service.owner
+    brief: >
+      Identifies the team or individual responsible for a service.
+    note: >
+      A `service.owner` provides organizational ownership information for a service,
+      enabling automation for notifications, incident response, and diagnostics.
+      While `service.namespace` groups services logically, `service.owner` explicitly
+      conveys who is responsible for the service and how to contact them.
+    stability: development
+    attributes:
+      - ref: service.owner.name
+        role: descriptive
+      - ref: service.owner.url
+        role: descriptive
+      - ref: service.owner.contact
+        role: descriptive
--- a/model/service/registry.yaml
+++ b/model/service/registry.yaml
@@ -69,3 +69,33 @@ groups:
           for that telemetry. This is typically the case for scraping receivers, as they know the target address and
           port.
         examples: ["627cc493-f310-47de-96bd-71410b7dec09"]
+      - id: service.owner.name
+        type: string
+        stability: development
+        brief: >
+          The name of the team or individual responsible for the service.
+        note: |
+          This attribute is used to identify the organizational owner of a service,
+          enabling automation for notifications, incident response, and diagnostics.
+          The value SHOULD be a human-readable name such as a team name.
+        examples: ["Checkout Team", "Platform Infrastructure", "john.doe"]
+      - id: service.owner.url
+        type: string
+        stability: development
+        brief: >
+          A URL or identifier pointing to the service's source code repository or documentation.
+        note: |
+          This attribute is typically used to link a service to its source control management system,
+          allowing developers and operators to quickly navigate to the relevant codebase.
+          The value SHOULD be a fully qualified URL when possible.
+        examples: ["https://github.com/example/checkout-service", "https://gitlab.example.com/platform/infra"]
+      - id: service.owner.contact
+        type: string
+        stability: development
+        brief: >
+          The contact point or communication channel for the team responsible for the service.
+        note: |
+          This attribute provides a way to reach the team or individual responsible for the service.
+          The value can be a chat channel URL, email address, or other contact information.
+          When using chat channels, the full URL is preferred for easy navigation.
+        examples: ["https://example.slack.com/channels/checkout-team", "#checkout-team", "checkout-team@example.com"]
--- a/.chloggen/service_owner_entity.yaml
+++ b/.chloggen/service_owner_entity.yaml
@@ -1,0 +1,24 @@
+# Use this changelog template to create an entry for release notes.
+#
+# If your change doesn't affect end users you should instead start
+# your pull request title with [chore] or use the "Skip Changelog" label.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: "enhancement"
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: "service"
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add service.owner entity with service.owner.name, service.owner.url, and service.owner.contact attributes"
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: [3101]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The new entity enables users to annotate services with organizational ownership information,
+  supporting automation for notifications, incident response, and diagnostics.
--- a/docs/registry/attributes/service.md
+++ b/docs/registry/attributes/service.md
@@ -14,6 +14,9 @@ A service instance.
 | <a id="service-instance-id" href="#service-instance-id">`service.instance.id`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The string ID of the service instance. [1] | `627cc493-f310-47de-96bd-71410b7dec09` |
 | <a id="service-name" href="#service-name">`service.name`</a> | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | string | Logical name of the service. [2] | `shoppingcart` |
 | <a id="service-namespace" href="#service-namespace">`service.namespace`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | A namespace for `service.name`. [3] | `Shop` |
+| <a id="service-owner-contact" href="#service-owner-contact">`service.owner.contact`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The contact point or communication channel for the team responsible for the service. [4] | `https://example.slack.com/channels/checkout-team`; `#checkout-team`; `checkout-team@example.com` |
+| <a id="service-owner-name" href="#service-owner-name">`service.owner.name`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | The name of the team or individual responsible for the service. [5] | `Checkout Team`; `Platform Infrastructure`; `john.doe` |
+| <a id="service-owner-url" href="#service-owner-url">`service.owner.url`</a> | ![Development](https://img.shields.io/badge/-development-blue) | string | A URL or identifier pointing to the service's source code repository or documentation. [6] | `https://github.com/example/checkout-service`; `https://gitlab.example.com/platform/infra` |
 | <a id="service-version" href="#service-version">`service.version`</a> | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | string | The version string of the service component. The format is not defined by these conventions. | `2.0.0`; `a01dbef8a` |
 
 **[1] `service.instance.id`:** MUST be unique for each instance of the same `service.namespace,service.name` pair (in other words
@@ -46,3 +49,15 @@ port.
 **[2] `service.name`:** MUST be the same for all instances of horizontally scaled services. If the value was not specified, SDKs MUST fallback to `unknown_service:` concatenated with [`process.executable.name`](process.md), e.g. `unknown_service:bash`. If `process.executable.name` is not available, the value MUST be set to `unknown_service`.
 
 **[3] `service.namespace`:** A string value having a meaning that helps to distinguish a group of services, for example the team name that owns a group of services. `service.name` is expected to be unique within the same namespace. If `service.namespace` is not specified in the Resource then `service.name` is expected to be unique for all services that have no explicit namespace defined (so the empty/unspecified namespace is simply one more valid namespace). Zero-length namespace string is assumed equal to unspecified namespace.
+
+**[4] `service.owner.contact`:** This attribute provides a way to reach the team or individual responsible for the service.
+The value can be a chat channel URL, email address, or other contact information.
+When using chat channels, the full URL is preferred for easy navigation.
+
+**[5] `service.owner.name`:** This attribute is used to identify the organizational owner of a service,
+enabling automation for notifications, incident response, and diagnostics.
+The value SHOULD be a human-readable name such as a team name.
+
+**[6] `service.owner.url`:** This attribute is typically used to link a service to its source control management system,
+allowing developers and operators to quickly navigate to the relevant codebase.
+The value SHOULD be a fully qualified URL when possible.
--- a/docs/registry/entities/README.md
+++ b/docs/registry/entities/README.md
@@ -90,6 +90,7 @@ Currently, the following namespaces exist:
 | | [service](service.md#service) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
 | | [service.instance](service.md#service-instance) | ![Development](https://img.shields.io/badge/-development-blue) |
 | | [service.namespace](service.md#service-namespace) | ![Development](https://img.shields.io/badge/-development-blue) |
+| | [service.owner](service.md#service-owner) | ![Development](https://img.shields.io/badge/-development-blue) |
 | Telemetry | | |
 | | [telemetry.distro](telemetry.md#telemetry-distro) | ![Development](https://img.shields.io/badge/-development-blue) |
 | | [telemetry.sdk](telemetry.md#telemetry-sdk) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) |
--- a/docs/registry/entities/service.md
+++ b/docs/registry/entities/service.md
@@ -76,3 +76,31 @@ port.
 | Identity | [`service.namespace`](/docs/registry/attributes/service.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | A namespace for `service.name`. [3] | `Shop` |
 
 **[3] `service.namespace`:** A string value having a meaning that helps to distinguish a group of services, for example the team name that owns a group of services. `service.name` is expected to be unique within the same namespace. If `service.namespace` is not specified in the Resource then `service.name` is expected to be unique for all services that have no explicit namespace defined (so the empty/unspecified namespace is simply one more valid namespace). Zero-length namespace string is assumed equal to unspecified namespace.
+
+## Service Owner
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+**type:** `service.owner`
+
+**Description:** Identifies the team or individual responsible for a service.
+
+**Attributes:**
+
+| Role | Key | Stability | [Requirement Level](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/) | Value Type | Description | Example Values |
+| --- | --- | --- | --- | --- | --- | --- |
+| Description | [`service.owner.contact`](/docs/registry/attributes/service.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | The contact point or communication channel for the team responsible for the service. [4] | `https://example.slack.com/channels/checkout-team`; `#checkout-team`; `checkout-team@example.com` |
+| Description | [`service.owner.name`](/docs/registry/attributes/service.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | The name of the team or individual responsible for the service. [5] | `Checkout Team`; `Platform Infrastructure`; `john.doe` |
+| Description | [`service.owner.url`](/docs/registry/attributes/service.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | A URL or identifier pointing to the service's source code repository or documentation. [6] | `https://github.com/example/checkout-service`; `https://gitlab.example.com/platform/infra` |
+
+**[4] `service.owner.contact`:** This attribute provides a way to reach the team or individual responsible for the service.
+The value can be a chat channel URL, email address, or other contact information.
+When using chat channels, the full URL is preferred for easy navigation.
+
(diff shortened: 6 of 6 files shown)


lmolkova:
@open-telemetry/semconv-service-and-deployment-approvers could you please review? thanks!

jpkrohling:
I see some conflicts here, and I'd love to get a first review before working on them.

jpkrohling:
Any news?

model/service/entities.yaml
```diff
@@ -41,3 +41,21 @@ groups:
...
+      A `service.owner` provides organizational ownership information for a service,
+      enabling automation for notifications, incident response, and diagnostics.
+      While `service.namespace` groups services logically, `service.owner` explicitly
+      conveys who is responsible for the service and how to contact them.
+    stability: development
+    attributes:
+      - ref: service.owner.name
+        role: descriptive
+      - ref: service.owner.url
+        role: descriptive
```
thompson-tomo, line 59: I see this as more being attribute of the `service` entity as it is unique per service.

model/service/entities.yaml
```diff
@@ -41,3 +41,21 @@ groups:
...
+      Identifies the team or individual responsible for a service.
+    note: >
+      A `service.owner` provides organizational ownership information for a service,
+      enabling automation for notifications, incident response, and diagnostics.
+      While `service.namespace` groups services logically, `service.owner` explicitly
+      conveys who is responsible for the service and how to contact them.
+    stability: development
+    attributes:
+      - ref: service.owner.name
+        role: descriptive
```
thompson-tomo, line 57: Should this be the identifying attribute? Or do we need to add an id?

model/service/entities.yaml
```diff
@@ -41,3 +41,21 @@ groups:
       - ref: service.namespace
         role: identifying
         requirement_level: required
+  - id: entity.service.owner
+    type: entity
+    name: service.owner
+    brief: >
+      Identifies the team or individual responsible for a service.
```
thompson-tomo, line 48: What isn't clear to me is if the owner is the one responsible for support or its the owner of the source code. I suspect it is support but we should make that clearer by updating name & description.

model/service/entities.yaml
```diff
@@ -41,3 +41,21 @@ groups:
       - ref: service.namespace
         role: identifying
         requirement_level: required
+  - id: entity.service.owner
+    type: entity
+    name: service.owner
```
thompson-tomo, line 46: How would we handle owner of hardware? Do we need to  mirror attributes into the hardware namespace? Or would it be better to rename this entity & corresponding attributes to `support.team` so it covers both use cases?

model/service/registry.yaml
```diff
@@ -69,3 +69,33 @@ groups:
...
+        type: string
+        stability: development
+        brief: >
+          The name of the team or individual responsible for the service.
+        note: |
+          This attribute is used to identify the organizational owner of a service,
+          enabling automation for notifications, incident response, and diagnostics.
+          The value SHOULD be a human-readable name such as a team name.
+        examples: ["Checkout Team", "Platform Infrastructure", "john.doe"]
+      - id: service.owner.url
```
thompson-tomo, line 82: based on moving attribute to the service entity

```suggestion
      - id: service.url.full
```

Otherwise

```suggestion
      - id: service.owner.url.full
```

jsuereth:
> @open-telemetry/semconv-service-and-deployment-approvers could you please review? thanks!

We did a review in the SIG two weeks back - there's a lot to discuss here.

Overall - understanding owner is big (i.e. we want this), but in the SIG we discussed understanding use cases for ownership, as some expressed that we may have a different notion of owner from each other.

The AI was for Semconv SIG to comment on this PR with use cases for owner to make sure we're modelling this correctly.

Notes: https://docs.google.com/document/d/1Fy6yXfZqrwN_oHw95Bdfg_0hzUgzlk3VO5wA1invgkI/edit?tab=t.0#heading=h.oer7fxi6z46u


From my perspective, I'm also curious where / how you plan to pull this information?  I.e. *how* would an OpenTelemetry "distribution" (e.g. someone using OTEL) pull this entity and report it on telemetry?  I'd love to see an example or prototype here.

While I realize that COULD just be manually annotating OTEL_RESOURCE_ATTRIBUTES, I think that's pretty high friction for this.  Would you see, e.g. a special hook in k8s resource detector to pull this information from labels, maybe an expansion on [kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels)?

model/service/entities.yaml
bgarcial: Opportune remark. In an ideal situation (under devops culture perspective) the team that writes and owns the source code should be the same team responsible for supporting it. Is good encouraging for them being the same actors; but we know silos are real in organizations. 
Do we want to assume or go here with the "you build it you run it" mindset?
If so, I see it as an an opportunity to intentionally not encourage to keep going with that distinction.

I am leaning for the above thinking, but just in case, if it is really necessary to make a differentiation here, the brief could include both roles or possibilities, something like:
>Identifies the team or individual responsible for developing, deploying and operating this service.

thompson-tomo: I want to avoid assumptions/ambiguity and in fact that proposed text doesn't make it clearer especially as we already have service.namespace which covers most of that.

> A string value having a meaning that helps to distinguish a group of services, for example the team name that owns a group of services. 

In reality what I think we should end up with is:
- `service.instance.support` attribute to capture who supports the instance
- `deployment.environment.owner` attribute to capture who owns the environment

These values can all be the same but could also be different but we avoid ambiguity.

I come from somewhere there is local first level support, regional second level & development do third level this is to ensure timely responses etc.

bgarcial: I understand that `service.namespace` is meant to provide ownership (at certain extent) by including the team name, but from what I have seen, there is a gap between what the attribute intends (under the semconv otel perspective) and how the namespace concept is actually used, including any additional metadata it provides. This is because product teams use a namespace for what K8s literally says it is for, which is logical grouping, and it is most often named as the product hosted there. Even when the team name is used as a prefix, that does not provide meaningful ownership information people can refer to, like a repo or communication channel. I recognize that some teams provide annotations with contact info in the namespace, but that is more the exception than the norm. Making it more explicit ensures that the contact details will be there.

Regarding `service.instance.support` + `deployment.environment.owner`, I am curious to hear more about this idea. I got your idea of providing some sort of structure (maybe a bit much?). What I am wondering is if having more than 1 ownership attribute could create some confusion, for example, during an incident, which attribute tells me who to page or escalate to?

thompson-tomo: I agree the usage and definition of service.namespace could be improved. I had attempted to introduce app.name to capture the product name but couldn't get enough people on board.

What I am not opposed to would be to also add service.owner if we adjust the description of service.namespace.

> What I am wondering is if having more than 1 ownership attribute could create some confusion, for example, during an incident, which attribute tells me who to page or escalate to?

I wouldn't think so as the `service.instance.support` is defined as who to contact about incidents. Even in a devops scenario it could be useful for having seperate attributes, for instance production incidents might go to 1 slack channel and staging might go to a different channel hence the split is useful.

bgarcial: >What I am not opposed to would be to also add service.owner if we adjust the description of service.namespace.

That could also be a good thing to bring to the table. I like it the way this ownership proposal here can help to refine the `service.namespace` convention! :) 
But how could we keep the discussion of `service.owner` forward without being dependent on `service.namespace` being adjusted first? I mean, the last one is already stable and adopted, which will imply a separate effort (and PR) to buy people in on the room (we are thinking/suggesting here), it has to improve. At the end, the goal is that both semconvs should complement each other and not block each other :P. 
If you think opening a conversation around `service.namespace` will help to refine this ownership initiative, what if we could track its description adjustment in a separate issue? Happy to help here if needed

bgarcial: >I wouldn't think so as the service.instance.support is defined as who to contact about incidents. Even in a devops scenario it could be useful for having seperate attributes, for instance production incidents might go to 1 slack channel and staging might go to a different channel hence the split is useful.

This is interesting, and got some questions? 
1. Am I understanding correctly that you are suggesting including `service.instance.support` attribute here alongside `service.owner.*`?
If so, we are talking about touching `service.instance` entity.  Would not that increase the scope of this PR proposal too much? (perhaps a similar situation with the idea of adjusting `service.namespace` description)

2. In the hypothetical situation this is introduced, in your opinion, what will be the value of per instance support this attribute would hold?
How I see it the info it could provide can be understood under routing perspective and not in delegating ownership to a separate group of people who did not build the service. What I mean by routing, is something like different channels of the same core team rather than entirely different teams.

(* I know at the end of the day, whatever value people set cannot be under strict control, it would be like a soft influence to encourage things, like `"The [recommended] value should/must be ..."`

Curious to hear your input.

thompson-tomo: > But how could we keep the discussion of service.owner forward without being dependent on service.namespace being adjusted first?

Hmm, the problem I forsee is without clarifying service.namespace there is ambiguity as to which attribute. This change could simply to update the note to

> A string value having a meaning that helps to distinguish a group of services. It is recommended to use the product/solution name but for historically reasons the team name that owns the service may be used.

In relation to your questions:

> Am I understanding correctly that you are suggesting including service.instance.support attribute here alongside service.owner.*?

They can be added via seperate PR, but would have both open & linked so that the justification of limiting owner to describing who owns source code and is not for support use cases.

> what will be the value of per instance support this attribute would hold?
> How I see it the info it could provide can be understood under routing perspective and not in delegating ownership to a separate group of people who did not build the service. What I mean by routing, is something like different channels of the same core team rather than entirely different teams.

Yes routing is the key thing and it nicely supports both approaches ie 1 team or dedicated teams as we know both are used in the field. Hence enables otel to avoid taking an opionated approach on how a business needs to set up their teams.

model/service/entities.yaml
jpkrohling: I don't think this would be identifying, I see this as metadata about the process only.

model/service/entities.yaml
jpkrohling: I took some time to read and digest this thread, and I'm still in favor of `service.owner`. The reason is that, as someone seeing this telemetry at 2am in the morning, I might need an easy way to find out who to ping next. Who owns this code/service? I realize that this doesn't fit _all_ cases, but I think it's a wide enough case to start as alpha and iterate based on real-world feedback. In fact, this is the pattern we've seen across three users by now at OllyGarden, and I bet there are way more out there with similar patterns.

model/service/entities.yaml
jpkrohling: I would probably leave that our for now, hardware can be a follow-up PR.

jpkrohling:
> From my perspective, I'm also curious where / how you plan to pull this information? I.e. how would an OpenTelemetry "distribution" (e.g. someone using OTEL) pull this entity and report it on telemetry? I'd love to see an example or prototype here.

In my view, this is highly relevant for exploratory cases, where platform engineers or operations folks are exploring data during an incident and need quick access to the team or source code that drives a specific service. This could lead to an internal IDP page with richer information about the team members, who's on call, and so on.

I could also see that tools, like incident management, AI assistants, and coding agents tools can use that information for better routing the alerts/issues. I don't think this would be the biggest usage of this information though.

> Would you see, e.g. a special hook in k8s resource detector to pull this information from labels, maybe an expansion on [kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels)?

Potentially, yes. The pattern I'm seeing though is observability teams instructing teams to add this to their SDK config (or otelconfig on my own case), and use governance tools (weaver in the future?) to enforce the existence of this attribute before an incident happens.

This is all perhaps to say that I believe most of the times, this information will come manually either added by the service owners or observability engineering teams.

model/service/entities.yaml
bgarcial: I am in with getting the focus on `service.owner` too, and starting it as alpha is a good baseline. Foremost, because I can relate to that pain of not knowing (at first glance) which external entity is flooding our backends with data, and having to do extra steps like looking for the `service.name` through a global org repositories search results, navigating several repos that mention it, until finding what I think is the correct one, and only then contacting the owners. Implementing this can save time, and most of the time, the actual fix from the owning team was applied quickly, often in less time than it took me to find who they were in the first place.
To recap, these would be my 5 cents:
-  I believe it is clear that `service.namespace` is serving a different purpose, so what would be the right thing to do here: do we want to open at the same time a pr for adjusting its description and link it to this? Or do we want to move forward with `service.owner` regardless of `service.namespace` clarification? I am leaning to the latter.

- In addition, I would leave `service.instance.support` out of this scope (as the same with hardware ownership). The reason is that adding a new attribute in a separate entity can deviate or extend the discussion. We can explore that in a follow-up PR, and that even might depend on the feedback received from `service.owner`. 

Any thoughts?

bgarcial:
I just noticed the discussion on the ownership topic on the SIG group document and what @jsuereth refers to. (I should have caught up on this earlier, apologies) 
It is very clear the different perspectives about how ownership should be addressed (including @thompson-tomo thoughts) and the ideas about how the proposed attribute values here should be populated and to which specific use cases give this priority (being incident response that one I can relate to). I am posting here to say that I will be following the SIG meetings to learn about the ongoing discussions and the consensus that needs to be reached in the community. :)

model/service/entities.yaml
thompson-tomo: But because we have modeled it as an entity each entity should be designed to have an id as that is a requirement for stability.

model/service/entities.yaml
thompson-tomo: It seems that we are talking about a few different use cases and trying to treat them as the same when they can differ.

> The reason is that, as someone seeing this telemetry at 2am in the morning, I might need an easy way to find out who to ping next.

> I can relate to that pain of not knowing (at first glance) which external entity is flooding our backends with data, and having to do extra steps like looking for the service.name through a global org repositories search results, navigating several repos that mention it, until finding what I think is the correct one, and only then contacting the owners.

I agree that the second use case would be addressed by `service.owner.*` and have no issue with adding it to

> enables users to annotate services with organizational ownership details

Especially if we simultaneously remove the ambiguity in service.namespace. 

We should however make it clear in the description of service.owner that it is to capture details of who owns the development of the software service generating the telemetry. This mirrors the discussion from the sig where the class it as the developer_owners.

We can then have a seperate PR to address the first use case of who to contact for support as even in devops, teams can often have different channels for support vs development and in some cases it will differ based on the instance. This matches operator_owners owner type from the sig doc.

pirgeo:
A few considerations:
- Should we be worried about personal information in these attributes (e.g., the name of a responsible person, or even a phone number or email address)? Many companies want to or have to avoid personal data in telemetry for privacy, legal, and compliance reasons.
- Who is responsible for updating the information as reality changes? Services are moved between teams, teams are renamed, and people leave the company. If these attributes are set close to the code or SDK, they will drift unless someone actively maintains them.
- Getting the data from a resource detector or other automation helps, but then we're implicitly assuming the existence of a separate, central source of truth (service catalog, IDP, on-call system) that needs to be kept accurate anyway. In that case, the OTel attributes are not the source of truth; they are just a projection.

I’ve read through the comments here and on #3101, and my understanding is that the main use case is essentially: “Who do I ping at 2am if something looks off?”. The answer is usually very specific to each organization’s incident process and structure (e.g. how on-call is handled). A close second is "Who do I assign to the ticket that is created from this?". Again, this is highly dependent on the organization. 

My biggest concern is how the “service owner” evolves over time and context. The correct person or team to contact may depend on
- Time of day/region (follow-the-sun on-call),
- Environment (prod vs staging vs test),
-  Support tier (L1 vs L2 vs L3 vs dev team).

A static `service.owner.name` feels too inflexible to model that use case. To me, it seems likely to become a snapshot that is quickly outdated.

Because of that, I think of `service.owner` more as a pointer to the source of truth, rather than the truth itself. For example, a `service.owner.url` or `service.owner.ref` pointing to an internal page (IDP, service catalog, runbook) seems extremely valuable. That page may contain: On-call contacts and escalation paths, debugging steps, known issues, links to the repo and backlog, etc. Crucially, that page can be kept up to date centrally when services or people move, without requiring any changes to the telemetry producers.

For a more machine-friendly variant, I could also imagine something like `service.owner.id`: A stable ID that is mapped internally to “whatever we currently consider the owner/support group for this service, given time, environment, etc.” That ID can be resolved by internal tooling to a service page, team, or on-call rotation. It also avoids putting PII directly into telemetry.

I think the service owner concept is a good idea, but it obviously has many interpretations. We need to be very careful how we define it here, so it has legs in the field. Interested to hear your thoughts!

jpkrohling:
> Should we be worried about personal information in these attributes

Personally Identifiable Information (PII) does NOT belong to telemetry. Not this one, not any one. We should have very strong language in the repo at a top-level if we don't have one already. 

> A static service.owner.name feels too inflexible to model that use case.

I'd say that we should only provide the metadata mechanisms for other systems to build on top of. Alert routing can use this as an input, for instance, and use the other dimensions (time of day?) for further refinement/decision.

jpkrohling:
Folks, I need some guidance here: is this something that is still under consideration to be pursued? If so, I'll address the suggestions, but if it's not under consideration, I'd rather invest time in other items.

pirgeo:
Personally, I think adding service owner information is worthwhile. However, I am not sure if the current proposal is very flexible. I'd love to hear from others who have used something like this in practice and see if there are additional use cases. In my mind, it would be a shame to introduce this if it turns out not to be really helpful.

jpkrohling:
>  I'd love to hear from others who have used something like this in practice

I agree, I'd love to hear from more people as well. This is a pattern we have seen in two companies using our product, and one other team said that they were having internal discussions on this exact matter, so having a semconv providing guidance would help them move forward.

ltflb-bgdi:
I am looking for a equivalent to the [elastic ECS organization field](https://www.elastic.co/docs/reference/ecs/ecs-organization). For us, `service.owner.name` would already be enough. Other fields like `service.owner.id`, `service.owner.url` etc. would be nice but not necessary, but our use case is quite simple. 

First we thought about using the `service.namespace` field, but as far as I get from the discussion, the purpose of this field is logical grouping, which would be a good replacement for our custom field name used to group services in what we call systems.

As mentioned above, our use-case is very simple compared to others. Still, a combination of `service.name`, `deployment.environment`, `service.namespace` and `service.owner.name` provides a lot of helpful information for filtering and routing without adding too much complexity.

A `service.owner` field may not be perfect for anybody, but it would help to clarify the meaning of `service.namespace` and save it from misuse due to absence of other fields.

bgarcial:
Similar situation here about @ltflb-bgdi describes, plus attributes like `service.owner.contact` and `service.owner.url` would be of great help.

jpkrohling:
@jsuereth , I need your assistance here: should I keep pursuing this?

jsuereth:
@jpkrohling I think we want to purse this.  It's an important thing to figure out how to communicate.  I think we need to work through the needs and requirements we're hearing from the community and determine a path forward - this kind of feedback is *standard* for a critical component in OpenTelemetry, and the disagreement on path forward, in my mind, an indicator of both the *need* to solve the problem as well as giving you an idea of how everyone has diverged in their treatment today.

From my perspective, I do have some concerns we should think about:

1. Should we be attaching owner to *every metric, log and trace* or should we just have the ability to report the association via Entities?
2. As discussed in Service + Deployment SIG - we need the ability to differentiate "Developer  Owner" vs. "Production Owner" vs. "Cost Center".  I.e. the word "owner" is somewhat nebulous, but your mission statement is NOT.  

Let's align on the mission - determining where notifications / incident response should be funneled.  I think if we limit "owner" to be more narrow - and leverage other attributes for other types of ownership, we can make progress here.

@jpkrohling If you'd like you could also hand this off to the Service + Deployment SIG.  It's possible someone there may be able to take what you have and refine the proposal to get alignment around the path forward here.

jpkrohling:
> If you'd like you could also hand this off to the Service + Deployment SIG. It's possible someone there may be able to take what you have and refine the proposal to get alignment around the path forward here.

I think this would be the right way to move forward. Thank you!

bgarcial:
Bumping this, so the `stale` label is (hopefully) removed from the PR

kenfinnigan:
Apologies, coming late to this one.

The one concern I'd have with embedding owner information, whether it be developer, production, or cost, is that information gets stale. Services move between teams, ownership shifts from development to SRE, many shifts can occur.

Is this information better suited to a service catalog, as opposed to telemetry of a service?

jsuereth:
For folks curious on the state of this PR - we're still actively working on this in the Service + Deployment SIG.

latest discussion is on a design document - https://docs.google.com/document/d/1aoanapxR6LMDOkLJYoYUgGyv3CIS9Ei9ERzcQYR6BvU/edit?tab=t.0 which can also be found in the SIG meetings notes.

I expect this PR or one similar will be revived as we hash out concerns raised on this PR.

PR #3282, opened 2026-01-14
Add note about preserving case to database stable semconv migration doc

trask:
Noticed this while updating Java database instrumentations.

### DIFF
--- a/docs/non-normative/db-migration.md
+++ b/docs/non-normative/db-migration.md
@@ -43,7 +43,7 @@ to
 
 ### Database client span attributes
 
-<!-- prettier-ignore-start -->
+<!-- markdownlint-disable-file MD060 -->
 | Change                                              | Comments                                                                                                    |
 |-----------------------------------------------------|-------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


docs/non-normative/db-migration.md
```diff
@@ -44,31 +44,31 @@ to
 ### Database client span attributes
 
 <!-- prettier-ignore-start -->
```
lmolkova:
```suggestion
<!-- markdownlint-disable-file MD060 -->
```

lmolkova (approved):

docs/non-normative/db-migration.md
trask: Is this preferred when the table is really wide? (makes sense to me)

kamphaus:
Is there a typo in the PR title?
I guess it should be `preserving` and not `perserving` or `persevering`.

docs/non-normative/db-migration.md
lmolkova: yeah, I think when table is too wide, it's hard to look at it in text mode anyway and aligning is not worth the effort (even when automated). personal preference, not a strong opinion

PR #3224, opened 2025-12-16
Add RPC migration doc

trask:
Resolves #3204

Marking as ready to start reviewing but will wait to merge until after #3223 is merged and that change incorporated.

PR #3283, opened 2026-01-14
Deprecate `rpc.message` event

### LINKED ISSUE #2865: Deprecate rpc.message.id/type attribute
### Area(s)

area:rpc

### What's missing?

The following attributes (`rpc.message.id`/`rpc.message.type`) provide limited if any added value or usefulness to sem conv so they should be deprecated.

The below attributes are deprecated with the  mentioned reasons:

- `rpc.message.id` -> obsoleted
- `rpc.message.type` -> use `network.io.direction` instead

### Tip

### LINKED ISSUE #3266: RPC: deprecate `rpc.message` events
This is a follow-up on https://github.com/open-telemetry/semantic-conventions/pull/3226#discussion_r2636675990

Today, RPC instrumentations emit span events for each message sent and received.
These messages allow us to see timestamps within a stream, but they don’t provide meaningful correlation between requests and responses.

With the deprecation of span events (and the lack of ability in semconv to define span events), this is no longer a viable long-term approach.

The current usefulness and correctness of message events is questionable:

* The only value they provide is timestamps and the count of messages during the call.
* Client-side: in case of retries, messages can be resent, and these should be attached to a per-try span (we don't define it!).
* Compressed vs. uncompressed size attributes need revision. Depending on the instrumentation point, you might only know the over-the-wire size; it might be difficult or impossible to know the uncompressed in the same instrumentation point. Native gRPC tracing instrumentation also [struggles with this]( https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md#tracing-information)

---

If we were to design this from scratch, we would:

* Document which layer records messages (per-call or per-attempt) on the client side and what to do when they are resent
* Document whether they should be log-based events, or potentially recommend that user apps use distributed tracing on a per-message basis when a request–response pattern is used within a stream. And/or have a metric for time between messages.
* Review existing instrumentations or prototype new ones to understand how to record message sizes and which ones matter (probably over-the-wire sizes - what value does uncompressed size provide?).

---

## The TL;DR and proposal:

* We’ll need a new way to record RPC messages, and it will be breaking (at least they won’t be span events).
* We don’t target streaming in the initial RPC stabilization effort.
* We’d like to deprecate message events, so that:
  * existing RPC instrumentations that follow the “old” semconv 1.37.0 or prior keep emitting them
  * RPC instrumentations that follow future, to-be-stable RPC conventions **will not emit message events at all**
* If we get feedback that message events are useful, they can always be added later in a non-breaking manner based on that feedback

lmolkova:
Fixes #3266, fixes #2865

Please check out #3266 for the context.


model/rpc/deprecated/registry-deprecated.yaml
```diff
@@ -284,3 +284,39 @@ groups:
           reason: uncategorized
           note: 'Value should be included in `rpc.method` which is expected
             to be a fully-qualified name.'
+      - id: rpc.message.type
```
lmolkova, line 287: these attributes were only used on the event

model/rpc/deprecated/registry-deprecated.yaml
```diff
@@ -14,31 +14,31 @@ groups:
               value: "RECEIVED"
               stability: development
         stability: development
-        brief: "Deprecated, use `rpc.message.type` instead."
+        brief: "Deprecated, no replacement at this time."
         deprecated:
-          reason: renamed
-          renamed_to: rpc.message.type
+          reason: obsoleted
+          note: Deprecated, no replacement at this time.
```
lmolkova, line 20: this needs to stay up-to-date (we have policies that check that renamed_to exists and is not deprecated)

trask (approved):

.chloggen/deprecate-rpc-message.yaml
```diff
@@ -0,0 +1,8 @@
+change_type: deprecation
+component: rpc
+note: Deprecate the `rpc.message` event and its attributes
```
lmolkova, line 3: if merged after https://github.com/open-telemetry/semantic-conventions/pull/3224, also need to update migration guide

kamphaus (approved):

PR #3295, opened 2026-01-16
[k8s] add k8s.service entity and metrics

### LINKED ISSUE #3294: [k8s] Add k8s.service entity
### Area(s)

area:k8s

### What's missing?

I'd like to propose adding a k8s.service entity and associated metrics, in line with existing k8s entities. While pod-level health metrics exist, k8s services can aggregate endpoints across multiple deployments via label selectors. The proposed metrics provide a direct view of service-level availability without requiring users to discover and aggregate all matching pods.

I would like to introduce the following attributes and metrics for a `k8s.service` entity.

### New Metrics

| Metric | Type | Description |
|--------|------|-------------|
| `k8s.service.endpoint.count` | Gauge | Count of endpoints by condition (ready/serving/terminating), address type, and zone |
| `k8s.service.load_balancer.ingress.count` | Gauge | Count of ingress entries provisioned by load balancer controller |

### New Attributes

**Metric attributes:**
- `k8s.service.endpoint.condition` (enum: ready, serving, terminating)
- `k8s.service.endpoint.address_type` (enum: IPv4, IPv6, FQDN)
- `k8s.service.endpoint.zone` (string)

**Service entity:**
- `k8s.service.traffic_distribution` (string, opt-in)
- `k8s.service.publish_not_ready_addresses` (bool, opt-in)

Listing out a few usecases for the proposed attributes and metrics -

- Service endpoint health during deployments - Track how many endpoints are ready to receive new traffic, still serving in-flight requests while terminating, or warming up. This is useful for assessing rolling update progress.
- LoadBalancer provisioning status - Confirm that external IPs or hostnames have been assigned by the load balancer controller, and detect when services lose their external endpoints.
- Zone-aware endpoint distribution - With topology-aware routing, operators need to verify each availability zone has sufficient endpoints and alert on zone imbalances.

### Tip

jinja2:
Fixes #3294

## Changes

Add `k8s.service` entity and metrics related to service endpoints and LB ingress. 

> [!IMPORTANT]


model/k8s/registry.yaml
```diff
@@ -793,3 +793,171 @@ groups:
...
+        brief: >
+          The address type of the service endpoint.
+        note: |
+          The network address family or type of the endpoint.
+          This attribute aligns with the `addressType` field of the
+          [K8s EndpointSlice](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoint-slice-v1/).
+          It is used to differentiate metrics when a Service is backed by multiple address types
+          (e.g., in dual-stack clusters).
+        examples: ['IPv4', 'IPv6']
+      - id: k8s.service.endpoint.zone
```
jinja2, line 949: We have `cloud.availability_zone` already but I wasn't sure if we should use this for k8s. The endpoint's zone corresponds to the topology.kubernetes.io/zone label on the node, which represents a failure domain and not necessarily a cloud availability zone for on-prem or bare-metal clusters. Open to feedback if this is too much splitting hairs and we should just use `cloud.availability_zone`, or any other suggestions - maybe `k8s.topology.zone`?

model/k8s/entities.yaml
```diff
@@ -238,3 +238,34 @@ groups:
...
+    name: k8s.service
+    brief: >
+      A Kubernetes Service object.
+    attributes:
+      - ref: k8s.service.uid
+        role: identifying
+      - ref: k8s.service.name
+        role: descriptive
+      - ref: k8s.namespace.name
+        role: descriptive
```
dmitryax: This is part of `k8s.namespace` entity. Should not be listed here

model/k8s/metrics.yaml
```diff
@@ -1773,3 +1773,68 @@ groups:
...
+    brief: "Number of endpoints for a service by condition and address type."
+    instrument: gauge
+    unit: "{endpoint}"
+    entity_associations:
+      - k8s.service
+    attributes:
+      - ref: k8s.service.endpoint.condition
+        requirement_level: required
+      - ref: k8s.service.endpoint.address_type
+        requirement_level: required
```
dmitryax, line 1794: Why these two have to be required?

model/k8s/registry.yaml
dmitryax: I think `k8s.service.endpoint.zone` is a pretty good name here

model/k8s/metrics.yaml
jinja2: The reason I went with required here is to prevent the metric from being misleading if a user aggregates it by dropping these dimensions.

For condition: endpoint conditions aren't mutually exclusive. During a rollout, a pod is often both serving and terminating at the same time. If we aggregate by dropping this attribute, the math breaks because one pod is counted twice.

It’s similar for address_type in dual-stack clusters. A single pod has both an IPv4 and an IPv6 address reported in separate EndpointSlices. Without this being required, a simple sum would make a 10-pod service look like it has 20 endpoints.

I think these two attributes are essential context for the counts to be meaningful

dmitryax: Sounds reasonable. Thanks!

model/k8s/metrics.yaml
```diff
@@ -1773,3 +1773,68 @@ groups:
...
         requirement_level: required
+
+  # k8s.service.* metrics
+  - id: metric.k8s.service.endpoint.count
+    type: metric
+    metric_name: k8s.service.endpoint.count
+    annotations:
+      code_generation:
+        metric_value_type: int
+    stability: experimental
```
ChrsMark: I see a mix of `experimental` and `development` within the PR. Is this on purpose? If I'm not mistaken we use `development`?

jinja2: Fixed the PR to only have `development` stability

ChrsMark (approved):
LGTM

jinja2:
@open-telemetry/specs-semconv-maintainers this one is approved by the K8s SIG and ready to be merged. Thanks!

model/k8s/registry.yaml
```diff
@@ -793,3 +793,171 @@ groups:
           The phase for the pod. Corresponds to the `phase` field of the:
           [K8s PodStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#podstatus-v1-core)
         examples: [ 'Pending', 'Running' ]
+      - id: k8s.service.uid
+        type: string
+        stability: development
+        brief: >
+          The UID of the Service.
+        examples: ['275ecb36-5aa8-4c2a-9c47-d8bb681b9aff']
+      - id: k8s.service.name
```
lmolkova, line 802: Given some intersection with `service.name` here and potential other intersections with service conventions (e.g. `k8s.service.endpoint.zone`), I'd love to get some feedback from @open-telemetry/semconv-service-and-deployment-approvers SIG on this

jinja2: For context, there's [non-normative K8s guidance](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-attributes.md) that clarifies how `service.*` attributes should be calculated in K8s environments. They're derived from workload objects (Deployment, StatefulSet, Pod names, etc.). The `k8s.service.*` here refers to the K8s Service object (a network abstraction for exposing pods), which is conceptually different to the logical application identity.

Happy to get more eyes on this if there are concerns!

jsuereth: For context - even the opentelemetry-operator project is using the deployment/workload-controller as the `service.name`.  I think `service` in opentelemetry represents a workload more than it does a network service, and the `k8s.service` is a very different thing.  If we tried to make them be the same, we'd really confuse users.

It'd be *more* accurate for us to try to rename `service` to be something more generic and less network-y, but we've crossed that bridge several times and decided not to.

PR #3215, opened 2025-12-15
Promote a selection of k8s and container attributes to alpha

ChrsMark:
## Changes

This is part of https://github.com/open-telemetry/semantic-conventions/issues/3120. The purpose of this PR is to advertise the intention and start collecting feedback.

All suggested attributes are already in use from several Collector components as described at https://github.com/open-telemetry/semantic-conventions/issues/3120#issue-3659290856:

> - [k8sattributes processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) (from [metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/metadata.yaml#L14C1-L135C19) and [extract::metadata](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/config.go#L133-L147))
> - [resourcedetection processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor)
> - [filelog receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver#file-log-receiver)

> [!IMPORTANT]
> Pull requests acceptance are subject to the triage process as described in [Issue and PR Triage Management](https://github.com/open-telemetry/semantic-conventions/blob/main/issue-management.md).
> PRs that do not follow the guidance above, may be automatically rejected and closed.

## Merge requirement checklist

* [x] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
  * If your PR does not need a change log, start the PR title with `[chore]`
* [x] Links to the prototypes or existing instrumentations (when adding or changing conventions)

Issue #3119, opened 2025-11-24
[k8s] Stability plan for semantic conventions


ChrsMark:
During KubeCon North America 2025 the group met and decided on the following regarding the stability plan:

### Phase 1

ref: https://github.com/open-telemetry/semantic-conventions/issues/3120

Focus on semantic conventions that are used by Collector components that target stability soon: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44130

These are the following:

- [k8sattributes processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor)
- [resourcedetection processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor)
- [filelog receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver#file-log-receiver)

The above probably rely on Resource Attributes and Entities only, not metrics.

Extra:
- [prometheus receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver#prometheus-receiver)

### Phase 2

Metrics Semantic Conventions for widely used Collector Components:

- [kubeletstats receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/kubeletstatsreceiver)
- [k8scluster receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sclusterreceiver)

### Phase 3

Additional components that use k8s semantic conventions are coming with lower priority and would not necessarily require for the group to meet regularly:  

- [k8sobjects receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver)
- [k8sevents receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8seventsreceiver)
- [dockerstats receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/dockerstatsreceiver)

/cc @open-telemetry/semconv-k8s-approvers @dashpole

PR #3304, opened 2026-01-19
Promote a selection of k8s and container attributes to beta

ChrsMark:
## Changes

Similar to https://github.com/open-telemetry/semantic-conventions/pull/3215.
This is part of https://github.com/open-telemetry/semantic-conventions/issues/3120. The purpose of this PR is to again advertise the intention towards having those attributes as Release Candidates soon.

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45447 aims to use the `alpha` version in the Collector Contrib already.

All suggested attributes are already in use from several Collector components as described at https://github.com/open-telemetry/semantic-conventions/issues/3120#issue-3659290856:

> - [k8sattributes processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) (from [metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/metadata.yaml#L14C1-L135C19) and [extract::metadata](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/config.go#L133-L147))
> - [resourcedetection processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor)
> - [filelog receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver#file-log-receiver)

Closes https://github.com/open-telemetry/semantic-conventions/issues/3381

> [!IMPORTANT]


templates/registry/markdown/stability.j2
```diff
@@ -5,6 +5,7 @@
 {%- elif stability == "release_candidate" %}![Release Candidate](https://img.shields.io/badge/-rc-mediumorchid)
 {%- elif stability == "deprecated" %}![Deprecated](https://img.shields.io/badge/-deprecated-red)
 {%- elif stability == "alpha" %}![Alpha](https://img.shields.io/badge/alpha-mediumpurple)
+{%- elif stability == "beta" %}![Beta](https://img.shields.io/badge/beta-mediumpink)
```
thompson-tomo:
```suggestion
{%- elif stability == "beta" %}![Beta](https://img.shields.io/badge/beta-lightpink)
```

templates/registry/markdown/stability.j2
```diff
@@ -18,6 +19,7 @@
 {%- elif stability == "release_candidate" %}<img src="https://img.shields.io/badge/-rc-mediumorchid" alt="Release Candidate"/>
 {%- elif stability == "deprecated" %}<img src="https://img.shields.io/badge/-deprecated-red" alt="Deprecated" />
 {%- elif stability == "alpha" %}<img src="https://img.shields.io/badge/alpha-mediumpurple" alt="Alpha"/>
+{%- elif stability == "beta" %}<img src="https://img.shields.io/badge/beta-mediumpink" alt="Beta"/>
```
thompson-tomo:
```suggestion
{%- elif stability == "beta" %}<img src="https://img.shields.io/badge/beta-lightpink" alt="Beta"/>
```

ChrsMark:
@open-telemetry/semconv-k8s-approvers PTAL

jinja2 (approved):

joaopgrassi:
@ChrsMark something looks wrong with the commit ids, the diff for the check fails. Maybe a rebase fixes it.

ChrsMark:
@joaopgrassi thank's for the heads-up, should be good now

trask:
before merging, I'd like to discuss since we haven't used beta in semconv previously

- what's the concrete value / implication of going from release to beta? if it's just to advertise the intention of going to release candidate soon, maybe a blog post would be better way to accomplish?
- what does this imply about all of the other semconv that are (still in) development instead of (the now used) beta

joaopgrassi:
While reviewing, I re-read https://opentelemetry.io/docs/specs/otel/document-status/. To me moving to Alpha sounds like the natural path. If now the sig believes that there's high confidence in the current design, meaning no new features/breaking changes, then I guess moving directly to RC makes sense.

ChrsMark:
> what's the concrete value / implication of going from release to beta? if it's just to advertise the intention of going to release candidate soon, maybe a blog post would be better way to accomplish

The idea is to go through the stability levels described at https://opentelemetry.io/docs/specs/semconv/general/semantic-convention-groups/#group-stability before graduating as `stable`. Moving from one stage to the next one per release allows us to bump the version in the Collector with PRs like https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45447. We could go directly to `release_candidate` skipping `alpha` and `beta`, but I felt that going through the full stability level steps will allow us to gradually update both the semconv part but the Collector part as well in a less "aggressive" way.

The ultimate goal (of the K8s SIG in alignment with the Collector SIG as per https://github.com/open-telemetry/semantic-conventions/issues/3119) is to transition the `k8sattributes` processor into using the `beta`/`release_candidate` respective semconvs behind a feature gates pair (see https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45307) following the Collector's [guidelines](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/semconv-feature-gates.md) as part of the [component's stability effort](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44483). This will then allow us to release these semconvs as `stable` once we use them for a while behind the feature gate pair.

Since we have lot's of work ahead I would advertise this change with a blogpost once we have these semantic conventions in `release_candidate` level and already introduced in the Collector's `k8sattributes` processor behind the feature gate pair.

I would target for the `release_candidate` before KubeConEU 26' if possible so it can also be announced there. (However this would imply that we will have 2 releases of SemConv project in the following 1.5 months, so not sure if that's realistic. We could have one next week (early Feb) and then one early March)

Let me know what you think @trask @joaopgrassi.

ChrsMark:
@open-telemetry/specs-semconv-approvers let us know if there is anything missing here

trask:
@ChrsMark thanks, sounds like a good plan 👍

ChrsMark:
> Should this also be moving the entities to beta?

Entities have not been included since they are not introduced yet in `k8sattributes` processor. For the Collector's component stability (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44483) we only need the attributes by now. We discussed that recently with @dmitryax and we agreed that we wouldn't like to block v1 of Collector's components on entities' work.

So @jsuereth I think entities can wait unless we have a specific need for this?

dmitryax:
I think it makes sense to promote the attribute names separately from the entities as a whole at this point. It’s important to provide more assurance to users that we’re not planning to change the names themselves in the future even as part of the entity work. I believe many systems rely on these naming conventions already.

jsuereth:
I'm just trying to make sure we have a consistent policy around stabilization here with attributes, entities + signals.

I'm fine stabilizing attributes then revisiting stabilization for entities.

PR #3305, opened 2026-01-19
semconv(error): add error.source attribute to error registry

bwdmr:


lmolkova (changes requested):
It seems instrumentation needs a lot of context and some non-trivial judgement to populate `error.source` attribute. Most of the generic instrumentations won't be able to. 

The only place I know where it's available is [Exception.Source](https://learn.microsoft.com/en-us/dotnet/api/system.exception.source?view=net-10.0) in .NET and then  it belongs  in the `exception` namespace. 

As the result, the definition is very vague - anything can be source (e.g. chatgpt thinks that source code or stack trace are related to it).  

If you need this attribute for your manual instrumentations and error tracking - go for it, but I don't believe we're ready to document it in OTel semantic conventions without some signals that it's necessary in 2+ generic instrumentations. We'd also need some prototypes and more details on how someone would set it.

bwdmr:
I would like to emphasize this point.

“Source” is not meant as the source code or the origin of an error, but rather the source of the affected change—that is, the place a user would need to look to identify the likely culprit. It therefore becomes orthogonal to the “type” dimension.
Introducing an additional dimension does fan out the available options for storage and filtering, but in this case that dimension is both intentional and fundamental.

If there is anything I can help with to steer this toward a clearer or more commonly understood description, I’m happy to do so.

lmolkova:
@bwdmr I don't doubt that it maybe useful, I doubt that's it's feasible to populate in generic manner. 

If you could come up with a couple of prototypes of how generic (not application-specific) instrumentations would populate such an attribute without any additional context, would be a good way to make progress.

bwdmr:
@lmolkova 

You’re right that “the true origin/root cause” isn’t generically knowable.
But for **`error.source`** we can define it as a *best-effort “where should I look first”* hint, and populate it generically with a clear precedence rule:

1. **Code location**: first non-instrumentation stack frame (function or file:line).
2. If it’s a boundary/dependency failure: **peer/endpoint** (peer.service / host / db.system…).
3. Else: `unknown`.

This is intentionally **orthogonal to `error.type`**:

* **`error.type`** answers *“what kind of failure is this?”* (timeout vs validation vs exception class, etc.) and is primarily about **classification** for grouping.
* **`error.source`** answers *“where should I start looking for the culprit?”* (emitting callsite, failing dependency, edge) and is primarily about **attribution/triage**.

Because they’re orthogonal, they complement each other: you can have the same `error.type=timeout` with different `error.source` values (Stripe vs Postgres vs internal callsite), and conversely the same `error.source` producing different `error.type` values over time. That’s exactly the dimension we’re trying to capture, and it’s feasible to populate generically as above while still allowing richer overrides when a subsystem has more context.

PR #3308, opened 2026-01-19
Deprecate error.message in favor of domain-specific attributes

### LINKED ISSUE #3307: Deprecate `error.mesage` and use domain-specific attributes instead
We've defined `error.message` in #1992 as a replacement for `feature_flag.evaluation.error.message`.

It turned out to be problematic to come up with requirements on when to use it vs `exception.message` - https://github.com/open-telemetry/semantic-conventions/pull/2296

It's also somewhat confusing when it comes to logs (is `error.message` related to log body with error severity).

So `error.message` is very generic. Unlike `error.type` it's not intended to be used on metrics or span -> metrics pipelines and the benefit of unification of error messages across domains is not huge.

At the same time, domain-specific error messages look more descriptive. Some candidate
- `(g)rpc.response.status_description` as a companion of `rpc.response.status_code` using domain-specific terminology
- `http.response.reason_phrase`
- `feature_flag.error.message` - can be documented as precisely as needed. This is also the only domain where `error.message` is used. 

Based on the discussions in logs sig, we're proposing deprecating `error.message` and replacing it with domain-specific attributes.

/cc @open-telemetry/specs-logs-approvers @open-telemetry/semconv-feature-flag-approvers

lmolkova:
Fixes #3307


model/feature-flags/registry.yaml
```diff
@@ -99,3 +99,9 @@ groups:
 
           It may be desirable to redact or otherwise limit the size and scope of `feature_flag.result.value` if possible.
           Because the evaluated flag value is unstructured and may be any type, it is left to the instrumentation author to determine how best to achieve this.
+      - id: feature_flag.error.message
```
lmolkova, line 102: relevant discussion: https://github.com/open-telemetry/semantic-conventions/pull/1990#discussion_r1991576639 entertaining `feature_flag.evaluation.error.message` vs `feature_flag.result.error.message`.

The `result` seems to be aligned with OpenFeature terminology: [structure representing the result of the 
flag evaluation process](https://openfeature.dev/specification/glossary#evaluating-flag-values), so error message is a property of result.

But in a practical sense, there is no need to use `evaluation` or `result`.

.chloggen/deprecate-error-message.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: breaking
+component: error, feature-flags
+note: >
+  Deprecate `error.message` in favor of using domain-specific codes such as `feature_flag.error.message`.
+issues: [3307, TODO]
```
lmolkova:
```suggestion
issues: [3307, 3308]
```

trask (approved):

docs/feature-flags/feature-flags-events.md
```diff
@@ -93,9 +93,9 @@ for referring to a value without including the value itself. This can
 provide additional context for understanding the meaning behind a value.
 For example, the variant `red` maybe be used for the value `#c05543`.
 
-**[7] `error.message`:** If and only if an error occurred during flag evaluation and `error.type` does not sufficiently describe the error.
+**[7] `feature_flag.error.message`:** If and only if an error occurred during flag evaluation and `error.type` does not sufficiently describe the error.
 
-**[8] `error.message`:** Should not simply duplicate the value of `error.type`, but should provide more context. For example, if `error.type` is `invalid_context` the `error.message` may enumerate which context keys are missing or invalid.
+**[8] `feature_flag.error.message`:** Should not simply duplicate the value of `error.type`, but should provide more context. For example, if `error.type` is `invalid_context` the `error.message` may enumerate which context keys are missing or invalid.
```
askpt:
```suggestion
**[8] `feature_flag.error.message`:** Should not simply duplicate the value of `error.type`, but should provide more context. For example, if `error.type` is `invalid_context` the `feature_flag.error.message` may enumerate which context keys are missing or invalid.
```

docs/registry/attributes/feature-flag.md
```diff
@@ -58,7 +59,7 @@ Describes deprecated feature flag attributes.
 
 | Key | Stability | Value Type | Description | Example Values |
 | --- | --- | --- | --- | --- |
-| <a id="feature-flag-evaluation-error-message" href="#feature-flag-evaluation-error-message">`feature_flag.evaluation.error.message`</a> | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `error.message`. | string | Deprecated, use `error.message` instead. | `Flag `header-color` expected type `string` but found type `number`` |
+| <a id="feature-flag-evaluation-error-message" href="#feature-flag-evaluation-error-message">`feature_flag.evaluation.error.message`</a> | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `feature_flag.error.message`. | string | Deprecated, use `error.message` instead. | `Flag `header-color` expected type `string` but found type `number`` |
```
askpt:
```suggestion
| <a id="feature-flag-evaluation-error-message" href="#feature-flag-evaluation-error-message">`feature_flag.evaluation.error.message`</a> | ![Deprecated](https://img.shields.io/badge/-deprecated-red)<br>Replaced by `feature_flag.error.message`. | string | Deprecated, use `feature_flag.error.message` instead. | `Flag `header-color` expected type `string` but found type `number`` |
```

docs/registry/attributes/feature-flag.md
askpt, line 54: Quick question, we are not adding `error.message` attribute here because it is part of the `error` registry, correct?

docs/feature-flags/feature-flags-events.md
lmolkova: great catch, thanks!

docs/registry/attributes/feature-flag.md
lmolkova: yep, this file lists all ff attributes. Deprecated `error.message` is still there in the corresponding `error` file. We no longer mention `error.message` on the ff evaluation events, so it just disappears from ff docs

beeme1mr:
After reviewing the linked issue, it's still not clear to me why `exception.message` is okay, but `error.message` is too generic. When we were putting together the feature flag SemCon, it was recommended to use a shared attribute, which seemed reasonable. I could honestly go either way, but I'd prefer not to depreciate an attribute to avoid unnecessary thrashing unless there's a very strong reason to make the change.

dyladan:
> This PR contains changes to area(s) that do not have an active SIG/project and will be auto-closed:
> 
> * error
> * error
> * feature-flags
> * feature-flags

I think the tooling here is a bit aggressive. It's expected that sigs wind down after doing their work. There isn't any reason for the feature flag sig to continue to meet when regular changes are not expected. 

>it's still not clear to me why `exception.message` is okay, but `error.message` is too generic

I agree with @beeme1mr it seems like this is a documentation issue. I found an attempt to document this at https://github.com/open-telemetry/semantic-conventions/pull/2296 which was closed as stale, but I don't see any unresolvable issues with it.

To me the biggest issue with using a generic attribute is what happens if multiple semconv areas need to use it on the same span. So far this has been avoided with `error.message` because it is really only used in feature flagging, and we are using it in an event which sidesteps this error because other semconv/instrumentations would emit separate events.

lmolkova:
@dyladan @beeme1mr thanks for the feedback. 

We've chatted about it in the SemConv call, documenting additional context here:

The `error.message` is not grounded in anything, unlike `exception.message`, which captures common property across most languages, is a stable and actively used attribute.
Aside from its name (`error.message`), it is fully convention-specific. When working on #2296 and #3256, we realized that it:

- would never appear on spans, as the message would become the span status description
- would never appear on metrics due to cardinality
- could only appear on logs / events. There, it could be confused with `logger.error(message)` - the log body, and `exception.message` (would be marginally different from `error.message` in many cases). Most importantly, there seems to be no point in grouping by `error.message` across all logs or showing this attribute on a generic log dashboard

So the practical benefit of having one attribute across conventions is low, while the ambiguity it introduces is high.

This is the only attribute that is in development state in feature flags and, at least in my head, was the main blocker preventing FF from going stable. While I understand it introduces churn, it also opens up a path forward for FF to stabilize at any time.

michaelsafyan:
> would never appear on spans, as the message would become the span status description

What about in the case of client errors, where span status is left unset, but where there is an error message (the client error)?

michaelsafyan:
One other concern about domain-specific error attributes: visualization. Won't this require visualization UIs to now look for "the error message" in more places? Wouldn't it be better if UIs could easily find a single attribute for "the error message"?

lmolkova:
> One other concern about domain-specific error attributes: visualization. Won't this require visualization UIs to now look for "the error message" in more places? Wouldn't it be better if UIs could easily find a single attribute for "the error message"?

@michaelsafyan do you have a suggestion of the specific user experience that would need it? We've been trying to come up with generic logs dashboard and came up with something like

- Timestamp
- Severity
- Maybe scope name
- Event name
- Maybe error.type - good grouping key
- Body (makes sense for logs dashboard but not for events)

So what would be the case when UX needs to visualize something around error message and would it make sense across all conventions?

lmolkova:
> What about in the case of client errors, where span status is left unset, but where there is an error message (the client error)?

it sounds like it wasn't an error if status was unset and it'd be a log record with appropriate severity. `warn.message` (or `exception.message`) if you will.

michaelsafyan:
In the case of a `SERVER` span with a client error (e.g. 4xx), `error.message` is a convenient way to pass the error, while not marking the entire span operation as an error (not setting the status).

Such an operation could have multiple logs correlated with the span of an appropriate severity, but there is no way to know if the log represents the message of the overall operation vs just a warning that happened within it.

One could imagine a view of the span which elevates error information to bring it front and center. While perhaps there are cases where domain-specific errors are appropriate (errors that are "fail open" and do not fail the operation), deprecating `error.type` (and specifically recommending many different attributes rather than a single attribute with common meaning) is inconvenient with respect to being able to trivially pull out the "operation-level error message".

lmolkova:
> In the case of a `SERVER` span with a client error (e.g. 4xx), `error.message` is a convenient way to pass the error, while not marking the entire span operation as an error (not setting the status).
> 
> Such an operation could have multiple logs correlated with the span of an appropriate severity, but there is no way to know if the log represents the message of the overall operation vs just a warning that happened within it.
> 
> One could imagine a view of the span which elevates error information to bring it front and center. While perhaps there are cases where domain-specific errors are appropriate (errors that are "fail open" and do not fail the operation), deprecating `error.type` (and specifically recommending many different attributes rather than a single attribute with common meaning) is inconvenient with respect to being able to trivially pull out the "operation-level error message".

@michaelsafyan I see your point and I think this scenario makes sense. Note that `error.message` was not intended to be that kind of thing though. What you're suggesting is something like `operation-ending-messaage`, not necessarily an error (which `error.message` would imply).

michaelsafyan:
Would it be possible to simply expand the meaning/scope of `error.message` (or add clarifying examples of when to use `error.message` vs `exception.message`) rather than deprecating it?

Note that OTel SemConv is not comprehensive or complete. There are uses of OTel SemConv in vendor-specific conventions that may not necessarily be documented in the OTel SemConv repo. That is, it is incorrect to conclude that lack of references within OTel SemConv necessarily means that the attribute is unused for the given meaning/purpose.

lmolkova:
> Would it be possible to simply expand the meaning/scope of `error.message` (or add clarifying examples of when to use `error.message` vs `exception.message`) rather than deprecating it?
> 
> Note that OTel SemConv is not comprehensive or complete. There are uses of OTel SemConv in vendor-specific conventions that may not necessarily be documented in the OTel SemConv repo. That is, it is incorrect to conclude that lack of references within OTel SemConv necessarily means that the attribute is unused for the given meaning/purpose.

As you can see, we tried to define this attribute and failed. Please create an issue and it'd also be a good change to rename this attribute to match the use-case.

We don't recommend external semconv registries to take dependency on experimental parts of OTel semconv.

lmolkova:
@open-telemetry/semconv-feature-flag-approvers Could you please take another look?

beeme1mr (approved):

Issue #1594, opened 2021-09-15
Proposal: Add Sensitive Data Labels


MovieStoreGuy:
## The Pitch
The idea is rather simple, extending the current OTLP to include an additional dictionary that is intended to store sensitive data. 

## The Rationale
Making the use of sensitive data being explicit instead implicit means that data vendors, data processors (pre and post) can first of all:
- Ensure any data regulations are followed (Thinking GDPR, CCPA, FedRamp to a degree)
- Allow for consumers of this data to handle the data correctly 

As we collect more telemetry from user's experiences, the likelihood of including UGC and PII increases and this should help those system that collect it to be clear on the actions they take on it.

## The outcome
The idea I have in mind and would love further discussion on is how this is consumed by the client (the application actually sending the data).

The idea being that the OTLP is extending to include another dictionary that is explicit in its sensitive nature and the SDK implement a `AddSensitiveLabel` method to make it clear on what it is.

From that the otel-collector could filter, drop, or transform those labels if they are known. 

From there, the exported data vendors / processes can set up ingestion policies on what should happen when sensitive data arrives.

lmolkova:
I believe it can be done today with annotations:

```yaml
- id: sensitive.attribute
  annotations:
    sensitivity: PII | secrets | all | none  # need to define and polish this enum
    reduction:  # this could be added later, need some design
      strategy: drop  # more strategies can be defined, e.g. regex, `strategy` serves as a discriminator for `reduction` type
```

svrnm:
> I believe it can be done today with annotations:
> 
> - id: sensitive.attribute
>   annotations:
>     sensitivity: PII | secrets | all | none  # need to define and polish this enum
>     reduction:  # this could be added later, need some design
>       strategy: drop  # more strategies can be defined, e.g. regex, `strategy` serves as a discriminator for `reduction` type

If we want to start with something simple, yes, this could work, but it's important to recognize, that sensitivity can be contextual, especially when it is PII, e.g. we had these examples with a user id or an IP address being sensitive based on the context. But maybe it's possible to encode that as well, I think a lot of it boils down to the question if there is a human person on the client side of the communication, or if any data that's added as an attribute refers to a real human.

PR #3310, opened 2026-01-19
Indicate that `exception.message` attribute may contain sensitive information

lmolkova:
Towards #2967

`exception.message` is recorded by default on span-events (for now) and logs. It's stable and too useful.
At the same time, there are known cases when exception message contains things like connection strings or user_ids. 

It is not recommended to include sensitive information in exception messages.

In many cases, exceptions are thrown and recorded as telemetry by  different components. As a result, it may not be possible for  instrumentation to prevent sensitive information from being recorded.

Additional processing in the telemetry pipeline may be necessary to remove sensitive information.

So adding sensitivity warning on the attribute. https://github.com/open-telemetry/semantic-conventions/issues/1594 would improve this story (by making sensitivity note machine readable).

### DIFF
--- a/.chloggen/exception-message-sensitive.yaml
+++ b/.chloggen/exception-message-sensitive.yaml
@@ -1,4 +1,4 @@
 change_type: enhancement
 component: exception
 note: Indicate that `exception.message` attribute may contain sensitive information.
-issues: [2967, TODO]
+issues: [2967, 3310]
--- a/docs/exceptions/exceptions-logs.md
+++ b/docs/exceptions/exceptions-logs.md
@@ -54,7 +54,7 @@ The table below indicates which attributes should be added to the
 
 > [!WARNING]
 >
-> This attribute contains sensitive (PII, secrets) information
+> This attribute may contain sensitive information.
 
 **[3] `exception.type`:** Required if `exception.message` is not set, recommended otherwise.
 
--- a/model/exceptions/registry.yaml
+++ b/model/exceptions/registry.yaml
@@ -23,7 +23,7 @@ groups:
         note: |
           > [!WARNING]
           >
-          > This attribute contains sensitive (PII, secrets) information
+          > This attribute may contain sensitive information.
       - id: exception.stacktrace
         type: string
         stability: stable
--- a/docs/exceptions/exceptions-spans.md
+++ b/docs/exceptions/exceptions-spans.md
@@ -44,7 +44,7 @@ This event describes a single exception.
 
 > [!WARNING]
 >
-> This attribute contains sensitive (PII, secrets) information
+> This attribute may contain sensitive information.
 
 **[3] `exception.type`:** Required if `exception.message` is not set, recommended otherwise.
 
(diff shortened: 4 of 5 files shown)


.chloggen/exception-message-sensitive.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: exception
+note: Indicate that `exception.message` attribute may contain sensitive information.
+issues: [2967, TODO]
```
lmolkova:
```suggestion
issues: [2967, 3310]
```

model/exceptions/registry.yaml
```diff
@@ -20,6 +20,20 @@ groups:
...
+          In many cases, exceptions are thrown and recorded as telemetry by
+          different components. As a result, it may not be possible for
+          instrumentation to prevent sensitive information from being recorded.
+
+          Additional processing in the telemetry pipeline may be necessary to
+          remove sensitive information.
+
+          > [!WARNING]
+          >
+          > This attribute contains sensitive (PII, secrets) information
```
arminru:
```suggestion
          > This attribute may contain sensitive information like PII or authentication secrets.
```
Unless you want to stick more closely to the example we currently give at https://github.com/open-telemetry/semantic-conventions/blob/main/docs/how-to-write-conventions/README.md#defining-attributes.

pellared: I was about to suggest almost the same

```suggestion
          > This attribute may contain sensitive (PII, secrets) information
```

lmolkova: thanks, great catch!  copied it from somewhere apparently without reading 🤦‍♀️

docs/exceptions/exceptions-logs.md
```diff
@@ -44,13 +44,27 @@ The table below indicates which attributes should be added to the
...
 **[1] `exception.message`:** Required if `exception.type` is not set, recommended otherwise.
 
-**[2] `exception.type`:** Required if `exception.message` is not set, recommended otherwise.
+**[2] `exception.message`:** It is NOT RECOMMENDED to include sensitive information in exception messages.
+
+In many cases, exceptions are thrown and recorded as telemetry by
+different components. As a result, it may not be possible for
+instrumentation to prevent sensitive information from being recorded.
+
+Additional processing in the telemetry pipeline may be necessary to
```
lmolkova: @trask wanted to make a suggestion,  leaving a comment so it's not get accidentally merged

model/exceptions/registry.yaml
```diff
@@ -20,6 +20,21 @@ groups:
...
+          instrumentation to prevent sensitive information from being recorded.
+
+          Additional processing in the telemetry pipeline may be necessary to
+          remove sensitive information.
+
+          > [!WARNING]
+          >
+          > This attribute may contain sensitive information like PII or
+          > authentication secrets.
+
```
trask: I kind of prefer just tagging it as "may contain sensitive information", wdyt?

```suggestion
        note: |
          > [!WARNING]
          >
          > This attribute may contain sensitive information.

```

PR #3311, opened 2026-01-19
Add event name and severity considerations to log-based exception events

### LINKED ISSUE #3100: [Logs] Recommend setting Severity Number
### Area(s)

area:log

### What's missing?

Logs without severity are bad. It's expensive to manage logs when you don't know if they can be dropped/sampled or not.

Original Slack thread: https://cloud-native.slack.com/archives/C062HUREGUV/p1762896560429319

It would be nice to encourage users to always set the Severity Number. This could be done by adding "general logs semantics" similarly to [general event semantics](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/events.md#general-event-semantics). 

### Tip

lmolkova:
Fixes #3100 

Extracted from https://github.com/open-telemetry/opentelemetry-specification/pull/4333

Related to https://github.com/open-telemetry/semantic-conventions/pull/3256

Summarizes guidance behind: #3414, #3426, #3435, #3436

The current guidance on recording exceptions on logs only tells how to attach exception-related details to a log record.
OTel instrumentations, when switching to log-based events from span events, would need to emit the whole record and need guidance on:
- whether to emit log or event. This PR recommends events.
- how to name events - use precise names like `http.client.request.exception`. OTel instrumentations that emit them should document them in central or instrumentation-specific docs. 
- how to set severity. This PR suggests to use Fatal or Error only when exception is not handled by application code. I.e. instrumentations that create server/consumer spans would be ones to ever use `error` severity. Client and producer should emit `warn`.

### DIFF
--- a/docs/exceptions/exceptions-logs.md
+++ b/docs/exceptions/exceptions-logs.md
@@ -95,7 +95,7 @@ For example, [`http.client.request.exception`](/docs/http/http-exceptions.md#htt
 represents exceptions that occur during an HTTP client request.
 
 Instrumentations that are not specific to a particular operation or domain, such as
-a global unhandled exception handler, SHOULD use the `exception` event name.
+global unhandled exception handlers, SHOULD use the `exception` event name.
 
 ### Severity
 
@@ -119,7 +119,7 @@ Examples:
 
 > [!NOTE]
 > Instrumentation SHOULD do the best effort to record such errors, but
-> OpenTelemetry SDK and exporters might not have a chance to actually export it.
+> OpenTelemetry SDK and exporters might not have a chance to actually export them.
 
 #### ERROR severity
 


.chloggen/exception-log-severity.yaml
```diff
@@ -0,0 +1,6 @@
+change_type: enhancement
+component: exception
+note: >
+  Improve exception (as log) definition: add guidance on when to use logs and
+  events, and how to set event name and severity.
+issues: [TODO]
```
lmolkova:
```suggestion
issues: [3311]
```

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +46,99 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+API, instrumentations SHOULD provide the exception instance rather than
+manually setting individual exception attributes.
+
+### Event name
+
+**Status**: [Development][DocumentStatus]
+
+It is RECOMMENDED to provide an
+[event name](/docs/general/events.md) that describes the instrumented operation
+and appends a `.exception` suffix.
```
lmolkova: update https://github.com/lmolkova/semantic-conventions/blob/7bfd63042d09fbe3571bd06f688f74f04086642f/docs/general/recording-errors.md to align

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +46,99 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+Semantic conventions that define `SERVER` or `CONSUMER` spans SHOULD also define
+a corresponding exception event and recommend using `ERROR` severity.
+
+Examples:
+
+- A background job terminates with an exception.
+- An HTTP server framework error handler catches an exception not handled by the
+  application code.
+
+Some libraries and frameworks generate artificial exceptions for operations that
```
lmolkova: apply to all exceptions, not just error
add a couple of examples

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +46,99 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+Examples:
+
+- A connection attempt to a remote service times out.
+- Writing data to a file results in an IO exception.
+- A remote service returns a 503 "Service Unavailable" response for 5 consecutive
+  attempts; retry attempts are exhausted and the operation fails.
+
+#### INFO and below severity
+
+Exceptions that don't indicate an actual issue SHOULD be recorded with severity
```
lmolkova, line 161: add some guidance that this takes higher priority than warn/error/fatal

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +46,99 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+The severity reflects the expected impact of the exception, not just its presence.
+
+[Severity Number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-severitynumber)
+SHOULD be provided on all exception events and SHOULD be set based on the context
+in which the exception occurs.
+
+#### FATAL severity
+
+Exceptions that usually result in application shutdown SHOULD be recorded with
+severity `FATAL` (severity number 21).
```
pellared, line 113: Shouldn't we allow/suggest using the whole FATAL range? The same comment applies to other severity levels.

lmolkova: good question.

Spec [says](https://github.com/open-telemetry/opentelemetry-specification/blob/550ddccf9fcfeb006d19abc1c92c1e51487262a0/specification/logs/data-model.md?plain=1#L290)

> If the source format has more than one severity that matches a single range in
this table then the severities of the source format must be assigned numerical
values from that range according to how severe (important) the source severity
is.
...
> If the source format has only a single severity that matches the meaning of the
range then it is recommended to assign that severity the smallest value of the
range.

So multiple numbers are primarily for log bridge mapping(?).

I think in practice, otel instrumentations have no means to understand if some exception is more fatal than another and having a single number provides more consistency and clarity.
At the same time, if some instrumentations do have context, this comes with SHOULD and they can use 22 instead of 21.

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +46,110 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+### Severity
+
+**Status**: [Development][DocumentStatus]
+
+The severity reflects the expected impact of the exception, not just its presence.
+
+[Severity Number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-severitynumber)
+SHOULD be provided on all exception events and SHOULD be set based on the context
+in which the exception occurs, following the guidance below.
+The lowest applicable severity number MUST be used.
```
lmolkova:
```suggestion
The lowest applicable severity number SHOULD be used.
```

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +46,110 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+- A background job terminates with an exception.
+- An HTTP server framework error handler catches an exception not handled by the
+  application code.
+
+#### WARN severity
+
+Exceptions that are expected to be handled by application code SHOULD be
+reported with severity `WARN` (severity number 13).
+
+Semantic conventions that define `CLIENT`, `PRODUCER`, or `INTERNAL` spans SHOULD
```
lmolkova:
```suggestion
Semantic conventions that define `CLIENT` or `PRODUCER` spans SHOULD
```

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +46,110 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+also define a corresponding exception event and recommend using `WARN` severity.
+
+Examples:
+
+- A connection attempt to a remote service times out.
+- Writing data to a file results in an IO exception.
+- A remote service returns a 503 "Service Unavailable" response for 5 consecutive
+  attempts; retry attempts are exhausted and the operation fails.
+
+#### INFO and below severity
```
lmolkova:
```suggestion
#### Other exceptions
```

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +46,110 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+
+Exceptions that don't indicate an actual issue SHOULD be recorded with severity
+`INFO` (severity number 9) or below.
+
+Examples:
+
+- An exception indicating that a request was cancelled on the client side is thrown on the
+  server and detected by the server instrumentation.
+- An error is returned when checking whether an optional dependency or resource exists.
```
lmolkova:
```suggestion
```

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +46,110 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+
+- A connection attempt to a remote service times out.
+- Writing data to a file results in an IO exception.
+- A remote service returns a 503 "Service Unavailable" response for 5 consecutive
+  attempts; retry attempts are exhausted and the operation fails.
+
+#### INFO and below severity
+
+Exceptions that don't indicate an actual issue SHOULD be recorded with severity
+`INFO` (severity number 9) or below.
```
lmolkova:
```suggestion
`DEBUG` (severity number 9).
```

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +47,108 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+
+[Severity Number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-severitynumber)
+SHOULD be provided on all exception events and SHOULD be set based on the context
+in which the exception occurs, following the guidance below.
+The lowest applicable severity number SHOULD be used.
+
+#### FATAL severity
+
+Exceptions that usually result in application shutdown SHOULD be recorded with
+severity `FATAL` (severity number 21).
```
cijothomas, line 113: if app is killed due to the `FATAL` exception, do we (sdk) live long enough to report it 🤣

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +47,108 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+
+Examples:
+
+- The application detects an invalid configuration at startup and shuts down.
+- The application encounters an out-of-memory condition.
```
cijothomas, line 118: Is it recommended to catch such OOM exception and report from with-in the process? Or did you mean something like "user configured something with a memory_size of 100 MB, but the app only has access to 5MB, so the app reports this event, and exits, without triggering OOM Exceptions"

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +47,108 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+reported with severity `WARN` (severity number 13).
+
+Semantic conventions that define `CLIENT` or `PRODUCER` spans SHOULD also define
+a corresponding exception event and recommend using `WARN` severity.
+
+Examples:
+
+- A connection attempt to a remote service times out.
+- Writing data to a file results in an I/O exception.
+- A remote service returns a 503 "Service Unavailable" response for 5 consecutive
```
cijothomas: is it the case that we SHOULD NOT use WARN when the 1st attempt failed, and wait until all retries are exhausted before doing WARN?

cijothomas: OTel Rust internaly uses DEBUG only at 1st failure, and does WARN when it runs out of retry attempts or the error is not-retry able.

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/retry.rs#L141
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/retry.rs#L167

Trying to make sure we are compliant with this recommendation!

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +47,108 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+[Emit a LogRecord](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/api.md#emit-a-logrecord)
+API, instrumentations SHOULD provide the exception instance rather than
+manually setting individual exception attributes.
+
+### Event name
+
+**Status**: [Development][DocumentStatus]
+
+It is RECOMMENDED to provide an [event name](/docs/general/events.md) that
+describes the instrumented operation with a `.exception` suffix.
```
cijothomas, line 92: nice.

docs/exceptions/exceptions-logs.md
```diff
@@ -40,18 +47,108 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+a corresponding exception event and recommend using `WARN` severity.
+
+Examples:
+
+- A connection attempt to a remote service times out.
+- Writing data to a file results in an I/O exception.
+- A remote service returns a 503 "Service Unavailable" response for 5 consecutive
+  attempts; retry attempts are exhausted and the operation fails.
+
+#### DEBUG severity
```
cijothomas, line 159: jump from WARN to DEBUG feels abrupt, as we skipped INFO. (I understand it is a never ending discussion likely to document when to use INFO!)

docs/exceptions/exceptions-logs.md
pellared: Note that shutdown may not be a crash. it may be a graceful shutdown that waits o logger provider shutdown being completed.
But even in case of a crash some "instrumentations" may be able to still report it: e.g. OBI :wink:

docs/exceptions/exceptions-logs.md
pellared: We discussed this today. Do you see a use case when an error/exception should have info level? Do you think we should clearly document it?

docs/exceptions/exceptions-logs.md
cijothomas: yea the examples below clarifies it!

docs/exceptions/exceptions-logs.md
pellared: I agree that this should be clarified.

docs/exceptions/exceptions-logs.md
cijothomas: https://github.com/open-telemetry/semantic-conventions/pull/3311#discussion_r2849211671
it is debatable if the Exception reporting from http call should be INFO or WARN, if its just a 1st attempt of many.

I don't have strong opinions on whether it should be INFO or DEBUG, but either way - lets make it clear in the doc,

trask: > I understand it is a never ending discussion likely to document when to use INFO!

indeed, that's a big part of why we skipped it 😄

> I don't have strong opinions on whether it should be INFO or DEBUG, but either way - lets make it clear in the doc

is your concern just that a reader might think that we forgot to include INFO?

docs/exceptions/exceptions-logs.md
trask: My reading of this is that it's an example where you would emit a (one) exception event with severity WARN.

docs/exceptions/exceptions-logs.md
hilmarf: How should logging bridges then react on Exceptions, that are originally logged on INFO level? Do they have to map them to WARN? Or to DEBUG? How should the bridge author know what's right? I believe only the original developer who decided to catch an exception and logged it on INFO would know that, right? For me it doesn't make sense to skip one level. Please support all official severities.

docs/exceptions/exceptions-logs.md
pellared: > My reading of this is that it's an example where you would emit a (one) exception event with severity WARN.

Probably this was the author's intention but this is not clearly described. One could understand that each attempt is logged with warn.

docs/exceptions/exceptions-logs.md
cijothomas: > is your concern just that a reader might think that we forgot to include INFO?

I think yes. We should make it clear that it is **intentional** to use DEBUG severity for the exception here.

cijothomas: > How should logging bridges then react on Exceptions, that are originally logged on INFO level?

Pass-through. This is semantic convention, which is prescribing how the original logger use severity. Logging-bridges should not make any "conversions". - This is my understanding.

docs/exceptions/exceptions-logs.md
lmolkova: added 

> [!NOTE]
> Instrumentation SHOULD do the best effort to record such errors, but
> OpenTelemetry SDK and exporters might not have a chance to actually export it

in https://github.com/open-telemetry/semantic-conventions/pull/3311/changes/a86e9298019647d2496bf2c417d6cb9a1e8cd83b

docs/exceptions/exceptions-logs.md
lmolkova: no, it's "if you record it, it should be fatal"

added

> [!NOTE]
> This document provides guidance to OpenTelemetry instrumentations that record
> exceptions and semantic conventions authors that define exception events.
> It does not apply to logging bridges.
> It prescribes how exception events should be recorded but does not prescribe
> when or if to record them.


in https://github.com/open-telemetry/semantic-conventions/pull/3311/changes/a86e9298019647d2496bf2c417d6cb9a1e8cd83b

docs/exceptions/exceptions-logs.md
lmolkova: addressed in https://github.com/open-telemetry/semantic-conventions/pull/3311/changes/a86e9298019647d2496bf2c417d6cb9a1e8cd83b

docs/exceptions/exceptions-logs.md
lmolkova: @cijothomas are you concerned with lack of INFO ? If you feel strong about it, I can add a section saying that we don't have a use-case for it (at least yet), but I think it's redundant and not necessary.

docs/exceptions/exceptions-logs.md
```diff
@@ -91,6 +98,10 @@ Examples:
 - The application detects an invalid configuration at startup and shuts down.
 - The application encounters an out-of-memory condition.
 
+> [!NOTE]
+> Instrumentation SHOULD do the best effort to record such errors, but
+> OpenTelemetry SDK and exporters might not have a chance to actually export it
```
pellared:
```suggestion
> OpenTelemetry SDK and exporters might not have a chance to actually export it.
```

docs/exceptions/exceptions-logs.md
```diff
@@ -38,20 +45,123 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
-## Attributes
+Exceptions that are unhandled by application code and don't result in application
+shutdown SHOULD be recorded with severity `ERROR` (severity number 17).
+
+Semantic conventions that define `SERVER` or `CONSUMER` spans SHOULD also define
+a corresponding exception event and recommend using `ERROR` severity.
+
+Examples:
+
+- A background job terminates with an exception.
```
lmolkova: maybe replace with messaging consumer

docs/exceptions/exceptions-logs.md
```diff
@@ -38,20 +45,123 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+`DEBUG` (severity number 9).
+
+For example, an exception indicating that a request was cancelled on the client
+side is thrown on the server and detected by the server instrumentation.
+
+#### TRACE severity
+
+Some libraries and frameworks generate artificial exceptions for operations that
+end with an unsuccessful error code. When possible, instrumentations SHOULD NOT
+record these artificial exceptions, or if recorded, SHOULD set the severity to
```
lmolkova: consider: semconv call: scala uses exception as normal mechanism, not emit this at all

docs/exceptions/exceptions-logs.md
```diff
@@ -13,7 +13,13 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
 <!-- toc -->
 
 - [Recording an exception](#recording-an-exception)
-- [Attributes](#attributes)
+  - [Event name](#event-name)
```
lmolkova, line 17:
```suggestion
  - [When not to record exceptions](#when-not-to-record-exceptions)
  - [Event name](#event-name)
```

docs/exceptions/exceptions-logs.md
```diff
@@ -38,20 +45,124 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+### Severity
+
+**Status**: [Development][DocumentStatus]
+
+The severity reflects the expected impact of the exception, not just its presence.
+
+[Severity Number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-severitynumber)
+SHOULD be provided on all exception events and SHOULD be set based on the context
+in which the exception occurs, following the guidance below.
+The lowest applicable severity number SHOULD be used.
```
pellared: TBH, I am not sure if this is needed. I feel that this may bring more confusion than clarity. Probably worth double-checking with other reviewers.

```suggestion
```

docs/exceptions/exceptions-logs.md
```diff
@@ -38,20 +45,124 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+- Writing data to a file results in an I/O exception.
+- A client library call fails after exhausting retries because the underlying
+  service is unavailable. The client library instrumentation records a single
+  WARN log; logging of individual retry attempts is left to the lower-level
+  instrumentation.
+
+#### DEBUG severity
+
+Exceptions that don't indicate an actual issue SHOULD be recorded with severity
+`DEBUG` (severity number 9).
```
pellared: fixing severity number for `DEBUG` and also I want to explicitly allow using `TRACE`. Maybe in future we are going to be able to have some better distinction between DEBUG and TRACE.

```suggestion
`DEBUG` (severity number 5) or even `TRACE` (severity number 1).
```

docs/exceptions/exceptions-logs.md
```diff
@@ -38,20 +45,124 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+code. Similarly, Spring in Java provides [`ResponseStatusException`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/server/ResponseStatusException.html).
+
+### Event name
+
+**Status**: [Development][DocumentStatus]
+
+It is RECOMMENDED to provide an [event name](/docs/general/events.md) that
+describes the instrumented operation with a `.exception` suffix.
+
+It is NOT RECOMMENDED to use a generic name such as `exception`.
```
lmolkova: update based on logs sig discussion - this is a fallback

docs/exceptions/exceptions-logs.md
```diff
@@ -13,7 +13,14 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
 
 - [Recording an exception](#recording-an-exception)
-- [Attributes](#attributes)
+  - [When not to record exceptions](#when-not-to-record-exceptions)
+  - [Event name](#event-name)
+  - [Severity](#severity)
+    - [FATAL severity](#fatal-severity)
+    - [ERROR severity](#error-severity)
+    - [WARN severity](#warn-severity)
+    - [DEBUG severity](#debug-severity)
```
lmolkova, line 23: TODO: add info and trace sections saying they are for app developers, not for semconv

docs/exceptions/exceptions-logs.md
```diff
@@ -38,20 +45,124 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
 
 ## Recording an exception
 
+> [!NOTE]
```
lmolkova, line 50: emphasize it better, maybe split into two sections 0- one for semconv authors and as a best practice, one for everybody with formal deifnitions

docs/exceptions/exceptions-logs.md
lmolkova: add info section in https://github.com/open-telemetry/semantic-conventions/pull/3311/changes/cc641b20e46b74e332dd5a0c8fb4c41a366e5655

> Application developers may use `INFO` severity (number 9) to record exceptions.

docs/exceptions/exceptions-logs.md
lmolkova: updated based on the SIG discussion to 

> #### TRACE severity
> 
> Application developers may use `TRACE` severity (number 1) to record exceptions.
> Semantic conventions authors SHOULD NOT define exception events with `TRACE`
> severity.

https://github.com/open-telemetry/semantic-conventions/pull/3311/changes/cc641b20e46b74e332dd5a0c8fb4c41a366e5655

docs/exceptions/exceptions-logs.md
```diff
@@ -38,20 +47,136 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+Instrumentations that are not specific to a particular operation or domain, such as
+a global unhandled exception handler, SHOULD use the `exception` event name.
+
+### Severity
+
+**Status**: [Development][DocumentStatus]
+
+The severity reflects the expected impact of the exception, not just its presence.
+
+[Severity Number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-severitynumber)
```
trask:
```suggestion
[Severity Number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.55.0/specification/logs/data-model.md#field-severitynumber)
```

docs/exceptions/exceptions-logs.md
```diff
@@ -38,20 +47,139 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+This document does not apply to logging bridges.
+
 Exceptions SHOULD be recorded as attributes on the
 [LogRecord](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.55.0/specification/logs/data-model.md#log-and-event-record-definition) passed to the [Logger](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.55.0/specification/logs/api.md#logger) emit
-operations. Exceptions MAY be recorded on "logs" or "events" depending on the
-context.
+operations.
+
+Exception events emitted by instrumentations that also record spans for the same
+operation MUST be associated with the corresponding span context.
```
lmolkova, line 66: @trask does this address  your comment https://github.com/open-telemetry/semantic-conventions/pull/3311#pullrequestreview-3918961518 ?

lmk if you want to bikeshed on MUST :)

trask: 👍

docs/exceptions/exceptions-logs.md
```diff
@@ -38,20 +47,139 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
-the language runtime.
+Examples:
 
-## Attributes
+- The application detects an invalid configuration at startup and shuts down.
+- The application encounters an out-of-memory condition.
+
+> [!NOTE]
+> Instrumentation SHOULD do the best effort to record such errors, but
+> OpenTelemetry SDK and exporters might not have a chance to actually export it.
```
lmolkova:
```suggestion
> OpenTelemetry SDK and exporters might not have a chance to actually export them.
```

docs/exceptions/exceptions-logs.md
```diff
@@ -38,20 +47,139 @@ emitted through the [Logger API](https://github.com/open-telemetry/opentelemetry
...
+**Status**: [Development][DocumentStatus]
+
+It is RECOMMENDED to provide an [event name](/docs/general/events.md) that
+describes the instrumented operation with a `.exception` suffix.
+
+For example, [`http.client.request.exception`](/docs/http/http-exceptions.md#http-client-request-exception)
+represents exceptions that occur during an HTTP client request.
+
+Instrumentations that are not specific to a particular operation or domain, such as
+a global unhandled exception handler, SHOULD use the `exception` event name.
```
lmolkova:
```suggestion
global unhandled exception handlers, SHOULD use the `exception` event name.
```

PR #3315, opened 2026-01-21
Update gen-ai-events.md - [chore] Fix typo in tool_call_response id example

robk777:
Removes leading space in tool_call_response id field in JSON example.

The id had `" call_..."` instead of `"call_..."` — the space would cause a mismatch with the original tool call id.


lmolkova:
Thanks @robk777 !

This file is auto-generated, please update the source - https://github.com/open-telemetry/semantic-conventions/blob/d77381ec96a6e1702905349370db09fdc0f8930e/model/gen-ai/registry.yaml#L564

You'd also need to run [`make table-generation registry-generation`](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#2-update-the-markdown-files) to update auto-generated content.

Thank you!

robk777:
Thanks @lmolkova! Got it, I'll fix the source in model/gen-ai/registry.yaml and run the table generation. Just waiting on CLA approval from my company (on hold due to an acquisition). Will push once that's sorted.

PR #3317, opened 2026-01-21
RPC: Clarify how to populate `server.address` and `server.port` when using client-side load balancing or service discovery

### LINKED ISSUE #3258: RPC: record url.scheme for grpc conventions to capture target fully
see https://grpc.io/docs/guides/custom-name-resolution/

lmolkova:
Fixes #3258


.chloggen/grpc-server-port-url.yaml
```diff
@@ -0,0 +1,6 @@
+change_type: enhancement
+component: rpc
+note: >
+  Clarify how to parse `server.address` and `server.port` from gRPC target URL.
+  Reference `url.scheme` on gRPC client span.
+issues: [3258, TODO]
```
lmolkova:
```suggestion
issues: [3258, 3317]
```

trask (approved):

model/rpc/spans.yaml
```diff
@@ -176,7 +176,54 @@ groups:
...
+          The address SHOULD be parsed from the target URL according to the
+          [gRPC Name Resolution specification](https://grpc.github.io/grpc/core/md_doc_naming.html) depending on the scheme used.
+
+          When the address is an IP address, instrumentations SHOULD NOT do a
+          reverse proxy lookup to obtain a DNS name and SHOULD set `server.address`
+          to the IP address provided.
+
+          If instrumentation cannot determine a single address from the target URL
+          (for example, when multiple IP addresses are provided),
+          it SHOULD NOT set this attribute.
```
lmolkova: another example (also applicable to dubbo) - https://github.com/makdharma/grpc-zookeeper-lb/blob/5eb44099baa9126928ccf9c649ad66bdeb5a768f/examples/src/main/java/io/grpc/examples/helloworld/HelloWorldClient.java#L253

E.g. url is `zk://zookeeper-host:2181`, when client/channel is created, client gets list of active instances from zookeper and uses client-side load balancing to work with them

Choice here:
- fallback to authority when server.address is not set (which could be zookeper or DNS server address). This is what happens today in Java gRPC instrumentation
- don't set `server.address | port`.

model/rpc/spans.yaml
```diff
@@ -176,7 +176,63 @@ groups:
...
+            - `url.scheme`: not set
+          - Given the target string `xds:///wallet.grpcwallet.io`, expected attributes:
+            - `server.address`: `wallet.grpcwallet.io`
+            - `server.port`: not set
+            - `url.scheme`: `xds`
+          - Given the target string `zk://zookeeper:2181`, expected attributes:
+            - `server.address`: `zookeeper`
+            - `server.port`: `2181`
+            - `url.scheme`: `zk`
+          - Given the target string `ipv4:198.51.100.123:50051,198.51.100.124:50051`, expected attributes:
```
lmolkova, line 227: - explore grpc.target idea
- check if dubbo has this case

matt-hensley (dismissed):

model/rpc/spans.yaml
```diff
@@ -176,7 +176,63 @@ groups:
...
+
+          - Given the target string `grpc.io::50051`, expected attributes:
+            - `server.address`: `grpc.io`
+            - `server.port`: `50051`
+            - `url.scheme`: not set
+          - Given the target string `xds:///wallet.grpcwallet.io`, expected attributes:
+            - `server.address`: `wallet.grpcwallet.io`
+            - `server.port`: not set
+            - `url.scheme`: `xds`
+          - Given the target string `zk://zookeeper:2181`, expected attributes:
```
steverao: This refers to the Span corresponding to the interaction between registration centers? In a client-to-server scenario, I understand that the target address cannot be the registration center's domain name.

model/rpc/spans.yaml
```diff
@@ -176,7 +176,59 @@ groups:
...
+          reverse proxy lookup to obtain a DNS name and SHOULD set `server.address`
+          to the IP address provided.
+
+          If the instrumentation cannot determine a single address from the target string
+          (for example, when multiple IP addresses are provided in the target string),
+          the instrumentation SHOULD NOT set this attribute.
+
+          Examples:
+
+          - Given the target string `grpc.io::50051`, expected attributes:
```
lmolkova:
```suggestion
          - Given the target string `grpc.io:50051`, expected attributes:
```

model/rpc/spans.yaml
```diff
@@ -176,7 +176,59 @@ groups:
...
+            - `server.address`: `"grpc.io"`
+            - `server.port`: `50051`
+            - `grpc.target`: not set
+          - Given the target string `xds:///wallet.grpcwallet.io`, expected attributes:
+            - `server.address`: `"wallet.grpcwallet.io"`
+            - `server.port`: not set
+            - `grpc.target`: `"xds:///wallet.grpcwallet.io"`
+          - Given the target string `zk://zookeeper:2181`, expected attributes:
```
lmolkova: todo: let's not populate server.address | port for zookeeper and other load-balancing/service discovery cases where we don't know if authority is a real endpoint

model/grpc/registry.yaml
```diff
@@ -0,0 +1,13 @@
...
+  display_name: gRPC Attributes
+  brief: gRPC-specific attributes
+  attributes:
+  - id: grpc.target
+    type: string
+    stability: development
+    examples: ["dns:///grpc.io:50051", "unix:///run/containerd/containerd.sock", "ipv4:198.51.100.123:50051"]
+    brief: >
+      The [gRPC target string](https://grpc.io/docs/guides/custom-name-resolution/)
+      specified when creating the gRPC channel.
```
trask: do we still want `grpc.target`?

lmolkova: great point! https://github.com/open-telemetry/semantic-conventions/pull/3317/changes/3b2c8c347743317da3d3154d943aeff22cc65de7

I think it still might be useful in corner cases. It could be added later even on metrics as opt in (it'd be a good default given that its value is limited).

trask (dismissed):

model/rpc/spans.yaml
```diff
@@ -176,7 +186,51 @@ groups:
...
+            - `server.address`: `"grpc.io"`
+            - `server.port`: `50051`
+          - Given the target string `dns://1.2.3.4/grpc.io:50051`, expected attributes:
+            - `server.address`: `"grpc.io"`
+            - `server.port`: `50051`
+          - Given the target string `unix:///run/containerd/containerd.sock`, expected attributes:
+            - `server.address`: `"/run/containerd/containerd.sock"`
+            - `server.port`: not set
+          - Given the target string `zk://zookeeper:2181/my-server`, expected attributes:
+            - `server.address`: "zk://zookeeper:2181/my-server"
```
steverao:
```suggestion
            - `server.address`: `"zk://zookeeper:2181/my-server"`
```

model/rpc/spans.yaml
```diff
@@ -176,7 +186,51 @@ groups:
...
+            - `server.port`: not set
+          - Given the target string `ipv4:198.51.100.123:50051,198.51.100.124:50051`, expected attributes:
+            - `server.address`: "ipv4:198.51.100.123:50051,198.51.100.124:50051"
```
steverao:
```suggestion
            - `server.address`: `"ipv4:198.51.100.123:50051,198.51.100.124:50051"`
```

trask:
Prototype for deriving `server.address` and `server.port` from grpc target: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/16161

model/rpc/spans.yaml
lmolkova: I think this can be resolved now that we use service registry address / path in server.address

trask (approved):

PR #3337, opened 2026-01-26
Add openai.api.type attribute to distinguish OpenAI API types

singankit:
Add openai.api.type attribute to distinguish OpenAI API types

Fixes #

## Changes

> [!IMPORTANT]


model/gen-ai/registry.yaml
```diff
@@ -386,6 +386,10 @@ groups:
               value: "execute_tool"
               brief: 'Execute a tool'
               stability: development
+            - id: responses
```
lmolkova: I don't think it can be an operation name - there are several operations in responses API: 

Image

I think a path forward might be an openai specific attribute that records an API type/version

singankit: My opinion is that `chat completion` and `responses` have very similar operations. [Reponses Migration guide](https://platform.openai.com/docs/guides/migrate-to-responses?update-item-definitions=chat-completions) also talks about the similarity of features between the two.

If `chat-completion` is available as an operation in Gen AI Sem Conv can `responses` be an operation too ?

image

singankit: Adding this screenshot from slack `# otel-genai-instrumentation` channel for context.
 
image

stephentoub: I don't think responses should be added in this way. It's the name of one implementation/protocol of a "chat" API. Each provider has one or more. From my perspective, the operation should continue being chat, and if the name of the endpoint needs to be tracked, it can be done in a separate tag.

model/gen-ai/registry.yaml
```diff
@@ -350,6 +350,14 @@ groups:
           The `gen_ai.data_source.id` SHOULD match the identifier used by the GenAI system rather than a name specific to the external storage, such as a database or
           object store. Semantic conventions referencing `gen_ai.data_source.id` MAY also leverage additional attributes, such as `db.*`, to further identify and describe the data source.
         examples: ['H7STPQYOND']
+      - id: gen_ai.api.type
```
lmolkova: I think it should be OpenAI specific, unless we have other cases when there are multiple types of the same API. 

```suggestion
      - id: openai.api.type
```

Also, even if there are multiple versions, we might find other means to record it for other systems (e.g. `anthropic.api.version = 1 | 2` or `gcp.api.protocol = grpc | rest`). we actually already do for GCP with `gen_ai.system.name = gcp.vertexai | gcp.gemini | gcp.gen_ai`.

model/gen-ai/registry.yaml
```diff
@@ -350,6 +350,14 @@ groups:
...
+        stability: development
+        type: string
+        brief: The API type or version used for the GenAI operation.
```
lmolkova: I think this should not include version. If you look into OpenAI API they could change version independently of type `https://api.openai.com/v1/responses`

model/gen-ai/registry.yaml
```diff
@@ -350,6 +350,14 @@ groups:
...
+        note: >
+          This attribute is used to distinguish between different API versions or types provided by the same provider
+          for similar operations. For example, OpenAI provides both "chat.completions" and "responses" APIs for chat operations.
+        examples: ['chat.completions', 'responses', 'completions']
```
lmolkova: please define strict enum for OpenAI attribute

docs/gen-ai/gen-ai-agent-spans.md
```diff
@@ -141,6 +141,7 @@ and SHOULD be provided **at span creation time** (if provided at all):
 | `execute_tool` | Execute a tool | ![Development](https://img.shields.io/badge/-development-blue) |
 | `generate_content` | Multimodal content generation operation such as [Gemini Generate Content](https://ai.google.dev/api/generate-content) | ![Development](https://img.shields.io/badge/-development-blue) |
 | `invoke_agent` | Invoke GenAI agent | ![Development](https://img.shields.io/badge/-development-blue) |
+| `responses` | Responses API operation such as [OpenAI Responses API](https://platform.openai.com/docs/api-reference/responses) | ![Development](https://img.shields.io/badge/-development-blue) |
```
lmolkova: probably need to regenerate this file

model/gen-ai/registry.yaml
lmolkova: could you please also update PR title and description accordingly?

model/openai/registry.yaml
```diff
@@ -19,6 +19,22 @@ groups:
               stability: development
         brief: The service tier requested. May be a specific tier, default, or auto.
         examples: ['auto', 'default']
+      - id: openai.api.type
```
lmolkova, line 22: we should reference this attribute on the openai client spans

model/openai/registry.yaml
```diff
@@ -19,6 +19,22 @@ groups:
...
+              brief: The OpenAI Chat Completions API.
+              stability: development
+            - id: responses
+              value: "responses"
+              brief: The OpenAI Responses API.
+              stability: development
+        brief: The type of OpenAI API being used.
+        note: >
+          This attribute is used to distinguish between different API types provided by OpenAI
+          for similar operations. For example, OpenAI provides both Chat Completions and Responses APIs for chat operations.
```
lmolkova:
```suggestion
        brief: The type of OpenAI API being used.
```

does it provide any information that's not available in the brief?

model/openai/registry.yaml
```diff
@@ -19,6 +19,22 @@ groups:
               stability: development
         brief: The service tier requested. May be a specific tier, default, or auto.
         examples: ['auto', 'default']
+      - id: openai.api.type
+        stability: development
+        type:
+          members:
+            - id: chat_completions
+              value: "chat_completions"
+              brief: The OpenAI Chat Completions API.
```
lmolkova:
```suggestion
              brief: The OpenAI [Chat Completions API](https://developers.openai.com/api/reference/chat-completions/overview)
```

model/openai/registry.yaml
```diff
@@ -19,6 +19,22 @@ groups:
...
+              stability: development
+            - id: responses
+              value: "responses"
+              brief: The OpenAI Responses API.
```
lmolkova:
```suggestion
              brief: The OpenAI [Responses API](https://developers.openai.com/api/reference/responses/overview).
```

.chloggen/gen-ai-api-type.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: openai
+note: Add `openai.api.type` attribute to distinguish between different OpenAI API types for similar operations (e.g., Chat Completions vs Responses APIs).
```
lmolkova:
```suggestion
note: Add `openai.api.type` attribute to distinguish between different OpenAI API types for similar operations.
```

model/openai/registry.yaml
singankit: Thanks for the feedback have added it.

model/gen-ai/registry.yaml
singankit: Thanks for the feedback. I have updated this PR to capture API type information on openai specific attribute.

model/gen-ai/registry.yaml
singankit: Have update the PR and its description.

docs/gen-ai/gen-ai-agent-spans.md
singankit: Regenerated the docs, thanks for catching it.

lmolkova (approved):

Issue #2910, opened 2025-10-14
Should `deployment.environment.name` be an enum?


joaopgrassi:
### Area(s)

area:deployment

### What's missing?

The `deployment.environment.name` attribute is currently defined as a free-form string, allowing any value. However, this flexibility introduces a common challenge that many developers can relate to: inconsistent naming conventions for environments. For example, the production environment might be labeled as prod, production, live, or something else entirely.

This inconsistency poses a unique problem because this attribute often plays a critical role in enabling observability platforms and vendors to build features based on its value. The lack of standardization makes it difficult to develop and innovate such features effectively.

Recognizing the importance of this attribute, the Service and Deployment SIG has discussed the idea of transitioning its value from a free-form string to an enumerated type (enum). The enum would define a set of well-known and commonly used environment names, such as dev, prod, staging, test, etc. While we acknowledge that naming conventions can be subjective and opinions may vary, it aligns with the project’s intention to provide clear and opinionated guidance to the community.

Goals of This Issue

- Raise Awareness: Inform the community about this proposed change.
 - Collect Feedback: Gather ideas, opinions, and suggestions from the community.
 - Reach Consensus: Work towards agreeing on a set of standardized enum values for the most common environment names.

We understand that this may be a contentious topic, but we believe that standardizing this attribute will benefit the ecosystem by making it easier for vendors and developers to build robust, environment-aware feature

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

thompson-tomo:
So my suggestion would be to introduce a new strongly typed attribute `deployment.environment.tier` which is has options like proposed. 

This solution would allow `deployment.environment.name` to be set based upon https://man.linuxreviews.org/man5/machine-info.5.html and the description of the attribute in sem conv to be updated.

It would also allow the CI/CD sem conv to use `deployment.environment.*` namespace to define the environment which the deployment is occurring to rather than needing a seperate sub namespace. It also increases interoperability capabilities with CD Events which uses environment to describe "the platform which may run a service".

How would we support scenarios where there are multiple environments running of the same tier which could be for redundancy or customer isolation? 

[...]

PR #3339, opened 2026-01-27
Stabilize deployment environment

Arnav-Bansal:
## Changes

This PR introduce changes to stabilize the deployment.environment.name field.

> [!IMPORTANT]


jsuereth:
This needs a changelog and rebase

docs/registry/entities/deployment.md
```diff
@@ -5,21 +5,17 @@
...
 ## Deployment
 
-**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+**Status:** ![Stable](https://img.shields.io/badge/-stable-lightgreen)
 
 **type:** `deployment`
 
 **Description:** The software deployment.
 
-> [!warning]
```
Kielek, line 14: Is it intentional change? Role is still marked as _Other_

lmolkova (changes requested):
We need to specify the role of attributes on the entity - see https://github.com/open-telemetry/semantic-conventions/pull/3339/changes#r2743336364

thompson-tomo:
The key thing which was missed is that the actual stability of the entity wasn't updated but rather just the stability in the doc was changed. This is why the policy passed.

.chloggen/deployment-environment-name-stable.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+# The name of the area of concern in the attributes-registry, (e.g. http, cloud, db)
+component: deployment
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Stabilize `deployment.environment.name` attribute.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+# The values here must be integers.
+issues: []
```
lmolkova:
```suggestion
issues: [3339]
```

model/deployment/entities.yaml
```diff
@@ -1,10 +1,11 @@
...
   - id: entity.deployment
     type: entity
-    stability: development
+    stability: stable
     name: deployment
     brief: >
       The software deployment.
     attributes:
       - ref: deployment.environment.name
+        role: descriptive
```
lmolkova: Should it be descriptive? As a result deployment entity does not have any identity attributes.


 /cc @open-telemetry/semconv-service-and-deployment-approvers

model/deployment/entities.yaml
```diff
@@ -1,10 +1,11 @@
 groups:
   - id: entity.deployment
     type: entity
-    stability: development
+    stability: stable
     name: deployment
```
lmolkova: Also, is this right that this entity is called `deployment` and not `deployment.environment`? 

If it's `deployment`, should it contain any other existing deployment attrs such as `deployment.name | id`

thompson-tomo: Agree it makes more sense to me to have deployment & deployment.environment entities.

Arnav-Bansal: Updated the file to make it more apt.

model/deployment/entities.yaml
Arnav-Bansal: Done. Added deployment.name as idenitify attribute as well.

trask:
> We can't stabilize this, until we have decided on how to deal with the values of it: #2910.

Good call! I've added this to the Service and Deployment SIG agenda for this Thu.

model/deployment/entities.yaml
jsuereth: So I'm not sure what happened with this change, but let's do the following:

- Have a single entity called `deployment.environment` (i.e. rename `deployment` entity)
- That entity has its "identity" as `deployment.environment.name` attribute.

I'm not aware of any need for other entities, namespaces or attributes.

Issue #1076, opened 2024-05-26
Add a human-readable message as an attribute representing the event


alexvanboxel:
### Area(s)

area:event

### Is your change request related to a problem? Please describe.

Many logging tools have a convention where, upon finding a "message" or "msg" field in a structured log, they'll display its contents prominently for easier user analysis. Since we're establishing event definitions, we can include a human-readable representation of each event. Think of it like a span name in a trace: it carries no inherent meaning but aids user comprehension.

### Describe the solution you'd like

Include an `event.message` attribute that provides a clear, concise, and human-readable description of the event.

### Describe alternatives you've considered

Following the convention used in some logging tools, we could try to utilize a `message` field in the event body. However, this clashes with the principle that each event can freely define its own fields.

Issue #2064, opened 2025-04-08
Conventions around logging string templates


AbhiPrasad:
### Propose new conventions

For structured log messages you’d often have a string template and template parameters attached as attributes alongside the computed log body.

Although things like https://messagetemplates.org/ do exist, it is not feasible to add to certain languages or environments, and given most languages support c style substitution or similar forms of structured logging mechanisms, I want to propose adding some conventions around describing template strings and their parameters.

1. `log.record.template`: The string template. This is useful for users to search on so that they don't need to do complex queries on the log body. For example, `User %s has logged in!`
2. `log.record.template.parameter.<key>`: The parameters of the string template. The `<key>` can either be an integer, representing the index where the parameter belongs in the string template, or it can be a string, representing the string that gets substituted in the string template

```json
"attributes": [
  {
    "key": "log.record.template",
    "value": { "string_value": "User %s has logged in!" }
  },
  {
    "key": "log.record.template.parameter.0",
    "value": { "string_value": "John" }
  }
],
```

```json
"attributes": [
  {
    "key": "log.record.template",
    "value": { "string_value": "Hit rate limit for endpoint: {endpoint}" }
  },
  {
    "key": "log.record.template.parameter.endpoint",
    "value": { "string_value": "/api/users/" }
  }
],
```

The reason we need to support both integer indexing and string name substitution for `log.record.template.parameter` is that many languages can only do integer indexing as the variable name might get minified or aliased away (minified JS, minified JVM bytecode, etc.). String name substitution is vastly more useful though.

I am happy to help own pushing forward these conventions.

nblumhardt:
Some related discussion: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4400

nblumhardt:
Awesome to see this moving forward 😎

Given logs are usually pretty heavily machine-processed, how should a consumer of these attributes determine the syntax used by the template, for example, to display a formatted version, without requiring the whole thing to be expanded in the log record somewhere?

The spec-required resource attributes - `telemetry.sdk.name`, `telemetry.sdk.language` in particular - seem like they might be reasonable for doing this heuristically, but is there a more precise mechanism that would fit with the design of the conventions?

E.g. additional, optional attribute might do it:

```js
"attributes": [
  {
    "key": "log.record.template",
    "value": { "string_value": "User %s has logged in!" }
  },
  {
    "key": "log.record.template.parameter.0",
    "value": { "string_value": "John" }
  },
[...]

Issue #1283, opened 2024-07-25
Emitting log template + parameters?


trask:
### Area(s)

area:log

### Is your change request related to a problem? Please describe.

Java log messages are often emitted as a template (with placeholders) + parameters, e.g.

```
log.info("Message {}, {} and {}", param1, param2, param3)
```

Using template + parameters is not really an indication from the user that this is a "structured log", as it is typically done as a performance optimization to avoid concatenating the string when the logging level is suppressed already (and also there are no parameter names).

Whereas there are other Java logging APIs which are intended for structured logging and using them is an indication from the user that this is a "structured log", e.g.

```
log.info(MapMessage)
```

So in the OpenTelemetry Java logging appenders we think it makes sense in the first case to emit the "formatted" (interpolated) message in the log body (string), and in the second case to emit a "structured log body".

The first purpose of this issue is to see if others agree with our thinking.

The second purpose of this issue is to see how we could _optionally_ capture the log template and parameters in the first case (e.g. to support https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/11865).

Our initial proposal would be something like:

* `log.record.template`
* `log.record.parameters`

where `log.record.parameters` would be a string array, though another option could be `log.record.parameters.<key>` to more generally support named parameters (do other languages have these? and if there are named parameters is that a reason to capture as structured log?)

Thanks!

### Describe the solution you'd like

not sure

lmolkova:
my 2c (only confirming that there are a lot of unknows on how to map logs to otel):

- the parameters are not strings/primitives and need to be flattened if they are recorded in the attributes. The body seems like a better place for them
- arguably people want to query by parameters, in some languages parameters are named (e.g. .NET allows to do  `logger.LogInfo("{foo}", "bar"))` and they work best if `"foo"="bar"` is preserved
- AFAIK there are corner cases like slf4j structured logging where `logger.atInfo().addKeyValue("foo", "bar").log("{}", "baz")` and it's not clear where the context ("foo"="bar") should go (attributes? body?)

igor-suhorukov:
>Using template + parameters is not really an indication from the user that this is a "structured log", as it is typically done as a performance optimization to avoid concatenating the string when the logging level is suppressed already

This seems like a valid use case, but in case of legacy applications and libraries, access to logger parameters is the only possible way to analyze structured application information. In this case, each logger call with an array can be considered as a structured message with unnamed parameters. 

[...]

PR #3343, opened 2026-01-27
Allow event body to represent a display message

pellared:
Fixes https://github.com/open-telemetry/semantic-conventions/issues/1076 (proposes  to use an attribute "event.message" for this, not Body)

Related to https://github.com/open-telemetry/semantic-conventions/issues/1283 

Related to https://github.com/open-telemetry/semantic-conventions/issues/2064

FWIK some tools use Body for displaying the log/event record.

[OTel Rust](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs#L75-L76) and [OTel .NET](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/Serializer/ProtobufOtlpLogSerializer.cs#L215-L217) use body to represent a display message. In those are two languages nearly every LogRecord is an Event.

It is also worth mentioning that most (maybe even all) log bridges use `Body` as the log message.

Moreover it follows https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-body:

> Can be for example a human-readable string message (including multi-line) describing the event in a free form

### DIFF
--- a/docs/general/events.md
+++ b/docs/general/events.md
@@ -33,7 +33,7 @@ Semantic conventions that define events MUST document the event name and its att
 * Events SHOULD specify a [severity number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-severitynumber).
 
 * Events SHOULD NOT use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-body)
-  except of a string display message of the event.
+  except to represent a string display message of the event.
 
 ## External event compatibility
 


pellared:
CC @cijothomas @lmolkova @trask @alexvanboxel @thompson-tomo

docs/general/events.md
```diff
@@ -30,10 +30,12 @@ Semantic conventions that define events MUST document the event name and its att
 * [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/logs/data-model.md#field-attributes)
   SHOULD be used to represent details and provide additional context about the event.
 
-* Events SHOULD NOT use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/logs/data-model.md#field-body).
-
 * Events SHOULD specify a [severity number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/logs/data-model.md#field-severitynumber).
 
+* Events MAY use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/logs/data-model.md#field-body)
+  to represent the display message of the event,
+  which is RECOMMENDED to be of string type.
```
trask: one use case that had been discussed in early days was using the event body for large unstructured content, e.g. http request/response bodies, I think part of the thought was that backends may have larger limits for body compared to attributes

pellared: Maybe we should document explicitly that specific semantic conventions (e.g. HTTP) may have other recommendations for "body"`?

cijothomas:
@pellared nit: The linked issue https://github.com/open-telemetry/semantic-conventions/issues/1076 is proposing to use an attribute "event.message" for this, not Body. You may want that clearly stated in the PR description.

docs/general/events.md
pellared: @trask, WDYT?

```suggestion
* Events MAY use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/logs/data-model.md#field-body)
  to represent the display message of the event,
  which is RECOMMENDED to be of string type.
  Some (e.g. HTTP) semantic conventions MAY
  have their own specific rules on using body.
```

docs/general/events.md
```diff
@@ -30,10 +30,12 @@ Semantic conventions that define events MUST document the event name and its att
 * [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/logs/data-model.md#field-attributes)
   SHOULD be used to represent details and provide additional context about the event.
 
-* Events SHOULD NOT use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/logs/data-model.md#field-body).
-
 * Events SHOULD specify a [severity number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/logs/data-model.md#field-severitynumber).
 
+* Events MAY use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/logs/data-model.md#field-body)
```
cijothomas: IIUC, this is something [OTel Rust](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs#L75-L76) and [OTel .NET](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/Serializer/ProtobufOtlpLogSerializer.cs#L215-L217) does already. 
(Those are two languages where nearly every LogRecord is an Event)

pellared: Added to the PR description

docs/general/events.md
cijothomas: The suggested text looks good to me.

trask: 👍

lmolkova: I think we're losing the "don't use body if you can avoid it" rule of thumb here.
~Bikeshedding~ Suggesting

```suggestion
* Events MAY use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/logs/data-model.md#field-body) 
  to represent the display message of the event,
  which is RECOMMENDED to be of string type.
  Semantic conventions MAY also define the use of the body field for 
  purposes other than a display message, in cases where representing 
  the same information as attributes is problematic
```

(also instead of next paragraph on HTTP)

trask: yeah, I'm also having second thoughts about this. I brought it up in last week's semconv meeting, and I left the meeting thinking that we lose out on the strength of this proposal (event body is just a display message, aligns nicely with classic logs) if we also allow it to be other things

trask: Also I think aligns with getting rid of `error.message` (#3308)

trask: Discussed in Log SIG. Long-term vision is to send large objects out-of-band with some kind of ref attribute. This removes worry about not using body for large objects. So let's lean into event body (if present) SHOULD represent a display message.

cijothomas: https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs#L268-L269 
OTel Arrow also does what this PR is proposing - user friendly display message is stored as Body.

pellared: Removed the following

> Some (e.g. HTTP) semantic conventions MAY
  have their own specific rules on using body.

in https://github.com/open-telemetry/semantic-conventions/pull/3343/commits/d4c6008b10a408d51977de731eee9e0140e8a870

pellared: Do we want to also change

> Events MAY use [body]

to 

> Events SHOULD use [body]

?

The "(if present)" from https://github.com/open-telemetry/semantic-conventions/pull/3343/files#r2789512730 is a little strange. Thus I decided to leave it as MAY.

docs/general/events.md
lmolkova: super-nit - i think we updated spec links 

```suggestion
* Events MAY use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-body)
```

docs/general/events.md
trask: this seems weaker now by removing "Events SHOULD NOT use body"

I was thinking more along the lines of "Events SHOULD NOT use body, except as a string display message"

docs/general/events.md
pellared: Fixed in https://github.com/open-telemetry/semantic-conventions/pull/3343/commits/151d4a67faf9ae5250069a419c8a09ae01e77cab

docs/general/events.md
pellared: I like this.
Change applied https://github.com/open-telemetry/semantic-conventions/pull/3343/commits/c07676a0a13548d5630e50a1dafb989c2946e39e

nit: Maybe we should specify even  "MUST NOT" instead of "SHOULD NOT" ?
Or is too restrictive? 
Note that these are semantic conventions and not specification, therefore I think we can use MUST NOT.
We can make it as a separate PR if this is controversial

lmolkova: I think we should keep SHOULD NOT. The world won't end if someone puts something into the body

lmolkova: Also, it's weird to say MUST NOT and immediately provide an exception.

docs/general/events.md
```diff
@@ -30,10 +30,11 @@ Semantic conventions that define events MUST document the event name and its att
 * [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-attributes)
   SHOULD be used to represent details and provide additional context about the event.
 
-* Events SHOULD NOT use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-body).
-
 * Events SHOULD specify a [severity number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-severitynumber).
 
+* Events SHOULD NOT use [body](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/logs/data-model.md#field-body)
+  except of a string display message of the event.
```
trask:
```suggestion
  except to represent a string display message of the event.
```

trask (approved):

trask:
@cijothomas can you take another look after latest changes? thanks!

PR #3346, opened 2026-01-28
feat: Add x64 as a valid host.arch value

### LINKED ISSUE #3345: Inconsistency with host.arch: 'x64' vs 'amd64' in .NET
### Area(s)

area:host

### Propose new conventions

I am opening this issue based on a report in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/3810.

The current semantic conventions in model/host/registry.yaml specify amd64 for 64-bit x86 architecture and do not list x64 as a valid member.
However, in .NET, RuntimeInformation.OSArchitecture returns x64, leading to the OpenTelemetry.Resources.Host component returning "x64", which violates the spec.

I am raising this here to discuss potential resolutions:

Should x64 be added as a valid alias/value for host.arch in the semantic conventions?
Or should the SDKs strictly map platform-specific values (like "x64") to the canonical "amd64" defined in the spec?
The original reporter noted that Go, Java, and Python SDKs correctly return amd64, suggesting .NET is the outlier, but "x64" is a very common identifier on Windows systems.

### Tip

AguirreMoy:
Fixes #3345

This PR adds `x64` as a valid member of `host.arch` in `model/host/registry.yaml`.


thompson-tomo (changes requested):
This introduces duplicated enum values which would be blocked once #2448 is fixed. Key thing is value what is used see line 42. Either value and brief changed or instrumentation use defined value.

AguirreMoy:
Fixing this on dotnet-contrib to be specific for x64=>amd64 instead of introducing it to otel semantics.

https://github.com/open-telemetry/semantic-conventions/issues/3345

PR #3348, opened 2026-01-28
fix: make criticality attribute opt-in #3340

### LINKED ISSUE #3340: The service entity contains development attribute as recommended
### Area(s)

area:service

### What's the problem?

The service entity contains the [service.criticality](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/service.md) attribute however it has the development stability level. This shouldn't be allowed due to entity being stable. The exclusion which let it through is legacy from prior to the entity being split hence should've been removed.

The exclusion removed from the policy and the attribute becomes opt-in.

### Tip

thompson-tomo:
Fixes #3340 

## Changes

This removes the attribute stability exclusion on the service entity which is not relevant since entity was split. What emerged was that the criticality attribute was now conflicting. To resolve criticality is now opt-in but can be promoted at a later point it is stablised.

A changelog has been left out as the attribute is not included in any semconv release as it has just been added. 

> [!IMPORTANT]

### DIFF
--- a/docs/registry/entities/service.md
+++ b/docs/registry/entities/service.md
@@ -17,7 +17,7 @@
 | --- | --- | --- | --- | --- | --- | --- |
 | Identity | [`service.name`](/docs/registry/attributes/service.md) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | `Required` | string | Logical name of the service. [1] | `shoppingcart` |
 | Description | [`service.version`](/docs/registry/attributes/service.md) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | `Recommended` | string | The version string of the service component. The format is not defined by these conventions. | `2.0.0`; `a01dbef8a` |
-| Description | [`service.criticality`](/docs/registry/attributes/service.md) | ![Development](https://img.shields.io/badge/-development-blue) | `opt-in` | string | The operational criticality of the service. [2] | `critical`; `high`; `medium`; `low` |
+| Description | [`service.criticality`](/docs/registry/attributes/service.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Opt-In` | string | The operational criticality of the service. [2] | `critical`; `high`; `medium`; `low` |
 
 **[1] `service.name`:** MUST be the same for all instances of horizontally scaled services. If the value was not specified, SDKs MUST fallback to `unknown_service:` concatenated with [`process.executable.name`](process.md), e.g. `unknown_service:bash`. If `process.executable.name` is not available, the value MUST be set to `unknown_service`.
 
--- a/docs/resource/service.md
+++ b/docs/resource/service.md
@@ -64,7 +64,7 @@ between them. Additionally, there's a single database instance.
 | --- | --- | --- | --- | --- | --- | --- |
 | Identity | [`service.name`](/docs/registry/attributes/service.md) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | `Required` | string | Logical name of the service. [1] | `shoppingcart` |
 | Description | [`service.version`](/docs/registry/attributes/service.md) | ![Stable](https://img.shields.io/badge/-stable-lightgreen) | `Recommended` | string | The version string of the service component. The format is not defined by these conventions. | `2.0.0`; `a01dbef8a` |
-| Description | [`service.criticality`](/docs/registry/attributes/service.md) | ![Development](https://img.shields.io/badge/-development-blue) | `opt-in` | string | The operational criticality of the service. [2] | `critical`; `high`; `medium`; `low` |
+| Description | [`service.criticality`](/docs/registry/attributes/service.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Opt-In` | string | The operational criticality of the service. [2] | `critical`; `high`; `medium`; `low` |
 
 **[1] `service.name`:** MUST be the same for all instances of horizontally scaled services. If the value was not specified, SDKs MUST fallback to `unknown_service:` concatenated with [`process.executable.name`](process.md), e.g. `unknown_service:bash`. If `process.executable.name` is not available, the value MUST be set to `unknown_service`.
 


jsuereth:
So we had a bit of discussion on this in Service + Deployment Semconv SIG.

For the purpose of Entities, descriptive attributes are somewhat implicitly opt-in. There's a bit of a mismatch in terms of what requirement_level means, regarding inclusion of entity attributes. Technically *all* attributes are opt-in in practice via `OTEL_RESOURCE_ATTRIBUTES` today.

I think we should consider this implication in our rules / policies around Entity *descriptive* attributes and what requirement_level means here.  cc @lmolkova for thoughts.

lmolkova:
@jsuereth 

Let's assume there is a resource detector that can determine service criticality based on something like priority class. Should this detector populate service criticality by default? I think yes. This is what recommended level means.

I think @thompson-tomo brings up different point - having recommended attribute that's in development is not allowed on stable group. 

https://github.com/open-telemetry/semantic-conventions/blob/08de1bce7fa19dd7710d50b2a4649bb67357332e/docs/general/semantic-convention-groups.md?plain=1#L70-L78

We should update policies to test for it - not sure what went wrong there and we we didn't detect it.

In this sense, a stable resource detector should NOT report criticality by default, but MAY report it when user opts into it (or all experimental features - this is also opt-in). 

The entity specifics I see is:
- all id attributes are required or conditionally required
- all descriptive attributes are recommended or opt-in

thompson-tomo:
> having recommended attribute that's in development is not allowed on stable group.

@lmolkova that is exactly my point.

> We should update policies to test for it - not sure what went wrong there and we we didn't detect it.

Already done so and it has been included in this PR. The issue  was that an exclusion for this entity was defined so that `service.instance.id` could be on the `service` entity but when that attribute was split into it's own entity, the exclusion remained hence when criticality was introduced it was allowed to be added to the stable entity.

jsuereth:
SO the primary issue here is `service.name` was marked stable prior to the entity actually existing, let alone being stable and defined, IMO.  When we started creating entity rules, we had an exception for `service` because the *attribute had to be stable* but the group was not.  We did, IMO, a poor thing and marked the Entity stable without actually stabilziing the entities.

In #3162 - we split the `service` entity - NOT something we would do if it was truly stable.  Additionally, it was pure coincidence that the remaining attributes on `service` were stable.  *only* the identity (`service.name`) is considered stable by the SIG.  We probably should have kept the entity unstable, but there's various nuance here due to the nature of how this all rolled out.

The SIG would like criticality to be recommended and stabilize with a set of attributes (we're still figuring out service.owner).

TL;DR: I believe the exception still applies - I also think `service` the *entity* is not actually stable yet, but should be shortly.

If you'd like us to, instead, treat service *the entity*, for which we still have no meaningful way in code to interact with it, to be declared stable, we can go through the process of using `opt_in`, which we'll then have to forever keep.  

However, given there exists NO code or instrumentation (outside of prototypes) today that *can* interact with Service as an entity - I fully believe the exception applies.  There is no user that this change would break.  There is no code that would generate this attribute in a stable way, and more importantly, there is no way the *entity* (not the attribute) can be produced in OTEL instrumentation today.

lmolkova:
@jsuereth agree with the exception. I don't think we can downgrade the entity back to development either. I think it's a good reminder for us that nothing should be marked stable without some instrumentations actually using it.

jsuereth:
@lmolkova This is one reason I think we need to discuss this more deeply.  Everything @thompson-tomo did here was reasonable.  We're just in an unreasonable state right now.

For example - with Entity + specifically *service* some aspects are stable and have stable things which depend on those attributes.  However, the Entity itself is not, and the expansion from "raw attribute" to "entity" has not happened in the spec.  We want to somehow thread the needle of:

- Don't break things users depend on.
- Allow stability to be defined for things that are so ingrained into OTEL and won't change (e.g. meaning of `service.name` / `service.instance.id`)
- Allow the evolution of `service` the entity.  (We're currently discussing `service.owner` and whether this is a separate entity or multiple descriptive attributes on service, e.g.).


For context - I'm fine if we decided to `opt_in` new attributes on service entity and mark them `recommended` when stabilizing - but I think we're lacking a full set of policies around entities here, and we likely need to expand "group" stability policy for that.  To some extent - we need to design/plan what this is.  For now - I'm just asking to keep the exception for service until we sort this out.  I want to avoid putting folks reading through semconv into a bunch of "churn", that actually isn't manifest in any instrumentation.

trask:
SIG discussion: @jsuereth to take this to Entity SIG to iron out plan before we decide whether to merge or not

jsuereth:
An update here - 

Given the state of the world *and* current state of entities - Here's our path forward.

1. We'd like to keep the `service` exception for now, and treat `service` entity as 'special', i.e. somewhere between stable and  unstable.  Stable attributes *must* be preserved and kept stable. New attributes will have more flexibility.  When we "remove the exception" that will be the equivalent of declaring the `service` entity stable.
2. We will provide some updating policy guidance for Entities in general.   E.g. the inherit nature of descriptive attributes means we should actually *prevent* them from being marked required. We'll be contributing this into semconv.
3. We need to sort out a path for experimental additions to stable entities.  Our goal is something like the following:
   - Define a new experimental attribute.
   - Add it to a stable entity. Mark the *desired final state* when the attribute is stable. It *can* be recommended, conditionally recommended or opt_in.
   - *require* some kind of opt-in flag for the experimental attribute while it remains experimental on the entity.
   - Stabilizing the attribute means you can remove the opt-in flag.
   
This is a bit of a blend between what @thompson-tomo did here, but with recording the *intention* of the requirement-level on stability.

What do we all think?

cc @dyladan @dmitryax from Entities SIG.

thompson-tomo:
I agree documenting intention is key.

Would the following be a useful option to support point 3.

We could use an annotation property on the attribute ref to be that flag. When we generate the docs we render a config table after the attributes table similiar to below

> **Configuration**
> | Setting | Category | Sources | Value Type | Example Value | Attributes |
> | --- | --- | --- | --- | --- | --- |
> | resource.entities.experimental | Experimental opt-in | config file | string[] | service | service.criticality |

To complement that table I would add to the attribute table a column indicating if there is configuration options with values required and none.

That way we make it clear that user must opt-in via the corresponding config option. We could even potentially eliminate the flag and simply use the definition to calculate if the config option is needed. This would tie in nicely to declarative config.

jsuereth:
@thompson-tomo Yes - that's exactly where my head was going for this. If you wanted to put a quick proposal for what that looks like, we can tie that into the stbale-by-default efforts.  I'd link you proposal as a comment on the stable-by-default OTEP

thompson-tomo:
All done @jsuereth via #3396 in particular https://github.com/thompson-tomo/semantic-conventions/blob/feat/%233395_AddsConfigurationTables/docs/registry/entities/service.md.

Note I opted to include the other declarative config options for resources to avoid battling whitespace issues.

As such I will close this PR & associated issue.

PR #3364, opened 2026-01-31
Clarify observability for decision-only (non-executed) operations

Nick-heo-eg:
This PR adds documentation clarifying that non-executed, decision-only
operations are valid observability targets.

It does not introduce new signals or semantics, but documents how existing
OpenTelemetry concepts can be applied consistently to decision outcomes
such as denials, holds, or policy evaluations.

### DIFF
--- a/docs/decision-observability.md
+++ b/docs/decision-observability.md
@@ -1,0 +1,103 @@
+# Observability for Non-Executed (Decision-Only) Operations
+
+## Overview
+
+In many real-world systems, requests or operations are evaluated but
+intentionally **not executed**.
+
+Common examples include policy denials, safety gates, compliance holds,
+approval workflows, and quota enforcement. In these cases, the system
+reaches a valid outcome without performing downstream work.
+
+Today, such **decision-only outcomes** are often missing from traces or
+represented inconsistently, despite being critical for auditability,
+incident analysis, and explainability.
+
+This document clarifies that non-executed operations are valid
+observability targets and provides guidance on representing them using
+existing OpenTelemetry concepts.
+
+---
+
+## Why This Matters
+
+Non-executed outcomes are frequently more important than successful
+execution paths:
+
+- From an audit or compliance perspective, *why something did not run*
+  is often the primary question.
+- From an operational perspective, blocked or denied requests may
+  indicate misconfiguration, policy drift, or emerging incidents.
+- From an explainability perspective, silently dropped decisions reduce
+  trust and make post-incident analysis difficult.
+
+If these outcomes are not consistently observable, systems lose
+important decision context even though a meaningful outcome occurred.
+
+---
+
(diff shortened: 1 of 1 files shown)


Nick-heo-eg:
This is a documentation-only clarification.

Scope:
- No new semantic attributes
- No new signals or metrics
- No behavior change for existing instrumentation

Goal:
Clarify that decision-only / non-executed operations (e.g. deny, hold, policy evaluation)
are valid observability targets using existing OpenTelemetry concepts.

This is intended to help users apply semantic conventions consistently
without implying execution or side effects.

docs/decision-observability.md
```diff
@@ -0,0 +1,103 @@
...
+In many real-world systems, requests or operations are evaluated but
+intentionally **not executed**.
+
+Common examples include policy denials, safety gates, compliance holds,
+approval workflows, and quota enforcement. In these cases, the system
+reaches a valid outcome without performing downstream work.
+
+Today, such **decision-only outcomes** are often missing from traces or
+represented inconsistently, despite being critical for auditability,
+incident analysis, and explainability.
```
joaopgrassi, line 14: How did you gauge that this is true? If an app is instrumented, say with HTTP instrumentation, and a policy/permission prevents it from executing a DB query, this SHOULD be already traced, at least the incoming server request if not more.

docs/decision-observability.md
kamphaus, line 1: This file looks like it should be an issue to discuss any proposal for conventions responding to your need.

Since this file does not contain any conventions it also does not belong under `docs/` (other than maybe under `docs/non-normative/`).

Nick-heo-eg:
Thanks for the feedback — this is helpful, and I agree this wasn’t framed clearly enough.

A few clarifications on my side:

1. Regarding the point about “decision-only outcomes are missing”

You’re absolutely right that in many systems, the request itself is already observable via standard HTTP or server instrumentation, even if execution is prevented by policy. What I was trying to point at (and didn’t scope precisely enough) is a narrower gap I’ve seen in practice:

- The trace shows that a request arrived. 
- But it does not reliably encode that a decision was evaluated and intentionally stopped as a first-class, queryable signal.
- That distinction matters when the absence of execution is itself the expected and correct outcome (e.g. safety gates, approval workflows, AI guardrails). I agree the current wording reads too broadly and overstates the claim.

2. On applicability and process

I agree this should have started as an issue-level discussion rather than a PR, and that this document does not define concrete semantic conventions.

I’m happy to close this PR and reframe the topic as a discussion issue. I’ll only revisit this with a PR if there’s clear consensus that a convention gap actually exists.

3. On docs placement

Agreed as well this does not belong under docs/ in its current form. If anything, it would fit better as non-normative background material tied to an issue. 

Thanks again for the pushback it helped surface where my assumptions weren’t obvious or well supported.

PR #3368, opened 2026-02-02
pprof: add scope attributes for sample type ordering and defaults

aalexand:
## Changes

The change adds two new pprof-specific attributes to support the lossless conversion between pprof and OTel profile formats. These attributes address the handling of pprof's `default_sample_type` and the preservation of the original sample type order.

New attributes:
- `pprof.scope.default_sample_type`: Records the default_sample_type from the original pprof profile.
- `pprof.scope.sample_type_order`: Records the indices of the sample types in the original profile to allow exact reconstruction of the pprof sample type array.

The new attributes use the `pprof.scope.*` naming to clearly indicate they apply to the `ScopeProfiles` level. This is similar naming-wise with the existing `pprof.mapping.*` and `pprof.location.*` attributes while avoiding ambiguity with `pprof.profile.*` (which applies to individual Profile messages).

For more details see the original [design notes](https://docs.google.com/document/d/1CF_xg0AFBGyFhkL1hdSSaOXW4pLApg39PwYdcM0xVW8/edit?tab=t.0).


aalexand:
@open-telemetry/profiling-approvers @florianl FYI

model/pprof/registry.yaml
```diff
@@ -70,3 +70,24 @@ groups:
         examples:
           - "http://pprof.example.com/cpu-profile.html"
         stability: development
+      - id: pprof.scope.default_sample_type
+        brief: >
+          Records the pprof's default_sample_type in the original profile. Not set
+          if the default sample type was missing.
+        note: >
+          This attribute SHOULD be set at the scope level
+          (resource_profiles[].scope_profiles[].scope.attributes[]).
```
christos68k: Should we use MUST here as it's the only acceptable place for this attribute?

```suggestion
          This attribute, if present, MUST be set at the scope level
          (resource_profiles[].scope_profiles[].scope.attributes[]).
```

model/pprof/registry.yaml
```diff
@@ -70,3 +70,24 @@ groups:
...
+        type: string
+        examples:
+          - "cpu"
+        stability: development
+      - id: pprof.scope.sample_type_order
+        brief: >
+          Records the indexes of the sample types in the original profile.
+        note: >
+          This attribute SHOULD be set at the scope level
+          (resource_profiles[].scope_profiles[].scope.attributes[]).
```
christos68k:
```suggestion
          This attribute, if present, MUST be set at the scope level
          (resource_profiles[].scope_profiles[].scope.attributes[]).
```

PR #3387, opened 2026-02-03
[chore] Fallback to podman if docker not installed

### LINKED ISSUE #3386: Make commands fail if docker is not found
### Area(s)

tooling

### What's the problem?

When you try to setup the repo to run the various commands defined in `Makefile`, it assumes that the shell can run `docker --version` to detect whether you're using Podman or Docker. But this assume that the `docker` command can be executed, which is not true if a) you don't have Docker installed (and configured properly); and b) you installed Podman but didn't alias or simlink `docker` to `podman` in such away that the shell command can pick it up so that `docker --version` returns the expected output.

The simplest solution seems to be verifying that Docker is available by checking that `docker --version` runs correctly. I have a PR ready to make that happen.

Alternatively, I can see the setup instructions include something about needing Docker or Podman installed and setup in such away that `docker --version` can be executed by the makefile, along with instructions on the recommended way to do that on a vanilla setup without either Docker or Podman installed.

My preferred solution is the code change plus some documentation about needing Docker or Podman installed, which I can do if the code-solution is OK by y'all.

### Tip

bidetofevil:
Fixes #3386

## Changes

Checks whether the `docker` command can be executed and fallback to using `podman` if it's not.

The logic could've been written in a number of ways (including bailing early if neither `podman` nor `docker` can be executed) but I figured this is the most straightforward. Totally cool to structure the if/else and fallback logic differently or error out early as well depending on what you all prefer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,20 +43,26 @@ OPA_CONTAINER=$(OPA_CONTAINER_REPOSITORY)/$(VERSIONED_OPA_CONTAINER_NO_REPO)
 
 # Determine if "docker" is actually podman
 DOCKER_VERSION_OUTPUT := $(shell docker --version 2>&1)
-DOCKER_IS_PODMAN := $(shell echo $(DOCKER_VERSION_OUTPUT) | grep -c podman)
+PODMAN_REFERENCES := $(shell echo $(DOCKER_VERSION_OUTPUT) | grep -ic podman)
 
 ifneq ($(strip $(DOCKER_VERSION_OUTPUT)),)
-ifeq ($(DOCKER_IS_PODMAN),0)
+ifeq ($(PODMAN_REFERENCES),0)
     DOCKER_COMMAND := docker
 endif
 endif
 
-DOCKER_COMMAND ?= podman
+ifndef DOCKER_COMMAND
+    ifneq ($(strip $(shell podman --version 2>&1)),)
+        DOCKER_COMMAND := podman
+    endif
+else
+    $(info Neither docker nor podman can be executed. Did you install and configure one of them to be used?)
+endif
 
 # Debug printing
 ifdef DEBUG
 $(info Docker version output: $(DOCKER_VERSION_OUTPUT))
-$(info Is Docker actually Podman? $(DOCKER_IS_PODMAN))
+$(info Podman references in docker version output: $(PODMAN_REFERENCES))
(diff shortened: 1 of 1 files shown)


trask:
cc @michaelsafyan in case you have a chance to review

bidetofevil:
Strange. I'll have a look. This has to be something dumb on my part lol.

bidetofevil:
Figured it out. Will put up a fix shortly.

bidetofevil:
It's my unfamiliarity of the makefile syntax 🙃 

Anyway, kept the original logic and added a condition that the `docker --version` output isn't blank, and pulled the setting of the command to podman out to be defined if it wasn't done already. If this is considered ugly and not idiomatic, let me know and I'll redo it.

jsuereth:
@bidetofevil I'm fine with ugly in the makefile, it just needs to work / pass the builds :)

Makefile
```diff
@@ -45,11 +45,13 @@ OPA_CONTAINER=$(OPA_CONTAINER_REPOSITORY)/$(VERSIONED_OPA_CONTAINER_NO_REPO)
...
 
+ifneq ($(strip $(DOCKER_VERSION_OUTPUT)),)
 ifeq ($(DOCKER_IS_PODMAN),0)
     DOCKER_COMMAND := docker
-else
-    DOCKER_COMMAND := podman
 endif
+endif
+
+DOCKER_COMMAND ?= podman
```
jsuereth: Nit: Can you also check if `podman` is installed and issue an error message that should install one or the other?

I realize we just fail big today, so this is a convenience, but it would definitely help folks know they don't *need* podman if docker isn't installed, that docker should be ok.

bidetofevil: Sure. I was avoiding that to minimize the change but being clear about why we are failing is a good idea.

I also didn't like the fallback being podman tbh.

bidetofevil: Refactored this to look for podman only if we're not using docker, and show error message if neither appears to be executable. 

Checking to see if the command is defined before execution would be nice but I'm not confident I can do it in an idiomatic way without touching a lot of things and learning a lot more about makefile syntax 😅

bidetofevil: Not sure if I'm supposed to resolve the convo or reviewers should, but feel free to re-open it if the changes didn't satisfy this.

PR #3389, opened 2026-02-04
Add generate-all (markdown) target

lmolkova:
small usability improvement: adds target to do all md generation (tables, registry, areas) at once.

### DIFF
--- a/model/README.md
+++ b/model/README.md
@@ -45,7 +45,7 @@ Or, for convenience, run
 make generate-all
 ```
 
-that executes both targets along with other markdown generation targets that
+which executes both targets along with other markdown generation targets that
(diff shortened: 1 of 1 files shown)


model/README.md
```diff
@@ -23,24 +23,33 @@ A schema file for VS code is configured in the `/.vscode/settings.json` of this
...
+
+Or, for convenience, run
+
+```bash
+make generate-all
 ```
 
-For more information, see the [Weaver](https://github.com/open-telemetry/weaver)
-as our code generations tool.
+that executes both targets along with other markdown generation targets that
```
arminru:
```suggestion
which executes both targets along with other markdown generation targets that
```

PR #3396, opened 2026-02-05
feat: Add in configuration tables for entities.

### LINKED ISSUE #3395: Document configuration options for Entities
Following on from the discussion in, one of the suggestions by @jsuereth was to document the configuration options and to make clear the behaviour of experimental attributes on stable entities.

This issue is for introducing a configuration table on the entity pages as well as making it clear when configuration is needed.

thompson-tomo:
Fixes #3395

## Changes

This introduces configuration tables to the entity pages split based on env vs file. For configuration options where no common setting is defined, it is specified as TBA. Also a note is added above the attribute table indicating the explicit opt-in nature.

## Open questions

- Should we convert the configuration sections into collapsible blocks?
- Should we limit which attributes can be defined via configuration ie https://github.com/open-telemetry/weaver/issues/854

> [!IMPORTANT]


thompson-tomo:
@jack-berg i'm interested in your thoughts on this proposal to incorporate declarative config definitions into the semconv documentation for entities.

docs/resource/webengine.md
```diff
@@ -17,11 +17,33 @@
...
+| `resource.attributes.[].name` | `Conditionally Required` | User Defined Resource | string | `webengine.name` | [`webengine.name`](/docs/registry/attributes/.md) |
+| `resource.attributes.[].type` | `Conditionally Recommended` | User Defined Resource | string | `string` | [`webengine.name`](/docs/registry/attributes/.md) |
+| `resource.attributes.[].value` | `Conditionally Required` | User Defined Resource | string | `WildFly` | [`webengine.name`](/docs/registry/attributes/.md) |
+| `resource.attributes.[].name` | `Conditionally Required` | User Defined Resource | string | `webengine.version` | [`webengine.version`](/docs/registry/attributes/.md) |
+| `resource.attributes.[].type` | `Conditionally Recommended` | User Defined Resource | string | `string` | [`webengine.version`](/docs/registry/attributes/.md) |
+| `resource.attributes.[].value` | `Conditionally Required` | User Defined Resource | string | `21.0.0` | [`webengine.version`](/docs/registry/attributes/.md) |
```
jack-berg: If I understand this right, this is acting as a demonstration on how a user can configure these resource attributes in declarative config?

docs/resource/webengine.md
```diff
@@ -17,11 +17,33 @@
...
+| Other | [`webengine.version`](/docs/registry/attributes/webengine.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | string | The version of the web engine. | `21.0.0` | `Opt-In` |
+
+**Configuration File Options:**
+
+| Setting | Requirement Level | Category | Value Type | Example Value | Attribute |
+| --- | --- | --- | --- | --- | --- |
+| `resource.attributes.[].name` | `Conditionally Required` | User Defined Resource | string | `webengine.description` | [`webengine.description`](/docs/registry/attributes/.md) |
+| `resource.attributes.[].type` | `Conditionally Recommended` | User Defined Resource | string | `string` | [`webengine.description`](/docs/registry/attributes/.md) |
+| `resource.attributes.[].value` | `Conditionally Required` | User Defined Resource | string | `WildFly Full 21.0.0.Final (WildFly Core 13.0.1.Final) - 2.2.2.Final` | [`webengine.description`](/docs/registry/attributes/.md) |
```
jack-berg: Maybe there's a way to collapse these into a single row per attribute, with an example like:

```
- name: webengine.description
  value: WildFly Full 21.0.0.Final (WildFly Core 13.0.1.Final) - 2.2.2.Final
```

Note I omitted `type` since this is a string attribute and `type: string` is the default: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/opentelemetry_configuration.json#L175

jack-berg (commented):
Would also be cool to see semantic conventions get more opinionated about the names of the resource detectors responsible for adding resource attributes.

Explicitly setting resource attributes is fine, but cumbersome. Right now, we have a limited set of [spec-defined resource detectors](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#resource-detector-name). These are much more convenient for users:

```
resource:
  detection/development: # /development properties may not be supported in all SDKs
    detectors:
      - service: 
      - host:
      - process:
      - container:
```

If semantic conventions defined further resource detectors we would include them in the declarative config schema as well known types, and the UX would improve.

docs/resource/webengine.md
thompson-tomo: Hmm, i'll see if I can workout a way to generate an example snippet.

docs/resource/webengine.md
thompson-tomo: It's seeking to describe the configuration options applicable to the entity. Ideally this pattern would Scale-out to the other signals.

jsuereth:
I really like the direction of this.

High level -

- As @jack-berg points out, It'd be nice to have an `annotation` that denotes which resource detector in OTEL config would detect the entity and its attributes.
- While OTEL_RESOURCE_ATTRIBUTES overrides everything now, I wonder if we should allow an annotation to *override* whether or not this is shown?  As entities and infra around them stabilize, I would like to push harder on resource detectors and the new ENV variable that is entity aware.
- I was *also* thinking we define the feature-flag for `opt_in` attributes.

I don't have any specific nits in this review, as I think we should discuss some high level points.  Thanks for putting together this proposal!

thompson-tomo:
Thanks for the feedback @jack-berg / @jsuereth what I have done based on your feedback is:
- consolidated the entries for setting attribute values into a single setting
- added the settings to include opt-in to attributes
- added the settings to exclude recommended attributes
- added the settings to enable resource detectors based on entity name

Open points:
- limiting what attributes are shown as user definable,  

What I think would be nice is if we could link the setting name to the corresponding definition

docs/resource/webengine.md
thompson-tomo: @jack-berg all done including the skip of default properties

templates/registry/markdown/entity_macros.j2
```diff
@@ -48,14 +48,51 @@
...
+{% if ((e.stability == "stable") and (attribute.stability != "stable")) %}| `TBA` | `Required` | Experimental Opt-in | TBA | {{ attrs.name_with_link(attribute, attribute_registry_base_url, e.lineage_attributes) }} |{% endif %}{% if ((attribute.requirement_level == "opt_in")) %}
+| `resource.experimentalresourcedetection.attributes.included[]` | `Conditionally Required` | Attribute Inclusion | `{{attribute.name}}` | {{ attrs.name_with_link(attribute, attribute_registry_base_url, e.lineage_attributes) }} |{% endif %}{% if ((attribute.requirement_level != "opt_in")) %}
+| `resource.experimentalresourcedetection.attributes.excluded[]` | `Conditionally Required` | Attribute Exclusion | `{{attribute.name}}` | {{ attrs.name_with_link(attribute, attribute_registry_base_url, e.lineage_attributes) }} |{% endif %}
+| `resource.attributes[]` | `Opt-In` | User Defined Resource | `- name: {{attribute.name}},`{% if attribute.type != "string" %}<br>&nbsp;&nbsp;&nbsp;`type:{{ attrs.type(attribute) }},`{%endif%}<br>&nbsp;&nbsp;&nbsp;`value: {{examples.first(attribute) | trim }}` | {{ attrs.name_with_link(attribute, attribute_registry_base_url, e.lineage_attributes) }} |
```
thompson-tomo:
```suggestion
| `resource.attributes[]` | `Opt-In` | User Defined Resource | `- name: {{attribute.name}},`{% if attrs.type(attribute) != "string" %}<br>&nbsp;&nbsp;&nbsp;`type:{{ attrs.type(attribute) }},`{%endif%}<br>&nbsp;&nbsp;&nbsp;`value: {{examples.first(attribute) | trim }}` | {{ attrs.name_with_link(attribute, attribute_registry_base_url, e.lineage_attributes) }} |
```

jsuereth:
> What I think would be nice is if we could link the setting name to the corresponding definition

+1

A few more thoughts for discussions:

- I'd love if we can make it clear what "requirement-level" means vs. "Config requirement level".
  To me this may be easier if we have some "entity-level" config setting, e.g. we can tie to the Configuration
  spec on how the entity is detected.  (@jack-berg - Are collector resource-detectors in Config spec yet? Or would we want ot track those separately).
- For entities specifically, *everything* is overridable via `OTEL_RESOURCE_ATTRIBUTES`.  We may want to use blanked language here for now, vs. encoding a ton of YAML in our model we'd have to clean up as Resource Detection, Entities + Configuration SIGs make progress on various components to make this situation better.
- For `opt_in` fields - I really like the tie into config / settings.  I think we'll need some kind of "model" documentation on what annotations are allowed, what various fields mean, etc.
- We should think about policy enforcement over time here.  I'd like to make sure the "model" of what we put in YAML is something we could "enforce" non-breaking changes on.  I haven't had time to think through the details, but briefly - some attributes will *always* be opt-in.  We should not conflate opt-in flags of "this is an experimental feature that will stabilize" with "this is a stable feature that requires an opt-in".  So we may want to have rules like "stable config values can only include stable opt-in attributes" or some such.

So some strawman proposals:

- We break the rendering into three pieces in your entity config table:
  - A blanket statement on `OTEL_RESOURCE_ATTRIBUTES` being applicable way of specifying all attributes
  - A call-out to resource detection available on *per-entity-type* (vs attribute) basis
  - See if this means we only need expanded config options for `opt_in` with the caveat we can use the blanked call-out.
- In `docs/how-to-...` we update a section for how to tie `opt_in` arguments to a config parameter.

thompson-tomo:
> I'd love if we can make it clear what "requirement-level" means vs. "Config requirement level".

Agree and is an open to do task

> To me this may be easier if we have some "entity-level" config setting, e.g. we can tie to the Configuration spec on how the entity is detected.

I couldn't find anything to this effect

> For entities specifically, everything is overridable via OTEL_RESOURCE_ATTRIBUTES. We may want to use blanked language here for now, 

I have adjusted the jinja so that only attributes defined as user settable will now be shown in the config tables rather than all attributes.

> encoding a ton of YAML in our model we'd have to clean

Actually the yaml in the docs are being dynamically generated

> For opt_in fields - I really like the tie into config / settings. I think we'll need some kind of "model" documentation on what annotations are allowed, what various fields mean, etc.

Agree on documentation, note that jinja is dynamically determining what settings are applicable and not using annotations.
 
> We should not conflate opt-in flags of "this is an experimental feature that will stabilize" with "this is a stable feature that requires an opt-in".

I think that this is already covered, if a stable entity has non stable attributes then the experimental flag setting needs to be set which is seperate to the requirement opt-in setting.

> A blanket statement on OTEL_RESOURCE_ATTRIBUTES being applicable way of specifying all attributes

Agree and will handle this seperately to this pr

> A call-out to resource detection available on per-entity-type (vs attribute) basis
> See if this means we only need expanded config options for opt_in with the caveat we can use the blanked call-out.

I have reworked the docs to optimise the content and  there is the setting to enable the resource detector. @jack-berg do we extend declarative config to exclude entities from resource detector?

> In docs/how-to-... we update a section for how to tie opt_in arguments to a config parameter.

This is automated using jinja so nothing needs to be done.

Note I have raised https://github.com/open-telemetry/weaver/issues/1230 to discuss improving the ability to define resource detectors.

PR #3403, opened 2026-02-05
Define declarative configuration path for `sensitive_query_parameters`

trask:
Similar to
- #3394 

This property doesn't have any precedent for naming, so would be worth considering other options besides `sensitive_query_parameters`.

cc @open-telemetry/configuration-approvers

### DIFF
--- a/docs/db/elasticsearch.md
+++ b/docs/db/elasticsearch.md
@@ -97,10 +97,13 @@ value `REDACTED`:
 
 This list is subject to change over time.
 
+Matching of query parameter keys against the sensitive list SHOULD be case-sensitive.
+
 ![Development](https://img.shields.io/badge/-development-blue)
 Instrumentation MAY provide a way to override this list via declarative configuration.
 If so, it SHOULD use the `sensitive_query_parameters` property
-(an array of strings with minimum items 0) under `.instrumentation/development.general.sanitization.url`.
+(an array of case-sensitive strings with minimum items 0) under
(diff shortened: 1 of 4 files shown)


docs/db/elasticsearch.md
```diff
@@ -97,6 +97,13 @@ value `REDACTED`:
 
 This list is subject to change over time.
 
+![Development](https://img.shields.io/badge/-development-blue)
+HTTP instrumentation MAY provide a way to override this list via declarative configuration.
+If so, it SHOULD use the `sensitive_query_parameters` property
+(an array of strings with minimum items 0) under `.instrumentation/development.general.http`.
```
lmolkova: there are other instrumentation that may populate `url.full` or `url.query` (e.g. elasticsearch does) - do we expect each of these instrumentations to define the section?

Alternative:

```yaml
general:
  sanitization:
    url:
      sensitive_query_parameters: []
    enduser:  # imaginary future
       id_obfuscation_algo:
```

There will also be more sanitization options, it kind of makes sense to group them together under `sanitization` section and then by attribute key/namespace rather than by instrumentation.

It does not affect HTTP headers or methods much because they are HTTP-specific anyway

trask: I like it: https://github.com/open-telemetry/semantic-conventions/pull/3403/commits/a0290751bb9f37e552388a290a61cbfe2f3a1697

PR #3404, opened 2026-02-06
[editorial] Use anchor id attribute instead of name

chalin:
- Updates `<a name="abc">` to be `<a id="abc">`. The `name` attribute is essentially obsolete in this context.


lmolkova (approved):

kamphaus (approved):

kamphaus:
You might want to prefix the PR title with `[chore] ` to clear the changelog check.

KalleOlaviNiemitalo:
I wonder if it should then be `span` rather than `a`.

kamphaus:
> I wonder if it should then be `span` rather than `a`.

In html semantics an `<a>` tag with `id=` attribute is common as a target for links.
https://www.w3schools.com/jsref/dom_obj_anchor.asp

While other tags, like `<span>` with `id` attribute, can be used for this purpose, an empty `<a />` tag makes it explicit that this is used as a link target.

PR #3414, opened 2026-02-09
Define HTTP exception events

trask:
Part of migrating from recording exceptions as span events to recording them as logs.

Assigning event names to exceptions which correspond to existing semantic conventions provides an additional level of correlation between the signals, and makes it easy to view these specific sets of exceptions.

The idea would be to also define these for Database, RPC, Messaging (at least).

Prototype at https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/16049

### DIFF
--- a/docs/http/http-exceptions.md
+++ b/docs/http/http-exceptions.md
@@ -70,7 +70,6 @@ This event represents an exception that occurred during HTTP server request proc
 
 This event SHOULD be recorded when an exception occurs during HTTP server request processing.
 Instrumentations SHOULD set the severity to ERROR (severity number 17) when recording this event.
-Some HTTP server frameworks generate artificial exceptions for certain HTTP status codes (e.g., 404 Not Found). When possible, instrumentations SHOULD NOT record these artificial exceptions, or SHOULD set the severity to DEBUG (severity number 5).
 Instrumentations MAY provide a configuration option to populate exception events with the attributes captured on the corresponding HTTP server span.
 
 **Attributes:**
--- a/model/http/events.yaml
+++ b/model/http/events.yaml
@@ -40,10 +40,6 @@ groups:
 
       Instrumentations SHOULD set the severity to ERROR (severity number 17) when recording this event.
 
-      Some HTTP server frameworks generate artificial exceptions for certain HTTP status codes
-      (e.g., 404 Not Found). When possible, instrumentations SHOULD NOT record these artificial
-      exceptions, or SHOULD set the severity to DEBUG (severity number 5).
-
       Instrumentations MAY provide a configuration option to populate exception events
       with the attributes captured on the corresponding HTTP server span.
     attributes:


lmolkova (approved):
I believe we'll need to define default severity for these events (before stability) and probably add more attributes, but it can be done incrementally

model/http/events.yaml
kamphaus, line 1: These events do not use any http.* attributes.
Is this intentional?
Eg. I assume `http.response.status_code` would be reported on the related http span.

docs/http/http-exceptions.md
```diff
@@ -0,0 +1,100 @@
...
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `http.client.exception`.
+
+This event represents an exception that occurred during an HTTP client request.
+
+This event SHOULD be recorded when an exception occurs during HTTP client operations, such as network failures, timeouts, or other errors that prevent the request from completing successfully.
```
lmolkova: from logs sig meeting:
  
should we consider mentioning that some exceptions are artificial (e.g. for 404) and, when possible, should not be recorded (or perhaps recorded with different severity)

https://github.com/lmolkova/opentelemetry-specification/blob/0be3bc3fdee1a6bdcf129ee6ceb19ff02d070e62/oteps/4333-recording-exceptions-on-logs.md?plain=1#L140

?

docs/http/http-exceptions.md
```diff
@@ -0,0 +1,100 @@
...
+- [HTTP client exception](#http-client-exception)
+- [HTTP server exception](#http-server-exception)
+
+<!-- tocstop -->
+
+## HTTP client exception
+
+The `http.client.exception` event represents an exception that occurred during
+an HTTP client request, such as network failures, timeouts, or other errors that
+prevent the request from completing successfully.
```
lmolkova: nit, it's already in the auto-generated text below, do we need it?

model/http/events.yaml
trask: This is an interesting question.

They should be available via the correlated HTTP span.

But it could definitely be useful in some cases to have them on the exception (log) record, e.g. if the span is sampled out, to enable exception-to-metrics pipelines, or probably other use cases too.

docs/http/http-exceptions.md
```diff
@@ -0,0 +1,100 @@
...
+
+<!-- semconv event.http.client.exception -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `http.client.exception`.
```
lmolkova: should we call it `http.client.request.exception` ? 

For the same reason we call metric `http.client.request.duration`. We need to include the operation being instrumented (request) - in theory, there could be other operations / targets: `http.client.connection_pool.exception`, `http.client.configuration.exception`, `http.client.stream.exception`. I don't know if we want this granularity, but I can make a case that these events would have different sets of attributes.

docs/http/http-exceptions.md
trask: https://github.com/open-telemetry/semantic-conventions/pull/3414/commits/5e30a4a0f306a793cbe860d30cae5382da956675

docs/http/http-exceptions.md
trask: https://github.com/open-telemetry/semantic-conventions/pull/3414/commits/8632b46f315e2928871e62cf389e5e05b8f8672d

docs/http/http-exceptions.md
trask: https://github.com/open-telemetry/semantic-conventions/pull/3414/commits/5cc96405b5a7a8185f2eacad61d108c994e8139b

trask:
> I believe we'll need to define default severity for these events

👍

Added in https://github.com/open-telemetry/semantic-conventions/pull/3414/commits/5cc96405b5a7a8185f2eacad61d108c994e8139b

> and probably add more attributes

yeah, I'm torn on this (https://github.com/open-telemetry/semantic-conventions/pull/3414#discussion_r2784077102), did you have any preliminary thoughts?

docs/http/http-exceptions.md
```diff
@@ -0,0 +1,96 @@
...
+<!-- prettier-ignore-start -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `http.client.request.exception`.
+
+This event represents an exception that occurred during an HTTP client request, such as network failures, timeouts, or other errors that prevent the request from completing successfully.
+
+This event SHOULD be recorded when an exception occurs during HTTP client operations.
+Instrumentations SHOULD set the severity to ERROR (severity number 17) when recording this event.
```
trask:
```suggestion
Instrumentations SHOULD set the severity to WARN (severity number 13?) when recording this event.
```

trask:
Discussed adding more attributes in Log SIG.

Sampling.

Adding attributes later wouldn't be breaking.

`error.type` is common across signals. In Java at least, in most cases `error.type` on exception would be same as `exception.type`.

Instrumentations MAY provide flag to copy all span attributes over to exception.

model/http/events.yaml
kamphaus: > Instrumentations MAY provide flag to copy all span attributes over to exception

I think that's a good idea.

To be included in the note text?

trask:
> Instrumentations MAY provide flag to copy all span attributes over to exception.

Added in https://github.com/open-telemetry/semantic-conventions/pull/3414/commits/281060393dd397a5f17e2ee4f058bf20f785f46a

docs/http/http-exceptions.md
trask: https://github.com/open-telemetry/semantic-conventions/pull/3414/commits/2b1236ffd3fde21184e9971871bea54a9c952520

model/http/events.yaml
trask: https://github.com/open-telemetry/semantic-conventions/commit/281060393dd397a5f17e2ee4f058bf20f785f46a

model/http/events.yaml
```diff
@@ -0,0 +1,56 @@
...
+      This event SHOULD be recorded when an exception occurs during HTTP client operations.
+
+      Instrumentations SHOULD set the severity to WARN (severity number 13) when recording this event.
+
+      Some HTTP client frameworks generate artificial exceptions for non-successful HTTP status codes
+      (e.g., 404 Not Found). When possible, instrumentations SHOULD NOT record these artificial
+      exceptions, or SHOULD set the severity to DEBUG (severity number 5).
+
+      Instrumentations MAY provide a configuration option to populate exception events
+      with the attributes captured on the corresponding http client span.
```
lmolkova:
```suggestion
      with the attributes captured on the corresponding HTTP client span.
```

model/http/events.yaml
```diff
@@ -0,0 +1,56 @@
...
+      This event SHOULD be recorded when an exception occurs during HTTP server request processing.
+
+      Instrumentations SHOULD set the severity to ERROR (severity number 17) when recording this event.
+
+      Some HTTP server frameworks generate artificial exceptions for certain HTTP status codes
+      (e.g., 404 Not Found). When possible, instrumentations SHOULD NOT record these artificial
+      exceptions, or SHOULD set the severity to DEBUG (severity number 5).
+
+      Instrumentations MAY provide a configuration option to populate exception events
+      with the attributes captured on the corresponding http server span.
```
lmolkova:
```suggestion
      with the attributes captured on the corresponding HTTP server span.
```

PR #3422, opened 2026-02-10
Add .NET Remoting to RPC spans documentation

lewis800:
This is a PR related to a new dotnet contrib instrumentation [PR](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3829), for .NET Framework Remoting which is a legacy protocol

## Changes

* Adds netframework_remoting as a new rpc.system.name enum value in the RPC registry
* Adds client and server span conventions for .NET Remoting under dotnet-remoting.md

* [X ] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [X ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
* [ X] Links to the prototypes or existing instrumentations (when adding or changing conventions)


model/rpc/spans.yaml
```diff
@@ -425,3 +425,57 @@ groups:
...
+      document for details on how to record span status.
+    extends: rpc
+    span_kind: client
+    attributes:
+      - ref: rpc.response.status_code
+        requirement_level:
+          conditionally_required: if available.
+        brief: The fully-qualified .NET exception type that was thrown by the .NET Remoting call, used as the status code.
+        note: |
+          All non-empty status codes SHOULD be considered errors.
```
thompson-tomo, line 450: We should just error.type and update the description  rather introducing a status code given that this rpc system doesn't have a concept of status codes.

lewis800:
> .NET remoting was deprecated many years ago, it's not supported by any of the supported .NET versions.
> 
> These conventions can be documented next to the instrumentation (if any), but I don't think we should have them in the central semantic convention repo

That's a fair point, and if that's how other legacy Instrumentations have been handled such as WCF and GrpcCore etc, happy to stick with that and close out the PR.

My only ask would be some feedback on the `rpc.system.name` that should be used for this instrumentation

* netframework_remoting < leaning toward this as the final value
* dotnet_remoting
* remoting

matt-hensley:
I vote for `dotnet_remoting`. It's simple and gives enough context for people looking a telemetry.

As a .NET practitioner, I can make an argument for `netfx_remoting`, since the the .NET vs .NET Framework distinction comes up often. However, I think that's only interesting and relevant to people fairly deep in the .NET ecosystem.

lmolkova:
thanks @lewis800 !

Following [system-specific naming](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#system-specific-naming), I think it could be 
- `dotnet.remoting` - it's a sub-component in .NET ecosystem
- `netfx|netframework.remoting` - this is something that old enough not to be considered part of .NET ecosystem 😅 

Either way, there are other `dotnet.*` and there could be more `netfx|netframework.*` things in the future, so it should be a namespace (`.` instead of  `_`).

I'd prefer `dotnet.remoting` (some things in the current dotnet will eventually become legacy, but they will emit telemetry for ages after)

I'm curious if @open-telemetry/semconv-dotnet-approver have preference.

lewis800:
Sold, dotnet.remoting it is!

Thank you all for your time

Cheers,

PR #3424, opened 2026-02-11
Declarative configuration for semantic convention version selection

trask:
Spinning out from https://github.com/open-telemetry/opentelemetry-configuration/pull/519#discussion_r2770162556

Defining the forward-looking declarative configuration superset of `OTEL_SEMCONV_STABILITY_OPT_IN`.

```
instrumentation/development:
  general:
    <domain>:
      semconv:
        version: 1
        experimental: true
        dual_emit: true
```

### DIFF
--- a/docs/configuration/version-selection.md
+++ b/docs/configuration/version-selection.md
@@ -99,9 +99,9 @@ When adding support for a new stable semantic convention version
 support the new conventions under `experimental: true` (even though the
 convention itself is stable) until the implementation is complete and
 validated. Once confident that the new version is fully supported, it
-SHOULD then begin supporting the stable flag (`experimental: false`) for
-that version. This ensures users are not exposed to incomplete or
-changing telemetry under the stable flag.
+SHOULD then begin supporting `experimental: false` for that version. This
+ensures users are not exposed to incomplete or changing telemetry when
+using the stable configuration (`experimental: false`).
 
 ## Relationship with `OTEL_SEMCONV_STABILITY_OPT_IN`
 


docs/configuration/version-selection.md
lmolkova, line 1: nit: how do you feel about starting a new `configuration` directory next to or under `general` - we'll probably have more about config there and `general` is a bit of a mess

It'd be great to add a README there and just point to declarative config docs, I'm pretty sure a lot of semconv readers don't have context and would appreciate some links

trask: yeah, I was struggling with where to land it anyways

ptal

https://github.com/open-telemetry/semantic-conventions/pull/3424/commits/bc2d3649eb04d6ac4fcd3916351e5da6be28ae09

docs/configuration/README.md
lmolkova, line 1: would it make sense to add @open-telemetry/configuration-approvers as co-code-owners for this dir?

docs/configuration/version-selection.md
```diff
@@ -0,0 +1,109 @@
...
+
+The target semantic convention version for this domain (e.g., `1`).
+
+**`experimental`**
+
+When `true`, include development-stage conventions for the selected
+`version`, regardless of whether that version is pre-stable or already
+stable.
+
+**`dup`**
```
mx-psi: I can't think of a better name for this (`emit_previous_and_new_names` sounds terrible) but I feel like it may not be entirely clear what `dup` means without reading the documentation

docs/configuration/version-selection.md
```diff
@@ -0,0 +1,109 @@
...
+- `messaging`
+- `rpc`
+
+The following properties SHOULD be supported:
+
+| Property       | Type    | Required | Default |
+| -------------- | ------- | -------- | ------- |
+| `version`      | integer | Yes      | (none)  |
+| `experimental` | boolean | No       | `false` |
+| `dup`          | boolean | No       | `false` |
```
thompson-tomo: I would much rather see the settings actually reside alongside the definitions which they apply to, rather then a manually maintained page. Happy to put a proposal together which is consistent with https://github.com/thompson-tomo/semantic-conventions/blob/feat/%233395_AddsConfigurationTables/docs/registry/entities/android.md aka #3396

docs/configuration/version-selection.md
trask: some options:

- `dual_emit`
- `emit_previous`
- `dual_emit_previous`

docs/configuration/version-selection.md
trask: you can put together a separate proposal if you'd like, but I don't think that discussion belongs on this PR

docs/configuration/version-selection.md
lmolkova: it's almost like `dup` is a value of version.

`version = dup # or dual` - emit previous and current major versions

trask: this might be confusing when instrumentations in your app support different versions (of course, you're already in a confusing situation then)

thompson-tomo:
As an alternative to this pr I have drafted #3432 which incorporates the relevant config options directly into the snippets that are rendered in the docs rather than creating seperate docs.

This ensures that a reader only has 1 location to look at for docs and avoids ambiguity by making clear what attributes are controlled by the setting.

docs/configuration/version-selection.md
mx-psi: I like `dual_emit` more than `dup`

thompson-tomo: What about deprecated/removed as it is enabling the attributes which have been deprecated/removed since last version to be emitted?

trask: Changed to `dual_emit`

docs/configuration/README.md
lmolkova: chatted with @jack-berg offline and he seems to be supportive of this

trask: https://github.com/open-telemetry/semantic-conventions/pull/3424/commits/a860cc1168024b17bbc2c9f5cef6ce2b79a9bb40

docs/configuration/version-selection.md
```diff
@@ -0,0 +1,109 @@
...
+**`experimental`**
+
+When `true`, include development-stage conventions for the selected
+`version`, regardless of whether that version is pre-stable or already
+stable.
+
+**`dual_emit`**
+
+When `true`, also emit the previous stable major version alongside the
+target version (e.g., `version=2, dual_emit=true` emits both v2 and v1).
```
mx-psi, line 61: I think we need some wording here similar to open-telemetry/opentelemetry-collector/pull/14538 as to what happens when the metric name is kept but the attributes change

docs/configuration/version-selection.md
```diff
@@ -0,0 +1,109 @@
...
+under the stable flag (`experimental: false`). See
+[Semantic Conventions Stability][SemConvStability] for the
+definition of stability guarantees and breaking changes.
+
+When adding support for a new stable semantic convention version
+(e.g., v1) to a stable instrumentation, the instrumentation SHOULD only
+support the new conventions under `experimental: true` (even though the
+convention itself is stable) until the implementation is complete and
+validated. Once confident that the new version is fully supported, it
+SHOULD then begin supporting the stable flag (`experimental: false`) for
```
mx-psi: It's a bit unclear to me what "the stable flag" means here. Are we talking about the ability to use `version` without setting `experimental` to true?

docs/configuration/version-selection.md
trask: https://github.com/open-telemetry/semantic-conventions/pull/3424/commits/60b07e5722ca6ae3145556c3fe8d96dcc6033680

docs/configuration/version-selection.md
trask: https://github.com/open-telemetry/semantic-conventions/pull/3424/commits/c2e0d3c787b213e5424eeb8bfac3d86be9334380

PR #3426, opened 2026-02-11
Define database and RPC exception events

trask:
Following pattern from
- #3414


model/rpc/events.yaml
```diff
@@ -0,0 +1,48 @@
...
+      This event SHOULD be recorded when an exception occurs during RPC client call operations.
+
+      Instrumentations SHOULD set the severity to WARN (severity number 13) when recording this event.
+
+      Instrumentations MAY provide a configuration option to populate exception events
+      with the attributes captured on the corresponding RPC client span.
+    attributes:
+      - ref: exception.type
+        requirement_level:
+          conditionally_required: Required if `exception.message` is not set, recommended otherwise.
```
lmolkova, line 20: I know it came from the original exception event definition and we already used it on HTTP, but does it make sense? 

proposing

```
exception.type - required
exception.message - recommended | required
exception.stacktrace- recommended
```

It's ok to follow up on this separately, I'd also be curious to hear from @pellared @cijothomas if `exception.type` or message could be unavailable or not meaningful in Go / Rust

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs#L92

Rust don't populate exception.type or stacktrace. We do exception.message. so we should be good from this guidance.

lmolkova: thanks @cijothomas!

it probably means that we can't make `exception.type` required currently.

wouldn't it make sense to populate the `exception.type` as well? Assuming it's common to inherit from `std::error::Error`, it  would be interesting

trask:
> Should the docs under `/docs/exceptions/` be updated to include an `exception-events.md` page?

I think that might be [/docs/exceptions/exceptions-logs.md](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-logs.md), though that page is missing some info still, e.g. event name and severity recommendations

model/rpc/events.yaml
lmolkova: I created https://github.com/open-telemetry/semantic-conventions/issues/3431 to track this separately - I think it should be a blocker for stability. Feel free to resolve this comment and merge

PR #3430, opened 2026-02-12
add jvm.file_descriptor.limit

SylvainJuge:
We currently have `jvm.file_descriptor.count` to provide the current open file descriptors as reported by the JVM.

A limit can be configured at the JVM or OS level, and this is exposed in the underlying JMX interface, thus it would be relevant to capture this as a `jvm.file_descriptor.limit` metric to complement the existing `jvm.file_descriptor.count`.


braydonk:
Yep, that's a generic concept that would make sense to add to `process` as well (instrumented via `ulimit` or `/proc/[pid]/limits`)

PR #3435, opened 2026-02-13
Define messaging exception events

trask:
Follow-up to

- #3414 
- #3426

Prototype: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/16049 (specifically [this commit](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/16049/commits/a341ddea4fef6f384a9852e78c378af1d02f6ff3))


docs/messaging/messaging-exceptions.md
```diff
@@ -0,0 +1,96 @@
...
+## Messaging client operation exception
+
+<!-- semconv event.messaging.client.operation.exception -->
+<!-- NOTE: THIS TEXT IS AUTOGENERATED. DO NOT EDIT BY HAND. -->
+<!-- see templates/registry/markdown/snippet.md.j2 -->
+<!-- prettier-ignore-start -->
+
+**Status:** ![Development](https://img.shields.io/badge/-development-blue)
+
+The event name MUST be `messaging.client.operation.exception`.
```
lmolkova: we have both, producer and client operations, I think both should have the same exception name, but we should probably document that it applies to both.

trask: https://github.com/open-telemetry/semantic-conventions/pull/3435/commits/af5c74f538f45daba2b85bfdf5183d012fa3463c

model/messaging/events.yaml
```diff
@@ -0,0 +1,50 @@
...
+    note: >
+      This event SHOULD be recorded when an exception occurs during messaging client operations
+      including both producer operations (e.g., `send`, `schedule`) and consumer operations
+      (e.g., `receive`, `peek`).
+
+      Instrumentations SHOULD set the severity to WARN (severity number 13) when recording this event.
+
+      Instrumentations MAY provide a configuration option to populate exception events
+      with the attributes captured on the corresponding messaging client span.
+    attributes:
```
kamphaus, line 81: Should the `messaging.operation.type` attribute be included to designate whether the exception occurred during a send or receive operation?

trask: From General semconv SIG meeting: Event name would be span "type/identity" that doesn't exist. So makes sense to split this event up into multiple event names based on `messaging.operation.type`.

trask: Split in https://github.com/open-telemetry/semantic-conventions/pull/3435/commits/63d04406af167558164b3f835bdb08ededb85e52

PR #3436, opened 2026-02-13
Define gen_ai exception event for client operations

trask:
Follow-up to

- #3414 
- #3426

Prototype: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/16049 (specifically [this commit](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/16049/commits/a341ddea4fef6f384a9852e78c378af1d02f6ff3))


docs/gen-ai/gen-ai-exceptions.md
```diff
@@ -0,0 +1,96 @@
...
+>
+> This attribute may contain sensitive information.
+
+**[3] `exception.type`:** Required if `exception.message` is not set, recommended otherwise.
+
+<!-- prettier-ignore-end -->
+<!-- END AUTOGENERATED TEXT -->
+<!-- endsemconv -->
+
+## Generative AI server operation exception
```
lmolkova: we don't currently have a definition of a GenAI server span (only metrics), I'd rather not define an event yet and would add the event when/if we add server span. WDYT?

trask: oh, I missed that, just saw the metric, will remove 👍

trask: https://github.com/open-telemetry/semantic-conventions/pull/3436/commits/55282f4ea40393b3b9e0592eb12931c8d7d227fe

lmolkova (approved):

tarekgh:
Q, I recall there were kind of plan to deprecate the span event. is this not the case anymore?

lmolkova:
@tarekgh that's the plan and this and other similar PRs are parts of the plan. This is a log-based event (not a span event). Check out this draft https://github.com/open-telemetry/opentelemetry.io/pull/9342 that summarizes the story and this https://github.com/open-telemetry/semantic-conventions/pull/3311 for semconv

PR #3448, opened 2026-02-17
Definition for a basic end-user app crash event

bidetofevil:
## Changes

Add a new event in the `app` namespace for a crash of an end-user facing app. Further specification of what a crash is defined as are provided in the event definition. There are implementations abound (async: [Embrace native crash impl](https://github.com/embrace-io/embrace-android-sdk/blob/main/embrace-android-instrumentation-crash-ndk/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/crash/ndk/NativeCrashDataSourceImpl.kt), synchronous: [OTel Android JVM crash impl](https://github.com/open-telemetry/opentelemetry-android/tree/main/instrumentation/crash)), but none strictly follow the convention as stated.

* [x ] [CONTRIBUTING.md](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) guidelines followed.
* [x ] Change log entry added, according to the guidelines in [When to add a changelog entry](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).
* [x ] Links to the prototypes or existing instrumentations (when adding or changing conventions)


model/app/events.yaml
```diff
@@ -1,4 +1,59 @@
...
+      If `exception` attributes are used, it indicates that the exception represented
+      by those attributes caused the crash. When used, these attributes SHOULD be
+      usable for debugging without additional data not specified in the event
+      (e.g. stacktrace already deobfuscated). If that is not the case, additional
+      information like `app.build.id` SHOULD be provided to make the data useful for
+      debugging, either as an attribute in this event or derived from the Resource if
+      `app.sync` is not true.
+    attributes:
+      - ref: app.async
+        requirement_level: opt_in
```
thompson-tomo:
```suggestion
```

I think it would be cleaner & simpler if we removed this attribute. I would then focus app.crash on the sync use case including removing the unnecessary attributes. We could then introduce a process.crash to handle the async use case.

model/app/events.yaml
```diff
@@ -1,4 +1,59 @@
...
+      that are normally provided by the SDK Resource. See the definition of
+      `app.async` for more details about how to use them as it pertains to crashes.
+      How the instrumentation will determine whether an instance of a crash has
+      already been reported and how the necessary data will be retrieved is left up
+      to the instrumentation. Providing enough data to dedupe is NOT REQUIRED.
+      If `exception` attributes are used, it indicates that the exception represented
+      by those attributes caused the crash. When used, these attributes SHOULD be
+      usable for debugging without additional data not specified in the event
+      (e.g. stacktrace already deobfuscated). If that is not the case, additional
+      information like `app.build.id` SHOULD be provided to make the data useful for
```
LikeTheSalad:
```suggestion
      information like `app.build_id` SHOULD be provided to make the data useful for
```

model/app/events.yaml
```diff
@@ -1,4 +1,59 @@
...
+      debugging, either as an attribute in this event or derived from the Resource if
+      `app.sync` is not true.
+    attributes:
+      - ref: app.async
+        requirement_level: opt_in
+      - ref: app.build_id
+        requirement_level:
+          conditionally_required: Recommended if `app.async` is set, opt_in otherwise.
+      - ref: app.crash.id
+        requirement_level: required
```
LikeTheSalad: I'm curious why this attribute is required? Based on its description, it seems like it could be used to reference it in other telemetry data. Is that for context purposes, similar to what the trace id value does?

LikeTheSalad (commented):
Thanks @bidetofevil, I think it's a great start! I just have one comment regarding `app.crash.id`, which, to me, is the only attribute that's not too clear what it's for. Apart from that, I think it's good to go.

model/app/events.yaml
bidetofevil: Yeah, this should probably be more "opt_in". It's meant to be a canonical ID for that crashed instance used for deduping and cross-referencing, like if you were getting this data from a source outside the app and don't want to report this more than once. But if the instrumentation is just capturing this live and creating an event for it, it's not strictly necessary.

model/app/events.yaml
bidetofevil: Will fix

model/app/events.yaml
bidetofevil: The async use case is very common, like with native crashes on Android. I think any app crash event needs to include this use case, as it compries a large swath of them.

And the reason it's in an attribute of its own is because while it's most prominent for crashes, other types of async-report telemetry exists, so we should define a framework for it outside of crashes.

thompson-tomo: I have no problem with the use case of async capturing hence the proposal of using app.* events for sync use case which we would eventually associate with an app entity and potentially the process. Then for async we use process.* events which we would eventually associate with a procees entity. This way the async events have access to a reproducible identity etc and we reduce the need for the conditionally attributes.

This gives you a framework of:
- app.* events to describe the inner workings of a process providing user access to an app by including app specific attributes
- process.* events to describe something related to a process running within a system
- system.* events to describe the something related to the system

This continues what we already have for metrics between system & process.

model/app/registry.yaml
```diff
@@ -6,6 +6,40 @@ groups:
...
+          may only be provided by OS after the app has been terminate via a tombstomb, so the only way to report
+          the event is by reading the tombstone the next time the app starts up.
+          If this attribute is provided and its value is `true`, additional attributes that define the state of
+          the app at the time when the event occurred normally taken from the Resource MUST be provided if
+          they differ from the attributes in the Resource of the SDK instance producing the signal.
+          In other words, attributes in the signal override their equivalents in the Resource.
+          Further, any related attributes not specified in the signal MUST be taken from the Resource (e.g.
+          if `service.version` is specified but `service.name` is not, it is assume it will be in the Resource).
+          For telemetry that omits this attribute, it SHOULD be assumed that it was NOT recorded asynchronously.
+      - id: app.crash.id
```
thompson-tomo, line 9: What about

```suggestion
      - id: exception.occurance.id
```
That way it can be better reused for things such as Grouping nested exceptions into a group using a shared attribute.

bidetofevil: A few reasons I want this in the app namespace:

- I want to restrict this to end-user facing apps, which the namespaces does. Crash has a special meaning in this space and I want to keep the semantics
- Exceptions aren't the only causes of crashes -- at least not ones that we can know the details of. 
- There are additional things that will be put on this. When part 2 of this PR comes out with additional details about the crash, it will become more apparent

model/app/events.yaml
bidetofevil: So the async crash may not have to do with the process at all, just a different mode of information-getting and telemetry recording. Making this an attribute that cuts across all events make more sense to me than carving out a special case in another namespace for which other async events have to duplicate as well.

thompson-tomo: I get that and i understand that it is a `different mode of information-getting and telemetry recording` which is why I am thinking about how this telemetry can be associated with entities.

This also reduces the need for so many conditional attributes based on the use case by having dedicated events

model/app/events.yaml
LikeTheSalad: Got it. I agree it shouldn't be required, so thanks for updating it. I also think it could be very useful for cases like grouping crashes, as mentioned by @thompson-tomo in another comment, but I want to make sure I fully understand it first. The description summarised reads that: 

- It's a unique identifier for a crash.
- It could come from outside of the instrumentation.
- It "COULD be meaningful" and referenced elsewhere.
- It can facilitate deduplication.

That info seems too generic for me to get a grasp of its usage. For example, regarding the "uniqueness" of it, does it mean that it must be globally unique or unique for the same or similar type of crash occurrences? If it's the latter, then it sounds like it could be used to group occurrences of the same crash?

Regarding "_It COULD come from a source external to the instrumentation_", can you mention an example of an external source?

What do you mean when you say that its value "_COULD be meaningful_"? For example, meaningful to group similar crashes as mentioned above? Or is it mostly meaningful to avoid exporting the same crash more than once? (in reference to the "facilitate deduplication" point)? Or both?

model/app/events.yaml
LikeTheSalad: It's probably worth mentioning that a "native crash" in this context most likely means a crash that comes from some C/C++ binaries packaged and executed from within the app. So if that's the case, then it makes sense for it to belong in the "app.*" namespace. I'm mentioning it because the term "native" seems to me to be overloaded.

Having said that, I tend to avoid the use of boolean flags unless there's no other option, but right now I don't think I'm aware of all the options we have, especially given that this is the first of a series of PRs, so maybe a few more details/examples could help us here, @bidetofevil?

For example, considering that there are 2 types of crashes, let's call them a "java" one, which is the default people tend to think about when they talk about Android crashes, and the "native" one that I mentioned earlier. How likely is it that it will be needed to set the `async` flag to "true" for a "java" crash? Or setting it to "false" for the "native" one? Also, which other "types of crashes" are you planning to account for in future PRs?

Depending on the answers to those questions, we might consider the idea of replacing the "async" flag with a "type" enum. Or maybe some nested namespaces for the event name? Such as "app.crash.java" and "app.crash.native".

atulgpt: Hi, `native` is indeed overloaded, if you talk to a react native developer who develops the code in ts or js then for them native is java iteself

bidetofevil:
So who can approve this or give me guidance on what is blocking it from approval? @open-telemetry/semconv-client-approvers?

LikeTheSalad:
> So who can approve this or give me guidance on what is blocking it from approval? @open-telemetry/semconv-client-approvers?

There are a couple of pending questions, at least from my side:

- Some regarding the `crash.id` attribute: https://github.com/open-telemetry/semantic-conventions/pull/3448#discussion_r2831948069 
- Some regarding the `async` flag, here: https://github.com/open-telemetry/semantic-conventions/pull/3448#discussion_r2832050252

Can you take a look at those, @bidetofevil ?

bidetofevil:
Oh shoot. Sorry I missed these @LikeTheSalad. I'll take a look.

model/app/events.yaml
bidetofevil: Sorry @LikeTheSalad and @atulgpt  - just getting back to this.

"Native", in this case, means anything (e.g. C++, Rust, etc.) compiled into a shared library file that is "native" to the device architecture that can be called through JNI (Java-Native Interface) and developed using the Android NDK (Native Development Kit). This is why the word native is used (on Android): it's literally what we call it.

But this is just the most common example: this is a boolean flag because it doesn't talk about the specific contents of the crash (e.g. C++ stacktraces), but rather the manner from which the data for the crash is obtained.

JVM crashes are the exception, where you can catch a crash before it happens (i.e. live), and record it synchronously with the same OTel SDK and app instance that is initialized with the CURRENT state of the app when the crash is taking place. Whereas async is the more common case, where you obtain a tombstone of information AFTER the app instance where the crash has taken place died, so you can't just use the reporting OTel SDK instance to provide the details of the app during the crash because it will be different, as the reporting is happening some time after.

Here's an example:

App version 1.1 produces a JVM crash, which is cause with the UncaughtExceptionHandler and written to an OTel event before the app crashes. In the OTLP, it will report that `app.version` of the crash is 1.1 because that's what the Resources say.

But say the same app version produced a C++ crash. The process dies, and the tombstone is written asynchronously to disk before the OTel event can be recorded. Before the app starts up next time, it is upgrade to 1.2, so when it starts up again, `app.version` of the new OTel instance will be 1.2. 

Having this flag will allow us to distinguish between these two types of crashes when they are processed in the backend, not only for crashes, but for other types of telemetry as well.

LikeTheSalad: Thanks for the thorough explanation, @bidetofevil. After going back through the thread, I think the main concern raised earlier is not about whether async crashes are a valid use case, but about whether this is the right design for them.

Your latest comment, combined with the rest of the thread, helped me get a better idea of the overall proposal and the reason for the questions and suggestions on the `async` flag. The issue is that this proposal seems to combine two different use cases into a single event. The question for me is whether `app.crash` should carry a boolean flag plus **5 conditional attributes** that become recommended when it is enabled. That starts to feel like two schemas inside one event, which is harder to reason about.

It's also hard to evaluate confidently at this stage because we haven't implemented this flow in OTel Android yet. So far, we've focused on JVM stacktraces, which are still the most common type of crash Android developers tend to deal with. From my perspective, that makes this feel a bit early to standardize, especially as part of a "basic" crash event that doesn't otherwise need these attributes.

The more I look at it, the more it seems like the async case may need its own event. That would keep `app.crash` (or even something like `app.crash.sync`) simple and focused, while giving the async use case a dedicated schema that can evolve independently.

Let me know if you have concerns with the approach of having a separate event for each use case.

bidetofevil: As discussed at the Android SIG meeting, I'll move the definition of `app.async` to another PR. I've modified the wording of the description to remove references to the attribute, but kept in the description the possibly of the event being logged async and included the opt_in attributes that are relevant in that case.

Let me know if this look OK @LikeTheSalad. We can have a discussion about the async stuff in a different PR :-)

bidetofevil:
I eff the commit history. Will fix this.

bidetofevil:
Reopen with proper rebasing and updates per discussion

model/app/events.yaml
LikeTheSalad: Thank you, Hanson!

LikeTheSalad (approved):
Thanks!

LikeTheSalad:
@bidetofevil, I think you need to regenerate the markdown files to sort the CI check error.

bidetofevil:
Ugh, I thought I checked already. Will try again.

bidetofevil:
Removed `service.name` as an optional attribute for now given that the markdown generated refers to a markdown file that doesn't exist. Having a different service name is really truly an async use case that needs to be called out, though, so removing it for now isn't a bad idea anyway.

breedx-splk (approved):

model/app/events.yaml
```diff
@@ -1,4 +1,51 @@
...
+      - ref: exception.message
+        requirement_level: opt_in
+      - ref: exception.stacktrace
+        requirement_level: opt_in
+      - ref: os.name
+        requirement_level: opt_in
+      - ref: os.version
+        requirement_level: opt_in
+      - ref: service.version
+        requirement_level: opt_in
```
thompson-tomo: Should this also be removed given we removed service.name via https://github.com/open-telemetry/semantic-conventions/commit/ff0fe1ccd69bb51f5a2d4de511f6ad7fbcf9a438 & we already have app.build_id?

fractalwrench (approved):
LGTM

LikeTheSalad:
@lmolkova is there anyone specifically that we should ask for approval?

model/app/events.yaml
bidetofevil: Build IDs aren't often used to track different version. There's also no sem-ver so you can only tell if one is different than the other, whereas version gives you a lot richer information. This is almost always preferred when folks are tagging telemetry from mobile apps as a point of code differentiation.

thompson-tomo: Yes but that meaning/understanding is only possible when you know what it corresponds to. In the case here we already removed service.name hence we don't know what it is for.

bidetofevil: The removal of `service.name` was because it was relatively low value and there's an unrelated issue that prevented the MD from being generated property (it contains text with a broken reference link when ported to this event). But the idea that there could be a different `service.version` still exists even if we don't explicitly say it was from an async recording of the event. In fact, it is still an important attribute - to know a crash came in a version that is different than the one producing the event.

model/app/registry.yaml
```diff
@@ -6,6 +6,18 @@ groups:
...
     attributes:
+      - id: app.crash.id
+        type: string
+        stability: development
+        brief: >
+          A unique identifier representing an instance of an end-user facing app crash.
+        note: |
+          Its value COULD be meaningful and be used as a reference for other telemetry and metadata recorded by
+          the same instrumentation (e.g. it is an ID generated by an external source that captured the crash).
+          It COULD come from a source external to the instrumentation such that it can be used to look up additional
```
lmolkova:
```suggestion
          It MAY come from a source external to the instrumentation such that it can be used to look up additional
```

should it be SHOULD? Where else could it come from? Would instrumentation generate it? let's document it.

model/app/registry.yaml
```diff
@@ -6,6 +6,18 @@ groups:
       Describes attributes related to client-side applications (e.g. web apps or mobile apps).
     stability: development
     attributes:
+      - id: app.crash.id
+        type: string
+        stability: development
+        brief: >
+          A unique identifier representing an instance of an end-user facing app crash.
+        note: |
+          Its value COULD be meaningful and be used as a reference for other telemetry and metadata recorded by
```
lmolkova:
```suggestion
          Its value MAY be meaningful and be used as a reference for telemetry and metadata recorded by
```
?

COULD is not an [RFC verb](https://datatracker.ietf.org/doc/html/rfc2119). 

This is attribute definition applicable to all telemetry items, "other telemetry" does not seem to be relevant here

model/app/events.yaml
```diff
@@ -1,4 +1,51 @@
...
+    brief: |
+      This event represents a termination of a user-facing application due to
+      unrecoverable programming errors such as exceptions or signals that
+      indicate an error has happened at a lower level.
+    note: |
+      This event identifies the crash of an end-user facing app, a specific type
+      of unexpected app termination defined in the `brief` section. It is intended
+      to provide a basic schema that can be used for all crashes, but it existence
+      does not preclude other, more narrow events to be defined with more extensive
+      schemas to handle specific crash scenarios on specific platforms.
```
lmolkova:
```suggestion
```

No need to repeat brief, also no need to mention that there could be other more narrow events in the future.

model/app/events.yaml
```diff
@@ -1,4 +1,51 @@
...
+      Crash events can be produced asynchronously by another SDK instance, so it
+      can contain attributes that are normally provided by the SDK Resource. The
```
lmolkova: does it mean that my backend service can report crash event for my mobile client? what does " another SDK" mean? OTel SDK ?

model/app/events.yaml
```diff
@@ -1,4 +1,51 @@
...
+      If `exception` attributes are used, it indicates that the exception represented
+      by those attributes caused the crash. When used, these attributes SHOULD be
+      usable for debugging without additional data not specified in the event
+      (e.g. stacktrace already deobfuscated). If that is not the case, additional
+      information like `app.build_id` SHOULD be provided to make the data useful for
+      debugging, either derived from the Resource or as attributes in this event if
+      they differ from what is defined there (in the asynchronous case).
+    attributes:
+      - ref: app.build_id
+        requirement_level: opt_in
```
lmolkova: it seems all the attributes except session id are opt-in (i.e. require explicit action to enable) - is this the intent?

model/app/events.yaml
```diff
@@ -1,4 +1,51 @@
...
+    note: |
+      This event identifies the crash of an end-user facing app, a specific type
+      of unexpected app termination defined in the `brief` section. It is intended
+      to provide a basic schema that can be used for all crashes, but it existence
+      does not preclude other, more narrow events to be defined with more extensive
+      schemas to handle specific crash scenarios on specific platforms.
+      Crash events can be produced asynchronously by another SDK instance, so it
+      can contain attributes that are normally provided by the SDK Resource. The
+      appearance of any of those attributes (e.g. app.build_id) means they should
+      be used instead of the ones provided in the Resource.
```
lmolkova: >  appearance of any of those attributes (e.g. app.build_id) means they should
>    be used instead of the ones provided in the Resource.

not sure i understand what it means. Which "those" ? "attributes that are normally provided by the SDK Resource" ? "means they should be used" where? on the event?

Can we leave a note on each affected attribute something like "when reporting crash event on behalf of another SDK instance, set it on the event  itself".

model/app/events.yaml
```diff
@@ -1,4 +1,51 @@
...
+      How the instrumentation will determine whether an instance of a crash has
+      already been reported and how the necessary data will be retrieved is left up
+      to the instrumentation. Providing enough data to dedupe is NOT REQUIRED.
+      If `exception` attributes are used, it indicates that the exception represented
+      by those attributes caused the crash. When used, these attributes SHOULD be
+      usable for debugging without additional data not specified in the event
+      (e.g. stacktrace already deobfuscated). If that is not the case, additional
```
lmolkova: >  When used, these attributes SHOULD be
> usable for debugging without additional data not specified in the event
> (e.g. stacktrace already deobfuscated).

Please leave is as a requirements on exception.stacktrace and if this sentence affects anything else, please leave corresponding notes.

It currently reads as "event instance should be self-sufficient" which is requirement to semconv authors rather than clarification for instrumentation authors or telemetry consumers.

model/app/registry.yaml
bidetofevil: Instrumentation can capture crashes at runtime (uncaught JVM exception) and record the event then. Or it can be processed and sent after the fact from a tombstone that recorded what happened. 

I believe I provided the examples in the description to some of the scope that was cut for now (e.g. `async` attribute) but I can add it back in.

model/app/events.yaml
bidetofevil: The main use case being alluded to here is at the next app launch, when the instrumentation running picks up the tombstone generated by a crashed app instance and record the event. But in theory, yeah, the tombstone can be processed by any number of processes running an OTel SDK instance and produce the same event - even from a completely separate machine in the backend.

I can be more explicit in description this.

model/app/events.yaml
bidetofevil: No, the intention is that the instrumentation is free to omit if it is not applicable (e.g. can be obtained from the Resource). 

Looking at the [docs](https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/), it seems like some should be "recommended" while others are "conditionally required". I'll modify that.

model/app/events.yaml
bidetofevil: I can do that. I just tried to not repeat too much but I'll make this explicit

model/app/events.yaml
bidetofevil: Will do.

bidetofevil:
I've updated per your comments @lmolkova . Please have another look when you have time next to see if it makes sense.

marandaneto (approved):

breedx-splk:
We have 4 approvals on this. Can we please move it forward? 🙏🏻

breedx-splk:
Really hoping for some progress on this. I put a block on https://github.com/open-telemetry/opentelemetry-android/pull/1691 more than a month ago....

It's not like we're trying to even go stable. Just trying to get the game started...

tedsuo (approved):

bidetofevil:
I've resolved all conversations so this should be mergable

model/app/events.yaml
```diff
@@ -1,4 +1,50 @@
...
+      to the instrumentation. Providing enough data to dedupe is NOT REQUIRED.
+    attributes:
+      - ref: app.build_id
+        requirement_level:
+          conditionally_required: If reporter is not the crashing application instance itself and it's available.
+      - ref: app.crash.id
+        requirement_level: recommended
+      - ref: exception.type
+        requirement_level: recommended
+        brief: If this contains obfuscated symbols, `app.build_id` SHOULD be provided so they can be deobfuscated.
```
lmolkova:
```suggestion
        note: If this contains obfuscated symbols, `app.build_id` SHOULD be provided so they can be deobfuscated.
```

this looks like a note rather than brief

model/app/events.yaml
```diff
@@ -1,4 +1,50 @@
...
+      - ref: exception.message
+        requirement_level: recommended
+        brief: If this contains obfuscated symbols, `app.build_id` SHOULD be provided so they can be deobfuscated.
```
lmolkova:
```suggestion
        note: If this contains obfuscated symbols, `app.build_id` SHOULD be provided so they can be deobfuscated.
```

model/app/events.yaml
```diff
@@ -1,4 +1,50 @@
...
+      - ref: exception.stacktrace
+        requirement_level: recommended
+        brief: If this contains obfuscated symbols, `app.build_id` SHOULD be provided so they can be deobfuscated.
```
lmolkova:
```suggestion
        note: If this contains obfuscated symbols, `app.build_id` SHOULD be provided so they can be deobfuscated.
```

model/app/events.yaml
```diff
@@ -1,4 +1,50 @@
...
+      - ref: os.name
+        requirement_level:
+          conditionally_required: If reporter is not the crashing application instance itself.
```
lmolkova: nit:

```suggestion
          conditionally_required: If reporter is not the crashing application instance itself and the OS name is available.
```

I assume it's not always available on the reporter?

(also on other attributes with this condition)

lmolkova (approved):

model/app/events.yaml
bidetofevil: Yeah on the client you can't assume anything, including details of the environment that it's running on.

bidetofevil:
Thanks @lmolkova - I've accepted all of your suggestions. Let me know if there's anything I need to do for this to be merged.

kamphaus:
Please fix the generated markdown files.

bidetofevil:
@kamphaus Done!

kamphaus (approved):

PR #3461, opened 2026-02-19
Update process attribute requirement levels

osullivandonal:
Related to #864

## Changes

- Set `process.pid` and `process.creation.time` as required (process Identity).
- ~~Set `process.executable.name` to required (Description).~~
- Set process descriptive attributes to Optional (`opt_in`).
- Update `docs/resource/process.md` to reflect the new requirement levels. 

This change ensures consistent process identification and description across all platforms.

<del>- The reason that the process identifier entities have been marked as `required` is they are consistently available across all platforms. These provide a solid way to identify a process. </del>
<del> - `process.executable.name` has been made required as this is also available across all platforms. </del>
   <del> - For example:
      - **Linux** this is available from either `/proc/(pid)/comm or /proc/(pid)/status`. 
      - **Windows** executable name via: [QueryFullProcessImageNameW](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-queryfullprocessimagenamew)
      - **Darwin (macOS)**: Available via [`proc_name`](https://developer.apple.com/documentation/kernel/proc_name) (macOS 10.4+).
      - **BSD (FreeBSD)**: Available via [`kinfo_getproc`](https://man.freebsd.org/cgi/man.cgi?query=kinfo_getproc) (the `ki_comm` field in the `kinfo_proc` structure).</del>

<del>[gopsutil](https://github.com/shirou/gopsutil), the tool we use in hostmetricsreceiver scrapers, provides all of the above.</del>

> [!NOTE]
> Initially, the idea was to make `process.executable.name` required as this is available across all platforms. However, it was observed that calling [execve()](https://man7.org/linux/man-pages/man2/execve.2.html) can replace the process at a certain PID, meaning the PID stays the same but other attributes such as `process.executable.name` can change. For this reason, the descriptive attributes for the process have been changed to `optional` and the identity attributes remain required.

> [!IMPORTANT]

### DIFF
--- a/docs/resource/process.md
+++ b/docs/resource/process.md
@@ -9,7 +9,7 @@ linkTitle: Process
 <!-- toc -->
 
 - [Process](#process)
-  - [Selecting process attributes](#selecting-process-attributes)
+  - [Selecting process command attributes](#selecting-process-command-attributes)
 - [Process runtimes](#process-runtimes)
   - [Erlang runtimes](#erlang-runtimes)
   - [Go runtimes](#go-runtimes)


model/process/entities.yaml
```diff
@@ -8,34 +8,30 @@ groups:
...
       - ref: process.executable.name
         role: descriptive
-        requirement_level:
-          conditionally_required: See [Selecting process attributes](/docs/resource/process.md#selecting-process-attributes) for details.
+        requirement_level: opt_in
       - ref: process.executable.path
         role: descriptive
-        requirement_level:
-          conditionally_required: See [Selecting process attributes](/docs/resource/process.md#selecting-process-attributes) for details.
+        requirement_level: opt_in
```
thompson-tomo, line 20: I would rather see these moved to a seperate entity instead of opt-in similar to what we have with process.runtime.

Especially in relation to:

> it was observed that calling execve() can replace the process at a certain PID

Wouldn't the creation time also change? If not couldn't execve also change the runtime hence puts us in the same situation.

model/process/entities.yaml
```diff
@@ -8,34 +8,30 @@ groups:
...
       - ref: process.command_args
         role: descriptive
-        requirement_level:
-          conditionally_required: See [Selecting process attributes](/docs/resource/process.md#selecting-process-attributes) for details.
+        requirement_level: opt_in
       - ref: process.args_count
         role: descriptive
-        requirement_level:
-          conditionally_required: See [Selecting process attributes](/docs/resource/process.md#selecting-process-attributes) for details.
+        requirement_level: opt_in
```
thompson-tomo, line 32: I think we lose the intention/use cases by making it opt-in. Note if the concern is that it can change, that is OK descriptive attributes are allowed to change.

docs/resource/process.md
```diff
@@ -88,7 +76,7 @@ Between `process.command_args` and `process.command_line`, usually `process.comm
 On Windows and other systems where the native format of process commands is a single string,
 `process.command_line` can additionally (or instead) be used.
 
-`process.args_count` can be additional be provided or instead when sanitization of the arguments is not being performed.
+`process.args_count` can be provided as an additional attribute, particularly when the full `process.command_args` cannot be collected due to sanitization or privacy concerns.
```
thompson-tomo: Not sure on the changes the following would be clearer to me.

```suggestion

### Selecting process command attributes

There are 3 attributes (see below) which describe the command used to launch a process and as such atleast 1 should be provided.

- process.command
- process.command_line
- process.command_args

Between process.command_args and process.command_line, process.command_args is the preferred attribute to be used. process.command_line is only preferred when the native format of process commands is a single string. 

In cases where sensitive data can be provided via command line, these values must be sanitised and if not done the process.command attribute must be used instead along with the process.args_count attribute.

```

docs/resource/process.md
```diff
@@ -38,45 +38,33 @@ linkTitle: Process
...
-| Description | [`process.executable.name`](/docs/registry/attributes/process.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` [6] | string | The name of the process executable. On Linux based systems, this SHOULD be set to the base name of the target of `/proc/[pid]/exe`. On Windows, this SHOULD be set to the base name of `GetProcessImageFileNameW`. | `otelcol` |
-| Description | [`process.executable.path`](/docs/registry/attributes/process.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` [7] | string | The full path to the process executable. On Linux based systems, can be set to the target of `proc/[pid]/exe`. On Windows, can be set to the result of `GetProcessImageFileNameW`. | `/usr/bin/cmd/otelcol` |
+| Identity | [`process.creation.time`](/docs/registry/attributes/process.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | The date and time the process was created, in ISO 8601 format. | `2023-11-21T09:25:34.853Z` |
```
thompson-tomo, line 41: Suggest adding a note, if value is not available use current time.

mx-psi: Under what circumstances would the value not be available?

thompson-tomo: So on dotnet the method to get the start time can throw a win32 exception -> https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.starttime?view=net-10.0

I know it should be rare but for me having the fallback of the current time would be preferred to leaving off the entity or leaving off the attribute as we would lose uniqueness.

mx-psi: I would prefer to do this on a separate PR but it sounds reasonable to me to add some sort of alternative if an exception can be thrown

mx-psi (approved):

docs/resource/process.md
thompson-tomo: The issue is by making it now required, we are being forced to address the what happens if we can't get it scenario.

mx-psi: Implementations are being forced to, but we are not forced to specify how they should do it

model/process/entities.yaml
rogercoll: > Wouldn't the creation time also change? If not couldn't execve also change the runtime hence puts us in the same situation.

The creation time would not change, from the kernel’s perspective, the same process structure (PID, start time, etc.) is preserved, only the memory mappings and execution context are replaced. See this test https://gist.github.com/rogercoll/916276a75416231e82f00ebd46e22452

> I would rather see these moved to a seperate entity instead of opt-in similar to what we have with process.runtime.

Sounds reasonable to me, but could you extend on what would be the benefits comparing to being `opt_in`? How will it relate to the process.entity?

docs/resource/process.md
rogercoll: > So on dotnet the method to get the start time can throw a win32 exception -> https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.starttime?view=net-10.0

It looks like this is a win32 implementation detail and the exception will be thrown when the process is not running (an available process should always have a start time). If the process is not running, it shouldn't be possible to create its entity (process does not exist?).

model/process/entities.yaml
thompson-tomo: Thanks for sharing that test.

Some of The benefits I see would be:

- simplicity in usage as we are not going to force users to now be required to enable the individual attributes, it is just a matter of adding the detector
- when we have entity relationships we would have a relationship between the process & process.executable entity just like we would have between process & process.runtime entity.
- we could see all processes using a single executable

Those are just a couple which come to mind

docs/resource/process.md
thompson-tomo: > Implementations are being forced to, but we are not forced to specify how they should do it

But we should be to ensure consistent identity generation across implementations especially given some solutions can cause issues.

> It looks like this is a win32 implementation detail and the exception will be thrown when the process is not running (an available process should always have a start time).

Actually that is InvalidOperationException. The Win32Exception is defined as:

> An error occurred in the call to the Windows function.

docs/resource/process.md
osullivandonal: Thanks, I agree the command attributes guidance could be clearer. A few thoughts on your suggestion:

1. If we're moving `process.executable.*` to a separate entity, then this section would naturally focus on the command attributes, so that makes sense to me.
2. However, saying "at least 1 should be provided" seems to conflict with these being `opt_in`.
3. I'd still like to keep the `execve` context somewhere - it explains why these attributes are modeled this way.

model/process/entities.yaml
osullivandonal: Thanks for the explanation. Since `execve` can change the executable while preserving the PID/creation time, there is an argument for making `process.executable` separate. I just have some clarifying questions:

1. Scope: This would mean creating a new `entity.process.executable` with attributes like `process.executable.name`, `process.executable.path`, etc?
2. What happens to `process.command.*`, I don't think this can be an attribute to `process.executable`?
   -  Is the idea to leave this as an opt in attribute to `process`?
3. Identifying attributes: What would the identifying attributes for `process.executable`? The execution path?
4. Migration path: Is there a concern for existing implementations that already emit these as process attributes?
5. Implementation: For SDK/Collector implementations, would this mean a separate detector that emits `entities.process.executable` or would a single process detector emit both entities?

docs/resource/process.md
osullivandonal: This feels like implementation details territory. This spec defines what the attributes mean and their requirement level, not how to handle platform specific exceptions. 

That said, I understand the concern about consistent identity generation. If we do add guidance, "use current time" seems problematic:

1. It's semantically incorrect
2. If the failure is intermittent, you'd generate inconsistent identities for the same process.

Looking at the .NET [docs](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.starttime?view=net-10.0), InvalidOperationException is thrown when the process isn't running - which isn't a practical concern since you can't create an entity for a
  non-existent process. The Win32Exception is a generic catch-all for Windows API errors. To me these come back to implementation details.

model/process/entities.yaml
thompson-tomo: > Scope: This would mean creating a new entity.process.executable with attributes like process.executable.name, process.executable.path, etc?

Yes

> What happens to process.command.*, I don't think this can be an attribute to process.executable?

Agree remain on process as it is describing the commands used for that process. One executable can be used with many commands.

> What would the identifying attributes for process.executable? The execution path?

As a starting point yes.

> Is there a concern for existing implementations that already emit these as process attributes?

Can't think of any as sdk's currently don't emit the entities the attributes are associated with.

> For SDK/Collector implementations, would this mean a separate detector that emits entities.process.executable or would a single process detector emit both entities?

Up to implementation but would likely follow what is done with process runtime.

docs/resource/process.md
thompson-tomo: See I am not sure why they need to be opt-in, descriptive attributes are allowed to change over the life cycle of the entity and recommended attributes can be excluded if wanted.

docs/resource/process.md
thompson-tomo: > This feels like implementation details territory. This spec defines what the attributes mean and their requirement level, not how to handle platform specific exceptions.

Agree however this is guidance is what to do when a piece of the identity is not available.

> If the failure is intermittent, you'd generate inconsistent identities for the same process.

But why would you be generating the identity again? The resource/entity is constructed at initialisation and then reused.

> InvalidOperationException is thrown when the process isn't running - which isn't a practical concern since you can't create an entity for a
non-existent process.

Agree, in that case exit.

> The Win32Exception is a generic catch-all for Windows API errors.

Agree and in that case you would have a running application.

mx-psi: > But we should be to ensure consistent identity generation across implementations especially given some solutions can cause issues.

My point is that I think this is something that we can deal with on a separate PR.

docs/registry/entities/process.md
```diff
@@ -15,38 +15,26 @@
...
-| Description | [`process.executable.path`](/docs/registry/attributes/process.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Conditionally Required` [7] | string | The full path to the process executable. On Linux based systems, can be set to the target of `proc/[pid]/exe`. On Windows, can be set to the result of `GetProcessImageFileNameW`. | `/usr/bin/cmd/otelcol` |
+| Identity | [`process.creation.time`](/docs/registry/attributes/process.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | string | The date and time the process was created, in ISO 8601 format. | `2023-11-21T09:25:34.853Z` |
+| Identity | [`process.pid`](/docs/registry/attributes/process.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Required` | int | Process identifier (PID). | `1234` |
 | Description | [`process.interactive`](/docs/registry/attributes/process.md) | ![Development](https://img.shields.io/badge/-development-blue) | `Recommended` | boolean | Whether the process is connected to an interactive shell. | |
```
ChrsMark: This is still `Recommended` but the PR description states `Set process descriptive attributes to Optional (opt_in).` If we don't explicitly set the levels the fallback is `Recommended`.

In this, is the intention to make all descriptive attributes `opt_in`? I wonder if we should decide on some that should actually be `Recommended`.  I would set `[process.owner](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/process.md)` as recommended at least given that it's important info and it's the first things you read in tools like `ps aux`. 
I would also recommend [process.command](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/process.md) too.
No strong preference here, I just think we should provide some more guidance here by recommending some more widely available/useful attributes.

model/process/entities.yaml
osullivandonal: Ok, I will take this feedback to the System semconv WG so we can discuss. Thanks for your input.

docs/resource/process.md
osullivandonal: Agreed @mx-psi , I will take this to the System semconv WG to discuss.

docs/registry/entities/process.md
osullivandonal: Good spot, this was an oversight on my part. I have made all the descriptive attributes `opt_in` as these are non identify, with the exception that `process.owner` and `process.command` are now recommended.

model/process/entities.yaml
osullivandonal: The idea here is that all non identifying attributes for process are to be made `optional`, as they can change.

thompson-tomo: But that is exactly the definition of descriptive attributes that they can change, hence we don't need to make them opt-in to support that use case. Keeping them recommended means that a user can opt-out from the attributes.

osullivandonal:
It was decided in the System semconv WG SIG that this PR is to just update the process attribute requirement levels, other work such as moving `process.executable` to its own entity can be discussed after this PR is merged in a new issue. cc. @open-telemetry/semconv-system-approvers

ChrsMark:
@open-telemetry/specs-semconv-maintainers this one is approved by the SIG, please take a look.

